PR #75, opened 2015-10-20
Fix crash when reloading app during Chrome debugging

### LINKED ISSUE #68: Reloading in chrome is broken
If you reload the app while in chrome debugging, you can end up in a situation where you have an old js context sending requests to a server associated with the reloaded context. 

The solution for this issue is to have a sessionID associated with each rpcServer, and to support simultaneous requests to multiple rpcServers both for the current and old remote jscontexts. We can support these on a single port so we only need one webserver instance.

appden:
Added a session ID, that actually tracks the Realm constructor in our RPC server's JS context. This conveniently let us remove the special casing for defaultPath as well as letting us call the Realm constructor with JSObjectCallAsConstructor.

Fixes #68

### DIFF
--- a/lib/realm.js
+++ b/lib/realm.js
@@ -119,12 +119,15 @@ Object.defineProperties(Realm, {
         value: Object.freeze(propTypes),
     },
     defaultPath: {
-        get: rpc.getDefaultPath,
-        set: rpc.setDefaultPath,
+        get: util.getterForProperty('defaultPath'),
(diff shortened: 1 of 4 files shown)


noldidrita:
Maybe it's just in my case, but I'm observing a crash on iOS during chrome debugging, on the latest release (5.0.2) after upgrading from 4.0.0-alpha2 (I believe).

PR #188, opened 2015-12-16
Fix crash caused by accessing invalid ObjectSchema

### LINKED ISSUE #181: Crashes running on device
I pulled in the new updates and was able to get the app running on my iPhone 6. But it crashes around the 2nd or 3rd time the app makes a database call. Let me know if I can provide more crash data.

screen shot 2015-12-08 at 6 48 30 pm

screen shot 2015-12-08 at 6 49 48 pm

appden:
These references would eventually become invalid. The quickest fix is for them to no longer be references. The longer-term fix might be to only store the object type and dynamically retrieve the ObjectSchema only when necessary.

Fixes #181


santhosh77h:
@appden how to solve this issue in v2.23.0 because i don't see the files which are updated by you here.

PR #267, opened 2016-02-19
[WIP] add task to ship debug & release shared object

### LINKED ISSUE #212: We should be supporting both debug and release builds
For now we only do what `installArchives` uses

nhachicha:
fixes #212

### DIFF
--- a/react-native/android/src/main/java/io/realm/react/RealmReactModule.java
+++ b/react-native/android/src/main/java/io/realm/react/RealmReactModule.java
@@ -18,18 +18,22 @@ import java.util.concurrent.CountDownLatch;
 import fi.iki.elonen.NanoHTTPD;
 
 public class RealmReactModule extends ReactContextBaseJavaModule {
+    private static final String DEBUG_SHARED_OBJECT = "realmreact-dbg";
+    private static final String RELEASE_SHARED_OBJECT = "realmreact";
     private static final int DEFAULT_PORT = 8082;
     private static boolean sentAnalytics = false;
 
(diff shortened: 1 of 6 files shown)


timanglade:
@appden What was the issue with this?

react-native/android/src/main/java/io/realm/react/RealmReactModule.java
```diff
@@ -18,19 +18,23 @@
...
-
-    public RealmReactModule(ReactApplicationContext reactContext) {
+    public RealmReactModule(ReactApplicationContext reactContext, boolean debug) {
         super(reactContext);
 
+        if (debug) {
+            SoLoader.loadLibrary(DEBUG_SHARED_OBJECT);
+        } else {
+            SoLoader.loadLibrary(RELEASE_SHARED_OBJECT);
+        }
```
nhachicha: you can't use reflection, because you don't know the package name of the app yet.

appden:
Sorry I neglected to respond to this last week. Here are some issues/concerns I'm hoping we can resolve before merging this PR:
- I believe this change would make us always embed _both_ the release and debug shared libraries into every application
  - **Possible solution**: Use Gradle build configurations to decide which library to embed
- I believe the debug configuration is still stripping symbols from the build product
  - **Solution**: Add `cmd-strip :=` override to stop stripping symbols (demonstrated in `sk-android-build-release` branch)

appden:
@nhachicha what do you think about my comments above?

Issue #235, opened 2016-02-11
`Symbol.iterator` should be installed on Results


alazier:
So we can do `for(var obj of results)`

PR #273, opened 2016-02-22
[WIP] Steal Array.prototype methods for List and Results

### LINKED ISSUE #18: Support all applicable JS Array methods on List and Results
From @appden :

> I'd like to modify the array and results implementations to properly inherit from the native JS Array so we can leverage all of its methods. To do this, we just need to implement getters (and setters in the case of Array) for indices and the "length" property. JavaScriptCore will handle everything if we just do that (I believe the spec enforces this behavior, but I'm not quite sure of the details). Its native class name should be "Array" as well since many libraries expect this to check if an object is an array.

alazier:
@appden What other methods are essential? Adding all methods would result in code bloat imo.

Fixes #18

### DIFF
--- a/src/js_util.cpp
+++ b/src/js_util.cpp
@@ -97,3 +97,23 @@ bool RJSIsValueDate(JSContextRef ctx, JSValueRef value) {
     static JSStringRef dateString = JSStringCreateWithUTF8CString("Date");
     return RJSIsValueObjectOfType(ctx, value, dateString);
 }
+
+void RJSCopyFunctionsFromPrototype(JSContextRef ctx, JSObjectRef srcObject, JSObjectRef destObject, std::vector<const char *> names) {
+    JSValueRef destPrototypeValue = JSObjectGetPrototype(ctx, destObject);
+    JSObjectRef destPrototypeObject = RJSValidatedValueToObject(ctx, destPrototypeValue);
+
+    JSValueRef srcPrototypeValue = JSObjectGetPrototype(ctx, srcObject);
+    JSObjectRef srcPrototypeObject = RJSValidatedValueToObject(ctx, srcPrototypeValue);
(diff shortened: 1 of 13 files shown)


src/js_util.cpp
```diff
@@ -97,3 +97,23 @@ bool RJSIsValueDate(JSContextRef ctx, JSValueRef value) {
...
+    JSValueRef exception = NULL;
+    for (auto name : names) {
+        JSStringRef jsName = JSStringCreateWithUTF8CString(name);
+        JSObjectSetProperty(ctx, destPrototypeObject, jsName, RJSValidatedPropertyValue(ctx, srcPrototypeObject, jsName),
+                            kJSPropertyAttributeReadOnly | kJSPropertyAttributeDontDelete | kJSPropertyAttributeDontEnum, &exception);
+        if (exception) {
+            throw RJSException(ctx, exception);
+        }
+        
+        JSStringRelease(jsName);
```
bdash: This leaks if an exception is thrown.

appden:
I think this stealing should be done in JS, so it can be shared between the JSC and Chrome Debugging, along with the future V8 implementation. In order to do that, we'd need to expose `Realm.List` and `Realm.Results` constructors (which should throw if actually called), which would also enable type checking (`object instanceof Realm.List`) as well.

We should steal these from `Array.prototype`:
- `join`
- `slice`
- `forEach`
- `every`
- `some`
- `filter`
- `find`
- `findIndex`
- `map`
- `reduce`
- `reduceRight`
- `entries`
- `keys`
- `values`

Also, we should steal `Symbol.iterator`, which depending on the platform may be polyfilled as `@@iterator` instead. That would resolve #235 as well.

alazier:
Putting on the backburner for the moement

PR #288, opened 2016-02-29
Steal Array.prototype methods for List and Results

appden:
This pretty much steals all non-mutating methods that were possible to take. The stealing is done in JS so it can be compatible with Chrome debugging as well as the future V8 binding. In order to do this, the `Realm.List` and `Realm.Results` constructors were exposed (meaning you can perform `instanceof` type checks now), which will throw if called directly.

This specifically excludes `indexOf`, `lastIndexOf`, and `includes` because those require will require special implementations in another PR.


alazier:
As we discussed it would be preferable to not spread the binding implementation across JS and c. I don't see this would not work in chrome debugging if we implemented this using JSC.

alazier:
I can see why we would need to do the same in chrome to make debugging work as passing callbacks over rpc isn't possible. Still not convinced we shouldn't be stealing the methods in jsc when running on the device.

lib/browser/results.js
```diff
@@ -18,21 +18,21 @@
...
+createMethods(Results.prototype, objectTypes.RESULTS, [
     'filtered',
     'sorted',
     'snapshot',
 ]);
 
-function create(realmId, info) {
-    return util.createList(Results.prototype, realmId, info);
+export function create(realmId, info) {
+    return createList(Results.prototype, realmId, info);
```
alazier: maybe this should named differently since it isn't creating a list

alazier:
Would make this a bit easier to review if the js/require changes were a separate pr. Fine for now but keep in mind for next time.

src/js_init.cpp
```diff
@@ -54,27 +56,36 @@ JSClassRef RJSRealmTypeClass() {
     return JSClassCreate(&realmTypesDefinition);
 }
 
+static JSObjectRef InvalidConstructor(JSContextRef ctx, JSObjectRef constructor, size_t argumentCount, const JSValueRef arguments[], JSValueRef* exception) {
```
alazier: This name is a bit confusing. Maybe call it UncallableConstructor

src/js_util.hpp
```diff
@@ -201,6 +201,14 @@ static inline size_t RJSValidatedListLength(JSContextRef ctx, JSObjectRef object
     return RJSValidatedValueToNumber(ctx, lengthValue);
 }
 
+static inline void RJSObjectSetProperty(JSContextRef ctx, JSObjectRef object, JSStringRef propertyName, JSValueRef value, JSPropertyAttributes attributes = 0) {
```
alazier: Should include the word Validated for naming consistency

alazier:
So other than my comments this looks good. CI is having issues so we should run locally to confirm everything is working.

appden:
All tests pass locally.

PR #289, opened 2016-02-29
Accept constructor in create() and objects() methods

### LINKED ISSUE #233: `create` should accept object name or constructor function

appden:
This will help the linter prevent spelling mistakes and make some code more easily readable.

Resolves #233

### DIFF
--- a/src/js_realm.cpp
+++ b/src/js_realm.cpp
@@ -249,11 +249,28 @@ JSValueRef RealmGetProperty(JSContextRef ctx, JSObjectRef object, JSStringRef pr
     return NULL;
 }
 
+std::string RealmObjectTypeForValue(SharedRealm &realm, JSContextRef ctx, JSValueRef value) {
+    if (JSValueIsObject(ctx, value) && JSObjectIsConstructor(ctx, (JSObjectRef)value)) {
+        JSObjectRef constructor = (JSObjectRef)value;
+
+        for (auto pair : RJSConstructors(realm.get())) {
+            if (pair.second == constructor) {
+                return pair.first;
+            }
+        }
+
+        throw std::runtime_error("Constructor was not registered in the schema for this Realm");
+    }
+
+    return RJSValidatedStringForValue(ctx, value, "objectType");
+}
+
 JSValueRef RealmObjects(JSContextRef ctx, JSObjectRef function, JSObjectRef thisObject, size_t argumentCount, const JSValueRef arguments[], JSValueRef* jsException) {
     try {
-        RJSValidateArgumentCount(1, argumentCount);
-        std::string className = RJSValidatedStringForValue(ctx, arguments[0], "objectType");
+        RJSValidateArgumentCount(argumentCount, 1);
+
         SharedRealm sharedRealm = *RJSGetInternal<SharedRealm *>(thisObject);
+        std::string className = RealmObjectTypeForValue(sharedRealm, ctx, arguments[0]);
         return RJSResultsCreate(ctx, sharedRealm, className);
     }
     catch (std::exception &exp) {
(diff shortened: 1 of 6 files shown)


src/js_realm.cpp
```diff
@@ -249,11 +249,28 @@ JSValueRef RealmGetProperty(JSContextRef ctx, JSObjectRef object, JSStringRef pr
     return NULL;
 }
 
+std::string RealmObjectTypeForValue(SharedRealm &realm, JSContextRef ctx, JSValueRef value) {
```
alazier: Name RJSValidatedObjectTypeForValue for consistency?

src/js_realm.cpp
```diff
@@ -249,11 +249,28 @@ JSValueRef RealmGetProperty(JSContextRef ctx, JSObjectRef object, JSStringRef pr
...
+    if (JSValueIsObject(ctx, value) && JSObjectIsConstructor(ctx, (JSObjectRef)value)) {
```
alazier: When we parse the schema we accept non-constructor functions which have a schema. Here we only accept constructors. Which is correct? The docs should probably explain what types of functions users can pass in...

tests/js/realm-tests.js
```diff
@@ -318,6 +318,13 @@ module.exports = BaseTest.extend({
...
             // Should have been multiplied by 100 in the constructor.
             TestCase.assertEqual(object.intCol, 100);
+
+            // Should be able to create object by passing in constructor.
+            object = realm.create(CustomObject, {intCol: 2});
+            TestCase.assertTrue(object instanceof CustomObject);
+            TestCase.assertTrue(Object.getPrototypeOf(object) == CustomObject.prototype);
+            TestCase.assertEqual(customCreated, 2);
+            TestCase.assertEqual(object.intCol, 200);
         });
```
alazier: We should probably test with constructors which aren't in the schema, and functions which aren't constructors.

tests/js/realm-tests.js
```diff
@@ -325,6 +332,16 @@ module.exports = BaseTest.extend({
                 realm.create('InvalidObject', {intCol: 1});
             });
         });
+
+        // Only the original constructor should be valid.
+        function WrongCustomObject() {}
+        WrongCustomObject.schema = CustomObject.schema;
```
alazier: Ok - looks like you are testing that here - WongCustomObject sounds off though, maybe InvalidCustomObject since it isn't wrong but simply not in the schema?

alazier: Kind of unrelated, but maybe we need to freeze the schema or constructor if a user uses it when opening a realm. Would be weird if a user changed the schema after opening a realm.

tests/js/realm-tests.js
```diff
@@ -397,18 +415,28 @@ module.exports = BaseTest.extend({
             realm.create('PersonObject', {name: 'Alex', age: 12, married: true});
         });
 
-        TestCase.assertThrows(function() { 
+        // Should be able to pass constructor for getting objects.
+        var objects = realm.objects(schemas.PersonObject);
```
alazier: Maybe move this down to the part of the testcase where this data is actually used.

src/js_realm.cpp
appden: `JSObjectIsConstructor` will return `true` if the function can be called with `new`, which basically only excludes arrow functions and native functions, so I'll update the schema parsing code to remove the `JSObjectIsFunction` check.

tests/js/realm-tests.js
appden: Since we don't retain that object, the user is free to do whatever they want to it from my perspective. It would be weird to freeze it.

tests/js/realm-tests.js
appden: This data isn't used at all in this test case. Only for the assert on the next line.

alazier:
change look good. waiting on ci.

alazier:
CI is good to go - only failure is from android with was fixed in the other pr. :+1:

alazier:
I retract the :+1: - update the docs please.

appden:
@alazier I updated the docs

alazier:
After one small fix in the docs :+1:

Issue #226, opened 2016-02-02
Support RNPM


appden:
It's becoming the de facto way for clients to link in React Native plugins. Supporting it shouldn't be too difficult: https://github.com/rnpm/rnpm

appden:
I think we should really add support for this, but I think it would require landing a PR to support embedded frameworks rather than static libraries. It's not too difficult given I already did something very similar in the Cordova branch.

alazier:
I think we should support this at some point but it is lower priority than supporting our current users or building user requested features.

jwhitley:
Until that time, I'll suggest that a prominent warning that rnpm doesn't work (yet) should be added to the getting started docs.  rnpm has gotten so good, that I often assume that projects which don't list it just haven't found out about it yet (which is exactly what has happened on several occasions so far).

Conversely, it's surprising when it fails, and that led me on a merry chase with essentially zero useful messaging that my project setup was borked.

appden:
@jwhitley good idea!

@alazier React Native is merging in `rnpm` to officially become part of the project, so its importance will suddenly rise and we may instead need to send the PR to RN instead.

PR #292, opened 2016-03-01
Use device IP addresses to connect to RPC host

### LINKED ISSUE #276: Improve error message when Chrome debugging does not work
Hi,

I keep getting this error when I try to debug the app in Chrome. 

screen shot 2016-02-23 at 10 56 29 pm

The first time I debug the app on a clean device it runs ok but then after I restart the app or the device it crashes.

Thank You!

### LINKED ISSUE #284: Error: Failed to load 'http://localhost:8082/create_session'
When running the application on dev mode from my phone I get the following error
```
Failed to execute 'send' on 'XMLHttpRequest': Failed to load 'http://localhost:8082/create_session'
```

If I run the same code on the simulator it works fine.

appden:
It still only tries localhost for the simulator, but for devices, we now get a list of possible IP addresses to attempt to connect to before throw a more helpful exception.

Resolves #284 and fixes #276

### DIFF
--- a/lib/browser/index.js
+++ b/lib/browser/index.js
@@ -18,6 +18,7 @@
 
 'use strict';
 
+import { NativeModules } from 'react-native';
 import { keys, propTypes, objectTypes } from './constants';
 import * as lists from './lists';
 import * as objects from './objects';
@@ -25,6 +26,7 @@ import * as results from './results';
 import * as rpc from './rpc';
 import * as util from './util';
 
+const {debugHosts, debugPort} = NativeModules.Realm;
 const listenersKey = Symbol();
 
 rpc.registerTypeConverter(objectTypes.LIST, lists.create);
@@ -162,5 +164,24 @@ Object.defineProperties(Realm, {
     },
 });
 
-// The session ID refers to the Realm constructor object in the RPC server.
-Realm[keys.id] = rpc.createSession();
+for (let i = 0, len = debugHosts.length; i < len; i++) {
+    try {
+        // The session ID refers to the Realm constructor object in the RPC server.
+        Realm[keys.id] = rpc.createSession(debugHosts[i] + ':' + debugPort);
+        break;
+    } catch (e) {
+        // Only throw exception after all hosts have been tried.
+        if (i < len - 1) {
+            continue;
+        }
+
+        // Log the original exception for debugging purposes.
+        console.error(e);
+
+        throw new Error(
+            'Realm failed to connect to the embedded debug server inside the app. ' +
+            'If attempting to use Chrome debugging from a device, ensure the device is ' +
+            'reachable on the same network as this machine.'
+        );
+    }
+}
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -21,12 +21,11 @@
 import * as base64 from './base64';
 import { keys, objectTypes } from './constants';
 
-const DEVICE_HOST = 'localhost:8082';
-
 const {id: idKey, realm: realmKey} = keys;
 const typeConverters = {};
 
 let XMLHttpRequest = global.originalXMLHttpRequest || global.XMLHttpRequest;
+let sessionHost;
 let sessionId;
 
 // Check if XMLHttpRequest has been overridden, and get the native one if that's the case.
@@ -45,8 +44,17 @@ export function registerTypeConverter(type, handler) {
     typeConverters[type] = handler;
 }
 
-export function createSession() {
-    sessionId = sendRequest('create_session');
+export function createSession(host) {
+    let oldHost = sessionHost;
+
+    try {
+        sessionHost = host;
+        sessionId = sendRequest('create_session');
+    } catch (e) {
+        sessionHost = oldHost;
+        throw e;
+    }
+
     return sessionId;
 }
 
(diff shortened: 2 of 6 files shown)


appden:
@alazier will you take a look at this

jwhitley:
FWIW, I'm blocked by #284 trying to evaluate realm-js for our needs.  In my case, I've only tried to run in the iOS Simulator so far.

I decided to give this branch a spin to provide feedback, but it's dying in the simulator at [index.js:29](https://github.com/realm/realm-js/blob/sk-debug-on-device/lib/browser/index.js#L29).  Screencap:

screenshot_3_7_16__2_32_pm

This is the first time I've tried to build Realm from source, and this smells like a build process issue rather than a code issue?  My local Realm build setup sequence follows:

Following the instructions at [Building Realm](https://github.com/realm/realm-js#building-realm), I built the RealmReact target in a clone checked out to this branch.  I modified my app's `package.json` to pull in Realm via:

`"realm": "file:/path/to/realm-js",`

Next, I ran `export SKIP_ANDROID_BUILD=1; npm install realm` in my app's root dir.  Finally, I rebuilt and ran my app with the updated Realm version.

Offhand, it's not apparent what I may have missed, but the bit where `NativeModules.Realm` is apparently `undefined` definitely seems like a smoking gun.  :-/

Pointers (and perhaps improved setup docs for outside contributors...) would be greatly appreciated.

appden:
@jwhitley thanks for the detailed report! Can you confirm that the `RealmReact.framework` is present in the `Embedded Binaries` and `Link Binary with Libraries` sections as describes in the [Getting Started](https://realm.io/docs/react-native/latest/#getting-started) documentation. If it isn't, then please follow those steps and let me know if that fixes things. I will improve the error message for this case.

jwhitley:
@appden, thanks, reviewing those settings fixed it.  I got lazy, only lightly skimmed the getting started steps, due to a bad assumption that [rnpm](https://github.com/rnpm/rnpm) had set me up correctly.  (Only because it's done the right thing infallibly every other time I've used it...)

So far, so good with this branch on the iOS Simulator.  I'll have a go with a development device tomorrow and report back.

appden:
@jwhitley I'm curious to hear how it worked for you on a device. We'd like to support `rnpm`, but will probably need to send them a PR to make it work (you can follow progress here: #226).

lib/browser/index.js
```diff
@@ -162,5 +164,24 @@ Object.defineProperties(Realm, {
...
+        Realm[keys.id] = rpc.createSession(debugHosts[i] + ':' + debugPort);
+        break;
+    } catch (e) {
+        // Only throw exception after all hosts have been tried.
+        if (i < len - 1) {
+            continue;
+        }
+
+        // Log the original exception for debugging purposes.
+        console.error(e);
```
alazier: This only logs the error of the last failure

lib/browser/rpc.js
```diff
@@ -45,8 +44,17 @@ export function registerTypeConverter(type, handler) {
...
 }
 
-export function createSession() {
-    sessionId = sendRequest('create_session');
+export function createSession(host) {
+    let oldHost = sessionHost;
+
+    try {
+        sessionHost = host;
+        sessionId = sendRequest('create_session');
```
alazier: if we pass `sessionHost` as a third argument to `send_request` to override the stored `sessionHost`. If we do this we don't need the try/catch here as we can set the stored `sessionHost` after the call.

react-native/RealmReact.mm
```diff
@@ -138,6 +161,60 @@ - (void)removeListenerForEvent:(NSString *)eventName handler:(RealmReactEventHan
 }
 
 #if DEBUG
+- (NSArray *)getIPAddresses {
```
alazier: would be really nice if we had tests for this.

react-native/android/src/main/java/io/realm/react/RealmReactModule.java
```diff
@@ -58,7 +65,18 @@ public String getName() {
         if (!isContextInjected()) {
             startWebServer();
         }
-        return Collections.EMPTY_MAP;
+
+        List<String> hosts;
```
alazier: Do we always need the following code or only when in debug mode?

lib/browser/index.js
appden: That's the point since it's almost always going to try multiple IP addresses, so we'd just be spamming the console if we logged every error.

lib/browser/rpc.js
appden: Ok, I'll make that change.

react-native/RealmReact.mm
appden: Yea but I don't think that's worth all the effort right now.

react-native/android/src/main/java/io/realm/react/RealmReactModule.java
appden: You're right, we only need just do this in the case where we call `startWebServer()`.

jwhitley:
@appden Sorry for the late response, but this change as merged on `master` is working fine on a device for me.  FWIW, this is with a modified `AppDelegate.m` in my project that injects the current build hosts' hostname in debug builds, instead of just hardcoding `localhost`.  Thus the on-device dev instance is also connected to the hosts' dev-mode packager.  (A change which I've *really* got to submit upstream...)

appden:
@jwhitley thanks for the update! I'd really appreciate you trying to use version `0.11.0-rc` from npm if you get the chance.

wulaizi:
Eh, now 10.0.0-rc.2 this problem is still not resolved

PR #301, opened 2016-03-03
Fix broken iterator methods on Android

appden:
The `ArrayIterator` implementation in the JSC version used on Android was broken, so we implement it ourselves. Also added `Symbol.iterator`, fixing #235.

### DIFF
--- a/tests/js/list-tests.js
+++ b/tests/js/list-tests.js
@@ -567,16 +567,39 @@ module.exports = BaseTest.extend({
             TestCase.assertEqual(list.reduce(function(n, p) {return n + p.age}, 0), 33);
             TestCase.assertEqual(list.reduceRight(function(n, p) {return n + p.age}, 0), 33);
 
-            // Some of these may not be present in every environment.
-            if (list.entries) {
-                TestCase.assertEqual(list.entries().next().value[1].name, 'Ari');
-            }
-            if (list.keys) {
-                TestCase.assertEqual(list.keys().next().value, 0);
-            }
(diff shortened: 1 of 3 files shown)


tests/js/list-tests.js
```diff
@@ -567,16 +567,44 @@ module.exports = BaseTest.extend({
...
-                TestCase.assertEqual(list.keys().next().value, 0);
-            }
-            if (list.values) {
-                TestCase.assertEqual(list.values().next().value.name, 'Ari');
-            }
+            [
+                'entries',
+                'keys',
+                'values',
+                typeof Symbol != 'undefined' && Symbol.iterator, // eslint-disable-line no-undef
```
alazier: im not sure what is going on here.

tests/js/list-tests.js
```diff
@@ -567,16 +567,44 @@ module.exports = BaseTest.extend({
...
+            ].forEach(function(methodName) {
+                if (!methodName) {
```
alazier: why do we need this?

alazier:
Looks like this fixes the android test failure we were seeing. :+1:

tests/js/list-tests.js
appden: It gets `Symbol.iterator` as long as `Symbol` is present

tests/js/list-tests.js
appden: Because `Symbol.iterator` may not exist.

PR #305, opened 2016-03-03
Latest object store changes and core

alazier:


appden:
@alazier CI is still failing, but after my fix, it seems to work fine on my machine. Can you confirm?

PR #308, opened 2016-03-04
Update projects to use React Native 0.21

appden:
Updated to mostly use the latest templates as well.


appden:
@alazier it's green!!!

appden:
@alazier can we merge this?

PR #312, opened 2016-03-06
Don't download/unpack core unnecessarily

alazier:
Check for core symlink before downloading/unpacking.


appden:
@alazier I updated the script and manually tested every code path.

PR #315, opened 2016-03-11
Case insensitive queries

### LINKED ISSUE #295: Support case insensitive queries
Requested by a user. We should probably come up with a syntax better than the [c] used in predicates. Thoughts @bdash @kristiandupont @jpsim @cmelchior ?

alazier:
Closes #295

### DIFF
--- a/src/object-store/src/parser/parser.cpp
+++ b/src/object-store/src/parser/parser.cpp
@@ -66,8 +66,9 @@ struct argument : seq< one< '$' >, must< argument_index > > {};
 
 // expressions and operators
 struct expr : sor< dq_string, sq_string, number, argument, true_value, false_value, key_path > {};
+struct case_insensitive : pegtl_istring_t("[c]"){};
 
-struct eq : sor< two< '=' >, one< '=' > > {};
+struct eq : seq< sor< two< '=' >, one< '=' > >, opt< case_insensitive > >{};
 struct noteq : pegtl::string< '!', '=' > {};
 struct lteq : pegtl::string< '<', '=' > {};
 struct lt : one< '<' > {};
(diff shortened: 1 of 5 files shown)


src/object-store/src/parser/parser.cpp
```diff
@@ -66,8 +66,9 @@ struct argument : seq< one< '$' >, must< argument_index > > {};
 
 // expressions and operators
 struct expr : sor< dq_string, sq_string, number, argument, true_value, false_value, key_path > {};
+struct case_insensitive : pegtl_istring_t("[c]"){};
```
appden: Only lowercase `[c]` should be allowed, like `NSPredicate`

appden: Nvm, they actually do allow `[C]` :smile:

src/object-store/tests/query.json
```diff
@@ -197,10 +197,10 @@
...
         ["QueryThrows", "StringObject", "stringCol CONTAINS $0", 1],
 
-        ["Disabled", "QueryCount", 3, "StringObject", "stringCol ==[c] 'a'"],
-        ["Disabled", "QueryCount", 5, "StringObject", "stringCol BEGINSWITH[c] 'A'"],
-        ["Disabled", "QueryCount", 4, "StringObject", "stringCol ENDSWITH[c] 'c'"],
-        ["Disabled", "QueryCount", 2, "StringObject", "stringCol CONTAINS[c] 'B'"]
+        ["QueryCount", 3, "StringObject", "stringCol ==[c] 'a'"],
+        ["QueryCount", 5, "StringObject", "stringCol BEGINSWITH[c] 'A'"],
+        ["QueryCount", 4, "StringObject", "stringCol ENDSWITH[c] 'c'"],
+        ["QueryCount", 2, "StringObject", "stringCol CONTAINS[c] 'B'"]
```
appden: I believe the existing code allows `[c]` on non-string comparisons and just ignores it. I think we really should be throwing an error in that case, and test here that we do.

alazier: This only applied to the `==` operator as the other operators throw for non-string properties. I don't think we need to thrown an exception in this case.

appden: Ok. I think for correctness sake we should throw in this case, but that's up to you.

appden:
Looks good to merge :+1:

PR #317, opened 2016-03-11
[WIP] Use proper hooks to initialize the JSContext

alazier:
@sk - this is your branch rebased.


alazier:
These new hooks do not work reliably. When refreshing on ios, many times I get the `Missing Realm constructor error`

alazier:
We are also getting test failures on android.

Issue #314, opened 2016-03-08
Android crashes with realm 0.10.0 and react-native


jrichardlai:
Hello,

I am testing on multiple android phone an app using Realm. I noticed that it crashes on some Android phones. Old one and recent ones.

What is the minimum version supported for Android ? Is there any code try catch to put on?

Not sure how to debug this, here is some stacktraces:

Galaxy S6 Edge 5.0.2

```
I/DEBUG   (30122): Revision: '10'
I/DEBUG   (30122): ABI: 'arm'
I/DEBUG   (30122): pid: 7917, tid: 7975, name: mqt_js  >>> xxx.package.app <<<
I/DEBUG   (30122): signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x4604b088
I/System.out(13346): KnoxVpnUidStorageknoxVpnSupported API value returned is false
I/qtaguid (13346): Tagging socket 53 with tag 3000120100000000{805310977,0} uid -1, pid: 13346, getuid(): 10016
I/DEBUG   (30122):     r0 00000000  r1 00000000  r2 e8d296d2  r3 e8d296d2
I/DEBUG   (30122):     r4 dfcf3d10  r5 f3fa3d34  r6 0000000b  r7 4604b089
I/DEBUG   (30122):     r8 f3fa3850  r9 dc519140  sl dc517070  fp daa01744
I/DEBUG   (30122):     ip dfcf3e34  sp daa01720  lr df9ecf54  pc 4604b088  cpsr 600b0030
I/DEBUG   (30122): 
I/DEBUG   (30122): backtrace:
I/DEBUG   (30122):     #00 pc 1344a088  /dev/ashmem/dalvik-main space (deleted)
I/DEBUG   (30122):     #01 pc 0017df50  /data/app/xxx.package.app-1/lib/arm/librealmreact.so
I/DEBUG   (30122):     #02 pc 00016bf7  /data/app/xxx.package.app-1/lib/arm/libreactnativejni.so (facebook::react::JSCExecutor::JSCExecutor(facebook::react::Bridge*, std::string const&)+198)
I/DEBUG   (30122):     #03 pc 00016d59  /data/app/xxx.package.app-1/lib/arm/libreactnativejni.so (facebook::react::JSCExecutorFactory::createJSExecutor(facebook::react::Bridge*)+20)
I/DEBUG   (30122):     #04 pc 0001369d  /data/app/xxx.package.app-1/lib/arm/libreactnativejni.so
[...]

PR #327, opened 2016-03-14
Fix ARM hook into JSGlobalContextCreateInGroup

appden:
@alazier @nhachicha I figured out the cause of the crash. Merging this would allow for us to continue to support React Native v0.20+, which I think would be good for us to do for at least one more release. I do think we should eventually drop this hack in favor of requiring v0.22, but I think it would be kinder to our existing users to wait another release to do that. Also, I think we should merge this even if we do move forward with #317 in case we ever need to revert back to using this hack if React Native stops supporting the new method to get direct access to the `JSGlobalContextRef`. It would be good to have a working version sitting in our git history.

We were loading into the program counter (`PC`), which is not good in THUMB mode, so we now instead load into R3, since it's a temp register not used by this function (it only takes two arguments). Also, when building this module in THUMB mode itself, we needed to clear the ARM instruction cache, which wasn't needed when switching from THUMB to ARM mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Support relative paths when opening realms
 
 ### Bugfixes
+* Fix for crash on Android when initializing the Realm module
 * Automatically forward port 8082 for Android
 * Fix broken iterator methods on Android
 * Don't download or unpack core libraries unnecessarily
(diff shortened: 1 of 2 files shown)


alazier:
You need to update the changelog.

I tried to test on the device I have only to realize I was only able to verify the crash on 22rc which no longer exists. @nhachicha do you have any of the devices with you on which we were seeing the crash?

alazier:
This seems to fix the crash on nabil's device. Lets update the changelog and merge.

alazier:
Closes #314

PR #330, opened 2016-03-14
fixes #329

### LINKED ISSUE #329: Tests failing on x86
After the change introduced in https://github.com/realm/realm-js/pull/304 the test suite is failing on x86 devices (Genymotion)

```
<testsuite name="RealmTests" tests="12" failures="3" timestamp="2016-01-22T14:40:44.874443-05:00">
    <testcase name="testRealmConstructor"/>
    <testcase name="testRealmConstructorPath">
      <error message="">'/data/data/io.realm.react.testapp/files/default.realm' does not equal expected value '/data/data/io.realm.react.testapp/files/test1.realm'</error>
    </testcase>
    <testcase name="testRealmConstructorSchemaVersion">
      <error message="">Realm at path already opened with different schema version.</error>
    </testcase>
    <testcase name="testRealmConstructorSchemaValidation"/>
    <testcase name="testDefaultPath">
      <error message="">should use updated default realm path - '/data/data/io.realm.react.testapp/files/default.realm' does not equal expected value '/data/data/io.realm.react.testapp/files/default2.realm'</error>
    </testcase>
```

nhachicha:
We found this fix with @tgoyne 
we suspect gcc to do an optimisation on x86 causing `config.path` in  `RealmCoordinator::get_coordinator(config.path)->get_realm(std::move(config));` to be empty, causing it to retrieve the default coordinator that points to a different path.

### DIFF
--- a/src/object-store/src/shared_realm.cpp
+++ b/src/object-store/src/shared_realm.cpp
@@ -184,7 +184,8 @@ Group *Realm::read_group()
 
 SharedRealm Realm::get_shared_realm(Config config)
 {
-    return RealmCoordinator::get_coordinator(config.path)->get_realm(std::move(config));
+    auto coordinator = RealmCoordinator::get_coordinator(config.path);
+    return coordinator->get_realm(std::move(config));
 }
 
 void Realm::update_schema(std::unique_ptr<Schema> schema, uint64_t version)


alazier:
Nice job figuring this out. :+1:

PR #336, opened 2016-03-17
List splice method now accepts single argument and is documented

appden:


docs/list.js
```diff
@@ -90,6 +89,19 @@ class List {
     push(...object) {}
 
     /**
+     * Changes the contents of the list by removing objects and/or inserting new objects.
+     * @see {@linkcode https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice Array.prototype.splice}
+     * @param {number} index - The start index. If greater than the length of the list,
+     *   the start index will be set to the length instead. If negative, then the start index
+     *   will be counted from the end of the list (e.g. `list.length - index`).
+     * @param {number} [count=length - start] - The number of objects to remove from the list.
```
alazier: It's not clear to me from the docs what happens if only one argument is passed

alazier: Maybe should be `count=length - index`? Or change the working in the last line to be consistent.

alazier:
Other than inconsistency in the wording in the docs :+1: 

Would be nice to get the chrome/debug tests working more reliably but no need to hold back this pr. Need to figure that out at some point though.

appden:
I clarified it, thanks for pointing that out!

PR #338, opened 2016-03-17
Support getting the schema version from unopened realms

### LINKED ISSUE #332: Expose Schema from Realm instances
Right now this can not be retrieved from an existing Realm.

alazier:
Stopgap until we have more complete migrations.

### DIFF
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -104,6 +104,14 @@ class Realm {
 }
 
 /**
+ * Get the current schema version of the Realm at the given path.
+ * @param {string} path - The path to the file where the
+ *   Realm database is stored.
+ * @returns {number} version of the schema, or `-1` if no Realm exists at `path`.
+ */
+Realm.schemaVersion = function(path) {};
+
+/**
  * The default path where to create and access the Realm file.
  * @type {string}
  */
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -107,9 +107,11 @@ class Realm {
  * Get the current schema version of the Realm at the given path.
  * @param {string} path - The path to the file where the
  *   Realm database is stored.
+ * @param {string} encryptionKey - Optional encryption key.
+ * @throws {Error} When passing an invalid or non-matching encryption key.
  * @returns {number} version of the schema, or `-1` if no Realm exists at `path`.
  */
-Realm.schemaVersion = function(path) {};
+Realm.schemaVersion = function(path, encryptionKey) {};
 
 /**
  * The default path where to create and access the Realm file.
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -107,7 +107,8 @@ class Realm {
  * Get the current schema version of the Realm at the given path.
  * @param {string} path - The path to the file where the
  *   Realm database is stored.
- * @param {string} encryptionKey - Optional encryption key.
+ * @param {ArrayBuffer|ArrayBufferView} [encryptionKey] - Required only when
+ *   accessing encrypted Realms.
  * @throws {Error} When passing an invalid or non-matching encryption key.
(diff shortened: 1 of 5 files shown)


src/js_realm.cpp
```diff
@@ -149,6 +149,13 @@ static bool SetDefaultPath(JSContextRef ctx, JSObjectRef object, JSStringRef pro
     return true;
 }
 
+inline std::string RJSNormalizePath(std::string path) {
```
appden: Shouldn't this be `static` too

tests/js/realm-tests.js
```diff
@@ -117,6 +117,16 @@ module.exports = BaseTest.extend({
         TestCase.assertEqual(Realm.defaultPath, newPath, "defaultPath should have been updated");
     },
 
+    testRealmSchemaVersion: function() {
+        TestCase.assertEqual(Realm.schemaVersion(Realm.defaultPath), 0xFFFFFFFFFFFFFFFF);
```
appden: I think this should either return `-1` or throw. I think my vote is on the former.

alazier: This is the version stored internally and is consistent with all the other bindings. Since realms should be compatible across bindings maybe we should leave it? Could also make it a constants.

appden:
@alazier I added an API doc, which you can adjust as needed. I don't believe this closes #332. Also, it's missing Chrome support, which I can add if you want.

alazier:
#332 was named incorrectly. Fixed it.

I'll fix chrome.

tests/js/realm-tests.js
appden: I don't think anything is stored internally for a Realm that doesn't exist. The other bindings have 64-bit integers, and JS doesn't really. I think `-1` perfectly reflects a `does not exist` state (like with `indexOf`), whereas `0xFFFFFFFFFFFFFFFF` is just totally foreign when it comes to JS APIs.

alazier: Used, not stored. I think you know what I meant.

appden: I didn't exactly know what you meant especially given your sentence about compatibility. :smile: I don't think this API means anything for Realms being compatible across bindings.

Each binding should make sense for its given language. That is why the Objective-C API returns a constant (`RLMNotVersioned`) and the Swift API returns `nil`. Each of those makes sense for their respective languages. For JS, I'd say the appropriate value is definitely `-1`, but `null` is also a possibility, though I'd say it's less good.

appden:
Also, this should support being passed the `encryptionKey` like the other bindings' APIs.

alazier:
Should we take a dictionary or a second argument for encryption?

appden:
I think an optional second argument is the way to go in this case.

src/js_realm.cpp
appden: Still think this should be `static`

appden:
Can you add `encryptionKey` and `@throws` to the API docs

docs/realm.js
```diff
@@ -106,9 +106,11 @@ class Realm {
  * Get the current schema version of the Realm at the given path.
  * @param {string} path - The path to the file where the
  *   Realm database is stored.
+ * @param {string} encryptionKey - Optional encryption key.
```
appden: Change to `[encryptionKey]` so it's properly marked as optional in the docs.

appden: And perhaps the comment should instead say that it's required if the Realm is encrypted.

appden:
After tweaking the `encryptionKey` API doc, looks good to merge. :+1:

docs/realm.js
appden: Oh, and the type should be `{ArrayBuffer|ArrayBufferView}`

PR #339, opened 2016-03-18
Support for indexed properties

### LINKED ISSUE #325: Support indexed properties
Requested by user.

alazier:
@appden 

Closes #325

### DIFF
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -28,7 +28,7 @@ import org.apache.tools.ant.filters.ReplaceTokens
 // We then copy both the downloaded code and our custom makefiles and headers into third-party-ndk.
 // After that we build native code from src/main/jni with module path pointing at third-party-ndk.
 
-ext.coreVersion = "$System.env.REALM_CORE_VERSION"
+ext.coreVersion = "npm --silent run get-core-version".execute().text.trim()
 def currentVersion = "npm --silent run get-version".execute().text.trim()
 def downloadsDir = new File("$projectDir/downloads")
 def jscDownloadDir = new File("$projectDir/src/main/jni/jsc")
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -27,8 +27,8 @@ import org.apache.tools.ant.filters.ReplaceTokens
 // We download various C++ open-source dependencies into downloads.
 // We then copy both the downloaded code and our custom makefiles and headers into third-party-ndk.
 // After that we build native code from src/main/jni with module path pointing at third-party-ndk.
-ext.coreVersion = '0.96.2'
 
+ext.coreVersion = "$System.env.REALM_CORE_VERSION"
 def currentVersion = "npm --silent run get-version".execute().text.trim()
 def downloadsDir = new File("$projectDir/downloads")
 def jscDownloadDir = new File("$projectDir/src/main/jni/jsc")
--- a/scripts/download-core.sh
+++ b/scripts/download-core.sh
@@ -3,7 +3,7 @@
 set -e
 set -o pipefail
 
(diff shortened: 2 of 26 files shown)


scripts/download-core.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 set -o pipefail
 
-: ${REALM_CORE_VERSION:=0.96.2} # set to "current" to always use the current build
+: ${REALM_CORE_VERSION:=0.97.0} # set to "current" to always use the current build
```
bdash: You'll want to skip core v0.97.0. It crashes on 32-bit iOS devices. v0.97.1 and newer have the fix.

alazier: Thanks for the info. We should update object store with the newer version as well (that's where I pulled this from).

src/object-store/src/property.hpp
```diff
@@ -55,6 +55,7 @@ namespace realm {
 
         size_t table_column = -1;
         bool requires_index() const { return is_primary || is_indexed; }
+        bool is_indexable() const { return type == PropertyTypeInt || type == PropertyTypeBool || type == PropertyTypeString || type == PropertyTypeDate; }
```
appden: If the plan is to migrate dates to double fields, then perhaps we shouldn't let them be indexed?

scripts/prepublish.sh
```diff
@@ -2,7 +2,8 @@
 
 set -e
 set -o pipefail
-
+export REALM_CORE_VERSION=`bash ./scripts/download-core.sh --version`
```
appden: Please use `$(...)` syntax and remove the `bash` part

scripts/prepublish.sh
```diff
@@ -2,7 +2,8 @@
...
+echo "Core Version:$REALM_CORE_VERSION"
```
appden: Space after colon :smile:

react-native/android/build.gradle
```diff
@@ -27,8 +27,8 @@ import org.apache.tools.ant.filters.ReplaceTokens
 // We download various C++ open-source dependencies into downloads.
 // We then copy both the downloaded code and our custom makefiles and headers into third-party-ndk.
 // After that we build native code from src/main/jni with module path pointing at third-party-ndk.
-ext.coreVersion = '0.96.2'
 
+ext.coreVersion = "$System.env.REALM_CORE_VERSION"
```
appden: Good idea, but we support calling gradle directly, so I would switch this to add an `npm run` script to `package.json` (like `get-core-version`, which calls `scripts/download-core.sh --version`) and run that just like the line below that sets `currentVersion`

tests/js/realm-tests.js
```diff
@@ -258,6 +258,52 @@ module.exports = BaseTest.extend({
...
+
+        TestCase.assertThrows(function() {
+            IndexedTypes.properties = { dataCol: {type: 'data', indexed: true} }
+            new Realm({schema: [IndexedTypes], path: '4'});
+        });
+
+        // primary key
+        IndexedTypes.primaryKey = 'boolCol';
+        IndexedTypes.properties = { boolCol: {type: 'bool', indexed: true} }
+        new Realm({schema: [IndexedTypes], path: '5'});
```
appden: Assuming this is to test that it does **not** throw an exception, so could you add that to the comment above

appden:
Also, can you update the API docs

src/object-store/src/property.hpp
alazier: You should join the object store repo so you can give this feedback to the commiter of this code.

appden: Haha ok, but if we foresee this changing sometime soon, then perhaps we should skip documenting it?

appden:
Looks good. I'm still wary about indexing dates, but if that's not going to change anytime soon, then we can leave it. :+1:

PR #340, opened 2016-03-20
Make List and Results inherit from Collection

appden:
This also adds documentation for all the `Array` methods that were added to `Collection`.


docs/collection.js
```diff
@@ -0,0 +1,279 @@
...
+     * The number of objects in the results.
+     * @type {number}
+     * @readonly
+     */
+    get length() {}
+
+    /**
+     * Returns new results that represent this collection being filtered by the provided query.
+     * @param {string} query - Query used to filter results.
+     * @param {...any} [arg] - Each subsequent argument is used by the placeholders
```
alazier: Should we instead/also document unordered placeholders? Which is the preferable method?

docs/collection.js
```diff
@@ -0,0 +1,279 @@
...
+     *   (e.g. `$0`, `$1`, `$2`, …) in the query.
+     * @throws {Error} If the query or any other argument passed into this method is invalid.
+     * @returns {Realm.Results} filtered according to the provided query.
+     * @example
+     * let merlots = wines.filtered('varietal == "Merlot" && vintage <= $0', maxYear);
```
alazier: nobody says `varietal`.

docs/collection.js
```diff
@@ -0,0 +1,279 @@
...
+class Collection {
+    /**
+     * The number of objects in the results.
+     * @type {number}
+     * @readonly
+     */
+    get length() {}
+
+    /**
+     * Returns new results that represent this collection being filtered by the provided query.
```
alazier: `results` should be capitalized and perhaps italicized. This goes for all the cases we use the work in this way.

docs/collection.js
```diff
@@ -0,0 +1,279 @@
...
+     *   (e.g. `$0`, `$1`, `$2`, …) in the query.
+     * @throws {Error} If the query or any other argument passed into this method is invalid.
+     * @returns {Realm.Results} filtered according to the provided query.
+     * @example
+     * let merlots = wines.filtered('varietal == "Merlot" && vintage <= $0', maxYear);
+     */
+    filtered(query, ...arg) {}
+
+    /**
+     * Returns new results that represent this collection being sorted by the provided property
```
alazier: `Results`

docs/collection.js
appden: Sure, I'll mention what those look like here

docs/collection.js
appden: That example was there before, but I'm happy to use an alternative if you like. Please suggest one. :smile:

docs/collection.js
appden: Ok will do

docs/collection.js
```diff
@@ -0,0 +1,279 @@
...
+     *   This function takes three arguments:
+     *   - `object` – The current object being processed in the collection.
+     *   - `index` – The index of the object being processed in the collection.
+     *   - `collection` – The collection itself.
+     * @param {object} [thisArg] - The value of `this` when `callback` is called.
+     * @returns {Realm.Object|undefined} if the `callback` did not return `true` for any object
+     *   in the collection.
+     * @since 0.11.0
+     */
+    find(callback, thisArg) {}
```
alazier: I think it's dangerous to have this without having a fast way to do the same thing using our apis. Users can do the same thing with filtered but will use this instead because it is easier without realizing how much slower it is.

docs/collection.js
alazier: `variety`? `type`? or just leave it as is...

docs/collection.js
appden: It's there for parity with `Array`. I don't think it's dangerous because users will just use `Array.prototype.find` or `for` loops instead if they need to. In some cases, you might need to use this over a query because queries can't do everything, and this method returns early after the first one is found.

lib/array-methods.js
```diff
@@ -20,6 +20,19 @@
 
 var arrayPrototype = Array.prototype;
```
alazier: maybe call this file collection-methods?

appden: Ok will do

lib/browser/results.js
```diff
@@ -18,13 +18,11 @@
...
+import Collection, { createCollection } from './collections';
 import { objectTypes } from './constants';
-import { createCollection, createMethods } from './util';
+import { createMethods } from './util';
 
-export class Results {
-    constructor() {
-        throw new TypeError('Illegal constructor');
-    }
+export default class Results extends Collection {
```
alazier: what does default do?

appden: So you can do `import Results from 'results'` rather than `import {Results} from 'results'`. It just the canonical way of exporting the primary object of a module.

alazier:
Everything else seems pretty much fine assuming the code that was moved around was not changed (I did not review that thoroughly). Once my other comments are addressed and ci is passing on the latest commit :+1:

appden:
Addressed all of the PR. All tests passed. :+1:

PR #344, opened 2016-03-22
Fix disabling of RCTDevMenu for tests on iOS 9.3

appden:
After upgrading to Xcode 7.3, the test bundle is being loaded before the rest of the app. We need to ensure +[RCTDevMenu load] is called before we remove it from the set of native modules.


appden:
@alazier all tests passed, will you take a look at this

alazier:
Looks good :+1:

PR #347, opened 2016-03-23
Change RealmReact to static library and support rnpm

appden:
We needed to do this due to the way React Native modules work. Our hacky solution for using a framework ended up falling apart due to it relying on symbols not getting stripped from the main application.

This static library is in its own Xcode project because rnpm links in *all* static libraries from a given Xcode project.

The postlink script takes care of linking in libc++ and libz. So now the only setup instructions will be: `rnpm link realm`


alazier:
Please make the rnpm stuff a separate pr. Not sure we need that right now and it will incur more testing which would be nice to not have to do asap.

alazier:
Can we have our other static libraries in subprojects? or is that what we are doing?

alazier:
I guess we can leave the rnmp stuff for now and not push a build until we are ready.

alazier:
I think ReactNative.xcodeproj should be at the root level.

appden:
All of our Android stuff lives under `react-native/android` so I think it makes the most sense to do the same for iOS

scripts/rnpm-postlink.js
```diff
@@ -0,0 +1,37 @@
...
+        var mainGroup = project.getFirstProject().firstProject.mainGroup;
+
+        project.getPBXGroupByKey(mainGroup).children.push({
+            value: group,
+            comment: 'Frameworks',
+        });
+    }
+
+    ['libc++', 'libz'].forEach(function(name) {
+        project.addFramework('usr/lib/' + name + '.tbd', {
```
alazier: what if the framework is already there?

alazier:
Where does our framework get added to the user's project?

scripts/rnpm-postlink.js
appden: It's a no-op. I confirmed both in testing and in the code.

appden:
`rnpm` does that automatically for every static library in the project

alazier:
I think once we add a test to make sure `rnpm link realm` works we can :+1:

PR #357, opened 2016-03-29
Remove changes_available implementation that asserts

### LINKED ISSUE #352: Remove `assert(0)` from `changes_available`
This prevents users from using Realm in swift/objc and RN simultaneously. Even without working/proper notifications we should remove the assert.

Would be good to do a quick point release if possible to support the user asking for this

appden:
This gets called when a Realm changes on another thread, but is called on the correct thread on iOS. It doesn't look like it will be called on Android, so we can unofficially support iOS accessing Realms from multiple bindings for now.

Fixes #352

### DIFF
--- a/src/js_realm.cpp
+++ b/src/js_realm.cpp
@@ -36,9 +36,6 @@ using RJSAccessor = realm::NativeAccessor<JSValueRef, JSContextRef>;
 
 class RJSRealmDelegate : public BindingContext {
 public:
-    virtual void changes_available() {
-        assert(0);
-    }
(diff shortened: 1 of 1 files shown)


appden:
The default implementation is empty

alazier:
I thought it was abstract. Looks good then :+1:

alazier:
Still might want to leave it since we need to fill it in later (and possibly leave a comment) but doesn't matter too much.

appden:
We won't need it unless we implement Realms that don't auto-refresh and we want to send out a notification that they should manually refresh themselves.

Issue #283, opened 2016-02-26
Use an implicit empty array as the default for list properties


alazier:
Users shouldn't have to specify an empty list during object creation.

appden:
@alazier This was fixed by merging #369, right?

alazier:
This has been fixed.

PR #369, opened 2016-04-01
Don't require optional or link properties when creating objects

### LINKED ISSUE #367: Missing property value for property name
Hi, I'm having problems with optional properties. When I mark property as optional and then I don't set it up on object being saved it throws exception
Unhandled JS Exception: Missing property value for property name

Here is the code I used for saving:
```javascript
'use strict';

import React, { Component, Text, View } from 'react-native';
import Realm from 'realm';

class Car {}
Car.schema = {
  name: 'Car',
  properties: {
    name: { type: 'string', optional: true },
    make: 'string',
    model: 'string'
  }
}
class TestRealm extends Component {
  render() {
    let realm = new Realm({schema: [Car]});
    let ferrari = new Car();
    ferrari.make = 'Ferrari';
    ferrari.model = 'Enzo';
    realm.write(() => {
      realm.create('Car', {
        make: 'Ferrari',
        model: 'Enzo'
      });
    });
    return (
      <View>
          <Text>Everything went better than expected</Text>
      </View>
    )
  }
}

export default TestRealm;
```

alazier:
Fixes #367 and #283 

@appden

### DIFF
--- a/src/object-store/src/object_accessor.hpp
+++ b/src/object-store/src/object_accessor.hpp
@@ -208,10 +208,12 @@ namespace realm {
             case PropertyTypeArray: {
                 realm::LinkViewRef link_view = m_row.get_linklist(column);
                 link_view->clear();
-                size_t count = Accessor::list_size(ctx, value);
-                for (size_t i = 0; i < count; i++) {
-                    ValueType element = Accessor::list_value_at_index(ctx, value, i);
-                    link_view->add(Accessor::to_object_index(ctx, m_realm, element, property.object_type, try_update));
+                if (!Accessor::is_null(ctx, value)) {
+                    size_t count = Accessor::list_size(ctx, value);
+                    for (size_t i = 0; i < count; i++) {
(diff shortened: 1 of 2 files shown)


src/object-store/src/object_accessor.hpp
```diff
@@ -310,6 +312,9 @@ namespace realm {
                     if (Accessor::has_default_value_for_property(ctx, realm.get(), object_schema, prop.name)) {
                         object.set_property_value_impl(ctx, prop, Accessor::default_value_for_property(ctx, realm.get(), object_schema, prop.name), try_update);
                     }
+                    else if (prop.is_nullable || prop.type == PropertyTypeArray) {
+                        object.set_property_value_impl(ctx, prop, Accessor::null_value(ctx), try_update);
```
appden: I'm just curious, is it necessary to explicitly null these properties?

appden:
Looks good :+1:

PR #376, opened 2016-04-08
Update rnpm version in package.json

alazier:
Users are still hitting the issue with Test in the project name as we are including the offending version of rnpm as a dependency. 

@appden

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "prepublish": "scripts/prepublish.sh"
   },
   "dependencies": {
-    "rnpm": "1.5.2",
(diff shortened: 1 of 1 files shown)


alazier:
@appden - do we need to have rnpm as a dependency since we tell people to install it globally? Removing this would require the need to keep updating the version...

appden:
@alazier we need the dependency to correctly parse project config for our `postlink` script. I'm going to integrate that functionality into `rnpm` this week with a PR to fix rnpm/rnpm#122, which I filed a couple weeks. Then we won't need this dependency.

PR #383, opened 2016-04-11
Basic data migrations

### LINKED ISSUE #333: Support for data migrations
Requested by user http://stackoverflow.com/questions/36022863/how-to-handle-realm-react-native-migrations-and-schemaversion-on-ios

alazier:
Closes #333 

Allows manual copying/renaming of properties without having to do an extra migrations. Design also supports stepwise migrations.

I think we should consider shipping this improvement without full support for property renames or schema access during migrations, as this eliminates the biggest pain points and current limitations.

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -295,7 +295,7 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
                         create_object<T, RealmClass<T>>(ctx, new SharedRealm(old_realm)),
                         create_object<T, RealmClass<T>>(ctx, new SharedRealm(realm))
                     };
-                    Function<T>::call(ctx, migration_function, nullptr, 2, arguments);
+                    Function<T>::call(ctx, migration_function, 2, arguments);
                 };
             }
             
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -245,7 +245,7 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
 }
 
 template<typename T>
-void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {    
+void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {
     static const String path_string = "path";
     static const String schema_string = "schema";
     static const String schema_version_string = "schemaVersion";
@@ -291,7 +291,7 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
             
             ValueType migration_value = Object::get_property(ctx, object, migration_string);
             if (!Value::is_undefined(ctx, migration_value)) {
-                FunctionType migration_function = Value::validated_to_function(ctx, migration_value);
+                FunctionType migration_function = Value::validated_to_function(ctx, migration_value, "migration");
                 config.migration_function = [=](SharedRealm old_realm, SharedRealm realm) {
                     ValueType arguments[2] = {
                         create_object<T, RealmClass<T>>(ctx, new SharedRealm(old_realm)),
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -243,11 +243,12 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
 }
 
 template<typename T>
-void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {
+void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {    
     static const String path_string = "path";
     static const String schema_string = "schema";
     static const String schema_version_string = "schemaVersion";
     static const String encryption_key_string = "encryptionKey";
+    static const String migration_string = "migration";
 
     realm::Realm::Config config;
     typename Schema<T>::ObjectDefaultsMap defaults;
@@ -284,6 +285,18 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
             }
             else {
                 config.schema_version = 0;
+            }
+            
+            ValueType migration_value = Object::get_property(ctx, object, migration_string);
+            if (!Value::is_undefined(ctx, migration_value)) {
+                FunctionType migration_function = Value::validated_to_function(ctx, migration_value);
+                config.migration_function = [=](SharedRealm old_realm, SharedRealm realm) {
+                    ValueType arguments[2] = {
+                        create_object<T, RealmClass<T>>(ctx, new SharedRealm(old_realm)),
+                        create_object<T, RealmClass<T>>(ctx, new SharedRealm(realm))
+                    };
+                    Function<T>::call(ctx, migration_function, nullptr, 2, arguments);
+                };
             }
             
             ValueType encryption_key_value = Object::get_property(ctx, object, encryption_key_string);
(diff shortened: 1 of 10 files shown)


alazier:
@appden ready for review now.

src/js_realm.hpp
```diff
@@ -243,11 +245,12 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
 }
 
 template<typename T>
-void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {
+void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {
```
appden: Accidental whitespace added

src/js_realm.hpp
```diff
@@ -286,6 +289,18 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
                 config.schema_version = 0;
             }
             
+            ValueType migration_value = Object::get_property(ctx, object, migration_string);
+            if (!Value::is_undefined(ctx, migration_value)) {
+                FunctionType migration_function = Value::validated_to_function(ctx, migration_value);
```
appden: Third argument to `migration_value` should be `"migration"` (or you can use `migration_string` if you can add something to coerce it to a `const char*`) for a nicer error message.

src/js_realm.hpp
```diff
@@ -363,6 +378,12 @@ void Realm<T>::get_schema_version(ContextType ctx, ObjectType object, ReturnValu
 }
 
 template<typename T>
+void Realm<T>::get_schema(ContextType ctx, ObjectType object, ReturnValue &return_value) {
+    auto schema = get_internal<T, RealmClass<T>>(object)->get()->config().schema.get();
```
appden: It doesn't really matter, but I don't think the `->get()` part is necessary.

alazier: it is necessary

src/js_schema.hpp
```diff
@@ -218,5 +222,67 @@ realm::Schema Schema<T>::parse_schema(ContextType ctx, ObjectType schema_object,
     return realm::Schema(schema);
 }
 
+template<typename T>
+typename T::Object Schema<T>::object_for_schema(ContextType ctx, const realm::Schema &schema) {
+    ObjectType object = Object::create_array(ctx);
+    uint32_t count = 0;
+    for (auto object_schema : schema) {
```
appden: You could do `auto&` here to avoid a copy.

src/js_schema.hpp
```diff
@@ -218,5 +222,67 @@ realm::Schema Schema<T>::parse_schema(ContextType ctx, ObjectType schema_object,
...
+
+template<typename T>
+typename T::Object Schema<T>::object_for_object_schema(ContextType ctx, const ObjectSchema &object_schema) {
+    ObjectType object = Object::create_empty(ctx);
+
+    static const String name_string = "name";
+    Object::set_property(ctx, object, name_string, Value::from_string(ctx, object_schema.name));
+
+    ObjectType properties = Object::create_empty(ctx);
+    for (auto property : object_schema.properties) {
```
appden: `auto&` here too

tests/js/migration-tests.js
```diff
@@ -0,0 +1,158 @@
...
+                var oldObjects = oldRealm.objects('TestObject');
+                var newObjects = newRealm.objects('TestObject');
+                TestCase.assertEqual(oldObjects.length, 1);
+                TestCase.assertEqual(newObjects.length, 1);
+
+                TestCase.assertEqual(oldObjects[0].prop0, 'stringValue');
+                TestCase.assertEqual(oldObjects[0].prop1, 1);
+                //TestCase.assertThrows(function() { oldObjects[0].renamed; });
+
+                //TestCase.assertThrows(function() { newObjects[0].prop0; });
```
appden: These commented out pieces should probably assert `undefined`

tests/js/migration-tests.js
```diff
@@ -0,0 +1,158 @@
...
+                TestCase.assertEqual(newObjects[0].renamed, '');
```
appden: Do new properties typically have empty default values like this?

alazier: if they aren't optional then they have to.

tests/js/migration-tests.js
```diff
@@ -0,0 +1,158 @@
...
+
+                newObjects[0].renamed = oldObjects[0].prop0;
+            }
+        });
+
+        var objects = realm.objects('TestObject');
+        TestCase.assertEqual(objects.length, 1);
+        TestCase.assertEqual(objects[0].renamed, 'stringValue');
+        TestCase.assertEqual(objects[0].prop1, 1);
+        TestCase.assertThrows(function() { newObjects[0].prop0; });
```
appden: Yikes, we throw on non-existent property values? I probably was never paying attention to that, but that's not good. :smile:

appden: On maybe this throws because `newObjects` is not defined here.

alazier: that would make sense. will fix that

alazier: `objects[0].prop0` is `undefined` in this case which is what we should be testing.

src/js_schema.hpp
```diff
@@ -218,5 +222,67 @@ realm::Schema Schema<T>::parse_schema(ContextType ctx, ObjectType schema_object,
...
+template<typename T>
+typename T::Object Schema<T>::object_for_property(ContextType ctx, const Property &property) {
+    ObjectType object = Object::create_empty(ctx);
+
+    static const String name_string = "name";
+    Object::set_property(ctx, object, name_string, Value::from_string(ctx, property.name));
+
+    static const String type_string = "type";
+    const std::string type = property.type != PropertyTypeArray ? string_for_property_type(property.type) : "list";
+    Object::set_property(ctx, object, type_string, Value::from_string(ctx, type));
```
appden: I would prefer if the `type` be the `objectType` for `type == "object"`, since we encourage that usage in our documentation to avoid confusion caused by `objectType` meaning two possible things.

tests/js/migration-tests.js
```diff
@@ -0,0 +1,158 @@
...
+var TestCase = require('./asserts');
+var Schemas = require('./schemas');
+
+module.exports = BaseTest.extend({
+    testMigrationFunction: function() {
+        var count = 0;
+        function migrationFunction(oldRealm, newRealm) {
+            TestCase.assertEqual(oldRealm.schemaVersion, 0);
+            TestCase.assertEqual(newRealm.schemaVersion, 1);
+            count++;
```
appden: As you modify the `oldRealm` inside of this function? If that would throw, then can we test that it throws (and doesn't crash) here?

tests/js/realm-tests.js
```diff
@@ -348,26 +351,26 @@ module.exports = BaseTest.extend({
         new Realm({schema: [NotIndexed], path: '1'});
 
         TestCase.assertThrows(function() {
-            IndexedTypes.properties = { floatCol: {type: 'float', indexed: true} }
-            new Realm({schema: [IndexedTypes], path: '2'});
+            schemas.IndexedTypes.properties = { floatCol: {type: 'float', indexed: true} }
```
appden: I'm concerned by modifying a shared object like this since it might be shared with other tests. Let's at least create a shallow copy of `schemas.IndexedTypes` at the beginning of this test and use that instead?

src/js_realm.hpp
appden: I get it now, since it's a pointer to a `shared_ptr`

src/js_schema.hpp
alazier: This makes code that is evaluating the schema much nastier - what you suggest would make it very difficult to check if something is an object type.

tests/js/realm-tests.js
alazier: this was an oversight. will just make a new object type.

tests/js/migration-tests.js
alazier: I added this check to the other test.

appden:
Looks good :+1:

Issue #391, opened 2016-04-18
Support checking if objects have been invalidated


alazier:
Right now there is no way to check if an object is valid.

appden:
Closed by merging #395

PR #395, opened 2016-04-19
Add `realm.isValid()` to support checking for deleted objects

### LINKED ISSUE #389: Delete an object in realm cause react native app crash
DOMException: Failed to execute 'send' on 'XMLHttpRequest': Failed to load 'http://localhost:8082/get_property'.
    at Error (native)
    at sendRequest (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:81928:9)
    at Object.getProperty (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:81848:1504)
    at Object.ORD_NO (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:81746:961)
    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true:80873:27
    at Object.changeOrderState (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:80874:19)
    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true:73377:3977
    at task$ (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:73377:1670)
    at tryCatch (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:10263:30)
    at GeneratorFunctionPrototype.invoke [as _invoke] (http://localhost:8081/index.ios.bundle?platform=ios&dev=true:10539:12)

alazier:
Fixes #389 and #391 

@appden

### DIFF
--- a/lib/browser/objects.js
+++ b/lib/browser/objects.js
@@ -23,10 +23,18 @@ import { getterForProperty, setterForProperty } from './util';
 
 const registeredConstructors = {};
 
+export default class RealmObject {
+}
+
+// Non-mutating methods:
+createMethods(RealmObject.prototype, objectTypes.OBJECT, [
+    'isValid',
+]);
+
 export function create(realmId, info) {
     let schema = info.schema;
     let constructor = (registeredConstructors[realmId] || {})[schema.name];
-    let object = constructor ? Object.create(constructor.prototype) : {};
+    let object =  Object.create(constructor ? constructor.prototype : RealmObject.prototype);
 
     object[keys.realm] = realmId;
     object[keys.id] = info.id;
--- a/lib/browser/objects.js
+++ b/lib/browser/objects.js
@@ -19,7 +19,7 @@
 'use strict';
 
 import { keys } from './constants';
-import { getterForProperty, setterForProperty } from './util';
+import { getterForProperty, setterForProperty, createMethods } from './util';
 
 const registeredConstructors = {};
 
--- a/lib/browser/objects.js
+++ b/lib/browser/objects.js
@@ -18,7 +18,7 @@
 
 'use strict';
 
-import { keys } from './constants';
+import { keys, objectTypes } from './constants';
 import { getterForProperty, setterForProperty, createMethods } from './util';
 
 const registeredConstructors = {};
--- a/lib/browser/objects.js
+++ b/lib/browser/objects.js
@@ -31,10 +31,10 @@ createMethods(RealmObject.prototype, objectTypes.OBJECT, [
     'isValid',
 ]);
 
-export function create(realmId, info) {
+export function createObject(realmId, info) {
     let schema = info.schema;
     let constructor = (registeredConstructors[realmId] || {})[schema.name];
-    let object =  Object.create(constructor ? constructor.prototype : RealmObject.prototype);
+    let object = Object.create(constructor ? constructor.prototype : RealmObject.prototype);
 
     object[keys.realm] = realmId;
     object[keys.id] = info.id;
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -138,7 +138,6 @@ class Realm {
     static void create(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void delete_one(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void delete_all(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
-    static void is_object_valid(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void write(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void add_listener(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void remove_listener(ContextType, ObjectType, size_t, const ValueType[], ReturnValue &);
@@ -213,7 +212,6 @@ struct RealmClass : ClassDefinition<T, SharedRealm> {
         {"create", wrap<Realm::create>},
         {"delete", wrap<Realm::delete_one>},
         {"deleteAll", wrap<Realm::delete_all>},
-        {"isValid", wrap<Realm::is_object_valid>},
         {"write", wrap<Realm::write>},
         {"addListener", wrap<Realm::add_listener>},
         {"removeListener", wrap<Realm::remove_listener>},
@@ -233,11 +231,13 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
     FunctionType collection_constructor = ObjectWrap<T, CollectionClass<T>>::create_constructor(ctx);
     FunctionType list_constructor = ObjectWrap<T, ListClass<T>>::create_constructor(ctx);
     FunctionType results_constructor = ObjectWrap<T, ResultsClass<T>>::create_constructor(ctx);
+    FunctionType realm_object_constructor = ObjectWrap<T, RealmObjectClass<T>>::create_constructor(ctx);
 
     PropertyAttributes attributes = PropertyAttributes(ReadOnly | DontEnum | DontDelete);
     Object::set_property(ctx, realm_constructor, "Collection", collection_constructor, attributes);
     Object::set_property(ctx, realm_constructor, "List", list_constructor, attributes);
     Object::set_property(ctx, realm_constructor, "Results", results_constructor, attributes);
+    Object::set_property(ctx, realm_constructor, "RealmObject", realm_object_constructor, attributes);
 
     return realm_constructor;
 }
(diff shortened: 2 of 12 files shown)


src/jsc/jsc_class.hpp
```diff
@@ -276,6 +278,10 @@ inline JSValueRef ObjectWrap<ClassType>::get_property(JSContextRef ctx, JSObject
             // Out-of-bounds index getters should just return undefined in JS.
             return Value::from_undefined(ctx);
         }
+        catch (Results::OutOfBoundsIndexException &) {
+            // Out-of-bounds index getters should just return undefined in JS.
+            return Value::from_undefined(ctx);
+        }
```
appden: This isn't necessary since `OutOfBoundsIndexException` inherits from `out_of_range`. I removed it locally and all tests still pass. I can commit that change if you want.

src/object-store/src/object_accessor.hpp
```diff
@@ -47,6 +47,8 @@ namespace realm {
         SharedRealm realm() { return m_realm; }
         const ObjectSchema &get_object_schema() { return *m_object_schema; }
         Row row() { return m_row; }
+        
+        bool is_valid() { return m_row.is_attached(); }
```
appden: This method can be `const`

src/js_realm.hpp
```diff
@@ -212,6 +213,7 @@ struct RealmClass : ClassDefinition<T, SharedRealm> {
         {"create", wrap<Realm::create>},
         {"delete", wrap<Realm::delete_one>},
         {"deleteAll", wrap<Realm::delete_all>},
+        {"isValid", wrap<Realm::is_object_valid>},
```
appden: This should be a static method: `Realm.isValid()`. That way you don't need a reference to the realm instance to make these checks, and this method doesn't verify that the object passed in is a part of this realm anyways.

appden:
We need API docs too.

src/js_realm.hpp
alazier: All the other methods which operate on objects are not static methods so perhaps we should keep this as is for consistency. I also wouldn't be opposed to also validating that the passed in object is from the given realm.

alazier:
@appden the more i think about this more the more I believe this should be a member method on `Realm.RealmObject`, `Realm.List`, and `Realm.Results` rather than a method on Realm or a Realm instance. Having a method on RealmObject makes the most sense and we only run into issue when users are using custom classes. In those cases things will be cleaner when we require users inherit from RealmObject, in which case they can call the super method if they have their own `is_valid`. Until then we can still support them calling `RealmObject.is_valid` on their custom objects.

src/js_realm.hpp
```diff
@@ -231,11 +231,13 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
     FunctionType collection_constructor = ObjectWrap<T, CollectionClass<T>>::create_constructor(ctx);
     FunctionType list_constructor = ObjectWrap<T, ListClass<T>>::create_constructor(ctx);
     FunctionType results_constructor = ObjectWrap<T, ResultsClass<T>>::create_constructor(ctx);
+    FunctionType realm_object_constructor = ObjectWrap<T, RealmObjectClass<T>>::create_constructor(ctx);
 
     PropertyAttributes attributes = PropertyAttributes(ReadOnly | DontEnum | DontDelete);
     Object::set_property(ctx, realm_constructor, "Collection", collection_constructor, attributes);
     Object::set_property(ctx, realm_constructor, "List", list_constructor, attributes);
     Object::set_property(ctx, realm_constructor, "Results", results_constructor, attributes);
+    Object::set_property(ctx, realm_constructor, "RealmObject", realm_object_constructor, attributes);
```
appden: This should just be `"Object"`, so it's `Realm.Object`

src/js_realm_object.hpp
```diff
@@ -59,9 +61,24 @@ struct RealmObjectClass : ClassDefinition<T, realm::Object> {
         wrap<RealmObject::set_property>,
         wrap<RealmObject::get_property_names>,
     };
+    
+    using ContextType = typename T::Context;
+    using ObjectType = typename T::Object;
+    using ValueType = typename T::Value;
+    using ReturnValue = js::ReturnValue<T>;
```
appden: This stuff appears out of place

src/jsc/jsc_class.hpp
```diff
@@ -23,6 +23,8 @@
 #include "js_class.hpp"
 #include "js_util.hpp"
 
+#include "results.hpp"
+
```
appden: This can be removed.

src/node/node_class.hpp
```diff
@@ -23,6 +23,8 @@
 #include "js_class.hpp"
 #include "js_util.hpp"
 
+#include "results.hpp"
+
```
appden: Remove here too.

src/object-store/src/object_accessor.hpp
```diff
@@ -136,7 +147,7 @@ namespace realm {
     //
     template <typename ValueType, typename ContextType>
     inline void Object::set_property_value(ContextType ctx, std::string prop_name, ValueType value, bool try_update)
-    {
+    {
```
appden: Extra whitespace

src/js_realm_object.hpp
```diff
@@ -46,10 +46,16 @@ class RealmObject {
...
+    
+    static void is_valid(ContextType, ObjectType, size_t, const ValueType [], ReturnValue &);
 };
 
 template<typename T>
 struct RealmObjectClass : ClassDefinition<T, realm::Object> {
+    using ContextType = typename T::Context;
+    using ObjectType = typename T::Object;
+    using ValueType = typename T::Value;
+    using ReturnValue = js::ReturnValue<T>;
```
appden: None of these need to be here

lib/browser/objects.js
```diff
@@ -18,15 +18,23 @@
...
+// Non-mutating methods:
+createMethods(RealmObject.prototype, objectTypes.OBJECT, [
+    'isValid',
+]);
+
 export function create(realmId, info) {
     let schema = info.schema;
     let constructor = (registeredConstructors[realmId] || {})[schema.name];
-    let object = constructor ? Object.create(constructor.prototype) : {};
+    let object =  Object.create(constructor ? constructor.prototype : RealmObject.prototype);
```
appden: Nbd, but extra space was inserted here.

appden:
I pushed a commit that fixes Chrome debug mode since I know you're having some weird issues on your machine running that stuff. If you could address the one remaining comment, then good to merge. :+1:

I will do a separate PR with API doc updates for everything we have done lately.

src/js_realm_object.hpp
alazier: yeah i was going to combine the classes but decided to revisit that later.

Chris-Petty:
Will we have to wait until 0.12.0 before we can use this?

alazier:
@Chris-Petty `0.12.0-rc` is available now. To use this feature you call `object.isValid()` on any fetched object to check if it can still be accessed. If you try this out let us know how it goes.

Chris-Petty:
So running 0.12.0-rc now. I figure that it would pay to reinstantiate the mock database using the ol' schema. Came to that conclusion as using the existing 0.11 DB, Red Screen of Death keep saying that isValid() was an undefined function.
So populating the realm using the same schema in 0.12.0-rc gives RSoD on app start up saying `'Property' must be of type: number`. The schema worked fine with 0.11. Throws the error on the following code (loop is just for making up lots of data):

```
    for (var i = 1; i < 1000; i+=2) {
      let itemOne = realm.create('Item', { // RSOD/traceback gives this line
        id: i + 'I',
        code: 'MI' + i,
        name: 'MockItem' + i,
        defaultPackSize: '100',
        lines: [],
        typeOf: 'Drug',
        department: pharmacyDept,
        description: 'Super good at being an Item',
        category: antidoteSpecficCat,
      });

// Schema associated with above object

class Item {};
Item.schema = {
  name: 'Item',
  primaryKey: 'id',
  properties: {
    id: 'string',
    code: 'string',
    name: 'string',
    defaultPackSize: 'double',
    lines: { type:'list', objectType: 'ItemLine' },
    typeOf: 'string',
    department: 'ItemDepartment',
    description: 'string',
    category: 'ItemCategory'
  }
}
```

appden:
@Chris-Petty It appears your `defaultPackSize` is a `double` but the code above is passing in the string `"100"`. If that's the issue, then obviously our error message is no good because it's supposed to say the name of the actual property, so please let me know if that was it and we'll fix up our error message!

Chris-Petty:
Thanks @appden that fixed it, went through the rest of the instantiating code and changed all the string literals to numbers. Once including a case that was actually meant to be a string, causing `'Property' must be of type: string` unsurprisingly, though still not providing which specific property or line the error was on, just the line where realm.create is located is given in the traceback.

I'm guessing in 0.11 JS was being forgiving and converting the strings to numbers, JS often likes to do magic like that. Something between 0.11 and 0.12 is changing that behaviour (perhaps intentionally, I would agree that it is safer).

Unfortunately once that problem is resolved, I am still getting undefined is not a function when using isValid(). Is the correct use
`realm.isValid(realmObjectToBeChecked);`

appden:
@Chris-Petty You're right, a refactor led to us being more strict with types, which probably is a good thing, but was not intentional. The `isValid()` method is actually on the new `Realm.Object` class, so your `Item` class should extend it like this: `class Item extends Realm.Object {}` and then you can call `realmObjectToBeChecked.isValid()`.

I'll fix our error messages to be more informative. I think we'll keep the extra type safety unless we hear some strong arguments against it. Also, FYI, the final `0.12.0` was released today. :smile:

Chris-Petty:
It doesn't appear [here](https://realm.io/docs/react-native/latest/), though `npm info realm` was saying 0.12.0 was the latest, which did give me some confusion.

appden:
You're right, we'll deploy the updated docs very soon!

Chris-Petty:
`Super expression must either be null or a function, not object.`
traceback gives some babelhelper, probably more useful is the first line of my schema that throws it. I expect every class is going to throw this error due to me not doing it quite right for 0.12.
```
class Item extends Realm.Object {}
Item.schema = {
  name: 'Item',
  primaryKey: 'id',
  properties: {
    id: 'string',
    code: 'string',
  ...
  },
};
```

Maybe I should wait for docs haha. In the mean time have changed all of the schema to just be declared with const syntax. Removed some redundant inheritance that was going on to allow this. Seems to be working as intended! Have not quite tried isValid yet, though.
```
const ItemSchema = {
  name: 'Item',
...
```

Chris-Petty:
isValid works 🎉 🎉 🎉

PR #396, opened 2016-04-20
Update test and example apps to use React Native 0.24.1

appden:
This might make tests run more reliably on iOS due to some internal changes, mainly "Stop the runloop from invalidate instead of dealloc" (99c7de2).

### DIFF
--- a/examples/ReactExample/package.json
+++ b/examples/ReactExample/package.json
@@ -7,7 +7,7 @@
   },
   "dependencies": {
     "react": "^0.14.5",
-    "react-native": "^0.22.0",
+    "react-native": "^0.24.1",
     "realm": "file:../.."
   }
 }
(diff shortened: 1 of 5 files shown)


alazier:
Looks good, but ci is still failing.

appden:
@alazier CI is still being inconsistent on iOS (Android is borked for a different reason on CI right now) and I'm really not sure why. Tests are running very consistently for me locally.

alazier:
Looks like things are blocked/hanging.

alazier:
My guess would be we are waiting for an executor to load which has already be loaded, or something like that.

appden:
Since I can't seem to reproduce locally, I'm going to push up a temporary commit with some extra logging to see what's happening. :smile:

appden:
@alazier I don't think I'm going to have luck getting CI to cooperate before our next release, but I'd really like to get this merged since it fixes the memory leak issue (#397).

alazier:
yeah lets merge it.

PR #405, opened 2016-04-28
Support for readOnly realms

### LINKED ISSUE #392: Support read only realms
Needed to support bundled realm files http://stackoverflow.com/questions/36682944/is-there-a-way-to-pre-package-realm-database-default-realm-through-react-nativ

alazier:
Closes #392

### DIFF
--- a/lib/browser/index.js
+++ b/lib/browser/index.js
@@ -71,6 +71,8 @@ export default class Realm {
 
         [
             'path',
+            'readOnly',
+            'schema',
             'schemaVersion',
         ].forEach((name) => {
             Object.defineProperty(this, name, {get: util.getterForProperty(name)});
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -246,12 +246,6 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
 
 template<typename T>
 void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc, const ValueType arguments[]) {
-    static const String path_string = "path";
-    static const String schema_string = "schema";
-    static const String schema_version_string = "schemaVersion";
-    static const String encryption_key_string = "encryptionKey";
-    static const String migration_string = "migration";
-
     realm::Realm::Config config;
     typename Schema<T>::ObjectDefaultsMap defaults;
     typename Schema<T>::ConstructorMap constructors;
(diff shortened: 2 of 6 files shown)


src/js_realm.hpp
```diff
@@ -267,28 +263,36 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
...
             ValueType version_value = Object::get_property(ctx, object, schema_version_string);
             if (!Value::is_undefined(ctx, version_value)) {
                 config.schema_version = Value::validated_to_number(ctx, version_value, "schemaVersion");
             }
             else {
                 config.schema_version = 0;
             }
-            
+
+	    	static const String migration_string = "migration";
```
appden: Over-indented

tests/js/realm-tests.js
```diff
@@ -119,6 +119,30 @@ module.exports = BaseTest.extend({
...
+        TestCase.assertEqual(realm.readOnly, true);
+
+        TestCase.assertThrows(function() {
+            realm.write(function() {});
+        });
+        realm.close();
+
+        realm = new Realm({readOnly: true});
+        TestCase.assertEqual(realm.schema.length, 1);
+        TestCase.assertEqual(realm.readOnly, true);
```
appden: Does can you re-open a read-only realm without it being read-only, and vice versa?

alazier: you should be able to. readonly realms are not cached

appden: Should we test this?

lib/browser/index.js
```diff
@@ -71,6 +71,8 @@ export default class Realm {
 
         [
             'path',
+            'readOnly',
```
appden: I think `readonly` would be nicer IMHO

alazier: I'm on the fence but all the other bindings use the current capitalization.

tests/js/realm-tests.js
alazier: yeah

lib/browser/index.js
appden: I understand, but I think we should do what feels right for our binding. If you feel strongly we should keep it this way, then that's fine.

appden:
Looks good :+1:

chaodoze:
Can you provide sample code and steps on how to create a pre-packaged read only realm, that can be bundled in react-native? Thanks!

Bogidon:
Does readOnly add performance/size optimizations (like removing locks)? Couldn't find a matching PR in [realm/realm-core](https://github.com/realm/realm-core).

Bogidon:
I think my previous comment was unclear. What I mean to ask is this: is there a way to turn a read-write realm it into a read-only realm and in the process remove any indexing/features required for write support? I think a lot of users of `realm-js` are looking to bundle their apps with pre-built databases, which (on iOS at least) must be copied to read-write directories due to OS restrictions. 

It would certainly be beneficial to trim the DB size as much as possible. I'm currently shipping a gzipped `.realm` to App Stores (with surprisingly good results), but I'm sure that's all that can be done.

PR #406, opened 2016-04-29
Add workspace and clean up xcproject layout

alazier:
@appden


alazier:
This change seems to reproduce the chrome debug test errors we have been seeing at least locally. Will see if the same happens on ci.

alazier:
Oddly enough the chrome tests worked fine on ci - the only failures were the expected test failures.

react-native/ios/RealmReact.xcodeproj/project.pbxproj
```diff
@@ -120,9 +120,9 @@
 		F60690241CA277CC0003FB26 /* Products */ = {
 			isa = PBXGroup;
 			children = (
-				F606902D1CA277CD0003FB26 /* libGCDWebServers.a */,
+				F606902D1CA277CD0003FB26 /* liblibGCDWebServers.a */,
```
appden: This doesn't look right...

react-native/ios/RealmReact.xcodeproj/project.pbxproj
```diff
@@ -345,7 +345,7 @@
 					"$(SRCROOT)/../../tests/react-test-app/node_modules/react-native/React/**",
 					"$(SRCROOT)/../../examples/ReactExample/node_modules/react-native/React/**",
 				);
-				OTHER_LIBTOOLFLAGS = "$(SRCROOT)/../../core/librealm-ios-bitcode.a $(BUILT_PRODUCTS_DIR)/libRealmJS.a $(BUILT_PRODUCTS_DIR)/libGCDWebServers.a";
+				OTHER_LIBTOOLFLAGS = "$(BUILT_PRODUCTS_DIR)/libRealmJS.a $(BUILT_PRODUCTS_DIR)/libGCDWebServers.a";
```
appden: This intentional?

react-native/ios/RealmReact.xcodeproj/project.pbxproj
alazier: Don't think that is used - I renamed it and changed it back so think it is just an artifact - will remove.

react-native/ios/RealmReact.xcodeproj/project.pbxproj
alazier: This is. Core is now linked into `libRealmJS` so that all users of that library don't also need to include it.

appden:
This one looks good :+1:

PR #407, opened 2016-04-29
All tests now pass in Chrome debug mode (including migration)

appden:
Support for synchronous callbacks was added, and adding future support for async callbacks should just be a matter of polling in JS along with setting up a `ALooper`/`CFRunLoop` for the newly dedicated RPC thread.

### DIFF
--- a/lib/browser/util.js
+++ b/lib/browser/util.js
@@ -46,13 +46,13 @@ export function createMethod(type, name, mutates) {
             throw new TypeError(name + ' method was called on an object of the wrong type!');
         }
 
-        let result = rpc.callMethod(realmId, id, name, Array.from(arguments));
-
-        if (mutates) {
-            fireMutationListeners(realmId);
+        try {
+            return rpc.callMethod(realmId, id, name, Array.from(arguments));
+        } finally {
+            if (mutates) {
+                fireMutationListeners(realmId);
+            }
         }
-
-        return result;
     };
 }
 
--- a/src/concurrent_deque.hpp
+++ b/src/concurrent_deque.hpp
@@ -1,0 +1,90 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <condition_variable>
+#include <deque>
+#include <exception>
+#include <mutex>
+
+namespace realm {
+
+class ConcurrentDequeTimeout : public std::exception {
+  public:
+    ConcurrentDequeTimeout() : std::exception() {}
+};
+
+template <typename T>
+class ConcurrentDeque {
+  public:
+    T pop_front(size_t timeout = 0) {
+        std::unique_lock<std::mutex> lock(m_mutex);
+        while (m_deque.empty()) {
+            wait(lock, timeout);
+        }
+        T item = std::move(m_deque.front());
+        m_deque.pop_front();
+        return item;
+    }
+
+    T pop_back(size_t timeout = 0) {
+        std::unique_lock<std::mutex> lock(m_mutex);
+        while (m_deque.empty()) {
+            wait(lock, timeout);
+        }
+        T item = std::move(m_deque.back());
+        m_deque.pop_back();
+        return item;
+    }
+
+    void push_front(T&& item) {
+        std::unique_lock<std::mutex> lock(m_mutex);
+        m_deque.push_front(std::move(item));
+        lock.unlock();
+        m_condition.notify_one();
+    }
+
+    void push_back(T&& item) {
+        std::unique_lock<std::mutex> lock(m_mutex);
+        m_deque.push_back(std::move(item));
+        lock.unlock();
+        m_condition.notify_one();
+    }
+
+    bool empty() {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        return m_deque.empty();
+    }
+
+    void wait(std::unique_lock<std::mutex> &lock, size_t timeout = 0) {
+        if (!timeout) {
+            m_condition.wait(lock);
+        }
+        else if (m_condition.wait_for(lock, std::chrono::milliseconds(timeout)) == std::cv_status::timeout) {
+            throw ConcurrentDequeTimeout();
+        }
+    }
+
+  private:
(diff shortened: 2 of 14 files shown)


src/rpc.hpp
```diff
@@ -28,20 +33,42 @@ class ObjectSchema;
 namespace rpc {
 
 using json = nlohmann::json;
+
 using RPCObjectID = u_int64_t;
 using RPCRequest = std::function<json(const json)>;
 
+class RPCWorker {
```
alazier: some comments would be nice

src/rpc.cpp
```diff
@@ -313,8 +412,12 @@ JSValueRef RPCServer::deserialize_json_value(const json dict) {
         std::string type_string = type.get<std::string>();
 
         if (type_string == RealmObjectTypesFunction) {
-            // FIXME: Make this actually call the function by its id once we need it to.
-            return JSObjectMakeFunction(m_context, NULL, 0, NULL, jsc::String(""), NULL, 1, NULL);
+            RPCObjectID callback_id = value.get<RPCObjectID>();
+            JSObjectRef callback = JSObjectMakeFunctionWithCallback(m_context, nullptr, js::wrap<run_callback>);
+
+            // Curry the first argument to be the callback id.
```
alazier: Curry?

src/concurrent_deque.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace realm {
+
+class ConcurrentDequeTimeout : public std::exception {
+  public:
+    ConcurrentDequeTimeout() : std::exception() {}
+};
+
+template <typename T>
+class ConcurrentDeque {
```
alazier: Is this really necessary? Couldn't we just use dispatch queues and avoid the complexity?

appden: Well we can't use dispatch queues on Android. Even so, this class it pretty essential to the whole thing working. This probably isn't the _only_ way, but it's the best way I could think of. :smile:

src/rpc.cpp
appden: It's pretty standard terminology. Example: https://robots.thoughtbot.com/introduction-to-function-currying-in-swift

src/rpc.hpp
appden: I tried to comment plenty inside the implementation file. You think I should do more here?

src/rpc.cpp
```diff
@@ -173,20 +215,71 @@ RPCServer::~RPCServer() {
...
+        return;
+    }
+
+    // The first argument was curried to be the callback id.
+    RPCObjectID callback_id = jsc::Value::to_number(ctx, arguments[0]);
+    JSObjectRef arguments_array = jsc::Object::create_array(ctx, argc - 1, argc == 1 ? nullptr : arguments + 1);
+    json arguments_json = server->serialize_json_value(arguments_array);
+
+    // The next task on the stack will instruct the JS to run this callback.
+    server->m_worker.add_task([&]() -> json {
```
alazier: why are we passing arguments by reference?

alazier: i guess if this is synchronous it is fine

alazier: but I don't think it is?

src/concurrent_deque.hpp
alazier: good point.

alazier: i was really hoping to not have to review the lock/condition logic 😄

src/rpc.cpp
```diff
@@ -173,20 +215,71 @@ RPCServer::~RPCServer() {
...
-            return {{"error", "Invalid session ID"}};
+            RPCRequest action = m_requests[name];
+            assert(action);
+
+            m_worker.add_task([=] {
+                return action(args);
+            });
         }
+    }
+    else {
```
alazier: Maybe put this condition first so we don't have to indent

src/rpc.cpp
```diff
@@ -43,10 +41,54 @@ static const char * const RealmObjectTypesFunction = "function";
...
+    try {
+        // Use a 10 millisecond timeout to keep this thread unblocked.
+        auto task = m_tasks.pop_back(10);
+        task();
+
+        // Since this can be called recursively, it must be pushed to the front of the queue *after* running the task.
+        m_futures.push_front(task.get_future());
+    }
+    catch (ConcurrentDequeTimeout &) {
+        // We tried.
```
alazier: maybe log a message here?

alazier: or do we expect to fail here?

src/rpc.cpp
```diff
@@ -173,20 +215,71 @@ RPCServer::~RPCServer() {
...
+        return {
+            {"callback", callback_id},
+            {"arguments", arguments_json},
+        };
+    });
+
+    // Wait for the next callback result to come off the result stack.
+    while (server->m_callback_results.empty()) {
+        // This may recursively bring us into another callback, hence the callback results being a stack.
+        server->m_worker.try_run_task();
```
alazier: I'm not sure why we need this - can't we just wait for the worker to finish since it is already running `try_run_task` internally?

src/rpc.cpp
appden: It is executed before leaving this function. I will add a comment that confirms that.

src/rpc.cpp
appden: Will do

src/rpc.cpp
appden: This is an expected thing that happens a lot.

src/rpc.cpp
appden: This function is run on the worker thread, so we need to "spin the run loop" here until the callback result comes of the stack.

alazier: This seems wrong. I would think `try_run_task` should only ever be called from the worker thread. With this code it seems possible the task could be run on this thread instead.

appden: This function is **always** run on the worker thread. So this is where the worker thread is blocked in the middle of executing a JS callback until that callback has actually run in Chrome and sent back a result.

src/rpc.cpp
alazier: Can you explain why? Relying on a timeout for the expected behavior seems like a design flaw or at the minimum suboptimal.

appden: Without it the thread can deadlock, we need the condition to timeout in order to keep the loop going and checking its underlying condition. It's a similar concept to when you use `[NSRunLoop runUntilDate:]`.

lib/browser/util.js
```diff
@@ -46,13 +46,13 @@ export function createMethod(type, name, mutates) {
             throw new TypeError(name + ' method was called on an object of the wrong type!');
         }
 
-        let result = rpc.callMethod(realmId, id, name, Array.from(arguments));
-
-        if (mutates) {
-            fireMutationListeners(realmId);
+        try {
+            return rpc.callMethod(realmId, id, name, Array.from(arguments));
```
alazier: what gets returns if this throws?

appden: Since there's no `catch` here, the exception will propagate.

alazier:
Looks good once we get things passing on ci. Looks like another spurious error so I will try re-running the job.

appden:
I'm working on getting CI tests passing with #396, so I'm going to merge this, then rebase that branch on top of this.

PR #411, opened 2016-05-02
Bump rnpm dependency to 1.6.5

appden:


alazier:
Why not use `"*"` and just always use the latest version?

appden:
We only use `rnpm` internally for its JS API to get project configuration for our `rnpm-postlink` script. Since they can change their JS API at any time, we need to pin to a specific version and test with it. The only reason I'm bumping this is that I got an indication from someone that they tweaked their project discovery logic.

Once rnpm/rnpm#122 if fixed (likely by me submitting a PR), then we can remove `rnpm` and `xcode` dependencies.

alazier:
ok 👍

Issue #130, opened 2015-11-11
Bad error messages when using invalid class


alazier:
We need to improve error messages when we encounter them so that they are actionable. Two bugs which required a debugger for me to decipher:

`var realm = new Realm({schema: [schemas.PersonObject, schemas.DefaultValuesObject]});
`

`var realm = new Realm({schema: [schemas.PersonObject, schemas.DefaultValues]});`

PR #412, opened 2016-05-02
Fix error message when specifying invalid schema

### LINKED ISSUE #253: Error when specifying invalid object type for list propery is wrong
```
        var personListSchema = {
            name: 'PersonList',
            properties: {
                list: {type: 'list', objectType: 'Persoon'}
            }
        };
```
prints out this message:

/Users/ari/src/realm/realm-js/tests/ios/../lib/list-tests.js:454: error: -[ListTests testListFiltered] : Migration is required due to the following errors: 

Should indicate object type is invalid

alazier:
Closes #253 and #130 

Now when specifying an invalid schema it does not tell you that you need a migration (which was confusing and incorrect). It now correctly says that the schema is invalid. For cases where the schema changes the migration message is displayed when needed.

@appden

### DIFF
--- a/src/object-store/src/object_store.cpp
+++ b/src/object-store/src/object_store.cpp
@@ -570,7 +570,7 @@ InvalidNullabilityException::InvalidNullabilityException(std::string const& obje
 MissingObjectTypeException::MissingObjectTypeException(std::string const& object_type, Property const& property) :
     ObjectSchemaPropertyException(object_type, property)
 {
-    m_what = "Property '" + property.name + "' has an invalid type '" + property.object_type + "'.'";
+    m_what = "Property '" + property.name + "' has an invalid type '" + property.object_type + "'.";
 }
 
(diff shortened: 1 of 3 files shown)


src/object-store/src/object_store.cpp
```diff
@@ -560,7 +570,7 @@ InvalidNullabilityException::InvalidNullabilityException(std::string const& obje
 MissingObjectTypeException::MissingObjectTypeException(std::string const& object_type, Property const& property) :
     ObjectSchemaPropertyException(object_type, property)
 {
-    m_what = "Target type '" + property.object_type + "' doesn't exist for property '" + property.name + "'.";
+    m_what = "Property '" + property.name + "' has an invalid type '" + property.object_type + "'.'";
```
appden: The single quote after the period should be removed

appden:
Looks good after fixing the quote. :+1:

PR #413, opened 2016-05-03
Add schema property to API docs and update CHANGELOG

appden:


CHANGELOG.md
```diff
@@ -6,12 +6,18 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Enhancements
 * Support for queries comparing optional properties to `null`
 * `object.isValid()` has been added to enable checking if an object has been deleted
+  - **Note:** If using a custom class for your objects, then they should inherit from `Realm.Object`
```
alazier: not sure this not is necessary. could say you need to inherit to use base class functionality, but this seems like it should be obvious.

appden:
@alazier This new wording work for you?

alazier:
Looks great 👍

PR #414, opened 2016-05-03
[WIP] Support providing functions as default values

appden:
Their return value will be used as the default value. This was easy to add now that Chrome debugging supports synchronous callbacks.

### DIFF
--- a/src/object-store/src/object_accessor.hpp
+++ b/src/object-store/src/object_accessor.hpp
@@ -286,31 +286,32 @@ namespace realm {
             throw MutationOutsideTransactionException("Can only create objects within a transaction.");
         }
 
-        // get or create our accessor
-        bool created;
-
         // try to get existing row if updating
         size_t row_index = realm::not_found;
         realm::TableRef table = ObjectStore::table_for_object_type(realm->read_group(), object_schema.name);
         const Property *primary_prop = object_schema.primary_key_property();
+
         if (primary_prop) {
             // search for existing object based on primary key type
             ValueType primary_value = Accessor::dict_value_for_key(ctx, value, object_schema.primary_key);
-            if (primary_prop->type == PropertyTypeString) {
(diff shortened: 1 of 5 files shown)


alazier:
Has this been requested by a user? Not sure this is highest priority at the moment. Could see how this could be used for auto-increment but beyond that what are the other use cases we are targeting?

appden:
I think it's incredibly useful for auto-incrementing ids (or randomly generated ones), and other similar use cases. I've seen lots of other ORMs have something similar to this. We don't need to it for the next release, but I put it up here since I know it's useful and it was easy to implement now.

src/object-store/src/object_accessor.hpp
```diff
@@ -286,31 +286,32 @@ namespace realm {
...
-                row_index = table->find_first_string(primary_prop->table_column, Accessor::to_string(ctx, primary_value));
-            }
-            else {
-                row_index = table->find_first_int(primary_prop->table_column, Accessor::to_long(ctx, primary_value));
-            }
 
-            if (!try_update && row_index != realm::not_found) {
-                throw DuplicatePrimaryKeyValueException(object_schema.name, *primary_prop,
-                    "Attempting to create an object of type '" + object_schema.name + "' with an exising primary key value.");
+            if (!Accessor::is_null(ctx, primary_value)) {
```
alazier: This is not correct - if primary value is null we still need to get the proper row_index. I also think we prevent null primary keys somewhere else in the codebase. This is supported on other platforms though so perhaps we should support it as well even though I don't think it is particularly useful.

appden: If `primaryKey` is null or undefined here, then that is caught later in the function if it has no default value.

appden:
@alazier I still believe this would be a nice API to have. I changed it to WIP though.

PR #416, opened 2016-05-03
Allow setting of non-persisted properties

### LINKED ISSUE #415: Port over ignoredProperties or allow non-schema properties
If I try to assign a non-schema property to a RealmObject, I currently get an exception in `Object::set_property_value` in file object_accessor.hpp.

This only happens on Release builds and was unexpected. If RealmJS want all properties to be listed in schema, then there should be an ignoredProperties field in schemas. Otherwise, it should not throw a InvalidPropertyException.

alazier:
Fixes #415 

@appden


appden:
:+1:

PR #417, opened 2016-05-04
Fix issues in Node 6

appden:
Had to disable the `join()` test in Node 6 because of a regression in v8. It fails to convert Realm objects to strings even though they have a `toString()` method in this case. It's not critical functionality, but I will file a bug soon about that.

### DIFF
--- a/tests/js/list-tests.js
+++ b/tests/js/list-tests.js
@@ -556,7 +556,7 @@ module.exports = BaseTest.extend({
 
             // A Node 6 regression in v8 causes an error when converting our objects to strings:
             // TypeError: Cannot convert a Symbol value to a string
-            if (typeof process != 'object' || ('' + process.version).indexOf('v6.') < 0) {
+            if (!TestCase.isNode6()) {
                 TestCase.assertEqual(list.join(' '), 'Ari Tim Bjarne');
             }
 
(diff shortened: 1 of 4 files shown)


tests/js/list-tests.js
```diff
@@ -554,7 +554,11 @@ module.exports = BaseTest.extend({
             TestCase.assertEqual(list.slice(1, 3).length, 2);
             TestCase.assertEqual(list.slice(1, 3)[1].age, 12);
 
-            TestCase.assertEqual(list.join(' '), 'Ari Tim Bjarne');
+            // A Node 6 regression in v8 causes an error when converting our objects to strings:
+            // TypeError: Cannot convert a Symbol value to a string
+            if (typeof process != 'object' || ('' + process.version).indexOf('v6.') < 0) {
```
alazier: We should add a utility method in `TestCase` or somewhere else to do this check. `TestCase.platform == 'node'`, `TestCase.isNode()` or something similar. Imagine there will be other platform specific tests in the future.

alazier: Can we make this fix and get this merged?

appden: Made the fix, let me know if it's good to merge

alazier:
👍

PR #418, opened 2016-05-04
Linux support

appden:


appden:
@alazier I'd like to get the merged at some point. It includes some commits for Node v6 support, but I will address your PR feedback in the other branch. It includes fixes for Node on Mac too, like a proper `platform.cpp` file and hooking into the `libuv` runloop.

alazier:
I don't see anything objectionable - we should definitely rebase and get tests running though before merging.

appden:
This branch is based on work that will never be merged into the object-store. A new branch will be created.

PR #419, opened 2016-05-04
Add close() method API doc

appden:


docs/realm.js
```diff
@@ -70,6 +70,12 @@ class Realm {
     constructor(config) {}
 
     /**
+     * Closes this Realm so it may be re-opened with a newer schema version.
```
alazier: this -> the ?

docs/realm.js
```diff
@@ -70,6 +70,12 @@ class Realm {
...
+     * All objects and collections in this Realm will no longer be valid until it is re-opened.
```
alazier: I don't think objects ever become valid again.

docs/realm.js
```diff
@@ -71,7 +71,7 @@ class Realm {
 
     /**
      * Closes this Realm so it may be re-opened with a newer schema version.
-     * All objects and collections in this Realm will no longer be valid until it is re-opened.
+     * All objects and collections from this Realm are no longer valid.
```
alazier: After calling all objects and collections from this Realm are no longer valid

PR #423, opened 2016-05-06
Use new timestamp format

### LINKED ISSUE #355: Storing date as milliseconds is not compatible with other bindings
Because dates are stored as milliseconds, realm files can not currently be shared across languages. We need to normalize to the new standard format.

alazier:
Fixes #355


src/ios/RealmJS.xcodeproj/project.pbxproj
```diff
@@ -246,7 +246,6 @@
 			children = (
 				F6BCCFDF1C83809A00FE31AE /* lib */,
 				F62A35131C18E6E2004A917D /* iOS */,
-				F60103051CC4ADE500EC01BA /* JS */,
```
appden: Why did you remove this group? It was helpful for me to keep the JS engine abstractions in a separate group from the stuff that uses those abstractions.

src/js_realm.hpp
```diff
@@ -297,11 +297,26 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
...
+                        create_object<T, RealmClass<T>>(ctx, old_realm_ptr),
+                        create_object<T, RealmClass<T>>(ctx, realm_ptr)
                     };
-                    Function<T>::call(ctx, migration_function, 2, arguments);
+                    
+                    try {
+                        Function<T>::call(ctx, migration_function, 2, arguments);
+                    }
+                    catch (...) {
+                        old_realm->close();
```
appden: I really think the `old_realm` should be closed in the object store layer.

src/js_realm.hpp
```diff
@@ -322,14 +337,17 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
...
     SharedRealm realm = realm::Realm::get_shared_realm(config);
-    auto delegate = new RealmDelegate<T>(realm, Context<T>::get_global_context(ctx));
-
-    if (!realm->m_binding_context) {
-        realm->m_binding_context.reset(delegate);
+    
+    // Fix for datetime -> timestamp conversion
+    if (realm->config().upgrade_initial_version != realm->config().upgrade_final_version &&
+        realm->config().upgrade_initial_version < 5) {
+        assert(0);
```
appden: Maybe this needs a comment, but what is so bad about this condition that it needs to crash?

src/object-store/src/shared_realm.hpp
```diff
@@ -82,7 +82,10 @@ namespace realm {
             // everything can be done deterministically on one thread, and
             // speeds up tests that don't need notifications.
             bool automatic_change_notifications = true;
-
+            // File format versions populated when a file format updrade takes place
+            // during realm opening
+            int upgrade_initial_version, upgrade_final_version;
```
appden: You could just default each of these to `0` here, and remove them from the initializer lists of the constructors.

src/object-store/src/shared_realm.hpp
```diff
@@ -82,7 +82,10 @@ namespace realm {
             // everything can be done deterministically on one thread, and
             // speeds up tests that don't need notifications.
             bool automatic_change_notifications = true;
-
+            // File format versions populated when a file format updrade takes place
```
appden: Typo: `updrade`

src/ios/RealmJS.xcodeproj/project.pbxproj
alazier: That abstraction was not clear at all. If we make all of are naming consistent I am ok with whatever but as it was I was having to look around in different groups to fine `js_` files which is confusing and annoying.

src/js_realm.hpp
alazier: You are reviewing a WIP which is not ready for review.

src/object-store/src/shared_realm.hpp
alazier: good point

scripts/test.sh
```diff
@@ -158,21 +158,9 @@ case "$TARGET" in
...
-  node-gyp configure
-
-  # Being explicit about debug mode rather than relying on defaults.
-  if [[ $CONFIGURATION == 'Debug' ]]; then
-    node-gyp build --debug
-  else
-    node-gyp build --no-debug
-  fi
-
+  bash build-node.sh $CONFIGURATION
```
appden: Could you just execute this directly rather than passing to `bash`

src/ios/platform.mm
```diff
@@ -64,6 +64,25 @@ void ensure_directory_exists_for_file(const std::string &fileName)
     }
 }
 
+void copy_bundled_realm_files()
+{
+    for (id bundle in [NSBundle allBundles]) {
+        NSString *resourcePath = [bundle resourcePath];
+        NSString *docsDir = @(default_realm_file_directory().c_str());
+        NSFileManager *manager = [NSFileManager defaultManager];
```
appden: These last two can be pulled out of the loop

src/ios/platform.mm
```diff
@@ -64,6 +64,25 @@ void ensure_directory_exists_for_file(const std::string &fileName)
...
+        
+        for (NSString *path in [manager enumeratorAtPath:resourcePath]) {
+            if (![path containsString:@".realm"]) {
+                continue;
+            }
+            NSError *error;
```
appden: This error isn't being used

src/js_realm.hpp
appden: Still think it makes sense to move this to object store lambda

src/js_realm.hpp
```diff
@@ -322,14 +339,35 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
...
-    auto delegate = new RealmDelegate<T>(realm, Context<T>::get_global_context(ctx));
-
-    if (!realm->m_binding_context) {
-        realm->m_binding_context.reset(delegate);
+    
+    // Fix for datetime -> timestamp conversion
+    if (realm->config().upgrade_initial_version != realm->config().upgrade_final_version &&
+        realm->config().upgrade_initial_version < 5) {
+        // any versions earlier than file format 5 are stored as milliseconds and need to be converted to the new format
+        for (auto object_schema : *realm->config().schema) {
```
appden: I think it would be nice to move this code to another function (even file)

appden: Also, I think `auto&` is better here

src/js_realm.hpp
```diff
@@ -322,14 +339,35 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
...
+            auto table = ObjectStore::table_for_object_type(realm->read_group(), object_schema.name);
+            for (auto property : object_schema.properties) {
```
appden: `auto&` here too

src/js_realm.hpp
```diff
@@ -322,14 +339,35 @@ void Realm<T>::constructor(ContextType ctx, ObjectType this_object, size_t argc,
...
+                if (property.type == PropertyTypeDate) {
+                    if (!realm->is_in_transaction()) {
+                        realm->begin_transaction();
+                    }
+                    
+                    for (size_t row_index = 0; row_index < table->size(); row_index++) {
+                        auto milliseconds = table->get_timestamp(property.table_column, row_index).get_seconds();
+                        table->set_timestamp(property.table_column, row_index, Timestamp(milliseconds / 1000, (milliseconds % 1000) * 1000000));
```
appden: I think we should do some kind of detection of the order of magnitude. I could see cases where somebody converts a Cocoa/Android app to RN, or just adds some RN code to an existing Cocoa/Android app, and this messes up the dates.

src/node/build-node.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -o pipefail
+set -e
```
appden: This should make sure the working directory is correct like other scripts with `cd "$(dirname "$0")"`

src/node/build-node.sh
```diff
@@ -0,0 +1,24 @@
...
+# Being explicit about debug mode rather than relying on defaults.
+if [[ $CONFIGURATION == 'Debug' ]]; then
+  node-gyp build --debug
+else
+  node-gyp build --no-debug
+fi
+
+# Link to the appropriate module in the build directory.
+mkdir -p ../../build
+ln -fs "../src/node/build/$CONFIGURATION/realm.node" ../../build
```
appden: The path is just `build/$CONFIGURATION/realm.node` here

src/object-store/src/property.hpp
```diff
@@ -23,25 +23,15 @@
...
         PropertyTypeString = 2,
-        /** Data type: NSData */
         PropertyTypeData   = 4,
-        /** Any type: id, **not supported in Swift** */
-        PropertyTypeAny    = 6,
-        /** Date type: NSDate */
-        PropertyTypeDate   = 7,
-        /** Object type. See [Realm Models](http://realm.io/docs/cocoa/latest/#models) */
+        PropertyTypeAny    = 6, // deprecated and will be removed in the future
+        PropertyTypeDate   = 8,
```
appden: `PropertyTypeDate` changed here from `7` to `8`. I'm assuming that's on purpose because of the format change, but then how does the migration code in `js_realm.hpp` work correctly when it's checking against `8`?

src/ios/platform.mm
```diff
@@ -64,6 +64,25 @@ void ensure_directory_exists_for_file(const std::string &fileName)
...
+            if (![manager copyItemAtPath:[resourcePath stringByAppendingPathComponent:path] toPath:[docsDir stringByAppendingPathComponent:path] error:&error]) {
```
appden: This method will fail if the file already exists, which would effectively mean that only one test could call `copyBundledRealmFiles()`. Perhaps we should overwrite it instead, or take a flag signaling whether to overwrite or not?

Realm.xcworkspace/contents.xcworkspacedata
```diff
@@ -13,4 +13,7 @@
    <FileRef
       location = "group:examples/ReactExample/ios/ReactExample.xcodeproj">
    </FileRef>
+   <FileRef
+      location = "group:examples/ReactExample/ios/../../../src/node/RealmNode.xcodeproj">
```
appden: Can you fix this path too?

src/node/RealmNode.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,184 @@
+// !$*UTF8*$!
```
appden: The `RealmNode` target still exists in `RealmJS.xcodeproj`

appden: Can you move the `RealmNode` target to this project since it's really nice to see syntax issues as you work on these files. The target here just runs a shell script.

scripts/test.sh
alazier: tried it doesn't work

src/js_realm.hpp
alazier: It shouldn't need to be closed at all...

src/js_realm.hpp
alazier: I agree this would be nice.

src/object-store/src/property.hpp
alazier: all columns are converted to 8 when the realm is opened by core

src/ios/platform.mm
alazier: or only copy if the files don't already exist? that is probably what the users really want

src/platform.hpp
```diff
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <string>
+#include <android/asset_manager.h>
```
appden: This file is shared across all platforms, so I would be fine with removing this include and just including the declaration in the file where it's used (`src/android/io_realm_react_RealmReactModule.cpp`).

src/android/platform.cpp
```diff
@@ -30,13 +44,50 @@ namespace realm {
...
+
+    }
+    
+    void copy_bundled_realm_files()
+    {
+        AAssetDir* assetDir = AAssetManager_openDir(androidAssetManager, "");
+
+        const char* filename = (const char*)NULL;
+        while ((filename = AAssetDir_getNextFileName(assetDir)) != NULL)
+        {
```
appden: Braces should be on same line

src/android/platform.cpp
```diff
@@ -30,13 +44,50 @@ namespace realm {
...
+          if (isRealmFile(filename))
+          {
+            AAsset* asset = AAssetManager_open(androidAssetManager, filename, AASSET_MODE_STREAMING);
```
appden: Could you change this to 4 space indentation

src/js_realm.hpp
appden: ...but apparently is does :smile:

alazier: It does in js becuase we can't control when objects are disposed. other platforms don't have this issue.

appden: That's not entirely true. Any platform with a GC, like when Java starts using this, will have an issue. Also, even on something like Cocoa, if you pass the old realm into a block provided by the user of our API, we have no control over whether they accidentally retain it or put it into an autorelease pool.

src/js_realm.hpp
```diff
@@ -253,6 +253,31 @@ inline typename T::Function Realm<T>::create_constructor(ContextType ctx) {
 
     return realm_constructor;
 }
+    
+void convert_outdated_datetime_columns(SharedRealm realm) {
```
appden: This needs to be `inline` or preferably `static inline` so it doesn't violate the One Definition Rule (ODR). Also, probably could take `const SharedRealm&` as its argument.

alazier: I made the method static which is all we really need.

PR #426, opened 2016-05-10
Plug RealmDelegate leak and support changing constructors/defaults

appden:
The primary reason for support changing constructors/defaults is for hot module loading to work correctly. Also, before doing `new Realm()` would wipe out the constructors and defaults, which was not intended. We now only swap them if a new `schema` was provided. I added some tests for this behavior.

### DIFF
--- a/tests/js/realm-tests.js
+++ b/tests/js/realm-tests.js
@@ -402,7 +402,7 @@ module.exports = BaseTest.extend({
     testRealmCreateWithDefaults: function() {
         var realm = new Realm({schema: [schemas.DefaultValues, schemas.TestObject]});
 
-        var writeCallback = function() {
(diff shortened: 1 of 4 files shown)


tests/js/realm-tests.js
```diff
@@ -418,15 +417,43 @@ module.exports = BaseTest.extend({
...
+        var objectSchema = {
+            name: 'IntObject',
+            properties: {
+                intCol: {type: 'int', default: 1},
+            }
+        };
+
+        var realm = new Realm({schema: [objectSchema]});
+
+        var writeCallback = function() {
```
alazier: maybe give this a more descriptive name `createAndTestObject` or something.

alazier:
looks good other than minor suggestions 👍

PR #427, opened 2016-05-10
Fix ES6 class inheritance of Realm.Object with Babel

appden:
From [a comment](https://github.com/realm/realm-js/pull/395#issuecomment-217060447) on #395, it was clear that we failed to test that `extends Realm.Object` would work. This makes it work and checks that it works in React Native with the example app test.


alazier:
👍

PR #436, opened 2016-05-16
Updates for React Native 0.25

### LINKED ISSUE #433: Fix Imports of React and React Native
See the deprecation notice on: https://github.com/facebook/react-native/releases/tag/v0.25.1:

Instead of:
```js
import React, { Component, View } from 'react-native';
```
you should now:
```js
import React, { Component } from 'react';
import { View } from 'react-native';
```

appden:
This fixes some warnings that our `ListView` generates (fixes #433).

### DIFF
--- a/examples/ReactExample/components/todo-app.js
+++ b/examples/ReactExample/components/todo-app.js
@@ -18,10 +18,13 @@
 
 'use strict';
 
-import React, {
+import React from 'react';
+
+import {
     Component,
     Navigator,
-    StatusBarIOS,
+    Platform,
+    StatusBar,
     Text,
     TouchableOpacity,
     View,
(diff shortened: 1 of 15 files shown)


alazier:
Can you fix the imports as requested in #433 ?

examples/ReactExample/components/todo-app.js
```diff
@@ -18,10 +18,13 @@
 
 'use strict';
 
-import React, {
+import React from 'react';
+
+import {
```
alazier: Do we need to change this?

appden:
I did

examples/ReactExample/components/todo-app.js
appden: Yes this change was very intentional

alazier: What I mean is do we need to change this so that each import is separate? Isn't that what was deprecated with the latest release?

appden: The deprecation is that the `react-native` module no longer re-exports `React`, so you need to import it separately.

alazier:
I didn't fully understand the deprecation. Now that I do 👍

PR #439, opened 2016-05-16
Add `isValid` method to Results and List

### LINKED ISSUE #420: Add isValid() method for collections
This would mirror the functionality of the `isValid()` method on `Realm.Object`

alazier:
closes #420 

Also updates to latest object store.

@appden


src/js_realm.hpp
```diff
@@ -116,15 +116,16 @@ class RealmDelegate : public BindingContext {
...
 };
 
 std::string default_path();
 void set_default_path(std::string path);
 void delete_all_realms();
 
 template<typename T>
-class Realm {
+class RealmClass : public ClassDefinition<T, SharedRealm> {
+  public:
```
appden: There's no reason to make all of these types public.

src/object-store/src/results.cpp
```diff
@@ -411,7 +411,7 @@ Query Results::get_query() const
             // The TableView has no associated query so create one with no conditions that is restricted
             // to the rows in the TableView.
             m_table_view.sync_if_needed();
-            return Query(*m_table, std::make_unique<TableView>(m_table_view));
+            return Query(*m_table, (std::unique_ptr<TableViewBase>)std::make_unique<TableView>(m_table_view));
```
appden: You can just do `make_unique<TableViewBase>` rather than use a C-style cast, right?

src/object-store/src/results.cpp
```diff
@@ -326,8 +408,21 @@ Query Results::get_query() const
...
+            // by TableView::get_query returning a Query with a non-null table.
+            Query query = m_table_view.get_query();
+            if (query.get_table()) {
+                return query;
+            }
+
+            // The TableView has no associated query so create one with no conditions that is restricted
+            // to the rows in the TableView.
+            m_table_view.sync_if_needed();
+            return Query(*m_table, (std::unique_ptr<TableViewBase>)std::make_unique<TableView>(m_table_view));
```
appden: You can just do `make_unique<TableViewBase>` rather than use a C-style cast, right?

bdash:
We're doing roughly the same thing for Cocoa in realm/realm-cocoa#3603. Note that we're also "fixing" `Results` to only be invalidated if `invalidate` is called on a Realm, rather than the current behavior where it's invalidated when it depends on an object that's deleted.

src/object-store/src/results.cpp
```diff
@@ -56,20 +57,56 @@ Results::Results(SharedRealm r, const ObjectSchema &o, Table& table)
...
-    if (m_background_query) {
-        m_background_query->unregister();
+    if (m_notifier) {
+        m_notifier->unregister();
     }
 }
 
+bool Results::is_valid() const
+{
+    if (m_table && !m_table->is_attached())
```
bdash: This should perform a thread check.

bdash:
Seems like it'd be a lot clearer to split updating the object store and adding new functionality into two PRs.

alazier:
@bdash agreed about separate pr being better but it would be a pain to separate at this point. As far as the corresponding changes to cocoa goes, I'm happy to move over to that results impl once it is on object store. For now will leave it as is.

src/node/binding.gyp
```diff
@@ -6,16 +6,20 @@
         "node_init.cpp",
         "platform.cpp",
         "../js_realm.cpp",
+	"../object-store/src/collection_notifications.cpp",
```
appden: This looks like a hard tab...

src/object-store/src/results.cpp
appden: Still think this one should be addressed too

alazier: doens't work

appden:
:+1:

PR #440, opened 2016-05-17
Consolidate RealmNode.xcodeproj into RealmJS.xcodeproj

appden:
Moved it up one level since it references the whole source tree inside the src/ directory. Added the functionality to run Node tests from Xcode on the existing RealmNode target.


alazier:
The react-native dir seems out of place now. Perhaps it should also be reorganized and moved under src? It is now the only src not under that directory.

appden:
We need the React Native code to be in a separate Xcode project for compatibility with `rnpm`. Moving it under `src/` kind of implies that `RealmJS.xcodeproj` would access them, but it strictly does not. Also, the JS files need to stay there so `require('realm/react-native')` works.

alazier:
Can the structure of our npm package can be different from the structure of our repo?

appden:
Kind of, but requires writing some scripts and would complicate things.

appden:
@alazier I would really like to merge this too

alazier:
I'm not happy with this new structure. If this is needed for the other prs we can merge it and change it again later but as is this is kind of confusing.

PR #441, opened 2016-05-17
Add a working WeakRealmNotifier for Node

appden:
Uses libuv APIs to async schedule onto the default Node run loop.

### DIFF
--- a/src/object-store/src/impl/node/weak_realm_notifier.cpp
+++ b/src/object-store/src/impl/node/weak_realm_notifier.cpp
@@ -16,6 +16,8 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#include <nan.h>
+
 #include "impl/weak_realm_notifier.hpp"
 
 #include "shared_realm.hpp"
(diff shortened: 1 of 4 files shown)


src/object-store/src/impl/node/weak_realm_notifier.cpp
```diff
@@ -0,0 +1,78 @@
...
+    rgt.m_handle = nullptr;
+
+    return *this;
+}
+
+WeakRealmNotifier::~WeakRealmNotifier()
+{
+    if (m_handle) {
+        uv_close((uv_handle_t*)m_handle, [](uv_handle_t* handle) {
+            auto realm_weak_ptr = static_cast<std::weak_ptr<Realm>*>(handle->data);
```
alazier: are these three calls safe/correct given that they are made async?

appden: Yes, the handle is created in one place and destroyed in one place. There's no concurrent access to it.

alazier:
What tests that this actually works properly?

appden:
The fact it compiles and doesn't crash. :smile: I have a branch where this is tested, but relies on merging other PRs, so I'm working toward getting that into master.

appden:
@alazier I would really like to merge this

alazier:
We need to have automated tests. What other prs do the tests rely on?

appden:
Once I incorporate async tests, then this will be tested. This PR does not affect a released product yet.

alazier:
Understood - but when I make changes I want to be able to verify that I am not breaking this functionality.

appden:
Async tests are coming in another PR very soon. I don't see how you'll break this in the meantime. :smile:

PR #446, opened 2016-05-19
Allow key paths to start with "not" in queries

### LINKED ISSUE #445: field name cannot start with 'not',  such as 'notic'.
there are two query string:

1. filtered( '$0==notic',notic)
2. filtered( 'notic==$0',notic)

first is correct, second is error. 
I think that  second string is equal with ' not  ic==$0'.

can you understand?

appden:
Fixes #445

### DIFF
--- a/src/object-store/src/parser/parser.cpp
+++ b/src/object-store/src/parser/parser.cpp
@@ -24,6 +24,8 @@
 #include <pegtl/analyze.hh>
 #include <pegtl/trace.hh>
 
+#define string_operator_t(s) seq< pegtl_istring_t(s), not_at< identifier_other > >
+
 using namespace pegtl;
 
 namespace realm {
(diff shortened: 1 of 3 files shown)


src/object-store/src/parser/parser.cpp
```diff
@@ -24,6 +24,8 @@
 #include <pegtl/analyze.hh>
 #include <pegtl/trace.hh>
 
+#define string_operator_t(s) seq< pegtl_istring_t(s), not_at< identifier_other > >
```
alazier: comment here would be nice.

alazier: naming is ambiguous as well

alazier:
Need to fix naming/comments for our macro. Actual logic looks good.

src/object-store/tests/parser.cpp
```diff
@@ -66,6 +66,9 @@ static std::vector<std::string> valid_queries = {
     "!(0=0)",
     "! (0=0)",
     "NOT0=0",   // keypath NOT0
+    "NOT0.a=0", // keypath NOT0
```
alazier: maybe add a similar test for AND/OR

src/object-store/src/parser/parser.cpp
```diff
@@ -93,11 +95,11 @@ struct group_pred : if_must< one< '(' >, pad< pred, blank >, one< ')' > > {};
...
 struct false_pred : pegtl_istring_t("falsepredicate") {};
 
-struct not_pre : seq< sor< one< '!' >, pegtl_istring_t("not") > > {};
+struct not_pre : seq< sor< one< '!' >, string_operator_t("not") > > {};
 struct atom_pred : seq< opt< not_pre >, pad< sor< group_pred, true_pred, false_pred, comparison_pred >, blank > > {};
 
-struct and_op : pad< sor< two< '&' >, pegtl_istring_t("and") >, blank > {};
-struct or_op : pad< sor< two< '|' >, pegtl_istring_t("or") >, blank > {};
+struct and_op : pad< sor< two< '&' >, string_operator_t("and") >, blank > {};
+struct or_op : pad< sor< two< '|' >, string_operator_t("or") >, blank > {};
```
bdash: Is a similar change needed for `TRUE`, `FALSE`, `NULL`, `TRUEPREDICATE`, `FALSEPREDICATE`, `CONTAINS`, `BEGINSWITH`, and `ENDSWITH`?

appden: @bdash I opened PR #456 that I believe fixes the rest of the issues you brought up.

PR #451, opened 2016-05-20
Improve JS tests

appden:
Fixed a number of minor outstanding issues.


tests/js/query-tests.js
```diff
@@ -77,18 +75,24 @@ function runQuerySuite(suite) {
 
     for (var index in suite.tests) {
         var test = suite.tests[index];
+        var type;
```
alazier: why this change? these variables are not used outside of this block. no reason to declare them up here.

alazier:
Everything else looks fine.

tests/js/query-tests.js
appden: Because `var` is function-scoped, not block scoped, and we can't use `let` in this file since it must be ES5.

alazier:
Why did react-tests Debug fail?

appden:
@alazier no idea, I'm re-running it :smile:

appden:
@alazier I rebased this branch and fixed a bug that was causing debug tests to fail. It's good to go now.

alazier:
👍

PR #456, opened 2016-05-26
Improve parsing of string tokens

appden:
All of the parsing behavior was confirmed to be compatible with the `NSPredicate` parser.

### DIFF
--- a/src/object-store/src/parser/parser.cpp
+++ b/src/object-store/src/parser/parser.cpp
@@ -82,11 +82,11 @@ struct contains : string_token_t("contains") {};
 struct begins : string_token_t("beginswith") {};
 struct ends : string_token_t("endswith") {};
 
-struct string_oper : pad< seq< sor< contains, begins, ends>, star< blank >, opt< case_insensitive > >, blank > {};
-struct symbolic_oper : pad< sor< eq, noteq, lteq, lt, gteq, gt >, blank > {};
+struct string_oper : seq< sor< contains, begins, ends>, star< blank >, opt< case_insensitive > > {};
+struct symbolic_oper : sor< eq, noteq, lteq, lt, gteq, gt > {};
 
 // predicates
-struct comparison_pred : seq< expr, sor< string_oper, symbolic_oper >, expr > {};
+struct comparison_pred : seq< expr, pad< sor< string_oper, symbolic_oper >, blank >, expr > {};
 
 struct pred;
 struct group_pred : if_must< one< '(' >, pad< pred, blank >, one< ')' > > {};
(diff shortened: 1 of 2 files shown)


src/object-store/tests/parser.cpp
```diff
@@ -34,10 +38,13 @@ static std::vector<std::string> valid_queries = {
     // operators
     "0=0",
     "0 = 0",
+    "0 =[c] 0",
     "0!=0",
     "0 != 0",
     "0==0",
     "0 == 0",
+    "0==[c]0",
+    "0 == [c] 0",
```
alazier: I don't think this should be valid. Spaces should not be allowed between modifiers and operators.

src/object-store/src/parser/parser.cpp
```diff
@@ -70,37 +70,34 @@ struct argument : seq< one< '$' >, must< argument_index > > {};
 
 // expressions and operators
 struct expr : sor< dq_string, sq_string, number, argument, true_value, false_value, null_value, key_path > {};
-struct case_insensitive : pegtl_istring_t("[c]"){};
+struct case_insensitive : pegtl_istring_t("[c]") {};
 
-struct eq : seq< sor< two< '=' >, one< '=' > >, opt< case_insensitive > >{};
+struct eq : seq< sor< two< '=' >, one< '=' > >, star< blank >, opt< case_insensitive > >{};
```
alazier: why do we want to allow spaces here?

src/object-store/src/parser/parser.cpp
```diff
@@ -70,37 +70,34 @@ struct argument : seq< one< '$' >, must< argument_index > > {};
...
-struct ends : pegtl_istring_t("endswith") {};
+struct contains : string_token_t("contains") {};
+struct begins : string_token_t("beginswith") {};
+struct ends : string_token_t("endswith") {};
 
-template<typename A, typename B>
-struct pad_plus : seq< plus< B >, A, plus< B > > {};
-
-struct padded_oper : pad_plus< seq< sor< contains, begins, ends>, opt< case_insensitive > >, blank > {};
+struct string_oper : pad< seq< sor< contains, begins, ends>, star< blank >, opt< case_insensitive > >, blank > {};
```
alazier: This line is pretty hard to read now. Might want to break it up into sub rules

src/object-store/src/parser/parser.cpp
appden: `NSPredicate` allows for spaces, and since we're adopting its syntax, we should too

src/object-store/tests/parser.cpp
appden: Like I said above, since we're using `NSPredicate` syntax, we shouldn't arbitrarily add whitespace restrictions to it.

src/object-store/src/parser/parser.cpp
appden: I factored out the `pad` and I think it reads better now.

alazier:
👍

PR #457, opened 2016-05-26
Fix for upgrading optional date properties

alazier:
@appden


appden:
It would be a lot easier if you made this PR on the `0.13.x` branch, since I intend on merging it there first and releasing another point release.

appden:
Published `0.13.2` with this fix

PR #462, opened 2016-05-27
PR feedback fixes from object store pr

alazier:


appden:
I think it's easier to just wait for that PR to land, and then just merge it into our repo. That way we don't have a bunch of duplicate commits.

alazier:
We won't have duplicate commits - we are using git subtree.

alazier:
@appden anything left to do here or are we good to merge?

appden:
I believe this is good to go!

PR #466, opened 2016-05-31
Remove unused GCov_Build configuration

appden:
We've never really used this, so let's remove it until we ever actually need something like this.


alazier:
👍

PR #467, opened 2016-05-31
Avoid copying `Object` and `List` instances unnecessarily

bdash:
Avoid copying `Object` and `List` instances when we can instead move them.


appden:
Looks good :+1:

alazier:
👍

PR #468, opened 2016-06-01
Add async inter-process change notification test for Node

### LINKED ISSUE #461: Add async tests to test WeakRealmNotifier
#441 adds this for Node, and should be tested.

appden:
This PR also lays the foundation for async tests to come to other platforms. It ensures the `ExternalCommitHelper` and `WeakRealmNotifier` are working as intended.

Resolves #461

### DIFF
--- a/tests/js/index.js
+++ b/tests/js/index.js
@@ -63,10 +63,24 @@ exports.runTest = function(suiteName, testName) {
         // Start fresh in case of a crash in a previous run.
         Realm.clearTestState();
 
+        var promise;
         try {
-            testMethod.call(testSuite);
+            promise = testMethod.call(testSuite);
+
+            // If the test returns a promise, then clear state on success or failure.
+            if (promise) {
+                promise.then(
+                    function() { Realm.clearTestState(); },
+                    function() { Realm.clearTestState(); }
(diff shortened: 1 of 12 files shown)


tests/js/index.js
```diff
@@ -63,10 +63,24 @@ exports.runTest = function(suiteName, testName) {
         // Start fresh in case of a crash in a previous run.
         Realm.clearTestState();
 
+        var promise;
         try {
-            testMethod.call(testSuite);
+            promise = testMethod.call(testSuite);
+
+            // If the test returns a promise, then clear state on success or failure.
```
alazier: why would the test not return a promise? wouldn't we still want to clear the state in that case?

appden: Most of our tests don't return promises since they are synchronous.

alazier: ok. i see the clear in the finally block. this looks fine.

alazier:
👍  once ci is passing.

appden:
@alazier Looks like we had a legit Chrome debug mode failure caused by merging #479. The rest is passing. So I'm going to merge, and will fix the legit issue in another PR.

PR #472, opened 2016-06-03
Implement objectForPrimaryKey() method

### LINKED ISSUE #328: Add method Realm.objectForPrimaryKey(objectType, primaryKey)
I would like to see a simple method to retrieve an object by primary key.

Ideally something similar to this would be simple to use.

`let myObject = results[ 'keyName' ];`

Or maybe what Swift developers have:

`public func objectForPrimaryKey<T: Object>(type: T.Type, key: AnyObject) -> T?`
https://realm.io/docs/swift/latest/api/Classes/Realm.html#/s:FC10RealmSwift5Realm19objectForPrimaryKeyuRdq_CS_6Object_FS0_FTMq_3keyPSs9AnyObject__GSqq__

Thank you for making realm easy to get going with on react-native!

appden:
Resolves #328

### DIFF
--- a/lib/browser/index.js
+++ b/lib/browser/index.js
@@ -56,6 +56,13 @@ function setupRealm(realm, realmId) {
     });
 }
 
+function getObjectType(realm, type) {
+    if (typeof type == 'function') {
+        return objects.typeForConstructor(realm[keys.realm], type);
+    }
+    return type;
+}
+
 export default class Realm {
     constructor(config) {
         let schemas = typeof config == 'object' && config.schema;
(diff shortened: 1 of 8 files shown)


lib/browser/index.js
```diff
@@ -106,6 +106,15 @@ export default class Realm {
         let method = util.createMethod(objectTypes.REALM, 'objects');
         return method.apply(this, [type, ...args]);
     }
+
+    objectForPrimaryKey(type, ...args) {
+        if (typeof type == 'function') {
```
alazier: We could just move the check for function into `typeForConstruction` to make this a one liner every time we need to do this.

src/js_realm.hpp
```diff
@@ -207,19 +209,35 @@ class RealmClass : public ClassDefinition<T, SharedRealm> {
     }
     
     // converts constructor object or type name to type name
```
alazier: this comments is no longer accurate

appden:
@alazier I made changes according to your feedback here.

src/js_realm.hpp
```diff
@@ -462,35 +479,45 @@ void RealmClass<T>::objects(ContextType ctx, ObjectType this_object, size_t argc
     validate_argument_count(argc, 1);
 
     SharedRealm realm = *get_internal<T, RealmClass<T>>(this_object);
-    std::string type = validated_object_type_for_value(realm, ctx, arguments[0]);
+    auto &object_schema = validated_object_schema_for_value(ctx, realm, arguments[0]);
```
alazier: do you need ```auto &``` since this returns a ref?

alazier:
everything else looks good 👍

src/js_realm.hpp
appden: `auto` doesn't become a reference without explicitly adding the `&`

PR #479, opened 2016-06-08
fix for setting properties to objects from other realms

alazier:
@appden @bigfish24

### DIFF
--- a/src/js_object_accessor.hpp
+++ b/src/js_object_accessor.hpp
@@ -120,10 +120,14 @@ struct NativeAccessor {
         auto child = realm::Object::create<ValueType>(ctx, realm, *object_schema, static_cast<ValueType>(object), try_update);
         return child.row().get_index();
     }
-    static size_t to_existing_object_index(ContextType ctx, ValueType &value) {
+    static size_t to_existing_object_index(ContextType ctx, SharedRealm realm, ValueType &value) {
         ObjectType object = Value::validated_to_object(ctx, value);
         if (Object::template is_instance<RealmObjectClass<T>>(ctx, object)) {
-            return get_internal<T, RealmObjectClass<T>>(object)->row().get_index();
+            auto realm_object = get_internal<T, RealmObjectClass<T>>(object);
+            if (realm_object->realm() == realm) {
+                return realm_object->row().get_index();
+            }
+            throw std::runtime_error("Realm object is from another Realm");
         }
         throw std::runtime_error("object is not a Realm Object");
     }
--- a/src/js_object_accessor.hpp
+++ b/src/js_object_accessor.hpp
@@ -122,14 +122,16 @@ struct NativeAccessor {
     }
     static size_t to_existing_object_index(ContextType ctx, SharedRealm realm, ValueType &value) {
         ObjectType object = Value::validated_to_object(ctx, value);
-        if (Object::template is_instance<RealmObjectClass<T>>(ctx, object)) {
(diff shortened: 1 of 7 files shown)


src/js_object_accessor.hpp
```diff
@@ -106,7 +106,10 @@ struct NativeAccessor {
     static size_t to_object_index(ContextType ctx, SharedRealm realm, ValueType &value, const std::string &type, bool try_update) {
         ObjectType object = Value::validated_to_object(ctx, value);
         if (Object::template is_instance<RealmObjectClass<T>>(ctx, object)) {
-            return get_internal<T, RealmObjectClass<T>>(object)->row().get_index();
+            auto realm_object = get_internal<T, RealmObjectClass<T>>(object);
+            if (realm_object->realm() == realm) {
+                return realm_object->row().get_index();
+            }
```
appden: Shouldn't this return an invalid index in an `else` clause here?

appden: Also, `to_existing_object_index()` should probably be modified as well, since I could see this potentially causing a crash when making queries.

alazier: In the else case the object is treated just like a json object which is the whole point of this change.

appden: Oh I get it, a new Realm object is created. Do other bindings behave this way? It feels a little like magic and that we should instead be throwing an error.

appden:
@alazier I just pushed a fix for queries onto this branch. You can review it and merge if you're happy with it.

tests/js/results-tests.js
```diff
@@ -167,9 +172,19 @@ module.exports = {
...
+        var realm = new Realm({schema: [schemas.LinkTypes, schemas.TestObject]});
+        var realm2 = new Realm({path: '2.realm', schema: realm.schema});
+        var object;
+
+        realm2.write(function() {
+            object = realm2.create('TestObject', {doubleCol: 1});
+        });
 
         TestCase.assertThrows(function() {
-            realm.objects('PersonObject').filtered("invalidQuery");
```
alazier: why did you get rid of this old test?

src/js_object_accessor.hpp
```diff
@@ -120,10 +120,14 @@ struct NativeAccessor {
         auto child = realm::Object::create<ValueType>(ctx, realm, *object_schema, static_cast<ValueType>(object), try_update);
         return child.row().get_index();
     }
-    static size_t to_existing_object_index(ContextType ctx, ValueType &value) {
+    static size_t to_existing_object_index(ContextType ctx, SharedRealm realm, ValueType &value) {
         ObjectType object = Value::validated_to_object(ctx, value);
         if (Object::template is_instance<RealmObjectClass<T>>(ctx, object)) {
-            return get_internal<T, RealmObjectClass<T>>(object)->row().get_index();
+            auto realm_object = get_internal<T, RealmObjectClass<T>>(object);
+            if (realm_object->realm() == realm) {
```
alazier: Would prefer to throw from in the ```if```

tests/js/results-tests.js
appden: I moved it up next to the other `assertThrows()` above. ⬆️

src/js_object_accessor.hpp
appden: Ok, will do!

alazier: ill fix this

appden:
Looks good :+1:

PR #480, opened 2016-06-09
Fix bug due to std::thread::id being reused

### LINKED ISSUE #473: Realm is already open in another context on this thread.
This message just started appearing. Sometimes refreshing the simulator will fix it but other times I have to quit the simulator and launch it again. 

Can you shed some light on what this error is and how to go about preventing it? 

Thanks. 

screenshot 2016-06-06 08 14 59

appden:
After a thread is destroyed, a thread id might be reused. This is true for pthread_self() and hence std::thread::id. We were hitting this in Chrome debug mode because the "worker" thread was being destroyed and a new one immediately created when reloading. The thread id would be the same, and therefore we'd get back the SharedRealm for the previously destroyed thread (which had yet to be garbage collected.

The new implementation uses an atomically incremented, thread-local identifier, which serves our needs very well.

Fixes #473

### DIFF
--- a/src/RealmJS.xcodeproj/project.pbxproj
+++ b/src/RealmJS.xcodeproj/project.pbxproj
@@ -189,6 +189,7 @@
 		F620F0571CB766DA0082977B /* node_init.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = node_init.cpp; sourceTree = "<group>"; };
 		F620F0591CB7B4C80082977B /* js_object_accessor.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = js_object_accessor.hpp; sourceTree = "<group>"; };
 		F620F0741CB9F60C0082977B /* CoreFoundation.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = CoreFoundation.framework; path = Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreFoundation.framework; sourceTree = DEVELOPER_DIR; };
+		F6242B291D08EE9600BE1E03 /* thread_id.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = thread_id.hpp; sourceTree = "<group>"; };
 		F6267BC91CADC30000AC36B1 /* js_util.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = js_util.hpp; sourceTree = "<group>"; };
 		F6267BCA1CADC49200AC36B1 /* node_dummy.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = node_dummy.cpp; sourceTree = "<group>"; };
 		F62BF8FB1CAC71780022BCDC /* libRealmNode.dylib */ = {isa = PBXFileReference; explicitFileType = "compiled.mach-o.dylib"; includeInIndex = 0; path = libRealmNode.dylib; sourceTree = BUILT_PRODUCTS_DIR; };
@@ -363,6 +364,7 @@
 				02E008D11D10AB1B00F3AA37 /* atomic_shared_ptr.hpp */,
 				02E008D21D10AB1B00F3AA37 /* format.cpp */,
 				02E008D31D10AB1B00F3AA37 /* format.hpp */,
+				F6242B291D08EE9600BE1E03 /* thread_id.hpp */,
 			);
 			name = util;
 			path = src/util;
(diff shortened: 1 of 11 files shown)


alazier:
What's going on with ci?

alazier:
The code looks fine but we should ensure the ci issues are not indicative of a real issue.

bdash:
The CI failure looks like a real issue:
```
/Users/realm/workspace/js_pr/configuration/Debug/target/realmjs/src/object-store/src/util/thread_id.hpp:36:12: error: thread-local storage is not supported for the current target
    static REALM_THREAD_LOCAL thread_id_t thread_id = 0;
           ^
In file included from /Users/realm/workspace/js_pr/configuration/Debug/target/realmjs/src/object-store/src/impl/list_notifier.cpp:19:
In file included from /Users/realm/workspace/js_pr/configuration/Debug/target/realmjs/src/object-store/src/impl/list_notifier.hpp:22:
In file included from /Users/realm/workspace/js_pr/configuration/Debug/target/realmjs/src/object-store/src/impl/collection_notifier.hpp:24:
In file included from ../core/include/realm/group_shared.hpp:29:
../core/include/realm/util/features.h:178:30: note: expanded from macro 'REALM_THREAD_LOCAL'
#  define REALM_THREAD_LOCAL __thread
```

appden:
This was caught because CI tests on iPhone 4s, which turns out to not even support `__thread`, same with iPhone 5 targets. I just pushed a fix for that.

src/object-store/src/util/thread_id.hpp
```diff
@@ -0,0 +1,104 @@
...
+#else
+#define REALM_HAS_THREAD_LOCAL 0
+#include <pthread.h>
+#endif
+
+namespace realm {
+
+using thread_id_t = std::size_t;
+
+namespace _impl {
```
bdash: IMO there's enough going on in this header now that it'd be worth moving most of it to a .cpp file.

appden: I actually disagree because `get_thread_id()` is a very hot function that should be inlined, and have its implementation inlined, as much as possible. Also, `ThreadLocal` is a class template.

bdash: I'm skeptical that `get_thread_id` is so hot that the cost of a single function call is unbearable. I'd love to see the data you've gathered on this.

appden: Honestly, I think `verify_thread()` should be inlined as well, and I think it should be called on property access to objects. I set a breakpoint and saw it called over 5000 times during our test suite. That's all the data I have. This is a pretty silly argument though. It's a very small utility function that makes sense to stay in a header given the existing style of the rest of the object store codebase.

bdash: I didn't expect this to be a contentious point. If it were only the ~5 line function itself I'd agree it's not worth it, but the fact there's 10x that in supporting code is why i suggested moving it to a .cpp file.

alazier: The other object store code base in headers is there because of the use of
templates. Anything more than a few lines should go in a cpp file when
possible.

src/object-store/src/util/thread_id.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+#if !REALM_HAS_THREAD_LOCAL
+template<typename T>
```
alazier: Does this even need to be templated?

src/RealmJS.xcodeproj/project.pbxproj
```diff
@@ -69,6 +69,8 @@
 		F63FF3261C1642BB00B3B8E0 /* GCDWebServerFileResponse.m in Sources */ = {isa = PBXBuildFile; fileRef = F63FF3181C1642BB00B3B8E0 /* GCDWebServerFileResponse.m */; };
 		F63FF3271C1642BB00B3B8E0 /* GCDWebServerStreamedResponse.m in Sources */ = {isa = PBXBuildFile; fileRef = F63FF31A1C1642BB00B3B8E0 /* GCDWebServerStreamedResponse.m */; };
 		F64A059B1D10D928004ACDBE /* format.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 02E008D21D10AB1B00F3AA37 /* format.cpp */; };
+		F64A059D1D13710C004ACDBE /* thread_id.cpp in Sources */ = {isa = PBXBuildFile; fileRef = F64A059C1D13710C004ACDBE /* thread_id.cpp */; };
```
alazier: why twice?

src/object-store/src/util/thread_id.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <atomic>
+
+using namespace realm;
+
+// Since std::thread::id may be reused after a thread is destroyed, we use
+// an atomically incremented, thread-local identifier instead.
+thread_id_t realm::util::get_thread_id() {
+    static std::atomic<thread_id_t> id_counter;
+    static REALM_THREAD_LOCAL_TYPE(thread_id_t) thread_id = 0;
```
alazier: would prefer `using` or `typedef` instead of the macro `REALM_THREAD_LOCAL_TYPE`

alazier:
merge this and we can revisit the threadlocal stuff again later.

bdash:
The [What's New In LLVM](https://developer.apple.com/videos/play/wwdc2016/405/) WWDC session suggests that while C++ `thread_local` was not supported, C11's `_Thread_local` is supported. A quick test shows that `_Thread_local` does not give the same "thread-local storage is not supported for the current target" errors that `thread_local` causes. That may result in a clean solution to this after all.

appden:
@bdash `_Thread_local` still fails for iPhone 5 and below on Xcode 7 (haven't installed Xcode 8 yet).

bdash:
Yeah, it does seem like 32-bit ARM devices only support `_Thread_local` when their minimum iOS version is set to iOS 9.0 or higher.

PR #489, opened 2016-06-14
Add ability to run Node tests on Linux

appden:
It uses the Android ExternalCommitHelper implementation.

The Dockerfile requires that you have a static library build of core for Linux with PIC enabled.


alazier:
Don't know how to review the docker stuff. Perhaps you could write up instructions on the wiki on how to use/test it? Everything else looks good.

JamesChristie:
I have made some attempts to get this working myself, but I'm unable to get the bindings to compile. Unfortunately it looks like the Linux specific platform file here https://github.com/realm/realm-js/pull/489/files#diff-3254677a7917c6c01f55212f86c57fbfR18 is not downloaded but needs to be available ahead of time locally? We've got CI for a project using realm running on Linux and this is a big blocker for us.

If there's anything else I should try to get this up and running let me know; I'm happy to help test.

alazier:
@JamesChristie we are working on releasing linux/node core releases the same way we do for other platforms. Not sure exactly when this will happen but should be pretty soon.

JamesChristie:
Excellent! Please feel free to @-ping me if you are in need of a guinea pig for anything.

alazier:
👍

alexsotocx:
@JamesChristie @alazier I have the same issue with the CI, We are using bamboo running in Ubuntu but couldn't compile it yet.

**Please** if you need someone to test out or when it's ready

PR #490, opened 2016-06-14
Fix test failure in Chrome debug mode

appden:
We had a new failure in Chrome debug mode in the `testLinkTypesPropertySetters` test. This fixes that.


alazier:
👍

appden:
These were cherry-picked into #462, so it'll be merged there.

PR #498, opened 2016-06-20
Remove management directories in node

### LINKED ISSUE #495: realm.management directrory is not deleted when clearing test state
This fails because we are trying to call unlink on a directory.

alazier:
Closes #495 

@appden

### DIFF
--- a/src/node/platform.cpp
+++ b/src/node/platform.cpp
@@ -58,7 +58,7 @@ void copy_bundled_realm_files()
 
 void remove_realm_files_from_directory(const std::string &dir_path)
 {
-    std::string delete_realms = "rm -rf " + dir_path + "/*.realm*";
+    std::string delete_realms = "rm -rf '" + dir_path + "'/*.realm*";
     system(delete_realms.c_str());
 }
 
--- a/src/node/platform.cpp
+++ b/src/node/platform.cpp
@@ -66,7 +66,12 @@ void remove_realm_files_from_directory(const std::string &dir_path)
     while (struct dirent *entry = readdir(dir)) {
         if (strstr(entry->d_name, ".realm")) {
             // Intentionally not complaining if there was an error.
-            unlink(entry->d_name);
+            if (entry->d_type == DT_DIR) {
+                rmdir(entry->d_name);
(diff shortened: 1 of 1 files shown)


src/node/platform.cpp
```diff
@@ -66,7 +66,12 @@ void remove_realm_files_from_directory(const std::string &dir_path)
     while (struct dirent *entry = readdir(dir)) {
         if (strstr(entry->d_name, ".realm")) {
             // Intentionally not complaining if there was an error.
-            unlink(entry->d_name);
+            if (entry->d_type == DT_DIR) {
+                rmdir(entry->d_name);
```
appden: I believe `rmdir` need the directory to be empty. Also, do we only need this for Node?

alazier: The objective c api uses `[NSFileManager removeItemAtPath]` which I believe is recursive.

alazier: Ok - fixed it for node now.

src/node/platform.cpp
```diff
@@ -58,24 +58,8 @@ void copy_bundled_realm_files()
...
-            }
-            else {
-                unlink(entry->d_name);
-            }
-        }
-    }
-
-    closedir(dir);
+    std::string delete_realms = "rm -rf " + dir_path + "/*.realm*";
+    system(delete_realms.c_str());
```
appden: This is dirty, fragile, and a seriously potential for security issues and data corruption. Let's just implement recursive deleting properly. It isn't that much code.

alazier: This is only used to clean up after tests. It is fine. If this were a documented/public api then I would agree with you.

appden: I am still against adding this, but if you absolutely insist on not just writing the recursive delete, then please at least single quote the `dir_path` like this: `"rm -rf '" + dir_path + "'/*realm*"`

alazier: You are welcome to make a pr with a proper fix if you would like. I have limited time today and am happy with this fix for the time being.

appden:
😞👍

PR #499, opened 2016-06-22
Remove rnpm postlink script

appden:
This should go out in the next release (`0.15.0`), so let's make sure point releases are branched from the `v0.14.0` tag and then merged back in.


alazier:
:+1:

Issue #503, opened 2016-06-23
v0.14.0 - Getting `strip failed with exit code 1` on Archive.


joshuapinter:
I wasn't getting this before but perhaps an issue since 0.14.0.

![screenshot 2016-06-23 07 07 10](https://cloud.githubusercontent.com/assets/180819/16304141/5a9928a2-3911-11e6-9b68-ea3d42c6faa4.png)

I've got all my strip build settings to no. Any idea what could be the issue? 

Thanks!

Josh

dorongutman:
Getting the same thing:

```
Strip /Users/doron/Library/Developer/Xcode/DerivedData/MyAppMobile-aluqrtvwwettmshhhguwuycjsecp/Build/Intermediates/ArchiveIntermediates/MyAppMobile/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/libRealmJS.a
    cd /Users/doron/repos/myapp-mobile/node_modules/realm/src
    export PATH="/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin:/Applications/Xcode.app/Contents/Developer/usr/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/strip -S /Users/doron/Library/Developer/Xcode/DerivedData/MyAppMobile-aluqrtvwwettmshhhguwuycjsecp/Build/Intermediates/ArchiveIntermediates/MyAppMobile/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/libRealmJS.a

0  0x10ee1597e  __assert_rtn + 144
1  0x10ee6ca06  ld::tool::SectionRelocationsAtom<arm64>::encodeSectionReloc(ld::Internal::FinalSection*, ld::tool::SectionRelocationsAtom<arm64>::Entry const&, std::__1::vector<macho_relocation_info<Pointer64<LittleEndian> >, std::__1::allocator<macho_relocation_info<Pointer64<LittleEndian> > > >&) + 1618
2  0x10ee9732a  ld::tool::SectionRelocationsAtom<arm64>::encode() + 62
3  0x10ee7276d  ld::tool::OutputFile::updateLINKEDITAddresses(ld::Internal&) + 413
4  0x10ee6ce89  ld::tool::OutputFile::write(ld::Internal&) + 167
[...]

Issue #482, opened 2016-06-10
Assert from accessing preloaded realm file: group.cpp:286: [realm-core-0.100.2] Assertion failed: top_size == 9 [5790035, 9]


nerdcrack:
See attached for the error log.
Realm React Native: 0.13.2
Realm Browser: 0.102.1 (73)
On devices running iOS 9.3.2

The error isn't spurious and is seen persistently during development.  It seems to be related to pre-packaged Realm database that is then copied from Bundle to Documents folder, and later accessed. 

It happens in two different setups. Setup 1 is where the pre-packaged realm file was first generated from Realm Browser importing from CSV files on a laptop, and then added to the app bundle. Setup 2 is where the pre-packaged realm file was generated from the app importing from CSV files at runtime on the device, and then downloaded back to the laptop for inclusion to the app bundle (i.e., bypassing the use of Realm Browser).

[realmjs_exception.txt](https://github.com/realm/realm-js/files/309762/realmjs_exception.txt)

bmunkholm:
@nerdcrack It's super important for us to get this reproduced and fixed! So if you are able to reproduce it somewhat reliably we are really appreciate your help to reproduce this. The details you provide is unfortunately not sufficient. The basic info I get is that the Browser is not causing this. But creating a file and opening it is. That's being doing a million times a day by others so your setup/data or the way you store or retrieve it is highly interesting for us. You can send data and code in full confidentiality to help@realm.io.

appden:
@nerdcrack We released `0.14.0-rc` and believe it may be fixed in that release. Will you please update to it and let us know whether or not that's the case. Thanks a lot!

nerdcrack:
Sure thing!

appden:
@nerdcrack Checking in to see if you've been able to update and confirm this is fixed. Thanks!

nerdcrack:
@appden Sorry I don't think I could get to it until later this week. Will let know no later than next Monday. Thanks!

nerdcrack:
@appden Got a link error when building my app (Xcode 7.3.1, iOS 9.3.2, iPhone 6) with Realm v0.14.0 
ld: targeted OS version does not support use of thread local variables in __ZN5realm4util13get_thread_idEv for architecture arm64

Reverted to v0.13.2 and everything worked fine again. Any idea? Thanks!

appden:
[...]

PR #506, opened 2016-06-27
[WIP] Fixes for 0.14.1

### LINKED ISSUE #509: Realm for react-native gives IOS version warning
After installing the latest version of Realm using npm, when I tried to run my react-native app
The build failed with the following error:

d: targeted OS version does not support use of thread local variables in __ZN5realm4util13get_thread_idEv for architecture x86_64 clang: error: linker command failed with exit code 1 (use -v to see invocation)

After changing the deployment target from 7 to 8.1
I found this error 

ld: targeted OS version does not support use of thread local variables in __ZN5realm4util13get_thread_idEv for architecture x86_64

appden:
This is **WIP** until #503 is solved, and hopefully we get confirmation on #482 being fixed.

The first fix in here for an issue reported in a comment in #482, which is described below…

By default, React Native projects target a minimum of iOS 7, but our projects were set to iOS 8 because previously we used a dynamic framework.

When building for an iOS 7 minimum target version, the linker would complain about thread local variables not working on iOS 7. So it was necessary to add even more checks (iOS >= 8 or OS X >= 10.7) before using native thread local variables rather than our shim.

This also changes our projects to target iOS 7 so we can catch these issues quicker in the future.

Fixes #509


emanuelez:
retest this please

emanuelez:
retest this please

PR #513, opened 2016-06-30
Aligning mixpanel implementation with Java & Cocoa

appden:
These changed were made by @nhachicha. Essentially the analytics are sent at compile-time rather than at runtime for Android.


appden:
@nhachicha I rebased this branch. I'd appreciate if you double-checked this still works properly now that it's rebased on `master`.

The changes look good to me, so I'll leave it up to you to merge once confirm it works! :+1:

nhachicha:
it works locally (when using local http server) 👍

PR #514, opened 2016-06-30
Make Results derived from List use its LinkViewRef

appden:
There are some advantages for the `Results` being built from a `LinkViewRef`, and the object store very much encourages that with its `filter` and `sort` methods on List.

### DIFF
--- a/src/js_results.hpp
+++ b/src/js_results.hpp
@@ -38,10 +38,9 @@ struct ResultsClass : ClassDefinition<T, realm::Results, CollectionClass<T>> {
     using Value = js::Value<T>;
     using ReturnValue = js::ReturnValue<T>;
 
-    static ObjectType create_instance(ContextType, const realm::Results &, bool live = true);
+    static ObjectType create_instance(ContextType, realm::Results, bool live = true);
     static ObjectType create_instance(ContextType, const realm::List &, bool live = true);
     static ObjectType create_instance(ContextType, SharedRealm, const ObjectSchema &, bool live = true);
-    static ObjectType create_instance(ContextType, SharedRealm, const ObjectSchema &, Query, bool live = true);
 
     template<typename U>
     static ObjectType create_filtered(ContextType, const U &, size_t, const ValueType[]);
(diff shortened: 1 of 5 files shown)


appden:
@alazier @bdash This needs review too

src/js_results.hpp
```diff
@@ -74,16 +73,16 @@ struct ResultsClass : ClassDefinition<T, realm::Results, CollectionClass<T>> {
...
+    auto new_results = new realm::Results(std::move(results));
     new_results->set_live(live);
 
     return create_object<T, ResultsClass<T>>(ctx, new_results);
 }
 
 template<typename T>
 typename T::Object ResultsClass<T>::create_instance(ContextType ctx, const realm::List &list, bool live) {
-    return create_instance(ctx, list.get_realm(), list.get_object_schema(), list.get_query(), live);
+    return create_instance(ctx, list.sort({}), live);
```
bdash: This feels like a bit of an abuse of `sort()`. Not a big deal for now though.

src/js_results.hpp
```diff
@@ -96,20 +95,13 @@ typename T::Object ResultsClass<T>::create_instance(ContextType ctx, SharedRealm
...
-}
-
-template<typename T>
 template<typename U>
 typename T::Object ResultsClass<T>::create_filtered(ContextType ctx, const U &collection, size_t argc, const ValueType arguments[]) {
-    auto query_string = Value::validated_to_string(ctx, arguments[0], "predicate");
-    auto query = collection.get_query();
     auto const &realm = collection.get_realm();
     auto const &object_schema = collection.get_object_schema();
+    auto query = collection.get_table()->where();
```
bdash: This is making the assumption that a collection always has an associated table. That's not true of a default-constructed `Results` or `List` instance. This may not be a problem in practice as realm-js appears to make the same assumption about collections always having an object schema, but I figured I'd mention it.

appden: It was definitely worth mentioning, but I think it won't be a problem since we don't want to represent default-constructed collections in the JS API.

src/js_results.hpp
appden: I wasn't sure if was worth adding a new API on `List` just for this, or to add either an implicit or explicit conversion.

bmunkholm:
Please move it to the Review column then :-)

PR #516, opened 2016-06-30
Fix crash from calling objects() after close()

appden:
The `read_group()` method was called when m_shared_group is NULL. The `verify_open()` check probably should be added in more places, but calling it all over the place is probably not ideal.

### DIFF
--- a/src/object-store/src/shared_realm.cpp
+++ b/src/object-store/src/shared_realm.cpp
@@ -205,6 +205,8 @@ Realm::~Realm()
 
 Group *Realm::read_group()
 {
+    verify_open();
+
     if (!m_group) {
         m_group = &const_cast<Group&>(m_shared_group->begin_read());
     }
@@ -315,6 +317,13 @@ void Realm::verify_in_write() const
 {
     if (!is_in_transaction()) {
         throw InvalidTransactionException("Cannot modify persisted objects outside of a write transaction.");
+    }
+}
+
+void Realm::verify_open() const
+{
+    if (is_closed()) {
+        throw InvalidTransactionException("Can't read a Realm that was closed.");
     }
 }
 
--- a/src/object-store/src/shared_realm.hpp
+++ b/src/object-store/src/shared_realm.hpp
@@ -133,6 +133,7 @@ namespace realm {
         thread_id_t thread_id() const { return m_thread_id; }
         void verify_thread() const;
         void verify_in_write() const;
+        void verify_open() const;
 
         bool can_deliver_notifications() const noexcept;
 
@@ -140,7 +141,7 @@ namespace realm {
         // Realm after closing it will produce undefined behavior.
         void close();
 
-        bool is_closed() { return !m_read_only_group && !m_shared_group; }
+        bool is_closed() const { return !m_read_only_group && !m_shared_group; }
 
         // returns the file format version upgraded from if an upgrade took place
         util::Optional<int> file_format_upgraded_from_version() const;
--- a/tests/js/realm-tests.js
+++ b/tests/js/realm-tests.js
@@ -654,6 +654,10 @@ module.exports = {
         TestCase.assertThrows(function() {
             realm.objects(InvalidPerson);
         });
+        TestCase.assertThrows(function() {
+            realm.close();
+            realm.objects('PersonObject');
+        });
     },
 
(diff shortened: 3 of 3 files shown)


src/object-store/src/shared_realm.cpp
```diff
@@ -318,6 +320,13 @@ void Realm::verify_in_write() const
     }
 }
 
+void Realm::verify_open() const
+{
+    if (is_closed()) {
+        throw InvalidTransactionException("Can't read a Realm that was closed.");
```
bdash: The comments in the header say that continuing to call methods on `Realm` after it is closed is undefined behavior. If we're going to change this policy, and adding checks of this nature feels like a change, I think we should do so consistently on public entrypoints of this class.

Since `begin_transaction` calls `read_group` without first checking `is_closed`, this could result in beginning a write transaction throwing an exception about reading a Realm. That seems like it could be confusing?

appden: I agree this should be improved, though I think we should add `verify_open()` guards in the right places around `SharedRealm`. There are so many avenues through which `SharedRealm` is accessed, that it would be burdensome for a binding to methodically check `is_closed()` for every relevant method call. I'm going to be MIA, so I leave this PR up to you and @alazier to find the best solution. The test I added below will cause a crash unless this is resolved somehow. :smile:

simonask:
So me and @bdash actually ran into this very issue yesterday.

Currently it is possible to generate a segfault by calling `close()` and then doing anything with the opened Realm. That's clearly not acceptable. I think the expected outcome of interacting with a closed Realm is an exception. :-)

The alternative is to remove the `close()` method.

alazier:
@simonask we can't remove the close method, as you need to be able to have control over when a file is releases as in JS garbage collection can't be triggered/forced. @kristiandupont is supposed to be working on this in the object store, ie adding checks before any methods which would be unsafe after calling close. Or if @bdash has time to tackle this that would work as well

bdash:
`Realm::close` is hard to use safely as-is given that `Realm` instances are cached per thread. If two different parts of one app happen to open the same Realm via separate calls to the JS Realm constructor, then one of them calls `close()` on the instance they receive, the other code will be in for a very bad time (it'll crash at present, but after a change like this it'd throw an exception). I think we need to find a less fragile way to solve the problem that `Realm::close` was added to solve.

alazier:
@bdash any suggestions on other solutions? I can't really think of any - there are simply times where a user needs to relinquish access to a Realm file. We could consider not caching/reusing Realms in JS as then the issue you bring up would no longer exist.

bdash:
My understanding is that what JS needs is the ability to disassociate the lifetime of the C++ objects from that of the JS wrappers around them, avoiding the need to wait until the JS engine's garbage collector happens to deallocator the wrapper object to release the C++ objects. One option would be to disable caching, as you mention. Another could be for the JS `close` method to drop the strong references that the JS wrappers have to their corresponding C++ objects, allowing their reference counts to drop ahead of the JS wrapper's deallocation.

simonask:
Note that several threads in the traditional sense are not required to trigger this unfortunate situation - Node.js apps with their events and callbacks etc. are defacto still concurrent despite being single-threaded, so the same issue can occur there.

A simple solution could be to change `close()` to really just decrement a refcount somewhere, as @bdash suggests. Note that the server side components of RMP really do need to be able to manage the lifetime of open Realm files, because it's easy to get into a situation where millions of Realm files are being opened, and the V8 GC has no chance to schedule a collection since it knows nothing about the mapped memory and allocation of file descriptors inside Realm.

alazier:
@simonask is turning off caching a better solution for RMP? It is definitely simpler and would result in resources actually being released when close is called. This is somehing @sorenvind had asked for and I don't think caching is necessary given that JS is single threaded and given the fact that there are no convenience methods which retrieve cached Realms.

simonask:
@alazier RMP actually already has some form of caching of open Realms internally, so perhaps it is exactly the right solution.

sorenvind:
Killing caching definitely removes a lot of possible complexity, which is why I asked for it. I'd much rather have the behaviour be explicit and predictable in the sense that opening a realm uses a bit more resources but can't influence existing instances.

kristiandupont:
#718 replaces this.

PR #523, opened 2016-07-08
Support latest react-native version

alazier:
Users have confirmed that this fixes issues with rn 0.28+


react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -55,20 +56,21 @@ - (JSGlobalContextRef)ctx;
...
-        assert(signature);
 
-        // React Native 0.14+ expects a JSContext here, but we also support 0.13.x, which takes a JSGlobalContextRef.
-        if (!strcmp([signature getArgumentTypeAtIndex:2], "@")) {
-            JSContext *context = [[JSContext alloc] init];
-            rctJSContext = [[RCTJavaScriptContext alloc] initWithJSContext:(void *)context];
+        NSMethodSignature *signature = [RCTJavaScriptContext instanceMethodSignatureForSelector:@selector(initWithJSContext:onThread:)];
+        if (signature) {
+            // for RN 0.28.0+
+            rctJSContext = [executor context];
```
heracek: In my opinion this is too hackish. You are deducing that when there is support for `initWithJSContext:onThread:` there is also `[executor context]`. This can break easily in future RN release.

I'd suggest to move `[executor context]` to the top of the function, because it should be **the** way to ask for context. See my [commit](https://github.com/heracek/realm-js/commit/fec5ab408f0b6d4ca18306bf105af96539d9e72c).

`[executor context]` was introduces in [this commit](https://github.com/facebook/react-native/commit/a5d82c282336c80addf1225db1346c764b6794f7#diff-b2c406c59f5d64abb6f8eca647b69e8aR218) and it seems that it [remained the same](https://github.com/facebook/react-native/blame/master/React/Executors/RCTJSCExecutor.mm#L376) until [this commit](https://github.com/facebook/react-native/commit/1ebd9c5deaa453a56e73f4fc34165f08de18bd79#diff-24171be54ca5e6987e6e9b2f870fe8a3L307) removed lazy creation of context (`if (!_context) {...}`).

Commit removing lazy creation of context is only on master for now. And maybe it would mitigate any problems with context being `nil`, because context is created as part of `setUp` method.

There is also other hack to ivar in my commit, but I don't know how now to get `javaScriptThread` other way.

Whet do you think?

heracek: And I must say that my commit is **not** tested. It just suggests how `RealmReactGetJSGlobalContextForExecutor` may look.

I don't dev environment set up for `realm-js`.
I'd need more time for that, I'll have time for that next week.

alazier: I agree that these changes can be cleaned up a bit - will incorporate some of your suggestions into the pr. Based on @jerson 's testing though we will still need to populate the `_jscWrapper` ivar as well as this is expected to exist on calls to `-context`. Alternatively we can revert to the old methodology we were using where we would create and set the context manually rather than depending on `-context`. Will hopefully get the pr updated today so you guys can test this out as I am still unable to trigger this code path.

In the longer term we need to be making a pr against ReactNative to remove all dependencies on RN implementation details. We tried this at one point the but unfortunately there was a race condition with the changes that were merged into RN, so we had to revert to this strategy to initialize the Realm apis.

appden: Just so you know, the `RCTJavaScriptContextCreatedNotification` had been reliably posted right after creating the `JSContext` since React Native v0.26.0 (see [e195bf9](https://github.com/facebook/react-native/commit/e195bf9495c5f2cb4d35b2b9f777dcd2464ec5b9)). Also, a "reliable hack" would be to swizzle `-[JSContext initWithVirtualMachine:]`. :smile:

alazier: @appden - last time I tried it `RCTJavaScriptContextCreatedNotification` is posted async and in some cases would arrive after RN tried to call our apis which obviously wouldn't work. Probably worth trying this out again since the RN code base has changed significantly since. The `JSContext` hack sounds terrible but I think your right about its reliability. Only issue is it might be too reliable - ie how do we prevent other `JSContext`s outside of RN from being initialized with our apis?

alazier:
@fealebenpae @bdash can one of you review?

react-native/android/src/main/jni/Application.mk
```diff
@@ -12,6 +12,7 @@ APP_CPPFLAGS := -std=c++14
 APP_CPPFLAGS += -frtti
 APP_CPPFLAGS += -fexceptions
 APP_CPPFLAGS += -DREALM_HAVE_CONFIG
+APP_CPPFLAGS += -fomit-frame-pointer
```
bdash: Why is this necessary?

react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -55,17 +56,18 @@ - (JSGlobalContextRef)ctx;
...
     if (!rctJSContext && create) {
         Class RCTJavaScriptContext = NSClassFromString(@"RCTJavaScriptContext");
-        NSMethodSignature *signature = [RCTJavaScriptContext instanceMethodSignatureForSelector:@selector(initWithJSContext:)];
-        assert(signature);
 
-        // React Native 0.14+ expects a JSContext here, but we also support 0.13.x, which takes a JSGlobalContextRef.
-        if (!strcmp([signature getArgumentTypeAtIndex:2], "@")) {
-            JSContext *context = [[JSContext alloc] init];
-            rctJSContext = [[RCTJavaScriptContext alloc] initWithJSContext:(void *)context];
+        NSMethodSignature *signature = [RCTJavaScriptContext instanceMethodSignatureForSelector:@selector(initWithJSContext:onThread:)];
```
bdash: Why is this using `+instanceMethodSignatureForSelector:`? Isn't `+instancesRespondToSelector:` what you actually want to know here, since you're not verifying anything about the selector beyond its existence?

react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -55,17 +56,18 @@ - (JSGlobalContextRef)ctx;
...
+        if (signature) {
+            // for RN 0.28.0+
+            rctJSContext = [[RCTJavaScriptContext alloc] initWithJSContext:[JSContext new] onThread:[NSThread currentThread]];
         }
         else {
-            JSGlobalContextRef ctx = JSGlobalContextCreate(NULL);
-            rctJSContext = [[RCTJavaScriptContext alloc] initWithJSContext:ctx];
+            // for RN < 0.28.0
+            NSMethodSignature *oldSignature = [RCTJavaScriptContext instanceMethodSignatureForSelector:@selector(initWithJSContext:)];
```
bdash: Ditto about `+instanceMethodSignatureForSelector:` vs `+instancesRespondToSelector:` here.

bdash:
Are the CI failures legit, or CI flakiness?

react-native/android/src/main/jni/Application.mk
alazier: Was getting compile errors in the assembly in this file https://github.com/realm/realm-js/blob/master/src/android/jsc_override.cpp before this change - I assume it is due to using a newer version of the ndk.

react-native/ios/RealmReact/RealmReact.mm
alazier: We were previously using `+instanceMethodSignatureForSelector:` as we were previously verifying the arguments and I simply left the code the same. Happy to change it but functionally this is still correct?

alazier:
@bdash the CI failures the the current bane of my existence. Adroid has been failing for a while now due to device issues. The react-tests/example failures are more recent and I haven't yet figured out the issue. If you have any ideas or ways to approach fixing this I'm all ears. All the tests currently pass locally.

react-native/ios/RealmReact/RealmReact.mm
bdash: It probably gives correct behavior as-is, it's just doing its work in a non-obvious way that performs unnecessary work. It'd be worth changing.

bdash:
👍

react-native/ios/RealmReact/RealmReact.mm
alazier: Will switch it over and merge.

PR #533, opened 2016-07-18
Remove usage of Realm.Types in the spec and deprecate it

mrackwitz:
Spotted that the source code says, `Realm.Types` is useless and should be removed. I took a stab at making this possible. The way to deprecate that in the last commit might be debatable. I've also considered to make the property `configurable` and let it overwrite itself, so that it prints this warning only once.

/c @alazier @appden

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -41,18 +41,29 @@ if (typeof Realm != 'undefined') {
 Object.defineProperties(realmConstructor.Collection.prototype, arrayMethods);
 
 // TODO: Remove this now useless object.
+var types = Object.freeze({
+    'BOOL': 'bool',
+    'INT': 'int',
+    'FLOAT': 'float',
+    'DOUBLE': 'double',
+    'STRING': 'string',
+    'DATE': 'date',
+    'DATA': 'data',
+    'OBJECT': 'object',
+    'LIST': 'list',
+});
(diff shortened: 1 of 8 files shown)


lib/index.js
```diff
@@ -41,18 +41,28 @@ if (typeof Realm != 'undefined') {
...
-        'DATA': 'data',
-        'OBJECT': 'object',
-        'LIST': 'list',
-    })
+    get: function() {
+        var stack = new Error().stack.split("\n").slice(2).join("\n");
+        var msg = '`Realm.Types` is deprecated! Please specify the type name as lowercase string instead!\n'+stack;
+        if (console.warn != undefined) {
+            console.warn(msg);
+        } else {
```
alazier: else should be on the next line.

alazier:
This looks good. Need to figure out what is going on with ci though.

alazier:
Need to update the changelog

mrackwitz:
I addressed your feedback here. Are the remaining CI failures unrelated?

alazier:
Changes look good. No idea what is going on with the failures for react-tests. Different tests fail for debug and release....

alazier:
retest this please

alazier:
retest this please

alazier:
retest this please

PR #535, opened 2016-07-18
Update to the latest objectstore code

tgoyne:
@alazier


src/js_list.hpp
```diff
@@ -207,7 +207,7 @@ void ListClass<T>::snapshot(ContextType ctx, ObjectType this_object, size_t argc
     validate_argument_count(argc, 0);
 
     auto list = get_internal<T, ListClass<T>>(this_object);
-    return_value.set(ResultsClass<T>::create_instance(ctx, *list, false));
+    return_value.set(ResultsClass<T>::create_instance(ctx, realm::Results(list->get_realm(), list->get_query()).snapshot()));
```
bdash: This should be able to use `List::snapshot`.

src/js_results.hpp
```diff
@@ -74,33 +72,14 @@ struct ResultsClass : ClassDefinition<T, realm::Results, CollectionClass<T>> {
...
-    return create_object<T, ResultsClass<T>>(ctx, results);
-}
-
-template<typename T>
-typename T::Object ResultsClass<T>::create_instance(ContextType ctx, SharedRealm realm, const ObjectSchema &object_schema, Query query, bool live) {
-    auto results = new realm::Results(realm, object_schema, std::move(query));
-    results->set_live(live);
-
-    return create_object<T, ResultsClass<T>>(ctx, results);
+    return create_object<T, ResultsClass<T>>(ctx, new realm::Results(realm, *table));
```
bdash: Could move `realm` here.

src/js_list.hpp
alazier: Does this get moved automatically?

src/js_results.hpp
```diff
@@ -111,18 +90,13 @@ typename T::Object ResultsClass<T>::create_filtered(ContextType ctx, const U &co
...
-        args.push_back(arguments[i]);
-    }
+    std::vector<ValueType> args(&arguments[1], &arguments[argc]);
 
     parser::Predicate predicate = parser::parse(query_string);
     query_builder::ArgumentConverter<ValueType, ContextType> converter(ctx, realm, args);
     query_builder::apply_predicate(query, predicate, converter, *realm->config().schema, object_schema.name);
 
-    return create_instance(ctx, realm, object_schema, std::move(query));
+    return create_instance(ctx, realm::Results(realm, std::move(query)));
```
alazier: Same question here about implicit move.

alazier:
JS changes look fine assuming the Results objects get moved automatically. Also just wondering, did you make these changes by using `git subtree` from master or with some other strategy?

There is something not working at the moment with ci and the react-native tests so will need to run those manually to make sure we aren't missing something.

alazier:
Looks like the `realmjs` tests failures are legitimate though https://ci.realm.io/job/js_pr/configuration=Debug,target=realmjs/942/console

alazier:
failure is in `testResultsDeletedObjects`

src/object-store/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ include(CompilerFlags)
 include(Sanitizers)
 
 include(RealmCore)
-set(REALM_CORE_VERSION "1.1.0" CACHE STRING "")
+set(REALM_CORE_VERSION "1.2.0" CACHE STRING "")
```
bdash: I suspect we'll want to update realm-js's core version to 1.2.0 as well.

bdash: This might explain the test failures. Object store no longer has the workaround for realm/realm-core#1837, as it was fixed in core v1.2.0.

src/js_list.hpp
tgoyne: Yes, it's an unnamed temporary and thus not an lvalue. Only things that can be assigned to need to be explicitly converted to rvalues.

tgoyne:
> Also just wondering, did you make these changes by using git subtree from master or with some other strategy?

I did `git -c /src/realm-object-store format-patch c16e841 --stdout | git am --directory src/object-store -c /src/realm-js --reject`.

alazier:
Is that going to cause any issues/conflicts when using subtree in the future, or is that basically the same thing?

Also @bdash was right, after updating to the latest core the tests seem to pass! I can push those changes if you would like.

tgoyne:
git-subtree appears to have last been used to pull in changes in November, so all of the things since then are going to break its automatic handling anyway.

alazier:
That's not true at all. We have been using `git subtree` consistently. Last time we updates was last month - https://github.com/realm/realm-js/commit/81a32909c170ee580c4ca1121bfade9d7cd81b5c

That being said things seem to be pretty similar. Trying to run `it subtree pull -P src/object-store/ object-store master` on your branch only produces two minor conflicts, so I don't think there is any major incompatibilities.

scripts/download-core.sh
```diff
@@ -4,7 +4,7 @@ set -e
 set -o pipefail
 
 # Set to "latest" for the latest build.
-: ${REALM_CORE_VERSION:=1.1.2}
+: ${REALM_CORE_VERSION:=1.4.0}
```
alazier: Do we need to change the downloaded tarball filename format as well? In the object-store changes we now seem to be using `xz` extension rather than `bz2`. Also we may need to replace the `gz` extensions for Darwin and Node with `xz`?

tgoyne:
Ah, git-subtree must have improved then, if it no longer needs the extra metadata in the commit message to avoid breaking horribly. Good to know.

bdash:
👍

alazier:
tests all pass locally 👍

Issue #521, opened 2016-07-06
React-Native - Error: Could not locate the bindings file. Tried:


secretyouth:
Hi all,

Hopefully someone can chime in and help me figure this out. I am using React-Native (0.28.0) & Realm (0.14.1) since installing ream my react native tests have been failing due to the following error.

I've literally hit a wall, when I remove the react components that are requiring Realm, they test fine. But with realm the error bellow is returned.

Any help or ideas are much appreciated, as I have tried remove node_modules, npm rebuild & downgraded realm. 

**Package.json**
{
  "name": "reactApp",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "test": "mocha --require react-native-mock/mock.js --compilers js:babel-core/register --recursive test/*.js"
  },
  "dependencies": {
    "lodash": "^4.13.1",
    "react": "15.2.0",
    "react-native": "^0.28.0",
    "realm": "^0.14.1"
  },
  "devDependencies": {
    "babel-core": "^6.10.4",
    "babel-preset-react-native": "^1.9.0",
    "chai": "^3.5.0",
    "enzyme": "^2.3.0",
    "mocha": "^2.5.3",
    "mockery": "^1.7.0",
    "react": "^15.2.0",
    "react-addons-test-utils": "^15.2.0",
    "react-dom": "^15.2.0",
    "react-native-mock": "^0.2.4"
  }
}

**Error when running npm test**

> mocha --require react-native-mock/mock.js --compilers js:babel-core/register --recursive test/*.js

/Users/jamiehayman/someApp/reactApp/node_modules/bindings/bindings.js:91
  throw err
  ^

Error: Could not locate the bindings file. Tried:
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/build/realm.node
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/build/Debug/realm.node
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/build/Release/realm.node
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/out/Debug/realm.node
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/Debug/realm.node
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/out/Release/realm.node
 → /Users/jamiehayman/someApp/reactApp/node_modules/realm/Release/realm.node
[...]

PR #537, opened 2016-07-21
Get the path to nan relatively

fealebenpae:
Under npm3 the `realm` and `nan` modules might end up as siblings so the best way to get the path to `nan` is to require it from within the `realm` module. The `nan` module prints its path so it can be captured by an external build system like in this patch.


fealebenpae:
Related to #521.

alazier:
👍

PR #546, opened 2016-08-01
Align build with downstream

fealebenpae:


alazier:
Changes look fine. Would be nice to have something testing all of these code paths those - think the Docker stuff is the only thing not being tested? We need to also run the react-tests locally before merging until we figure out why we keep getting failures on ci.

fealebenpae:
retest this please

alazier:
retest this please

PR #547, opened 2016-08-02
Update to latest, greatest object store

fealebenpae:


bdash:
Let's switch to pulling in the object store as a submodule instead?

alazier:
@bdash we plan to. Have some open prs right now and want to get a sync release out in the next few days- plan to make the switch after that to avoid any unnecessary delays.

PR #548, opened 2016-08-02
Switch to pulling the object store in as a submodule

bdash:
This is using the `js` branch from realm-object-store. It'll switch to master once we've merged those changes in.

/cc @fealebenpae @alazier


alazier:
JS changes look good 👍

PR #569, opened 2016-08-16
Make it possible to subclass Realm.Object on node

### LINKED ISSUE #565: Testing with mocha, Error: Illegal constructor in native code
I'm using the instructions here to test with mocha: https://github.com/realm/realm-js/issues/521

I am seeing an error when I on write(), I'm only seeing this when I test, the app (ios) works.

Here's what my test is doing:

```
import Realm from 'realm';
class CurrentUser extends Realm.Object {}
CurrentUser.schema = {
    name: 'CurrentUser',
    properties: {
        username:  'string'
    }
};
export default new Realm({schema: [CurrentUser, User, Test, Athlete, Team]});

// this goes in the describe block
       realm.write(() => {
            currentUser = realm.create('CurrentUser', {
                username: 'hi'
            });
        });
```

when I run it, I get:

`
try realm:
     Error: Illegal constructor
      at Error (native)
`

fealebenpae:
/cc @alazier 

Fixes #565

### DIFF
--- a/src/node/node_class.hpp
+++ b/src/node/node_class.hpp
@@ -223,17 +223,17 @@ inline void ObjectWrap<ClassType>::setup_property(v8::Local<TargetType> target, 
 
 template<typename ClassType>
 inline void ObjectWrap<ClassType>::construct(Nan::NAN_METHOD_ARGS_TYPE info) {
-    if (!info.IsConstructCall()) {
-        Nan::ThrowError("Constructor must be called with new");
-    }
     if (reinterpret_cast<void*>(s_class.constructor)) {
         auto isolate = info.GetIsolate();
         auto arguments = get_arguments(info);
         v8::Local<v8::Object> this_object = info.This();
         info.GetReturnValue().Set(this_object);
 
-        auto wrap = new ObjectWrap<ClassType>();
-        wrap->Wrap(this_object);
+        // only initialize this if this is a direct construct call, not a super() call
+        if (info.IsConstructCall()) {
+            auto wrap = new ObjectWrap<ClassType>();
+            wrap->Wrap(this_object);
+        }
 
         try {
             s_class.constructor(isolate, this_object, arguments.size(), arguments.data());
@@ -242,7 +242,7 @@ inline void ObjectWrap<ClassType>::construct(Nan::NAN_METHOD_ARGS_TYPE info) {
             Nan::ThrowError(node::Exception::value(isolate, e));
         }
     }
-    else {
+    else if (info.IsConstructCall()) {
         Nan::ThrowError("Illegal constructor");
     }
 }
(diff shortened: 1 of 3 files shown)


fealebenpae:
please retest this

fealebenpae:
retest this please

src/node/node_class.hpp
```diff
@@ -223,17 +223,17 @@ inline void ObjectWrap<ClassType>::setup_property(v8::Local<TargetType> target,
...
 template<typename ClassType>
 inline void ObjectWrap<ClassType>::construct(Nan::NAN_METHOD_ARGS_TYPE info) {
-    if (!info.IsConstructCall()) {
-        Nan::ThrowError("Constructor must be called with new");
-    }
     if (reinterpret_cast<void*>(s_class.constructor)) {
         auto isolate = info.GetIsolate();
         auto arguments = get_arguments(info);
         v8::Local<v8::Object> this_object = info.This();
         info.GetReturnValue().Set(this_object);
```
alazier: should the constructor also return this?

tests/js/object-tests.js
```diff
@@ -515,5 +515,30 @@ module.exports = {
         TestCase.assertEqual(realm.objects('Date')[1].currentDate.getTime(), -10000);
         TestCase.assertEqual(realm.objects('Date')[2].currentDate.getTime(), 1000000000000);
         TestCase.assertEqual(realm.objects('Date')[3].currentDate.getTime(), -1000000000000);
+    },
+    
+    testInheritance: function() {
+        // the equivalent of `class InheritingObject extends Realm.Object`
+        function InheritingObject() {
```
alazier: why not just define the class normally?

fealebenpae: Although that's not the case on React Native on older iOS versions, so babel kicks in there. Should we (or do we?) run ours tests through babel?

alazier: We could do that, or only run the ES6 tests for environments where it is supported.

src/node/node_class.hpp
fealebenpae: I don't think so - the return value of the constructor function is disregarded in `new Foo()` and `Foo.apply(this, arguments)` expressions.

I guess I can remove this.

tests/js/object-tests.js
fealebenpae: Personally I think we're fine with the "manually transpiled" code right now, but it is a good idea to add a native `class` declaration because it might have different semantics. I'll see how easy it would be to add an ES6-only test suite.

fealebenpae: It seems like we already have tests for this on React with ES6 classes, courtesy of @appden: https://github.com/realm/realm-js/commit/26d2c169a9adb4f541433c46d02b5351a941c08b.

Too bad we weren't running the React tests on Node 😭

appden: @fealebenpae @alazier FYI, native ES6 classes will have issues in Node, and I think fixing it will require removing the "feature" where the constructor is automatically called.

fealebenpae: Ah, yes, in ES6 class constructors cannot be called as functions. But that would be a problem on iOS as well, once people start using native ES6 classes there.

appden: Exactly, so I seriously doubt many people are using this relatively undocumented "feature", so I definitely suggest dropping it and maybe in the future provide an optional special method that could be called whenever a new Realm object is created if somebody really badly needs this functionality.

PR #572, opened 2016-08-19
Don't run npm install multiple times

alazier:
@fealebenpae 

Right now `npm install` is run multiple times unnecessarily when running tests. This reduces the number of runs to 1 per test.


fealebenpae:
👍

alazier:
retest this please

PR #576, opened 2016-08-24
Update to latest object store

fealebenpae:


alazier:
👍 assuming tests pass. should probably run the android and react-tests locally

PR #609, opened 2016-10-10
Make the node addon embeddable

fealebenpae:
So that it can be linked in other node addons

/cc @alazier


src/node/node_class.hpp
```diff
@@ -49,6 +49,8 @@ class ObjectWrap : public Nan::ObjectWrap {
         return Nan::New(js_template);
     }
 
+    static void construct(const v8::FunctionCallbackInfo<v8::Value>&);
```
alazier: Instead of hardcoding this to v8, we should instead get these types from the `Types` metatype. This way we can reuse all of this code for v8 and nan by just changing the `Types` type.

fealebenpae: We should always use the v8 callback types over the Nan ones. We won't need to implement the callbacks for both v8 and Nan. Besides - there isn't an alias for `const v8::FunctionCallbackInfo<v8::Value>&` in Types.

alazier:
We need to update this to use a submodule commit from the public repo. The latest changes have been merged to the js branch.

binding.gyp
```diff
@@ -1,50 +1,46 @@
...
-          "src/object-store/src/impl/apple",
-          "src/object-store/src/parser",
-          "src/object-store/external/pegtl"
-        ]
-      },
       "target_name": "object-store",
-      "dependencies": [ "realm-sync" ], # sync also includes core
+      "dependencies": [ "realm-core" ],
       "type": "static_library",
-      "include_dirs": [ "<@(object-store-include-dirs)" ],
```
alazier: Why do we need to separate out all dirs individually?

scripts/download-core.sh
```diff
@@ -105,6 +103,7 @@ else
     echo "The core library seems to be up to date."
 fi
 
+[ -z "$SYNC_DIR" ] && exit 0 # not downloading sync
```
fealebenpae: We don't have a publicly-accessible build of sync for node.

src/js_realm.hpp
```diff
@@ -29,13 +29,16 @@
 #include "js_results.hpp"
 #include "js_schema.hpp"
 #include "js_observable.hpp"
+
+#if REALM_ENABLE_SYNC
```
alazier: We should always be building with sync enabled.

scripts/download-core.sh
alazier: We and CI should download the privately available version.

binding.gyp
fealebenpae: You'll notice I list the include directories twice - once here and once in `all_dependent_settings`. I used to use a variable and reference that, but gyp doesn't resolve the absolute paths of elements in list variables when expanding them in the downstream target via `all_dependent_settings` - instead of using the `object-store` target as a base directory, it uses the importing target, which results in wrong absolute paths.

alazier:
I think we might want a test which ensures that things remain submodulable when we make future changes to the build system.

alazier:
Can we just remove the changes which disable building for node and then merge this?

fealebenpae:
Let's discuss this in real time at the next status hangout.

PR #645, opened 2016-11-09
User persistence and sync tests

### LINKED ISSUE #636: Get node-pre-gyp working with react-native
We need to support building a combined react-native/node binding on the public repo so we can release the node binding publicly. I assume this means porting over the node-pre-gyp build configuration to the public repo.

@radu-tutueanu @fealebenpae - any guidance or help on getting this working would be greatly appreciated, as I think you are most familiar with how things are working in the private repo?

### LINKED ISSUE #637: Test node binding on linux and mac on ci
Right now we run tests from the `scripts/test.sh` script. Do we want to switch over to using a jenkinsfile on the public repo?

@alebsack @sorenvind @radu-tutueanu

### LINKED ISSUE #641: support building for linux with sync disabled

### LINKED ISSUE #646: Load the native module in the case we are running on node
We need to add code to `index.js` to detect the proper environment and load the native module in the case we are running on node.

alazier:
@sorenvind lets merge this once we get the node/sync tests passing on ci.

Closes #636, closes #646, closes #641, closes #637


tests/js/user-tests.js
```diff
@@ -0,0 +1,279 @@
...
+    throw new Error(`Error ${error} was not expected`);
+  }
+}
+
+// Test the given requestFunc, passing it the given callback after it's been wrapped appropriately.
+// This function makes sure that errors thrown in the async callback rejects the promise (making tests actually run).
+function callbackTest(requestFunc, callback) {
+  return new Promise((resolve, reject) => {
+    function callbackWrapper() {
+      try {
```
mrackwitz: Promises automatically catch errors thrown in the callback provided to their constructor.

test.js
```diff
@@ -23,33 +23,29 @@ const prompt = require('prompt');
 const mkdirp  = require('mkdirp');
```
mrackwitz: We probably want to remove this at some point.

alazier: For sure. We should replace it with a proper example in the examples folder.

tests/js/user-tests.js
alazier: So you are saying this isn't needed?

sorenvind: Yes, that's what Marius is saying. However this doesn't hurt, it just makes it slightly clearer what's going on if you don't know the specific api in promise creation.

mrackwitz: I don't think so. In my opinion this construct unnecessarily obfuscates the program flow. That was also the reason why I noticed it at all in the first place.

PR #647, opened 2016-11-09
Fix node tests

radu-tutueanu:
We still need to be able to build node without sync for linux.

This is why I kept the "--realm_enable_sync" flag. It might need to be added to a few of the other npm installs (react)

The download of the sync library could be improved, I want to change the format just like the core one.

@sorenvind @alazier

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -32,7 +32,7 @@
     },
     {
       "target_name": "object-store",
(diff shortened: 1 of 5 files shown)


radu-tutueanu:
At the moment, this fixes the node macOS tests.

PR #649, opened 2016-11-10
Fix node tests for linux

radu-tutueanu:


alebsack:
@alazier this is ready for review.  node-linux tests pass: https://ci.realm.io/job/js_pr/1047/

alazier:
The test and download script changes look good. I don't know enough about docker and the other stuff to review, but it seems to work!

PR #653, opened 2016-11-14
Nh/update android

### LINKED ISSUE #639: get react-native working reliably on ci

nhachicha:
* Using `dependencies.list` to figure out the Core/Sync version
* Update `ReactExample`
* Add a switch to build RN Android with Sync or just core
     - `/realm-js/react-native/android/gradlew clean publishAndroid` will build using Core as a dependency only
   -  `/realm-js/react-native/android/gradlew clean publishAndroid  -PbuildWithSync=true` will pull Core + Sync & build them as static library (note: this only compile using Sync static library it doesn't enable sync yet)

@alazier @kristiandupont 

This fixes #639

### DIFF
--- a/tests/react-test-app/package.json
+++ b/tests/react-test-app/package.json
@@ -8,6 +8,7 @@
   "dependencies": {
     "react": "15.3.2",
     "react-native": "^0.37.0",
+    "react-native-fs": "^1.1.0",
     "realm": "file:../..",
     "realm-tests": "file:../js",
     "xmlbuilder": "^4.2.1"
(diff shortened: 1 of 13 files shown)


tests/react-test-app/package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "react": "15.3.2",
     "react-native": "^0.37.0",
+    "react-native-fs": "^1.1.0",
```
kristiandupont: What's this for?

nhachicha: this is needed for the `react-test-app` which run tests for Android `tests/react-test-app/android/settings.gradle`. We use it in `tests/react-test-app/index.android.js` to export the test result to the SD card on Android emulator.

PR #655, opened 2016-11-15
Do not build node with sync by default

radu-tutueanu:
@bigfish24 Will clarify if this is needed or not.
Not sure if anything needs to happen on the react side as well.

@alazier @kristiandupont

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -123,7 +123,7 @@ case "$TARGET" in
   ;;
 "react-tests")
   if ! [ -z "${JENKINS_HOME}" ]; then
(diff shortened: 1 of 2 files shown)


alazier:
Shouldn't we only set this for linux? Is that even possible with how the build system is set up?

radu-tutueanu:
The idea was to maybe set it to OS X as well for tomorrow. I changed the tests to run "npm install --realm_enble_sync", so the tests would not be affected.

bigfish24:
@radu-tutueanu Paul said we should hold back sync from NPM install until Thursday. Keeps the message focused.

PR #662, opened 2016-11-16
Add runtime analytics for node

kristiandupont:


examples/NodeInterprocess/listener.js
```diff
@@ -10,6 +28,8 @@ let winstonRealm = new Realm({
 winstonRealm.objects('Log').filtered('level = "error"').addListener((logs, changes) => {
   changes.insertions.map((index) => {
     let log = logs[index];
+
+    // eslint-disable-next-line no-console
```
alazier: can we just get rid of this rule?

PR #664, opened 2016-11-17
Added tests directory to package.json

alebsack:
@alazier this will help test the node modules on different node versions and platforms after building.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "react-native",
     "scripts",
     "src",
+    "tests",
     "vendor",
     "binding.gyp",
     "dependencies.list"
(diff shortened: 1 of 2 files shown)


alazier:
I just pushed a changed to remove the ios files/app from the package. Think we should do some more cleanup before shipping this but overall I think this is good as long as it doesn't increase the package size. Does anyone else have thoughts on this?

alebsack:
After installing the npm module for node, I'd like to be able to:
```
cd tests
npm install
npm run test-nosync
```

alebsack:
@fealebenpae @radu-tutueanu Do you have any thoughts?

radu-tutueanu:
That sounds good to me. 👍

PR #666, opened 2016-11-17
Jenkinsfile to replace js_pr

alebsack:
This should be a starting point for replacing `js_pr` in CI.  It's not aimed at fixing all of the failing scenarios, but I'd like to get this in as a way to move forward with fixing them.

@emanuelez @alazier @radu-tutueanu @nhachicha


Dockerfile
```diff
@@ -1,8 +1,34 @@
...
 
-RUN apt-get update && \
-    apt-get install -y curl && \
+# Install the JDK
+# We are going to need some 32 bit binaries because aapt requires it
+# file is need by the script that creates NDK toolchains
+ENV DEBIAN_FRONTEND noninteractive
+RUN dpkg --add-architecture i386 && \
+    apt-get update -qq && \
+    apt-get install -y file git curl wget zip unzip bsdmainutils strace lsof \
```
emanuelez: I prefer to have one package per line and in alphabetical order. That usually helps while checking if a particular package is installed and avoiding duplicates.

Jenkinsfile
```diff
@@ -0,0 +1,142 @@
...
+      } else {
+        echo "Building release: '${gitTag}'"
+        setBuildName("Tag ${gitTag}")
+      }
+    }
+    echo "version: ${version}"
+  }
+}
+
+def reportStatus(target, state, message) {
```
emanuelez: Why do we need this?

alebsack: We can set the individual status of each branch of execution in the PR.

Dockerfile
```diff
@@ -1,8 +1,34 @@
 FROM ubuntu:xenial
 
-RUN apt-get update && \
-    apt-get install -y curl && \
+# Install the JDK
+# We are going to need some 32 bit binaries because aapt requires it
+# file is need by the script that creates NDK toolchains
+ENV DEBIAN_FRONTEND noninteractive
+RUN dpkg --add-architecture i386 && \
```
nhachicha: We also need to install Android NDK & SDK similar to https://github.com/realm/realm-java/blob/master/Dockerfile for Android

Jenkinsfile
```diff
@@ -0,0 +1,142 @@
+#!groovy
```
nhachicha: for Android we need to have access to the device to run tests inside, currently this is how `realm-java` is setup https://github.com/realm/realm-java/blob/master/Jenkinsfile#L43

The current android test also extract a file `tests/react-test-app/tests.xml` that the current conf in CI uses to publish Unit Test result
screen shot 2016-11-23 at 18 44 09

Dockerfile
```diff
@@ -1,8 +1,75 @@
 FROM ubuntu:xenial
 
-RUN apt-get update && \
-    apt-get install -y curl && \
+# Install the JDK
+# We are going to need some 32 bit binaries because apt requires it
```
emanuelez: It's actually `aapt`, not `apt`. It stands for Android Asset Packaging Tool.

Jenkinsfile
```diff
@@ -0,0 +1,142 @@
...
+    $class: 'GitHubCommitStatusSetter',
+    contextSource: [$class: 'ManuallyEnteredCommitContextSource', context: target],
+    statusResultSource: [$class: 'ConditionalStatusResultSource', results: [[
+      $class: 'AnyBuildResult', message: message, state: state]]
+    ],
+    reposSource: [$class: 'ManuallyEnteredRepositorySource', url: 'https://github.com/realm/realm-js']
+  ])
+}
+
+def doInside(script, target, postStep = null) {
```
emanuelez: Why are we doing docker stuff in bash? Is it something we're gonna need on a developer's desk?

Dockerfile
alebsack: That's not confusing at all!

Jenkinsfile
alebsack: IMO it's functionally the same, but yes, it allows developers to run any of the `test.sh` stuff in a local docker container.

Jenkinsfile
radu-tutueanu: Nice. Though i guess that if we still see issues with the GitHub api rate limit, we might want to disable it

alebsack: I don't think it's making any more GitHub API requests than the existing js_pr job.  If so, we could eliminate the PENDING status.

radu-tutueanu: ok, we will see. I don't think js_pr jobs have been an issue until now.

alebsack:
This is RTM.  The MacOS tests are still a nondeterministic.  🎱

PR #667, opened 2016-11-17
Support running realm-js in Electron renderer

### LINKED ISSUE #673: Node (v8) Environment Checks
In `lib/index.js` an environment check is performed. 
`if (typeof Realm != 'undefined') {
} else if (typeof window != 'undefined') {
} else if (typeof process == 'object' && (('' + process) == '[object process]' || typeof jest == 'object'){}`
Since window is checked before process, applications such as electron (browser/node hybrid) are unable to run Realm. The solution would be to check for process before window. I can submit a PR for this if that would be helpful :).

BrandonSmith:
Electron is a combination of a node and
Chromium processes. Chromium processes
can have node bindings enabled and can
therefore run most npm modules seamlessly.

The current `lib/index.js` bootstrapping
logic results in a ReferenceError on the
following statement:

    typeof Realm != 'undefined'

Thus, checking whether running under node
first allows realm-js to run inside the
Electron renderer (Chromium) process.

Attached is a screenshot of the modified example demonstrated in the [realm.io blog post](https://realm.io/news/first-object-database-realm-node-js-server/) announcing Node support. Namely, a realm database is shared between Electron's main node process and the Chromium [BrowserWindow](http://electron.atom.io/docs/api/browser-window/) process that has node integration.

vh9gxfxuuq

Electron + Realm is potentially a big win. Being able to concurrently read and write to the same database is quite powerful. Chromium, of course, has IndexedDB, but it is inaccessible to the Electron's main node process without RPC orchestration. Likewise, many Electron projects leverage node sqlite bindings or PouchDB (with LevelDB bindings). But those also face the same problem of concurrency and RPC orchestration.

This fixes #673

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -22,24 +22,36 @@ function node_require(module) {
     return require(module);
 }
 
-var realmConstructor;
+function isNode() {
+    return typeof process == 'object' && (('' + process) == '[object process]' || typeof jest == 'object')
+}
 
-if (typeof Realm != 'undefined') {
-    // The global Realm constructor should be available on device (using JavaScriptCore).
-    realmConstructor = Realm;  // eslint-disable-line no-undef
-// eslint-disable-next-line
-} else if (typeof window != 'undefined') {
-    // The userAgent will be defined when running in a browser (such as Chrome debugging mode).
-    realmConstructor = require('./browser').default; // (exported as ES6 module)
-// eslint-disable-next-line
-} else if (typeof process == 'object' && (('' + process) == '[object process]' || typeof jest == 'object')) {
+// function isElectronRenderer() {
+//     return isNode() && process.type === 'renderer'
+// }
+
+var realmConstructor;
+if (isNode()) {
     // If process is defined, we're running in node.
     // Prevent React Native packager from seeing this module.
     var binary = node_require('node-pre-gyp');
     var path = node_require('path');
-    var binding_path = binary.find(path.resolve(path.join(__dirname,'../package.json')));
+    var pkg = path.resolve(path.join(__dirname,'../package.json'));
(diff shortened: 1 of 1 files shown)


alazier:
This is awesome. We had plans to do this but you beat us to it! We would need to to fill out our CLA before accepting contributions: https://github.com/realm/realm-js/blob/master/CONTRIBUTING.md#cla

pwdl:
Do you have an idea when this functionality (i.e. electron support) will be included in an upcoming release?

If not, do you suggest pulling an 'release candidate' branch?

kristiandupont:
@pwdl it's merged to master so it will be in the next release. So you should have access to this soon.

pwdl:
@kristiandupont - Thanks. It was the 'soon' I wasn't sure about. No worries. I will hang tight or look at details of change...

BrandonSmith:
@pwdl The way I solved it in the interim is a private module `electron-realm` that replaces `lib/index.js` from this repo and bootstraps directly.

alazier:
There is also an rc release `0.15.1-rc` available which should have this fix.

Steviey:
So does this mean we can use realm.js in Electron and use the same code base on mobile with React-Native?

kristiandupont:
Electron is still not officially supported, but people are making it work. Depending on your installed node versions etc., you might have some hurdles to work through. We will offer official support soon, but we can't say exactly when yet. (And yes, the API is the same so you can share your code base)

Steviey:
Great, since every SQlite-implementation differs, this would be a very big plus point and motivation to switch to Realm. Is there a place where I can follow the Electron-progress?

kristiandupont:
#765 is probably a good issue to track if you want to stay informed regarding Electron.

PR #675, opened 2016-11-20
call cleanup on exceptions and ctr-c

alazier:
@kristiandupont @bigfish24

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -30,6 +30,15 @@ module.exports = function(realmConstructor) {
         if (realmConstructor.Sync.cleanup) {
             // FIXME: DOES THIS WORK ON BOTH NODE AND REACT NATIVE?
             process.on('exit', realmConstructor.Sync.cleanup);
+            process.on('SIGINT', function () {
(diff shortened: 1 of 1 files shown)


mrackwitz:
@sorenvind and I just discussed this change and we decided that this should be reverted, because otherwise this could to situations where failed application-logic need to be torn down and only after that the sync session may be cleaned up. So this should be handled specifically in the test harness of the JS SDKs and any other consumer of this library. We should advise in the docs, that users are responsible themselves to call `Realm.Sync.cleanup()`.

sorenvind:
Let me clarify this a bit: Currently realm-js attaches a signal handler as a side-effect of simply require-ing `Realm`. Depending on the order in which the signal handlers are called, this means that downstream users either 1) can't clean up themselves (because process.exit() is called in the signal handlers), or 2) specific exit codes are forced by realm-js. 

The API i'd like is a synchronous function `Realm.cleanup()` or `Realm.exitHandler()` that downstream users can call when they need to on exit.

alazier:
This shouldn't be reverted, but we can discuss alternative approaches.

PR #679, opened 2016-11-21
Only set the path derived from the SyncManager if none was provided

### LINKED ISSUE #678: Setting the path for a sync realm does not work
A synced realm will always be opened at a path like ./realm-object-server `./realm-object-server/7884aaaa-101e-469e-ad06-0f99e7467eb2/realm%3A%2F%2Flocalhost%3A7800%2FrealmName`

This makes it hard for the cleanup in the ROS, we know where we open realms and make sure to cleanup when needed.

@alazier

mrackwitz:
Fixes #678. /c @alazier 

* [ ] Tests
* [ ] Changelog entry

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -328,12 +328,16 @@ void RealmClass<T>::constructor(ContextType ctx, ObjectType this_object, size_t 
         else if (Value::is_object(ctx, value)) {
             ObjectType object = Value::validated_to_object(ctx, value);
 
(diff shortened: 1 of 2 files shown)


mrackwitz:
So it turns out that the object store currently has the restriction that a given user can only have the same synced URL open once as seen in #243. This doesn't necessarily need to block this PR though afaict.

alazier:
I just rebased on master. As long as our only failures are react-android and react-example we are good to merge.

PR #693, opened 2016-11-24
Test deleted files

radu-tutueanu:


radu-tutueanu:
The failures do not seem related to the changes

PR #696, opened 2016-11-24
Update OS

radu-tutueanu:
Remove duplicate files from the submodule.


kristiandupont:
Does this move us forward to the version without the duplicate files?

radu-tutueanu:
Yes, that is right

PR #707, opened 2016-11-30
Fix Android compilation (Using  EPOLL commit helper)

nhachicha:

PR #708, opened 2016-11-30
Use epoll ECH

kristiandupont:


kristiandupont:
Ah, I see now that this duplicates #707.

PR #710, opened 2016-12-01
improve resetting the iOS simulator

larkost:
This lines up our usage of the iOS simulator so that we will only start a single instance when run in CI. Prior to this up to 3 different instances could be started, but only one (unknown selection process) was actually used. This also should clean up our process of switching from one version of the simulator to another. In order to make this more predictable I have moved the reset behavior from a separate file, into the `test.sh` file.

These improvements are solely for CI runs, and should not have any impact on personal runs.

Additionally I added a couple of items to the `.gitignore` file to keep them out.


larkost:
Looking at the output of the runs:
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-js/detail/PR-710/1/pipeline
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-js/detail/PR-710/2/pipeline
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-js/detail/PR-710/3/pipeline

I think together they combine to show that this PR looks good (or at least has not done more damage). @jpsim would you mind looking over this?

kristiandupont:
@jpsim ping

scripts/test.sh
```diff
@@ -6,9 +6,13 @@ set -e
 export TEST_SCRIPT=1
 export NPM_CONFIG_PROGRESS=false
 
-TARGET="$1"
-CONFIGURATION="${2:-"Release"}"
-DESTINATION=
+TARGET=$1
+CONFIGURATION=${2:-Release}
```
jpsim: why were the quotes removed here?

scripts/test.sh
```diff
@@ -6,9 +6,13 @@ set -e
...
+
+IOS_SIM_DEVICE=${IOS_SIM_DEVICE:-} # use preferentially, otherwise will be set and re-exported
+ios_sim_default_device_type=${IOS_SIM_DEVICE_TYPE:-iPhone 5s}
```
jpsim: won't this cause issues due to the space and lack of quoting?

scripts/test.sh
```diff
@@ -44,16 +48,33 @@ stop_server() {
...
+log_temp=
+test_temp_dir=
 cleanup() {
   # Kill started object server
-  stop_server
-
+  stop_server || true
+  
+  # Give the simulator a chance to go down gracefully
+  if $startedSimulator; then
```
jpsim: why is this preferable to the `while pgrep`/`pkill` loop that was previously in `reset-simulators.sh`?

scripts/test.sh
```diff
@@ -44,16 +48,33 @@ stop_server() {
...
+    osascript -e 'tell app "Simulator" to quit without saving' || true
+    sleep 0.25 # otherwise the pkill following will get it
```
jpsim: Shouldn't this instead run a `pgrep` loop as was previously used in `reset-simulators.sh` to wait until the process is truly gone? Otherwise, this 0.25s wait seems very fragile, especially when running on resource constrained VMs under pressure...

scripts/test.sh
```diff
@@ -44,16 +48,33 @@ stop_server() {
...
   # Kill all other child processes.
   pkill -P $$ || true
 
   # Kill react native packager
   pkill node || true
   rm -f "$PACKAGER_OUT" "$LOGCAT_OUT"
+  
+  # Cleanup temp files
+  if [ -n "$log_temp" ] && [ -e "$log_temp" ]; then
+    rm "$log_temp" || true
```
jpsim: IIRC `rm -f` always returns successfully, even if the file doesn't exist, so you could use that and even forgo the `-e` check above.

scripts/test.sh
```diff
@@ -44,16 +48,33 @@ stop_server() {
...
+  fi
+  if [ -n "$test_temp_dir" ] && [ -e "$test_temp_dir" ]; then
+    rm -rf "$test_temp_dir" || true
```
jpsim: the `-f` flag IIRC means that the `|| true` or the `-e` check above shouldn't be necessary

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
 }
 
 xctest() {
-  local dest="$(xcrun simctl list devices | grep -v unavailable | grep -m 1 -o '[0-9A-F\-]\{36\}')"
+  # note: the DEVELOPER_DIR and IOS_SIM_DEVICE sections sould be kept in sync with setup_ios_simulator
```
jpsim: What does "kept in sync" mean? They shouldn't be the same value, obviously. Does this refer to different versions of Xcode including different versions of available devices? If so, this could be reworded to say something along the lines of "This script may need to be updated to use available iOS Simulator devices if the Xcode version used changes.".

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
...
         exit $EXITCODE
-	}
+	  }
   fi
 }
 
+setup_ios_simulator() {
+  # - Ensure one version of xcode is chosen by all tools
+  
+  if [[ -z "$DEVELOPER_DIR" ]]; then
```
jpsim: this is taken care of earlier in the script

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
...
+  launchctl remove com.apple.CoreSimulator.CoreSimulatorService 2>/dev/null || true
+  sleep 0.25 # launchtl can take a small moment to kill services
+  
+  # kill them with fire
+  while pgrep -qx Simulator com.apple.CoreSimulator.CoreSimulatorService; do
+    pkill -9 -x Simulator com.apple.CoreSimulator.CoreSimulatorService || true
+    sleep 0.05
+  done
+  
+  # - Prod `simctl` a few times to make sure it is ready to go
```
jpsim: why? what happens when it's not ready to go? is there a better metric than running this continuously for 5 seconds to indicate when it's "ready"?

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
...
+  local deadline=$((SECONDS+5))
+  while [ -z "`xcrun simctl list devices 2>/dev/null`" ] && [ $SECONDS -lt $deadline ]; do
+    : # nothing to see here
+  done
+  
+  # - Choose a device, if it has not already been chosen
```
jpsim: this is copied verbatim from another part of the script. consider extracting into a function?

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
...
 }
 
+setup_ios_simulator() {
+  # - Ensure one version of xcode is chosen by all tools
+  
+  if [[ -z "$DEVELOPER_DIR" ]]; then
+    export DEVELOPER_DIR="$(xcode-select -p)"
+  fi
+  
+  # - Kill the Simulator to ensure we are running the correct one
```
jpsim: given the two ways this script kills the simulator, it might be worth adding a comment explaining why both different approaches are needed, when to use one over the other, etc. And perhaps refactoring into a `kill_simulators(with_fire: true/false)` method.

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
...
+    export IOS_SIM_DEVICE
+  done
+  if [ -z "$IOS_SIM_DEVICE" ]; then
+    echo "*** Failed to determine the iOS Simulator device to use ***"
+    exit 1
+  fi
+  
+  # - Reset the device we will be using
+  
+  xcrun simctl shutdown "$IOS_SIM_DEVICE" 1>/dev/null 2>/dev/null || true # sometimes simctl gets confused
```
jpsim: `>/dev/null 2>&1` rather than `1>/dev/null 2>/dev/null`?

scripts/test.sh
```diff
@@ -81,26 +102,118 @@ start_packager() {
...
+  xcrun simctl erase "$IOS_SIM_DEVICE"
+  
+  # - Start the target in Simulator.app
+  # Note: as of Xcode 7.3.1 `simctl` can not completely boot a simulator, specifically it can not bring up backboard, so GUI apps can not run.
+  #       This is fixed in version 8 of Xcode, but we still need the compatibility
```
jpsim: Do we still need the compatibility? Does Realm React Native support Xcode 7?

scripts/test.sh
larkost: For setting variables there is no difference, so the quotes were superfluous.

jpsim: ok, didn't know that. 👌

scripts/test.sh
larkost: No, not for setting variables. And for the default (`:-`) the end curly brace is the implicit end-quote. E.g.:
```bash
bob=${IOS_SIM_DEVICE_TYPE:-iPhone 5s}
echo "a - $bob - b"
```
Will get you `a - iPhone 5s - b`

jpsim: 👍

scripts/test.sh
larkost: Using `osascript` is just to give it a chance to go down nicely, reducing the amount of error clutter in stderr/stdout (before this was always giving nasty-looking messages). So the .25 secs is more of a grace period before the following `pkill -P $$` (it is now a child of ours). Additionally, since we are in `cleanup` I would rather not wait infinitely for something that is not going to go down from a `SIGHUP`. That sort of behavior is more appropriate in startup than shutdown (i.e.: line 164).

scripts/test.sh
larkost: True if the file is missing, but not if there is an I/O or permission error. Since we are in cleanup I would rather not fail things for those reasons. You are correct and the `-e` check is superfluous, and more a habit from other languages (Python in this case).

scripts/test.sh
larkost: Another case of me handling unexpected problems (like IO errors).

scripts/test.sh
jpsim: Sounds good! Thanks for explaining.

scripts/test.sh
larkost: Because the Jenkins and non-Jenkins (read: developer workstation) routes are different, they both need to have the DEVELOPER_DIR and IOS_SIM_DEVICE variables setting code. That should stay in sync to keep the behaviors the same.

scripts/test.sh
larkost: As indicated above, not in the non-Jekins path through the script.

scripts/test.sh
jpsim: ahhh, so it's not that `DEVELOPER_DIR` needs to be synced with `IOS_SIM_DEVICE`, but rather both places where they're set (Jenkins vs non-Jenkins), if you change one you should change the other? In any case, another reason why an explanatory comment is useful here.

scripts/test.sh
larkost: This is where the first time you run `simctl` after a different version is run, sometimes the `CoreSimulatorService` setup was not quite right and it fails to startup the first time.

And note that this tries it until it succeeds (i.e.: returns stdout), which in most cases is just once. So we are rarely retying for anywhere near the 5 second deadline.

jpsim: explanatory comment would be nice

scripts/test.sh
larkost: Why? This is just more explicit about what is happening. I am not aware of any performance differences, are you?

scripts/test.sh
larkost: They are not using it, but I want to share as much of this logic between projects.

scripts/test.sh
jpsim: it's just strange to see different syntaxes doing the same thing, for no apparent reason, which leads me (and likely other readers) to wonder why this is different, and if that difference is significant.

scripts/test.sh
jpsim: given that this is weaved into React Native's custom test runner script, as opposed to being in a self-contained file like it was before, don't you think this hurts our ability to reuse this?

kristiandupont:
What's the status of this, @jpsim and @larkost?

scripts/test.sh
larkost: Note that this goes with the group below.

jpsim:
Still waiting for some of my feedback to be addressed.

scripts/test.sh
larkost: Trying to put that much context in the comments feels a bit odd given that one is in a method called `cleanup`, and the other in `setup_ios_simulator`, with a section comment explaining why we are quitting Simulator.app.

And there are meaningful differences in the way we kill these based on those circumstances (clearing the way before us vs. cleaning up our own mess) that would make trying to combine this awkward.

scripts/test.sh
larkost: It is not going to be straight re-use, but rather copy-paste.

larkost:
Here is a new version. I merged some logic as you wanted, and that would up removing the need for some branching in places.

jpsim:
Which CI job output should I be looking at here? For example, all of the `react-tests` and the `react-example` jobs are failing, and I would have expected those to be the ones we're trying to make more reliable with the changes in this PR...

larkost:
If you look at the results of those runs, all of the problems are with missing symbols, namely `_RCTSetLogFunction`. That was an existing problem when I branched (I don't immediately know if that is fixed elsewhere yet).

jpsim:
If this helps make CI jobs with the Simulator more reliable, don't let me stop you from merging this in.

But I'm still not convinced that we wouldn't be better off building a self-contained wrapper around `xcodebuild` that works around iOS Simulator issues in a more reusable package, so that we could consolidate our efforts across projects and reuse that wrapper, rather than having the fixes interspersed into the test runner that's custom and unique to this project.

larkost:
I am not against making it a wrapper around `xcodebuild`, but that is not where we were coming from. The existing method (both here and in `realm-cocoa` has an external process that is called, and so cannot do things like intelligently choose the simulator we should be running in.

It is unclear from your comment (and marking in the Reviewers section) if you are approving this.

PR #712, opened 2016-12-01
use stash/unstash rather than multiple checkouts

larkost:
Trying to solve realm/ci#17


larkost:
@emanuelez I think that this one is read to go. There are three failures, but only the [`linux_test_runners`](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-js/detail/PR-712/25/pipeline/46/) one looks like it could be associated with this. So I think we should merge this one in.

larkost:
Checks out the repo on one machine, then uses stash/unstash to get that checkout onto all the other machines. Additionally adds a timeout to the android build to prevent those from hanging machines indefinitely and does some general cleanup on the Jenkinsfile.

PR #718, opened 2016-12-05
Add tests for closed-realm access protection

kristiandupont:


alazier:
We need to test all methods/properties available on realm including transactions.

src/js_realm.hpp
```diff
@@ -639,6 +639,9 @@ void RealmClass<T>::add_listener(ContextType ctx, ObjectType this_object, size_t
     auto callback = Value::validated_to_function(ctx, arguments[1]);
 
     SharedRealm realm = *get_internal<T, RealmClass<T>>(this_object);
+    if (realm->is_closed()) {
```
alazier: should we check for remove listener and remove all as well?

kristiandupont: Yeah, let me add that.

kristiandupont:
Yes, we should probably merge the https://github.com/realm/realm-object-store/pull/278 PR into our JS branch

PR #724, opened 2016-12-07
doc additions

alazier:
@kristiandupont @bigfish24


bigfish24:
The parameter in the callback should be named change_event to match the naming right?

PR #728, opened 2016-12-09
Allow use of a custom sync dir

### LINKED ISSUE #720: Set the root dir used for the object server including the metadata dir
According to https://github.com/realm/realm-object-store/issues/263#issuecomment-263949456
We need support for setting the root dir used by OS
We currently have a workaround for the ROS to have the server running without being able to set the path.

@alazier

alazier:
This closes #720

@kristiandupont


radu-tutueanu:
Merging, as we need it for a new version for https://github.com/realm/realm-sync-services/pull/621

PR #737, opened 2016-12-13
run notification tests in process

alazier:


blagoev:
@alazier Hey Ari, Do we need this still ?

PR #743, opened 2016-12-16
Nh/fixes 740

### LINKED ISSUE #740: Android tests on CI tend to hang forever
A timeout would be nice, but we need to figure out the realm issue

nhachicha:
Trying to fix #740 
- Relying on the React Catalyst to be active before running the tests
- Fix Android Test App package
- Adding release config for Android Test App


nhachicha:
retest this please

scripts/test.sh
```diff
@@ -169,14 +169,16 @@ case "$TARGET" in
   pushd tests/react-test-app
 
   npm install
-  open_chrome
-  start_packager
```
alazier: I think we still want to run the packager internally? Otherwise I think it pops open on a new terminal on ci and we lose control over it.

PR #744, opened 2016-12-16
Fix test names timeout

### LINKED ISSUE #711: Timeouts waiting for realm-test-names
In testing `macos_react_tests` both `debug` and `release` have been sporadically failing with the message:
```*** Terminating app due to uncaught exception 'ConditionTimeout', reason: 'Timed out waiting for: realm-test-names'```

This appears to be a legitimate (i.e. not CI-related) problem, probably in the test. An example of it can be found in the runs of either of those in this CI run:

https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-js/detail/master/6/pipeline/59

kristiandupont:
Fixes #711.

(No progress so far, just debugging CI.)


alazier:
I think the problem is a race condition - ie the notification is sent before we start listening. There is a complilation warning saying we should be inheriting from some eventing interface, maybe that would help.

PR #757, opened 2016-12-21
Disable dead code elimination

kristiandupont:
This fixes the linker error that we've started seeing, probably because of xcode versioning:
```
Undefined symbols for architecture i386:
1581 	  "_RCTSetLogFunction", referenced from:
1582 	      -[RealmReactTests invokeTest] in RealmReactTests.o
1583 	  "_RCTAddLogFunction", referenced from:
1584 	      +[RealmReactTests load] in RealmReactTests.o
```


mrackwitz:
Shouldn't we rather use `-force_load` when linking Realm on iOS instead of forcing our users to disable `DEAD_CODE_STRIPPING` for the whole target?

fealebenpae:
It seems to be the favored solution according to https://github.com/facebook/react-native/issues/2685. Besides, the symbols the linker complains about come from `libReact.a` and shouldn't be affected by how Realm is linked into the target.

PR #758, opened 2016-12-21
Introduce Function::make_callback

kristiandupont:


alazier:
Is there a reason we never finished this? @fealebenpae Would this have fixed the issues we have where we are required to call `setTimeout` to make come js callbacks to fire?

fealebenpae:
@alazier we already have https://github.com/realm/realm-js/blob/master/src/node/node_function.hpp#L40 - it's likely it's not used everywhere, like in the GlobalNotifier, though.

alazier:
@fealebenpae I can fix the adapter/gn. Thanks.

PR #770, opened 2017-01-02
Catch analytics errors

### LINKED ISSUE #769: Deactivate mixpanel
Hello, 

I just tried to run my nodejs project without internet and it crashed right away since realm can't make a request to api.mixpanel.com.

Is there a way to handle the error so it doesn't crash ? or remove analytic if there is no internet connection ?

```js
events.js:160
throw er; // Unhandled 'error' event
^
Error: getaddrinfo ENOTFOUND api.mixpanel.com api.mixpanel.com:443
at errnoException (dns.js:28:10)
at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)
```

kristiandupont:
This resolves #769

### DIFF
--- a/lib/submit-analytics.js
+++ b/lib/submit-analytics.js
@@ -89,7 +89,9 @@ module.exports = function(eventName) {
         }
     };
 
-    request(`https://api.mixpanel.com/track/?data=${Buffer.from(JSON.stringify(payload), 'utf8').toString('base64')}&ip=1`);
+    try {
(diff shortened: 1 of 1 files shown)


fealebenpae:
I think the proper fix is to add the second parameter to the `request` call, which is the callback that will receive the error.

kristiandupont:
retest this please

Issue #752, opened 2016-12-20
Add test harness for Electron


kristiandupont:

kristiandupont:
Possibly using [Spectron](http://electron.atom.io/spectron/)

kraenhansen:
Closed with https://github.com/realm/realm-js/pull/1196

PR #774, opened 2017-01-03
Fix constructor init

### LINKED ISSUE #766: Realm is not defined (React-native + Debug JS Remotely)
Hi!
When I turn on the Debug JS Remotely feature on my Android emulator, I get this error

err

When the Debug JS Remotely is Off, I get no errors

Android: API 17
React-native: 0.39.2
Realm: 0.15.2

kristiandupont:
This resolves #766

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,16 +27,14 @@ function nodeRequire(module) {
 function getContext() {
     if (typeof process === 'object' && process + '' === '[object process]') {
         return process.type === 'renderer' ? 'electron' : 'nodejs';
-    }        
+    }
 
-    if (typeof navigator !== 'undefined') {
-        if (navigator.product === 'ReactNative') { // eslint-disable-line no-undef
-            return 'reactnative';
-        }
+    if (typeof jest === 'object') {
+        return 'nodejs';
+    }
 
-        if (/Chrome/.test(navigator.userAgent)) { // eslint-disable-line no-undef
-            return 'chromedebugger';
-        }
+    if (typeof Realm !== 'undefined') {
+        return 'jscore';
     }
 
     if (typeof global !== 'undefined') {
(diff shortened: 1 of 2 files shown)


kristiandupont:
retest this please

tests/react-test-app/package.json
```diff
@@ -6,8 +6,9 @@
     "start": "react-native start"
   },
   "dependencies": {
-    "react": "15.3.2",
-    "react-native": "^0.37.0",
+    "invariant": "^2.2.2",
```
radu-tutueanu: Why add invariant? Was this missing?

kristiandupont: Yes. Though it's not a direct dependency, it seems to be the way people resolve it in https://github.com/facebook/react-native/issues/11327. Maybe there is a nicer solution, this is a bit hacky..

alazier: It sounds like this could be a dev dependency

alazier:
How do we test that this is actually fixed?

lib/index.js
```diff
@@ -27,16 +27,14 @@ function nodeRequire(module) {
...
-            return 'reactnative';
-        }
+    if (typeof jest === 'object') {
+        return 'nodejs';
+    }
 
-        if (/Chrome/.test(navigator.userAgent)) { // eslint-disable-line no-undef
-            return 'chromedebugger';
-        }
+    if (typeof Realm !== 'undefined') {
```
alazier: Does this reordering matter/make a difference? Might be useful to add some comments.

kristiandupont:
We've tried it manually with Electron and it doesn't break anything. You are right that we need tests (It's tracked with #752) but as it is, Electron is not officially supported and it doesn't work out of the box so I think it's safe to make this update.

tests/react-test-app/package.json
kristiandupont: Sure, I can make it a dev dependency. It doesn't make much of a difference unless someone wants to deploy the test app somewhere though.

lib/index.js
kristiandupont: It matters because during chrome debugging, `navigator.product` is now also set to `ReactNative` so if we just test for that first, `getContext()` will report `reactnative` also in chrome. I'll add a comment.

PR #775, opened 2017-01-03
Disable caching of realm instances

kristiandupont:
This might be the way to resolve the initialize without schema version issue among other things.


kristiandupont:
This is discarded as https://github.com/realm/realm-object-store/pull/314 will provide the proper solution.

PR #776, opened 2017-01-04
Update to sync v1.0.0-BETA-6.3

bdash:
This required updating to current object store master to accommodate the changes to sync error reporting.

/cc @radu-tutueanu


scripts/test.sh
```diff
@@ -69,7 +69,7 @@ cleanup() {
   pkill -9 -P $$ || true
 
   # Kill react native packager
-  pkill node || true
+  pkill -x node || true
```
bdash: This isn't related to the rest of the change, but it fixes warnings that are printed about trying and failing to kill root-owned `ionodecache` processes.

radu-tutueanu:
Thanks, @bdash and @kristiandupont  ! There are still some failures, I'll take a look a little later.

kristiandupont:
CI fails with the known issue. Tests succeed on my machine now, so I will merge this.

PR #778, opened 2017-01-04
Update to object store master to pick up the fix for the global notifier

bdash:


tests/js/realm-tests.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
         TestCase.assertEqual(defaultRealm.schemaVersion, 0);
 
         TestCase.assertThrows(function() {
-            new Realm({schemaVersion: 1});
+            new Realm({schemaVersion: 1, schema: []});
```
bdash: This change is necessary because the schema version is ignored unless a schema is specified. I think that has always been the intent.

PR #780, opened 2017-01-05
Update ObjectStore

radu-tutueanu:
To add proper support for opening a Realm without a schema and then with a schema


radu-tutueanu:
Already done in #778

PR #785, opened 2017-01-09
Improve error message for typeerrors

### LINKED ISSUE #730: Ambiguous error - Error: 'Property' must be of type: string
Hi,

I have an issue with the error messages. I am getting the following error:

Error: 'Property' must be of type: string

I obviously have a problem with my data but I have no clue _which_ property is being provided the wrong data. Is there anyway the message could be updated to indicate where the problem lies? Or, if I'm just being dumb, point me to where I can see what property is having issues. 

Thanks,

Sam

kristiandupont:
This resolves #730

### DIFF
--- a/tests/js/realm-tests.js
+++ b/tests/js/realm-tests.js
@@ -866,4 +866,21 @@ module.exports = {
         realm = new Realm({path: 'dates-v5.realm', schema: [schemas.DateObject]});
         TestCase.assertEqual(realm.objects('Date')[0].currentDate.getTime(), 1);
     },
+
+    testErrorMessageFromInvalidWrite: function() {
+        var realm = new Realm({schema: [schemas.PersonObject]});
+        
+        TestCase.assertThrowsException(function() {
+            realm.write(function () {
+                var p1 = realm.create('PersonObject', { name: 'Ari', age: 10 });
+                p1.age = "Ten";
+            });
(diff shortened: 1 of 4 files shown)


kristiandupont:
This only resolves the issue partly. #787 covers the rest, which requires changes to object store. I think we could merge this one for the current release and make the full solution after that.

tests/js/realm-tests.js
```diff
@@ -866,4 +866,21 @@ module.exports = {
...
+        var realm = new Realm({schema: [schemas.PersonObject]});
+        
+        TestCase.assertThrowsException(function() {
+            realm.write(function () {
+                var p1 = realm.create('PersonObject', { name: 'Ari', age: 10 });
+                p1.age = "Ten";
+            });
+        }, new Error("PersonObject.age must be of type: number"));
+
+        //TestCase.assertThrowsException(function() {
```
alazier: do we still want this?

kristiandupont: You are right, I should remove it and reintroduce it when we solve #787.

kristiandupont:
@fealebenpae I agree that `TypeError` would be better but we aren't using specialized error classes anywhere and introducing them here seems out of scope for this PR.

jbrower95:
is this in 0.15?

kristiandupont:
@jbrower95 yes, 0.15.4.

PR #797, opened 2017-01-12
Update Object Store Submodule

radu-tutueanu:


radu-tutueanu:
Blocken on https://github.com/realm/realm-object-store/pull/341

PR #800, opened 2017-01-13
the react example made another RealmTasks implementation

gritzko:
Pending issues:
* magic IP/URI discovery
* React Native 0.40 has some antics: item list starts at a negative Y (maybe need a RN expert here, or just wait for a new RN version)


examples/ReactExample/components/login.js
```diff
@@ -0,0 +1,102 @@
...
+import Realm from 'realm';
+import styles from './styles';
+import TodoApp from './todo-app';
+
+
+export default class LoginScreen extends React.Component {
+
+    constructor (props) {
+        super(props);
+        this.state = {login: 'user', password: '...'};
```
kristiandupont: Is the "..." password a placeholder? We should probably use the `placeholder` attribute of the `TextInput` element instead.

examples/ReactExample/components/login.js
```diff
@@ -0,0 +1,102 @@
...
+                    password: value.password||'password'
+                });
+            }
+        } ).done();
+    }
+
+    submit () {
+        console.log('LOGIN')
+        Realm.Sync.User.login(
+            "http://192.168.157.64:9080/",
```
kristiandupont: I suppose we should extract this to somewhere. Where is it stored in the other examples?

examples/ReactExample/components/login.js
```diff
@@ -0,0 +1,102 @@
...
+            this.state.password,
+            (error, user) => {
+                if (error) {
+                    this.state.error = error;
+                    this.forceUpdate();
+                } else {
+                    delete this.state.error;
+                    AsyncStorage.setItem('creds', JSON.stringify(this.state));
+
+                    console.log('BINGO');
```
kristiandupont: We should probably remove these before merging

examples/ReactExample/components/realm.js
```diff
@@ -20,23 +20,45 @@
...
+    name: 'TaskListList',
+    properties: {
+        id: 'int',
+        items: {type: 'list', objectType: 'TaskList'},
     },
+    primaryKey: 'id'
 };
 
-export default new Realm({schema: [Todo, TodoList]});
+var adminToken = "ewoJImlkZW50aXR5IjogImFkbWluIiwKCSJhY2Nlc3MiOiBbInVwbG9hZCIsICJkb3dubG9hZCIsICJtYW5hZ2UiXQp9Cg==:I1mEfddNdQo4wS/7dVgEIEm9pv7pLEyyx3zmGgWMCxkEHOxz8DxLjm/yLD8OD7iE9XcCHy4xs6SOu2ZbbRysezssyY+3z/anrz7u/EBDvyfHPRakaw3rTpD6RcTb4SaAeLBagtDP7YqycneZrGd3oUXofvJdWg4YpDOvr/+1zwntt3DGg5D1ghKzSWcF7FnosswjiqGk91t2hbZbPYNEPEJKwkCPj0exOZqQ73627rD9Jzbr3CcHpp0V4U1BEhVfgZyo9dchWYT9Qw0rkqjqc5ylvR73Ihpa9hN9+wKxhU1K6nNzsbhrD0sMfOOlxsuduWHrhiT8vkTL/WFOO4vC1Q==";
```
kristiandupont: How is this stored in the other examples?

examples/ReactExample/components/login.js
gritzko: There is some magic; they take ip from ifconfig and substitute it there.
That potentially given some weird errors if your WiFi address changes.
There must be a better solution.

examples/ReactExample/components/login.js
gritzko: Thanks.

examples/ReactExample/components/realm.js
gritzko: Oops, not needed anymore.

examples/ReactExample/components/login.js
gritzko: I think, I'll put all such things into a separate config file.

PR #801, opened 2017-01-17
Add missing chrome debugging apis for Sync.User

### LINKED ISSUE #791: Add missing API's for chrome debugging
A number of API's, especially related to sync, are missing from the chrome debugging proxies.

kristiandupont:
This resolves #791.


src/rpc.cpp
```diff
@@ -133,6 +134,27 @@ RPCServer::RPCServer() {
         RPCObjectID realm_id = store_object(realm_object);
         return (json){{"result", realm_id}};
     };
+    m_requests["/create_user"] = [this](const json dict) {
```
alazier: Why not just forward these requests to the Sync object as we do with all other methods? Same question for the rest of new sync methods.

kristiandupont: Maybe I don't understand the implementation well enough, but if we create the user simply with a `/call_method` request, the object wont be stored in `m_objects`, so we would not be able to access it afterwards? Or are you suggesting that I add that ability to `call_method`?

alazier: Objects are stored when they are serialized before they are returned. Not sure why this wouldn't just work - although I haven't looked at this code for quite a while.

alazier: Looks like you are already calling `store_object` in `serialize_json_value`, so the object should be stored.

kristiandupont: Yes, I think you are right. I am trying out just sending the `Sync.User` object in `create_session` and use it like the other objects.

tests/js/index.js
```diff
@@ -28,8 +28,8 @@ var TESTS = {
     QueryTests: require('./query-tests'),
     EncryptionTests: require('./encryption-tests'),
     MigrationTests: require('./migration-tests'),
-    UserTests: require('./user-tests'),
-    AsyncTests: require('./async-tests')
+    UserTests: require('./user-tests')
+    /*AsyncTests: require('./async-tests')*/
```
alazier: why are we disabling these? isn't this the only place we actually test collection notifications?

PR #809, opened 2017-01-20
Always invoke sync callbacks on the JavaScript thread

### LINKED ISSUE #804: Marshal sync callbacks onto the JS thread
Without this we get assertion failures in React Native on iOS and segmentation faults in JavaScriptCore on Android because the JavaScript state is not meant to be accessed on other threads.

fealebenpae:
Fixes #804

### DIFF
--- a/src/event_loop_dispatcher.hpp
+++ b/src/event_loop_dispatcher.hpp
@@ -71,6 +71,8 @@ private:
     };
     const std::shared_ptr<EventLoopSignal<Callback>> m_signal;
     
+    const std::thread::id m_thread = std::this_thread::get_id();
+    
 public:
     EventLoopDispatcher(std::function<void(Args...)> func)
     : m_state(std::make_shared<State>(func))
@@ -81,6 +83,11 @@ public:
     
     void operator()(Args... args)
     {
+        if (m_thread == std::this_thread::get_id()) {
+            m_state->m_func(args...);
+            return;
+        }
+        
         {
             std::unique_lock<std::mutex> lock(m_state->m_mutex);
             m_state->m_invocations.push(std::make_tuple(args...));
--- a/src/event_loop_dispatcher.hpp
+++ b/src/event_loop_dispatcher.hpp
@@ -24,7 +24,8 @@
 
 #include "util/event_loop_signal.hpp"
 
-namespace __apply_polyfill {
+// FIXME: remove once we switch to C++ 17 where we can use std::apply
+namespace _apply_polyfill {
 template <class Tuple, class F, size_t... Is>
 constexpr auto apply_impl(Tuple t, F f, std::index_sequence<Is...>) {
     return f(std::get<Is>(t)...);
@@ -64,7 +65,7 @@ private:
                 std::unique_lock<std::mutex> lock(state->m_mutex);
                 while (!state->m_invocations.empty()) {
                     auto& tuple = state->m_invocations.front();
-                    ::__apply_polyfill::apply(tuple, state->m_func);
+                    ::_apply_polyfill::apply(tuple, state->m_func);
                     state->m_invocations.pop();
                 }
             }
--- a/src/event_loop_dispatcher.hpp
+++ b/src/event_loop_dispatcher.hpp
@@ -56,18 +56,16 @@ private:
     const std::shared_ptr<State> m_state;
     
     struct Callback {
-        std::weak_ptr<State> m_state;
+        std::shared_ptr<State> m_state;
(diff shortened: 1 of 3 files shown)


src/event_loop_dispatcher.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+#pragma once
+
+#include <queue>
+#include <tuple>
+#include <mutex>
+
+#include "util/event_loop_signal.hpp"
+
+namespace __apply_polyfill {
```
simonask: Identifiers beginning with double underscores are reserved by the language (used by compilers for things like `__builtin_foo`).

src/event_loop_dispatcher.hpp
```diff
@@ -0,0 +1,92 @@
...
+template <class Tuple, class F, size_t... Is>
+constexpr auto apply_impl(Tuple t, F f, std::index_sequence<Is...>) {
+    return f(std::get<Is>(t)...);
+}
+
+template <class Tuple, class F>
+constexpr auto apply(Tuple t, F f) {
```
simonask: Add FIXME about removing this when `std::apply` is available. :)

src/event_loop_dispatcher.hpp
```diff
@@ -0,0 +1,92 @@
...
+private:
+    struct State {
+    public:
+        State(std::function<void(Args...)> func) : m_func(func) { }
+        
+        const std::function<void(Args...)> m_func;
+        std::queue<Tuple> m_invocations;
+        std::mutex m_mutex;
+    };
+    const std::shared_ptr<State> m_state;
```
simonask: Why `std::shared_ptr` and not a member variable?

src/event_loop_dispatcher.hpp
```diff
@@ -0,0 +1,92 @@
...
+                std::unique_lock<std::mutex> lock(state->m_mutex);
+                while (!state->m_invocations.empty()) {
+                    auto& tuple = state->m_invocations.front();
+                    ::__apply_polyfill::apply(tuple, state->m_func);
+                    state->m_invocations.pop();
+                }
+            }
+        }
+    };
+    const std::shared_ptr<EventLoopSignal<Callback>> m_signal;
```
simonask: Same, why `std::shared_ptr`?

src/event_loop_dispatcher.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+template <typename... Args>
+class EventLoopDispatcher<void(Args...)> {
+    using Tuple = std::tuple<typename std::remove_reference<Args>::type...>;
+private:
+    struct State {
+    public:
+        State(std::function<void(Args...)> func) : m_func(func) { }
+        
+        const std::function<void(Args...)> m_func;
```
simonask: Wrapping things in an `std::function` seems unnecessary since you don't really use the `Args...` template pack for anything other than passing along to `std::function`. I suppose you could just replace `std::function<void(Args...)>` with a template paramter `F` as in the declaration.

fealebenpae: My template parameter is exactly `void(const std::string&, const SyncConfig&, std::shared_ptr<SyncSession>)`, not the inferred type of a lambda or other callable. I don't think I'd be able to use it as storage.

src/event_loop_dispatcher.hpp
fealebenpae: The `EventLoopDispatcher` instance can be copied a number of times when stuffed in a `std::function` but I want it to use the same state.

src/event_loop_dispatcher.hpp
fealebenpae: `EventLoopSignal` is neither copyable, nor movable. What's more, the Android implementation inherits from `std::enable_shared_from_this` and relies on being stored in a `shared_ptr`.

src/event_loop_dispatcher.hpp
simonask: Oh okay, so the intention is that a user of the class will define the signature up front?

src/event_loop_dispatcher.hpp
simonask: :+1:

src/event_loop_dispatcher.hpp
fealebenpae: Yup, like in https://github.com/realm/realm-js/blob/8c61cc9d41aa302de700e58edb06786972b99aec/src/js_sync.hpp#L268.

PR #828, opened 2017-02-01
Add Session and some missing User docs

kristiandupont:


docs/sync.js
```diff
@@ -181,10 +182,78 @@ class User {
...
+	openManagementRealm() {}
+}
+
+/**
+ * An object encapsulating a Realm Object Server session. Sessions represent the communication between the 
+ * client (and a local Realm file on disk), and the server (and a remote Realm at a given URL stored on a Realm Object Server).
+ * Sessions are always created by the SDK and vended out through various APIs. The lifespans of sessions 
+ * associated with Realms are managed automatically.
+ */
+class Session {
```
fealebenpae: I think this is missing `@memberof Realm.Sync` like User.

docs/permission.js
```diff
@@ -0,0 +1,232 @@
+import {typeForConstructor} from '../lib/browser/objects';
```
fealebenpae: Does this belong here?

PR #849, opened 2017-02-08
Replace `npm run get-version` with JSON parser gradle function

### LINKED ISSUE #799: error when install Realm on window
Hi all,

I have try to install Realm on window for new project react-native, but it fall

step I did: 

create new react-native project
 `react-native init TestRealm`
`cd TestRealm`

And then install Realm

`npm install --save realm`

And link

`react-native link realm`

Then run 

`react-native run-android`

but it have error

```
C:\Users\keyyu\Atom\Workspace\TestRealm>react-native run-android
Starting JS server...
'adb' is not recognized as an internal or external command,
operable program or batch file.
Building and installing the app on the device (cd android && gradlew.bat installDebug)...

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\keyyu\Atom\Workspace\TestRealm\node_modules\realm\android\build.gradle' line: 71

* What went wrong:
A problem occurred evaluating project ':realm'.
> Cannot run program "npm" (in directory "C:\Users\keyyu\Atom\Workspace\TestRealm\node_modules\realm\android"): CreateProcess error=2, The system cannot find the file specified

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 22.896 secs
Could not install the app on the device, read the error above for details.
Make sure you have an Android emulator running or a device connected and have
set up your Android development environment:
https://facebook.github.io/react-native/docs/android-setup.html

(node:10272) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: Cannot read property 'message' of undefined
```

I did step by step like [https://realm.io/docs/react-native/latest/](https://realm.io/docs/react-native/latest/).

my enviroment

"node": v6.9.2
"npm":  v3.10.9
"react": "15.4.2",
"react-native": "0.40.0",
"realm": "^0.15.4"

Please help. 

Thank you!

dbhowell:
**Goal:** This PR is to resolve #799 
**Expected Results**: When building a react-native app that links `realm` for Android on a Microsoft Windows platform, the gradle task of SendAnalytics should send the realm package version to realm for analytic purposes.
**Actual Result**: The command `npm run get-version --silent` appears to fail as the `node` executable isn't in the gradles PATH. (https://github.com/realm/realm-js/blob/master/react-native/android/publish_android_template#L71)

This PR resolves the issue by parsing the `package.json` directly with a JSON parser instead of using the `npm run get-version` script, bypassing the need for the npm script.

It uses a gradle function created by Andrew Jack (@AndrewJack, see https://github.com/AndrewJack/versioning-react-native-app).

This one is difficult to test, however should work across all supported platforms (Microsoft Windows, Linux and MacOS).

(CLA has been signed)

### DIFF
--- a/react-native/android/publish_android_template
+++ b/react-native/android/publish_android_template
@@ -66,9 +66,17 @@ String getAppId () {
 
 apply from: 'analytics.gradle'
 
+import groovy.json.JsonSlurper
+
+def getNpmVersion() {
+    def inputFile = new File("../package.json")
+    def packageJson = new JsonSlurper().parseText(inputFile.text)
+    return packageJson["version"]
+}
+
 task send(type: SendAnalyticsTask) {
     applicationId = getAppId()
-    version = "npm --silent run get-version".execute(null, projectDir).text.trim()
+    version = getNpmVersion().trim()
 }
 
 preBuild.dependsOn send
--- a/react-native/android/publish_android_template
+++ b/react-native/android/publish_android_template
@@ -69,7 +69,7 @@ apply from: 'analytics.gradle'
 import groovy.json.JsonSlurper
 
 def getNpmVersion() {
-    def inputFile = new File("../package.json")
+    def inputFile = new File(buildscript.sourceFile.getParent() + "/../package.json")
     def packageJson = new JsonSlurper().parseText(inputFile.text)
     return packageJson["version"]
 }


kristiandupont:
This is awesome, thank you for contributing @dbhowell!

iSimonWeb:
Can we merge this?

kristiandupont:
Hi @iSimonWeb sorry, Nabil who needs to review this is very busy at the moment. We will get to it asap.

nhachicha:
Hi @dbhowell 
thanks for the PR and the detailed explanation.

I like the idea of using `JsonSlurper` to parse the `package.json` to get the version, we should probably move towards this solution, however, the current implementation will report the _user's_ version not the `Realm` module version.

ex: If you use the `realm-js/examples/ReactExample`, you'll see the `send` task parsing `realm-js/examples/ReactExample/package.json` with `0.0.1` not the one for the Realm module (located in `realm-js/examples/ReactExample/node_modules/realm/package.json`  with `1.0.1`)
Hope this could help.

Cheers,

dbhowell:
Hi @nhachicha , I thought that might be the case. I'll change it to look for the `node_modules/realm/package.json`. Would that be acceptable?

The assumption is then: it is always located in the local `node_modules` folder (which may not always be the case)?

dbhowell:
@nhachicha ignore my last comment, I think I've discovered a way to get the current gradle source file and can use that as the base path (`buildscript.sourceFile`)

dbhowell:
@nhachicha, please see my last commit to address https://github.com/realm/realm-js/pull/849#issuecomment-279795929

Issue #736, opened 2016-12-13
VS code debugging crashes with react native and realm-js, also debug crashes on nuclide


deodeveloper:
Actual Behavior

debugging crashes on react native 0.39, vscode 1.7.2, realm 0.15.0
Similar crashing happens on nuclide too.
Expected Behavior

Could able to debug the app with Realm-JS
Software versions

React-Native VS Code extension version: 1.7.2
VSCode version: 1.7.2
OS platform and version: mac os 10.12.1
NodeJS version: 7.2.1
React Native version: 0.39
Outputs (Include if relevant)

[vscode-react-native] Starting debugger app worker.
[vscode-react-native] Established a connection with the Proxy (Packager) to the React Native application
[vscode-react-native] Debugging session started successfully.
Error: socket hang up
at doRequest (/Users/admin/Documents/Development/react-native/lang-app/languageApp/node_modules/sync-request/index.js:37:11)
at makeRequest (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:73531:14)
at sendRequest (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:73561:14)
at Object.createSession (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:73452:1020)
at /Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:73080:31
at loadModuleImplementation (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:188:1)
at guardedLoadModule (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:140:8)
at _require (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:124:1)
at /Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:72899:18
at loadModuleImplementation (/Users/admin/Documents/Development/react-native/lang-app/languageApp/.vscode/.react/index.ios.bundle:188:1)
Realm failed to connect to the embedded debug server inside the app. If attempting to use Chrome debugging from a device, ensure the device is reachable on the same network as this machine.

deodeveloper:
[...]

PR #854, opened 2017-02-09
Fix VScode debugging for React Native

mjbader:
With our React Native project I looked into why Realm refused to work in the VScode debugging set up using react native tools.  Despite being on the latest version I kept running into #736.  After doing some proper digging I found out that the rpc call was timing out when the handler attempted to get the text from the request because the content type was "application/json" and not "text/plain" like the chrome debugger.  This caused GCDWebserver to call GWS_DNOT_REACHED which caused the server to abort and a response to not be sent.  After setting the content type to text/plain, I ran into some issues with the constructor and removing the .default for the browser import resolved the missing constructor issue.


gritzko:
see https://github.com/realm/realm-js/pull/855

PR #867, opened 2017-02-15
Nh/fixes 866

### LINKED ISSUE #866: Analytics task `send` could be triggered without a build
Current Gradle task `send` which sends analytics, should be triggered usually when the user perform a build, example: `react-native run-android`  or `./gradlew assembleDebug`

Currently the task is defined to execute during the Gradle configuration phase, this will be for example triggered when the user invoke `./gradlew tasks`

nhachicha:
- Fixes #866 
- Update Changelog, for the next release
- Add _Credits_ section  to the Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+1.0.3 Release notes
+=============================================================
+### Breaking changes
+* None
+
+### Enhancements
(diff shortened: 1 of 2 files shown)


nhachicha:
cc @dbhowell

PR #892, opened 2017-03-01
Make Realm creation available extenally, update OS and Sync versions

alazier:
@kristiandupont @fealebenpae @gritzko @radu-tutueanu


radu-tutueanu:
There are the breaking changes in realm-core 2.3.2 : https://github.com/realm/realm-core/commit/4322f25731a9eb44ae133b0a7a57bb547f83d377. ObjectStore unit tests do not compile `src/object_accessor.hpp` as far as I can tell.
You should update the core version in `dependencies.list` as well, used when building node without sync.

radu-tutueanu:
Looks fine to me otherwise. Where would we use this function?

alazier:
@radu-tutueanu this is used to create Realms in the correct execution context from within the global notifier/adapter.

alazier:
Without this I don't think registering notifications on realms created from the notifier/adapter would work. Does this sound correct @fealebenpae ?

radu-tutueanu:
👍 
@fealebenpae is on vacation until next week.

PR #900, opened 2017-03-09
Remove need for bash script on android builds

dbhowell:
**Goal**: Remove the need to use a bash script (`find`, `sed` & `mv`) to rename `*-arm-*` to `*-armeabi-*` (progressing to a platform independence when building realm from source)

Currently you are using a bash script at the point untarring the realm core. It's purpose is to rename any file containing `-arm-` and replace it with `-armeabi-`, otherwise the build fails.

We can utilise gradle's rename instead.

There may be a reason you've used the bash script, but I can't spot it.


kristiandupont:
Great, thank you for this, @dbhowell! We'll take a look at it.

PR #915, opened 2017-03-15
Add optional userInfo param to registerWithProvider

### LINKED ISSUE #909: Add `userInfo` map to `User.registerWithProvider`
Like the other SDK's (like [.Net](https://realm.io/docs/xamarin/1.1.0/api/reference/Realms.Sync.Credentials.html#Realms_Sync_Credentials_Custom_System_String_System_String_System_Collections_Generic_IDictionary_System_String_System_Object__))

kristiandupont:
This resolves #909

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -1,3 +1,4 @@
+import {stringify} from 'ini';
 ////////////////////////////////////////////////////////////////////////////
 //
 // Copyright 2016 Realm Inc.
@@ -139,15 +140,17 @@ class User {
 	static login(server, username, password, callback) {}
 
 	/**
-	 * Login a sync user using an external login provider.
+	 * Register/login a sync user using an external login provider.
 	 * @param {string} server - authentication server
-	 * @param {string} provider - The provider type
-	 * @param {string} providerToken - The access token for the given provider
+     * @param {object} options - options, containing the following:
+	 * @param {string} options.provider - The provider type
+	 * @param {string} options.providerToken - The access token for the given provider
+     * @param {object} [options.userInfo] - A map containing additional data required by the provider
 	 * @param {function(error, User)} callback - called with the following arguments:
 	 *   - `error` - an Error object is provided on failure
 	 *   - `user` - a valid User object on success
 	 */
-	static registerWithProvider(server, provider, providerToken, callback) {}
+	static registerWithProvider(server, options, callback) {}
 
 	/**
 	 * Register a sync user with username and password.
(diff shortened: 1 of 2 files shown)


docs/sync.js
```diff
@@ -139,15 +140,17 @@ class User {
 	static login(server, username, password, callback) {}
 
 	/**
-	 * Login a sync user using an external login provider.
+	 * Register/login a sync user using an external login provider.
 	 * @param {string} server - authentication server
-	 * @param {string} provider - The provider type
-	 * @param {string} providerToken - The access token for the given provider
+     * @param {object} options - options, containing the following:
```
fealebenpae: Indentation seems to be off.

lib/user-methods.js
```diff
@@ -157,11 +157,28 @@ module.exports = {
...
+                };
+                callback = arguments[3];    
+            }
+
+            let reqOptions = {
+                provider: options.provider,
+                data: options.providerToken,
+            };
+
+            if (options.userData) {
```
fealebenpae: This should be `options.userInfo`

docs/sync.js
```diff
@@ -139,15 +140,17 @@ class User {
...
+	 * @param {string} options.provider - The provider type
+	 * @param {string} options.providerToken - The access token for the given provider
+     * @param {object} [options.userInfo] - A map containing additional data required by the provider
```
alazier: Why do we need a separate internal map for userInfo? Could we just make the whole options map userInfo? Not sure what the motivation is for having nested maps.

teotwaki: Probably to handle both the old API exposed by the JS client, and the new one introduced by this change. Also, because the userInfo is completely optional, and unused by most authentication delegates.

kristiandupont: Also, this is aligned with the other SDK's.

docs/sync.js
```diff
@@ -139,15 +140,17 @@ class User {
 	static login(server, username, password, callback) {}
 
 	/**
-	 * Login a sync user using an external login provider.
+	 * Register/login a sync user using an external login provider.
 	 * @param {string} server - authentication server
-	 * @param {string} provider - The provider type
-	 * @param {string} providerToken - The access token for the given provider
+   * @param {object} options - options, containing the following:
```
bmunkholm: is indentation a bit off here?

kristiandupont: It is! Even after I fixed it. It looks like the file actually has tabs rather than spaces. I'll fix the whole thing.

PR #917, opened 2017-03-15
fix test-runners test on windows

blagoev:


blagoev:
closing this since we have different plans for running tests on windows

PR #918, opened 2017-03-15
Use sync 1.4.0

radu-tutueanu:


radu-tutueanu:
@fealebenpae All the react tests fail. Is that a known issue?

src/js_realm.hpp
```diff
@@ -44,6 +44,20 @@
 namespace realm {
 namespace js {
 
+static std::string normalize_realm_path(std::string path) {
```
kristiandupont: Shouldn't this live in object store now, if we are sharing it?

radu-tutueanu: I think you are right, it does make sense to move it to ObjectStore

kristiandupont: But for this release we can go with this I suppose.

PR #921, opened 2017-03-16
make tests Debug arg case insensitive

blagoev:


scripts/test.sh
```diff
@@ -9,6 +9,10 @@ export NPM_CONFIG_PROGRESS=false
 TARGET=$1
 CONFIGURATION=${2:-Release}
 
+if echo $CONFIGURATION | grep -i "^Debug$" > /dev/null ; then
```
fealebenpae: Why not use `if [[ $CONFIGURATION =~ "^[Dd]ebug$" ]]; then` instead?

blagoev: the grep version catches debUG as valid aswell. also debuG and so on

fealebenpae: Right. Makes sense.

PR #924, opened 2017-03-17
Validate user-methods parameters

### LINKED ISSUE #890: Add argument checks for methods in user-methods.js
As is nothing happens which is confusing. We should probably throw an error for an invalid number of arguments. I saw this behavior in node.

kristiandupont:
Resolves #890

### DIFF
--- a/lib/user-methods.js
+++ b/lib/user-methods.js
@@ -24,14 +24,10 @@ function node_require(module) {
     return require(module);
 }
 
-function typeOf(obj) {
-    return ({}).toString.call(obj).match(/\s(\w+)/)[1].toLowerCase();
-}
-
 function checkTypes(args, types) {
     args = [].slice.call(args);
     for (var i = 0; i < types.length; ++i) {
-        if (typeOf(args[i]) !== types[i]) {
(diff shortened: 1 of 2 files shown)


lib/user-methods.js
```diff
@@ -24,6 +24,19 @@ function node_require(module) {
     return require(module);
 }
 
+function typeOf(obj) {
+    return ({}).toString.call(obj).match(/\s(\w+)/)[1].toLowerCase();
```
fealebenpae: How about `Object.prototype.toString`? Or using the `typeof` operator?

kristiandupont: I don't know what I was thinking here -- something about types not being available in the arguments array, which doesn't make any sense..

lib/user-methods.js
```diff
@@ -24,6 +24,15 @@ function node_require(module) {
     return require(module);
 }
 
+function checkTypes(args, types) {
+    args = [].slice.call(args);
```
fealebenpae: This should be `Array.prototype.slice`.

kristiandupont: I guess you are right :)

PR #925, opened 2017-03-17
Add missing sync error

### LINKED ISSUE #923: Descriptive error message trying to access Sync on Linux
As Sync is not available in the free edition on node on linux, we should produce an error rather than just a missing sync object.

kristiandupont:
This resolves #923

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -69,7 +69,7 @@ module.exports = function(realmConstructor) {
     } else {
         Object.defineProperty(realmConstructor, 'Sync', {
             get: function () {
-                throw new Error("Sync is not enabled. Note that the developer edition of the Node.JS SDK for Realm does not include sync on Linux.");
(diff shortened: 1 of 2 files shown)


lib/extensions.js
```diff
@@ -66,6 +66,12 @@ module.exports = function(realmConstructor) {
 
         setConstructorOnPrototype(realmConstructor.Sync.User);
         setConstructorOnPrototype(realmConstructor.Sync.Session);
+    } else {
+        Object.defineProperty(realmConstructor, 'Sync', {
+            get: function () {
+                throw new Error("Sync is not enabled. Note that the developer edition of the Node.JS SDK for Realm does not include sync on Linux.");
```
blagoev: Maybe wording could be improved. Is this "not enabled" or it is "not available". Not sure about that.

PR #930, opened 2017-03-20
Adapter Docs

alazier:
@bigfish24 @kristiandupont @pixelchef

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -314,49 +314,49 @@ class Adapter {
  * that will exist in the instruction object.
  * @typedef Realm.Sync.Adapter~Instruction
  * @type {(INSERT|SET|DELETE|CLEAR|CHANGE_IDENTITY|LIST_SET|LIST_INSERT|LIST_ERASE|LIST_CLEAR|ADD_TYPE|ADD_PROPERTY)}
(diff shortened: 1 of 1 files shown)


alazier:
For some reason Instruction doesn't show up as `TypeDefinition` under `Adatper`. Any idea how to fix this @kristiandupont ?

docs/sync.js
```diff
@@ -261,3 +261,104 @@ class Session {
...
+
+
+/**
+ * Class for creating custom Data Connectors. Only available in the Enterprise Edition.
+ * @memberof Realm.Sync
+ */
+class Adapter {
+	/**
+	 * Create a new Adapter to moitor and process changes made across multiple Realms
+	 * @param {string} local_path - the local path where realm files are stored
```
kristiandupont: Is this where the synced realms that are being monitored will be stored? Or is it for some metadata realm? Does this mean that my server running the adapter should potentially have storage available comparable to my actual database server (or, in case they're the same, twice the space necessary for the data)?

kristiandupont:
Oh, but one more thing: is there a particular reason why we keep everyting sync-related in one file?It seems like it would be nicer to separate it.

PR #933, opened 2017-03-22
Platform agnostic test commands

blagoev:


blagoev:
retest this please

tests/test-runners/ava/package.json
```diff
@@ -3,6 +3,7 @@
     "test": "ava"
   },
   "devDependencies": {
-    "ava": "^0.16.0"
+    "ava": "^0.16.0",
+    "realm": "../../.."
```
kristiandupont: Why was this change necessary?

blagoev: removed the need to do a symlink to the realm in the parent dir

PR #938, opened 2017-03-24
Bump up test timeout

alebsack:
We've been having a lot of test failures due to this test not waiting long enough.


alebsack:
Already tried it!

PR #940, opened 2017-03-24
Client permission APIs

alazier:


alazier:
This is ready for review

alazier:
This is really ready for review now as the OS pr is merged. @kristiandupont @blagoev

src/js_permission_results.hpp
```diff
@@ -0,0 +1,150 @@
...
+                Object::set_property(ctx, js_permission, "access", Value::from_string(ctx, "Admin"));
+            else
+                Object::set_property(ctx, js_permission, "access", Value::from_string(ctx, "None"));
+
+            return_value.set(js_permission);
+        }
+
+        template<typename T>
+        void PermissionResultsClass<T>::add_listener(ContextType ctx, FunctionType, ObjectType this_object, size_t argc, const ValueType arguments[], ReturnValue &return_value) {
+            validate_argument_count(argc, 1);
```
kristiandupont: We've done this elsewhere already so maybe the point it moot, but I dislike using a standard name like `addListener` with a non-standard signature (which would be eventName, handler). I think we should either use another name or take an ignored (or maybe asserted-on only) first parameter with an event name that must be `"change"` or whatever.

tests/js/user-tests.js
```diff
@@ -87,6 +87,41 @@ function callbackTest(requestFunc, callback) {
   });
 }
 
+function getAndCheckPermission(user, expected, callback) {
+  user.getPermissions((error, permissions) => {
+    failOnError(error);
+    setTimeout(() => {
```
kristiandupont: Why is this necessary? Is it the callback problem again?

tests/js/user-tests.js
```diff
@@ -87,6 +87,41 @@ function callbackTest(requestFunc, callback) {
...
+      console.log(permissions);
+
+      failOnError(error);
+
+      TestCase.assertEqual(expected.length, permissions.length);
```
kristiandupont: Could we not use `assertArrayEqual` here?

src/js_sync.hpp
```diff
@@ -166,6 +230,102 @@ void UserClass<T>::logout(ContextType ctx, FunctionType, ObjectType this_object,
...
+                error = Value::from_undefined(protected_ctx);
+            } catch(const std::exception& e) {
+                error = Exception<T>::value(protected_ctx, e.what());
+            }
+            Function::call(protected_ctx, protected_callback, 1, &error);
+        },
+        [ctx] (auto user, auto url) {
+            Realm::Config config;
+            populate_sync_config_impl<T>(ctx, config, std::move(user), url, [](auto, auto) {});
+            return config;
```
kristiandupont: I can't quite follow the flow here (but it looks like it's correct, judging from the tests) -- why are we giving `set_permission` a callback that creates a configuration object?

alazier:
Why wasn't this ever merged? It's really frustrating to have done this work, have everything tested, and then have things change underneath the hood.

kristiandupont:
Our process has always been to merge our own PR's. Also, I had comments on it. It's fair if you disagree but you could at least let us know.

alazier:
Yeah that's totally fair. For some reason I thought this had been merged but I guess it wasn't done. I have things working again with setting permissions by email, but it needs a bit more work before we can go ahead with it.

src/js_permission_results.hpp
```diff
@@ -0,0 +1,150 @@
...
+        template<typename T>
+        void PermissionResultsClass<T>::get_index(ContextType ctx, ObjectType object, uint32_t index, ReturnValue &return_value) {
+            auto results = get_internal<T, PermissionResultsClass>(object);
+            auto permission = results->get(index);
+            auto js_permission = Object::create_empty(ctx);
+            Object::set_property(ctx, js_permission, "path", Value::from_string(ctx, permission.path));
+            Object::set_property(ctx, js_permission, "userId", Value::from_string(ctx, permission.condition.user_id));
+
+            if (permission.access == Permission::AccessLevel::Read)
+                Object::set_property(ctx, js_permission, "access", Value::from_string(ctx, "Read"));
```
alazier: How would that look?

src/js_permission_results.hpp
alazier: Which do you prefer?

tests/js/user-tests.js
alazier: I think I tried that and was having issues?

alazier:
Ok I pushed some fixes and changes needed to support permissions based on email. The newest test won't pass until we have ROS support merged and released.

src/js_permission_results.hpp
kristiandupont: Actually, I now remember that RN doesn't support this, so strike that for now.

src/js_permission_results.hpp
kristiandupont: If it's the same to you, let's add a first parameter `eventName` that is ignored (or just asserted on) for now.

ianpward:
@bmunkholm can this be merged?

bmunkholm:
With functions out now, I'm sure @kristiandupont and @blagoev can get this reviewed and merged.
@alazier are you done now?

alazier:
I think I fixed most of what was requested. Looks like this pr needs to be rebased again. Would be really helpful to split sync features into multiple files to minimize unnecessary merges in the future.

kristiandupont:
Closed and replaced with #1054

PR #948, opened 2017-03-28
enable js debugging of node tests from within vs code

blagoev:


.vscode/launch.json
```diff
@@ -0,0 +1,50 @@
+{
```
kristiandupont: Hm, this brings up an interesting point. I would normally vouch for .gitignoring the `.vscode` folder entirely, but clearly this file makes sense to keep in there. The same goes for the settings file that I have which states the unusual tab size of 4

blagoev: Before commit I checked some MS repos for that it seems they commit launch.json and tasks.json. I think we can continue to ignore .vscode and specifically settings.json there. Thus I explicitly commited these files only.

blagoev: .vscode is ignored in .gitignore currently. I vote to leave it there

kristiandupont: Agreed.

PR #959, opened 2017-04-05
Construct Results with object type name only

fealebenpae:
Fixes https://github.com/realm/realm-browser-js/issues/12

### DIFF
--- a/src/js_results.hpp
+++ b/src/js_results.hpp
@@ -55,7 +55,7 @@ struct ResultsClass : ClassDefinition<T, realm::js::Results<T>, CollectionClass<
     using ReturnValue = js::ReturnValue<T>;
 
     static ObjectType create_instance(ContextType, realm::Results);
-    static ObjectType create_instance(ContextType, SharedRealm, const ObjectSchema &);
+    static ObjectType create_instance(ContextType, SharedRealm, const std::string &object_type);
 
     template<typename U>
     static ObjectType create_filtered(ContextType, const U &, size_t, const ValueType[]);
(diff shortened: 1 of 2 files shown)


src/js_results.hpp
```diff
@@ -101,8 +101,11 @@ typename T::Object ResultsClass<T>::create_instance(ContextType ctx, realm::Resu
 }
 
 template<typename T>
-typename T::Object ResultsClass<T>::create_instance(ContextType ctx, SharedRealm realm, const ObjectSchema &object_schema) {
-    auto table = ObjectStore::table_for_object_type(realm->read_group(), object_schema.name);
+typename T::Object ResultsClass<T>::create_instance(ContextType ctx, SharedRealm realm, const std::string &object_type) {
```
alazier: Shouldn't we just fix this in object store when creating a Results object? We have run into this issue multiple times and should fix it at the root of the problem to prevent it from happening again in the future.

fealebenpae: The issue appears local to realm-js. `Results` takes a `SharedRealm` and a `Table&`, it's the table lookup in realm-js that fails.

The chain of events in this particular bug is the following:
1. `RealmClass<T>::objects` calls into `validated_object_schema_for_value` which tries to find the object type in the schema and returns a `ObjectSchema&`.
2. `ResultsClass<T>::create_instance` reads the `ObjectSchema&` name and looks up the table.
3. However, because the `ObjectSchema&` is empty, we end up with a null `TableRef` and the `Results` constructor segfaults.

Why and how the `ObjectSchema&` becomes empty I have no idea. But the fact is we only need the object type name to construct a `Results` instance, not the schema itself, so we can get away with ignoring the empty `ObjectSchema&`.

The ROS team needs this workaround for a release which is why we just patched realm-js instead of investigating any deeper underlying cause. But you're right - this is not the first time we have seen table lookup fail like that. Any insight you can provide into why this is happening and how to fix it properly would be welcomed.

alazier: Yeah we should definitely ship this fix. Was just wondering if there was a more general fix in object store since we have hit this multiple times now.

radu-tutueanu:
We should merge the fix to master and then try to move it to OS in a separate PR.

PR #960, opened 2017-04-05
Fail on Sync use when Sync is not available

blagoev:


kristiandupont:
I don't understand why this is necessary as we should not be callig the extensions more than once, but the fix is fine and I guess if this solves it we're good.

blagoev:
This does not work on node v4. I will look into different approach for this one

kristiandupont:
How does it fail?

blagoev:
It fails with Proxy is not defined.

kristiandupont:
This has been solved by unified packaging

PR #962, opened 2017-04-06
Windows android build

blagoev:
This switches every symlink to a directory junctions on Windows on pre-build step. Junctions do not require elevated process. The problem with having junctions is that git will show there are new files at the symlink place. Thus junctions are not good to be used by post-checkout script etc. Thus here they are used only on build so the build can succeed and then junctions are reverted back to symlinks as a post-build step.

So after build finishes it reverts back the the junctions to symlinks. If the build fails the revert logic is still executed.
If build is force stopped (CTRL+C) the junction points are left behind. 
This means user need to use "npm run win-revert-symlinks" to manually revert junctions to symlinks. There is a script for manually fixing symlinks if one needs to "npm run win-fix-symlinks"

The scripts for the junction-symlink convertion are from here http://stackoverflow.com/questions/5917249/git-symlinks-in-windows 

Although seemingly complex I think this handles the situation pretty fine. I think a section in the readme describing how things work will be enough to prepare the developer.

Also currently sync tests (users-tests.js and session-tests.js) are failing since we can't start the ros on windows. Maybe a way to switch these tests off if realm-js is build and run on Windows will be ok.

Let's review and discuss this and then I can update the readme with instructions.

### DIFF
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -31,10 +31,17 @@ import org.apache.tools.ant.filters.ReplaceTokens
 // We download various C++ open-source dependencies into downloads.
 // We then copy both the downloaded code and our custom makefiles and headers into third-party-ndk.
 // After that we build native code from src/main/jni with module path pointing at third-party-ndk.
+def npmExecutable = 'npm';
 
-ext.coreVersion = "npm --silent run get-core-version".execute().text.trim()
-ext.syncVersion = "npm --silent run get-sync-version".execute().text.trim()
-def currentVersion = "npm --silent run get-version".execute().text.trim()
+if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        npmExecutable = 'npm.cmd'
+}
+ext.coreVersion = "$npmExecutable --silent run get-core-version".execute().text.trim()
+ext.syncVersion = "$npmExecutable --silent run get-sync-version".execute().text.trim()
+println "Realm Core Version: $ext.coreVersion"
+println "Realm Sync Version: $ext.syncVersion"
+
+def currentVersion = "$npmExecutable --silent run get-version".execute().text.trim()
 def downloadsDir = new File("$projectDir/downloads")
 def jscDownloadDir = new File("$projectDir/src/main/jni/jsc")
 def coreDownloadDir = new File("$projectDir/src/main/jni")
--- a/scripts/react-tests-android.js
+++ b/scripts/react-tests-android.js
@@ -1,0 +1,48 @@
+var shell = require("shelljs");
+
+shell.set("-e");
+
+shell.pushd("./tests/react-test-app");
+
+shell.exec("npm install");
+
+if (process.platform !== "win32") { 
+    shell.env["PATH"] = "/opt/android-sdk-linux/platform-tools:$PATH"
+}
+
+if (shell.env["REALM_BUILD_ANDROID"]) {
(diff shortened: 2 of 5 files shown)


scripts/react-tests-android.js
```diff
@@ -0,0 +1,49 @@
+var shell = require("shelljs");
+
+shell.set("-e");
+
+shell.pushd("./tests/react-test-app");
+
+shell.exec("npm install");
```
fealebenpae: `yarn` is not a required dependency of React Native, so better not rely on it.

react-native/android/build.gradle
```diff
@@ -31,10 +31,17 @@ import org.apache.tools.ant.filters.ReplaceTokens
...
 // After that we build native code from src/main/jni with module path pointing at third-party-ndk.
+def npmExecutable = 'npm';
 
-ext.coreVersion = "npm --silent run get-core-version".execute().text.trim()
-ext.syncVersion = "npm --silent run get-sync-version".execute().text.trim()
-def currentVersion = "npm --silent run get-version".execute().text.trim()
+if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        npmExecutable = 'npm.cmd'
+}
+ext.coreVersion = "$npmExecutable --silent run get-core-version".execute().text.trim()
```
fealebenpae: We carry `dependencies.list` in the npm package. Wouldn't it be simpler to just parse the file (it's ini)?

blagoev: I tried to reduce the churn here and preserve the logic as much as possible. These lines were failing on windows hence this fix. Also we are using the values from the dependencies.list as environment variables on some places during the build so it will require us to do the same thing that the npm helper package does for us.

blagoev: Also If we can remove the symlinks in the repo it will remove the need to do this pre/post build actjions which will be way more better. Not sure if we can do it easely. 
Currently these are the symlinks

examples/ReactNativeBenchmarks/.eslintrc.json <<===>> ../ReactExample/.eslintrc.json
react-native/android/src/main/jni/src <<===>> ../../../../../src/
react-native/android/src/main/jni/vendor <<===>> ../../../../../vendor
tests/react-test-app/android/app/src/main/assets <<===>> ../../../../../data

dbhowell: #972 may assist with this

blagoev: @dbhowell The changes in this PR actually make the use of npm platform independent
The build.gradle still uses npm but the npm script is changed like so
https://github.com/realm/realm-js/pull/962/commits/55bc8869048b98bd1bfdccc36db862947bd240de#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R49

You have implemented what @fealebenpae have suggested above for these exact npm scripts, so we can merge your pr with this one.

Still we need to decide what to do with the symlinks in the repo before merging this PR

@fealebenpae @kristiandupont What you guys think?

dbhowell: While I agree that using `npm` makes it platform independent, it also increases the complexity (more points of failure, having to test for platform type, more `npm` dependencies etc).

Personally, I think having all gradle builds independent of both platform AND `npm` is a 'Good Thing'. That's just my opinion and completely understand that other's will disagree.

blagoev: @dbhowell  I totally agree with you. This npm redirection is not the best thing. I will try merging your PR when I make the builds green. thanks a lot.

PR #968, opened 2017-04-12
Removed superfluous process handlers

alebsack:
I noticed this when developing RaaS -- loading the realm-js in node is a bit intrusive when setting up process event handlers.

The `exit` handler is always called, regardless of what you do in `SIGINT` and `uncaughtException`, so we don't need to do `Sync.cleanup()` in those.  

In the case of `SIGINT`, the code was exiting with an error status.  Catching ctrl-C (interrupt) and exiting should not be considered an error condition (or we should not be forcing this on the users of realm-js).

I think we should also leave the `uncaughtException` handler undefined, as we do not want to force stack trace outputs on our users -- they should be handling process lifecycle themselves.

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -52,16 +52,6 @@ module.exports = function(realmConstructor) {
         if (realmConstructor.Sync.cleanup) {
             // FIXME: DOES THIS WORK ON BOTH NODE AND REACT NATIVE?
             process.on('exit', realmConstructor.Sync.cleanup);
-            process.on('SIGINT', function () {
-                realmConstructor.Sync.cleanup();
-                process.exit(2);
-            });
-            process.on('uncaughtException', function(e) {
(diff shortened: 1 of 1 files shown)


alebsack:
@alazier Do you think we can at least do without the `SIGINT` handler?  Cleanup should still be called on the `exit` handler.

alazier:
@alebsack that was the original motivation for adding this as C-c would cause segfaults. If we can find another way to prevent the gn and other sync users from crashing when C-c then that would obviously be a preferable solution. Perhaps @kristiandupont or someone else on the js team can dig in if it is high priority.

kristiandupont:
I don't feel that this is important enough to go over again? It could be improved but it works and there are a number of things that are higher priorities.

blagoev:
Are we working on this or we can close the PR.

PR #972, opened 2017-04-19
Use dependencies.list file instead of npm for build.gradle

dbhowell:
**Goals**: To remove the need to use npm in build.gradle which makes the build process more platform independent.

**Expected Results**: `./gradlew publishAndroid` to continue to build on MacOS & Linux and closer to building on Microsoft Windows (bar symlink issues)

This may assist with #962 and also https://github.com/realm/realm-js/pull/962#discussion_r110385252

While the helper function loads and parses the dependencies.list each time it's used, the overhead isn't noticeable and arguably faster than using the `npm.execute()` call.


blagoev:
I can confirm the change is working on windows. Will proceed to merging

PR #989, opened 2017-04-28
report undefined deletions on deleteAll instead of max int count indexes

blagoev:
This resolves an issue where a call to deleteAll reports max int rows deleted and js binding consumes all memory trying to report all indexes in the range of 0 to max int. 
This is a fix to a limitation of the current implementation which may be changed in the future.

It means users will receive undefined value as a single result in deletions property  in their event handler

### DIFF
--- a/src/js_collection.hpp
+++ b/src/js_collection.hpp
@@ -50,7 +50,7 @@ typename T::Value CollectionClass<T>::create_collection_change_set(ContextType c
     std::vector<ValueType> deletions, insertions, modifications;
     
     if (change_set.deletions.count() == std::numeric_limits<size_t>::max()) {
-        deletions.push_back(Value::from_undefined(ctx));
+        deletions.push_back(Value::from_null(ctx));
     }
     else {
         for (auto index : change_set.deletions.as_indexes()) {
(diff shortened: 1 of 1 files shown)


src/js_collection.hpp
```diff
@@ -48,8 +48,14 @@ typename T::Value CollectionClass<T>::create_collection_change_set(ContextType c
 {
     ObjectType object = Object::create_empty(ctx);
     std::vector<ValueType> deletions, insertions, modifications;
-    for (auto index : change_set.deletions.as_indexes()) {
-        deletions.push_back(Value::from_number(ctx, index));
+    
+    if (change_set.deletions.count() == std::numeric_limits<size_t>::max()) {
+        deletions.push_back(Value::from_undefined(ctx));
```
kristiandupont: I guess from Max' comments, perhaps we should stick with `null`?

blagoev: I thought about that as well. Will asking him for input here. We should pick the most meaningful value here. I choose 'undefined' since the naming is more meaningful to the natural English language.  It kinda falls more to the result we are trying to communicate. The server knows there are deletions but their count is "undefined". Not sure this is a correct/best way to communicate this. I wanted to escape from things like -1 or Positive infinity or some other obscure value.

mbalex99: `undefined` is usually developer "written code". I have a **strong** opinion that libraries or frameworks should never return to the caller anything that is `undefined` unless it's introspecting metadata of some sort. 

1. `undefined` in JavaScript means that the variable has been declared but not assigned anything.
2. `null` means a representation of no value.

blagoev: thanks will switch to null value here

PR #1000, opened 2017-05-04
Use sync 1.8.3

radu-tutueanu:


radu-tutueanu:
Lost my patience and started a new build, but looks like https://ci.realm.io/job/realm/job/realm-js/job/PR-1000/5/ passed. Merging.

PR #1005, opened 2017-05-06
Download api (aka open async api)

### LINKED ISSUE #973: Download Realm API

blagoev:
This is the related discussion https://github.com/realm/realm-js/issues/973

Includes Realm.open and Realm.openAsync methods. 
Let's decide which one we like to keep. We can shape this in many ways.

currently does not pass the error code to the user. Will be fixing it.

This diverges somewhat from the cocoa implementation which fails when there is no sync session in place. 
Here instead I have done this to resolve the promise (succeed the callback) so it can be used with local realms as well.

Note:  Currently if the callback throws an exception it is propagated down to C++ and it brings the process down from here  https://github.com/realm/realm-js/pull/1005/commits/44432660eae01defa807837c58b37db9f9ce00fd#diff-3a80f0d3292e26debf0ac0175cacbc0aR47

This resolves #973


blagoev:
Also includes https://github.com/realm/realm-js-private/pull/246

this comment
https://github.com/realm/realm-js-private/pull/246#issuecomment-267154104

blagoev:
Ok on RN the callback is taking to much time when a promise is resolved. Exactly the same as the issue of MakeCallback on node but if you send any event from the RN app it seems to be able to continue. This issue needs to be addressed.

It seems the call to JSObjectCallAsFunction returns as normal.

src/js_realm.hpp
```diff
@@ -220,6 +223,7 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
         {"deleteAll", wrap<delete_all>},
         {"write", wrap<write>},
         {"addListener", wrap<add_listener>},
+        {"wait", wrap<wait_for_download_completion>},
```
kristiandupont: I get that this is not the real api, but "wait" seems a bit generic to me. Is there any reason it's not simply `onRealmSynchronized` or something similar? I mean, it *is* public, right?

blagoev: yes this is public. I just kept in sync to the other methods. We can shape it better I guess. 
My reasoning here is user is not aware what wait_for_download means. So generic wait method after open seemed better. like Realm.open(config).wait().
And also wait_for_download does not make sense in local realms but I made it possible to use wait for that as well. 
If this is exposed as onRealmSynchronized it may leave the user with impression it can be called multiple times during the app lifetime. And still the name don't make sense when opening local realms.

blagoev:
I did two version of the same method open that returns Promise and openAsync which uses callbacks with the intention to remove one of them after discussions. Also we may choose to leave the callback one and rename it to "open"

kristiandupont:
I wrote that comment before realizing that you had already implemented both. I don't think it's a problem to offer both -- it would have been a problem to only offer the promise one.

lib/extensions.js
```diff
@@ -40,6 +40,34 @@ module.exports = function(realmConstructor) {
     setConstructorOnPrototype(realmConstructor.Results);
     setConstructorOnPrototype(realmConstructor.Object);
 
+    //Add async open API
+    Object.defineProperties(realmConstructor, getOwnPropertyDescriptors({
+        open(config) {
+            return new Promise((resolve, reject) => {
+                const realm = new realmConstructor(config);
```
fealebenpae: I don't think that's correct. This will open a new Realm instance with the schema which will push to the server. Cocoa actually creates a special Realm that doesn't own a schema and wait on its sync session before creating a regular Realm - https://github.com/realm/realm-cocoa/blob/master/Realm/RLMRealm.mm#L193.

I believe what we ought to do is add a `Realm.Sync._getSession(syncConfig)` which maps to `SyncManager::get_session)`, call `_waitForDownload(callback)` on the Session and in the callback resolve the promise with a new Realm. Then this code becomes:
```js
const session = realmConstructor.Sync._getSession(config.sync);
session._waitForDownload((error) => {
    if (error) reject(error);
    else resolve(new realmConstructor(config));
});
```

src/js_realm.hpp
fealebenpae: I think this should be a method on `Realm.Sync.Session` instead.

lib/extensions.js
blagoev: Big oversight from me. Will be reworking this. Thanks

blagoev: @fealebenpae 
/cc @kristiandupont 

While I am working on using a schema-less Realm as the cocoa approach, isn't their approach achievable by doing 

```JavaScript
open(config) {
                const realm = new realmConstructor();
                realm.wait((error) => {
                    if (error) {
                        reject(new Error("Cannot asynchronously open synced Realm, because the associated session previously experienced a fatal error"));
                    }

                    const syncedRealm = new realmConstructor(config);
                    resolve(syncedRealm);
                });
            });
        },
```

It boils down to the question what var realm = new Realm() is doing. Isn't this creating a schema-less local realm. By looking at the code it seems it does exactly that.

The end result of my query to you is that this change can help keep the current proposed api Realm.open. It's the internal implementation that will change a bit.

What do you think?

blagoev: I am going to test this anyways

blagoev: Ok maybe this does not work that way. Will need to create an actual local Realm with no schema as the cocoa approach

blagoev:
I switched to Realm._waitForDownload
We may choose to remove the underscore. Still I think users should not use this method. They still can if needed for advanced scenario/workaround

Also Realm.open is the one that returns a Promise
Realm.openAsync is the one with the callback.

Do we need to switch the naming on these two?
I personally like it open to be the Promise version = more future proof

tests/js/session-tests.js
```diff
@@ -70,18 +111,129 @@ module.exports = {
...
+
+        return new Promise((resolve, reject) => {
+            const child = execFile('node', [tmpFile.name, username, realmName, REALM_MODULE_PATH], { cwd: tmpDir.name }, (error, stdout, stderr) => {
+                if (error) {
+                    reject(new Error('Error executing download api helper' + error));
+                }
+                resolve();
+            });
+        })
+            .then(() => {
```
alazier: spacing seems weird.

tests/js/session-tests.js
```diff
@@ -70,18 +111,129 @@ module.exports = {
...
+        const realmName = uuid();
+        const expectedObjectsCount = 3;
+
+        let tmpDir = tmp.dirSync();
+        let content = fs.readFileSync(__dirname + '/download-api-helper.js', 'utf8');
+        let tmpFile = tmp.fileSync({ dir: tmpDir.name });
+        fs.appendFileSync(tmpFile.fd, content, { encoding: 'utf8' });
+
+        return new Promise((resolve, reject) => {
+            const child = execFile('node', [tmpFile.name, username, realmName, REALM_MODULE_PATH], { cwd: tmpDir.name }, (error, stdout, stderr) => {
```
alazier: Might be nicer if we commented to say what this other script does.

tests/js/download-api-helper.js
```diff
@@ -0,0 +1,31 @@
...
+        var realm = new Realm(config);
+
+        realm.write(() => {
+            for (let i = 1; i <= 3; i++) {
+                realm.create('Dog', { name: `Lassy ${i}` });
+            }
+        });
+
+        console.log("Dogs count " + realm.objects('Dog').length);
+        setTimeout(_ => process.exit(0), 3000);
```
alazier: its really annoying to have to wait 3 seconds every time you run a test. Would be nicer if we had a notification which we could do with the global notifier.

tests/js/session-tests.js
```diff
@@ -70,18 +111,129 @@ module.exports = {
...
+        const username2 = uuid();
+        const realmName = uuid();
+        const expectedObjectsCount = 3;
+
+        let tmpDir = tmp.dirSync();
+        let content = fs.readFileSync(__dirname + '/download-api-helper.js', 'utf8');
+        let tmpFile = tmp.fileSync({ dir: tmpDir.name });
+        fs.appendFileSync(tmpFile.fd, content, { encoding: 'utf8' });
+
+        return new Promise((resolve, reject) => {
```
alazier: lots of code duplication with the other open test should consider refactoring.

tests/js/session-tests.js
```diff
@@ -70,18 +111,129 @@ module.exports = {
...
+                    return new Promise((resolve, reject) => {
+                        const accessTokenRefreshed = this;
+                        let successCounter = 0;
+
+                        let config = {
+                            sync: { user, url: `realm://localhost:9080/~/${realmName}` },
+                            schema: [{ name: 'Dog', properties: { name: 'string' } }],
+                        };
+
+                        Realm.openAsync(config, (error, realm) => {
```
alazier: So this is the same as `open` but with different error handling?

lib/extensions.js
```diff
@@ -40,6 +40,35 @@ module.exports = function(realmConstructor) {
...
+    Object.defineProperties(realmConstructor, getOwnPropertyDescriptors({
+        open(config) {
+            return new Promise((resolve, reject) => {
+                realmConstructor._waitForDownload(config, (error) => {
+                    if (error) {
+                        reject(error);
+                    }
+
+                    let syncedRealm = new realmConstructor(config);
+                    //FIXME: RN hangs here. Remove when node's makeCallback alternative is implemented
```
alazier: we should really just get the makeCallback pr implemented/merged - think the work is already done?

src/js_realm.hpp
```diff
@@ -206,6 +209,7 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
         {"schemaVersion", wrap<schema_version>},
         {"clearTestState", wrap<clear_test_state>},
         {"copyBundledRealmFiles", wrap<copy_bundled_realm_files>},
+        {"_waitForDownload", wrap<wait_for_download_completion>},
```
alazier: Do we need to use the `_`? Any reason this shouldn't be something users could call themselves?

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+
+template<typename T>
+void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType, ObjectType this_object, size_t argc, const ValueType arguments[], ReturnValue &return_value) {
+    validate_argument_count(argc, 2);
+    auto config_object = Value::validated_to_object(ctx, arguments[0]);
+    auto callback_function = Value::validated_to_function(ctx, arguments[1]);
+
+    ValueType sync_config_value = Object::get_property(ctx, config_object, "sync");
+    if (!Value::is_undefined(ctx, sync_config_value))
+    {
```
alazier: In the rest of the codebase brackets are not on newlines.

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+        realm::Realm::Config config;
+        static const String encryption_key_string = "encryptionKey";
+        ValueType encryption_key_value = Object::get_property(ctx, config_object, encryption_key_string);
```
alazier: Isn't this parsed out somewhere else? Would be nice to only have a single code path for parsing config objects.

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+        if (!Value::is_undefined(ctx, encryption_key_value))
+        {
```
alazier: brackets

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+                ValueType callback_arguments[1];
+                callback_arguments[0] = object;
+                Function<T>::callback(protected_ctx, protected_callback, protected_this, 1, callback_arguments);
+            }
+        });
+        std::function<WaitHandler> waitFunc = std::move(wait_handler);
+
+        auto realm = realm::Realm::get_shared_realm(config);
+        if (auto sync_config = config.sync_config)
+        {
```
alazier: brackets

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+        Protected<ObjectType> thiz(ctx, this_object);
+        SyncClass<T>::populate_sync_config(ctx, thiz, config_object, config);
+
+        Protected<FunctionType> protected_callback(ctx, callback_function);
+        Protected<ObjectType> protected_this(ctx, this_object);
+        Protected<typename T::GlobalContext> protected_ctx(Context<T>::get_global_context(ctx));
+
+        EventLoopDispatcher<WaitHandler> wait_handler([=](std::error_code error_code) {
+            HANDLESCOPE
+            if (error_code == std::error_code{}) {
```
alazier: Is there not a cleaner way to check for no error?

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+            ObjectType object = Object::create_empty(protected_ctx);
+            Object::set_property(protected_ctx, object, "message", Value::from_string(protected_ctx, "Cannot asynchronously open synced Realm, because the associated session previously experienced a fatal error"));
+            Object::set_property(protected_ctx, object, "errorCode", Value::from_number(protected_ctx, 1));
+
+            ValueType callback_arguments[1];
+            callback_arguments[0] = object;
+            Function<T>::call(protected_ctx, protected_callback, protected_this, 1, callback_arguments);
+            return;
+        }
+
```
alazier: extra line

src/js_realm.hpp
```diff
@@ -537,6 +541,84 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
...
+                ValueType callback_arguments[1];
+                callback_arguments[0] = object;
+                Function<T>::callback(protected_ctx, protected_callback, protected_this, 1, callback_arguments);
+            }
+        });
+        std::function<WaitHandler> waitFunc = std::move(wait_handler);
+
+        auto realm = realm::Realm::get_shared_realm(config);
+        if (auto sync_config = config.sync_config)
```
alazier: wont this always be true at this point?

tests/js/session-tests.js
blagoev: VS Code formater does that. Will try to fix it

tests/js/session-tests.js
blagoev: openAsync provides different API to the user, the open method returns a Promise. Check this comment for why that is https://github.com/realm/realm-js/pull/1005#issuecomment-299821078

tests/js/download-api-helper.js
blagoev: The wait is there to allow the syncing to happen, ie not to end the process before the sync is done. Otherwise the test will fail always since we want to test the download api waited for the exact number of records.

lib/extensions.js
blagoev: The PR is usint the makeCallback implementation which is done as a commit in this PR. It will be merged when this PR merges.

The issue is with ReactNative where there is no makeCallback available and this hangs. The code that we use seems all fine but still we hang on this piece of code. There are still ways to implement this correctly but they need more research. Hence the FIXME comment.

src/js_realm.hpp
blagoev: Check this comment on the underscore name. https://github.com/realm/realm-js/pull/1005#issuecomment-300915003

src/js_realm.hpp
blagoev: VS Code C++ formatter. will fix it

src/js_realm.hpp
blagoev: no its not. This is always done before the call to parse the config object. This is taken from here: 
https://github.com/realm/realm-js/blob/master/src/js_realm.hpp#L349-L357

I have included it because the end result should be valid realm::Realm::Config config; object the sync properties.

src/js_realm.hpp
blagoev: will fix

src/js_realm.hpp
blagoev: Not sure. Any suggestions will help. The cocoa implementation is doing this exactly like that.

src/js_realm.hpp
blagoev: I don't see what is guaranteeing it in the code above.

src/js_realm.hpp
blagoev: will fix

src/js_realm.hpp
blagoev: will fix

src/js_realm.hpp
bdash: `std::error_code` provides an `operator bool`, so you could do `if (!error_code) {`

blagoev: fixed

astigsen:
Does the current API support marking the realm as read-only?

blagoev:
@astigsen Merged before noticing your comment. 

Yes the current api supports opening the realm as read only using the configuration object readOnly property. This PR uses the same config object so it should be working.

PR #1023, opened 2017-05-17
Update to work with new object store object accessor API

tgoyne:
This does the bare minimum to get things to compile and pass tests with objectstore master. There's probably more that should be done, but I won't have time to do it.


bdash:
The object accessor refactoring lays the groundwork for arrays of primitives. It's necessary for realm-js to update to the new API to pull in any new version of object store, including the soon-to-be-merged stable ID work.

bdash:
This is ready to go once we have a core release that includes realm/realm-core#2624.

bdash:
Core v2.8.1 includes realm/realm-core#2624. A Sync release that picks up that version of core should be available tomorrow.

radu-tutueanu:
We need to update the node-server-sdk to work with the object accessor as well: https://github.com/realm/realm-js-private/pull/278

PR #1027, opened 2017-05-18
Update docs for Realm.open and Realm.openAsync

kristiandupont:


docs/realm.js
```diff
@@ -76,6 +76,25 @@ class Realm {
     constructor(config) {}
 
     /**
+     * Open a realm asynchronuously with a promise. If the realm is synced, it will be fully
```
blagoev: asynchronuously = asynchronously

PR #1032, opened 2017-05-25
Adding TypeScript Definitions 

mbalex99:
@alazier maybe you can help me with this. 

1. I added the `tsconfig.json`
2. I added the `index.d.ts` 
3. I added `index.d.ts` to a reference in the `package.json` 

However the final artifacts seem to be different from the `realm-js` repo. How would I tell the publishing of the files to include `index.d.ts` and `tsconfig` at the root level of the installed package?


index.d.ts
```diff
@@ -0,0 +1,469 @@
...
+// Definitions by: Akim <https://github.com/Akim95> & Maximilian Alexander <https://github.com/mbalex99> 
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
+
+declare namespace Realm {
+    /**
+     * PropertyType
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.html#~PropertyType }
+     */
+    type PropertyType = string | 'bool' | 'int' | 'float' | 'double' | 'string' | 'data' | 'date' | 'list';
```
kristiandupont: This should include `Object`, shouldn't it?

bang88:
Maybe we can rewrite js using TypeScript so we can generate type definitions automatically, and then we can keep sync with master js files ,no need maintain a separate type definition file ?

index.d.ts
mbalex99: The docs specify only 

Type: "bool" or "int" or "float" or "double" or "string" or "date" or "data" or "list" or "<ObjectType>"

"<ObjectType>" is a string for another name of an object. Thus, that's why this TypeScript string literal has an accepted `string` native type.

mbalex99:
@bang88 I don't think we can do this without a very complicated build process. We are invoking untranspiled native JS from C++.

Looks like the easiest way to give happy TypeScript users support is to maintain the .d.ts file for now.

bang88:
By the way, we should update type definitions support https://realm.io/docs/javascript/latest/#models constructor

```js
class Person {
  get ageSeconds() {
    return Math.floor((Date.now() - this.birthday.getTime()));
  }
  get age() {
    return ageSeconds() / 31557600000;
  }
}

Person.schema = PersonSchema;

// Note here we are passing in the `Person` constructor
let realm = new Realm({schema: [CarSchema, Person]});
```
the code will have a error when use typescript , due to `Person` not assignable to type  `ObjectSchema`, Person not contains the keys which ObjectSchema defined.

I would like to using class as a Model but sucked at here. any chance to support this situation ?

Sorry for comments here. 

Thanks again.

mbalex99:
We are definitely looking at this at the moment! Great 👍 idea!

bdash:
The commit history on this branch looks a bit whacky now. Can we clean it up before it is merged?

mbalex99:
Closing in favor of #1076

PR #1055, opened 2017-06-09
Add objectSchema method to Realm Object

bigfish24:
Not sure if this is correct, since I couldn't get the tests to run.

### DIFF
--- a/tests/js/object-tests.js
+++ b/tests/js/object-tests.js
@@ -470,6 +470,14 @@ module.exports = {
             obj.doubleCol;
         });
     },
+    
+    testObjectSchema: function() {
+        var realm = new Realm({schema: [schemas.TestObject]});
+        realm.write(function() {
+            var obj = realm.create('TestObject', {doubleCol: 1});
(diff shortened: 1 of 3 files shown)


alazier:
You should use a method `objectSchema()` rather than a property.

tests/js/object-tests.js
```diff
@@ -470,6 +470,14 @@ module.exports = {
             obj.doubleCol;
         });
     },
+    
+    testObjectSchema: function() {
+        var realm = new Realm({schema: [schemas.TestObject]});
+        realm.write(function() {
+            var obj = realm.create('TestObject', {doubleCol: 1});
+            TestCase.assertEqual(obj.objectSchema, schemas.TestObject);
```
alazier: Need to change this to a method.

bigfish24:
@blagoev I can't get the tests to run and this fails CI... not sure what to do?

blagoev:
@bigfish24 I will take a look why that is

blagoev:
maybe we need to add it to the docs as well

PR #1065, opened 2017-06-13
Support the new React Native C++ bridge

### LINKED ISSUE #1049: [React Native v0.45.0-rc.3] [iOS] Realm has no setter or ivar for its bridge
When trying to use Realm in the soon-to-be-stable v0.45 of React Native the `setBridge` function fails with the following (which causes the exception in the issue title to show in a red box) on https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmReact.mm#L275

```
[<RCTBridge 0x6100000a5e20> valueForUndefinedKey:]: this class is not key value coding-compliant for the key javaScriptExecutor.
```

I'm assuming this has to do with the newer C++ bridge. 

---

```
Realm: 1.3.1
React Native: v0.45.0-rc.3
Platform: iOS (Android seems to work great)
```

fealebenpae:
Fixes #1049

### DIFF
--- a/react-native/ios/RealmReact/RealmReact.mm
+++ b/react-native/ios/RealmReact/RealmReact.mm
@@ -265,6 +265,22 @@ RCT_REMAP_METHOD(emit, emitEvent:(NSString *)eventName withObject:(id)object) {
     [self performSelectorOnMainThread:@selector(invalidate) withObject:nil waitUntilDone:YES];
 }
 
+typedef JSGlobalContextRef (^JSContextRefExtractor)();
+
+void _initializeOnJSThread(JSContextRefExtractor jsContextExtractor) {
+    // Make sure the previous JS thread is completely finished before continuing.
+    static __weak NSThread *s_currentJSThread;
+    while (s_currentJSThread && !s_currentJSThread.finished) {
+        [NSThread sleepForTimeInterval:0.1];
+    }
+    s_currentJSThread = [NSThread currentThread];
+
+    // Close all cached Realms from the previous JS thread.
+    realm::_impl::RealmCoordinator::clear_all_caches();
+
+    RJSInitializeInContext(jsContextExtractor());
+}
+
 - (void)setBridge:(RCTBridge *)bridge {
     _bridge = bridge;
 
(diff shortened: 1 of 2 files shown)


react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -265,23 +265,54 @@ - (void)dealloc {
...
         @throw [NSException exceptionWithName:@"Invalid Executor" reason:@"Chrome debug mode not supported in Release builds" userInfo:nil];
 #endif
-    }
-    else {
+    } else if ([bridge isKindOfClass:objc_lookUpClass("RCTCxxBridge")]) {
+        // probe for the new C++ bridge in React Native 0.45+
+
+        __weak __typeof__(self) weakSelf = self;
+        __weak __typeof__(bridge) weakBridge = bridge;
+        [bridge performSelector:@selector(executeBlockOnJavaScriptThread:) withObject:^{
```
bdash: Is there a reason to use `performSelector:withObject:` here rather than just doing `[bridge executeBlockOnJavaScriptThread:^{ … }]`?

react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -265,23 +265,54 @@ - (void)dealloc {
...
+            __typeof__(self) self = weakSelf;
+            __typeof__(bridge) bridge = weakBridge;
+            if (!self || !bridge) {
+                return;
+            }
+
+            _initializeOnJSThread(^ {
+                return (__bridge JSGlobalContextRef)[bridge performSelector:@selector(jsContextRef)];
```
bdash: Ditto about just doing `return [bridge jsContextRef]` instead.

react-native/ios/RealmReact/RealmReact.mm
fealebenpae: `executeBlockOnJavaScriptThread:` isn't declared publicly on `RCTBridge` - it's a method on `RCTCxxBridge` which is privately-declared.

react-native/ios/RealmReact/RealmReact.mm
fealebenpae: Same as above.

fealebenpae:
@bdash does it make sense for me to declare a category on `RCTBridge` with those two methods so I can invoke them directly? `setBridge:` is not on a hot path, so we don't care about the overhead from `performSelector:`, but if the other way's more correct...

bdash:
Declaring the methods in a category would be more correct, and makes the call sites more readable.

thatjuan:
After applying this patch directly to a 1.3.1 install on RN 0.45.1, I get the following error:

```
template <class T>
constexpr Optional<T>::operator bool() const
{
    return m_engaged;  // EXC_BAD_ACCESS
}
```
And this (potentially unrelated) output:

```
[warn][tid:NSOperationQueue 0x60800003ef20 :: NSOperation 0x610000052c90 (QOS: DEFAULT)][RCTModuleData.mm:232] RCTBridge required dispatch_sync to load RCTDevLoadingView. This may lead to deadlocks
```

fealebenpae:
Thanks, @thatjuan, I'll try to reproduce it. Could you also attach a full stack trace, please?

thatjuan:
@fealebenpae There is a chance that error was caused by xcode 8.3.3. I switched back to  8.2.1 and it gets past that. However, I'm now stuck at:

```2017-06-13 23:55:56.217 [fatal][tid:main] mutex lock failed: Invalid argument```

I can't tell if this is caused by Realm or not. I'm running this on a clean simulator and it consistently crashes there.

screenshot 2017-06-14 00 03 35

fealebenpae:
@thatjuan you can add a break on exceptions breakpoint in Xcode to see where the error might occur.

fealebenpae:
@thatjuan I haven't been able to reproduce this. It would be best if you can add an exception breakpoint in Xcode and see where the error originates.

dzunglht:
Build release doesn't work, app crashes !
Thanks

fealebenpae:
@dzunglht can you share more details? Xcode version, React Native version, etc.

dzunglht:
It works OK in debug mode but app gets crashed in release mode
RN: 0.45.1
Xcode:  8.3.2 (8E2002)
P/S: i already applied this patch: https://raw.githubusercontent.com/realm/realm-js/822b7f600d87a7f2d82fc9ba73815b9846b54809/react-native/ios/RealmReact/RealmReact.mm

image

fealebenpae:
Did you apply https://github.com/realm/realm-js/pull/1065/commits/3cba96e6ff543eb87886c5c493c42e1070d1e16f as well?

dzunglht:
@fealebenpae Yes, i did, how many patches i must apply ? Would you please list them out ?
Many thanks

fealebenpae:
Could you please archive your project and share it so I can debug it? Alternatively, add an exception breakpoint in Xcode and inspect the underlying error.

thatjuan:
@fealebenpae Sorry it took me awhile. Now my project is crashing here:

screenshot 2017-06-15 19 47 51

Here's the variable inspector at the crashpoint:

screenshot 2017-06-15 19 55 36

I'm going to try to pinpoint if a specific call from the JS side is triggering this crash, but it looks like it happens at startup.

I'm running this on xcode 8.2.1 on an iOS 10.3 simulator.

bdash:
@thatjuan, may I suggest that you file a new GitHub issue and include all of the information that you've outlined so far? That'll ensure that the details of the problem you're seeing don't get lost in this now-merged pull request.

PR #1046, opened 2017-06-06
Descriptive error message on create

blagoev:
This is related to https://github.com/realm/realm-object-store/pull/468

PR #1069, opened 2017-06-14
Include property names in type error exceptions thrown by Realm.create

### LINKED ISSUE #787: More descriptive error message when creating objects
If setting a property to the wrong type when creating an object, you still get the vague error message `'Property' must be of type: string` etc.

bdash:
This is an alternative approach to that taken in #1046.

Fixes #787.

It could still do with more thorough tests.


bdash:
It looks like there's a failure when building for Node on Windows. I'll look into that tomorrow.

bdash:
Windows issue is fixed, so this is now complete and awaiting review.

PR #1076, opened 2017-06-15
adding ts and package.json typings

mbalex99:
We just have to make sure that the `package.json` has a `typings` property that points to `index.d.ts` and make sure that's discoverable in the finalized package.


index.d.ts
```diff
@@ -0,0 +1,471 @@
+// Type definitions for realm-js 1.8.0
```
bdash: Is it necessary to include the version number in this file? It seems likely to get out of sync unless it's automatically updated as part of the release process.

index.d.ts
```diff
@@ -0,0 +1,471 @@
...
+// Project: https://github.com/realm/realm-js
+// Definitions by: Maximilian Alexander <https://github.com/mbalex99>
```
bdash: Shouldn't this instead have the normal Realm copyright / license notice?

index.d.ts
mbalex99: 👍

index.d.ts
mbalex99: 👍

bdash:
Is it not possible to include our [normal license header](https://github.com/realm/realm-js/blob/master/lib/index.js#L1-L17) in the .d.ts file?

mbalex99:
Isn't it redundant because it's in this repo?

bdash:
It's present as a header in every other source file in the repository. If there's not a technical reason for it to be omitted here, it should be included.

index.d.ts
```diff
@@ -0,0 +1,485 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
bdash: 2017 :-)

mbalex99:
@kristiandupont and @blagoev can you guys take a look at the `index.d.ts`

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     * @param  {any} encryptionKey?
+     * @returns number
+     */
+    static schemaVersion(path: string, encryptionKey?: any): number;
+
+    /**
+     * @param  {Realm.Configuration} config?
+     */
+    constructor(config?: Realm.Configuration);
+
```
kristiandupont: The alternative `open` and `openAsync` factories seem to be missing.
https://github.com/realm/realm-js/blob/master/lib/extensions.js#L45

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     * SortDescriptor
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Collection.html#~SortDescriptor }
+     */
+    type SortDescriptor = string | [string, boolean] | any[];
+
+    /**
+     * Collection
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Collection.html }
+     */
+    interface Collection<T> {
```
fealebenpae: This should be a class.

We copy a bunch of `Array` methods on our `Collection` prototype so I suggest we make our type implement [`ReadonlyArray<T>`](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L986) and of course add any methods that are missing, such as `some`, `reduce`, `filter`, etc.

Some of the definitions here are wrong - for example, the `map` callback doesn't return `void`.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+        static adminUser(adminToken: string): User;
+        static login(server: string, username: string, password: string, callback: (error: any, user: User) => void): void;
+        static loginWithProvider(server: string, provider: string, providerToken: string, callback: (error: any, user: User) => void): void;
+        static register(server: string, username: string, password: string, callback: (error: any, user: User) => void): void;
+        static registerWithProvider(server: string, options: { provider: string, providerToken: string, userInfo: any }, callback: (error: Error | null, user: User | null) => void): void;
+        logout(): void;
+        openManagementRealm(): Realm;
+    }
+
+    interface SyncConfiguration {
```
fealebenpae: Also needs
```ts
validate_ssl: boolean;
ssl_trust_certificate_path: string;
```

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+        user: User;
+        url: string;
+    }
+
+    /**
+    * Session
+    * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.Session.html }
+    */
+    interface Session {
+        readonly config: any;
```
fealebenpae: The type here should be `SyncConfiguration`.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+        readonly state: string;
```
fealebenpae: Type should be `"invalid" | "valid" | "inactive"`.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+    interface SyncConfiguration {
+        user: User;
+        url: string;
+    }
+
+    /**
+    * Session
+    * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.Session.html }
+    */
+    interface Session {
```
fealebenpae: This should be a class.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+         * @returns number
+         */
+        unshift(object: T): number;
+    }
+
+    /**
+     * Results
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Results.html }
+     */
+    type Results<T> = Collection<T>;
```
fealebenpae: This should be a class.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+         * @returns boolean
+         */
+        isValid(): boolean;
+    }
+
+    /**
+     * List
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.List.html }
+     */
+    interface List<T> extends Collection<T> {
```
fealebenpae: This should be a class.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+         * @returns void
+         */
+        removeListener(callback: () => void): void;
+    }
+
+    /**
+     * Object
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Object.html }
+     */
+    interface Object {
```
fealebenpae: This should be a class.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+        /**
+         * @returns void
+         */
+        removeAllListeners(): void;
+
+        /**
+         * @param  {()=>void} callback
+         * @returns void
+         */
+        removeListener(callback: () => void): void;
```
fealebenpae: Shouldn't the `callback` type here match the one in `addListener`?

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+        readonly state: string;
+        readonly url: string;
+        readonly user: User;
+    }
+
+    /**
+    * AuthError
+    * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.AuthError.html }
+    */
+    interface AuthError {
```
fealebenpae: This is actually a class.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     */
+    type Results<T> = Collection<T>;
+
+
+
+    /**
+     * LogLevel
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.html#~LogLevel }
+     */
+    type LogLevelType = 'error' | 'info' | 'debug';
```
fealebenpae: Is this used anywhere else than in `setLogLevel`? Does it need its own typedef?

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.ChangeEvent.html }
+     */
+    interface ChangeEvent {
+        readonly changes: { [object_type: string]: { insertions: number[], deletions: number[], modifications: number[] } };
+        readonly oldRealm: Realm;
+        readonly path: string;
+        readonly realm: Realm;
+    }
+
+    function addListener(serverURL: string, adminUser: Realm.Sync.User, regex: string, name: string, changeCallback: () => void): void;
```
fealebenpae: `changeCallback` here should take a `ChangeEvent` parameter.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+    function removeAllListeners(name?: string[]): void;
+    function removeListener(regex: string, name: string, changeCallback: () => void): void;
```
fealebenpae: Ditto.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+
+declare class Realm {
+    static defaultPath: string;
+
+    readonly path: string;
+    readonly readOnly: boolean;
+    readonly schema: Realm.ObjectSchema[];
+    readonly schemaVersion: number;
+
+    syncSession: Realm.Sync.Session;
```
fealebenpae: `Realm.Sync.Session | null` - when the realm is not synced this is null.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     * @param  {string} path
+     * @param  {any} encryptionKey?
+     * @returns number
+     */
+    static schemaVersion(path: string, encryptionKey?: any): number;
+
+    /**
+     * @param  {Realm.Configuration} config?
+     */
+    constructor(config?: Realm.Configuration);
```
fealebenpae: This has one extra overload that takes `path?: string`.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     */
+    close(): void;
+
+    /**
+     * @param  {string|Realm.ObjectType} type
+     * @param  {T&Realm.ObjectPropsType} properties
+     * @param  {boolean} update?
+     * @returns T
+     */
+    create<T>(type: string | Realm.ObjectType, properties: T & Realm.ObjectPropsType, update?: boolean): T;
```
fealebenpae: `type` can also be a constructor function.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+     * @returns void
+     */
+    deleteAll(): void;
+
+    /**
+     * @param  {string|Realm.ObjectType} type
+     * @param  {number|string} key
+     * @returns T
+     */
+    objectForPrimaryKey<T>(type: string | Realm.ObjectType, key: number | string): T | void;
```
fealebenpae: `type` can also be a constructor function.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    /**
+     * @param  {string|Realm.ObjectType} type
+     * @returns Realm
+     */
+    objects<T>(type: string | Realm.ObjectType): Realm.ObjectType & Realm.Results<T>;
```
fealebenpae: `type` can also be a constructor function.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    /**
+     * @param  {string} name
+     * @param  {()=>void} callback
+     * @returns void
+     */
+    addListener(name: string, callback: () => void): void;
```
fealebenpae: `callback` should be `(sender: Realm, eventName: string) => void`. Or rather `(sender: Realm, event: 'change') => void` because that's all we support (which means `name: string` should be `event: 'change'`, too)

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    /**
+     * @param  {string} name
+     * @param  {()=>void} callback
+     * @returns void
+     */
+    removeListener(name: string, callback: () => void): void;
```
fealebenpae: Ditto.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    /**
+     * @param  {string[]} name?
+     * @returns void
+     */
+    removeAllListeners(name?: string[]): void;
```
fealebenpae: `removeAllListeners` only takes an optional string, not a string array.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+    interface ObjectType {
+        type: ObjectClass;
+    }
+
+    /**
+     * realm configuration
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.html#~Configuration }
+     */
+    interface Configuration {
+        encryptionKey?: any;
```
fealebenpae: `encryptionKey?` should be `ArrayBuffer | ArrayBufferView` or better yet just `Int8Array`.

index.d.ts
```diff
@@ -0,0 +1,486 @@
...
+        migration?: any;
```
fealebenpae: `migration?` should be `(oldRealm: Realm, newRealm: Realm) => void`

index.d.ts
```diff
@@ -0,0 +1,440 @@
...
+        ssl_trust_certificate_path: string;
+    }
+
+    /**
+    * Session
+    * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.Session.html }
+    */
+    class Session {
+        readonly config: SyncConfiguration;
+        readonly state: 'invalid' | 'valid' | 'inactive';
```
fealebenpae: `valid` => `active`

index.d.ts
```diff
@@ -0,0 +1,440 @@
...
+
+declare class Realm {
+    static defaultPath: string;
+
+    readonly path: string;
+    readonly readOnly: boolean;
+    readonly schema: Realm.ObjectSchema[];
+    readonly schemaVersion: number;
+
+    syncSession: Realm.Sync.Session | null;
```
fealebenpae: readonly

index.d.ts
```diff
@@ -0,0 +1,440 @@
...
+     * Open a realm asynchronously with a promise. If the realm is synced, it will be fully synchronized before it is available.
+     * @param {Configuration} config 
+     */
+    static open(config: Realm.Configuration): Promise<Realm>
+    /**
+     * Open a realm asynchronously with a callback. If the realm is synced, it will be fully synchronized before it is available.
+     * @param {Configuration} config 
+     * @param {Function} callback will be called when the realm is ready.
+     */
+    static openAsync(config: Realm.Configuration, callback: ((realm: Realm) => void))
```
fealebenpae: `callback: (error: any, realm: Realm) => void`

index.d.ts
```diff
@@ -0,0 +1,440 @@
...
+
+    syncSession: Realm.Sync.Session | null;
+
+    /**
+     * Get the current schema version of the Realm at the given path.
+     * @param  {string} path
+     * @param  {any} encryptionKey?
+     * @returns number
+     */
+    static schemaVersion(path: string, encryptionKey?: any): number;
```
fealebenpae: `encryptionKey?: ArrayBuffer | ArrayBufferView`

index.d.ts
```diff
@@ -0,0 +1,440 @@
...
+     * @param  {number|string} key
+     * @returns T
+     */
+    objectForPrimaryKey<T>(type: string | Realm.ObjectType | Function, key: number | string): T | void;
+
+    /**
+     * @param  {string|Realm.ObjectType|Function} type
+     * @returns Realm
+     */
+    objects<T>(type: string | Realm.ObjectType | Function): Realm.ObjectType & Realm.Results<T>;
```
fealebenpae: Should the return type be just `Result<T>`?

alazier:
Can we get this merged and included in the next release @fealebenpae I saw you mention we are doing another one soon? Would be helpful for what I am working on as well as any future ts projects.

fealebenpae:
@alazier we can merge this as soon as all the comments are addressed, so it would be helpful if you could also go through the PR and maybe fix some of the issues I've pointed out if you have the time.

PR #1083, opened 2017-06-19
Use an uncached realm instance for _waitForDownload

### LINKED ISSUE #1061: Realm.open returns realm with inactive sync session
Not sure if https://github.com/realm/realm-js/blob/master/lib/extensions.js#L53 is about the same issue but `Realm.openAsync` seems unreliable.

I have a reproduction case for this, please ping me if you need more info.

```Javascript
var Realm = require('realm');

var adminToken = 'SECRET_TOKEN';
var realmURL = 'SOME_SPECIFIC_URL';

Realm.openAsync({
  sync: {
    user: Realm.Sync.User.adminUser(adminToken),
    url: realmURL,
  },
}, (error, realm) => {
  if (error) {
    console.log(error);
  } else {
    console.log(realm);
  }
});
```

When try to open `SOME_SPECIFIC_URL` `Realm.openAsync` either doesn't execute callback at all or executes the callback and immediately closes realm (even if I keep a strong reference on it).

Using 1.3.1.

fealebenpae:
Fixes #1061

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -550,6 +550,7 @@ void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType, 
     ValueType sync_config_value = Object::get_property(ctx, config_object, "sync");
     if (!Value::is_undefined(ctx, sync_config_value)) {
         realm::Realm::Config config;
+        config.cache = false;
         static const String encryption_key_string = "encryptionKey";
         ValueType encryption_key_value = Object::get_property(ctx, config_object, encryption_key_string);
         if (!Value::is_undefined(ctx, encryption_key_value)) {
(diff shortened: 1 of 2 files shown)


src/js_realm.hpp
```diff
@@ -590,7 +591,7 @@ void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType,
             if (user && user->state() != SyncUser::State::Error) {
                 if (auto session = user->session_for_on_disk_path(config.path)) {
                     session->wait_for_download_completion([=](std::error_code error_code) {
-                        realm->config(); //capture and keep realm instance for till here
+                        realm->close(); //capture and keep realm instance for till here
```
kristiandupont: I am not sure what happens here but is the comment correct?

fealebenpae: ah, I might get it because of my Bulgarian background, but basically the idea is to keep the realm alive up to that point and then dispose it.

blagoev: it might be mine. feel free to remove it :)

PR #1094, opened 2017-06-24
Added method to find index of object in collections

astigsen:
This method allows you to find the index of an object within a collection, which is otherwise impossible as js apparently does not allow us to overload the equality operator (so there is no way to use `findIndex()` for this, as you would have to be able to compare two instances of the same object).


ianpward:
This method is needed for 
https://gist.github.com/astigsen/971b75d22d22eeed0961caa16aef994b

We should get this merged so that we can use this as a debugging tool in the future.

ohenepee:
What happened to this pull and what's the replacement to find the index of an object?

Issue #343, opened 2016-03-22
Support accessing/querying backlinks


alazier:
We will probably want to support virtual properties the same as cocoa.

joshuapinter:
Any timeline on this, Alazier? Would be great to have in there. Thanks!

joshuapinter:
@alazier Also, I'm storing the `parentId` as a property of a child object and then manually looking up the `parent` object from that `parentId` value. 

Is there a better way to do this? I've seen some existence of linking in other documentation, like [Swift](https://realm.io/docs/swift/latest/#inverse-relationships), but not sure if Realm JS supports this. Can you shed some light on this and maybe a best practice that we can use in the meantime.

Thanks!

alazier:
Linking works on the other bindings but we haven't added the feature to JS
yet as we would like to do it right. Don't have a timeline but we can bump
up the priority. Most of the underlying functionality is built and just
needs to be exposed so hopefully this won't take too long to build.

astigsen:
>  I'm storing the parentId as a property of a child object and then manually looking up the parent object from that parentId value.

While implicit backlinks are not supported yet, you can still do direct linking with object properties. That will at least remove the need for parentId's and manual lookups:
https://realm.io/docs/react-native/latest/#object-properties

joshuapinter:
@astigsen Right, so instead of using `parentId` I could just use `parent` and then link to the `Parent` object? I imagine this doesn't work when creating a `Parent` record along with the nested `Children` record all in one write, like so: 

```js
realm.write(() => {
  realm.create('Person', {
    name: 'Joe',
    car: {make: 'Honda', model: 'Accord', drive: 'awd', person: ...},
  });
});
```

You would have to create the `Person` object first and then link to it, like this: 

```js
realm.write(() => {
  let person = realm.create('Person', {
    name: 'Joe',
    car: {make: 'Honda', model: 'Accord', drive: 'awd'}
  });

  person.car.person = person;
});
```

Does that look right?

appden:
@joshuapinter That looks right to me.

joshuapinter:
Thanks @appden! Would be good to throw this in the docs as well.

martigasco:
@astigsen I'm thinking your suggested solution would take a heavy penalty at creation time, especially in mass insertion scenarios (like syncing with a remote API), wouldn't it?

[...]

Issue #1096, opened 2017-06-26
Realm.List inherited functions seem missing in jest test


peteroid:
### Summary
When I was testing my model under Realm using Jest, it seems some functions were missing. For instance, `map()` and `keys()` would be missing and the test will throw the error saying `TypeError: xxx.map is not a function`. However, this error only happens in the jest but not in the normal execution.

After some trials, I found that replacing the map function from `xxx.map(fn)` to `Realm.List.prototype.map.call(xxx, fn)` could be working in both the jest test and normal execution.

### Expected results
`map` function should be defined in the List object

### Actual results
Jest threw errors like `TypeError: xxx.map is not a function`

### Version
* realm: 1.3.1
* node: 7.10.0
* jest: 20.0.4

John251314:
I met the same problem. what is the solution, in addition to the above method. Thanks.

benweidig:
We ran into this exact issue today, but we needed some(...) which isn't 
callable like that.

kneth:
See also https://github.com/facebook/jest/issues/4603

kneth:
@kraenhansen You have a hypothesis why it isn't working, right?

kraenhansen:
> @kraenhansen You have a hypothesis why it isn't working, right?

I believed we have a similar issue with `for(let ... of ...) loop` on React Native. Originating from the way Realm JS is looping over the methods to export here: https://github.com/realm/realm-js/blob/master/lib/collection-methods.js#L49. I have a suspicion that some environments call the callback of `forEach` asynchronously.

But ... reading over the Jest issues, that you @kneth linked to, this does indeed seem like a known Jest bug. Perhaps a workaround could be found, but it seems more valuable to get the root cause fixed.

takameyer:
@peteroid  We took some time to try and reproduce this issue and failed to reproduce it.  Can you try updating Jest and Realm, to see if the issue is resolved?

takameyer:
The issues has been closed, as we are not able to reproduce it and no one in the community has reported that it is still an issue.  If someone is seeing this issue, feel free to yell and we will reopen the issue.

PR #1101, opened 2017-06-27
Add support for accessing linking objects / backlinks

mrackwitz:
Coming from the Cocoa SDKs, I was missing backlinks when working with the JS SDK, so I dove in my spare time a little into the codebase and wanted to see how it could be done.

This is related to #343.

/c @kristiandupont @bmunkholm


ianpward:
Excellent!

src/js_schema.hpp
```diff
@@ -177,14 +187,27 @@ ObjectSchema Schema<T>::parse_object_schema(ContextType ctx, ObjectType object_s
         for (uint32_t i = 0; i < length; i++) {
             ObjectType property_object = Object::validated_get_object(ctx, properties_object, i);
             std::string property_name = Object::validated_get_string(ctx, property_object, name_string);
-            object_schema.persisted_properties.emplace_back(parse_property(ctx, property_object, property_name, object_defaults));
+            Property property = parse_property(ctx, property_object, property_name, object_defaults);
+            if (property.type == realm::PropertyType::LinkingObjects) {
+                object_schema.computed_properties.emplace_back(property);
```
fealebenpae: We should `std::move` the property.

tests/js/linkingobjects-tests.js
```diff
@@ -0,0 +1,85 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
fealebenpae: Should be `2017`

fealebenpae:
There seem to be some test failures in the Chrome debugger. Could be related to #1096. I'll investigate.

I'll also add API docs.

astigsen:
In js it is super usual to open the realms without specifying a schema (since it can be inferred from the file itself), so LinkingProperties should also work dynamically without a predefined schema.

Issue #794, opened 2017-01-11
Progress notification


kristiandupont:

kristiandupont:
Docs for the .net api: https://realm.io/docs/xamarin/latest/#progress-notifications

agersoncgps:
Do we now have access to progress notifications?

blagoev:
It got merged, so next release will have it.
PR here https://github.com/realm/realm-js/pull/1279

PR #1105, opened 2017-06-27
Add progress notifications

blagoev:
Support download progress notifications for Realm open and openAsync methods

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -586,7 +586,7 @@ void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType, 
         auto realm = realm::Realm::get_shared_realm(config);
         if (auto sync_config = config.sync_config)
         {
-            static const String progressFuncName = "_onProgress";
+            static const String progressFuncName = "_onDownloadProgress";
             bool progressFuncDefined = false;
             if (!Value::is_boolean(ctx, sync_config_value) && !Value::is_undefined(ctx, sync_config_value))
             {
@@ -615,11 +615,8 @@ void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType, 
             if (user && user->state() != SyncUser::State::Error) {
                 if (auto session = user->session_for_on_disk_path(config.path)) {
                     if (progressFuncDefined) {
-                        session->register_progress_notifier([=](uint64_t transferred_bytes, uint64_t transferrable_bytes) {
-                               progressFunc(transferred_bytes, transferrable_bytes);
-                        }, SyncSession::NotifierType::download, false);
-                    }
-                    
+                        session->register_progress_notifier(std::move(progressFunc), SyncSession::NotifierType::download, false);
+                    } 
                     
                     session->wait_for_download_completion([=](std::error_code error_code) {
                         realm->close(); //capture and keep realm instance for until here
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -159,6 +159,8 @@ public:
     using ConstructorMap = typename Schema<T>::ConstructorMap;
     
     using WaitHandler = void(std::error_code);
+    using ProgressHandler = void(uint64_t transferred_bytes, uint64_t transferrable_bytes);
+
 
     static FunctionType create_constructor(ContextType);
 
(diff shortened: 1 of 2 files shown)


blagoev:
In short method _onProgress can be defined in the sync config object which will be called for download progress notifications. This allows open and openAsync to work.

tests/js/session-tests.js
```diff
@@ -237,6 +237,114 @@ module.exports = {
...
+            return promisifiedLogin('http://localhost:9080', username, 'password').then(user => {
+                const accessTokenRefreshed = this;
+                let successCounter = 0;
+                let progressNotificationCalled = false;
+                let config = {
+                    sync: { 
+                        user, 
+                        url: `realm://localhost:9080/~/${realmName}`,
+                        _onProgress: (transferred, remaining) => { 
+                            progressNotificationCalled = true
```
kristiandupont: Maybe we could also assert that `transferred + remaining === expectedSize` if we know what that is?

tests/js/session-tests.js
```diff
@@ -237,6 +237,114 @@ module.exports = {
...
+    testProgressNotificationsForRealmOpen() {
+        if (!isNodeProccess) {
+            return Promise.resolve();
+        }
+
+        const username = uuid();
+        const realmName = uuid();
+        const expectedObjectsCount = 3;
+
+        let tmpDir = tmp.dirSync();
```
kristiandupont: Seems like it's about time we abstract all of this into a `runInSeparateProcess('download-realm-api.js')` helper call or something like that?

src/js_realm.hpp
```diff
@@ -579,13 +581,48 @@ void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType,
...
 
+        std::function<ProgressHandler> progressFunc; 
+
         auto realm = realm::Realm::get_shared_realm(config);
-        if (auto sync_config = config.sync_config) {
+        if (auto sync_config = config.sync_config)
+        {
+            static const String progressFuncName = "_onProgress";
+            bool progressFuncDefined = false;
+            if (!Value::is_boolean(ctx, sync_config_value) && !Value::is_undefined(ctx, sync_config_value))
```
blagoev: I took this check from populate sync config
https://github.com/realm/realm-js/blob/master/src/js_sync.hpp#L373-L375

It seems we support sync config as boolean to force sync on history

tests/js/session-tests.js
blagoev: Indeed. will refactor it

tests/js/session-tests.js
blagoev: you mean to have third parameter with the expectedSize calculated to ease the user. I can add this. On our test the method is called always once with size 166. I guess it is synced at once if the realm is small

kristiandupont: No, I mean the test must know what the expected size is and I guess the sum of the parameters will match that. I don't know if it's deterministic when the notification is triggered so we can assert on the individual values, but this would at least be something

src/js_realm.hpp
kristiandupont: Hm ok.

tests/js/session-tests.js
blagoev: got it. will try to see how deterministic are the values on our test and will assert on them as well

src/js_realm.hpp
```diff
@@ -579,13 +581,48 @@ void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType,
...
+
+                    progressFunc = std::move(progress_handler);
+                }
+            }
+
             std::shared_ptr<SyncUser> user = sync_config->user;
             if (user && user->state() != SyncUser::State::Error) {
                 if (auto session = user->session_for_on_disk_path(config.path)) {
+                    if (progressFuncDefined) {
+                        session->register_progress_notifier([=](uint64_t transferred_bytes, uint64_t transferrable_bytes) {
```
fealebenpae: You can just `std::move(progressFunc)` instead of creating a lambda that captures a copy of the `std::func` you already have.

blagoev: good point. changing it

blagoev: @fealebenpae Do you think we need the _ before the name. or to name it just onDownloadProgress

fealebenpae: Depends on whether we want this to be public API. It's fine for internal use (it was originally requested by the Browser), but I don't think it should be public, so the underscore is good IMO.

blagoev: ok.  the latest changes include the underscore

hth190393:
This not working now, i tried but proccess download notification not show. Help please!

kristiandupont:
@hth190393 this is only an internal API. The public API is not yet exposed, you can follow it through #794.

PR #1111, opened 2017-06-30
Add a new optional server parameter to User.adminUser

### LINKED ISSUE #963: Opening management realm throws
I'd like to open the management realm from a Node.js script on the server like it's described in the JS docs here: https://realm.io/docs/swift/latest/#access-control

The expected result would be: the script will not crash
The result is: the script crashes

The code to reproduce:
```js
var Realm = require('realm');
const token = require('fs').readFileSync('admin_token.base64').toString()

var admin = Realm.Sync.User.adminUser(token);
admin.openManagementRealm();
```

The crash output:

```
node crash.js
Error: Unexpected user auth url:
    at User.openManagementRealm (.../realm-object-server-1.4.1/node_modules/realm/public/lib/user-methods.js:213:23)
    at Object.<anonymous> (.../realm-object-server-1.4.1/crash.js:5:7)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:418:7)
    at startup (bootstrap_node.js:139:9)
```

fealebenpae:
This enables `User.openManagementRealm()` for admin token users.

Fixes #963

### DIFF
--- a/src/js_sync.hpp
+++ b/src/js_sync.hpp
@@ -21,6 +21,7 @@
 #include <list>
 #include <map>
 #include <set>
+#include <regex>
 
 #include "event_loop_dispatcher.hpp"
 #include "platform.hpp"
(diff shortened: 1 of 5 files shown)


src/js_sync.hpp
```diff
@@ -404,7 +405,11 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
         }
 
         std::string raw_realm_url = Object::validated_get_string(ctx, sync_config_object, "url");
-
+        if (shared_user->token_type() == SyncUser::TokenType::Admin) {
+            static std::regex tilde("/~/");
+            raw_realm_url = std::regex_replace(raw_realm_url, tilde, "/__auth/");
```
mrackwitz: Unless the identity is not deserialized by object store when it is an admin token, we could rely here on the identity info in `shared_user` instead of hard-coding `__auth`.

fealebenpae: The identity is always a guid (because that’s what we provide on the JS side). I was originally thinking about using `__auth` as the identity but then there would be conflicts if you try to use an admin token user from two or more servers.

PR #1113, opened 2017-07-03
Add support for getting user id

blagoev:
For #1090


lib/browser/user.js
```diff
@@ -35,7 +35,8 @@ export default class User {
 
 createMethods(User.prototype, objectTypes.USER, [
     'logout',
-    '_sessionForOnDiskPath'
+    '_sessionForOnDiskPath',
+    'retrieveAccount'
```
fealebenpae: This is only needed for methods implemented in C++.

blagoev:
@fealebenpae  can you please review the last changes.

lib/user-methods.js
```diff
@@ -229,6 +229,33 @@ module.exports = {
...
             });
-        }
-    }
+        },
+        retrieveAccount(provider, provider_id) {
+            checkTypes(arguments, ['string', 'string']);
+
+            let url = url_parse(this.server);
+            url.set('pathname', `/api/providers/${provider}/accounts/${provider_id}`);
+            let headers = {
```
kristiandupont: `const`

lib/user-methods.js
```diff
@@ -229,6 +229,33 @@ module.exports = {
...
+                Authorization: this.token
+            };
+
+            const options = {
+                method: 'GET',
+                headers: headers,
```
kristiandupont: Could just be `headers`

tests/js/admin-user-helper.js
```diff
@@ -0,0 +1,88 @@
...
+                        url: `realm://localhost:9080/__admin`,
+                        error: err =>
+                         console.log('Error opening __admin realm ' + err.user  + ' ' + err.url + ' ' + err.state),
+                    }
+                };
+
+                Realm.open(config).then(realm => {
+                    if (userIdentity == admin_token_user.identity)
+                    {
+                        console.log(".");
```
kristiandupont: What's the purpose of this?

tests/js/admin-user-helper.js
```diff
@@ -0,0 +1,88 @@
...
+                        }
+
+                        Realm.Sync.User.login('http://localhost:9080', newAdminName, password, (error, newAdminUser) => {
+                            if (error) {
+                                reject(error);
+                            } else {
+                                let isAdmin = newAdminUser.isAdmin;
+                                user.logout();
+                                if (!isAdmin) {
+                                    setTimeout(_ => {
```
kristiandupont: Wouldn't `() => {...}` be better?

tests/js/admin-user-helper.js
```diff
@@ -0,0 +1,88 @@
...
+                                        waitForServerToUpdateAdminUser();
+                                    }, 500);
+                                    return;
+                                }
+
+                                resolve({ 
+                                    username: newAdminName, 
+                                    password: password
```
kristiandupont: Could just be `password`

tests/js/index.js
```diff
@@ -41,10 +41,11 @@ if (Realm.Sync) {
     TESTS.SessionTests = require('./session-tests');
 }
 
-function node_require(module) { return require(module); }
+function node_require(module) { return require(module); }
 
 // If on node, run the async tests
-if (typeof process === 'object' && process + '' === '[object process]') {
+global.isNodeProcess = typeof process === 'object' && process + '' === '[object process]'
```
kristiandupont: I think this runs in the browser as well, when running tests with chrome debugging? In which case, I guess `global` is undefined.

tests/js/admin-user-helper.js
```diff
@@ -0,0 +1,88 @@
...
+                    }
+                };
+
+                Realm.open(config).then(realm => {
+                    if (userIdentity == admin_token_user.identity)
+                    {
+                        console.log(".");
+                    }
+
+                    let pendingAdminUser = realm.objects('User').filtered(`id == "${userIdentity}"`)[0];
```
fealebenpae: Is `id` a primary key? We could potentially use `objectForPrimaryKey` then.

tests/js/admin-user-helper.js
fealebenpae: What's this?

tests/js/index.js
```diff
@@ -73,6 +74,22 @@ exports.registerTests = function(tests) {
...
 };
 
+exports.prepare = function(done) {
+    if (!global.isNodeProcess || global.testAdminUserInfo) {
+        done();
+    }
+
+    let helper = require('./admin-user-helper');
+    helper.createAdminUser().then(userInfo => {
+        global.testAdminUserInfo = userInfo;
```
fealebenpae: What if we have the tests call `createAdminUser()` directly to always get a fresh admin user?

tests/js/admin-user-helper.js
blagoev: left over. forgot to remove

tests/js/admin-user-helper.js
blagoev: using it as a shorthand for fast typing. indeed _ variable is not needed

tests/js/index.js
blagoev: indeed. will refactor

tests/js/admin-user-helper.js
blagoev: left over + checking anyone pays attention :). removing it right away

tests/js/admin-user-helper.js
blagoev: just copied cocoa logic. will try with objectForPrimaryKey

tests/js/index.js
blagoev: I guess the idea is to use only the global admin user. which is always fresh for every test run. I think cocoa reuses this user throughout their tests

fealebenpae: I see. Let's leave it as it is and revisit if we need to later on.

tests/js/admin-user-helper.js
```diff
@@ -65,15 +58,15 @@ exports.createAdminUser = function () {
                                 let isAdmin = newAdminUser.isAdmin;
                                 user.logout();
                                 if (!isAdmin) {
-                                    setTimeout(_ => {
+                                    setTimeout(() => {
```
fealebenpae: I think this could be just `setTimeout(waitForServerToUpdateAdminUser, 500)`

fealebenpae:
Oh, I totally forgot - could you also please make a new PR with a changelog entry and an update to the API docs?

blagoev:
sure thing. i forgot as well thanks for reminding.

blagoev:
@fealebenpae made this one https://github.com/realm/realm-js/pull/1127

PR #1135, opened 2017-07-11
Rethrow callback errors as fatal Node.js errors

fealebenpae:
Instead of rethrowing the JavaScript error emitted from a callback function, we should instead pass it to `node::FatalException`.

Callbacks are employed when there is no other JavaScript frame on the stack prior so rethrowing the JS error as a C++ exception is not going to propagate the error back to JavaScript. `node::FatalException` will raise the `uncaughtError` event on the `process` object , print the error and stacktrace, and alert the debugger is there is one attached. This would make our async callbacks behave consistently with Node’s own async callbacks such as `setTimeout` when encountering an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+vNext Release notes (TBD)
+=============================================================
+### Breaking changes
+* None
+
+### Enhancements
+* None
+
+### Bug fixes
+* Fix crash on Node.js when a listener callback throws an error.
+  The error will now be forwarded to Node's fatal error handling facilities. This means better error reporting,
+  the ability to debug such errors in a Node.js debugger, and proper invocation of the `uncaughtError` event on the `process` object.
+
 1.9.0 Release notes (2017-7-10)
 =============================================================
 ### Breaking changes
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -49,9 +49,13 @@ module.exports = function(realmConstructor) {
                         reject(error);
                     }
 
-                    let syncedRealm = new realmConstructor(config);
-                    //FIXME: RN hangs here. Remove when node's makeCallback alternative is implemented
-                    setTimeout(() => { resolve(syncedRealm); }, 1);                     
+                    try {
+                        let syncedRealm = new this(config);
+                        //FIXME: RN hangs here. Remove when node's makeCallback alternative is implemented
+                        setTimeout(() => { resolve(syncedRealm); }, 1);
+                    } catch (e) {
+                        reject(e);
+                    }
                 });
             });
         },
@@ -62,9 +66,13 @@ module.exports = function(realmConstructor) {
                         callback(error);
                     }
 
-                    let syncedRealm = new realmConstructor(config);
-                    //FIXME: RN hangs here. Remove when node's makeCallback alternative is implemented
-                    setTimeout(() => { callback(null, syncedRealm); }, 1);
+                    try {
+                        let syncedRealm = new this(config);
+                        //FIXME: RN hangs here. Remove when node's makeCallback alternative is implemented
+                        setTimeout(() => { callback(null, syncedRealm); }, 1);
+                    } catch (e) {
+                        callback(e);
+                    }
                 });
         },
     }));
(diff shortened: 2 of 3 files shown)


blagoev:
Looks good with one note:

For the event subscription cases which use the callback method this is ok.
For the cases where a Promise is returned this might be problematic. I think the only such case right now is Realm.open()
If the callback throws a JS exception I think the promise should be rejected instead redirecting to node::FatalException. Either we fix Realm.open to catch all JS exceptions thus preventing unhandled exceptions or we will need to use the previous logic of throwing c++ exceptions, catching them and rejecting the promises. Handling all js exceptions should be easy. The other way requires capturing the promise instance.

What do you think?

fealebenpae:
> For the cases where a Promise is returned this might be problematic.

Welp, let's try to break this down. Suppose we have the following JavaScript and C++ code:

```js
function foo() {
    _fooNative(function callbackCalledInCpp(error) {
      if (error) {
        reject(error);
      } else {
        resolve();
      }
    });
  });
}

function fooAsync(callback) {
    _fooNative(function callbackCalledInCpp(error) {
        callback(error);
    });
}
```

```c++
void foo(ContextType ctx, FunctionType, ObjectType, size_t argc, const ValueType arguments[], ReturnValue &return_value)
{
    Protected<GlobalContextType> protected_ctx(Context<T>::get_global_context(ctx));
    Protected<FunctionType> protected_callback(ctx, Value::validated_to_function(ctx, arguments[0]);

    EventLoopDispatcher<void(std::error_code)> dispatcher([protected_ctx, protected_callback](std::error_code error) {
        HANDLESCOPE;
        if (error) {
            ObjectType error_object = Object::create_empty(protected_ctx);
            // populate error properties

            ValueType callback_arguments[1];
            callback_arguments[0] = error_object;
            Function<T>::callback(protected_ctx, protected_callback, ObjectType(), 1, callback_arguments);
        } else {
            Function<T>::callback(protected_ctx, protected_callback, ObjectType(), 0, nullptr);
        }
    });

    SomeApi::foo(std::move(dispatcher));
}
```

Both `foo` and `fooAsync` call into our C++ method. Notice the `callbackCalledInCpp` function in JavaScript. This is the function that `Function<T>::callback` will execute. What happens when this function throws an error?

Well, for the Promise-returning `foo` we should make sure that `callbackCalledInCpp` never throws. `resolve` and `reject` are guaranteed never to throw with the way Promises are implemented so we're safe there.

> If the callback throws a JS exception I think the promise should be rejected instead redirecting to node::FatalException

Note that the callback here is the function that resolves or reject a promise. If this callback throws, there's no way for us to do anything to the promise.

However, `fooAsync`'s `callbackCalledInCpp` directly calls a user callback. If the user callback throws then we need to properly communicate that error back to the user, which I think the switch to `node::FatalError` helps with.

blagoev:
I am referring to this line. If it happens to throw an JS exception we will call the fatal error handler instead of rejecting the promise. Or maybe I am wrong about that.

fealebenpae:
So it turns out that in quite a few places where we use `Function<T>::callback` there are cases where we already have a JavaScript stack to execute on. One example is `Realm.addListener` which gets invoked indirectly by `Realm.write` when advancing the read version as part of the write transaction. https://github.com/realm/realm-js/pull/1135/commits/0e6ba0cd752e407fcf64ddadd79c8bceffe67019 identifies an optimization where `Function<T>::callback` can fall back to `Function<T>::call`. We actually had tests that expected this to be the case and were failing with the `node::FatalException` change. Now `FatalException` will only be called for truly asynchronous callbacks.

To be honest, it bothers me that the same callback on `Realm.addListener` can throw an error synchronously if using `Realm.write` or asynchronously with `FatalException` if the change is driven by the `WeakRealmNotifier` for example. I suppose we would want to always asynchronously invoke listener callbacks (maybe with `v8::Isolate::EnqueueMicrotask`), but that would be a breaking change in behavior we should consider for 2.0.

fealebenpae:
> I am referring to this line. If it happens to throw an JS exception we will call the fatal error handler instead of rejecting the promise.

I'm sorry, which line?

blagoev:
i forgot to copy it the first time https://github.com/realm/realm-js/blob/master/lib/extensions.js#L52

fealebenpae:
Ah, I see now. Thanks, that's a very good catch! Actually @kvap told me over Slack just yesterday that errors throws by opening the Realm (incompatible schema changes, in his case) do not reject the promise but are instead logged as unhandled rejections by Node.js. We should have surrounded that line with try/catch to begin with. I'll add one now.

We should really make sure we don't throw in callbacks that resolve promises.

PR #1138, opened 2017-07-12
Add `Realm.empty`

fealebenpae:
## What, How & Why?

Add a new property on `Realm`'s prototype that indicates whether a realm contains any objects or not.

/cc @mbalex99 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ vNext Release notes (TBD)
 * None
 
 ### Enhancements
-* None
+* Added `Realm.prototype.empty` which is a property that indicates whether or not the realm has any objects in it.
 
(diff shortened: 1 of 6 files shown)


bmunkholm:
In other SDK's it's called `isEmpty, so unless there is a JS idiomatic expectation that it must be called `empty`please rename to `isEmpty`.

fealebenpae:
@bmunkholm We don't use `is` or `are` prefixes for boolean properties in JavaScript, only for methods that return booleans.

bmunkholm:
Ahh, sorry, didn't notice it was a property :-)

PR #1195, opened 2017-08-09
Fix Object.setPrototypeOf issue on Android

kristiandupont:
## What, How & Why?

`Object.setPrototypeOf` doesn't exist in JSC on Android, causing some tests to crash.
This workaround fixes that.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/tests/js/schemas.js
+++ b/tests/js/schemas.js
@@ -44,8 +44,15 @@ PersonObject.prototype.description = function() {
 PersonObject.prototype.toString = function() {
     return this.name;
 };
-Object.setPrototypeOf(PersonObject, Realm.Object);
-Object.setPrototypeOf(PersonObject.prototype, Realm.Object.prototype);
+
+// Object.setPrototypeOf doesn't work on JSC on Android. The code below achieves the same thing.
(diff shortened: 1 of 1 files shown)


tests/js/schemas.js
```diff
@@ -44,8 +44,15 @@ PersonObject.prototype.description = function() {
...
     return this.name;
 };
-Object.setPrototypeOf(PersonObject, Realm.Object);
-Object.setPrototypeOf(PersonObject.prototype, Realm.Object.prototype);
+
+// Object.setPrototypeOf doesn't work on JSC on Android. The code below achieves the same thing.
+//Object.setPrototypeOf(PersonObject, Realm.Object);
+//Object.setPrototypeOf(PersonObject.prototype, Realm.Object.prototype);
+
+PersonObject.__proto__ = Realm.Object.__proto__;
```
fealebenpae: The right-hand part shouldn't use `__proto__`.
Should be just `PersonObject.__proto__ = Realm.Object`.

kristiandupont: Of course! I'll fix that quickly

PR #1196, opened 2017-08-09
Running js tests in Electron renderer process

kraenhansen:
## What, How & Why?

This introduces tests of `realm-js` in an Electron environment.

This PR introduces a `tests/electron` directory with:
- An electron app in `tests/electron/app` which has
  - `jasmine.js` that imports the jasmin lib, setup a console logger and exports an execute function.
  - `main.js` which starts a hidden `BrowserWindow` and either runs the tests itself (see --main flag below) or lets
    the render process do the heavy lifting.
  - `renderer.js` detects if it's supposed to run the tests and does that using the `jasmine.js`.
- `spec.js` in which imports and executes the tests exported by `tests/js/index.js`.
- A `test/electron/runner.js` script, which uses [spectron](https://www.npmjs.com/package/spectron) to start the Electron app and read out the console from the Electron process, and console logging it.

## Flags

To use these flags, you need to prepend them when calling `npm test` after the `--`, which indicates that the flag is
not ment for npm.

### Process

You can specify in which Electron process to run the tests:
- `--process=main` for the main process or
- `--process=render` the render process (which is default)

As an example, this runs all tests in the main process:

    npm test -- --process=main

### Filter

If you want to run only a subset of the tests, use the `--filter` flag, ex:

As an example, this runs only the suite named "UserTests":

    npm test -- --filter=UserTests

## Failing tests

These tests are failing at the moment:
- SessionTests (because REALM_MODULE_PATH is missing, due to `tests/spec/helpers` not loading correctly.
- AsyncTests (because of the same reason as SessionTests)
- GarbageCollectionTests (due to https://github.com/realm/realm-js/issues/1197)

The text above has been added to a [README.md](https://github.com/realm/realm-js/blob/4d59a377b47ff9aa60b1b2af39f20a1b1e788811/tests/electron/README.md) as well.

## ☑️ ToDos
* [ ] Fix the SessionTests
* [ ] Fix the AsyncTests
* [x] Verify that the latest version on master fixes the GarbageCollectionTests
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/scripts/download-object-server.sh
+++ b/scripts/download-object-server.sh
@@ -20,5 +20,7 @@ mkdir object-server-for-testing
 tar -C object-server-for-testing -xf "$object_server_bundle"
 rm "$object_server_bundle"
 
-echo -e "enterprise:\n  skip_setup: true\n" >> "object-server-for-testing/object-server/configuration.yml"
+echo "enterprise:\n  skip_setup: true\n" >> "object-server-for-testing/object-server/configuration.yml"
+# Change to a "warn" level
+sed -i -- "s/# level: 'info'/level: 'warn'/g" object-server-for-testing/object-server/configuration.yml
 touch "object-server-for-testing/object-server/do_not_open_browser"
--- a/tests/js/garbage-collection.js
+++ b/tests/js/garbage-collection.js
@@ -1,0 +1,62 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+'use strict';
+
+const Realm = require('realm');
+const TestCase = require('./asserts');
+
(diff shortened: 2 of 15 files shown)


scripts/download-object-server.sh
```diff
@@ -20,5 +20,7 @@ mkdir object-server-for-testing
 tar -C object-server-for-testing -xf "$object_server_bundle"
 rm "$object_server_bundle"
 
-echo -e "enterprise:\n  skip_setup: true\n" >> "object-server-for-testing/object-server/configuration.yml"
-touch "object-server-for-testing/object-server/do_not_open_browser"
\ No newline at end of file
+echo "enterprise:\n  skip_setup: true\n" >> "object-server-for-testing/object-server/configuration.yml"
+# Change to a "warn" level
+sed -i -- "s/# level: 'info'/level: 'warn'/g" object-server-for-testing/object-server/configuration.yml
+touch "object-server-for-testing/object-server/do_not_open_browser"
```
kristiandupont: Nice!

tests/js/garbage-collection.js
```diff
@@ -0,0 +1,62 @@
...
+'use strict';
+
+const Realm = require('realm');
+const TestCase = require('./asserts');
+
+const NUMBER_OF_OBJECTS = 1000;
+const BUFFER_LENGTH = 1024;
+const READ_CYCLES = 10;
+
+module.exports = {
```
kristiandupont: Perhaps it would make sense with a brief comment about why we have this? Including a link to the issue in the Electron repo

kraenhansen: :+1:

kraenhansen:
It seems that the `GarbageCollectionTests` are currently timing-out on when run in the react native context https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-js/detail/PR-1196/7/pipeline/47#step-265-log-207.
I suspect this is due to the debugging making the execution very slow.

PR #1209, opened 2017-08-14
Add shouldCompactOnLaunch option to configuration

### LINKED ISSUE #507: Expose compact() method on Realm instances
We can call it something else in JS, but essentially it would just call `SharedRealm::compact()`. This would alleviate concerns for app developers over ever-growing Realm file sizes.

This issue was brought up in #504.

kneth:
## What, How & Why?

Expose a way to compact (remove unused space) Realms.

Closes #507.

## ☑️ ToDos
* [x] jsdoc files updated

Question: should we also expose an unconditional compaction method?


src/js_realm.hpp
```diff
@@ -390,6 +393,28 @@ void RealmClass<T>::constructor(ContextType ctx, ObjectType this_object, size_t
...
+                }
+
+                FunctionType should_compact_on_launch_function = Value::validated_to_function(ctx, compact_value, "shouldCompactOnLaunch");
+                config.should_compact_on_launch_function = [=](uint64_t total_bytes, uint64_t unused_bytes) {
+                    ValueType arguments[2] = {
+                        Value::from_number(ctx, total_bytes),
+                        Value::from_number(ctx, unused_bytes)
+                    };
+
+                    ValueType should_compact = Function<T>::call(ctx, should_compact_on_launch_function, 2, arguments);
```
kristiandupont: I think this should be `callback` instead of `call`.

tests/js/realm-tests.js
```diff
@@ -18,6 +18,7 @@
 
 'use strict';
 
+var fs = require('fs');
```
kristiandupont: This won't work in React Native. Maybe use `react-native-fs`

src/js_realm.hpp
```diff
@@ -176,6 +176,8 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
     static void remove_listener(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void remove_all_listeners(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void close(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
+    static void compact(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
```
kristiandupont: This was never implemented?

src/js_realm.hpp
```diff
@@ -390,6 +393,28 @@ void RealmClass<T>::constructor(ContextType ctx, ObjectType this_object, size_t
                 config.schema_version = 0;
             }
 
+            static const String compact_on_launch_string = "shouldCompactOnLaunch";
+            ValueType compact_value = Object::get_property(ctx, object, compact_on_launch_string);
```
kristiandupont: I am not sure if this will work automatically with the chrome debugger proxies, try running the tests with 
```
scripts/test.sh react-tests Debug
```

src/js_realm.hpp
kneth: It is related to my question. Let us not expose a `Realm.compact()` method.

kristiandupont: Ah right. Yeah, I think the config option is fine for now.

nirinchev:
Personally, I vote for exposing the `compact` function :)

kristiandupont:
@nirinchev is there a specific reason for this or is it just a vague preference?

nirinchev:
On one hand, it's easier to write a command line tool that imports data into a Realm file, then calls compact to make sure it's as small as possible, before, say, bundling it in the app. Obviously, this can be done via the browser, or by opening the realm a second time with a `shouldCompactOnLaunch` block, but calling compact would be much more straightforward.

Second, I see this as a way to provide app developers with the tool to force deterministic compaction. They understand their apps much better than we do, so there may be natural places in the app where forced compaction may make sense. In a way this is analogous to GC vs Dispose - both make sense.

Third, .NET and Java have explicit `compact` API, so exposing it in js would reduce the differences between bindings.

Finally, I can't see a compelling reason not to expose it - it's a method and a unit test. If people don't want to use it, fine, doesn't cost us anything in terms of maintenance. If they want to use it, fine again, it's there 😃 What cons do you see in exposing the API?

tests/js/realm-tests.js
```diff
@@ -970,8 +968,7 @@ module.exports = {
         const realm2 = new Realm({schema: [schemas.StringOnly], shouldCompactOnLaunch: shouldCompact});
         TestCase.assertTrue(wasCalled);
         TestCase.assertEqual(realm2.objects('StringOnlyObject').length, count + 2);
-        var afterSize = fs.statSync(realm2.path).size;
-        TestCase.assertTrue(beforeSize >= afterSize);
+        // we don't check if the file is smaller as we assume that Object Store does it
```
nirinchev: How are we going to validate that `compact` was actually called?

kneth: Here we only check that the callback given to Object Store is actually called.IMHO, we don't have to test if Object Store is doing the right thing - that's should be done by Object Store's tests.

nirinchev: Yeah, it's just a nitpick - we don't have a test that validates that the result (`true`) that we set in the js callback is what's getting delivered to the OS callback. I'm sure it is, but would be nice to test it :) In any case, if it's hard, I guess we can skip it and assume it works.

kneth:
@nirinchev @kristiandupont I have been through your comments. `Realm.compact()` has been added.

misterlib:
@kristiandupont any idea when this will get pushed?

kristiandupont:
@misterlib it didn't make it into yesterdays release, sorry. We will make another release soon but it won't happen this week.

misterlib:
@kristiandupont thanks for the info. i think we're going to fork it and try it out. on a time crunch and already late. but appreciate you getting back to me!

jasonmerino:
@kneth @kristiandupont Do you guys have an estimated date when the next version of the JS SDK is going to be released? Would love to be able to take advantage of compacting.

kneth:
@jasonmerino No, but I hope we can do one later this week.

PR #1210, opened 2017-08-14
Depreciate setAccessToken

blagoev:
rename setAccessToken to setFeatureToken
expose both functions while setAccessToken is depreciated
Trim token argument string

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -114,7 +114,7 @@ module.exports = function(realmConstructor) {
                 realmConstructor.Sync._setFeatureToken(featureToken.trim());
             }
 
-            //enable depreciated setAccessToken
+            //enable deprecated setAccessToken
             realmConstructor.Sync.setAccessToken = realmConstructor.Sync.setFeatureToken;
         }
     }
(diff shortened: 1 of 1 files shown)


lib/extensions.js
```diff
@@ -104,6 +104,19 @@ module.exports = function(realmConstructor) {
...
+        if (realmConstructor.Sync._setFeatureToken) {
+            realmConstructor.Sync.setFeatureToken = function(featureToken) {
+                if (typeof featureToken !== 'string' || !featureToken instanceof String) {
+                    throw new Error("featureToken should be a string");
+                }
+
+                realmConstructor.Sync._setFeatureToken(featureToken.trim());
+            }
+
+            //enable depreciated setAccessToken
```
kneth: My personal taste: "depreciated" -> "deprecated"

bdash: "Depreciated" is simply the incorrect term to use in this context. "Deprecated" is correct.

blagoev: will fix 10x

blagoev: done. 
clearly I don't know how to write since my branches are named like that as well :)

Issue #1188, opened 2017-08-05
Realm is already in a write transaction inside live collection listener!?


sheshavpd:
Hello,
I'm using the latest realm js lib for my react native app. Thanks for the great library.
When` realm.write` emits a notification, my collection listener is catching it as expected.
I'm changing some values of other collection inside listener again.
But, in one of my Android device which is super fast(OnePlus 5) there's no error.
But in my other device( LETV le2), which is considerably slower than Oneplus, the error `Realm is already in a write transaction triggers up`. Am I missing something?

kneth:
@sheshavpd Thanks for reporting it. It is a bit hard to say anything definitely without seeing the code. It is possible for you to share it (or part of it)?

sheshavpd:
@kneth, hello, let me explain this as clearly as I can.
As #927 suggests, there's a bug where the notification for current transaction is triggered when the next realm transaction is called.
I've a `collection of users` and `list of messages` affiliated to that user.
When the message is received from internet, I need to marked it as `read` when user is online and has seen it.
So this is what I'm doing right now:
```
function onMessageReceived(userId,message)
{
   let user=realm.objects("User").filtered("id==$0",userId);
   realm.write(()=>{
       user.messages.push(message);
  });
   //#927 bug's workaround
   realm.write(()=>{});
}
```
and in the below snippet, I've attached a listener to messages of user in a component:
`let messages=realm.objects("User").filtered("id==$0",userId)[0].messages;
messages.addListenet(function(messages,changes){
   //Some code to push message to listview here on new insertions
   .......
  //End of UI tweeks
  //Here I showed the message and should mark it as delivered
   changes.insertions.forEach(function(index){
      if(userInThisScreen)
      {
        //Here is where I'm getting the error
        realm.write(()=>{
         messages[index].seen_time=Date.now();
        });
        //#927 bug's workaround
        realm.write(()=>{});
      }
   });
});
`
So, this code snippet is not producing error for OnePlus5, which is apparently very fast. But when it's run(fresh installation) on Le2, it says realm is already in a write transaction in the line I've shown above.

This happens to me whenever I need to change the Realm data inside listeners. Here's what I use as workaround:

`let messages=realm.objects("User").filtered("id==$0",userId)[0].messages;
messages.addListenet(function(messages,changes){
   //Some code to push message to listview here on new insertions
   .......
  //End of UI tweeks
  //Here I showed the message and should mark it as delivered
   changes.insertions.forEach(function(index){
[...]

Issue #815, opened 2017-01-26
Manual transactions


astigsen:
Add methods for manual transactions: `startTransaction()`, `commitTransaction()` and `rollbackTransaction()`.

These methods are needed when you have transactions that spans multiple callbacks that you cannot wrap in a single closure.

kneth:
Similar to #715.

kneth:
Fixed by #1216.

Issue #715, opened 2016-12-03
Exposing begin_transaction & close_transaction


marcus433:
By having explicit control over opening and closing transactions it can be possible to run asynchronous code inside a transaction. 

e.g. currently the below code would close the transaction before the coroutine completes as it isn't synchronous.
```
realm.write(() => {
  Promise.coroutine(() => {
    var somedata = yield x();
    realm.create('objname', somedata);
  })();
})
```

However, with explicit control of begin / close we could still have synchronous transactions while allowing for asynchronous functions to run.
```
Promise.coroutine(() => {
  realm.begin_transaction();
  var somedata = yield x();
  realm.create('objname', somedata);
  realm.close_transaction();
})();
```

Of course the above could also be solved by fetching the data outside of the transaction and placing the write inside the coroutine. However, situations can exist where the yielded function also depends on the transaction being ran. E.g. `function x()` may run `realm.create` itself rather than return data.

Side note: Wonderful job Realm Team!

kristiandupont:
This is an interesting idea. Thank you for the feedback and kind words!

kneth:
I have merged #1216 which implements manual transactions.

Udbhav12:
How to use begin transaction and close transaction? Where can I find it's documentation?

I am using realm 2.26.1

kneth:
@Udbhav12 

You can use [`beginTransaction`](https://realm.io/docs/javascript/latest/api/Realm.html#beginTransaction), [`cancelTransaction`](https://realm.io/docs/javascript/latest/api/Realm.html#cancelTransaction), and [`commitTransaction`](https://realm.io/docs/javascript/latest/api/Realm.html#commitTransaction).

PR #1216, opened 2017-08-15
Adding methods and property to manually control write transactions

kneth:
## What, How & Why?

Write transactions are currently only done through the `Realm.write()` function. Sometimes it is useful to be able to control the write transactions manually. This PR adds a set of methods.

See also #715, #815, and #1188.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


lib/index.d.ts
```diff
@@ -452,6 +453,21 @@ declare class Realm {
      * @returns void
      */
     write(callback: () => void): void;
+
+    /**
+     * @returns void
```
kristiandupont: We should add the comments from the documentation .js file here as well -- hopefully at some point, this file can be the single source of documentation both for the website and for typescript-aware ide's

kneth: Isn't it better do redo/unite the documentation in a separate PR?

kristiandupont: If this applies to other methods than just this one, then yes. I didn't check for that. :-)

kneth:
@kristiandupont @blagoev I have been through the PR feedback, and I believe it is ready to be merged.

PR #1221, opened 2017-08-16
Update object store

blagoev:
updated object-store

related commit https://github.com/realm/realm-object-store/commit/22420458794e0eabf032fd0e406e28fd06157f41#diff-99a0771b4ad4157fdbcf45dfdc8f1e24R151


src/js_types.hpp
```diff
@@ -351,11 +351,12 @@ REALM_JS_INLINE void set_internal(const typename T::Object &object, typename Cla
...
-{
-    if (prop.is_nullable && (is_null(context, value) || is_undefined(context, value))) {
+{ 
+    if (is_nullable(prop.type) && (is_null(context, value) || is_undefined(context, value))) {
         return true;
     }
 
+    //lubo
     using PropertyType = realm::PropertyType;
     switch (prop.type) {
```
bdash: `PropertyType::Array` and `PropertyType::LinkingObjects` should be removed from this switch, and instead tested for with `is_array(prop.type)` then testing for `PropertyType::Object` and `PropertyType::LinkingObjects` respectively.

bdash: And the switch statement would need to be on `property.type & ~PropertyType::Flags` instead, to remove `PropertyType::Nullable` from any nullable options.

src/js_sync.hpp
```diff
@@ -119,11 +119,14 @@ void UserClass<T>::is_admin(ContextType ctx, ObjectType object, ReturnValue &ret
 template<typename T>
 void UserClass<T>::create_user(ContextType ctx, FunctionType, ObjectType this_object, size_t argc, const ValueType arguments[], ReturnValue &return_value) {
     validate_argument_count(argc, 3, 5);
+    SyncUserIdentifier userIdentifier {
+        (std::string)Value::validated_to_string(ctx, arguments[1], "identity"),
```
kneth: Why do you need to cast?
And why use this C like cast?

src/js_types.hpp
blagoev: I tried with property.type & ~PropertyType::Flags and the compiler warning didn't go away. Will change it like that anyways.

//lubo comment is left over - will remove

src/js_sync.hpp
blagoev: copied from here https://github.com/realm/realm-js/pull/1221/files/09c0390220dd881053dd93b86e5375acf8ef5324#diff-59d314428ce563ce4238ac4e1295e609L125

the new version of get_user made the compiler pick the wrong overload. 

probably not needed for SyncUserIdentifier

src/js_types.hpp
blagoev: also thanks for picking up this for review. I was doing the changes based on the changes in object store. I noticed the is_array change here https://github.com/realm/realm-object-store/commit/22420458794e0eabf032fd0e406e28fd06157f41#diff-99a0771b4ad4157fdbcf45dfdc8f1e24R151

but PropertyType: Object is left inside the switch statement.

blagoev: got what you meant for  property.type & ~PropertyType::Flags

blagoev:
@kneth  fixed thanks

@bdash fixed and found another place which needed same changes. right ?
here: https://github.com/realm/realm-js/pull/1221/commits/35e6c9a6b22c7ad4d99b6ac678c28912da9c0a9a

also not sure about it will ever get to here 
https://github.com/realm/realm-js/pull/1221/commits/88e46bc12de436ddab4d20d3659f4bc481e88532#diff-67ffcaed4b7c898e9846b21c2024a6c5R371

src/js_types.hpp
```diff
@@ -351,13 +351,28 @@ REALM_JS_INLINE void set_internal(const typename T::Object &object, typename Cla
...
 
+    if (realm::is_array(prop.type)) {
+        if (prop.type == PropertyType::Object) {
+            auto object = to_object(context, value);
+            return Object<T>::template is_instance<ResultsClass<T>>(context, object)
+                || Object<T>::template is_instance<ListClass<T>>(context, object);
+        } else if (prop.type == PropertyType::LinkingObjects) {
+            return false;
+        }
+        else if (prop.type == PropertyType::Array) {
```
bdash: I don't think this check makes sense. `is_array` checks if the `PropertyType::Array` is set, and if it is you then test against the various types to determine what the array contains (objects, linking objects, and in the future the various primitive types).

src/js_types.hpp
```diff
@@ -397,6 +397,15 @@ inline bool Value<T>::is_valid_for_property(ContextType context, const ValueType
 
 inline std::string js_type_name_for_property_type(PropertyType type)
 {
+    if (realm::is_array(type)) {
+       if (type == PropertyType::LinkingObjects) {
+            throw std::runtime_error("LinkingObjects' type is not supported");
```
bdash: Hmm, wasn't support added for linking objects in the last few months? Surely this should now be returning a valid string rather than throwing?

blagoev:
I missed that this was continued here: https://github.com/realm/realm-js/pull/1263
so closing this one

PR #1243, opened 2017-08-23
Authentication Promise or Callback Style APIs

mbalex99:
## What, How & Why?

All http authentication methods are now `callback` or `Promise` styles

```javascript
Realm.Sync.User.login(url, username, password, (err, user) => {
    // normal callback
})

// if callback is not specified a Promise<User> is returned
Realm.Sync.User.login(url, username, password)
    .then(user => {})
    .catch(err => {})

// this is great for async await style of programming
try {
     await Realm.Sync.User.login(url, username, password)
} catch (err)  {

}
```

`Realm.open` is now favored over `Realm.openAsync`. I've marked in the comments that `Realm.openAsync` will be deprecated in the future. 
```javascript
Realm.open(config)
    .then(realm => {})
    .catch(err => {})

// if you like callbacks

Realm.open(config, (err, realm) => {
     // this just like `Realm.openAsync`
})
```

Reference to the issue(s) addressed by this PR: #1182 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [x] 🚦 Tests
* [x] 📝 Public documentation PR created or is not necessary
* [x] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [x] typescript definitions file is updated
* [x] jsdoc files updated
* [ ] Chrome debug API is updated if API is available on React Native

PR #1249, opened 2017-08-27
adding promisified apis

mbalex99:
Continuation of #1243 but this time history is off of master branch as @fealebenpae


blagoev:
@mbalex99 Can we mark the callback ones as deprecated before merging this.

blagoev:
I see that two people approved this but it has some bugs in the code. I will try fixing this and then post a newer PR with the changes. (a new PR since the merges with master were not correct and and broken.)

blagoev:
closing in favor of #1302

PR #1263, opened 2017-08-29
Update to latest objectstore master and core/sync

tgoyne:
Pulls in the changes from https://github.com/realm/realm-js/pull/1221 plus further changes for the latest objectstore.


tgoyne:
Remaining CI failures are due to needing a release with https://github.com/realm/realm-core/pull/2823 and objectstore master requiring vs2017.

blagoev:
@tgoyne  Can you define what is needed to finish this since we have only this week. Maybe we can discuss the vs 2017 issue

fealebenpae:
I updated our windows build nodes to the latest npm and node-gyp and now the build picks up Visual Studio 2017. However, there are still build errors.

blagoev:
it seems related to the unreleased BinaryData change here https://github.com/realm/realm-core/pull/2823

src/js_schema.hpp
```diff
@@ -76,56 +76,59 @@ Property Schema<T>::parse_property(ContextType ctx, ValueType attributes, std::s
...
     Property prop;
     prop.name = property_name;
-    
+
     ObjectType property_object = {};
     std::string type;
-    
+
+    using realm::PropertyType;
+    PropertyType is_optional = PropertyType::Required;
```
kneth: Maybe `is_optional` is not the best name. Reading "optional is equal to required" was too hard for me :smile: On the other hand, `modifier` is too generic.

src/js_types.hpp
```diff
@@ -80,18 +81,16 @@ struct Context {
...
-        m_prefix(std::move(prefix)),
-        m_type(std::move(type)) 
-        {}
-
-private:
-    std::string m_prefix;
-    std::string m_type;
+    TypeErrorException(StringData object_type, StringData property,
+                       std::string const& type, std::string const& value)
+    : std::invalid_argument(util::format("%1.%2 must be of type '%3', got (%4)",
```
kneth: Why not "'%4'"?

src/js_types.hpp
```diff
@@ -352,12 +350,31 @@ REALM_JS_INLINE void set_internal(const typename T::Object &object, typename Cla
...
 
-    using PropertyType = realm::PropertyType;
-    switch (prop.type) {
+    using realm::PropertyType;
+    if (realm::is_array(prop.type)) {
+        if (prop.type != PropertyType::Object) {
+            return false;
+        }
+
+        // FIXME: Do we need to validate the types of the contained objects?
```
kneth: How much does core/object store check?

tgoyne: All of this validation is just to provide better error messages; none of it is semantically required.

src/js_types.hpp
tgoyne: Results in some misleading messages when the value is an array (looks too much like the value provided was just a string).

tgoyne:
Windows node tests are hitting a call to `abort()` inside the first call to `SharedGroup::do_open()`. The Windows core build doesn't have debug symbols or assertions enabled so it's sort of hard to figure out what's going wrong.

tgoyne:
Sync 2.0.0-rc18 appears to be incompatible with ROS 2.0.0-fc2, so I'm not sure how to handle testing the update.

src/js_types.hpp
kneth: Let's leave it as an exercise for future contributors :smile:

tgoyne:
Windows crashes magically went away with the new version of core, so now the only failing jobs are the ones which need a ROS download (which is currently a 404).

bdash:
I think a new ROS requires getting a new release of realm-js that uses the updated version of sync, so it may not be possible to test the sync aspects of this right now.

blagoev:
I plan to update the downloading and starting of the new ROS to switch to npm for 2.0 and above. Will update this PR with that

bdash:
@blagoev, how does that help? There's no ROS release that uses a new enough version of sync at this point. Updating to use ROS from npm will be necessary when a new ROS release is available, but one inherently cannot be available until after this PR is merged and a realm-js release is cut (unless I'm completely misunderstanding how things work?).

blagoev:
i got too optimistic. I think we will need npm download in any case, but maybe we will need to release without these tests running. Maybe an internal release will do it both for realm-js and then ros, so realm-js can be tested with it afterwards

kneth:
As @bdash points out, this PR depends on a release of ROS with this PR included. I have tried to do a rc release based on this PR but failed.

kneth:
Yesterday I released 2.0.0-rc6 based on this PR (+ `master`).

tgoyne:
The ROS download for rc6 doesn't appear to exist.

bdash:
You can find the [latest ROS release on their releases page](https://github.com/realm/ros/releases). Note that the new ROS uses different packaging, so quite likely requires changes to the test harness before it can be used in tests.

blagoev:
we have an issue for that here https://github.com/realm/realm-js-private/issues/326

blagoev:
https://github.com/realm/realm-js-private/issues/326#issuecomment-328976487

kneth:
I guess we need to update once https://github.com/realm/realm-object-store/pull/546 is merged.

tgoyne:
I've updated this to just skip running the sync tests for now so that it can be merged without waiting for https://github.com/realm/realm-js/pull/1322 to be ready.

blagoev:
actually this is already merged in 2.0 branch. the #1322 is ready but we have a problem building for iOS from 2.0 branch so that takes high priority than #1322

PR #1274, opened 2017-08-31
Don't throw an exception every 10ms in the RPC worker

tgoyne:
Cuts the time taken to run the react native tests on my machine from 198 seconds to 41 seconds. Probably has a similar impact on the react native debugger performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+X.Y.Z Release notes
+=============================================================
+### Breaking changes
+* None
+
+### Enhancements
+* Improve performance of the RPC worker for chrome debugging.
+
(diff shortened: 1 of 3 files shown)


blagoev:
need to add it to changes.md as well

kneth:
@tgoyne I am not sure why CI doesn't pass. The offending line is `TestCase.assertEqual(permissions[1].path, `/${user1.identity}/test`);` - https://github.com/realm/realm-js/blob/master/tests/js/permission-tests.js#L64

tgoyne:
That appears to be just a pre-existing flakey test. I got the same failure on master after rerunning the tests a few times and it passed on CI after a rerun.

PR #1279, opened 2017-09-01
Progress api

blagoev:
Fixes #1279

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -332,9 +332,10 @@ declare namespace Realm.Sync {
         ssl_trust_certificate_path?: string;
     }
 
-    type ProgressNotificationCallback = (transferred: number, total: number) => void;
+    type ProgressNotificationCallback = (transferred: number, transferable: number) => void;
     type ProgressDirection = 'download' | 'upload';
     type ProgressMode = 'reportIndefinitely' | 'forCurrentlyOutstandingWork';
+    
     /**
     * Session
     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.Session.html }
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -332,6 +332,9 @@ declare namespace Realm.Sync {
         ssl_trust_certificate_path?: string;
     }
 
+    type ProgressNotificationCallback = (transferred: number, total: number) => void;
+    type ProgressDirection = 'download' | 'upload';
+    type ProgressMode = 'reportIndefinitely' | 'forCurrentlyOutstandingWork';
     /**
     * Session
     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.Session.html }
@@ -341,6 +344,9 @@ declare namespace Realm.Sync {
         readonly state: 'invalid' | 'active' | 'inactive';
         readonly url: string;
         readonly user: User;
+
+        addProgressNotification(direction: ProgressDirection, mode: ProgressMode, progressCallback: ProgressNotificationCallback): void;
+        removeProgressNotification(progressCallback: ProgressNotificationCallback): void;
     }
 
     /**
@@ -402,6 +408,11 @@ declare namespace Realm.Sync {
     }
 }
 
+
+interface ProgressPromise  extends Promise<Realm> {
+    progress(callback: Realm.Sync.ProgressNotificationCallback) : Promise<Realm>
+}
+
 declare class Realm {
     static defaultPath: string;
 
@@ -422,17 +433,20 @@ declare class Realm {
      */
     static schemaVersion(path: string, encryptionKey?: ArrayBuffer | ArrayBufferView): number;
 
+    
+
     /**
      * Open a realm asynchronously with a promise. If the realm is synced, it will be fully synchronized before it is available.
      * @param {Configuration} config 
      */
-    static open(config: Realm.Configuration): Promise<Realm>
+    static open(config: Realm.Configuration): ProgressPromise;
     /**
      * Open a realm asynchronously with a callback. If the realm is synced, it will be fully synchronized before it is available.
      * @param {Configuration} config 
+     * @param {ProgressNotificationCallback} progressCallback? a progress notification callback for 'download' direction and 'forCurrentlyOutstandingWork' mode 
      * @param {Function} callback will be called when the realm is ready.
      */
-    static openAsync(config: Realm.Configuration, callback: (error: any, realm: Realm) => void): void
+    static openAsync(config: Realm.Configuration, progressCallback?: Realm.Sync.ProgressNotificationCallback, callback: (error: any, realm: Realm) => void): void
 
     /**
      * Delete the Realm file for the given configuration.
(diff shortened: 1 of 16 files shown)


blagoev:
adding for review without the docs, so we can skip updating documentation if something is changed. Use the typescript definition for quick reference.
documentation is still in progress.

blagoev:
For reviewers,  pay attention to how Realm.open and Realm.openAsync work. Is this a good way to do it or we could support all progress modes/directions in there.

lib/index.d.ts
```diff
@@ -411,17 +422,20 @@ declare class Realm {
...
-    static open(config: Realm.Configuration): Promise<Realm>
+    static open(config: Realm.Configuration): ProgressPromise;
     /**
      * Open a realm asynchronously with a callback. If the realm is synced, it will be fully synchronized before it is available.
      * @param {Configuration} config 
+     * @param {ProgressNotificationCallback} progressCallback? a progress notification callback for 'download' direction and 'forCurrentlyOutstandingWork' mode 
      * @param {Function} callback will be called when the realm is ready.
      */
-    static openAsync(config: Realm.Configuration, callback: (error: any, realm: Realm) => void): void
+    static openAsync(config: Realm.Configuration, progressCallback?: Realm.Sync.ProgressNotificationCallback, callback: (error: any, realm: Realm) => void): void
```
kneth: Would be better to have the optional argument as the last argument?

src/js_realm.hpp
```diff
@@ -586,9 +586,14 @@ void RealmClass<T>::get_sync_session(ContextType ctx, ObjectType object, ReturnV
 
 template<typename T>
 void RealmClass<T>::wait_for_download_completion(ContextType ctx, FunctionType, ObjectType this_object, size_t argc, const ValueType arguments[], ReturnValue &return_value) {
-    validate_argument_count(argc, 2);
+    validate_argument_count(argc, 2, 3);
     auto config_object = Value::validated_to_object(ctx, arguments[0]);
-    auto callback_function = Value::validated_to_function(ctx, arguments[1]);
+    auto callback_function = Value::validated_to_function(ctx, arguments[argc - 1]);
```
kneth: Would be better to have the optional argument as the last argument?

src/js_sync.hpp
```diff
@@ -170,6 +171,10 @@ class SessionClass : public ClassDefinition<T, WeakSession> {
 
     static void simulate_error(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void refresh_access_token(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
+#if REALM_ENABLE_SYNC
```
kneth: `src/js_sync.hpp` is only included if `REALM_ENABLE_SYNC` is defined so I think we don't need this.

tests/js/download-api-helper.js
```diff
@@ -2,33 +2,49 @@
...
+}
 
+Realm.Sync.User.register('http://localhost:9080', username, 'password', (error, registeredUser) => {
+    if (error) {
+        const registrationError = JSON.stringify(error);
+        Realm.Sync.User.login('http://localhost:9080', username, 'password', (err, loggedUser) => {
+            if (err) {
+                const loginError = JSON.stringify(err);
+                console.error("download-api-helper failed:\n User.register() error:\n" + registrationError + "\n User.login() error:\n" + loginError);
+                process.exit(-2);
```
kneth: Maybe throw an exception. Termination does look a bit too rough to me.

blagoev: this file is run out of process so it's on purpose

lib/index.d.ts
blagoev: well the callback is the last as per nodejs convention. So this is stuffed before that. Also the js code allows to call the function with openAsync(config, callback) as before. Maybe I could change this to a default initialized parameter

src/js_sync.hpp
blagoev: will remove

lib/index.d.ts
kneth: My question has clearly non-JS roots :smile: Let's leave it as it is.

blagoev:
docs https://github.com/realm/realm.io/pull/3793

src/js_sync.hpp
```diff
@@ -308,7 +313,83 @@ void SessionClass<T>::refresh_access_token(ContextType ctx, FunctionType, Object
...
+        Protected<typename T::GlobalContext> protected_ctx(Context<T>::get_global_context(ctx));
+        std::function<ProgressHandler> progressFunc; 
+
+        EventLoopDispatcher<ProgressHandler> progress_handler([=](uint64_t transferred_bytes, uint64_t transferrable_bytes) {
+            HANDLESCOPE
+            ValueType callback_arguments[2];
+            callback_arguments[0] = Value::from_number(protected_ctx, transferred_bytes);
+            callback_arguments[1] = Value::from_number(protected_ctx, transferrable_bytes);
+
+            Function<T>::callback(protected_ctx, protected_callback, protected_this, 2, callback_arguments);
```
fealebenpae: This way the `this` of the callback is the session, but if the callback is an arrow function that's sort of useless. Wouldn't it be better to pass it as an argument instead?

src/js_sync.hpp
```diff
@@ -308,7 +313,83 @@ void SessionClass<T>::refresh_access_token(ContextType ctx, FunctionType, Object
...
+        });
+
+        progressFunc = std::move(progress_handler);
+
+        
+        auto registrationToken = session->register_progress_notifier(std::move(progressFunc), notifierType, false);
+
+        auto syncSession = create_object<T, SessionClass<T>>(ctx, new WeakSession(session));
+        PropertyAttributes attributes = ReadOnly | DontEnum | DontDelete;
+        Object::set_property(ctx, callback_function, "syncSession", syncSession, attributes);
```
fealebenpae: I know the properties are non-enumerable (though I think they would still show in a debugger), but I suggest prepending the names with an underscore. I'm not sure what your thoughts on the matter are, but I treat underscored member names in JavaScript as a convention to demonstrate that something is private.

src/js_sync.hpp
blagoev: which `this` should be better used here for the callback when called. (from user perspective)?

src/js_sync.hpp
blagoev: sure will fix

src/js_sync.hpp
fealebenpae: I think not specifying any `this` at this level would be best.

blagoev: 👍

kneth:
The failing test is:

```
PermissionTests : testApplyAndGetGrantedPermissions
   TypeError: Cannot read property 'path' of undefined
       at user.applyPermissions.then.then.permissions (/Users/realm/workspace/realm_realm-js_PR-1279-V5WM5CA2KFXWHKK7MY3PF3THQSTPR7NRC3MKNYVZK7FID7SDH6KQ/tests/js/permission-tests.js:64:50)
```

blagoev:
yes indeed. the reason is permission tests are failing sporadically. will do a rebuild. we will need to look at that if this happens too much.

Issue #573, opened 2016-08-20
Migration doesn't remove undeclared schemas


victoriafrench:
Not sure if this is a bug (though I feel it should be) and it's driving me crazy.

So I have my v1 schema that has let's say ['Dog', 'Cat']
Then I have v2 schema (with migration) lets say ['Cat', 'Mouse]

Migration works great... but the database now contains Dog, Cat, and Mouse. I would think it would now only be Cat and Mouse as I am no longer declaring Dog as part of the schema. But the table is still there.

I see in the iOS, Java, and Swift kit there is the capability to delete a table during migration. But none of those examples work for react-native. So how is this done or when will it be available.

Yes I did scour the internet for the past two days trying to find an answer and 'delete your database' is not acceptable as an answer.

Thanks.

victoriafrench:
So here is a real test scenario with interesting results:

```javascript
import Realm from 'realm';

const KeyCode_v1 = {
  name: 'KeyCode',
  primaryKey: 'code',
  properties: {
    code: 'string',
    formid: { type: 'string', indexed: true }
  }
};

const FormInfo_v1 = {
  name: 'FormInfo',
  primaryKey: 'id',
  properties: {
    id: 'string',
    deleted: 'bool',
    locked: 'bool',
    name: 'string',
    keycode: 'string'
  }
};

const KeyCodes = {
  name: 'KeyCodes',
  primaryKey: 'keycode',
  properties: {
    keycode: 'string',
    name: 'string',
    formsec: 'string',
    locked: 'bool',
    deleted: 'bool'
  }
};

function migrateV2(oldRealm, newRealm) {
  if (oldRealm.schemaVersion < 2) {
    var oldFormInfo = oldRealm.objects('FormInfo');
    var newKeyCodes = newRealm.objects('KeyCodes');
    for (let i = 0; i < oldFormInfo.length; i++) {
      let wf = oldFormInfo[i];
      let nf = {
        keycode: wf.keycode,
        name: wf.name,
        formsec: wf.id,
        locked: wf.locked || false,
        deleted: wf.deleted || false
      };
      let index = newRealm.objects('KeyCodes').filtered(`keycode == "${nf.keycode}"`);
      if ( index.length === 0 ) {
        newRealm.create('KeyCodes', nf);
      } else if (nf.locked === true && index[0].locked === false) {
        // record the new item only if it is locked and the other one isn't
        newRealm.delete(index);
        newRealm.create('KeyCodes', nf);
      }
    }
  }
[...]

PR #1290, opened 2017-09-11
Add Realm.deleteModel API (#573)

ashwinphatak:
## What, How & Why?

Reference to the issue(s) addressed by this PR: #573

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


ashwinphatak:
Added a test case to see how this works and it turns out that:

1. Deleting a model which is a target of linkingObjects results in an exception (from object-store)
2. Deleting a model which isn't a target of linkingObjects works fine

If this is the expected behaviour, I guess no changes are required.

kneth:
@ashwinphatak I will try to restart testing (it seems to be related to iOS simulator).

PR #1303, opened 2017-09-12
Use the objectstore keypath sorting functionality

### LINKED ISSUE #437: Sorting by sub objects
Assume I have the following schema:

```
var Car = {
   id: 'int',
   make: 'string',
   model: 'string',
};

var Person = {
   id: 'int',
   first_name: 'string',
   last_name: 'string',
};

var Passenger = {
   id: 'int',
   person: {type: 'Person'},
   car: {type: 'Car'},
};
```

If I do the following, I get the error Property 'person.last_name' does not exist on object type 'Passenger'.

`realm.objects('Passenger').sorted('person.last_name')`

Is my syntax incorrect, or is sorting by sub-objects not supported?  If it's not supported, are there plans to support it?

tgoyne:
Adds support for sorting on values over links, reduces the amount of code in the binding, and is required for sorting non-object lists.

Fixes #437.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+vNext (TBD)
+=============================================================
+### Breaking changes
+* None
+
+### Enhancements
+* Add support for sorting Lists and Results on values from linked objects.
+
+### Bug fixes
+* None
+
 2.0.0 Release notes (2017-9-19)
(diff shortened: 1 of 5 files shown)


kneth:
CI failures are due to missing ROS.

PR #1315, opened 2017-09-14
Check the exception message in all Realm tests which assert an exception is thrown

tgoyne:
And fix some particularly bad error messages.

I started adding tests to verify that the error messages when creating invalid primitive arrays were sensible and found that most of the code I had to modify was fairly unrelated to primitive arrays, so I went ahead and split these changes off.


tgoyne:
react-test failures are because the chrome debugging code has different error messages for a few things. I'll take care of that tomorrow.

PR #1321, opened 2017-09-18
v1 -> v2 upgrade path

kneth:
## What, How & Why?

Reference to the issue(s) addressed by this PR: https://github.com/realm/realm-js-private/issues/323

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

**Waiting for enabling ROS testing**: #1338


tests/js/session-tests.js
```diff
@@ -461,6 +461,25 @@ module.exports = {
...
+    testSyncV1() {
+        return Realm.Sync.User.register('http://localhost:9080', uuid(), 'password').then(user => {
+            return new Promise((resolve, _reject) => {
+                const config = { path: 'sync-1.x.realm', sync: { user, url: 'realm:://localhost:9080/~/sync-1.x' } };
+                try {
+                    const realm = new Realm(config);
+                }
+                catch (e) {
+                    if (e.match('/IncompatibleSyncedRealm: /')) {
+                        const backupConfig = { config: e.substring('IncompatibleSyncedRealm: '.length), sync: { user, url: 'realm:://localhost:9080/~/sync-1.x' } };
```
bdash: Uhh, why did we switch to parsing paths out of an error message and assembling a configuration by hand? That seems incredibly ugly and error-prone. Why are we not throwing an object that exposes a pre-built configuration?

Note that the configuration that's being assembled here isn't correct: including a sync configuration will guarantee that it'll fail to open in exactly the same way that it did originally.

kneth: Parsing strings is ugly but I don't think we can throw anything but strings (see for example https://github.com/realm/realm-js/blob/master/src/node/node_class.hpp#L309). My initial through was to throw a configuration object.

tests/js/session-tests.js
```diff
@@ -461,6 +461,25 @@ module.exports = {
         });
     },
 
+    testSyncV1() {
```
bmunkholm: Perhaps this in not the norm for the tests, but I would recommend a comment on what this test does, so I don't have to check the code.
"SyncV1" doesn't tell me enough.

bmunkholm: If you want to expose that in the naming: "testThrowsWhenOpeningIncompatibleRealms" might do.

kneth:
I ask for suggestion where and how to document it.

tests/js/session-tests.js
```diff
@@ -461,6 +461,25 @@ module.exports = {
...
+    testIncompatibleSyncedRealm() {
+        return Realm.Sync.User.register('http://localhost:9080', uuid(), 'password').then(user => {
+            return new Promise((resolve, _reject) => {
+                const config = { path: 'sync-1.x.realm', sync: { user, url: 'realm:://localhost:9080/~/sync-1.x' } };
+                try {
+                    const realm = new Realm(config);
+                }
+                catch (e) {
+                    if (e.match('/IncompatibleSyncedRealm: /')) {
+                        const backupConfig = { path: e.substring('IncompatibleSyncedRealm: '.length) };
```
bdash: Doesn't this configuration need to specify that the Realm should be opened as read-only? If you don't do that you should get an incompatible histories exception from opening a synced Realm as a writeable non-synced Realm.

bdash:
I strongly feel this needs to throw an object that exposes a configuration rather than forcing users to parse a path out of an error message.

See `v8::Isolate::ThrowException` for how to throw an arbitrary JS object using V8. In JavaScriptCore I believe it's simply a matter of assigning an object to the relevant exception output parameter.

docs/realm.js
```diff
@@ -88,24 +88,34 @@ class Realm {
...
      * @returns {ProgressPromise} - a promise that will be resolved with the realm instance when it's available.
      */
     static open(config) {}
 
     /**
+<<<<<<< HEAD
+     * Open a realm asynchronously with a callback. If the realm is synced, it will be fully
+=======
      * Open a Realm asynchronously with a callback. If the Realm is synced, it will be fully
+>>>>>>> 590f5845af4d8fd0bd82f280de4ec72cbcbba044
```
bdash: Looks like conflict resolution went awry in a few places.

blagoev:
async test is failing looking into this

blagoev:
It seems the RN debug target fails cause the rpc does not handle the exception correctly. Will continue to investigate

blagoev:
@kneth can I get a review

lib/browser/rpc.js
```diff
@@ -226,9 +243,21 @@ function sendRequest(command, data, host = sessionHost) {
         let error = response && response.error;
 
         // Remove the type prefix from the error message (e.g. "Error: ").
-        if (error) {
+        if (error && error.replace) {
             error = error.replace(/^[a-z]+: /i, '');
         }
+        else if (error.type && error.type == 'dict') {
```
kneth: `===`?

PR #1332, opened 2017-09-21
Fixes for latest react-native / metro-bundler

tqc:
The react native packager does not support dynamic require calls (facebook/react-native#6391, facebook/metro-bundler#52). Previous versions would only fail if the dynamic require was actually called, but the latest version detects the invalid syntax during build.

This update adds a replacement entry point for react-native that does not include the node-specific code.

### DIFF
--- a/lib/react-native.js
+++ b/lib/react-native.js
@@ -18,36 +18,12 @@
 
 'use strict';
 
-// Prevent React Native packager from seeing modules required with this
-function nodeRequire(module) {
-    return require(module);
-}
-
-/* global fetch */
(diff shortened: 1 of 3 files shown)


bmunkholm:
Hey @tqx!
Welcome and thanks a lot for the contribution!!!
I assume you already signed the CLA from CONTRIBUTING.md?

ashwinphatak:
Fixed in https://github.com/realm/realm-js/commit/e1b70c2f573e51dfbd6b8b42e3c8d8851744bb6c

blagoev:
closing this one since there is a fix for it. We could revisit how this is done in the future

PR #1337, opened 2017-09-22
The sync config object gets a new property called ssl_verify_callback.

morten-krogh:
The callback is used to let JS code decide whether the server's SSL
certificate should be accepted or rejected. The ssl_verify_callback
function is only applicable for platforms using OpenSSL.

This callback makes it possible to use SSL with arbitrary servers. 
It does not address the reason that some certificates were rejected by OpenSSL.
The callback can be used in those circumstances.

The callback has been tested manually. 
There are no automatic tests, since there does not seem to be any SSL tests in the repo
at the moment. Maybe this can be addressed later.

## What, How & Why?

Reference to the issue(s) addressed by this PR: #https://github.com/realm/realm-sync/issues/1601

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+NEXT RELEASE
+=============================================================
+
+### Enhancements
+* Add a callback function used to verify SSL certificates in
+  the sync config.
+
+
 2.0.0-rc10 Release notes (2017-9-19)
 =============================================================
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@ NEXT RELEASE
 =============================================================
 
 ### Enhancements
-* Add a callback function used to verify SSL certificates in
-  the sync config.
+* Add a callback function used to verify SSL certificates in the sync config.
 
 
 2.0.0-rc10 Release notes (2017-9-19)
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -311,7 +311,9 @@ Realm.defaultPath;
  *        The return value of open_ssl_verify_callback decides whether the certificate is accepted (true)
  *        or rejected (false). The open_ssl_verify_callback function is only respected on platforms where
  *        OpenSSL is used for the sync client, e.g. Linux. The open_ssl_verify_callback function is not
- *        allowed to throw exceptions.
+ *        allowed to throw exceptions. If the operations needed to verify the certificate lead to an exception,
+ *        the exception must be caught explicitly before returning. The return value would typically be false
+ *        in case of an exception.
  *
  *        When the sync client has received the server's certificate chain, it presents every certificate in
  *        the chain to the open_ssl_verify_callback function. The depth argument specifies the position of the
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -292,41 +292,43 @@ Realm.defaultPath;
  *   child properties:
  *   - `user` - A `User` object obtained by calling `Realm.Sync.User.login`
  *   - `url` - A `string` which contains a valid Realm Sync url
- *   - `error` - A callback function which is called in error situations
+ *   - `error` - A callback function which is called in error situations.
+ *        The `error` callback can take up to four optional arguments: `message`, `isFatal`,
+ *        `category`, and `code`.
  *   - `validate_ssl` - Indicating if SSL certificates must be validated
  *   - `ssl_trust_certificate_path` - A path where to find trusted SSL certificates
- *   - `ssl_verify_callback` - A callback function used to accept or reject the server's
- *      SSL certificate. ssl_verify_callback is called with an object of type
- *      {
- *        serverAddress: String
- *        serverPort: Number
- *        pemCertificate: String
- *        preverifyOk: Number
- *        depth: Number
- *      }
+ *   - `open_ssl_verify_callback` - A callback function used to accept or reject the server's
+ *        SSL certificate. open_ssl_verify_callback is called with an object of type
+ *        <code>
+ *          {
+ *            serverAddress: String,
+ *            serverPort: Number,
+ *            pemCertificate: String,
+ *            acceptedByOpenSSL: Boolean,
+ *            depth: Number
+ *          }
+ *        </code>
+ *        The return value of open_ssl_verify_callback decides whether the certificate is accepted (true)
+ *        or rejected (false). The open_ssl_verify_callback function is only respected on platforms where
+ *        OpenSSL is used for the sync client, e.g. Linux. The open_ssl_verify_callback function is not
+ *        allowed to throw exceptions.
  *
- *      The return value of ssl_verify_callback decides whether the certificate is accepted (true)
- *      or rejected (false). The ssl_verify_callback function is only respected on platforms where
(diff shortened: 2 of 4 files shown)


docs/realm.js
```diff
@@ -288,13 +288,44 @@ Realm.defaultPath;
...
+ *        depth: Number
+ *      }
+ *
+ *      The return value of ssl_verify_callback decides whether the certificate is accepted (true)
+ *      or rejected (false). The ssl_verify_callback function is only respected on platforms where
+ *      OpenSSL is used for the sync client, e.g. Linux.
+ *
+ *      When the sync client has received the server's certificate chain, it presents every certificate in
+ *      the chain to the ssl_verify_callback function. The depth argument specifies the position of the
+ *      certificate in the chain. depth = 0 represents the actual server certificate and the root
```
kneth: "and the" -> "The"???

docs/realm.js
```diff
@@ -288,13 +288,44 @@ Realm.defaultPath;
...
+ *
+ *      preverifyOk is 1 if OpenSSL has accepted the certificate, and
+ *      0 if OpenSSL has rejected it. It is generally safe to return true when preverifyOk is 1. If
+ *      preverifyOk is 0, an independent verification should be made.
+ *
+ *      One possible way of using the ssl_verify_callback function is to embed the known server certificate
+ *      in the client and accept the presented certificate if and only if it is equal to the known
+ *      certificate.
+ *
+ *      The purpose of ssl_verify_callback is to enable custom certificate pinning and to solve cases where
```
kneth: Maybe a link to a general description of certificate pinning could help the reader.

src/js_sync.hpp
```diff
@@ -240,6 +242,98 @@ class SyncSessionErrorHandlerFunctor {
...
+        typename T::Object ssl_certificate_object = Object<T>::create_empty(ctx);
+        Object<T>::set_property(ctx, ssl_certificate_object, "serverAddress", Value<T>::from_string(ctx, server_address));
+        Object<T>::set_property(ctx, ssl_certificate_object, "serverPort", Value<T>::from_number(ctx, double(server_port)));
+        Object<T>::set_property(ctx, ssl_certificate_object, "pemCertificate", Value<T>::from_string(ctx, pem_certificate));
+        Object<T>::set_property(ctx, ssl_certificate_object, "preverifyOk", Value<T>::from_number(ctx, double(preverify_ok)));
+        Object<T>::set_property(ctx, ssl_certificate_object, "depth", Value<T>::from_number(ctx, double(depth)));
+
+        const int argc = 1;
+        typename T::Value arguments[argc] = { ssl_certificate_object };
+        typename T::Value ret_val = Function<T>::callback(ctx, this_object->m_func, typename T::Object(), 1, arguments);
```
kneth: What if the callback function throws a JS exception?

docs/realm.js
```diff
@@ -288,13 +288,44 @@ Realm.defaultPath;
...
  *   - `error` - A callback function which is called in error situations
  *   - `validate_ssl` - Indicating if SSL certificates must be validated
  *   - `ssl_trust_certificate_path` - A path where to find trusted SSL certificates
+ *   - `ssl_verify_callback` - A callback function used to accept or reject the server's
+ *      SSL certificate. ssl_verify_callback is called with an object of type
+ *      {
+ *        serverAddress: String
+ *        serverPort: Number
+ *        pemCertificate: String
+ *        preverifyOk: Number
```
blagoev: why not boolean value instead of Number. by the comments below seems like a boolean will suffice

blagoev: for the preverifyOk

docs/realm.js
```diff
@@ -288,13 +288,44 @@ Realm.defaultPath;
...
  *   changing the `schema`.
- * @property {Object} [sync] - Sync configuration parameters with the following 
+ * @property {Object} [sync] - Sync configuration parameters with the following
  *   child properties:
  *   - `user` - A `User` object obtained by calling `Realm.Sync.User.login`
  *   - `url` - A `string` which contains a valid Realm Sync url
  *   - `error` - A callback function which is called in error situations
  *   - `validate_ssl` - Indicating if SSL certificates must be validated
  *   - `ssl_trust_certificate_path` - A path where to find trusted SSL certificates
+ *   - `ssl_verify_callback` - A callback function used to accept or reject the server's
```
blagoev: why not call this open_ssl_verify_callback if this is respected only for open_ssl systems.

docs/realm.js
blagoev: also why not change this name to openSSLAccepted, acceptedbyOpenSSL,or something like mustVerify

morten-krogh: Yes, I was very close to making it a boolean. I just kept the OpenSSL API.  
I agree it is better with a boolean. 
I have changed it to boolean and called it acceptedByOpenSSL.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+NEXT RELEASE
+=============================================================
+
+### Enhancements
+* Add a callback function used to verify SSL certificates in
```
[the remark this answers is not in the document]
morten-krogh: Done

docs/realm.js
morten-krogh: Changed.

docs/realm.js
morten-krogh: I changed "pinning" to "handling". 

I think, it would be somewhat out of place to link to some random page about certificate pinning. 

I just want to convey the message that this callback can be used to handle every type of
custom situation.

Feel free to come with suggestions if this is not good enough.

src/js_sync.hpp
morten-krogh: I think it is reasonable to require that the callback doesn't throw. 
The user can always catch exceptions and return false.

I will add the requirement of no except in the description.

morten-krogh: What about the error handler? I guess it is not allowed to throw either.

docs/realm.js
morten-krogh: Good idea. I will do that.

The idea was that it might be implemented for other crypto libraries in the future, but
that is probably not going to happen soon. I will rename it.

docs/realm.js
```diff
@@ -288,13 +288,44 @@ Realm.defaultPath;
...
+ *      SSL certificate. ssl_verify_callback is called with an object of type
```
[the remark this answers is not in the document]
morten-krogh: Thanks. I hadn't checked.

I checked it and reformatted a bit. 

Take a look.

morten-krogh:
@blagoev @kneth 
I have solved or replied to all your review comments.
Please review again.

src/js_sync.hpp
kneth: I agree: error handlers shouldn't throw.

My concern with your callback is that an I/O operation might fail. But if you add something to the doc about catching exceptions and return false, my concern isn't valid :smile:.

morten-krogh: I will add a discussion about exceptions.

blagoev:
@morten-krogh seems all good

blagoev:
@morten-krogh I will delete the remote branch.

PR #1345, opened 2017-09-25
Add support for lists of things other than objects

### LINKED ISSUE #1229: Add support for Arrays of primitives
## Description

Related to realm/product#59

## TODO

* [ ] Update to new object store (waiting for https://github.com/realm/realm-object-store/pull/512)
* [ ] Add changelog entry
* [ ] Schema: translate JS schema to object store
* [ ] Creating objects
* [ ] Tests

tgoyne:
Still needs docs and such, but the code/tests are ready for review. The commits should probably be reviewed independently; all of the ones before the actual AoP commit are mostly unrelated things that caused problems for me when trying to debug failing tests.

Fixes #1229.

-----------------------------------
Todo:
- [ ] Documentation
- [ ] Changelog
- [ ] Typescript definitions

### DIFF
--- a/src/js_schema.hpp
+++ b/src/js_schema.hpp
@@ -23,6 +23,8 @@
 #include "js_types.hpp"
 #include "schema.hpp"
 
+#include "util/format.hpp"
+
 namespace realm {
 namespace js {
 
@@ -41,7 +43,7 @@ struct Schema {
     using ConstructorMap = std::map<std::string, Protected<FunctionType>>;
 
     static ObjectType dict_for_property_array(ContextType, const ObjectSchema &, ObjectType);
-    static Property parse_property(ContextType, ValueType, std::string, ObjectDefaults &);
+    static Property parse_property(ContextType, ValueType, StringData, std::string, ObjectDefaults &);
     static ObjectSchema parse_object_schema(ContextType, ObjectType, ObjectDefaultsMap &, ConstructorMap &);
     static realm::Schema parse_schema(ContextType, ObjectType, ObjectDefaultsMap &, ConstructorMap &);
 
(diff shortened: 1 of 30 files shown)


tgoyne:
React native assertion failures are just because the RPC worker needs to be updated.

blagoev:
@bmunkholm I will review now, but without the RPC worker it means RN app can't be debugged

tests/ios/RealmJSCoreTests.m
```diff
@@ -48,6 +48,8 @@ + (XCTestSuite *)defaultTestSuite {
         context[@"Promise"] = promiseModule[@"Promise"];
     }
 
+    context[@"global"] = [JSValue valueWithNewObjectInContext:context];
```
blagoev: Wonder why do we need this. I am afraid it might hurt us if 'global' should not be present in that context.

src/js_schema.hpp
```diff
@@ -68,89 +70,92 @@ typename T::Object Schema<T>::dict_for_property_array(ContextType ctx, const Obj
...
-        prop.type = PropertyType::Data | is_optional;
+        prop.type |= PropertyType::Data;
     }
     else if (type == "list") {
-        if (!Value::is_valid(property_object)) {
-            throw std::runtime_error("List property must specify 'objectType'");
-        }
-        prop.type = PropertyType::Object | PropertyType::Array;
-        prop.object_type = Object::validated_get_string(ctx, property_object, object_type_string);
+        prop.type |= PropertyType::Object | PropertyType::Array;
```
kneth: Anything that prevent `list[]?`

src/jsc/jsc_class.hpp
```diff
@@ -300,20 +300,43 @@ inline void ObjectWrap<ClassType>::get_property_names(JSContextRef ctx, JSObject
...
 }
 
+static inline bool try_get_int(JSStringRef property, int64_t& value) {
+    value = 0;
+    auto str = JSStringGetCharactersPtr(property);
+    auto end = str + JSStringGetLength(property);
+    while (str != end && iswspace(*str)) {
+        ++str;
+    }
+    bool negative = false;
```
kneth: You could use `int sign = 1`, assign `-1` below and avoid branch in line 322 (line 323: `value *= sign`).

tests/js/asserts.js
```diff
@@ -19,75 +19,122 @@
...
-                message = 'Array contents not equal at index ' + i + ' (' + val1[i] + ' != ' + val2[i] + ')';
+        let compare;
+        if (val1.type === "data") {
+            compare = (i, a, b) => a === b || this.assertArraysEqual(new Uint8Array(a), b, `Data elements at index ${i}`, 1) || true;
+        }
+        else if (val1.type === "date") {
+            compare = (i, a, b) => (a && a.getTime()) === (b && b.getTime());
+        }
+        else if (val1.type === "float" || val1.type === "double") {
+            compare = (i, a, b) => a >= b - 0.000001 && a <= b + 0.000001;
```
kneth: If a and b are much larger than 1, this is equivalent to `a == b`. But it's fine as none of our tests use large value.

tests/ios/RealmJSCoreTests.m
tgoyne: The tests expect to be able to pass state between modules by assigning to properties of a variable named `global`, and that doesn't actually exist in RN (but does in everything else we run tests in).

blagoev: yes my concern is we may introduce bugs where RN (non test) code referring to 'global' might succeed in our tests but can fail elsewhere. Maybe we need to fix our tests instead. But  probably should create an issue for that

tgoyne: That's a valid concern, but I think waiting until after the release to deal with that sounds like the right choice.

src/js_schema.hpp
tgoyne: Fixed in c9a3548.

blagoev:
all is good

Issue #1342, opened 2017-09-25
"require() must have a single string literal argument" (React Native 0.49)


rh389:
It seems the version of `metro-bundler` used for `react-native@0.49-rc.5` chokes during transformation on any instance of `require` called with a variable argument, specifically: https://github.com/realm/realm-js/blob/f8f6661fe21b0abf8c876f1cef0a45ece85d5145/lib/index.js#L23 and https://github.com/realm/realm-js/blob/f8f6661fe21b0abf8c876f1cef0a45ece85d5145/lib/user-methods.js#L25

I haven't found a cross-platform solution, but for anyone looking for a quick fix I've hacked around it for RN only at https://github.com/rh389/realm-js-flat / `yarn add realm@rh389/realm-js-flat` (I've had to flatten git submodules to allow npm/yarn to work directly from github).

## Version of Realm and Tooling

- Realm JS SDK Version: master
- Node or React Native: React native 0.49.0-rc.5
- Client OS & Version: N/A
- Which debugger for React Native: Chrome

rh389:
I've just spotted that this is addressed by https://github.com/realm/realm-js/pull/1332

ashwinphatak:
Fix merged into master.

PR #1346, opened 2017-09-26
Workaround for RN >= 0.49 metro-bundler check for single string liter…

ashwinphatak:
…al argument to require() (#1342)

## What, How & Why?

Reference to the issue(s) addressed by this PR: # ???

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ X.Y.Z Release notes
 
 ### Bug fixes
 * Fixed port conflict between RN >= 0.48 inspector proxy and RPC server used for Chrome debugging (#1294).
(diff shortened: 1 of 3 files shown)


blagoev:
although this fixes the issue. The #1332 got a point that we should not be fighting the RN but working with it. These workarounds and node requires work but maybe it's worth investigating how we can remove them since they are workarounds still. 

@ashwinphatak  got a good point that #1332 may introduce a breaking change, but isn't it worth investigating how we can have #1332 in and not introducing the breaking change at the same time.

in short this fix just fixes the workaround

PR #1349, opened 2017-09-26
Sync for Windows

fealebenpae:


bmunkholm:
@fealebenpae Will you please add the outstanding tasks as "todo"?

fealebenpae:
This is generally blocked on https://github.com/realm/realm-object-store/issues/557, but we can release Sync for Windows support as a beta even with failing tests in Object Store.

The last remaining blocker for this PR is running the sync tests on Windows - they expect a ROS instance but we have no way of running one on Windows. @kneth should we disable sync tests on Windows, or is there an easy way to run ROS on another machine and use that for tests?

kneth:
@fealebenpae Let's disable the sync tests for now.

kneth:
@fealebenpae We need to rebase this on `master` instead.

Issue #1351, opened 2017-09-27
Lazily enable SyncManager


bigfish24:
When Node evaluates `const Realm = require("realm");` it creates the `/realm-object-server/io.realm.object-server-utility/metadata/sync_metadata.realm`

According to @bdash this is because:
>`SyncManager::shared().configure_file_system(…)` is called within `SyncClass<T>::create_constructor`
>Cocoa uses a bottleneck function that all accesses to `SyncManager::shared()` go through which take care of ensuring that `configure_file_system` is called. This lets the initialization happen lazily, but ensures it always happens.

This has caused problems in tests where you want to clean the local state. However, if you delete `realm-object-server` directory after `const Realm = require("realm");` you will get `mkdir` errors since the code expects the `io.realm.object-server-utility` exists.

kneth:
A fix has been merged

PR #1353, opened 2017-09-27
Enable SyncManager late

kneth:
## What, How & Why?

Reference to the issue(s) addressed by this PR: #1351

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


bdash:
Why require that a user call this explicitly rather than just doing the work ourselves the first time the user calls into any of the sync-related functionality?

/cc @blagoev

src/js_sync.hpp
```diff
@@ -39,6 +39,19 @@
 namespace realm {
 namespace js {
 
+static bool config_fs_done = false;
+
+realm::SyncManager& syncManagerShared() {
+    realm::SyncManager& shared = SyncManager::shared();
+    if (!config_fs_done) {
```
bdash: If you tweak this slightly to be:
```
SyncManager& syncManagerShared() {
    static bool configured = []{
        ensure_directory_exists_for_file(default_realm_file_directory());
        SyncManager::shared().configure_file_system(default_realm_file_directory(), SyncManager::MetadataMode::NoEncryption);
        return true;
    }();
    return SyncManager::shared();
}
```
It'll have the advantage of being thread safe by construction and slightly shorter as well.

kneth:
Nothing in particular to ask the users to do it explicitly. I have changed it so it is done every time the `SyncManager` is requested (but only doing it the first time).

src/js_sync.hpp
bdash: It also looks like `clear_test_state` still calls `SyncManager::shared()` directly, which may result in problems if `clear_test_state` is called at a point when no sync functionality has otherwise initialized sync.

bdash: This should either be declared `inline` or have the body of the function moved to a .cpp to avoid the linker errors that CI is currently seeing.

kneth: Thanks for the suggestion. Also the example is failing on CI for some reason.

bdash: The CI failure I looked at was due to the lack of `inline` I mentioned above. The js_sync.hpp header ends up included in multiple compilation units so you end up with duplicate copies of the symbol. Marking it as `inline` tells the linker to coalesce them and pick one copy. Moving the definition out of the header prevents there from being duplicates in the first place. Either should address the problem.

kneth: `clear_test_state` is call before and after every test.

blagoev:
@kneth there was a discussion on slack about handling this ourselves this boils down to doing something similar to what cocoa does. It means replacing every call to ```SyncManager::shared() ``` with a call to a function that will get the SyncManager instance and will initialize the file system

@bdash suggested this

```
SyncManager& syncManager()
{
    static bool configured = []{
        SyncManager::shared().configure_file_system(default_realm_file_directory(), SyncManager::MetadataMode::NoEncryption);
        return true;
    }();
    return SyncManager::shared();
}
```

and then everywhere we use ```shared()``` to use ```syncManager()``` instead. 

This will prepare the filesystem the moment we use the SyncManager as opposed to when the Realm constructor is created (usually on require)

The code above maybe lazy but exposes the function in JavaScript land which is not necessary.

blagoev:
oh I saw your last commit is doing exactly this. ignore the above

PR #1357, opened 2017-09-27
Update JSDocs for arrays of primitives

tgoyne:


docs/list.js
```diff
@@ -19,59 +19,70 @@
 /**
  * Instances of this class will be returned when accessing object properties whose type is `"list"`
  * (see {@linkplain Realm~ObjectSchemaProperty ObjectSchemaProperty}).
- * The objects contained in a list are accessible through its index properties and may only be
- * modified inside a {@linkplain Realm#write write} transaction.
+ *
+ * List mostly behaves like a normal Javascript Array, except for that it can
```
kneth: "Lists mostly behave" maybe

package.json
```diff
@@ -97,7 +97,7 @@
     "eslint": "^3.2.2",
     "eslint-plugin-jasmine": "^2.1.0",
     "eslint-plugin-react": "^6.7.0",
-    "jsdoc": "^3.4.0",
+    "jsdoc": "git://github.com/jsdoc3/jsdoc#b21427343c7294bbf1f14c718a390f3e955e37cb",
```
kneth: What does it gives up?

tgoyne: It turns out that jsdoc actually does nothing with @template and the commit adding support for it just made it ignore it instead of throwing an error. I was sort of expecting the fact that the type is generic to appear in the docs somewhere.

PR #1361, opened 2017-09-29
Partial sync

kneth:
## What, How & Why?

Reference to the issue(s) addressed by this PR: We don't have an issue tracking it.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


bdash:
This just looks to add the object store source files to the projects, but doesn’t expose them via the JS API. Will that come in a different PR?

kneth:
@bdash I plan to add the functionality to this PR later - that's why it is still "in active progress". I created the PR so I don't forget to implement it.

docs/realm.js
```diff
@@ -231,6 +231,17 @@ class Realm {
...
+    /**
+     * If the Realm is a partially synchronized Realm, fetch and synchronize the objects
+     * of a given object type that match the given query (in string format).
+     *
+     * **Partial synchronization is a tech preview. Its APIs are subject to change.**
+     * @param {Realm~ObjectType} type - The type of Realm objects to retrieve.
+     * @param {string} query - Query used to filter objects.
+     * @param {callback} - function to call when objects are created, modified, or deleted.
+     */
+    subscribeToObjects(className, query, callback) {}
```
bdash: Should this return a promise rather than taking a callback?

The callback taken by `partial_sync::register_query` is called precisely once: when the server has computed the initial result set, or determined that the query had an error. It's not called again when the result set changes.

src/js_sync.hpp
```diff
@@ -33,6 +33,7 @@
 #include "sync/sync_config.hpp"
 #include "sync/sync_session.hpp"
 #include "sync/sync_user.hpp"
+#include "sync/partial_sync.hpp"
```
bdash: Is this header needed here? It doesn't look like anything it provides is used here.

src/js_sync.hpp
```diff
@@ -633,14 +634,21 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
...
         config.sync_config = std::shared_ptr<SyncConfig>(new SyncConfig{shared_user, raw_realm_url,
                                                                         SyncSessionStopPolicy::AfterChangesUploaded,
                                                                         std::move(bind), std::move(error_handler),
                                                                         nullptr, util::none,
                                                                         client_validate_ssl, ssl_trust_certificate_path,
-                                                                        std::move(ssl_verify_callback)});
-
+                                                                        std::move(ssl_verify_callback),
+                                                                        is_partial,
+                                                                        util::none});
```
bdash: `util::none` can be omitted given that it's the default value for this argument.

CHANGELOG.md
```diff
@@ -1,9 +1,21 @@
+X.Z.Y Release notes
+=============================================================
+### Breaking changes
+
+### Enhancements
+* Added `Realm.subscribeToObjects()` to listen for changes in partial sync'ed Realms.
```
bdash: "partial sync'ed" -> "partially synced"?

src/js_sync.hpp
kneth: No, it should be included in `js_realm.hpp` instead.

docs/realm.js
```diff
@@ -331,7 +342,10 @@ Realm.defaultPath;
  *        The purpose of open_ssl_verify_callback is to enable custom certificate handling and to solve cases where
  *        OpenSSL erroneously rejects valid certificates possibly because OpenSSL doesn't have access to the
  *        proper trust certificates.
- *
+ *   - `partial` - Whether this Realm should be opened in 'partial synchronization' mode.
+ *        Partial synchronization mode means that no objects are synchronized from the remote Realm
```
bmunkholm: How about reversing the explanation? Instead of "no objects, except", then "only those matching...". e.g.
"Partial synchronisation only synchronizes those objects that match the query specified in contrast to the normal mode of operation that synchronises all objects in a remote Realm."

docs/realm.js
```diff
@@ -231,6 +231,17 @@ class Realm {
...
+    /**
+     * If the Realm is a partially synchronized Realm, fetch and synchronize the objects
+     * of a given object type that match the given query (in string format).
+     *
+     * **Partial synchronization is a tech preview. Its APIs are subject to change.**
+     * @param {Realm~ObjectType} type - The type of Realm objects to retrieve.
+     * @param {string} query - Query used to filter objects.
+     * @return {Promise} - a promise that will be resolved with the Realm.Results instance when it's available.
+     */
+    subscribeToObjects(className, query, callback) {}
```
blagoev: why not Realm.Sync.partialSync(className, query) or or if this must be on instance realm.syncSession.partialSync() or realm.syncSession.sync()

subscribeToObjects sounds more like notify for changes on these objects.

lib/extensions.js
```diff
@@ -141,6 +141,24 @@ module.exports = function(realmConstructor) {
             //enable deprecated setAccessToken
             realmConstructor.Sync.setAccessToken = realmConstructor.Sync.setFeatureToken;
         }
+
+
+        // instance methods
+        Object.defineProperties(realmConstructor.prototype, {
+            subscribeToObjects: function(objectType, query) {
```
blagoev: needs a change in lib/browser as well otherwise RN debugging fails

lib/extensions.js
```diff
@@ -141,6 +141,24 @@ module.exports = function(realmConstructor) {
...
+                let realm = this;
```
blagoev: const

lib/index.d.ts
```diff
@@ -599,6 +599,11 @@ declare class Realm {
      * @returns boolean
      */
     compact(): boolean;
+
+    /**
+     * @returns void
+     */
+    subscribeToObjects<T>(objectType: string, query: string, callback: (results: Realm.Results<T>, error: string) => void): void;
```
blagoev: this returns a Promise right?

src/js_realm.hpp
```diff
@@ -977,5 +985,39 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments a
...
+    auto cb = [=](realm::Results results, std::exception_ptr err) {
+        if (err) {
+            try {
+                rethrow_exception(err);
+            }
+            catch (const std::exception& e) {
+                typename T::Value arguments[2];
+                arguments[0] = Value::from_null(ctx);
+                arguments[1] = Value::from_string(ctx, e.what());
+                Function<T>::callback(ctx, callback, this_object, 2, arguments);
```
blagoev: i think you need to capture the `ctx` as Protected<ContectType>, also the `this_object`

blagoev: unless this is called immediately before unwinding the stack

tests/js/session-tests.js
```diff
@@ -646,4 +646,67 @@ module.exports = {
...
+                            sync: {
+                                user: user2,
+                                url: `realm://localhost:9080/~/${realmName}`,
+                                partial: true,
+                            },
+                            schema: [{ name: 'Integer', properties: { value: 'int' } }],
+                        };
+
+                        const realm2 = new Realm(config2);
+                        realm2.subscribeToObjects('Integer', 'value > 5').then((results, error) => {
```
blagoev: is this test really testing the partial sync. Seems to me it tests that realm.write did write more objects the the local Realm. I think real tests should be if an out of process client write some objects with type X and Y then the local already synced realm should receive only the X ones. So all the X should be present  and  the Y's should not be present

docs/realm.js
kneth: I simply used the same name as Realm Cocoa and Realm Java.

tests/js/session-tests.js
kneth: The test logs in, creates 10 objects, and logs out. Then it logs in, set up partial sync and receive 4 objects. Or at least, that's what it is supposed to do :smile:

kneth: Notice that the URI of the Realm is not the same when `partial` is `true` (it is rewritten in object store).

docs/realm.js
blagoev: its a poor name for that, but i guess we need to stick with it. is it exposed on the realm there as well? ( I haven't checked at all what other bindings are doing)

tests/js/session-tests.js
nirinchev: Shouldn't you wait for upload to complete after creating the objects. Additionally, might be a good idea to delete the Realm after the user logs out and then wait for download to complete after opening the realm the second time.

blagoev: the url seems the same to me. Looks like someone had worked over the weekend. Let's find who :)

docs/realm.js
kneth: Yes, it's an instance method on `RLMRealm`.

bdash: It can’t be on the session as there’d be know way to know which thread’s Realm the results should be associated with.

tests/js/session-tests.js
nirinchev: At least in .NET though, the file path for partial and non-partial realms is the same, so they resolve to the same path on disk (which is why I suggested deleting the Realm).

docs/realm.js
blagoev: @bdash  I didn't not get that. Isn't realm.syncSession enough to designate the target Realm. I think I didn't understand your comment correctly.

tests/js/session-tests.js
blagoev: yes I like this too. Clear state = clear test results. It seems the test needs some touch up.

kneth: I clearly misunderstood https://github.com/realm/realm-object-store/blob/master/src/sync/sync_config.hpp#L106

docs/realm.js
bdash: A single sync session is shared by all Realm instances for the same sync server URL. It’s necessary to know the specific Realm from which the matching objects should be returned (and also in which Realm we should perform the modifications to register for partial sync).

blagoev: @bdash  It is clear now. thanks. 

@kneth So these suggestions `realm.syncSession` can't work without passing an argument for the realm which will be awkward. `(realm.syncSession.subscribeToObjects(realm...)`

 so I guess either we leave it as it is or we do expose it on Realm.Sync ie `Realm.Sync.subscribeToObjects(realm, className, xx)` this has a better indication that it is related to sync than realm.subscribeToObjects which totally hides the fact, but is a mouthful.

tests/js/session-tests.js
```diff
@@ -655,29 +655,29 @@ module.exports = {
...
                                 for(let i = 0; i < 10; i++) {
                                     realm1.write(() => {
                                         realm1.create('Integer', {value: i});
                                     });
                                 }
-                            })
-                    })
-                    user1.logout();
+                                user1.logout();
+                                realm.deleteFile(config1);
```
blagoev: I think you need to wait for the changes to upload using the progress api and then do the deletion and logout.

docs/realm.js
bdash: I would strongly suggest following the lead of the other bindings for now rather than inventing something different. The API is just a preview and will be revised in the future.

blagoev: indeed staying as it is seems the best approach right now

blagoev:
👍

kneth:
I am merging the PR. The remaining test is stock downloading ROS. Since the test is disabled for RN, there is no point waiting for it.

PR #1363, opened 2017-09-30
Update the include and library search paths used when building against a local version of core / sync

bdash:
Core switched to CMake a while back, resulting in build products (generated headers, libraries) ending up in different locations than before. This looks for those files in the same locations that the sync build process expects to find them (`build.debug` for debug builds, `build.release` for release).

I'm not sure what the purpose of the `+` suffix was on some of the build settings related to the `realm-sync` target, but if I left those in place the build settings were entirely ignored. Everything appears to work correctly with those suffixes removed.

### DIFF
--- a/realm.gypi
+++ b/realm.gypi
@@ -14,11 +14,13 @@
     }],
     ["use_realm_debug", {
       "variables": {
-        "debug_library_suffix": "-dbg"
+        "debug_library_suffix": "-dbg",
+        "build_directory": "build.debug",
       }
     }, {
       "variables": {
-        "debug_library_suffix": ""
+        "debug_library_suffix": "",
+        "build_directory": "build.release",
       }
     }]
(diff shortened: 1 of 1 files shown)


blagoev:
somehow tests for Realm.open and Realm.openAsync fail with wait_for_download_completion callback not called. I can try locally later today to investigate

bdash:
The tests were failing because the branch was based on a commit of master between #1362 and #1364, so the sync version used by ROS didn't match the sync version used by Realm JS. I rebased it a few minutes ago so we should see those failures disappear.

bdash:
💥

blagoev:
👍

blagoev:
i suspect the + were there intentionally to either disable some configuration for testing locally or maybe this target was not supposed to get build.

Issue #1366, opened 2017-09-30
Ensure that a collection stays stable during iteration


bigfish24:
This is a side effect of Realm's live object behavior where the following is possible:
```js
// results.length == 2
const results = realm.objects('MyObject').filtered('myProp == null')

for (let object of results) {
    object.myProp = 'A String'
    // This runs only once rather than twice!
}
```
This is due to the fact that the query updates after the first iteration, leading to `results` to only have a length of 1. There is no error produced and the loop simply ends, leading to extremely difficult behavior to diagnose.

To fix you can simply use `snapshot()` or `slice()` on the `results`:
```js
// results.length == 2
const results = realm.objects('MyObject').filtered('myProp == null')

for (let object of results.snapshot()) {
    object.myProp = 'A String'
    // This runs twice
}
```

But it would be much better for Realm to just handle this for the developer so this mistake isn't possible to make.

PR #1368, opened 2017-10-01
Frozen results during iteration.

kneth:
## What, How & Why?

This change will align the iterator behaviour with Realm Cocoa.

Reference to the issue(s) addressed by this PR: #1366

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] ~~typescript definitions file is updated~~
* [x] jsdoc files updated
* [ ] ~~Chrome debug API is updated if API is available on React Native~~


bigfish24:
We need some doc updates to emphasize when this won’t help you.

blagoev:
@bigfish24 @fealebenpae @kneth what do you think on my last comment. (pinging here to accelerate the PR)

bigfish24:
Yes to adding to other methods as well. Sorry I couldn’t tell from Yavor’s comment if we could add to others too.

fealebenpae:
@blagoev the iterated collection is only going to change if you cause it to mutate in the body of the iteration as in the example in #1366. I think the collection methods are a lot less likely to do that as users don't normally mutate the objects in the collection in `map` or `find`. IMO we can just document this behavior for now - we implement `Symbol.iterator` ourselves so it was easy to fix the `for-of` iteration, but we directly copy a bunch of Array methods on our Collection prototype and making them use `snapshot()` would be more involved.

blagoev:
indeed users will not modify the collection during these calls. at least it is not common. my vision is if these calls like find get messed up by a background sync for example

blagoev:
we could merge this and then later revisit to broaden the fix

fealebenpae:
Another thread writing to the realm file cannot cause this as the main thread will be too busy executing the iteration to run the event loop and advance to a newer read transaction. That leaves us with the user manually mutating the collection or manually advancing the realm transaction (in the case where there is another writer that will have mutated the collection).

kneth:
As it is a breaking change, I would like to include it in 2.0.

blagoev:
👍

blagoev:
@fealebenpae still the forEach method is a candidate to be included. but lets leave it as it is for now

kneth:
Merging - works on my machine.

Issue #502, opened 2016-06-23
Delete existing realm when migration needed


stevenandianto:
In Realm for ios, there is configuration for this thing through RLMRealmConfiguration.deleteRealmIfMigrationNeeded, is there any setting that we can use for realm react native?

alazier:
This does not yet exist but we will add it at some point. In the meantime, you have two options:

 1. Update the schemaVersion/run a migration and delete all objects in the migration block to accomplish the same thing:
```
var realm = new Realm({
  schema: ...,
  schemaVersion: newestSchemaVersion,
  migration: function(oldRealm, newRealm) {
    newRealm.deleteAll();
});
```

2. There is a method that we are currently using for testing which deletes all existing Realm files in the documents directory. You could run this as a one-off to delete all Realm files whether or not a migration needs to be performed: ```Realm.clearTestState()```. Beware this method could change or be removed in a future in a release, and you probably shouldn't ship an app which is relying on this method.

jniemin:
@alazier, I would have another use case for possibility to delete Realm feature. My plan is to use separate encrypted Realm for storing user sensitive data that is retrieved from backend. I was planning to create secret on the fly on device and store that securely (device keychain). If for some reason the secret disappears from keychain I could always recover by deleting the existing encrypted Realm, creating a new secret and creating a new Realm with this newly created secret and populating that information from backend.

alazier:
@jniemin - I think what you are asking for is the same as #363 . Hopefully we can get that into a release in the near future.

alazier:
@jniemin for now you can manually delete the Realm files and associated files using something like RNFS

ashwinphatak:
Merged https://github.com/realm/realm-js/commit/670f7507f2becaa9080a7bcec54de0f0e314982f into master

Andersonfrfilho:
============>your app
```
const realm = await getRealm();
console.log(realm.path);
```
====================> In Terminal
```
adb shell rm /data/data/com.example.package/files/default.realm
```

PR #1370, opened 2017-10-02
Add 'deleteRealmIfMigrationNeeded' to configuration to delete a Realm…

ashwinphatak:
… if migration needed (#502)

## What, How & Why?

Reference to the issue(s) addressed by this PR: #502 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ X.Y.Z Release notes
 
 ### Enhancements
 * Added aggregate functions `min()`, `max()`, `sum()`, and `avg()` to `Realm.Results` and `Realm.List` (#807).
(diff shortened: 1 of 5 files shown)


ashwinphatak:
I'll add a check for that. 

Might also make sense to add another check such that `readOnly` and `deleteRealmIfMigrationNeeded` are not set at the same time.

ashwinphatak:
Better now?

pmiri:
Can this include an update to the documentation? Thanks!

kneth:
It is documented in https://realm.io/docs/javascript/latest/api/Realm.html#~Configuration. Of course, we could mention it in https://realm.io/docs/javascript/latest#opening-realms.

PR #1375, opened 2017-10-03
Deprecate node4 5

blagoev:


blagoev:
good catch https://github.com/realm/realm-js-private/pull/353

PR #1379, opened 2017-10-03
fix sync error

blagoev:
related docs fix https://github.com/realm/realm.io/pull/3876

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -361,14 +361,14 @@ declare namespace Realm.Sync {
         expiresAt?: Date;
     }
 
-    class SyncError {
+    interface SyncError {
         message: string; 
         isFatal: boolean 
         category?: string 
         code: number;
     }
-    
-    type ErrorCallback = (sender, error: SyncError) => void;
+
+    type ErrorCallback = (session: Session, error: SyncError) => void;
     type SSLVerifyCallback = (serverAddress: string, serverPort: number, pemCertificate: string, preverifyOk: number, depth: number) => boolean;
 
     interface SyncConfiguration {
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -361,7 +361,14 @@ declare namespace Realm.Sync {
         expiresAt?: Date;
     }
 
-    type ErrorCallback = (message?: string, isFatal?: boolean, category?: string, code?: number) => void;
+    class SyncError {
```
fealebenpae: I think this should be just an interface.

lib/index.d.ts
```diff
@@ -361,7 +361,14 @@ declare namespace Realm.Sync {
...
+        message: string; 
+        isFatal: boolean 
+        category?: string 
+        code: number;
+    }
+    
+    type ErrorCallback = (sender, error: SyncError) => void;
```
fealebenpae: `sender` should be `sender: Session`, I think.

blagoev: thanks, if we are sure this is always the case I can do it. I was thinking to do it `sender: any` and then i missed to do it. 
Why not call it `session: Session` then

lib/index.d.ts
blagoev: right will do

lib/index.d.ts
```diff
@@ -361,7 +361,14 @@ declare namespace Realm.Sync {
         expiresAt?: Date;
     }
 
-    type ErrorCallback = (message?: string, isFatal?: boolean, category?: string, code?: number) => void;
+    interface SyncError {
+        message: string;
```
kneth: Is only `category` optional?

blagoev: i think so. from the code it seems all other will be present with some value

Issue #808, opened 2017-01-20
Bulk update of collections


astigsen:
Like you can do in Cocoa with KVC, you should be able to call something like `.update(property, value)` on any collection to do bulk updates.

ashwinphatak:
Merged https://github.com/realm/realm-js/commit/4bcef8baff4af13722cfa7f8853fdb1c977d1565 into master.

PR #1381, opened 2017-10-04
bulk updates

ashwinphatak:
## What, How & Why?

Reference to the issue(s) addressed by this PR: #808

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/results.js
+++ b/docs/results.js
@@ -33,7 +33,7 @@ class Results extends Collection {
      * @param {string} property - The name of the property.
      * @param {string} value - The updated property value.
      * @throws {Error} If no property with the name exists.
-     * @since 1.12.1
+     * @since 2.0.0-rc20
      */
     update(property, value) {}
 }
--- a/docs/results.js
+++ b/docs/results.js
@@ -28,4 +28,12 @@
  * @memberof Realm
  */
 class Results extends Collection {
+    /**
+     * Bulk update objects in the collection.
+     * @param {string} property - The name of the property.
+     * @param {string} value - The updated property value.
+     * @throws {Error} If no property with the name exists.
+     * @since 1.12.1
+     */
+    update(property, value) {}
 }
--- a/lib/browser/results.js
+++ b/lib/browser/results.js
@@ -25,6 +25,7 @@ import { createMethods } from './util';
 export default class Results extends Collection {
 }
 
+// Non-mutating methods:
 createMethods(Results.prototype, objectTypes.RESULTS, [
     'filtered',
     'sorted',
@@ -39,6 +40,11 @@ createMethods(Results.prototype, objectTypes.RESULTS, [
     'removeListener',
     'removeAllListeners',
 ]);
+
+// Mutating methods:
+createMethods(Results.prototype, objectTypes.RESULTS, [
+    'update',
+], true);
 
 export function createResults(realmId, info) {
     return createCollection(Results.prototype, realmId, info);
--- a/src/js_results.hpp
+++ b/src/js_results.hpp
@@ -87,6 +87,8 @@ struct ResultsClass : ClassDefinition<T, realm::js::Results<T>, CollectionClass<
     template<typename Fn>
     static void index_of(ContextType, Fn&, Arguments, ReturnValue &);
 
+    static void update(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
+
     // aggregate functions
     static void min(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
     static void max(ContextType, FunctionType, ObjectType, size_t, const ValueType[], ReturnValue &);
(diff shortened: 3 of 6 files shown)


ashwinphatak:
@kneth WIP, but would love to get initial thoughts on the approach.

bmunkholm:
Supporting updates of multiple properties in a list of objects would also be cool. But not sure that's worth it, and it could make the API slightly more complicated.

blagoev:
@bmunkholm did you close this or you wanted to hit `Comment`

bmunkholm:
Ups - sorry. Got to install that github helper browser plugin that moves that damn button way to the left instead... :-)

blagoev:
feel you, mee too

lib/browser/results.js
```diff
@@ -40,6 +41,11 @@ createMethods(Results.prototype, objectTypes.RESULTS, [
     'removeAllListeners',
 ]);
 
+// Mutating methods:
+createMethods(Results.prototype, objectTypes.RESULTS, [
+    'update',
```
ashwinphatak: Am I correct in assuming that this should be in the 'mutating' methods section?

ashwinphatak:
@bmunkholm like this? -> `.update({ prop1: val1, prop2: val2, prop3: val3 })`

Indeed would be nice. Should I implement it as part of this PR?

bmunkholm:
@ashwinphatak Yes, I would actually suggest implementing it at the ObjectStore or even core layer, but that's likely a bit more involved. So doing it here is fine as a first attempt. But leave a TODO to move it to a lower layer for much better performance.

ashwinphatak:
@bmunkholm Sure, will do.

blagoev:
could you rebase this onto 2.0.x branch
I don't think we should do more features in 1.0

docs/results.js
```diff
@@ -27,4 +27,12 @@
  * @memberof Realm
  */
 class Results extends Collection {
+    /**
+     * Bulk update objects in the collection.
+     * @param {string} property - The name of the property.
+     * @param {string} value - The updated property value.
+     * @throws {Error} If no property with the name exists.
+     * @since 1.12.1
```
blagoev: update  when rebasing to 2.0

lib/browser/results.js
blagoev: yes

docs/results.js
ashwinphatak: Ok, will do!

src/js_results.hpp
```diff
@@ -302,7 +304,31 @@ void ResultsClass<T>::index_of(ContextType ctx, FunctionType, ObjectType this_ob
...
+    if (!schema.property_for_name(StringData(property))) {
+        throw std::invalid_argument(util::format("No such property: %1", property));
+    }
+
+    auto realm = results->get_realm();
+    if (!realm->is_in_transaction()) {
+        throw std::runtime_error("Can only 'update' objects within a transaction.");
+    }
+
+    for (auto i = results->size(); i > 0; i--) {
```
blagoev: I am missing something here. How this is different from doing it in JS like 
```
var results = realm.objects('NullableBasicTypesObject');
         realm.write(() => {
                 for (k i n results)
                      k.intCol = 20;
         });
```
minus the JS transitions.

can't we do a more optimized version that goes directly into object store so we gain from bulk updates

ashwinphatak: It's just moving the loop to C++ for now, with some performance gains. Ideally yes, this would be implemented in OS or Core in a "native" manner for much better performance, but using this approach to keep things simple to start with.

If I'm not mistaken, @bmunkholm was alluding to the same in his last comment. I'm going to leave a TODO to move it to a lower layer.

kneth: A optimization for the future is to operate directly on `TableView`. But that might be better done in object store than here.

blagoev: I think we don't talk about the same thing as @bmunkholm 
My reasoning is that this creates instances of RealmObjectClass<T> just to call set_property on it which I think is the same as doing it in JS except the JS transitions. What I am wondering is could we possible refactor `set_property` or user NativeAccessor directly to implement this. I think cocoa is doing something similar. Maybe I am over engineering this but the cocoa approach seems much more optimized

blagoev: here is what cocoa is doing

https://github.com/realm/realm-cocoa/blob/9df4bcb39459bb58a376a4ef0e088f8fc2376a83/Realm/RLMCollection.mm#L231

blagoev: I think @bmunkholm  suggest we have multiple property changes in one call which currently can be done in JS like so
```
results.update({ prop1: val1, prop2: val2, prop3: val3 })
(for  updatePair in arguments )
(pair => results.update(updatePair))
```
which is a "convenience method" if done like so in JS
but which can be optimized if core provided us with an `updateMany` method that will be a single call in realm-js as well

ashwinphatak: @blagoev I see your point. Will investigate how to avoid creating `RealmObjectClass` instances inside the loop, using `TableView` directly like Cocoa.

ashwinphatak:
Keeping approach unchanged per discussion on Slack. Does this look good to merge?

PR #1386, opened 2017-10-05
Merging `2.0.x` into `master`

kneth:
Let's merge `2.0.x` into `master` to simplify our processes.

Should we clean up `CHANGELOG.md`? Instead of having all the `rc`s, we could put them into one.


blagoev:
@kneth Why not merge changes.md when we do the final 2.0 release. For now I think it's ok to have it like this

PR #1396, opened 2017-10-09
Add support for codecov

emanuelez:


kneth:
:+1: when CI passes

bmunkholm:
It should exclude Test files and include src files. Looking at the Codecov report it looks like it's opposite.

kneth:
Most of our code base is in C++, and I think we can only measure the coverage in JavaScript (or is there a way to combine `istanbul` with `gcov`? See also https://stackoverflow.com/questions/21549912/how-can-i-post-code-coverage-from-a-combined-c-javascript-node-js-addon-to-cov

PR #1399, opened 2017-10-10
Use the official Node image

emanuelez:
The master branch is broken and:
 1) there are too many layers of abstraction and finding the root cause is tedious
 2) the Docker image currently being used for node builds includes way too much (ANDROID NDK, React Native etc), making those images very big and slow to build.

Here I propose to use the official Node image and cut a few layers of abstraction.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -197,14 +197,6 @@ def doDockerBuild(target, postStep = null) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


bmunkholm:
@emanuelez Please make a description that inform why this change is needed or the effect of it :-)

PR #1408, opened 2017-10-11
Add the ability to override the server that's used by a given sync session

bdash:
This is for internal use by ROS as part of its support for failover. If the session is active, we ask it to disconnect then reconnect to the new server. We also stash away the overridden server's details so it can be used again we later disconnect and reconnect. If the session is not yet active, we simply stash away the details to be used when establishing the connection.

The API looks like this:

```typescript
class Session {
    _overrideServer(address: String, port: number): void;
}
```

Fixes https://github.com/realm/realm-js-private/issues/348.

This is blocked on https://github.com/realm/realm-object-store/pull/568 being merged.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests


bigfish24:
This looks perfect!

bdash:
Writing unit tests for this is a little bit complicated. I've tested it manually for now, and hopefully ROS will be able to test it and verify it does what they need prior to this PR being merged.

alebsack:
@bdash how can I get a build of this in order to test it locally?

kneth:
@alebsack I will do a release in a few hours.

bigfish24:
I think the hope was we could test with building from source before a full releasez

bmunkholm:
@kneth We would need to include a new sync 2.0.2 that @kvap is making

PR #1409, opened 2017-10-12
adminUser() throws an exception if token or url is invalid

### LINKED ISSUE #1119: openAsync: Invalid token does not return error
Using this code with realm-1.8.1:

```
const accessToken = "INVALID TOKEN";
const syncUser = Realm.Sync.User.adminUser(accessToken);
const url = 'realm://localhost:9080/fakerealm2'

Realm.openAsync({ 
  sync: {
    user: syncUser,
    url: url
  }, 
}, (error, realm) => {
  if (error) {
    console.log(error);
  } else {
    console.log("opened realm!");
  }
});
```

I don't get an error message.  We need to properly handle this error condition.

kneth:
## What, How & Why?

If either token or server url is invalid, the returned admin user is an empty object. Throwing an exception is better since errors are raised early as possible.

This closes #1119 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests
* [ ] 📝 ~Public documentation PR created or is not necessary~

* [x] jsdoc files updated


docs/sync.js
```diff
@@ -209,6 +209,7 @@ class User {
      * @param {string} adminToken - existing admin token
      * @param {string} server - authentication server
      * @return {User} - admin user populated with the given token and server
+     * @throws {Error} If token or server is invalid.
```
bmunkholm: token -> adminToken

src/js_sync.hpp
```diff
@@ -150,7 +150,10 @@ void UserClass<T>::admin_user(ContextType ctx, FunctionType, ObjectType this_obj
         Value::validated_to_string(ctx, arguments[0], "authServerUrl"),
         Value::validated_to_string(ctx, arguments[1], "refreshToken")
     ));
-    return_value.set(create_object<T, UserClass<T>>(ctx, user));
+    if (user) {
+        return_value.set(create_object<T, UserClass<T>>(ctx, user));
+    }
+    throw std::runtime_error("Invalid token or server.");
```
bmunkholm: token -> adminToken

tests/js/session-tests.js
```diff
@@ -249,6 +249,31 @@ module.exports = {
...
+        return runOutOfProcess(__dirname + '/download-api-helper.js', username, realmName, REALM_MODULE_PATH)
+            .then(() => {
+                const invalidFeatureToken = 'THIS_IS_INVALID';
+                const url = `http://localhost:9080/~/${realmName}`;
+                return new Promise((resolve, reject) => {
+                    try {
+                        const user = Realm.Sync.User.adminUser(invalidFeatureToken, url);
+                    }
+                    catch (e) {
+                        TestCase.assertTrue(e == 'Error: Invalid token or server.');
```
bmunkholm: token -> adminToken

tests/js/session-tests.js
```diff
@@ -249,6 +249,31 @@ module.exports = {
...
+        const username = uuid();
+        const realmName = uuid();
+
+        return runOutOfProcess(__dirname + '/download-api-helper.js', username, realmName, REALM_MODULE_PATH)
+            .then(() => {
+                const invalidFeatureToken = 'THIS_IS_INVALID';
+                const url = `http://localhost:9080/~/${realmName}`;
+                return new Promise((resolve, reject) => {
+                    try {
+                        const user = Realm.Sync.User.adminUser(invalidFeatureToken, url);
```
bmunkholm: FeatureToken? The parameter is called adminToken...

bmunkholm: Need another testcase for invalid url

bmunkholm:
From Kenneth: The test for validating an unknown user or invalid Token doesn't seem to work as expected.

kneth:
I have made a very ugly hack - primarily because I don't understand the underlaying API. When I request an admin user with an invalid admin token (I believe that `THIS_IS_INVALID` is indeed invalid), I get an empty object. So the ugly hack is to throw an exception in that case.

The user returned by object store seems to be a valid (`state()` does not return `SyncUser::State::Error`) when I use my invalid admin token.

kneth:
@alebsack I am trying to solve #1119, and since you are member of Realm JS internal focus group, your opinion matters 😈

alebsack:
Works for me 👍

bmunkholm:
@kneth Is this ready for review/merge?

tests/js/user-tests.js
```diff
@@ -371,6 +371,38 @@ module.exports = {
...
+          try {
+              let user = Realm.Sync.User.adminUser('THIS_IS_INVALID', 'http://localhost:9080');
+          }
+          catch (e) {
+              didThrow = true;
+              TestCase.assertTrue(e.message === 'Invalid adminToken or server.');
+          }
+          TestCase.assertTrue(didThrow);
+
+          // FIXME: find a way to set up a proper admin token
```
blagoev: we should skip checking the good path and rename the test method to indicate it tests the failure path only.

bmunkholm:
@blagoev If it's good then merge and add another issue for what should be fixed later.

bmunkholm:
How does it affect ROS when released? @alebsack

alebsack:
I don't see it affecting ROS, as this is just a better way to highlight coding mistakes, such as not passing a proper token.

PR #1420, opened 2017-10-16
refactor OpenSSL static linking for Electron

fealebenpae:
This ensures the correct order of linker flags on Linux


bmunkholm:
@fealebenpae Who can review this?

PR #1428, opened 2017-10-24
Don't download + extract unnecessarily at build time

rh389:
## What, How & Why?
`download-realm.js` is run every time a mobile target is built. This will often download a large archive (unless it has already been downloaded since the OS temp was last cleared), and will always re-extract the archive into the target. This causes project files to be touched and (at least) Xcode to rebuild RealmJS and its dependencies.

This PR modifies `download-realm.js` to write a manifest file into the target directory containing metadata about the contents. If a manifest is present and matches what's being requested, the directory is left untouched (unless `--force` is passed). This skips an unnecessary download/extraction/build and greatly reduces build times when working on an app.

The diff looks huge because of the refactoring required to collect all the information needed before deciding whether to empty the target directory. I've changed as little as I could get away with.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/scripts/download-realm.js
+++ b/scripts/download-realm.js
@@ -23,8 +23,10 @@ const path = require('path');
 const os = require('os');
 const child_process = require('child_process');
 const fetch = require('node-fetch');
-const ini = require('ini').parse;
+const ini = require('ini');
 const decompress = require('decompress');
+
+const MANIFEST_FILENAME = '.manifest.list';
 
 function exec() {
     const args = Array.from(arguments);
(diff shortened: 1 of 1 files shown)


kneth:
@rh389 Thanks. I'll take a closer look later today. Please fill out our CLA (Contributor License Agreement - see https://github.com/realm/realm-js/blob/master/CONTRIBUTING.md).

rh389:
Thanks, CLA signed

scripts/download-realm.js
```diff
@@ -156,43 +137,100 @@ function getSyncCommitSha(version) {
...
+function readManifest(target) {
+    try {
+        return ini.parse(fs.readFileSync(path.resolve(target, MANIFEST_FILENAME), 'utf8'));
+    } catch (e) {
+        return false;
     }
+}
+
+function shouldSkipAcquire(target, desiredManifest, force) {
+    const existingManifest = readManifest(target);
```
blagoev: check `force` first if set return `false` and skip doing additional work

scripts/download-realm.js
```diff
@@ -156,43 +137,100 @@ function getSyncCommitSha(version) {
...
 
-    const downloadedArchive = path.resolve(os.tmpdir(), archive);
-    return promise
-           .then(() => download(serverFolder, archive, downloadedArchive))
-           .then(() => extract(downloadedArchive, target, archiveRootFolder));
+    if (!existingManifest) {
+        console.log('No manifest at the target, proceeding.');
+    } else if (!Object.keys(desiredManifest).every(key => existingManifest[key] === desiredManifest[key])) {
+        console.log('Target has is non-empty but has a different manifest, overwriting.');
```
blagoev: fix console message `has is`

scripts/download-realm.js
```diff
@@ -156,43 +137,100 @@ function getSyncCommitSha(version) {
...
+
+function writeManifest(target, manifest) {
+    return fs.writeFile(path.resolve(target, MANIFEST_FILENAME), ini.encode(manifest));
+}
+
+function readManifest(target) {
+    try {
+        return ini.parse(fs.readFileSync(path.resolve(target, MANIFEST_FILENAME), 'utf8'));
+    } catch (e) {
+        return false;
```
blagoev: return null instead

scripts/download-realm.js
```diff
@@ -23,9 +23,11 @@ const path = require('path');
 const os = require('os');
 const child_process = require('child_process');
 const fetch = require('node-fetch');
-const ini = require('ini').parse;
+const ini = require('ini');
 const decompress = require('decompress');
 
+const MANIFEST_FILENAME = '.manifest.list';
```
blagoev: maybe rename this file to .download-realm.list to better state what's its purpose and who is creating it

rh389: I had the same thought after submitting. Actually I wondered about making in `download-realm.lock` like Cocoapods / yarn / npm, and maybe in JSON format?

scripts/download-realm.js
```diff
@@ -156,43 +137,100 @@ function getSyncCommitSha(version) {
...
-    default:
-        throw new Error(`Unsupported sync platform '${options.platform}'`);
+        case 'mac':
+            desired.SYNC_ARCHIVE = `realm-sync-node-cocoa-${dependencies.REALM_SYNC_VERSION}.tar.gz`;
+            desired.SYNC_ARCHIVE_ROOT = `realm-sync-node-cocoa-${dependencies.REALM_SYNC_VERSION}`;
+            return Promise.resolve();
+        case 'ios':
+            desired.SYNC_ARCHIVE = `realm-sync-cocoa-${dependencies.REALM_SYNC_VERSION}.tar.xz`;
+            desired.SYNC_ARCHIVE_ROOT = `core`;
+            return Promise.resolve();
```
blagoev: It seems the old code did this but why not resolve with the `desired` object and then skip the additional `.then( () => desired)`. below

blagoev: at line 197

blagoev:
good job
(added some comments)

scripts/download-realm.js
blagoev: yep `download-realm.lock` seems fine. format can be either json or ini, whichever is prefered. I think we should gitignore it as well.

rh389:
Think that's all sorted, thanks for the review - I've left the `ini` format but happy to change if you wish, and I've changed "manifest" to "lockfile" (which now that I think about it is much more canonical for this kind of thing.)

PR #1435, opened 2017-10-25
remove setTimeout to repro issue 1255

ashwinphatak:
## What, How & Why?

PR to run all tests (especially on iOS, which is giving build errors locally) and check if we still need the `setTimeout` hack.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -71,13 +71,13 @@ module.exports = function(realmConstructor) {
                     },
                     (error) => {
                         if (error) {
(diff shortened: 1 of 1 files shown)


blagoev:
we need the current functionality as it is now to support the `progress` method on the returned `Promise`. if the setTimeout is not there the progress callback is invoked after the promise is resolved which is not desired most of the time.
Let's close this until we can rewrite the support for `progress` method on the `Promise` object

PR #1442, opened 2017-10-27
Update object store

tgoyne:
Mostly to pull in https://github.com/realm/realm-object-store/pull/587.


tgoyne:
Looking into why the windows build is failing. The files it's failing on haven't changed for a long time...

PR #1457, opened 2017-11-03
"open_ssl_verify_callback" => "ssl_verify_callback"

kraenhansen:
## What, How & Why?

The latest documentation mentions a `open_ssl_verify_callback` on the sync configuration object, but the typescript types says this is named `ssl_verify_callback`.

.... but should it actually be renamed in the types? It looks like the "open_ssl_verify_callback" is in-fact used from the c++ code https://github.com/realm/realm-js/blob/master/src/js_sync.hpp#L669

### DIFF
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -313,8 +313,8 @@ Realm.defaultPath;
  *        `category`, and `code`.
  *   - `validate_ssl` - Indicating if SSL certificates must be validated
  *   - `ssl_trust_certificate_path` - A path where to find trusted SSL certificates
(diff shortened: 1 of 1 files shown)


morten-krogh:
When I made the original PR, I called it ssl_verify_callback or something like that.
There review comments that preferred it to be called openssl_verify_callback. 
This makes sense, since the callback is only applicable for openssl. We have no
implementation for other ssl libraries.

I must have forgotten to update the typescript definition at that point.

Given that, we should probably update the typescript definition instead.
What do you think?

kraenhansen:
I agree.

Issue #1372, opened 2017-10-02
Address some comments about the client reset functionality


blagoev:
Some of the comments about the Client reset were submitted after the PR got merged. So probably we can discuss them here and decide how to proceed if needed)

https://github.com/realm/realm-js/pull/1341

PR #1459, opened 2017-11-03
Post-release feedback on client reset (#1372)

kneth:
## What, How & Why?

Addressing post-release feedback on client reset.

This closes # 1372

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/tests/js/session-tests.js
+++ b/tests/js/session-tests.js
@@ -742,13 +742,14 @@ module.exports = {
                 const config = { sync: { user, url: 'realm://localhost:9080/~/myrealm' } };
                 config.sync.error = (sender, error) => {
                     try {
+                        TestCase.assertEqual(error.name, 'ClientReset');
                         TestCase.assertEqual(error.code, 7); // 7 -> client reset
                         TestCase.assertDefined(error.config);
                         TestCase.assertNotEqual(error.config.path, '');
-                        const original_path = realm.path;
+                        const path = realm.path;
                         realm.close();
-                        Realm.Sync.initiateClientReset(original_path);
-                        // copy required objects from Realm at error.config.path
+                        Realm.Sync.initiateClientReset(path);
+                        // open Realm with error.config, and copy required objects a Realm at `path`
                         resolve();
                     }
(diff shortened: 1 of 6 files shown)


tests/js/session-tests.js
```diff
@@ -742,13 +742,14 @@ module.exports = {
                 const config = { sync: { user, url: 'realm://localhost:9080/~/myrealm' } };
                 config.sync.error = (sender, error) => {
                     try {
+                        TestCase.assertEqual(error.name, 'ClientReset');
                         TestCase.assertEqual(error.code, 7); // 7 -> client reset
```
blagoev: do we need to `error.code` if we have the `error.name` here.

kneth: If we should get rid of `error.code`, we need to map all error codes to names.

blagoev: I thought we introduced this one only here. I see your point.I guess if we decide to leave it then this error code should be coming from the source. I mean isn't it dangerous to assign some value here ourselves. since it may collide with others coming from the sync session handling code.

blagoev: Ok now I checked what `is_client_reset_requested` does here
https://github.com/realm/realm-db/blob/b0dd1381d89958474cce5584eee56d78e530a3e2/src/realm/object-store/sync/sync_config.hpp#L88

it seems it is better to leave the code as it is and client code to always compare with the name as `IncompatibleSyncedRealmError` does

I am seeing some value of having the original code present since it might help in certain scenarios. I mean to give the user these values might be of benefit to him (debugging is one of them)
``` 
ProtocolError::bad_server_file_ident
ProtocolError::bad_client_file_ident
ProtocolError::bad_server_version
ProtocolError::diverging_histories 
```

kneth: I think the current functionality is closer to our other products: https://deploy-preview-4096--realm.netlify.com/docs/java/4.1.1/api/. This will make it easier for developers.

PR #1460, opened 2017-11-03
Expose an object's internal object ID, and allow fetching an object by its object ID

bdash:
This adds an `_objectId()` method to each object, and adds `_objectForObjectId(type, id)` to the Realm.

I threw this together to help with some work I'm doing on the data connector.

It still needs:
- [x] To figure out the best behavior when used on non-synced Realms (currently both methods return undefined).
- [x] To figure out the best behavior when building without sync (currently both methods return undefined).
- [x] Tests.

Fixes https://github.com/realm/realm-js-private/issues/381.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


kneth:
@bdash Maybe the quick solution to 1. and 2. is to not expose it for non-sync builds?

simonask:
> To figure out the best behavior when used on non-synced Realms (currently both methods return undefined).

Very good question. Throwing an exception seems to be the alternative. I might have a slight preference for that, but certainly no strong opinions.

src/js_realm.hpp
```diff
@@ -987,6 +989,60 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments a
...
+    if (high_string.size() == 0 || high_string.size() > 16 || low_string.size() == 0 || low_string.size() > 16)
+        throw std::invalid_argument("Invalid object ID.");
+
+    auto isxdigit = static_cast<int(*)(int)>(std::isxdigit);
+    if (!std::all_of(high_string.begin(), high_string.end(), isxdigit) ||
+        !std::all_of(low_string.begin(), low_string.end(), isxdigit)) {
+        throw std::invalid_argument("Invalid object ID.");
+    }
+
+    return sync::ObjectID(std::stoi(high_string, nullptr, 16), std::stoi(low_string, nullptr, 16));
```
simonask: `stoi` returns an `int`, but each part of the object ID is up to 64 bits (unsigned) — use `std::stoull` instead. :-)

Any object with a string primary key is likely to have these upper bits set.

bdash: If only sync provided a function that handles this correctly so I don't have to implement it poorly myself 😄 (https://github.com/realm/realm-sync/issues/1796)

bdash:
This is now good to go, with the exception of a test failing in Chrome debug mode. I'm not quite sure what's required to teach it about the new methods that were added.

src/js_realm.hpp
simonask: Acknowledeged ;)

Shuffling this around is one of my motivations for reversing the dependency between Object Store and Sync. Particularly, I want to put the "Object ID" concept into Object Store, so everything is at least in one place.

kneth:
I have added the TypeScript definitions.

lib/index.d.ts
```diff
@@ -111,6 +111,17 @@ declare namespace Realm {
          * @returns Results<T>
          */
         linkingObjects<T>(objectType: string, property: string): Results<T>;
+
+        /**
+         * @returns string
+         */
+        _objectId(): string;
```
blagoev: are we going to document this method. If it is going to be public lets rename it `objectId`. same for the other methods

lib/index.d.ts
```diff
@@ -111,6 +111,17 @@ declare namespace Realm {
...
+
+        /**
+         * @param Object otherObject
+         * @returns boolean
+         */
+        isSameObject(otherObject: Object): boolean;
```
blagoev: isn't this with underscore in the name

blagoev:
the debuger changes  are alright with the `lib/browser` changes. I restarted the build to see whats going on

kneth:
I added the definitions to TypeScript so the ROS team can use them. But I didn't document them - on purpose.

lib/index.d.ts
kneth: I think we should keep the underscore and not document it.

lib/index.d.ts
kneth: I'll add the underscore and merge it.

blagoev:
I don't think we should add TS definitions for underscore methods. ROS team can use them with a `any` cast. Since this way we are exposing them as methods to the public in which case we should remove the underscore. That's also the reason we don't have _waitForDownload in the index.d.ts.

I hope this helps

blagoev:
I think Realm.open is causing the tests to timeout in debug here 
https://github.com/realm/realm-js/pull/1460/files#diff-4e6dfcfd63643ecc25714dc39281ccc2R55
thats the bug I am working on. I maybe wrong though. Will need to run them locally to see what's going on exactly

bdash:
I agree that we should omit internal methods from the TypeScript definitions.

lib/index.d.ts
kraenhansen: Are there good reasons this is not just going public right away?

bdash: The only current demand for this is from internal users. I expect we'll make this public once core supports stable IDs and the functions become applicable to all Realms, rather than only synced ones.

bdash:
I went ahead and skipped the test that's timing out for non-Node for now since it seems troublesome to figure out what's going wrong with it. This is consistent with some of the other sync tests that use `Realm.open`.

bdash:
I'll let folks on the JS team decide what to do about the TypeScript interface for these methods.

kneth:
I have reverted the TypeScript definition.

PR #1484, opened 2017-11-10
Update tests to not use deprecated APIs

tgoyne:
Switch to the newer promise-based APIs rather than the callback versions. The only deprecation warnings I now get when running the tests locally are for the `openAsync()` tests (which it would make no sense to switch).


blagoev:
for openAsync the plan is to remove them when we remove `openAsync` itself.

Issue #1369, opened 2017-10-01
Synced Realm not loading on Android, works fine on iOS


agersoncgps:
"realm": "^1.12.0"
"react-native": "^0.47.2",

https://pastebin.com/t6c2xNpA

This code works great in the iOS simulator and on a real iOS device. It outputs a valid object for the realm user, the realm object and it prints the correct number of records from the parents table:

User {identity: "**********", isAdmin: false, server: "https://realm.mydomain.org:9443", token: "*********…***********", Symbol(realm): "(User object)"…}

Database.js:57 Realm {…}

Database.js:59 realm-- 2567

However, when I run it on the Android simulator or on a real Android device I get 0 records synced:

User {identity: "**********", isAdmin: false, server: "https://realm.mydomain.org:9443", token:
"************…***************==", Symbol(realm): "(User object)"…}

Database.js:57 Realm {…}

Database.js:59 realm-- 0

blagoev:
Hi, thanks for reporting. We will need to debug this since from the code there is no obvious reason for sync not happening.  If there is a simple app that reproduces this it will help us a lot. 

Also you can check at your side if the realm that is returned on the Android device has increased in size and probably report what are the file sizes. 
There is a new progress API available which you can use to see if any bytes are indeed synced. The way you do it is like so"

`Realm.open(config).progress()`

here is for example one of our tests that you can use to quickly check the syntax
https://github.com/realm/realm-js/blob/master/tests/js/session-tests.js#L588-L595

agersoncgps:
I tried adding .progress() to my Realm.open under iOS where I know everything is working. No data is output to the console from the .progress() block. Did I implement it correctly?

https://pastebin.com/rETyUxDN

blagoev:
Seems all right. with one minor thing that it should be 
```
return this.getEncryptionKeyFromApi(jwt).then(data => {..............
and then 
return Realm.open
```

otherwise on error it will `leak` a rejected promise and the `catch` block will not be called. 

Still it should be calling the progress callback if it is wokring

agersoncgps:
Thanks, I added the return, but still no progress data...

blagoev:
realm-js 2.0.7 was just released with a fix for this.
Please verify it is working for you and thanks for reporting.

kneth:
@agersoncgps I close the issue but don't hesitate to reopen if you experience it after upgrading.

agersoncgps:
I updated to v2 and now get this error:

[...]

PR #1488, opened 2017-11-13
Fix realm open android debugger

### LINKED ISSUE #1248: Realm.open() hangs on Android
EDIT: This also happens after reloading the emulator on React Native, in a very simple setup (see below https://github.com/realm/realm-js/issues/1248#issuecomment-328313637)

## Goals

Get an error message when opening a synced realm if server not available or user is offline.

## Expected Results

Realm.open() returns a promise that gets rejected with useful error message.

## Actual Results

The promise simply stays in pending status indefinitely.

This callback does not get called:

https://github.com/realm/realm-js/blob/416632c9adc7350cdca12dc792edcf3e9d0b30d8/lib/extensions.js#L47

Hence the promise never returns.

## Steps to Reproduce

Run this code after log in with valid user:

```
const openRealm = (user, url, schema) => {
  return Realm.open({
    sync: { user, url },
    schema,
    schemaVersion: 1
  });
};

```

- Realm JS SDK Version: 1.10.3
- React Native: 0.44.3

blagoev:
this fixes #1248 #1369 and allows the debugger to work with arbitrary async callbacks

### DIFF
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -229,7 +229,7 @@ function deserialize_json_value(value) {
    
     return result;
 }
-let commandCounter = 0;
+
 function sendRequest(command, data, host = sessionHost) {
     clearTimeout(pollTimeoutId);
     try {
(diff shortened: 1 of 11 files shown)


lib/browser/rpc.js
```diff
@@ -230,17 +231,19 @@ function deserialize_json_value(value) {
...
+        
+        let url = 'http://' + host + '/' + command;
+        let response = makeRequest(url, data);
 
-    if (!response || response.error) {
-        let error = response && response.error;
+        if (!response || response.error) {
+            let error = response && response.error;
 
         // Remove the type prefix from the error message (e.g. "Error: ").
```
kneth: Indentation seems to be wrong.

kneth:
@blagoev Please remember to update `CHANGELOG.md`.

PR #1499, opened 2017-11-16
Use the sync label as the multiplex identifier

tgoyne:
This makes is so that when session multiplexing is enabled a unique connection is used for each backend shard, giving some measure of load balancing with multiplexing enabled.

No public docs/changelog for this since it's only relevant to a currently undocumented feature in combination with EE.

### DIFF
--- a/lib/user-methods.js
+++ b/lib/user-methods.js
@@ -92,32 +92,39 @@ function refreshAccessToken(user, localRealmPath, realmUrl) {
             // We do this because in React Native Remote Debugging
             // `Realm.clearTestState()` will have invalidated the user object
             let newUser = user.constructor.all[user.identity];
-            if (newUser) {
-                let session = newUser._sessionForOnDiskPath(localRealmPath);
-                if (session) {
-                    const errorHandler = session.config.error;
-                    if (response.status != 200) {
-                        let error = new AuthError(json);
-                        if (errorHandler) {
-                            errorHandler(session, error);
-                        } else {
-                            print_error('Unhandled session token refresh error', error);
-                        }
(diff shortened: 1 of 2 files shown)


tgoyne:
@nirinchev The .NET server stuff probably wants to do something similar to this.

lib/user-methods.js
```diff
@@ -92,32 +92,39 @@ function refreshAccessToken(user, localRealmPath, realmUrl) {
...
+                return;
+            }
+            if (session.state === 'invalid') {
+                return;
+            }
+
+            const tokenData = json.access_token.token_data;
+
+            parsedRealmUrl.set('pathname', tokenData.path);
+            session._refreshAccessToken(json.access_token.token, parsedRealmUrl.href, tokenData.sync_label);
```
bdash: If I'm reading the diff correctly, the only relevant portion of this change is the addition of this argument here, with the remainder simply being improving the formatting?

tgoyne: Yes. I started out thinking it needed significantly more changes, then it turned out that the data I needed was already there...

lib/user-methods.js
```diff
@@ -92,32 +92,39 @@ function refreshAccessToken(user, localRealmPath, realmUrl) {
...
+
+            if (errorHandler && errorHandler._notifyOnAccessTokenRefreshed) {
```
kneth: Maybe calling `print_error` like above if no `errorHandler`?

bmunkholm: @kneth Can you add that so we can merge it and include it in a release?

PR #1508, opened 2017-11-20
Update Readme.md to include debugging disclaimer

martsie:
## What, How & Why?
Some users on #491 have now reported investing hours of dev time to integrate Realm before having to remove it altogether due to the debugger being too slow for their team or project to use. We need to provide an honest disclaimer to developers until this issue is fixed as Realm is breaking a core, built-in piece of React Native functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ npm install --build-from-source=realm
 
         - Check [node-gyp](https://github.com/nodejs/node-gyp) manual for custom installation procedure for Windows
 
+## Issues with debugging
+Some users have reported the Chrome debugging being too slow to use after integrating Realm into their react-native project. This is due to the blocking nature of the RPC calls made through the Realm library and is an ongoing issue. If your team or project relies on debugging then please consider an alternative database library. See https://github.com/realm/realm-js/issues/491 for more information.
 
 ## Running the tests
 


martsie:
To say the project is open source and that you welcome contributions and in the same issue reject a contribution to the Readme that better informs developers is a bit of a contradiction.

Might I suggest we work on this PR together - what is some alternative wording we can use? Are there alternative step-through/step-into debuggers that don't slow down with Realm?

Often projects will state caveats and disclaimers in their Readme to better inform developers when they make their decision. This is not a nitpicking issue, here are some direct quotes from developers who have tried to implement Realm in their projects: 

"It is extremely slow on iOS.
Basically it can't be used as development is not feasible."

"I wonder how anyone can possibly use Realm since Chrome debugging is essential for developing a React native app..."

"Any update on this? This is critical for me, as I've got a pretty complex app with multiple Schemas (i have more than 30000 requests to get_property, it's been 25 minutes and I can't even start the app to debug it properly...)"

"Pretty sure realm wants to know if they are losing customers over a developer pain point. They lost me over this one 🤷‍♂️"

"I want to know if theres some plan or something that you guys are doing about this. I want to help if possible, I think that this is VERY important to all companies relying in Realm, if we can't debug, we can't scale and deliver our product."

"it's not a problem with Realm not being "debugable", the real problem is that Realm don't let me debug my application...Right now I'm having a problem and I can even start the Chrome "

"totally agree with @fidelisrafael.
@kristiandupont you should maybe warn users in your documentation that debugging in chrome is currently very unstable. I'm giving it an extra day or two to try to fix my debugging issues with realm, but if I have to sacrifice debugging vs realm, at this point, it's much simpler for me to go for another db."

"I'm loving realm for its fastness. But it's very hard to compromise for 5+ minutes to debug the app, since react native needs debugging in the first place. I'm having no options but to move away from realm 😞"

"Are you guys really looking at this? RIght now my application is still using Realm(it's amazing!), but this will be a real problem in a few months when our team grows up. Let's try to look a this which is one of the major problems(for me the unique "real") of Realm today..."

... (it goes on) https://github.com/realm/realm-js/issues/491

cwagner22:
@blagoev 
> The issue arise only when Realm objects are expanded and inspected

I don't think it's true, the app is slow 100% of the time when debugging is enabled.

martsie:
I echo that, on our app, initial boot time with the debugger takes around 30 seconds longer because our app does some queries to the database in order to populate the main menu. It's not that Realm fires a network request for every property access on a realm object that's the problem. The problem seems to be that it blocks the main js thread whilst waiting for the data to come back.

If Realm js was async the RPC calls wouldn't need to block the main js thread. Making async database abstraction layers for js is always the way to go, blocking the main js thread to wait for network requests is a recipe for performance issues.

At a minimum what would be great is to make all the async realm calls use async RPC calls. I'm happy to give it a go and put in a PR but it's hard to justify using my spare time if even a simple Readme PR gets rejected by the developers.

The line "suggesting alternative libraries undermines the real effort and countless hours the team behind the project" sounds like this to me "we're trying to build a competitive commercial product here, not contribute to the js open source database community", am I being cynical by thinking that?

blagoev:
Yeah, sorry, I think I got carried away there.
We obviously can’t put in notes for every issue, as there are widely different viewpoints on what is the current “biggest” issue. However, in this case, I think it would be fine as this issue has been there for way too long and is affecting most. What I think will be more constructive and helpful is to suggest alternatives/workarounds instead of saying “go somewhere else”. 

And again, we REALLY do appreciate help in tackling this issue as well as others. But let’s take that in the issue itself. Thanks!

PR #1382, opened 2017-10-04
fix node-pre-gyp not found

blagoev:
This is a naive fix for node-pre-gyp not found thrown when installing realm. 
The reason for the error is unknown since node-pre-gyp is a dependency of realm-js.
 This is a PR just to get opinions and get the conversation started about this. 

I think right now it reproduces very easily on blank RN project with
react-native init testPreGyp
npm install realm -- fails to find node-pre-gyp

PR #1511, opened 2017-11-20
use full path to node-pre-gyp

blagoev:
continuation from here https://github.com/realm/realm-js/pull/1382


blagoev:
This is caused by the use of lzma module which installs node-pre-gyp in it's node_modules dir and in turn causes node-pre-gyp symlink in .bin to not be present in realm/node_modues/.bin which results in realm-js `install` script to fail with node-pre-gyp not found

The change that removes the `tar.xz` support can be tracked here https://github.com/realm/ci/issues/247

Postponing this until the linked issue fixes the use of `tar.xz` for the sync library

blagoev:
converted to an issue

PR #1540, opened 2017-11-28
Add the header files and scripts to the gyp file

tgoyne:
This makes the generated project much more useful and otherwise has no impact on the build.


bmunkholm:
fealebenpae was quick there :-) Not sure if it should already have been in 'review' @tgoyne?

PR #1550, opened 2017-12-04
Enable force sync history option for TypeScript (Realm-Studio)

nhachicha:
- The ability to force a sync history to open a sync Realm file was added in https://github.com/realm/realm-js/pull/966/files however, it's not currently possible to set the value of `sync` to `true` when used in typescript. This is needed to fix https://github.com/realm/realm-studio/issues/392

Note: instead of updating the type to 
```
sync?: Realm.Sync.SyncConfiguration | boolean;
```
I've deliberately chosen to set it to `true` only since passing `sync:false` is redundant (since this property is optional and you can completely omit it)  

- Also fixed a side issue with `jsdoc` task not working with node 8 see https://realmio.slack.com/archives/C04R271HS/p1512409645000423?thread_ts=1512408576.000186&cid=C04R271HS

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -83,7 +83,7 @@ declare namespace Realm {
         inMemory?: boolean;
         schema?: ObjectClass[] | ObjectSchema[];
         schemaVersion?: number;
-        sync?: Realm.Sync.SyncConfiguration;
+        sync?: Realm.Sync.SyncConfiguration | boolean;
         deleteRealmIfMigrationNeeded?: boolean;
     }
 
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -83,7 +83,7 @@ declare namespace Realm {
         inMemory?: boolean;
         schema?: ObjectClass[] | ObjectSchema[];
         schemaVersion?: number;
-        sync?: Realm.Sync.SyncConfiguration | boolean;
+        sync?: Realm.Sync.SyncConfiguration | true;
         deleteRealmIfMigrationNeeded?: boolean;
     }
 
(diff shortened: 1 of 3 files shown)


bdash:
Can't Studio instead cast the config to `any` to achieve this rather than updating the TypeScript definition?

My understanding is that the TypeScript definitions only include API we consider to be public and supported. Does forcing use of a sync history really fit into that category?

nhachicha:
good point.

I'm fine either way, we can use `sync:true as any` as mentioned WDYT @kraenhansen ?

lib/index.d.ts
```diff
@@ -83,7 +83,7 @@ declare namespace Realm {
         inMemory?: boolean;
         schema?: ObjectClass[] | ObjectSchema[];
         schemaVersion?: number;
-        sync?: Realm.Sync.SyncConfiguration;
+        sync?: Realm.Sync.SyncConfiguration | true;
```
kneth: Not `boolean`?

nhachicha: :) you didn't read my PR description 🙃

blagoev:
@nhachicha please close this if this is not going to land

nhachicha:
closing this as the option should not be exposed as public API, jsdoc changes are tracked here https://github.com/realm/realm-js/pull/1553

kraenhansen:
😢 would have liked this ... why is it that we need to have some docs public and some private? It doesn't really make sense to me ...

bdash:
@kraenhansen, what part of it doesn't make sense? There's functionality that is useful to us internally when building ROS or Realm Studio that we don't wish to spend resources on documenting / supporting in an ongoing fashion for external users.

PR #1556, opened 2017-12-05
Checking the content-type of authentication responses

### LINKED ISSUE #1555: "Unexpected token < in JSON at position 0" when connecting to a broken ROS
## Goals

I want a better error message when issuing an /auth request to a broken instance of ROS.

## Expected Results

An error with the following message:

> Could not authenticate: Realm Object Server didn't respond with valid JSON.

## Actual Results

I get a syntax error with this message:

> Unexpected token d in JSON at position 0

## Steps to Reproduce

Try to authenticate with a ROS instance that does not have the auth service started or to any server that does not implement the ROS protocol.

## Code Sample

    Realm.Sync.User.login('http://realm.io', 'user', 'password').then(console.log, console.error);

outpus this on node.js:

> { FetchError: invalid json response body at https://realm.io/404?domain=realm.io&path=%2Fauth reason: Unexpected token < in JSON at position 0
    at /Users/kraenhansen/Repositories/realm-studio/node_modules/node-fetch/lib/body.js:48:31
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)
  name: 'FetchError',
  message: 'invalid json response body at https://realm.io/404?domain=realm.io&path=%2Fauth reason: Unexpected token < in JSON at position 0', type: 'invalid-json' }

And this from an Electron renderer process:

> SyntaxError: Unexpected token < in JSON at position 0

Ideally Realm JS should check if the response from the server has the expected content-type before parsing it as JSON (here https://github.com/realm/realm-js/blob/master/lib/user-methods.js#L157-L159):

    if (response.headers.get('Content-Type') !== 'application/json') {
      return response.text().then((body) => {
        throw new AuthError({
          title: `Could not authenticate: Realm Object Server didn't respond with valid JSON`,
          body,
        });
      });
    } else if (!response.ok) {
      return response.json().then((body) => Promise.reject(new AuthError(body)));
    } else {
      // ...
    }

- Realm JS SDK Version: 2.0.11
- Node or React Native: Both
- Client OS & Version: Mac

kraenhansen:
## What, How & Why?

This closes #1555 and https://github.com/realm/realm-studio/issues/417

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/tests/js/user-tests.js
+++ b/tests/js/user-tests.js
@@ -160,6 +160,20 @@ module.exports = {
       .catch((e) => assertIsError(e));
   },
 
+  testLoginTowardsMisbehavingServcer() {
+    const username = uuid();
+
+    // Try authenticating towards a server thats clearly not ROS
+    return Realm.Sync.User.register('https://github.com/realm/realm-js', username, 'user')
+      .catch((e) => {
+        assertIsError(e);
+        TestCase.assertEqual(
+          e.message,
+          "Could not authenticate: Realm Object Server didn't respond with valid JSON"
+        );
+      });
+  },
+
   testAll() {
     const all = Realm.Sync.User.all;
     TestCase.assertArrayLength(Object.keys(all), 0);
@@ -324,4 +338,3 @@ module.exports = {
   }, */
 
 };
-
--- a/tests/js/user-tests.js
+++ b/tests/js/user-tests.js
@@ -160,7 +160,7 @@ module.exports = {
       .catch((e) => assertIsError(e));
   },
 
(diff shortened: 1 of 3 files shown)


tests/js/user-tests.js
```diff
@@ -160,6 +160,20 @@ module.exports = {
       .catch((e) => assertIsError(e));
   },
 
+  testLoginTowardsMisbehavingServcer() {
```
fealebenpae: typo in `Servcer`.

PR #1557, opened 2017-12-06
Implement implicit property conversion for date and binary

### LINKED ISSUE #1542: Relax type restriction for date fields
To enable importing directly from json, it would be nice to allow dates to be set as strings, e.g. I'd like to parse the following json and pass it directly to `realm.create`:

```
{
  "id": 1,
  "foo": "bla bla",
  "createdAt": "2017-11-28T10:15:10.916Z"
}
```

### LINKED ISSUE #1551: Relax type restrictions for binary fields
To make it easier to support json (de)serialization, it'd be nice if we have implicit conversion between binary and base64 strings.

fealebenpae:
Closes #1542
Closes #1551

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


bdash:
I'm not really a fan of supporting these types of implicit conversions. Other bindings have repeatedly declined to provide them.

That said, if we strongly feel that we should add this to realm-js, I'd argue that we need to do it in a way that doesn't give different behavior depending on whether sync is available or not. These conversions have nothing to do with sync, so having different behavior on that basis is awful.

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+
+### Enchancements
+* None
+
+### Bug fixes
+* None.
+
+### Internal
+* Strings can now be assigned to Date columns. When that happens the JavaScript Date constructor will be invoked to parse the string.
+* Base64 strings can now be assigned to Data columns.
```
bdash: Why are these listed as "Internal"? They're user-visible changes in behavior, so should presumably be listed under "Enhancements".

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+vNext Release notes (TBD)
+=============================================================
+### Breaking changes
+* None.
+
+### Enchancements
```
bdash: `s/Enchancements/Enhancements/`

kneth:
I can see that this will make parsing JSON easier, and therefore I am not against adding the enhancements.

But I agree with @bdash that sync and non-sync versions should behave the same. I realize that it requires that our base64 decoder is moved from sync to core.

nirinchev:
I am not against moving the base64 decoder to core, but that'll require releases of core and sync, which we don't have the time to wait for. I'll advocate for merging this as is because it's needed for the GraphQL launch next week and the GraphQL service will only open synced realms. Then, once the necessary releases have been cut, we can enable it for non-synced realms and document it publicly.

fealebenpae:
We don’t ship a version of realm-js without sync on any platform. The sync/not-sync distinction is just not there in the final package.

fealebenpae:
> Why are these listed as "Internal"?

Because I think we shouldn’t have implicit conversions, either, but unfortunately we need them for another product. I’d rather include this as a private and undocumented feature that we can get rid of as soon as possible.

kneth:
@nirinchev @fealebenpae If you leave a big `FIXME` saying that the base64 decoder just be moved, I am fine with merging it.

blagoev:
shipz it

PR #1560, opened 2017-12-12
Adding property Realm.isClosed.

kneth:
## What, How & Why?

Sometimes it is useful to check if a Realm instance is closed or not. The readonly property `Realm.isClosed` is added.

`isClosed` has been chosen as it adhere to the naming in our other SDKs:

* https://realm.io/docs/dotnet/2.1.0/api/reference/Realms.Realm.html#Realms_Realm_IsClosed
* https://realm.io/docs/java/4.3.1/api/io/realm/Realm.html#isClosed--
* Realm Cocoa (Swift and Objective-C) can auto-close instances

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -73,7 +73,7 @@ class Realm {
     get isInTransaction() {}
 
     /**
-     * Indicated is this Realm is closed.
+     * Indicates if this Realm has been closed.
      * @type {boolean}
      * @readonly
      * @since 2.1.0
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -73,6 +73,14 @@ class Realm {
     get isInTransaction() {}
 
     /**
+     * Indicated is this Realm is closed.
+     * @type {boolean}
+     * @readonly
+     * @since 2.1.0
+     */
+     get isClosed() {}
+
+    /**
      * Gets the sync session if this is a synced Realm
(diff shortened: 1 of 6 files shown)


bmunkholm:
Why not `isOpen` :-)
Seriously - are there any precedence from other SDKs?

docs/realm.js
```diff
@@ -72,6 +72,14 @@ class Realm {
      */
     get isInTransaction() {}
 
+    /**
+     * Indicated is this Realm is closed.
```
nirinchev: `Indicates if this Realm has been closed.`?

nirinchev:
.NET has [`IsClosed`](https://realm.io/docs/dotnet/latest/api/reference/Realms.Realm.html#Realms_Realm_IsClosed).

kneth:
@fealebenpae I think `Realm.closed` and `Realm.close()` are too close (pun not intended).

fealebenpae:
That's true. While we can always have `open`, I don't insist on either.

kraenhansen:
For what it's worth, I think its good convention to have boolean properties prefixed with `is`, I've seen that a lot elsewhere and it makes me expect that its a pure accessor.

PR #1575, opened 2017-12-15
Updating object store

kneth:
Sync 2.1.10 requires an update to object store too.


blagoev:
also make sure android builds locally

PR #1580, opened 2017-12-18
Fix Realm.open complaining about the Realm already being open with a different schema version

### LINKED ISSUE #1478: Realm at path already opened with different schema version.
## Goals

migrate the databse

## Expected Results
database migrate
## Actual Results

 Realm at path '/data/data/com.realmmigration/files/default.realm' already opened with different schema version.

## Steps to Reproduce
clone the project start the app on master then close the app 
switch to branch version2 and restart the app

## Code Sample
[repo](https://github.com/artis-auxilium/realm-migration)

- Realm JS SDK Version: 2.0.4
- React Native: 0.49
- Client OS & Version: android 6.0 (not tested on ios)

bdash:
In order to correctly open read-only synchronized Realms, `Realm.open` would open the Realm without specifying a schema or schema version, wait for any remote changes to be downloaded (if appropriate), and then re-open the Realm with the specified schema and schema version. This would lead to an exception about the Realm being open with a different schema version if the Realm had previously been opened with a different schema version, due to the way `RealmCoordinator` caches information about the schema of open Realms.

We address this by making two changes:
1. `Realm.open` for non-synchronized Realms no longer goes through `_waitForDownload`. This means the dance described above where the Realm is opened twice is not used for local Realms.
2. `_waitForDownload` no longer keeps the `Realm` alive until after its callback has returned. It instead keeps the `SyncSession` alive. This is sufficient to avoid the connection being torn down and having to reconnect when `_waitForDownload`'s callback later opens the Realm with the correct schema and schema version, while also allowing for `RealmCoordinator`'s cached information to be cleared when the schemaless Realm is closed prior to the Realm being reopened.

In addition, tests have been added that reproduced the problem in both a local and sync context.

Fixes #1478.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+X.Y.Z Release notes
+=============================================================
+### Breaking changes
+* None.
+
+### Enhancements
+* None.
+
+### Bug fixes
+* Fix a bug where `Realm.open` could unexpectedly raise a "Realm at path ... already opened with different schema version" error.
+
+### Internal
+* None
+
 2.1.1 Release notes (2017-12-15)
 =============================================================
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bug fixes
 * [Object Server] Fixed a bug where long reconnection happens when a proxy in front of the sync worker returns one of those.
-* Fix a bug where `Realm.open` could unexpectedly raise an "Realm at path ... already opened with different schema version" error.
+* Fix a bug where `Realm.open` could unexpectedly raise a "Realm at path ... already opened with different schema version" error.
 
 ### Internal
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Bug fixes
 * [Object Server] Fixed a bug where long reconnection happens when a proxy in front of the sync worker returns one of those.
+* Fix a bug where `Realm.open` could unexpectedly raise a "Realm at path ... already opened with different schema version" error.
```
bmunkholm: This has to be added to a 2.1.2 section, as 2.1.1 was already released.

bdash: Ugh, I’m used to empty change log sections being added after a release is made, so I didn’t even notice this section was for a version that was already released.

blagoev: i will handle it

Issue #1568, opened 2017-12-14
Partial Sync: Support named queries


cmelchior:
See https://github.com/realm/product/issues/390 for details

PR #1583, opened 2017-12-19
Partial sync

### LINKED ISSUE #1570: Partial Sync: Unify notification system
See https://github.com/realm/product/issues/392 for details

kneth:
## What, How & Why?

Closes #1570, #1568, #1569 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


lib/browser/results.js
```diff
@@ -30,6 +30,7 @@ createMethods(Results.prototype, objectTypes.RESULTS, [
     'filtered',
     'sorted',
     'snapshot',
+    'subscribe',
```
nirinchev: Have we reached binding-level agreement on the `subscribe` terminology? My only concern is that it doesn't seem descriptive enough and may be confused with subscribing for notifications.

lib/extensions.js
```diff
@@ -159,18 +159,13 @@ module.exports = function(realmConstructor) {
...
-                        reject(err);
-                    } else {
-                        resolve(results);
-                    }
-                });
-            });
-            return promise;
+        // Keep these value in sync with subscription_state.hpp
+        realmConstructor.Sync.SubscriptionState = {
+            Undefined: -3,         // Unknown which state Partial Sync is in.
```
nirinchev: Are these going to be exposed to end-users? If yes, `unknown which state Partial Sync is in` is too vague.

lib/extensions.js
```diff
@@ -159,18 +159,13 @@ module.exports = function(realmConstructor) {
...
+            NotSupported: - 2,     // Partial Sync not supported.
+            Error: -1,             // An error was detect in Partial Sync.
+            Uninitialized: 0,      // The subscription was just created, but not handled by sync yet.
+            Initialized: 1,        // The subscription has been initialized successfully and are syncing data to the dev
```
nirinchev: `to the dev`?

src/js_collection.hpp
```diff
@@ -23,6 +23,9 @@
 #include "js_observable.hpp"
 
 #include "collection_notifications.hpp"
+#if REALM_ENABLE_SYNC
```
nirinchev: Is there a case where the js binding has this set to false?

kneth: I believe that we test both with and without sync on CI.

lib/extensions.js
kneth: `ice`

lib/extensions.js
kneth: Agree, we should document them more carefully them in the API documentation.

src/js_collection.hpp
nirinchev: Seems like wasted cycles if we don't release it after testing 😛

lib/browser/results.js
bdash: I don't think there's been any serious discussion about the specific naming used here. If you're concerned about this naming, we have very little time to resolve it, so I'd suggest bringing it up in Slack so we can hammer it out promptly.

src/js_collection.hpp
bdash: Isn't realm-js on Linux built without sync unless you're installing one of the paid editions?

nirinchev: I don't think so - we have a single package and feature tokens to unlock functionality.

src/js_realm.hpp
```diff
@@ -248,7 +248,7 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
         {"_objectForObjectId", wrap<object_for_object_id>},
  #if REALM_ENABLE_SYNC
         {"_waitForDownload", wrap<wait_for_download_completion>},
-        {"_subscribeToObjects", wrap<subscribe_to_objects>},
+        {"getQueryStatus", wrap<get_query_status>},
```
bdash: Is this used for anything? I cannot see any references to `get_query_status`.

src/js_results.hpp
```diff
@@ -254,6 +263,31 @@ void ResultsClass<T>::is_valid(ContextType ctx, ObjectType this_object, Argument
...
+#if REALM_ENABLE_SYNC
+template<typename T>
+void ResultsClass<T>::subscribe(ContextType ctx, ObjectType this_object, Arguments args, ReturnValue &return_value) {
+    args.validate_maximum(1);
+
+    auto results = get_internal<T, ResultsClass<T>>(this_object);
+    auto realm = results->get_realm();
+    auto sync_config = realm->config().sync_config;
+    if (!sync_config || !sync_config->is_partial) {
+        throw std::logic_error("A partial sync query can only be registered in a partially synced Realm.");
```
bdash: `partial_sync::subscribe` performs this check internally, so I don't think it's necessary to repeat it here?

docs/sync.js
```diff
@@ -468,6 +468,58 @@ class Session {
...
+
+    /**
+     * Gets the error message. `undefined` if no error.
+     * @type {string}
+     */
+    get error() {}
+
+    /**
+     * Unsubscribe a partial synced `Realm.Results`. The state will change to `Realm.Sync.SubscriptionState.Invalidated`.
+     * The `Realm.Results` will not produce any meaningful values.
```
bdash: It'd be worth mentioning here that any objects matching this query will be removed if they are not matched by any other query, and that they'll be removed asynchronously with respect to this method.

docs/sync.js
```diff
@@ -468,6 +468,58 @@ class Session {
...
+    unsubscribe() {}
+
+    /**
+     * Adds a listener `callback` which will be called when the state of the subscription changes.
+     * @param {function(state)} callback - A function to be called when changes occur.
+     * @throws {Error} If `callback` is not a function.
+     * @example
+     * subscription.addListener((subscription, state) => {
+     *     if (state == Realm.Sync.SubscriptionState.Complete) {
+     *         var partial_results = subscription.results;
```
bdash: I think you removed the `results` property for now?

Issue #1586, opened 2017-12-22
Realm.Sync.User.all returns only a subset of users when multiple server addresses are involved


bdash:
As discussed in https://github.com/realm/realm-object-store/issues/618#issuecomment-353628602 and https://github.com/realm/realm-object-store/issues/618#issuecomment-353629464, `Realm.Sync.User.all` returns a dictionary keyed by user identity, which assumes that the user identity is unique across all servers. That's not the case if you connect to the same server via multiple hostnames and log the same user in on each. This results in only a subset of the available users being returned.

See https://github.com/realm/realm-object-store/issues/618#issuecomment-353623813 for a code snippet that reproduces the issue with Realm JS v2.1.1.

PR #1587, opened 2017-12-22
When refreshing the token, look up the user in a way that doesn't suffer from #1586

bdash:
Expose a means of looking up a user by identity and server to avoid problems if the same user identity exists for multiple servers, which can happen when connecting to the same server via different hostnames.

Works around #1586 in the token refresh subsystem.

This may fix https://github.com/realm/realm-object-server-private/issues/790.

## ☑️ ToDos

* [x] Should the new method be public? **No**
* [x] Should the new method be able to return a user that's logged out? **No**
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] ~jsdoc files updated~


bdash:
@kneth, do you have any thoughts on the two questions asked in the ToDo list above?

kneth:
@bdash I don't think it should be public. So far, we are the only one who has experienced it. And I doubt it is useful to obtain a logged-out user. So I say "no" to both questions.

nirinchev:
@kneth what's the status of this? It appears to be the cause of somewhat frequent test failures in ROS and would be great if we could get it with the next js release.

kneth:
@nirinchev Today is big merge day which means that it will be merge soon. I hope to do a release either later today or tomorrow.

bdash:
Thanks for getting this merged, @kneth!

PR #1594, opened 2018-01-04
Add basic docs for multiprocess notifications

tgoyne:
A first stab at documentation for https://github.com/realm/realm-js-private/pull/391

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -41,8 +41,16 @@ class Sync {
      *
      * Currently supported events:
      *
-     *  * "available": Emitted whenever a new Realm is registered with the server. Passes the virtual path of the new Realm to the callback.
-     *  * "change": Emitted whenever a Realm has been changed. Passes a ChangeEvent object to the callback.
+     *  * `'available'`: Emitted whenever there is a new Realm which has a virtual
+     *    path matching the filter regex, either due to the Realm being newly created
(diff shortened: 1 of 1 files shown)


kneth:
I think we should make it clear which cases (operating systems, platforms, licenses) multiprocess notifications are supported.

kneth:
@tgoyne I have merged #1602 which should avoid the CI failures in this PR.

Is the PR ready to be reviewed and merged?

docs/sync.js
```diff
@@ -81,41 +116,60 @@ class Sync {
...
     get oldRealm() {}
 
     /**
-     * The change indexes for all added, removed, and modified objects in the changed Realm.
-     * This object is a hashmap of object types to arrays of indexes for all changed objects:
+     * The change indexes for all added, removed, and modified objects in the
+     * changed Realm. This object is a hashmap of object types to arrays of
+     * indexes for all changed objects:
+     *
+     * Note that deleting an object in Realm can cause other objects in the
```
kneth: Consider moving the notes to come after the example.

tgoyne: That doesn't appear to be possible with jsdoc. There's no `@endexample` or equivalent.

kneth: Sadly true, http://usejsdoc.org/tags-example.html

PR #1596, opened 2018-01-05
Explicitly send register: false during login

nirinchev:
## What, How & Why?

This fixes a bug that will be introduced with next version of ROS where not sending a value for `register` will be interpreted as `createOrUpdate`.

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ X.Y.Z Release notes
   should help with failing token refreshes on a loaded server.
 
 ### Internal
-* None.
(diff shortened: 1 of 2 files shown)


nirinchev:
@kneth what do you think the behavior of the newly introduced `authenticate` should be? Currently, it doesn't send `register`, which means that when ROS is updated, it'll create the user if it doesn't exist. I feel this is fine but wanted to double check first

nirinchev:
The CI failures seem to be unrelated to this PR

kneth:
@nirinchev Good point with `authenticate()`. We could add an extra (optional) parameter to it?

And I know how to fix the CI, so I'll do that.

PR #1599, opened 2018-01-06
Perform a pseudo-refresh for admin token users

tgoyne:
Although we don't need to actually refresh the token, when multiplexing is enabled we do still need to hit ROS to let it know we're accessing a Realm and get the sync label.

JS part of the fix for https://github.com/realm/realm-object-server-private/issues/836.


bigfish24:
@kneth @bmunkholm this is a bug affecting RapidFunnel that needs fixed. Can we review and create a release?

bigfish24:
The original issue is tracked here: https://github.com/realm/realm-object-server-private/issues/836#issuecomment-355704046

tgoyne:
This currently makes the realm-js-private tests fail. I'm working towards fixing that in https://github.com/realm/realm-js-private/tree/tg/test-proxy, and that shouldn't prevent reviewing this.

bmunkholm:
@tgoyne So I assume we can move this to "Review"?

lib/user-methods.js
```diff
@@ -63,13 +60,63 @@ function print_error() {
...
+
+function refreshAdminToken(user, localRealmPath, realmUrl) {
+    const token = user.token;
+    const server = user.server;
+
+    // We don't need to actually refresh the token, but we need to let ROS know
+    // we're accessing the file and get the sync label for multiplexing
+    let parsedRealmUrl = url_parse(realmUrl);
+    const url = append_url(user.server, 'realms/files/' + encodeURIComponent(parsedRealmUrl.pathname));
+    performFetch(url, {method: 'GET', timeout: 1000.0, headers: {Authorization: user.token}})
```
kneth: In #1584 the timeout is increased to 10 secs.

PR #1621, opened 2018-01-15
Copying definition of functions

kneth:
## What, How & Why?

Methods like `Realm.Sync.addListener` is now implemented in JavaScript, and JS files must be distributed as part of a release.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -120,13 +120,10 @@ if (!realmConstructor) {
 require('./extensions')(realmConstructor);
 
 if (realmConstructor.Sync) {
-  switch(getContext()) {
-    case 'nodejs':
-    case 'electron':
-        require('./notifier')(realmConstructor);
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -120,8 +120,13 @@ if (!realmConstructor) {
 require('./extensions')(realmConstructor);
 
 if (realmConstructor.Sync) {
-  require('./notifier')(realmConstructor);
-  Object.defineProperty(realmConstructor, 'Worker', {value: require('./worker')});
+  switch(getContext()) {
+    case 'nodejs':
+    case 'electron':
```
fealebenpae: Electron doesn’t have the global notifier, so we shouldn’t apply this for it.

kneth: :+1:

PR #1626, opened 2018-01-17
Add support for async sync change listeners

tgoyne:


tgoyne:
Tests for this added in https://github.com/realm/realm-js-private/pull/400.

PR #1643, opened 2018-01-29
Exposing get_number_of_versions

kneth:
## What, How & Why?

To get a better understanding on why some customers see large file sizes when using the data adapter.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -470,6 +470,7 @@ declare class Realm {
     readonly schemaVersion: number;
     readonly isInTransaction: boolean;
     readonly isClosed: boolean;
+    readonly _numberOfVersions: number;
 
     readonly syncSession: Realm.Sync.Session | null;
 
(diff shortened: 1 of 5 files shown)


lib/index.d.ts
```diff
@@ -470,6 +470,7 @@ declare class Realm {
     readonly schemaVersion: number;
     readonly isInTransaction: boolean;
     readonly isClosed: boolean;
+    readonly _numberOfVersions: number;
```
fealebenpae: I believe you can also use the `private` modifier here.

tests/js/realm-tests.js
```diff
@@ -1029,6 +1029,18 @@ module.exports = {
         TestCase.assertTrue(realm1.compact());
     },
 
+    testNumberOfVersions: function() {
+        const realm1 = new Realm({schema: [schemas.StringOnly]});
+        const realm2 = new Realm({schema: [schemas.StringOnly]});
+
```
bmunkholm: assert versions == 1 here?

kneth:
I don't think this feature will provide any useful information.

PR #1648, opened 2018-01-30
Wait for pending notifications to complete when removing a sync listener

tgoyne:
With async callbacks, in-process listeners need to be able to wait for anything currently in progress to complete just like out-of-process listeners. In addition, removeListener() needs to be declared as returning a Promise so that typescript code can actually use it.


bmunkholm:
@kneth Do add a release note entry please

PR #1653, opened 2018-02-06
Hook up the sort/distinct orderings from parsed queries

ironage:
This depends on https://github.com/realm/realm-object-store/pull/631

It gives users an ability to apply sort and distinct from inside the query string. Any externally applied sort/distinct calls will chain on top of the ones applied in the string.

I think having this ability gives maximum flexibility, but it is possible that it could also add confusion. If we didn't want JS to support sort/distinct inside the query string, we could simply check that the ordering result is empty and throw otherwise. Input welcome.


ironage:
Updated the changelog. It does support multiple properties, I added some more tests for that.
@kneth is this PR merging to the correct branch?

kneth:
@ironage Looks perfect. Thanks.

PR #1655, opened 2018-02-06
Add Schema Name to Error Message with Primary Key

### LINKED ISSUE #1654: Need to specify schema in error message when missing primary key.
## Goals

Clearer error messages when realm-js schema is all of a sudden missing primary key

## Expected Results

The error message should specify the schema name. 

## Actual Results
```
Missing primary key property 'ID'
```

It doesn't say which schema

## Steps to Reproduce

See code sample

## Code Sample

```typescript
import * as Realm from 'realm'

const TruckType = {
  name: 'TruckType',
  primaryKey: 'ID',
  properties: {
      ID: 'int',
      Name: 'string',
      MaintenanceIntervalMile: { type: 'int', optional: true },
      MaintenanceIntervalDay: { type: 'int', optional: true },
      MaintenanceIntervalYear: { type: 'int', optional: true },
      Description: { type: 'string', optional: true },
  }
};

const TruckType2 = {
  name: 'TruckType',
  primaryKey: 'ID',
  properties: {
      Name: 'string',
      MaintenanceIntervalMile: { type: 'int', optional: true },
      MaintenanceIntervalDay: { type: 'int', optional: true },
      MaintenanceIntervalYear: { type: 'int', optional: true },
      Description: { type: 'string', optional: true },
  }
};

const realm = new Realm({schema: [TruckType]})
realm.close()

const again = new Realm({schema: [TruckType2]})
```

- Realm JS SDK Version: 2.2.6
- Node or React Native: 8.9.0
- Client OS & Version: MacOSX, Linux

mbalex99:
## What, How & Why?

Adds "Schema with name" and object schema name in error message.

This closes #1654 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/src/js_schema.hpp
+++ b/src/js_schema.hpp
@@ -247,7 +247,7 @@ ObjectSchema Schema<T>::parse_object_schema(ContextType ctx, ObjectType object_s
         object_schema.primary_key = Value::validated_to_string(ctx, primary_value);
         Property *property = object_schema.primary_key_property();
         if (!property) {
-            throw std::runtime_error("Schema with name '" + object_schema.name + "' is missing primary key property '" + object_schema.primary_key + "'");
+            throw std::runtime_error("Schema named '" + object_schema.name + "' specifies primary key of '" + object_schema.primary_key + "' but does not declare a property of that name.");
         }
         property->is_primary = true;
     }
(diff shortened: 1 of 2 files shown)


src/js_schema.hpp
```diff
@@ -247,7 +247,7 @@ ObjectSchema Schema<T>::parse_object_schema(ContextType ctx, ObjectType object_s
         object_schema.primary_key = Value::validated_to_string(ctx, primary_value);
         Property *property = object_schema.primary_key_property();
         if (!property) {
-            throw std::runtime_error("Missing primary key property '" + object_schema.primary_key + "'");
+            throw std::runtime_error("Schema with name '" + object_schema.name + "' is missing primary key property '" + object_schema.primary_key + "'");
```
bdash: It might be clearer to say something like "Schema named X specifies primary key of Y, but does not declare a property of that name".

PR #1441, opened 2017-10-27
Update React Native Android buildToolsVersion to 26.0.2

noymer:
## What, How & Why?

In #992 buildToolsVersion was updated to 25.0.2,
but for build as below, buildToolsVersion in `publish_android_template` should also be updated.

>cd react-native/android
>./gradlew publishAndroid
>
>https://github.com/realm/realm-js#building-realm

And I update to the latest buildtools version: 26.0.2

PR #1656, opened 2018-02-07
TypeScript: change Realm.Object interface to class

noymer:
## What, How & Why?

When extending Realm.Object, typescript compile shows error.

```
// Person.ts
import Realm from 'realm'

class Person extends Realm.Object {
}
```

``` 
// tsc output
Type '{ readonly prototype: Object; }' is not a constructor function type.

class Person extends Realm.Object {
                                   ~~~~~~~~~~~~
```

To fix this, I have changed `lib/index.d.ts`.

--

I don't know how to use `Realm.Object` except `extends Realm.Object`,
if there are any other use cases, the code should be modified.

env
* TypeScript: 2.7.1
* React Native: 0.52.0
* realm-js: 2.1.1

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -114,9 +114,7 @@ declare namespace Realm {
         linkingObjects<T>(objectType: string, property: string): Results<T>;
     }
 
-    const Object: {
-        readonly prototype: Object;
-    }
+    class Object: { }
 
     /**
      * SortDescriptor


kneth:
@noymer Sorry for the late reply. Our CI system is not triggered by non-Realm users so I have recreated your PR in #1679. Did you sign the CLA?

noymer:
@kneth 

Yes, I have signed the CLA.
I have sent a pull request  once #1441

kneth:
Since any `Realm.Object` is a managed object (that is, directly tied to the Realm files and managed by Realm Core), extending it is not supported. I guess that your code is looking something like:

```
import * as Realm from "realm";

class Person extends Realm.Object {
    schema: any
};
Person.schema = {
    name: 'Person',
    properties: {
        name: 'string',
        age: 'int'
    }
};

let realm = new Realm({schema: [Person.schema]});
realm.write(() => {
    realm.create(Person.schema.name, { name: 'John', age: 42 });
});
console.log(realm.objects(Person.schema.name).length);
```

What is support is to let the schema be the centre instead:

```
import * as Realm from "realm";

let Person = {
    name: 'Person',
    properties: {
        name: 'string',
        age: 'int'
    }
} as Realm.ObjectSchema;

let realm = new Realm({ schema: [Person] });
realm.write(() => {
    realm.create(Person.name, { name: 'John', age: 42 });
});
console.log(realm.objects(Person.name).length);
```

PR #1660, opened 2018-02-14
Preliminary work for queries over named backlinks

ironage:
This should add the ability to query over named backlinks as described in #1498

This depends on a core release including https://github.com/realm/realm-core/pull/2989

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Better tests

PR #1677, opened 2018-02-27
Update query docs

astigsen:
Updated query docs in API reference to reflect the features of the new query parser.

### DIFF
--- a/docs/tutorials/query-language.md
+++ b/docs/tutorials/query-language.md
@@ -82,10 +82,10 @@ You can query on aggregates over properties in the lists using the aggregate ope
 Example:
 ```JS
 // Find contacts without friends
-let teens = realm.objects('Contact').filtered('friends.@count == 0');
+let lonely = realm.objects('Contact').filtered('friends.@count == 0');
 
 // Find contacts where the average age of their friends is above 40
-let adults = realm.objects('Contact').filtered('friends.@avg.age > 40"');
+let adults = realm.objects('Contact').filtered('friends.@avg.age > 40');
 ```
 
 Subqueries using the `SUBQUERY` operator allows you to filter the lists across multiple parameters while querying them.
@@ -93,7 +93,7 @@ Subqueries using the `SUBQUERY` operator allows you to filter the lists across m
 Example:
 ```JS
 // Find contacts with friends above 21 in SF
-let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = 'SF').@count > 0');
+let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = "SF").@count > 0');
 ```
 
 ### Backlink queries
--- a/docs/tutorials/query-language.md
+++ b/docs/tutorials/query-language.md
@@ -103,8 +103,8 @@ Other objects can link to an object and you can query on that releationship usin
 Example:
 ```JS
 // Find contacts with no incomming links
-let lonely = realm.objects('Contact').filtered('@links.@count == 0);
+let lonely = realm.objects('Contact').filtered('@links.@count == 0');
 
 // Find contacts where someone from SF has them as friends
-realm.objects('Contact').filtered('@links.Contact.friends.city == "SF");
+realm.objects('Contact').filtered('@links.Contact.friends.city == "SF"');
 ```
--- a/docs/tutorials/query-language.md
+++ b/docs/tutorials/query-language.md
@@ -95,16 +95,3 @@ Example:
 // Find contacts with friends above 21 in SF
 let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = "SF").@count > 0');
 ```
-
-### Backlink queries
-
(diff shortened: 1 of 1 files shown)


docs/tutorials/query-language.md
```diff
@@ -53,6 +53,58 @@ let peopleWhoseNameContainsA = realm.objects('Contact').filtered('name CONTAINS[
...
+
+Example:
+```JS
+// Find contacts with friends above 21 in SF
+let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = 'SF').@count > 0');
+```
+
+### Backlink queries
+
+Other objects can link to an object and you can query on that releationship using the `@links` and `@links.ClassName.PropertyName` syntax:
```
ironage: `@links` by itself is not currently supported, it must be proceeded by a class name and property name. Is this something that would return the sum of all backlink columns? If that is desired, let's open a core issue to track it.

Additionally, the class name part must be prefixed by "class_" which exposes something that should be kept internal. I would actually prefer that we keep backlink queries undocumented if possible, since we will have support for querying over named backlinks in https://github.com/realm/realm-js/pull/1660

docs/tutorials/query-language.md
```diff
@@ -53,6 +53,58 @@ let peopleWhoseNameContainsA = realm.objects('Contact').filtered('name CONTAINS[
...
+// Matches "John" and "Johnny"
+let Johns = realm.objects('Contact').filtered('name LIKE "John*"');
+```
+
 ### Composition
 Use parentheses and the `&&`/`AND` and `||`/`OR` operators to compose queries. You can negate a predicate with `!`/`NOT`.
 
+### Queries on collections
+
+When objects contain lists you can query into them using the collection operators `ANY`, `ALL` and `NONE`.
```
ironage: I discovered that `ALL` and `NONE` queries are matching objects with no items in the list which I believe is incorrect. A workaround is to append `AND friends.@count > 0` to `ALL` and `NONE` queries, but I wonder if we should fix this behaviour in core before documenting these?

astigsen: We should check what the NSPredicate behavior is here. cc @bdash

docs/tutorials/query-language.md
astigsen: I'll create an issue for standalone links queries.

>  would actually prefer that we keep backlink queries undocumented if possible, since we will have support for querying over named backlinks 

In js it is common to open realms without specifying a schema, relying on the dynamic nature of the language, so you cannot rely on there being named backlinks defined. Could we filter out the "class_" prefix in the parser?

ironage: Yes, it should be simple to try to match the exact class name and if that fails add a "class_" prefix automatically.

astigsen: It might be enough if we just add an option to auto-add the prefix that can be enabled when bindings use the parser (since the "class_" prefix is never exposed at the user level).

docs/tutorials/query-language.md
```diff
@@ -53,6 +53,58 @@ let peopleWhoseNameContainsA = realm.objects('Contact').filtered('name CONTAINS[
...
+let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = "SF").@count > 0');
+```
+
+### Backlink queries
+
+Other objects can link to an object and you can query on that releationship using the `@links` and `@links.ClassName.PropertyName` syntax:
+
+Example:
+```JS
+// Find contacts with no incomming links
```
kneth: "incomming" -> incoming"

nirinchev:
Would be nice to show how to query primitive collections - e.g. check if a list of numbers contains any number larger than x, or if a list of strings contains any string that starts with something (not sure what is supported).

docs/tutorials/query-language.md
```diff
@@ -53,6 +53,58 @@ let peopleWhoseNameContainsA = realm.objects('Contact').filtered('name CONTAINS[
...
 Use parentheses and the `&&`/`AND` and `||`/`OR` operators to compose queries. You can negate a predicate with `!`/`NOT`.
 
+### Queries on collections
+
+When objects contain lists you can query into them using the collection operators `ANY`, `ALL` and `NONE`.
+
+Example:
+```JS
+// Find contacts with one or more teenage friends
+let teens = realm.objects('Contact').filtered('ANY friends.age < 14');
```
ironage: @nirinchev There's an example of a query operating on a list of numbers is here, if it's not clear that the string operators are supported, we could add something like `ANY friends.name CONTAINS[c] "rob"`?

astigsen:
> Would be nice to show how to query primitive collections - e.g. check if a list of numbers contains any number larger than x, or if a list of strings contains any string that starts with something (not sure what is supported).

Queries against lists of literals is not supported yet as far as I know.

astigsen:
I removed the section about backlink queries for now. We can add it in again when we have fixed how class names are referenced.

docs/tutorials/query-language.md
nirinchev: I was thinking of an example for a query against a list of strings, e.g. `'ANY keywords CONTAINS[c] "apple"'` but @astigsen mentioned these are not supported :( If that's the case probably we should mention it explicitly?

ironage: Sadly, querying lists of primitives is not yet supported :(

PR #1680, opened 2018-02-28
Avoid Table::clear() when using partial sync

kneth:


src/js_realm.hpp
```diff
@@ -898,7 +898,13 @@ void RealmClass<T>::delete_all(ContextType ctx, ObjectType this_object, Argument
     }
 
     for (auto objectSchema : realm->schema()) {
-        ObjectStore::table_for_object_type(realm->read_group(), objectSchema.name)->clear();
+        auto table = ObjectStore::table_for_object_type(realm->read_group(), objectSchema.name);
+        if (realm->is_partial()) {
```
bdash: FWIW, in the equivalent Cocoa code I opted to do `Results(realm, *table).clear()` so as not to have to duplicate this logic in the binding.

PR #1693, opened 2018-03-08
Default sync realm

### LINKED ISSUE #1688: Default Sync Realm API
As part of the upcoming release of partial sync, Realm Platform Self-Hosted and Cloud will offer the concept of a default synced Realm. This is designed to simplify how developers start working with synced data. Specifically it will fix or improve the following:

1. Simplify how developers think about their data model. Previously developers needed to split their data into multiple Realms to control data access. Now with partial sync and the new permission system this is not necessary.
2. Mirror the ease of use of Realm Database which offers a similar default Realm convenience API
3. Limit the code needed to open a synced Realm - now the user simply needs to login with just the auth URL instead of needing to worry about Realm URLs and paths
4. The default synced Realm will include the master User list of the application allowing developers to incorporate relationships to users simplifying how to think about user metadata.

API Proposal:

```
/*
  Open the default synced Realm
  - Has a dedicated API so that we can have a default configuration using:
    - SyncUser.current
    - Default Realm URL based off SyncUser auth URL (supplied within the refresh token)
    - Hardcoded path
    - Setup the Realm with a sync history type

  - If opened before there is a SyncUser.current one of two things can happen depending on the implementation:
    - The simple implementation is to throw if there is no SyncUser.current
    - The advanced implementation is to open the Realm without a sync session allowing offline access
*/

// Java
Realm realm = Realm.getDefaultSyncInstance();

// Swift
let realm = try! Realm.defaultSync()

// Objective-C
RLMRealm *realm = [RLMRealm defaultSyncRealm];

// JS
// Currently lacks a default Realm, so we should add
var realm = new Realm.defaultSync()

// .Net
var realm = Realm.GetDefeaultSyncInstance();

/*
  Login Example
*/

let credentials = SyncCredentials.nickname(nickname: "Adam", isAdmin: false)
let serverURL = NSURL(“http://testapp.us1.cloud.realm.io”)
SyncUser.logIn(with: credentials,
               server: serverURL) { user, error in
    if let user = user {
        // can now open the default synchronized Realm with this user
        let realm = try! Realm.defaultSync()
    } else if let error = error {
        // handle error
    }
}

```

kneth:
## What, How & Why?

This closes #1688 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   - Method `Realm.privilges()` to compute privileges on a Realm, a Realm object schema, or a Realm object. The method returns either a `Realm.Permissions.Realm` or `Realm.Permissions.Class` object.
   - For non-synced Realms, all privileges are always granted.
   - For more details, please read the reference documentation.
+* Added `Realm.default()` which will return the default synced Realm instance (#1688).
 
 ### Internal
 * Updated to Realm Core 5.3.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
   - Method `Realm.privilges()` to compute privileges on a Realm, a Realm object schema, or a Realm object. The method returns either a `Realm.Permissions.Realm` or `Realm.Permissions.Class` object.
   - For non-synced Realms, all privileges are always granted.
   - For more details, please read the reference documentation.
-* Added `Realm.default()` which will return the default synced Realm instance (#1688).
+* Added `Realm.defaultSyncConfiguration()` which will return the configuration for a default synced Realm (#1688).
 
 ### Internal
 * Updated to Realm Core 5.3.0.
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -271,8 +271,8 @@ class Realm {
      * then replacing the database with the temporary one.
      * @returns {true} if compaction succeeds.
      */
-    compact() {}}
-
+    compact() {}
+}
 
 /**
  * Get the current schema version of the Realm at the given path.
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -120,14 +120,13 @@ class Realm {
     static openAsync(config, callback, progressCallback) {}
 
     /**
-     * Open the default Realm. The server URL for the current user will be used as base for
-     * the URL for the synced Realm. If {Realm.Sync} is not enabled, the default local Realm will
-     * be opened.
+     * Return a configuration for a default synced Realm. The server URL for the current user will be used as base for
+     * the URL for the synced Realm.
      * @throws {Error} if zero or multiple users are logged in
-     * @returns {ProgressPromise} - a promise that will be resolved with the Realm instance when it's available.
+     * @returns {Realm~Configuration} - a configuration matching a default synced Realm.
      * @since 2.3.0
      */
-    static default() {}
+    static defaultSyncConfiguration() {}
 
     /**
      * Closes this Realm so it may be re-opened with a newer schema version.
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -120,6 +120,16 @@ class Realm {
     static openAsync(config, callback, progressCallback) {}
 
     /**
+     * Open the default Realm. The server URL for the current user will be used as base for
+     * the URL for the synced Realm. If {Realm.Sync} is not enabled, the default local Realm will
+     * be opened.
+     * @throws {Error} if zero or multiple users are logged in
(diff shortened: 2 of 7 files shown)


docs/realm.js
```diff
@@ -261,8 +271,8 @@ class Realm {
      * then replacing the database with the temporary one.
      * @returns {true} if compaction succeeds.
      */
-    compact() {}
-}
+    compact() {}}
```
fealebenpae: Seems like formatting got messed up here.

lib/extensions.js
```diff
@@ -231,6 +233,35 @@ module.exports = function(realmConstructor) {
         });
     }
 
+
+    // A default Realm instance
+    realmConstructor.default = function() {
+        var user;
+        var config;
+
+        if (realmConstructor.Sync) {
```
fealebenpae: I believe we should use `this` here instead of capturing `realmConstructor`.

lib/extensions.js
```diff
@@ -231,6 +233,35 @@ module.exports = function(realmConstructor) {
...
+            user = realmConstructor.Sync.User.current;
+            if (user === undefined) {
+                throw new Error("No user is logged in");
```
fealebenpae: Since this method returns a Promise I think it’s better to `return Promise.reject(new Error())` instead of throwing.

fealebenpae: And because `User.current` also throws it might be best to use `User.all` instead, or surround this with a try/catch that returns a rejected promise. Or just surround the body of the method in `new Promise` and that will take care of handling throw nerrors.

lib/index.d.ts
```diff
@@ -561,6 +561,11 @@ declare class Realm {
      */
     static openAsync(config: Realm.Configuration, callback: (error: any, realm: Realm) => void, progressCallback?: Realm.Sync.ProgressNotificationCallback): void
 
+    /**
+     * Open the default Realm.
+     */
+    static default(): ProgressPromise;
```
fealebenpae: Shouldn’t this be `ProgressPromise<Realm>`?

kneth: `interface ProgressPromise extends Promise<Realm>`

docs/realm.js
```diff
@@ -119,6 +119,16 @@ class Realm {
...
 
+    /**
+     * Open the default Realm. The server URL for the current user will be used as base for
+     * the URL for the synced Realm. If {Realm.Sync} is not enabled, the default local Realm will
+     * be opened.
+     * @throws {Error} if zero or multiple users are logged in
+     * @returns {ProgressPromise} - a promise that will be resolved with the Realm instance when it's available.
+     * @since 2.3.0
+     */
+    static default() {}
```
bdash: This naming and behavior seems very confusing. There's no indication that what you're getting back is a synced Realm, and it will sometimes give back a non-synced Realm?

This pattern of providing a factory method was discussed by other bindings on Slack, and we instead decided to provide a factory for a _configuration_, since that makes it much more flexible to use, and allows using it for other places that need a configuration.

docs/realm.js
```diff
@@ -119,6 +119,15 @@ class Realm {
...
     static openAsync(config, callback, progressCallback) {}
 
+    /**
+     * Return a configuration for a default synced Realm. The server URL for the current user will be used as base for
+     * the URL for the synced Realm.
+     * @throws {Error} if zero or multiple users are logged in
+     * @returns {Realm~Configuration} - a configuration matching a default synced Realm.
+     * @since 2.3.0
+     */
+    static defaultSyncConfiguration() {}
```
bdash: It looks like you renamed this in the documentation, but not in the change log or implementation.

kneth: Damn, forgot to push

CHANGELOG.md
```diff
@@ -23,7 +23,8 @@
   - Method `Realm.privilges()` to compute privileges on a Realm, a Realm object schema, or a Realm object. The method returns either a `Realm.Permissions.Realm` or `Realm.Permissions.Class` object.
   - For non-synced Realms, all privileges are always granted.
   - For more details, please read the reference documentation.
-* Added `Realm.default()` which will return the default synced Realm instance (#1688).
+* Added `Realm.defaultSyncConfiguration()` which will return the configuration for a default synced Realm (#1688).
+* [Sync] Decrepated `Realm.Sync.setFeatureToken` (#1689).
```
bdash: Decrepated -> Deprecated

Issue #1635, opened 2018-01-23
Jest tests with jest.resetModules() fails after upgrade to 2.2.4


jerdie:
## Goals

update to realm@2.2.4

## Expected Results

Successful test

## Actual Results

got error from test with realm imported
```
    TypeError: Cannot redefine property: Worker
        at Function.defineProperty (native)

    > 1 | import Realm from 'realm';
```

## Steps to Reproduce

1. install realm2.2.4
2. 
```
import Realm from 'realm';
describe("Testing realm", () => {
    beforeEach(() => {
        jest.resetModules();
    });
    test("realm", () => {
        const realm2 = require('realm')
    })
})
```
3. run test

## Version of Realm and Tooling

- Realm JS SDK Version: 2.2.4
- Node or React Native: React native jest
- Client OS & Version: Windows
- Which debugger for React Native: None

Adding configurable: true in realm/lib/index.js's Object.defineProperty(realmConstructor fixes the issue.
Not sure if this is a bug. Can you provide a workaround?

jerdie:
Update:

I noticed that this happens on tests that have mocks. Error occurs when you do a jest.resetModules() or jest.resetAllMocks() and then do a require('realm').

kneth:
I can easily imagine that mocking can lead to troubles. A Realm object is not a plain JavaScript object as the object is managed by Realm and not the JS engine. 

We have to investigate a bit further and I label it as a bug for now.

Thanks for the report and the suggested fix.

jdelafon:
> Error occurs when the test has import, do a jest.resetModules() or jest.resetAllMocks() and then do a require('realm').

Sorry but I don't understand what the workaround is.

I import Realm in a service that I want to test, I cannot just `require('realm')` in each one of my tests. 

I am not using Jest mocks, but I get the same error on all test suites involving Realm.

It happens randomly, too, while adding code and without changing any lib version. I reverted my work to a commit where the tests did pass, but now they do not anymore. Any cache to reset maybe (I tried --clearCache) ?

jerdie:
Still haven't found a workaround for this. I just stated how to replicate it in my last comment. Once i get the error, all of my test that has a module that imports realm also gets the error. I'm stuck using "realm": "~2.1.1".

benweidig:
Running into the same issue, some of my services import Realm directly so jest.resetModules() isn't fixing it 100% for me

kneth:
I would like to know if anyone have tried to upgrade to 2.2.14?

[...]

PR #1695, opened 2018-03-08
Stop worker being redefined and breaking Jest

alexbassy:
## What, How & Why?
I was experiencing a bug whereby I am:
- Running Jest tests for a class that uses Realm
- When I make a file change, I get an error message `Cannot redefine property: Worker` and have to restart the test watcher

Fix: I wrapped it in an if-statement. I don't know what `realmConstructor.Sync` or `realmConstructor.Worker` is, so I don't know if this breaks something yet 🤓 

This addresses #1635, and _for me_ it fixes the problem

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -122,7 +122,9 @@ require('./extensions')(realmConstructor);
 if (realmConstructor.Sync) {
     if (getContext() === 'nodejs') {
       nodeRequire('./notifier')(realmConstructor);
-      Object.defineProperty(realmConstructor, 'Worker', {value: nodeRequire('./worker')});
+      if (!realmConstructor.Worker) {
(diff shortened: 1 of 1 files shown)


tgoyne:
This seems like it should be harmless as a workaround.

The root problem is that `jest.resetModules()` can't reload the native module, so the Javascript module initialization code is being called on the native object that's already been initialized. I'm not sure how confident I am that the rest of the init code handles this case correctly, but I guess if it appears to work it can't be completely wrong? We could probably do more to handle this case well (and maybe even call our internal reset-for-testing function when we detect that it's happened).

PR #1696, opened 2018-03-08
Revoke refresh token upon logout.

### LINKED ISSUE #1354: Revoke refresh token upon Logout
On Logout, we should revoke the user's refresh token: realm/ros#290

kneth:
## What, How & Why?

This closes #1354 

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


lib/user-methods.js
```diff
@@ -348,6 +348,23 @@ const staticMethods = {
...
 const instanceMethods = {
+    logout() {
+        this._logout();
+        const url = url_parse(this.server);
+        url.set('pathname', '/auth/revoke');
+        const headers = {
+            Authorization: this.token
+        };
+        const options = {
+            method: 'GET',
```
nirinchev: I think this should be POST: https://github.com/realm/realm-object-server-private/blob/master/src/services/AuthService.ts#L582

lib/user-methods.js
```diff
@@ -348,6 +348,23 @@ const staticMethods = {
...
 const instanceMethods = {
+    logout() {
+        this._logout();
+        const url = url_parse(this.server);
+        url.set('pathname', '/auth/revoke');
+        const headers = {
+            Authorization: this.token
+        };
+        const options = {
+            method: 'POST',
```
nirinchev: This should send the token in the body (the idea is that with a valid token, you could invalidate another token):

https://github.com/realm/realm-object-server-private/blob/master/src/services/AuthService.ts#L587

PR #1704, opened 2018-03-12
Kneth/allow   partial in urls

kneth:
## What, How & Why?

This closes https://github.com/realm/realm-studio/issues/658

## ☑️ ToDos

* ~[ ] 📝 Changelog entry~
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~

### DIFF
--- a/src/js_sync.hpp
+++ b/src/js_sync.hpp
@@ -824,12 +824,12 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
         }
 
         bool disable_partial_sync_url_checks = false;
-        ValueType disable_partial_sync_url_value = Object::get_property(ctx, sync_config_object, "_disablePartialSyncUrlChecks");
-        if (!Value::is_undefined(ctx, disable_partial_sync_url_value)) {
-            disable_partial_sync_url_urls = Value::validated_to_boolean(ctx, disable_partial_sync_url_value);
+        ValueType disable_partial_sync_url_checks_value = Object::get_property(ctx, sync_config_object, "_disablePartialSyncUrlChecks");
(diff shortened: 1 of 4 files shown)


src/js_sync.hpp
```diff
@@ -823,7 +823,13 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
             is_partial = Value::validated_to_boolean(ctx, partial_value);
         }
 
-        config.sync_config = std::make_shared<SyncConfig>(shared_user, std::move(raw_realm_url));
+        bool force_open_realms = false;
+        ValueType force_open_realms_value = Object::get_property(ctx, sync_config_object, "forceOpenRealms");
```
bdash: This seems like an awfully general configuration option name for such a specific purpose.

kneth: True, maybe `allowOpenPartialSyncURLs` is better.

kneth: I have on purpose not updated `CHANGELOG.md` or API doc. @kraenhansen isn't happy about underscores.

bmunkholm: We can update CHANGELOG.md under #internal. I think it's fine to have some way of making unsupported internal APIs clear - e.g. with _. Do we have others like that?

bmunkholm:
@kneth I liked your other naming suggestion better. "checks" is unclear. "_allowOpeningPartialSyncUrl" is more descriptive. Not that it's critical for internal use though.

kneth:
@bmunkholm I agree with @bdash on this one: `_allowOpeningPartialSyncUrl` almost implies that you have to set it to true if you are using partial sync. Maybe "validation" is better than "checks".

bmunkholm:
Never mind. But perhaps a comment next to the field "Private - Only for internal use" might be good.

PR #1713, opened 2018-03-14
Fixing TS definition for open_ssl_verify_callback.

### LINKED ISSUE #1652: `open_ssl_verify_callback` has incorrect TypeScript types
## Goals

TypeScript types for the `open_ssl_verify_callback` that match the docs and runtime behaviour.

## Expected Results

I expect the `SSLVerifyCallback` to have a single argument in the types as described by the documentation: https://realm.io/docs/javascript/latest/api/Realm.html#~Configuration:

> open_ssl_verify_callback - A callback function used to accept or reject the server's SSL certificate. open_ssl_verify_callback is called with an object of type
> 
>     {
>       serverAddress: String,
>       serverPort: Number,
>       pemCertificate: String,
>       acceptedByOpenSSL: Boolean,
>       depth: Number
>     }

## Actual Results

The types expect the verify callback to have a bunch of arguments - not just a single object:

>     type SSLVerifyCallback = (serverAddress: string, serverPort: number, pemCertificate: string, preverifyOk: number, depth: number) => boolean;
> 
>     interface SyncConfiguration {
>       ...
>       open_ssl_verify_callback?: SSLVerifyCallback;
>       ...
>     }

- Realm JS SDK Version: 2.2.5 (latest)
- Node or React Native: N/A
- Client OS & Version: N/A

kneth:
## What, How & Why?

This closes #1652

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


kraenhansen:
I don't think this is fixing the original issue.

kraenhansen:
The callback still has a lot of arguments - I would expect that it took only a single object as argument as the docs suggests (and the runtime behaviour confirms).

kneth:
Ah, they are all transferred as one object with many properties: https://github.com/realm/realm-js/blob/master/src/js_sync.hpp#L341

PR #1715, opened 2018-03-15
Added `writeCopyTo` functionality

### LINKED ISSUE #1678: Feature request: provide API for `write_copy`
## Goals

It would be great to have `write_copy` accessible from JS world. In our use case this might allow to create CLI apps to generate and/or preprocess realms for testing as it's much easier to do in JS than by using "native" version.

## Expected Results

Since `write` is already taken and looks more like transaction in Swift APIs, I expect to have API like `realm.saveCopy(path, encryptionKey)` or `realm.duplicateTo(path, encryptionKey)`. 
(but have to note that using of `encryptionKey` for us is definitely overkill, so that can be omitted).

## Actual Results

There is no way to create a copy.

vhbit:
Fixes #1678

### DIFF
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -272,6 +272,17 @@ class Realm {
      * @returns {true} if compaction succeeds.
      */
     compact() {}
+
+    /**
+     * Writes a compacted copy of the Realm to the given path.
+     *
+     * The destination file cannot already exist.
+     *
+     * Note that if this method is called from within a write transaction, the current data is written,
+     * not the data from the point when the previous write transaction was committed.
+     * @param {string} path path to save the Realm to
+     */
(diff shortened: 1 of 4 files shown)


vhbit:
@kneth please review

kneth:
It looks fine to me. But if you add a second optional argument, we can also support changing encryption key (like Realm Cocoa: https://realm.io/docs/objc/3.2.0/api/Classes/RLMRealm.html#/c:objc(cs)RLMRealm(im)writeCopyToURL:encryptionKey:error:).

Did you sign the Contributor License Agreement (CLA)?

vhbit:
@kneth 
> Did you sign the Contributor License Agreement (CLA)?
yup

> But if you add a second optional argument
done

PR #1717, opened 2018-03-16
Throttle HTTP requests made to the auth server

tgoyne:
Making thousands of HTTP requests simultaneously inevitably results in some failing or the server being overwhelmed, so limit how many can be made at once (currently to the arbitrary cap of 5) and enqueue any further requests.

*Should* fix https://github.com/realm/realm-object-server-private/issues/957, and I've verified that it does fix another customer's similar issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ X.Y.Z Release notes
 * Added an optional user as argument to `Realm.automaticSyncConfiguration` (#1708).
 
 ### Bug fixes
-* None.
+* [Sync] Avoid hammering the ROS authentication service when large numbers of Realms are opened at once.
 
 ### Internal
 * None.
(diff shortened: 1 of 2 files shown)


kneth:
@tgoyne Is there any way to find the optimal value of `maxCount`?

bmunkholm:
@tgoyne How long can that queue get? Is there any way to throttle back the source?

tgoyne:
> Is there any way to find the optimal value of maxCount?

No, there inherently is no optimal value as it depends on how many clients are accessing the server, what requests are being made, and how many cpu cores the server have. Most modern browsers have settled on a limit of 6.

> How long can that queue get? Is there any way to throttle back the source?

When opening 10k realms, it will be 10k requests long as they all get enqueued long before the first one finishes. I don't see any advantage to adding queuing in multiple places.

PR #1724, opened 2018-03-22
Avoid leaking realms when an error occurs within an event handler

tgoyne:
Previously returning a rejecting promise from an event handler would result in silently never receiving events for that Realm again. This makes it close the Realm and rethrow the error instead so that things will either crash or continue somewhat sensibly.


bmunkholm:
Add changelog.

kneth:
@bmunkholm I will do that when I release.

kneth:
Again the react-native in release mode hangs while in debug mode completes. I am merging.

PR #1725, opened 2018-03-23
Avoid leaking realms when an error occurs within an event handler

kneth:


kneth:
I am merging. The react-tests for Mac is passing in debug but release seems to hang.

PR #1731, opened 2018-03-28
Stringify fetch body in logout

thebnich:
In React Native, calling `user.logout()` throws an `Unsupported BodyInit type.` error. The body should be stringified here like it is elsewhere throughout this file.

I noticed there are already [logout tests](https://github.com/realm/realm-js/blob/master/tests/js/user-tests.js#L70); I didn't try to run them, but I assume they're passing because they're using a different `fetch` dependency?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+X.Y.Z Release notes
+=============================================================
+### Breaking changes
+* None.
+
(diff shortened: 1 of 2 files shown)


thebnich:
@kneth Looks like a regression from 9d8a677ecff1b367f3bee86ac4b8603e8248bcf6.

PR #1747, opened 2018-04-11
Retry on network errors when refreshing admin tokens

tgoyne:
This just copies the logic used for non-admin token user fetches, which seems to already work fine in practice.

Fixes https://github.com/realm/realm-js-private/issues/433.


bmunkholm:
needs changelog update.

kneth:
@bmunkholm I have updated `CHANGELOG.md`.

PR #1757, opened 2018-04-20
Add password reset wrappers (#1699).

### LINKED ISSUE #1699: Add wrappers for new password provider methods
We’ve added a few additional methods in ROS that allow people to implement password reset and email confirmation flows.
While these are simple HTTP API, it’d be nice to create wrappers for them. All of these rely on the idea that the
username/password provider is actually used as email/password (which we probably have to highlight in the method docs).

The url for all actions is: `POST https://server-url/auth/password/updateAccount`.

The body looks like:

```js
{
    "provider_id":"user's email - for some actions it's optional",
    "data": {
        "action":"%ACTION%",
        // Other fields that depend on action
    }
}
```

All return an empty response and will not provide indication if the user doesn't exist, but will return errors if
some of the parameters are invalid.

1. Request password reset: a reset password email is sent to the user's email with some opaque token.
    - `provider_id` is **required**
    - `action` is `reset_password`
    - **No** other fields in `data`
2. Confirm password reset: after obtaining the token from the email, the user can complete the password reset flow.
By default, ROS will send an email that will redirect to `http://ros-url/some-page?token=%TOKEN%` where a generic
UI will be shown to complete the password reset. However, it's possible to override this URL and e.g. change it to
`myapp://reset-password/%TOKEN%` so that it's handled by the app itself. It's the developer's responsibility to obtain
the token if they want to complete the reset flow within the app.
    - `provider_id` should not be provided
    - `action` is `complete_reset`
    - `data` contains the following additional fields:
        - `token`: the token sent to the user's email address.
        - `new_password`: the new password of the user.
3. Initiate email confirmation: for a lot of use cases, people may want to know that the user has confirmed their email
before allowing them to take certain actions in the app. An email will be sent to the user's email address with an opaque token.
    - `provider_id` is **required**
    - `action` is `request_email_confirmation`
    - **No** other fields in `data`
4. Confirm email: similarly to the password reset flow, the user obtains the token from the email somehow any must pass
it to the server. By default, we'll redirect to a generic page hosted in ROS.
    - `provider_id` should not be provided
    - `action` is `confirm_email`
    - `data` contains the following additional fields:
        - `token`: the token sent to the user's email address.
    - **Note**: After the email is confirmed, the json response after logging in with username/password will contain `isEmailConfirmed` field set to `true`. We should probably surface that on the `User` object.

Implementation/testing notes: This depends on https://github.com/realm/realm-object-server-private/pull/920 being merged and released.

nhachicha:
## What, How & Why?

This closes #1699

* [x] jsdoc files updated

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -271,6 +271,54 @@ class User {
     static register(server, username, password, callback) {}
 
     /**
+     * Request a password reset email to be sent to a user's email.
+     * This will not throw an exception, even if the email doesn't belong to a Realm Object Server user.
+     * 
+     * This can only be used for users who authenticated with the 'password' provider, and passed a valid email address as a username.
+     * 
+     * @param {string} server - authentication server
+     * @param {string} email - The email that corresponds to the user's username.
+     * @return {Promise<void>} A promise which is resolved when the request has been sent.
+     */
+    static requestPasswordReset(server, email) {}
+    
+    /**
+     * Complete the password reset flow by using the reset token sent to the user's email as a one-time authorization token to change the password.
+     * 
+     * By default, the link that will be sent to the user's email will redirect to a webpage where, they can enter their new password.
+     * If you wish to provide a native UX, you may wish to modify the url to use deep linking to open the app, extract the token,
+     * and navigate to a view that allows them to change their password within the app.
+     * 
+     * @param {string} server - authentication server
+     * @param {string} reset_token - The token that was sent to the user's email address.
+     * @param {string} new_passwd - The user's new password.
+     * @return {Promise<void>} A promise which is resolved when the request has been sent.
+     */
+    static completePasswordReset(server, reset_token, new_passwd) {}
+
+    /**
+     * Request an email confirmation email to be sent to a user's email.
+     * This will not throw an exception, even if the email doesn't belong to a Realm Object Server user.
+     * 
+     * @param {string} server - authentication server
+     * @param {string} email - The email that corresponds to the user's username.
+     * @return {Promise<void>} A promise which is resolved when the request has been sent.
+     */
+    static requestEmailConfirmation(server, email) {}
+    
+    /**
+     * Complete the password reset flow by using the confirmation token sent to the user's email as a one-time authorization token to confirm their email.
+     * 
+     * By default, the link that will be sent to the user's email will redirect to a webpage where they'll see a generic "Thank you for confirming" text.
+     * If you wish to provide a native UX, you may wish to modify the url to use deep linking to open the app, extract the token, and inform them that their email has been confirmed.
+     * 
+     * @param {string} server - authentication server
+     * @param {string} confirmation_token - The token that was sent to the user's email address.
+     * @return {Promise<void>} A promise which is resolved when the request has been sent.
+     */
+    static confirmEmail(server, confirmation_token) {}
+
+    /**
      * Create an admin user for the given authentication server with an existing token
      * @param {string} adminToken - existing admin token
      * @param {string} server - authentication server
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -296,6 +296,14 @@ declare namespace Realm.Sync {
         static registerWithProvider(server: string, options: { provider: string, providerToken: string, userInfo: any }, callback: (error: Error | null, user: User | null) => void): void;
         static registerWithProvider(server: string, options: { provider: string, providerToken: string, userInfo: any }): Promise<Realm.Sync.User>;
 
+        static requestPasswordReset(server: string, email: string): Promise<Void>;
+
+        static completePasswordReset(server:string, reset_token:string, new_passwd:string): Promise<Void>;
+        
+        static requestEmailConfirmation(server:string, email:string): Promise<Void>;
+
+        static confirmEmail(server:string, confirmation_token:string): Promise<Void>;
+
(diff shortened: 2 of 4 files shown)


nhachicha:
cc @nirinchev

nhachicha:
@kneth
Unfortunately, we can not write integration tests, since we can not instrument the ROS instance to send/extract the tokens. However, I tested the endpoints locally, using a ROS server pre-configured to send emails to `ethereal.email` to capture tokens then manually use the various token to complete email confirmation or reset.

You can also use a cloud instance which has the SMPT configured

lib/user-methods.js
```diff
@@ -275,6 +275,32 @@ function _authenticate(userConstructor, server, json, callback) {
...
+function _updateAccount(userConstructor, server, json) {
+    const url = append_url(server, 'auth/password/updateAccount');
+    const options = {
+        method: 'POST',
+        body: JSON.stringify(json),
+        headers: postHeaders,
+        open_timeout: 5000
+    };
+
+    const promise = performFetch(url, options)
```
nirinchev: No need to assign a new local for that - you can simply return.

lib/user-methods.js
```diff
@@ -275,6 +275,32 @@ function _authenticate(userConstructor, server, json, callback) {
...
+        .then((response) => {
+            const contentType = response.headers.get('Content-Type');
+            if (contentType.indexOf('application/json') === -1) {
+                return response.text().then((body) => {
+                    throw new AuthError({
```
nirinchev: Why are you throwing in this branch, but returning a rejected promise in the other?

lib/user-methods.js
```diff
@@ -275,6 +275,32 @@ function _authenticate(userConstructor, server, json, callback) {
...
+                        title: `Could not update user account: Realm Object Server didn't respond with valid JSON`,
+                        body,
+                    });
+                });
+            } else if (!response.ok) {
```
nirinchev: The `else` is redundant here as there is a `return` in the previous `if` statement.

lib/index.d.ts
```diff
@@ -296,6 +296,14 @@ declare namespace Realm.Sync {
         static registerWithProvider(server: string, options: { provider: string, providerToken: string, userInfo: any }, callback: (error: Error | null, user: User | null) => void): void;
         static registerWithProvider(server: string, options: { provider: string, providerToken: string, userInfo: any }): Promise<Realm.Sync.User>;
 
+        static requestPasswordReset(server: string, email: string): Promise<Void>;
+
+        static completePasswordReset(server:string, reset_token:string, new_passwd:string): Promise<Void>;
```
nirinchev: Why `new_passwd`? It's not hard to guess what it means, but still, don't see much value in saving two characters. Also, it's inconsistent with other methods that have `password` spelled out correctly.

lib/user-methods.js
nhachicha: hmm, I tried to follow the same logic as in https://github.com/realm/realm-js/blob/d0b0adabc92bc9da2da9d881a5d099a8ec4596ce/lib/user-methods.js#L243

lib/user-methods.js
nhachicha: ditto

lib/user-methods.js
nirinchev: In `_authenticate`, the promise is assigned to a local because there may be a callback that needs to get invoked. In the current case, there's no such thing, so there's no point to assign it just to return it on the next line :)

docs/sync.js
```diff
@@ -270,6 +270,54 @@ class User {
...
+     * @param {string} email - The email that corresponds to the user's username.
+     * @return {Promise<void>} A promise which is resolved when the request has been sent.
+     */
+    static requestPasswordReset(server, email) {}
+    
+    /**
+     * Complete the password reset flow by using the reset token sent to the user's email as a one-time authorization token to change the password.
+     * 
+     * By default, the link that will be sent to the user's email will redirect to a webpage where, they can enter their new password.
+     * If you wish to provide a native UX, you may wish to modify the url to use deep linking to open the app, extract the token,
```
tgoyne: How? There's no parameter here that allows customizing the url so I'm guessing it's configured somewhere else, and this should say where that somewhere else it.

docs/sync.js
```diff
@@ -270,6 +270,54 @@ class User {
...
+     * This will not throw an exception, even if the email doesn't belong to a Realm Object Server user.
+     * 
+     * @param {string} server - authentication server
+     * @param {string} email - The email that corresponds to the user's username.
+     * @return {Promise<void>} A promise which is resolved when the request has been sent.
+     */
+    static requestEmailConfirmation(server, email) {}
+    
+    /**
+     * Complete the password reset flow by using the confirmation token sent to the user's email as a one-time authorization token to confirm their email.
```
tgoyne: I think the first half of this sentence is wrong?

lib/user-methods.js
```diff
@@ -379,6 +405,53 @@ const staticMethods = {
...
+
+            return _updateAccount(this, server, json);
+        },
+
+        requestEmailConfirmation(server, email) {
+            checkTypes(arguments, ['string', 'string']);
+            const json = {                
+                data: { 
+                    action: 'request_email_confirmation',
+                    provider_id: email
```
tgoyne: I think this field is outside `data`?

nhachicha: Yes good catch

PR #1761, opened 2018-04-23
Add a certificate validation using Android Keystore for RN

### LINKED ISSUE #1743: [Bugs] [react-native] [android] Sync config error callback spam call with code 336134278
## Bugs: Sync config error callback spam call with code 336134278

**realm 2.2.15
react-native: 0.51.0
react-object-server: 2**

similar any sync config of realm js. We build a config object and pass it to constructor of Realm when our users login. 

```
function realmErrorHandle(obj, options) {
  console.log('=======> REALM ERR!', options);
}

  const syncConfig = {
      sync: {
        user,
        url: realmUrl(),
        error: realmErrorHandle,
      },
      path: dbLocation(userID),
      ...currentSchema,
    }

realmDB = new Realm(syncConfig)
```
**note: all parameter being ensure correct and work well on Ios version**

## Expected Results
  realmErrorHandle never being called and everything work expected.

## Actual Results
Realm still working but realmErrorHandle be call very spamly (maybe one times per second) and does not seem to will stop.

and describe following Error: 
```
{   
  name: 'Error',
  message: 'Unknown OpenSSL error: 336134278',
  isFatal: false,
  category: 'openssl',
  code: 336134278,
  userInfo: {} 
}
```
**the code of error 336134278 can not be find in any Realm documents or OpenSSL documents.**  

We use SSL for realm object server but realm-js client worked expected without any SSL config on Ios. But unlucky, we met this error on Android.

The lastest version which dont have this error is 2.0.13

## Steps to Reproduce
~~Happen when we register new realm user with ```Realm.Sync.User.registerWithProvider```. 
When use available user realmErrorHandle hasn't be call.~~

Update: This happen when first time launch application. In debug mode, If reload by double tap R, no log more!

  

## Update:
**We just upgrade realm js version to 2.2.15 but can not resolve this problem**

- Realm JS SDK Version: **2.2.15**
- Node or React Native: **React Native 0.51.0**
- Client OS & Version: **Mac OS 10.12.6**
- Which debugger for React Native:  **None**

nhachicha:
## What, How & Why?

This closes #1743 & https://github.com/realm/realm-js-private/issues/428

This reuse the current Realm-Java certificate validation method in order to let the `open_ssl_verify_callback` to validate the chain certificates using the Android Keystore

a first attempt was made to implement this using a `@ReactMethod` so 
```
Sync client (C++)  --> open_ssl_verify_callback (JS) --> React Native Bridge --> Java react method 
```
However since the React native bridge is asynchronous we can not return immediately to the Sync Client (C++) the result (The Java React Method can return a promise, but this need to change the code using  `JavaScriptCore` in C++ to work with promise when invoking the `open_ssl_verify_callback`.

The solution here is to cut the middle man and call the Java method directly from the Sync client using JNI (which is faster compared to the initial approach (no bridge/marshalling involved)) 

we end up with 
```
Sync client (C++)  --> JNI --> Java react method 
```

### DIFF
--- a/react-native/android/src/main/java/io/realm/react/util/SSLHelper.java
+++ b/react-native/android/src/main/java/io/realm/react/util/SSLHelper.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.realm.react.util;
 
 import android.util.Log;
--- a/src/android/jni_utils.cpp
+++ b/src/android/jni_utils.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Realm Inc.
+ * Copyright 2018 Realm Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/android/jni_utils.hpp
+++ b/src/android/jni_utils.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Realm Inc.
+ * Copyright 2018 Realm Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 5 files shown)


src/android/jni_utils.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2017 Realm Inc.
```
kneth: 2018?

src/android/jni_utils.hpp
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2017 Realm Inc.
```
kneth: 2018

react-native/android/src/main/java/io/realm/react/util/SSLHelper.java
```diff
@@ -0,0 +1,136 @@
+package io.realm.react.util;
```
kneth: Please add standard header.

bmunkholm: @kneth When it's not code changes, I would recommend the reviewer to just fix the issues versus waiting for the submitter to get it done.

PR #1764, opened 2018-04-24
Don't hang forever when the React tests fail to start

tgoyne:
This doesn't actually fix the tests not working, but does at least make it so that it doesn't sit there with simctl using 100% CPU until someone manually kills the job.

### DIFF
--- a/lib/notification-worker.js
+++ b/lib/notification-worker.js
@@ -25,7 +25,7 @@ function nodeRequire(module) {
     return require_method(module);
 }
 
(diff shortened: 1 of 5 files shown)


tgoyne:
Well this wasn't *supposed* to fix the tests, but it seems to have made some of them work and the hanging one is at least running some of the tests before hanging?

tgoyne:
I'm not sure that I believe that macos_react_test_release actually passed considering it took over an hour to run.

PR #1780, opened 2018-05-04
Track Realms seen in the notifier by id rather than path

tgoyne:
If a Realm is deleted and then a new one is created at the same path it needs to be treated as a new Realm.

Required for https://github.com/realm/realm-js-private/pull/443 and should have no impact outside of that.


bmunkholm:
@kneth Please remember to check for changelog.md updates :-)
That's missing

kneth:
@bmunkholm I typically copy over entries from `realm-js-private` when doing a release.

PR #1789, opened 2018-05-11
Add support for the 'delete' event

tgoyne:
See https://github.com/realm/realm-js-private/pull/445.


kneth:
I will add an entry to `CHANGELOG.md` (based on https://github.com/realm/realm-js-private/pull/445) when releasing.

PR #1790, opened 2018-05-14
Group SSL configuration options in a new config object

### LINKED ISSUE #1465: Sync SSL configuration don't feels JavaScript'ish
## Goals

I want a more native feeling when configuring Realm JS.

## Expected Results

I would expect all parameters that configure sync use of SSL have parameters that are:

1. Short
2. camelCased
3. Grouped by their context:
    - Either they all start with "ssl": `sslValidate`, `sslCertificatePath`, `sslValidateCallback`,
    - Or (even better) they are all fields on a shared object named "ssl":
    ```
    {
      sync: {
        ssl: {
          validate: true,
          validateCallback: () => true,
          certificatePath: '../../'
        }
      }
    }
    ```

While we're modifying the interface I would highly recommend that the `open_ssl_verify_callback` gets called with a single object instead of 5 individual arguments. If this is already the case - the type definitions are wrong and should be updated.

Also - is there any reason for the difference in wording "validate" and "verify"? Could we pick one and stick with it?

## Actual Results

The configuration for sync SSL uses long underscore-cased names for:

- `validate_ssl`
- `ssl_trust_certificate_path`
- `open_ssl_verify_callback`

- Realm JS SDK Version: 2.0.2

fealebenpae:
Fixes #1465

This change groups together the SSL options of the Sync configuration object in a new object, making the API feel more JavaScript-y.

The code that parses the new SSL configuration format is refactored so that the Adapter can use it as well.

I also refactored the JSDoc for the Sync configuration API. Please take a look.


ianpward:
WIll this solve the issue of using realm-js with SSL on Windows?

fealebenpae:
This and https://github.com/realm/realm-js-private/pull/446 will solve https://github.com/realm/realm-js-private/issues/438.

kneth:
@fealebenpae I like the refactoring of the doc - thanks!

kraenhansen:
🙏 thanks! This was an awesome solution. Great work!

Issue #1748, opened 2018-04-12
`writeCopyTo` on Linux


kneth:
`writeCopyTo` has been recently added. I have observed that the encryption part of the test (https://github.com/realm/realm-js/blob/master/tests/js/realm-tests.js#L1208) fails on Linux (but only on Linux). This must be investigated!

kneth:
#2020 fixes the issues and it will be released soon.

PR #1792, opened 2018-05-15
Add Object.linkingObjectCount()

astigsen:
## What, How & Why?

This adds `Object.linkingObjectCount()` method to get the total number of incoming links to an object.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/object.js
+++ b/docs/object.js
@@ -52,5 +52,5 @@ class Object {
      * @returns {number} number of links to this object.
      * @since 2.6.0
      */
-    linkingObjectsCount(objectType, property) {}
+    linkingObjectsCount() {}
 }
--- a/docs/object.js
+++ b/docs/object.js
@@ -46,4 +46,11 @@ class Object {
      * @since 1.9.0
      */
     linkingObjects(objectType, property) {}
+
+    /**
+     * Returns the total count of incoming links to this object
+     * @returns {number} number of links to this object.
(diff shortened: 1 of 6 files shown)


astigsen:
I also thought of naming it `backlinksCount()`, but in the end I found it more correct to align the naming with the other method on the object which is called `linkingObjects()` (in .NET this looks to be named `GetBacklinks()`, so there it would match up).

docs/object.js
```diff
@@ -46,4 +46,11 @@ class Object {
      * @since 1.9.0
      */
     linkingObjects(objectType, property) {}
+
+    /**
+     * Returns the total count of incoming links to this object
+     * @returns {number} number of links to this object.
+     * @since 2.6.0
+     */
+    linkingObjectsCount(objectType, property) {}
```
kneth: The parameter should be documented too. They can be copied from the method above.

kneth:
One - unrelated - test is failing in one build. I have seen it before, and I haven't a clear understanding of why it fails. I believe we can ignore it for now (see also #1748).

docs/object.js
astigsen: Yes, it should not take any parameters. It was a copy/paste mistake. Fixed.

Issue #1821, opened 2018-05-27
Reduce Runtime Dependencies


heri16:
This npm package seems to have lots of dependency bloat that could either be removed or moved into peerDependencies.

Why the need for both request and node-fetch?
Command-line-args is needed for runtime? How so?
Includes an unsafe package like sync-request? Really?

```
  "dependencies": {
    "command-line-args": "^4.0.6",
    "decompress": "^4.2.0",
    "fs-extra": "^4.0.2",
    "ini": "^1.3.4",
    "nan": "2.8.0",
    "node-fetch": "^1.6.3",
    "node-pre-gyp": "^0.6.36",
    "progress": "^2.0.0",
    "prop-types": "^15.5.10",
    "request": "^2.78.0",
    "stream-counter": "^1.0.0",
    "sync-request": "^3.0.1",
    "url-parse": "^1.2.0"`
```

kneth:
@heri16 Thanks for your PRs. I plan to get to them by the end of the week. Please sign our [CLA](https://github.com/realm/realm-js/blob/master/CONTRIBUTING.md#cla).

PR #1822, opened 2018-05-28
Drop request package dependency

heri16:
## What, How & Why?
Drop unnecessary request package dependency and switched to node-fetch.
See Issue: #1821


kneth:
CLA has not been signed.

PR #1823, opened 2018-05-28
Moved sync-request package into devDependencies

heri16:
Moved sync-request package into devDependencies.
See Issue: #1821


kneth:
CLA has not been signed.

PR #1824, opened 2018-05-28
Moved prop-types into peerDependencies

heri16:
See Issue #1821


kneth:
CLA has not been signed.

PR #1828, opened 2018-05-28
Split Install-deps from Runtime-deps

heri16:
Move packages only required in `scripts/download-realm.js` to the preinstall hook in `package.json`.
This is more idiomatic way so as not to pollute the runtime-dependencies that are normally considered by module bundlers such as browserify/webpack.

Running rebuilds from source should still work.


kneth:
CLA has not been signed.

PR #1830, opened 2018-05-28
Query-based sync as the default sync mode

### LINKED ISSUE #1793: Make Query-based Sync the default mode for syncing data
All bindings need to adopt Query-based synchronization (formerly named Partial Sync) as the default mode of synching. This is most likely a breaking API change which requires a major API bump.

Java merged their PR here: realm/realm-java#5893

In particular the setting on SyncConfiguration was renamed from partialRealm() to fullSynchronization().

- [ ] Change API
- [ ] Update documentation to use Query-based Sync instead of Partial Sync

cmelchior:
Closes #1793 

As discussed, this PR add new API's for creating a SyncConfiguration with Query-based Sync as the default setting.

STILL IN PROGRESS

TODO
* [x] Verify with @kneth / @kraenhansen this the correct public API we want
* [x] Convert tests to use new API
* [x] Verify that public Docs are correct


src/js_sync.hpp
```diff
@@ -833,19 +833,26 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
...
+
         if (!Value::is_undefined(ctx, partial_value)) {
             is_partial = Value::validated_to_boolean(ctx, partial_value);
         }
 
-        bool disable_partial_sync_url_checks = false;
-        ValueType disable_partial_sync_url_checks_value = Object::get_property(ctx, sync_config_object, "_disablePartialSyncUrlChecks");
-        if (!Value::is_undefined(ctx, disable_partial_sync_url_checks_value)) {
-            disable_partial_sync_url_checks = Value::validated_to_boolean(ctx, disable_partial_sync_url_checks_value);
+        // In the deprecation period setting full_synchronization takes precedence over partial
```
kneth: I would rather throw an exception.

docs/sync.js
```diff
@@ -35,10 +35,14 @@
...
- *    Partial synchronisation only synchronizes those objects that match the query specified in contrast
+ * @deprecated
+ * @property {boolean} [partial] - Whether this Realm should be opened in 'query-based synchronization' mode.
+ *    Query-based synchronisation only synchronizes those objects that match the query specified in contrast
  *    to the normal mode of operation that synchronises all objects in a remote Realm.
- *    **Partial synchronization is a tech preview. Its APIs are subject to change.**
+ * @property {boolean} [full_synchronization] - Whether this Realm should be opened in query-based or full
+ *    synchronization mode. The default is query-based mode which only synchronizes objects that have been subscribed to.
+ *    A fully synchronized Realm will synchronize the entire Realm in the background, irrespectively of the data being
+ *    used or not.
```
kraenhansen: Are `partial` and `full_synchronization` mutually exclusive? If so - we should prefer one over the other by warning when both are set. And perhaps we should add a `@deprecated` annotation on `partial`?

src/js_sync.hpp
kraenhansen: I agree if both `partial` and `full_synchronization` is sat by the user - that should throw.

docs/sync.js
kraenhansen: JavaScript land prefers camelCasing - I believe this should be renamed to `fullSynchronization`.

src/js_sync.hpp
cmelchior: You mean throw an exception if both `partial` and `fullSynchronization` is set? That works for me as well

lib/index.d.ts
```diff
@@ -83,11 +83,19 @@ declare namespace Realm {
         inMemory?: boolean;
         schema?: (ObjectClass | ObjectSchema)[];
         schemaVersion?: number;
-        sync?: Realm.Sync.SyncConfiguration;
+        sync?: Partial<Realm.Sync.SyncConfiguration>;
```
kraenhansen: When you have `PartialConfiguration` 👇 I don't see the need for this to be `Partial` too?

lib/index.d.ts
```diff
@@ -410,7 +419,8 @@ declare namespace Realm.Sync {
         ssl?: SSLConfiguration;
         error?: ErrorCallback;
         partial?: boolean;
-        _disablePartialSyncUrlChecks?:boolean;
+        fullSynchronization?: boolean;
```
kraenhansen: I don't understand why it's `full_synchronization` elsewhere and `fullSynchronization` here - is that a bug? I prefer the latter of the two.

cmelchior: The JS name should be `fullSynchronization`, in C++ that turns into `full_synchronization`. If that isn't the case it is a bug.

docs/sync.js
cmelchior: `partial` was marked deprecated here: https://github.com/realm/realm-js/pull/1830/files#diff-b6a6b946756f354e3a99acfa65414e2eR38

that the variable is named `full_synchronization` is a bug.

CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
+3.0.0 Release notes (YYYY-MM-DD)
+=============================================================
+### Compatibility
+* Sync protocol: 24
+* Server-side history format: 4
+* File format: 7
+* Realm Object Server: 3.0.0 or later
+
+The feature known as Partial synchronization has been renamed to Query-based synchronization and is now the default mode
```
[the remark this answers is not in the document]
cmelchior: Fixed

lib/index.d.ts
```diff
@@ -83,11 +83,19 @@ declare namespace Realm {
...
         schemaVersion?: number;
-        sync?: Realm.Sync.SyncConfiguration;
+        sync?: Partial<Realm.Sync.SyncConfiguration>;
         deleteRealmIfMigrationNeeded?: boolean;
         disableFormatUpgrade?: boolean;
     }
 
+    /**
+     * realm configuration used for overriding default configuration values.
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.html#~Configuration }
```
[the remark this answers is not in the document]
cmelchior: All other links in this document are formatted with whitespace, so I would prefer to keep the same style.

CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
...
+for synchronized Realms. This has impacted a number of APIs. See below for the details.
+
+### Breaking changes
+
+* [Sync] `Realm.Configuration.SyncConfig._disablePartialSyncUrlChecks` has been renamed to `Realm.Configuration.sync._disableQueryBasedSyncUrlChecks`.
```
kneth: This is an internal/private setting so I will not consider it breaking. I think it can be move to the *Internal* section.

lib/index.d.ts
```diff
@@ -605,6 +612,7 @@ declare class Realm {
     static openAsync(config: Realm.Configuration, callback: (error: any, realm: Realm) => void, progressCallback?: Realm.Sync.ProgressNotificationCallback): void
 
     /**
+     * @deprecated in favor of `Realm.Sync.User.createConfiguration()`.
```
kneth: Using `{@link Realm.Sync.User#createConfiguration}` will create a link.

lib/user-methods.js
```diff
@@ -520,6 +522,42 @@ const instanceMethods = {
...
                 });
         },
+        createConfiguration(config) {
+
+            if (config && config.sync) {
+                if (console.warn !== undefined) {
+                    console.warn(`'user' property will be overridden by ${this.identity}`);
+                }
+                if (config.sync.partial !== undefined && config.sync.fullSynchronization !== undefined) {
+                    throw new Error("'partial' and 'fullSynchronization' was both set. 'partial' has been deprecated, use only 'fullSynchronization'");
```
kneth: "was" -> "were"

tests/js/user-tests.js
```diff
@@ -318,6 +319,58 @@ module.exports = {
...
+          TestCase.assertThrowsContaining(() => {
+                  let config = { 
+                    sync: { 
+                      url: 'http://localhost:9080/~/default',
+                      partial: true, 
+                      fullSynchronization: false 
+                    } 
+                  };
+                  user.createConfiguration(config);
+          }, "'partial' and 'fullSynchronization' was both set. 'partial' has been deprecated, use only 'fullSynchronization'");
```
kneth: "was" -> "were"

tests/js/user-tests.js
```diff
@@ -318,6 +319,58 @@ module.exports = {
...
+          TestCase.assertThrowsContaining(() => {
+              new Realm({
+                  sync: {
+                    user: user,
+                    url: 'http://localhost:9080/~/default',
+                    partial: false,
+                    fullSynchronization: true
+                  }
+              })
+          }, "'partial' and 'fullSynchronization' was both set. 'partial' has been deprecated, use only 'fullSynchronization'");
```
kneth: "was" -> "were"

PR #1831, opened 2018-05-28
React to schema changes

### LINKED ISSUE #1825: Update schema when the schema on disk changes
When the schema on disk changes (e.g. due to sync adding new classes/properties), realm-js should update the `schema` property on the Realm, even if it is open. This is needed for the GraphQL service which caches realms to avoid opening them for every request.

kneth:
## What, How & Why?

This closes #1825

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -209,8 +209,8 @@ class Realm {
     /**
      * Add a listener `callback` for the specified event `name`.
      * @param {string} name - The name of event that should cause the callback to be called.
-     *   _Currently, only the "change" event supported_.
-     * @param {callback(Realm, string)} callback - Function to be called when the event occurs.
+     *   _Currently, only the "change" or "schema" events supported_.
+     * @param {callback(Realm, string)|callback(Realm, Schema)} callback - Function to be called when a change event occurs.
      *   Each callback will only be called once per event, regardless of the number of times
      *   it was added.
      * @throws {Error} If an invalid event `name` is supplied, or if `callback` is not a function.
@@ -220,8 +220,8 @@ class Realm {
    /**
     * Remove the listener `callback` for the specfied event `name`.
     * @param {string} name - The event name.
-    *   _Currently, only the "change" event supported_.
-    * @param {callback(Realm, string)} callback - Function that was previously added as a
+    *   _Currently, only the "change" or "schema" events supported_.
+    * @param {callback(Realm, string)|callback(Realm, Schema)} callback - Function that was previously added as a
     *   listener for this event through the {@link Realm#addListener addListener} method.
     * @throws {Error} If an invalid event `name` is supplied, or if `callback` is not a function.
     */
@@ -230,7 +230,7 @@ class Realm {
    /**
     * Remove all event listeners (restricted to the event `name`, if provided).
     * @param {string} [name] - The name of the event whose listeners should be removed.
-    *   _Currently, only the "change" event supported_.
+    *   _Currently, only the "change" or "schema" events supported_.
     * @throws {Error} When invalid event `name` is supplied
     */
     removeAllListeners(name) {}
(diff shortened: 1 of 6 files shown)


docs/realm.js
```diff
@@ -209,8 +209,8 @@ class Realm {
     /**
      * Add a listener `callback` for the specified event `name`.
      * @param {string} name - The name of event that should cause the callback to be called.
-     *   _Currently, only the "change" event supported_.
-     * @param {callback(Realm, string)} callback - Function to be called when the event occurs.
+     *   _Currently, only the "change" or "schema" events supported_.
```
tgoyne: "change" *and* "schema" events

(A bunch of these; not marking them all).

lib/index.d.ts
```diff
@@ -669,13 +669,15 @@ declare class Realm {
      * @returns void
      */
     addListener(name: string, callback: (sender: Realm, event: 'change') => void): void;
+    addListener(name: string, callback: (sender: Realm, schema: Realm.ObjectSchema[]) => void): void;
```
tgoyne: The callback should have an `event: 'schema'` parameter.

tests/js/realm-tests.js
```diff
@@ -1220,6 +1220,51 @@ module.exports = {
...
+                TestCase.assertEqual(realm1.schema.length, 1);
+                TestCase.assertEqual(realm.schema[0].name, 'TestObject');
+
+                if (calls == 2) {
+                    resolve();
+                }
+            });
+
+            realm1.addListener('change', (realm, event) => {
+                calls++;
```
nirinchev: This is minor, but the way this test is structured, if for some reason we get two notifications for changes, this will pass the test, even though we didn't get any schema notifications. I would suggest using a hash (or an object) to store the types of event notifications we've received and only resolve when we've received both `change` and `schema`.

tests/js/realm-tests.js
```diff
@@ -1220,6 +1220,51 @@ module.exports = {
...
+            const schema = [{
+                name: 'TestObject',
+                properties: {
+                    prop0: 'string',
+                }
+            }];
+            let realm2 = new Realm({ schema: schema, _cache: false });
+            TestCase.assertEqual(realm1.schema.length, 0); // not yet updated
+            TestCase.assertEqual(realm2.schema.length, 1);
+            realm2.write(() => {
```
nirinchev: This also has the potential of giving us false positives in the case where there's a regression and the schema notifications only get triggered after a write transaction is committed to the table that has been added/changed.

Additionally, it would be nice to test that notifications are delivered both when a new class is added to the schema and when an existing class is updated with a new property.

docs/realm.js
```diff
@@ -209,8 +209,8 @@ class Realm {
     /**
      * Add a listener `callback` for the specified event `name`.
      * @param {string} name - The name of event that should cause the callback to be called.
-     *   _Currently, only the "change" event supported_.
-     * @param {callback(Realm, string)} callback - Function to be called when the event occurs.
+     *   _Currently, only the "change" and "schema" events supported_.
```
bmunkholm: missing "are" after "events".

docs/realm.js
```diff
@@ -220,8 +220,8 @@ class Realm {
    /**
     * Remove the listener `callback` for the specfied event `name`.
     * @param {string} name - The event name.
-    *   _Currently, only the "change" event supported_.
-    * @param {callback(Realm, string)} callback - Function that was previously added as a
+    *   _Currently, only the "change" and "schema" events supported_.
```
bmunkholm: missing "are" after "events".

docs/realm.js
```diff
@@ -230,7 +230,7 @@ class Realm {
    /**
     * Remove all event listeners (restricted to the event `name`, if provided).
     * @param {string} [name] - The name of the event whose listeners should be removed.
-    *   _Currently, only the "change" event supported_.
+    *   _Currently, only the "change" and "schema" events supported_.
```
bmunkholm: missing "are" after "events".

PR #1849, opened 2018-06-06
Expose both newModifications and oldModifications on change events

tgoyne:
I added this to the 2.8.0 changelog section as that appears to have not been released yet.

### DIFF
--- a/docs/collection.js
+++ b/docs/collection.js
@@ -427,9 +427,12 @@ class Collection {
      * @param {function(collection, changes)} callback - A function to be called when changes occur.
      *   The callback function is called with two arguments:
      *   - `collection`: the collection instance that changed,
-     *   - `changes`: a dictionary with keys `insertions`, `modifications` and `deletions`,
-     *      each containing a list of indices that were inserted, updated or deleted respectively. If
-     *      partial sync is enabled, an additional key `partial_sync` is added.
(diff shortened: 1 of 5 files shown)


docs/collection.js
```diff
@@ -427,9 +427,12 @@ class Collection {
...
      *   The callback function is called with two arguments:
      *   - `collection`: the collection instance that changed,
-     *   - `changes`: a dictionary with keys `insertions`, `modifications` and `deletions`,
-     *      each containing a list of indices that were inserted, updated or deleted respectively. If
-     *      partial sync is enabled, an additional key `partial_sync` is added.
+     *   - `changes`: a dictionary with keys `insertions`, `newModifications`, `oldModifications`
+     *      and `deletions`, each containing a list of indices in the collection that were
+     *      inserted, updated or deleted respectively. `deletions` and `oldModifications` are
+     *      indices into the collection before the change happened, while `insertions` and
+     *      `newModifications` are indices into the new version of the collection. If partial sync
```
kneth: I suggest that we use `query_based_sync` instead of `partial_sync`.

kneth: Btw, we still have `modifications` so we should document them.

kneth: Looks like I forgot to clean up a PR back in February. The reference to partial sync can be removed.

PR #1852, opened 2018-06-08
Better representation of array of primitive types for  (#1847).

### LINKED ISSUE #1847: Cannot open a realm with a schema read from another Realm with list of a primitive type
## Goals

I want to:

- Open a Realm which schemas contain a list of primitives,
- Read and hang onto its schemas,
- Close the Realm,
- Open another (or the same Realm) with the schemas.

## Expected Results

I expect to be able to read a realms schema and use that to open another Realm.

## Actual Results

If for example I have a schema named "Person" with a list of string `names`, the schema read from the realm will be:

```
{
  name: "Person"
  properties: {
    names: {
      indexed: false,
      name: "names",
      objectType: "string",
      optional: false,
      type: "list"
    }
  }
}
```

Which will yeild the following error when opening the Realm:

> Schema validation failed due to the following errors:
> - Property 'Person.names' of type 'array' has unknown object type 'string'

I believe the underlying issue is that the schema does not recognize primitives as a valid `objectType` https://github.com/realm/realm-js/blob/master/src/js_schema.hpp#L107-L115.

See https://runkit.com/kraenhansen/realm-list-of-primitives-failure

- Realm JS SDK Version: 2.2.8
- Node or React Native: Node
- Client OS & Version: N/A

kneth:
## What, How & Why?

This closes #1847 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


src/js_schema.hpp
```diff
@@ -338,7 +339,17 @@ typename T::Object Schema<T>::object_for_property(ContextType ctx, const Propert
         Object::set_property(ctx, object, object_type_string, Value::from_string(ctx, property.object_type));
     }
     else if (is_array(property.type)) {
-        Object::set_property(ctx, object, object_type_string, Value::from_string(ctx, string_for_property_type(property.type & ~realm::PropertyType::Flags)));
+        auto tmp = string_for_property_type(property.type & ~realm::PropertyType::Flags);
+        if (strncmp(tmp, "string", 6) || strncmp(tmp, "int", 3) || strncmp(tmp, "bool", 4)
```
tgoyne: Isn't this just a long-winded way of saying `if (property.type != realm::PropertyType::Object)`?

PR #1873, opened 2018-06-18
Allow setting shouldCompactOnLaunch on synchronized Realms

tgoyne:
This is now supported by core.

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -518,10 +518,10 @@ void RealmClass<T>::constructor(ContextType ctx, ObjectType this_object, size_t 
                 }
 
                 FunctionType should_compact_on_launch_function = Value::validated_to_function(ctx, compact_value, "shouldCompactOnLaunch");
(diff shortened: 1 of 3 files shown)


src/js_realm.hpp
```diff
@@ -571,6 +568,12 @@ void RealmClass<T>::constructor(ContextType ctx, ObjectType this_object, size_t
                 config.cache = Value::validated_to_boolean(ctx, cache_value, "_cache");
             }
 
+            static const String automatic_change_notifications_string = "_automaticChangeNotifications";
+            ValueType automatic_change_notifications_value = Object::get_property(ctx, object, automatic_change_notifications_string);
+            if (!Value::is_undefined(ctx, automatic_change_notifications_value)) {
+                config.automatic_change_notifications = Value::validated_to_boolean(ctx, automatic_change_notifications_value, "_automaticChangeNotifications");
```
bmunkholm: Any reason not to use the constant you made (automatic_change_notifications_string)?

tgoyne: It's just a copy and paste from the code above to read other properties, and I'm not sure why the existing code does things the way it does.

PR #1874, opened 2018-06-18
Add support for creating template objects

### LINKED ISSUE #1870: Realm.createTemplateObject() for creating object with all default values
**Use case:**
I want to create a Realm object that has many required fields (50+). Even though I just want all these fields to contain the default value for the type: `0`, `""` and `false`. Currently, I am required to manually fill out all these fields since Realm does not allow these fields to be undefined.

```
realm.createObject('Person', {
 field1: false,
 field2: false,
 field3: false,
 field4: false,
 field5: false,
 // ... many more fields.
});
```

**Wanted:**
Instead, it would be nice if Realm could provide a method for creating an object with all default values. Something like:

```
let template = Realm.createTemplateObject(objectSchema);

// Create object with all defaults
realm.create(objectSchema.name, template);

// Create an object with all defaults, but a select few overrides
realm.create(objectSchema.name, Object.assign(template, { field1: true });
```

cmelchior:
Closes #1870


bmunkholm:
Shouldn't we create this as a managed object instead? Or is there a special need for an unmanaged object?
And then use the default values defined for the schema?

cmelchior:
No, the object should be unmanaged. The idea is make it easy to create objects with large amounts of required fields, but you really only want to override a few.

I ran into this when testing schemas with close to 20-30 required fields where most where booleans that I didn't care about.

The pattern will look something like this:
```
let template = Realm.createTemplateObject(schema);

// Override just the fields I care about, all other required fields will have the default value.
realm.create(schema.name, Object.assign(template, { col: true});
```

The reason we want it to be explicit is that otherwise, you might miss it if you spell a property wrong (and the real one then get set to the default value).

PR #1882, opened 2018-06-20
Compute realm size

### LINKED ISSUE #1881: Expose `calculate_aggregated_byte_size` to get the internal size of a Realm file
## Goals

I would like the Realm Core method `Group::compute_aggregated_byte_size()` exposed on an open Realm so the Realm Object Server can calculate the aggregated state size of Realms on demand.

Bonus points for an asynchronous implementation that doesn't block the main thread.

## Expected Results

After opening a Realm I would expect a method like `.computeSize()` which returned the size of the state of the Realm, in bytes.

- Realm JS SDK Version: v2.10.0
- Node or React Native: Node
- Client OS & Version: N/A

kraenhansen:
## What, How & Why?

This adds a `computeSize` method on the Realm object.

This closes #1881.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+X.Y.Z Release notes
+=============================================================
+### Compatibility
+* Sync protocol: 24
+* Server-side history format: 4
+* File format: 7
+* Realm Object Server: 3.0.0 or later
+
+### Breaking changes
+* None.
+
+### Enhancements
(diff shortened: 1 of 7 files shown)


kraenhansen:
We realized half way through the implementation of this that core includes the history and state of the current "snapshot" (I believe that was the word Kristian used). Hence https://github.com/realm/realm-core/issues/3063 was created.

PR #1884, opened 2018-06-21
Improve performance of `realm.create()` when running in node

tgoyne:
In total these changes cut the runtime of it by ~25% on the test case I've been measuring.

### DIFF
--- a/src/js_object_accessor.hpp
+++ b/src/js_object_accessor.hpp
@@ -68,10 +68,10 @@ public:
 
     OptionalValue value_for_property(ValueType dict, std::string const& prop_name, size_t prop_index) {
         ObjectType object = Value::validated_to_object(m_ctx, dict);
-        if (!Object::has_property(m_ctx, object, prop_name)) {
+        ValueType value = Object::get_property(m_ctx, object, prop_name);
+        if (Value::is_undefined(m_ctx, value)) {
             return util::none;
         }
-        ValueType value = Object::get_property(m_ctx, object, prop_name);
         const auto& prop = m_object_schema->persisted_properties[prop_index];
         if (!Value::is_valid_for_property(m_ctx, value, prop)) {
             throw TypeErrorException(*this, m_object_schema->name, prop, value);
--- a/src/node/node_string.hpp
+++ b/src/node/node_string.hpp
@@ -30,16 +30,35 @@ class String<node::Types> {
   public:
     String(const char* s) : m_str(s) {}
     String(const std::string &s) : m_str(s) {}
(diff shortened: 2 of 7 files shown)


src/js_object_accessor.hpp
```diff
@@ -68,10 +68,10 @@ class NativeAccessor {
 
     OptionalValue value_for_property(ValueType dict, std::string const& prop_name, size_t prop_index) {
         ObjectType object = Value::validated_to_object(m_ctx, dict);
-        if (!Object::has_property(m_ctx, object, prop_name)) {
+        ValueType value = Object::get_property(m_ctx, object, prop_name);
```
kneth: We could easily apply this pattern many other places 😄

src/node/node_string.hpp
```diff
@@ -30,16 +30,35 @@ class String<node::Types> {
...
     operator v8::Local<v8::String>() const {
         return Nan::New(m_str).ToLocalChecked();
     }
 };
 
+inline String<node::Types>::String(const v8::Local<v8::String> &s) {
+    if (s.IsEmpty() || s->Length() == 0) {
+        return;
+    }
+    if (s->IsOneByte()) {
```
kneth: `ContainsOnlyOneByte` might be safer (but slower)

src/js_object_accessor.hpp
kneth: But now we cannot know if we have the property or the value in `undefined`.

src/node/node_string.hpp
tgoyne: No, it isn't safer. `IsOneByte()` checks the type of the string, while  `ContainsOnlyOneByte()` checks the type of the string, and then if it isn't one-byte then checks the contents of the string to see if it actually could be one-byte. As such, `IsOneByte()` has false negatives, not false positives.

Calling `Utf8Length()` after verifying it's the same thing as `Length()` is just a pointless pessimization.

src/js_object_accessor.hpp
tgoyne: Drawing a distinction between no value and `undefined` is pretty strange for JS to begin with, and most of the API already didn't. I don't think this is a behavioral change to any existing working code; it's purely a widening in what sorts of objects we accept to create().

src/node/node_string.hpp
tgoyne: Okay, I see why the test was failing without these changes: internal and external one-byte strings are actually different in v8, and the assumption that `Length() == Utf8Length()` only holds for external one-byte strings.

I think trying to do different things for one-byte and two-byte strings isn't worth it given this.

PR #1888, opened 2018-06-24
When forcing sync on a local Realm, use an additive schema

ironage:
Since this is the expected mode when opening a sync Realm. An alternative would be to expose the schema mode to allow users to set this manually.


bmunkholm:
Missing changlog.md update

kneth:
@bmunkholm This is an undocumented feature which was introduced to help Realm Studio. As such, change log shouldn't be updated.

PR #1898, opened 2018-06-27
Expose custom HTTP headers in sync configuration

### LINKED ISSUE #1897: Expose API to set custom headers on the sync config
It appears we can set a map of custom headers on the sync client - it will be beneficial if we exposed an API for ROS to do that when opening Realms internally.

kneth:
## What, How & Why?
Realm Object Store provides an API to specify custom HTTP headers when communicating with the sync server.

This closes #1897 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -427,7 +427,7 @@ declare namespace Realm.Sync {
         partial?: boolean;
         fullSynchronization?: boolean;
         _disableQueryBasedSyncUrlChecks?:boolean;
-        custom_http_headers?: any;
+        custom_http_headers?: { [header: string]: string };
     }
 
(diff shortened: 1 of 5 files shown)


lib/index.d.ts
```diff
@@ -427,6 +427,7 @@ declare namespace Realm.Sync {
         partial?: boolean;
         fullSynchronization?: boolean;
         _disableQueryBasedSyncUrlChecks?:boolean;
+        custom_http_headers?: any;
```
nirinchev: Shouldn't this be

```ts
{ [header: string]: string }
```

I don't suppose we can pass in a complex object.

kneth: True, your suggestion is more accurate.

PR #1903, opened 2018-06-29
Add object._setLink()

tgoyne:
`object._setLink('prop', pk)` is semantically equivalent to `object.prop = realm.objectForPrimaryKey('ClassName', pk);` but roughly twice as fast due to a lot fewer intermediate steps and temporary objects. This gives a 10-20% overall speedup to the second pass of the data loader.

I tried making this just happen for normal property assignments rather than a separate function (i.e. `obj.prop = pk`) but it made the whole property validation and conversion pipeline a lot more complicated, and while it's behavior that seems useful in more cases than just the data loader it might be too weird to be worth it.

### DIFF
--- a/src/js_realm_object.hpp
+++ b/src/js_realm_object.hpp
@@ -56,6 +56,7 @@ struct RealmObjectClass : ClassDefinition<T, realm::Object> {
     static void linking_objects_count(ContextType, FunctionType, ObjectType, size_t, const ValueType [], ReturnValue &);
     static void get_object_id(ContextType, ObjectType, Arguments, ReturnValue &);
     static void is_same_object(ContextType, ObjectType, Arguments, ReturnValue &);
+    static void set_link(ContextType, ObjectType, Arguments, ReturnValue &);
 
     const std::string name = "RealmObject";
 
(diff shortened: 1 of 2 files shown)


src/js_realm_object.hpp
```diff
@@ -140,6 +142,50 @@ bool RealmObjectClass<T>::set_property(ContextType ctx, ObjectType object, const
...
+                                           accessor.template unbox<util::Optional<int64_t>>(args[1]));
+    }
+    else {
+        row_ndx = linked_table->find_first(linked_pk->table_column,
+                                           accessor.template unbox<int64_t>(args[1]));
+    }
+
+    if (row_ndx != realm::not_found) {
+        realm_object->row().set_link(prop->table_column, row_ndx);
+    }
```
kneth: Should we throw an exception here?

kneth:
It appears that CI is failing at an unrelated test.

src/js_realm_object.hpp
tgoyne: The object not existing is valid (it may have been excluded by a filter), but it probably should set_null() in that case.

PR #1906, opened 2018-07-03
Upgrading to Realm Core 3.5.4 and Realm Sync 3.6.0

kneth:


bmunkholm:
@kneth The title looks very wrong in the versions :-) And a new version is on it's way today, so I suggest waiting for that?

PR #1937, opened 2018-07-24
Fix Node.js 10 support

fealebenpae:
## What, How & Why?
Node.js 10 brought us several API breaking changes in v8, most notably removing `FunctionCallbackInfo::Callee()`. This patch fixes most of the errors by using `Nan::` APIs in the place of `v8::` APIs that were removed after being deprecated (such as the `v8::Object::SetAccessor` overload that didn't require a `v8::Context`). However, `FunctionCallbackInfo::Callee()` has no substitute and its usage had to be removed. After doing that I refactored `ConstructorType`, `MethodType`, and `ArgumentsMethodType` by removing the variants that took arguments as `int argc, ValueType args[]` and left only the `js::Arguments<T> &args` variant.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [x] 🚦 Tests


kneth:
The rename `arguments` to `args` makes the PR a bit noisy :-)

But will we get node 10 support before https://github.com/JoshKaufman/ursa/pull/169 is released?

fealebenpae:
Perhaps not in ROS, but nothing in realm-js depends on Ursa AFAIK.

PR #1941, opened 2018-07-24
Disable GNU dialect when building for Node 10 with GCC

fealebenpae:


target_defaults.gypi
```diff
@@ -11,7 +11,8 @@
     "cflags_cc!": [ # turn off default flags on older nodes on linux
       "-fno-exceptions",
       "-fno-rtti",
-      "-std=gnu++0x"
+      "-std=gnu++0x",
```
kneth: Should this one be removed?

kneth: Sorry, I missed that it is the flags you turn off :-)

PR #1968, opened 2018-08-16
Improve handling of parallel CI jobs.

mandrigin:
## What, How & Why?

1. Sometimes, it is necessary to run multiple parallel CI jobs on the same physical machine w/o any virtualization or containerization (a good example of that is building an iOS app). To work properly in that environment, realm should download a temp file to it's own unique temp folder to avoid clashes (that are actually fail with `symlink` syscall failing with `EEXISTS`).

2. To avoid an ever-growing temp folder, a new environment variable was introduced, that you can set to be inside a job's folder. That way, CI will be responsible for cleaning up after itself and keeping only the necessary amount of files.

Theoretically, all of that can be solved by overriding `TMP` env
variable but in practice in many-many environments it breaks the other
part of toolchain (even Xcode builds themselves sometimes).

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/scripts/download-realm.js
+++ b/scripts/download-realm.js
@@ -25,8 +25,35 @@ const child_process = require('child_process');
 const fetch = require('node-fetch');
 const ini = require('ini');
 const decompress = require('decompress');
+const crypto = require('crypto');
 
 const LOCKFILE_NAME = 'download-realm.lock';
+
+const TEMP_DIR_SUFFIX = generateRandomString();
+
+function generateRandomString() {
+    return crypto.randomBytes(20).toString('hex');
+}
+
+function ensureDirExists(dir) {
+    if (!fs.existsSync(dir)) {
+        fs.mkdirSync(dir);
+    }
+}
+
+function getTempDir() {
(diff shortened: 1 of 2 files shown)


kneth:
@mandrigin Thanks for the PR.

mandrigin:
@kneth no problem! We were hunting down our seemingly random iOS CI failures for a while, the error message really didn't make any sense. I think it worth sharing to the broader community.

scripts/download-realm.js
```diff
@@ -25,9 +25,36 @@ const child_process = require('child_process');
...
 
+const TEMP_DIR_SUFFIX = generateRandomString();
+
+function generateRandomString() {
+    return crypto.randomBytes(20).toString('hex');
+}
+
+function ensureDirExists(dir) {
+    if (!fs.existsSync(dir)) {
+        fs.mkdir(dir);
```
cmelchior: Shouldn't this be `mkdirSync` otherwise it might not be ready exiting this method?

PR #1975, opened 2018-08-21
Update devtoolset from 3 to 6

emanuelez:
Centos6 removed the packages for devtoolsets 3, 4 and 5
devtoolset-6 still uses glibc 2.12 and ships with gcc 6.3


bmunkholm:
just add a changelog update under internal.

Issue #1977, opened 2018-08-21
Realm.Sync.User.login fails


jigajames:
## Goals

Logging into realm fails

## Actual Results

On version 2.11.0 -> 2.14.2 the following code fails but returns correctly on 2.10.0

```
await Realm.Sync.User.login('https://rentro.***.realm.io', '***', '***')
```

```
ExceptionsManager.js:84 Unhandled promise rejection TypeError: Cannot set property identity of #<User> which has only a getter
    at assign (_object-assign.js:32)
    at _extends (user.js:24)
    at createUser (user.js:64)
    at deserialize (rpc.js:160)
    at createUser (rpc.js:71)
    at Function.createUser (user.js:28)
    at user-methods.js:264
    at run (es6.promise.js:66)
    at es6.promise.js:83
    at flush (_microtask.js:18)
```

And on the version 2.10.0 I get a segfault when subscribing for sync. 

## Version of Realm and Tooling

ream-js 2.11.0 - 2.14.2

jigajames:
The segfault on 2.10.0 was caused by not providing schemas to user.createConfiguration

kneth:
@jigajames It is actually the same as #1976. The plan is to add a check, and throw an exception when you try to open without a schema.

PR #1985, opened 2018-08-23
A schema is required when Realm is opened in a query-based sync situation

### LINKED ISSUE #1976: Query-based sync requires a schema
## Goals

Either fail with an exception or succeed when subscribing.

## Expected Results

When opening a schema-less query-based synced Realm, an exception should be thrown.

## Actual Results

An segmentation fault happens when `subscribe()` is called.

## Steps to Reproduce

See below.

## Code Sample

```
var Realm = require('realm');

// the URL to the Realm Object Server
var SERVER_URL = 'localhost:9080';
let userUuid = 'realm-admin';

Realm.Sync.User.login('http://' + SERVER_URL, userUuid, '').then(user => {
  const config = user.createConfiguration({
      sync: { url: 'realm://' + SERVER_URL + '/my-test-app',
          error: err => console.log(err)
        },
  });
  Realm.open(config).then(realm => {
      let subscription = realm.objects('Test').subscribe(); // segfault!!!
    })
});
```

- Realm JS SDK Version: 2.14.2
- Node or React Native: node
- Client OS & Version: any

kneth:
## What, How & Why?

When a Realm is opened for query-based sync, `subscribe()` will fail if no schema is defined.

This closes #1976 and #1977 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


kneth:
@nirinchev True but only if your configuration is generated by `createConfiguration` and you haven't modified the `schema` part.

nirinchev:
But if you modified the schema part, doesn't that mean that 1) you have a schema, and 2) that the default schema should get merged with yours?

nirinchev:
To be clear, I think that while the changes in this PR are fine, but a real fix would be to make sure we add the permission classes always as this will address both this issue and the one reported by Christian.

kneth:
The check here will be replaced by https://github.com/realm/realm-js/issues/1988

PR #1990, opened 2018-08-24
fullSynchronization might not be defined

### LINKED ISSUE #1989: false negative "Query-based sync requires a schema" error thrown with 2.15.2
The check introduced with https://github.com/realm/realm-js/pull/1985 will incorrectly trigger the error for full sync realms when the configuration is created using the constructor.

kneth:
## What, How & Why?

This closes #1989 

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+X.Y.Z Release notes
+=============================================================
+### Compatibility
+* Sync protocol: 24
+* Server-side history format: 4
+* File format: 7
+* Realm Object Server: 3.0.0 or later
+
+### Breaking changes
+* None.
+
+### Enhancements
+* None.
+
+### Bug fixes
+* Fixed a bug if `fullSynchronized` is not defined. This could lead to an exception when opening a synced Realm.
+
+### Internal
+* Realm Core v5.7.2.
+* Realm Sync v3.9.1.
+
 2.15.2 Release notes (2018-8-24)
 =============================================================
 ### Compatibility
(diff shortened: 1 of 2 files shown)


lib/extensions.js
```diff
@@ -80,7 +80,7 @@ module.exports = function(realmConstructor) {
             // result in sync rejecting the writes. `_waitForDownload` ensures that the session is kept
             // alive until our callback has returned, which prevents it from being torn down and recreated
             // when we close the schemaless Realm and open it with the correct schema.
-            if (!config.sync.fullSynchronization && config.schema === undefined) {
+            if (config.sync.fullSynchronized !== undefined && !config.sync.fullSynchronization && config.schema === undefined) {
```
nirinchev: I think the check should be `config.sync.fullSynchronization === false` - otherwise this will get triggered even when full sync is `true`.

kneth:
Comment by @nirinchev has been addressed

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* Realm Object Server: 3.0.0 or later
+
+### Breaking changes
+* None.
+
+### Enhancements
+* None.
+
+### Bug fixes
+* Fixed a bug if `fullSynchronized` is not defined. This could lead to an exception when opening a synced Realm.
```
bmunkholm: @kneth Please specify where "fullSynchonized" should be defined. And which version was this introduced?

bmunkholm: And also mention which version the bug was introduced in and include a reference to the issue/PR number.

kneth: See https://github.com/realm/realm-js/pull/1998

PR #1996, opened 2018-08-27
Expose user serialize and deserialize methods

### LINKED ISSUE #1276: Serializing and deserializing a User
## Goals

In a multi process environment, like an electron app, it would be valuable to be able to pass users between processes in a serialized form.

## Expected Results

I would expect that when logging in to ROS and serializing the resulting user I would get an object as follows:

    Realm.Sync.User.register("http://localhost:9080", username, password, (err, user) => {
        const userString = JSON.stringify(user);
        // user === "{token: "...", identity: "some-username", server: "http://localhost:9080", isAdmin: false }"

This could be fixed by implementing a `toJSON` method on the `Realm.Sync.User` class: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior

In addition to this, it would be valuable to create a `Realm.Sync.User` from a JSON representation. This could be implemented with a `fromJSON` method on the user.

## Actual Results

Serializing a Realm.Sync.User (`JSON.stringify(user)`) gives you the string representation of an empty object (`{}`) and the fields of a `User` are all implemented as getters that cannot be set.

## Steps to Reproduce

1. Login or register a user through the Realm Object Server.
2. Serialize the user using `JSON.stringify(user)`
3. Observe the disappointing result

## Code Sample

See "Expected Results"

- Realm JS SDK Version: `2.0.0-rc5`

nirinchev:
## What, How & Why?

This closes #1276
It also removes an annoying false negative warning when using `User.createConfiguration`.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/lib/user-methods.js
+++ b/lib/user-methods.js
@@ -38,6 +38,25 @@ function checkTypes(args, types) {
     }
 }
 
+function checkObjectTypes(obj, types) {
+    for (const name of Object.getOwnPropertyNames(types)) {
+        const actualType = typeof obj[name];
+        let targetType = types[name];
+        const isOptional = targetType[targetType.length - 1] === '?';
+        if (isOptional) {
+            targetType = targetType.slice(0, -1);
+        }
+
+        if (!isOptional && actualType === 'undefined') {
+            throw new Error(`${name} is required, but a value was not provided.`);
+        }
+
+        if (actualType !== targetType) {
+            throw new TypeError(`${name} must be of type '${targetType}' but was of type '${actualType}' instead.`);
+        }
+    }
+}
+
 // Perform a HTTP request, enqueuing it if too many requests are already in
 // progress to avoid hammering the server.
 const performFetch = (function() {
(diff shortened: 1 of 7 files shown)


lib/user-methods.js
```diff
@@ -478,98 +482,106 @@ const instanceMethods = {
             .then(() => console.log('User is logged out'))
             .catch((e) => print_error(e));
     },
-        openManagementRealm() {
```
nirinchev: Below is a whitespace change except for the `config.sync.user && console.warn !== undefined` line. I advise viewing with "hide whitespace changes" enabled.

nirinchev:
@kneth since this is purely a js implementation, I'm not sure if the chrome rpc API need to be updated.

kneth:
@nirinchev I agree, that you don't need to do anything related for the debugger since you are working in pure JS.

nirinchev:
Ping @kraenhansen @kneth for the input validation logic

Issue #1579, opened 2017-12-18
Align sync auth with other SDKs


kneth:
The current auth API is hard to new users to understand and use. I suggest that Realm JS is aligned with the other Realm SDKs.

* The current methods (`registerWithProvider`, `login`) are deprecated and to be removed as soon as possible.
* A new class`Realm.Sync.Credentials` is added. It will interface supported 3rd party provides, e.g., JWT, Google and Facebook.
* Methods for logging in and and logging out are added to `Realm.Sync.User`.

An example:

```
credentials = Realm.Sync.Credentials.facebook(getFacebookToken())
Realm.Sync.User.login(credentials, "http://objectserver.realm.io/auth").then(user => {
    // user is now logged and and can open a Realm
}).catch(error => {
    // an error has occurred and be handled
})
```

nirinchev:
This could be related: https://github.com/realm/realm-js/issues/1571

bigfish24:
This is a good discussion to have since with the cloud I think we should simplify the auth to just:

1. Username/password
2. JWT (custom)
3. Anonymous

The extra providers are not worth maintaining or building UIs to customize them. You can authenticate with Facebook via the JWT method since it should cover any auth flow.

I think just having 1 method like `authenticate` or `login` would be sufficient to pass a credential and return a user.

nirinchev:
I'm not sure about that - if the goal is to allow people to bootstrap quickly, the JWT flow isn't exactly ideal. It requires writing code to issue my own tokens and hosting this on another service. If I already have that in place, great, but if I don't, it's fairly annoying.

bigfish24:
Sure, but we have to balance priorities. I don't think supporting social auths are very high on the list of things to work on for cloud. Any effort to build a UI around them or maintain them could definitely be spent on higher value things. If we were confident in our webdev skills to build an amazing onboarding, build out cloud.realm.io with billing, etc. then I could see it as adding value, but it seems to take away resources from these efforts.

kneth:
@bigfish24 I would assume that Facebook and Google auth are important in a React Native context?

@nirinchev Will `credentials = Realm.Sync.Credentials.JWT(getJWTToken())` make it harder to get started with JWT?

nirinchev:
[...]

PR #2002, opened 2018-08-31
Redesign user auth

nirinchev:
## What, How & Why?

This aligns the authentication API with the other SDKs.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,23 @@ X.Y.Z Release notes
 * Realm Object Server: 3.0.0 or later
 
 ### Breaking changes
-* None.
+* The authentication API have been completely revamped.
+  * The following methods have been removed:
+    * `Realm.Sync.User.login`
+    * `Realm.Sync.User.register`
+    * `Realm.Sync.User.authenticate`
+    * `Realm.Sync.User.registerWithProvider`
+    * `Realm.Sync.User.adminUser`
+  * A new `Realm.Sync.User.login` method has been added that accepts the server url and a credentials object.
+  * A new class - `Realm.Sync.Credentials` has been added that contains factory methods to create credentials
+  with all supported providers.
+  * Here are some examples on how to transform your old code to use the new API:
+
+  | Old | New |
+  | - | - |
+  | <pre><code>const user = await Realm.Sync.User.login(serverUrl, 'username', 'password');</code></pre> | <pre><code>const credentials = Realm.Sync.Credentials.usernamePassword('username', 'password');<br>const user = await Realm.Sync.User.login(serverUrl, credentials); </code></pre> |
+  | <pre><code>const jwtToken = 'acc3ssT0ken...';<br>const user = await Realm.Sync.User.registerWithProvider(serverUrl, 'jwt', jwtToken);</code></pre> | <pre><code>const jwtToken = 'acc3ssT0ken...';<br>const credentials = Realm.Sync.Credentials.jwt(jwtToken);<br>const user = await Realm.Sync.User.login(serverUrl, credentials);</code></pre> |
+  | <pre><code>const customToken = 'acc3ssT0ken...';<br>const userInfo = { someValue: true };<br>const user = await Realm.Sync.User.registerWithProvider(serverUrl, 'custom/fooauth', customToken, userInfo);</code></pre> | <pre><code>const customToken = 'acc3ssT0ken...';<br>const userInfo = { someValue: true };<br>const credentials = Realm.Sync.Credentials.custom('custom/fooauth', customToken, userInfo);<br>const user = await Realm.Sync.User.login(serverUrl, credentials);</code></pre> |
 
 ### Enhancements
 * Exposed `User.serialize` to create a persistable representation of a user instance, as well as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ X.Y.Z Release notes
 * Realm Object Server: 3.0.0 or later
 
 ### Breaking changes
-* The authentication API have been completely revamped.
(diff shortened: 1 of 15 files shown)


lib/user-methods.js
```diff
@@ -609,10 +515,64 @@ const instanceMethods = {
...
+        return new Credentials('azuread', token)
+    },
+
+    jwt(token) {
+        return new Credentials('jwt', token);
+    },
+
+    adminToken(token) {
+        return new Credentials('adminToken', token);
+    },
```
kraenhansen: I guess we could use `checkTypes` for `nickname`, `azureAD`, `jwt` and `adminToken` as well?

lib/user-methods.js
```diff
@@ -609,10 +515,64 @@ const instanceMethods = {
...
+    nickname(value, isAdmin) {
+        return new Credentials('nickname', value, { is_admin: isAdmin || false });
+    },
+
+    azureAD(token) {
+        return new Credentials('azuread', token)
+    },
+
+    jwt(token) {
+        return new Credentials('jwt', token);
```
kraenhansen: Should this take an optional `provider` too?

docs/sync.js
```diff
@@ -288,58 +288,105 @@ class IncompatibleSyncedRealmError {
...
+    /**
+     * Gets additional user information associated with the credentials.
+     * @returns {object} A dictionary, containing the additional information.
+     */
+    get userInfo();
+
+}
+
+/**
+ * Class for logging in and managing Sync users.
```
kneth: Just leave out "logging in and".

kneth:
@nirinchev I have connected your PR with #1579 which I believe is the correct issue. To me, it is important that we get something which is as close to our other SDKs as possible.

nirinchev:
I agree - this very closely mirrors the .NET API, which if I remember correctly is aligned with Cocoa and Java (maybe with a few casing differences). @cmelchior can you do a high level review and see if the new API diverges in significant ways from Java?

lib/index.d.ts
```diff
@@ -290,6 +290,21 @@ declare namespace Realm.Sync {
         isAdmin: boolean;
     }
 
+    class Credentials {
```
cmelchior: In Java this is called `SyncCredentials` and `SyncUser`, here it is called `Credentials` and `User` but I guess that is due it being inside the Realm.Sync namespace.

lib/index.d.ts
```diff
@@ -290,6 +290,21 @@ declare namespace Realm.Sync {
...
+        static usernamePassword(username: string, password: string, createUser?: boolean): Credentials;
+        static facebook(token: string): Credentials;
+        static google(token: string): Credentials;
+        static anonymous(): Credentials;
+        static nickname(value: string, isAdmin?: boolean): Credentials;
+        static azureAD(token: string): Credentials;
+        static jwt(token: string): Credentials;
+        static adminToken(token: string): Credentials;
+
+        readonly provider: string;
```
cmelchior: Java uses the term `identityProvider`

lib/index.d.ts
```diff
@@ -290,6 +290,21 @@ declare namespace Realm.Sync {
...
+        readonly token: string;
```
cmelchior: And this is called `userIdentifier`

lib/index.d.ts
nirinchev: .NET uses `IdentityProvider` as well, Cocoa uses `provider`, but Cocoa's version is internal, so people can't actually access it. I guess, I'll change it to align with Java and .NET

lib/index.d.ts
nirinchev: In .NET and Cocoa we use `token`, although as mentioned, Cocoa's properties are private. I'll speak with Kenneth and see which one he likes better - I don't have a strong opinion on the matter :)

bmunkholm:
@nirinchev The "pipeline" in zenhub tracks the state - doing that in the subject with "[WIP]" is redundant and not a scheme that is used by others either, so I've removed "[WIP]" :-)

CHANGELOG.md
```diff
@@ -7,7 +7,23 @@ X.Y.Z Release notes
 * Realm Object Server: 3.0.0 or later
 
 ### Breaking changes
-* None.
+* The authentication API have been completely revamped.
```
kneth: "have" -> "has"

CHANGELOG.md
```diff
@@ -7,7 +7,23 @@ X.Y.Z Release notes
...
+  * The following methods have been removed:
```
kneth: "removed" -> "deprecated and will be removed at next major version"

docs/sync.js
```diff
@@ -288,58 +288,120 @@ class IncompatibleSyncedRealmError {
...
+    /**
+     * Creates credentials based on an Active Directory login.
+     * @param {string} token An access token, obtained by logging into Azure Active Directory.
+     * @return {Credentials} An instance of `Credentials` that can be used in {@linkcode Realm.Sync.User.login|User.login}.
+     */
+    static azureAD(token) {};
+
+    /**
+     * Creates credentials based on a JWT login.
+     * @param {string} token A Json Web Token, that will be validated against the server's configured rules.
```
kneth: "Json" -> "JSON"

CHANGELOG.md
nirinchev: I was using the plural version of API 🤓

kneth:
The CI errors are `The @deprecated, tag is not a known tag. File: sync.js` - which is strange as it is a [valid tag](http://usejsdoc.org/tags-deprecated.html).

PR #2003, opened 2018-08-31
Add param type to addConnectionNotification/removeConnectionNotification

gabro:
## What, How & Why?
This PR adds the `ConnectionNotificationCallback` type to the `callback` parameter of both `Session#addConnectionNotification` and `Session#removeConnectionNotification`.

Since the `ConnectionNotificationCallback` type alias was present but unused, I suspect this has been an oversight when adding those methods to the typings.

Without an explicit type, the `callback` parameter is implicitly inferred as `any`, causing a TypeScript build to fail if `--strict=true`

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Server and open synchronized Realms (#1276).
 ### Bug fixes
 * Removed a false negative warning when using `User.createConfiguration`.
 * Fixed the type definition for `User.authenticate`.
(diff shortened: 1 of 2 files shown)


gabro:
Thanks for the quick review :)

kneth:
@gabro Thanks  for the contribution. Please sign our [CLA](https://github.com/realm/realm-js/blob/master/CONTRIBUTING.md#cla)

gabro:
Done

PR #2013, opened 2018-09-09
Fixed the type definition for Realm.Permissions.User

### LINKED ISSUE #2012: Wrong typescript definition for Realm.Permissions.User
The type definition for the type `Realm.Permissions.User` is as follows:

```
class User {
        static schema: ObjectSchema;
        identity: string;
    }
```
The `identity` field I think is wrong because in realm studio the table contains an `id` field instead

- Realm JS SDK Version: 2.15.3

apperside:
The `identity` field was wrong, it should have been `id` instead because that is the name of the field in Realm studio

## What, How & Why?
The `identity` field was wrong, it should have been `id` instead

This closes #2012 

* [x] jsdoc files updated

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -578,7 +578,7 @@ declare namespace Realm.Permissions {
 
     class User {
         static schema: ObjectSchema;
-        identity: string;
+        id: string;
     }
 
     class Role {


apperside:
@kneth I completed the submission of the contribution agreement form

PR #2019, opened 2018-09-11
Add support for sync pause/resume

### LINKED ISSUE #2014: Add support for sync pause/resume

emanuelez:
## What, How & Why?

This closes #2014 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


bmunkholm:
Should it be start/stop or pause/resume as in Cocoa?

emanuelez:
It's start/stop in Java

tests/js/session-tests.js
```diff
@@ -1041,6 +1051,114 @@ module.exports = {
...
+                    }
+                };
+
+                const checks = {
+                    started: false,
+                    stopped: false,
+                    restarted: false,
+                }
+
+                Realm.open(config).then(realm => {
```
nirinchev: You can move the `Realm.open` outside the promise constructor. Something like:

```js
Realm.Sync.User.register(...)
    .then((user) => {
        const config = { ... };
        return Realm.open(config);
    })
    .then((realm) => {
        return new Promise(...);
    });
```

tests/js/session-tests.js
```diff
@@ -1041,6 +1051,114 @@ module.exports = {
...
+                    const session = realm.syncSession;
+
+                    session.addConnectionNotification((newState, oldState) => {
+                        if (newState === Realm.Sync.ConnectionState.Connected && checks.started === false) { checks.started = true; session.stop(); }
```
nirinchev: This is very convoluted to read. Any reason not to spread them on multiple lines?

tests/js/session-tests.js
```diff
@@ -1041,6 +1051,114 @@ module.exports = {
...
+            return new Promise((resolve, reject) => {
+                let config = {
+                    sync: {
+                        user: u,
+                        url: `realm://localhost:9080/~/${uuid()}`,
+                        fullSynchronization: true,
+                    }
+                };
+
+                Realm.open(config).then(realm => {
```
nirinchev: Same here - `Realm.open` can be moved outside the promise constructor

tests/js/session-tests.js
```diff
@@ -1041,6 +1051,114 @@ module.exports = {
...
+    },
+
+    testMultipleStarts() {
+        if(!isNodeProccess) {
+            return;
+        }
+
+        return Realm.Sync.User.register('http://localhost:9080', uuid(), 'password').then((u) => {
+            return new Promise((resolve, reject) => {
+                let config = {
```
nirinchev: Should this be `const`?

tests/js/session-tests.js
emanuelez: I see all the other tests in the file are structured as mine. Also, I cannot concatenate a `.catch(error => reject(error));` at the end of the chain you propose.

nirinchev: You don't need to concatenate it because the promise will be rejected anyway.

nirinchev: With the current code, you reject a promise if `Realm.open` fails. If you return it as I suggest, you'll end up with the same logical flow without having a codepath for promise rejection.

PR #2030, opened 2018-09-19
Fix typescript definitions for privileges() methods

cmelchior:
Even though you can access all properties in the `RealmPrivileges/ClassPrivileges/ObjectPrivileges` classes, I only exposed those with actual meaning in the definitions. This is in line with what Swift/Java does.


bmunkholm:
@kneth @cmelchior  Please take our kind bots suggestion to heart :-)
All issues and PR's must have a T: label.

PR #2038, opened 2018-09-21
Logging in using admin token should also return a promise

### LINKED ISSUE #2037: Realm.Sync.User.login() does not always return a promise
The return value from `Realm.Sync.User.login()` is wrong when the credentials is created from an admin token. The user is returned and not a promise. The following code snippets will throw an exception saying `then` is not a function.

```
let credentials = Realm.Sync.User.adminToken(admin_token);
Realm.Sync.User.login(serverUrl, credentials).then((user => { ... });
```

kneth:
## What, How & Why?

This closes #2037

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x[ 🚦 Tests

* ~[ ] jsdoc files updated~

### DIFF
--- a/tests/js/user-tests.js
+++ b/tests/js/user-tests.js
@@ -20,6 +20,7 @@
 
 'use strict';
 
+const fs = require('fs');
 const Realm = require('realm');
 const TestCase = require('./asserts');
 const isNodeProcess = typeof process === 'object' && process + '' === '[object process]';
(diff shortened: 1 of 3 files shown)


tests/js/user-tests.js
```diff
@@ -186,6 +196,23 @@ module.exports = {
       .catch((e) => { Promise.reject(e) } )
   },
 
+  testAuthenticateAdminToken() {
+    if (!isNodeProcess) {
+      return
+    }
+    // read admin token from ROS
+    let obj = JSON.parse(fs.readFileSync('../realm-object-server-data/keys/admin.json', 'utf8'));
```
nhachicha: if it's the only usage why not declare it as 
`let obj = JSON.parse(require('fs').readFileSync('../realm-object-server-data/keys/admin.json', 'utf8'));` you can avoid the declaration in https://github.com/realm/realm-js/pull/2038/files#diff-f24a77a955cf2ca9d84fe9d6d07fe08bR33

kneth: I believe that the RN packager will pick up `fs` anyway when it scans the files.

nirinchev:
This was intentional as there are some codepaths in ROS that rely on synchronously creating an admin user. I don't suppose it will be incredibly hard to rewrite, but considering the fact that the old API was synchronous, it didn't seem like a problem. If we want to keep this change though, we need to update the typescript definitions and the API docs.

kneth:
@nirinchev I see. I might have been too fast. I saw a typing error, and wanted to add a test. I should probably rewrite the test :-)

nirinchev:
No worries, I'm not strongly opposed to the new behavior, but let's sync offline on the implications this might have for the adapters and ROS.

PR #2056, opened 2018-10-08
Clarifying licensing per #1987

### LINKED ISSUE #1987: Confused with the license
Hi Realm folk,

I was looking at the README which says "Realm JS is published under the Apache 2.0 license. Realm Core is also published under the Apache 2.0 license and is available here. "

When I look at the LICENSE file it also has a section regarding "REALM COMPONENTS - This software contains components with separate copyright and license terms". This appears to have been added to realm-cocoa to match 'core licensing' in 2014, but in 2016 it was removed from realm-core.

Could you help me understand if that text applies to realm-js and which components it applies to?

Thanks.

hyandell:
The README doesn't mention all of the licensing for this project.

## What, How & Why?

This closes #1987

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,8 +169,8 @@ See [CONTRIBUTING.md](https://github.com/realm/realm-js/blob/master/CONTRIBUTING
 
 ## License
 
-Realm JS is published under the Apache 2.0 license.
-Realm Core is also published under the Apache 2.0 license and is available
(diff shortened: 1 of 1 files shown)


hyandell:
CLA signed, while noting that this contribution has no expressive content. Please also let me know what the process is to discontinue a CLA is.

PR #2060, opened 2018-10-09
Enhancements of subscription api

### LINKED ISSUE #2052: Subscription enhancements request
To provide a nicer subscribe/unsubscribe API in the GraphQL service, we need the following enhancements to the subscription functionality in realm-js:

- [ ] Ability to unsubscribe by name. There should be no reason to subscribe to something only to later unsubscribe. 
- [ ] Ability to list all active subscriptions. We already have that information in the partial Realm, so there should be no reason not to expose it. I need just the name and the query and not a reference to the subscription instances or anything like that. 
- [ ] A nicer wrapper around waiting for subscription - something like:

```js
  const listener = (__, state) => {
    switch (state) {
      case Realm.Sync.SubscriptionState.Complete:
        subscription.removeListener(listener);
        resolve();
        break;
      case Realm.Sync.SubscriptionState.Error:
        subscription.removeListener(listener);
        reject(subscription.error);
        break;
    }
  };
  subscription.addListener(listener);
});
```

kneth:
## What, How & Why?

This closes #2052

Requires: https://github.com/realm/realm-object-store/pull/698

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


src/js_sync.hpp
```diff
@@ -762,6 +765,18 @@ void SubscriptionClass<T>::get_error(ContextType ctx, ObjectType object, ReturnV
     }
 }
 
+template<typename T>
+void SubscriptionClass<T>::get_name(ContextType ctx, ObjectType object, ReturnValue &return_value) {
+    auto subscription = get_internal<T, SubscriptionClass<T>>(object);
+    if (subscription->m_name == util::none) {
+        // FIXME: should we reconstruct the name to match the one stored in __ResultSets?
```
nirinchev: I don't think there's currently a way to get an existing subscription, so not sure if we need to do anything there.

lib/extensions.js
```diff
@@ -379,6 +403,45 @@ module.exports = function(realmConstructor) {
...
+                    if (!this.isInTransaction) {
+                        this.beginTransaction();
+                        doCommit = true;
+                    }
+                    this.delete(named_subscriptions);
+                    if (doCommit) {
+                        this.commitTransaction();
+                    }
+                } else {
+                    throw new Error(`string expected - got ${typeof(string)}`);
```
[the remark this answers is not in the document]
nirinchev: Does the empty string evaluate to false in js? If so, I imagine we can get here, even if a string was provided.

lib/extensions.js
```diff
@@ -379,6 +403,45 @@ module.exports = function(realmConstructor) {
...
+                        query: all_subscriptions[i]['query'],
+                    }
+                    listOfSubscriptions.push(sub);
+                }
+                return listOfSubscriptions;
+            },
+
+            unsubscribe(name) {
+                if (name && typeof(name) === 'string') {
+                    let named_subscriptions = this.objects('__ResultSets').filtered(`name == '${name}'`);
```
nirinchev: A minor optimization would be to return early if there are no matches.

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added `Realm.subscriptions()` to query active query-based sync subscriptions. ([#2052](https://github.com/realm/realm-js/issues/2052))
+* Added `Realm.unsubscribe()` to unsubscribe ny name an active query-based sync subscriptions. ([#2052](https://github.com/realm/realm-js/issues/2052))
```
cmelchior: `ny` ?

lib/extensions.js
```diff
@@ -379,6 +403,45 @@ module.exports = function(realmConstructor) {
...
+            subscriptions(arg) {
+                let all_subscriptions = this.objects('__ResultSets');
+                if (arg) {
+                    if (arg.includes('*') || arg.includes('?')) {
+                        all_subscriptions = all_subscriptions.filtered(`name LIKE '${arg}'`);
+                    } else {
+                        all_subscriptions = all_subscriptions.filtered(`name == '${arg}'`);
+                    }
+                }
+                let listOfSubscriptions = [];
```
cmelchior: So you went with a simple wrapper class after all? Wasn't part of the original use case to get notifications for the Subscription state, which seems impossible on the objects returned here?  Not saying this cannot be done, but you just need to make it _very_ clear in docs, that this is just an immutable snapshot that doesn't change.

nirinchev: The use case we have - exposing the existing subscriptions via GraphQL - doesn't require notifications. I'm fine with both but considering this is mainly catering to an internal use case, I'd say 🚢 it as it is and let's improve and align with other SDKs later. I'm also okay with leaving the API undocumented or documented with a beta label if we feel we'll want to change them soon.

CHANGELOG.md
kneth: Damn, `b` is too close to `n` on my keyboard.

lib/extensions.js
kneth: Yeah, I plan to wrap up and ship it as 2.19.0-rc.2.

lib/extensions.js
```diff
@@ -379,6 +403,45 @@ module.exports = function(realmConstructor) {
...
+                for (var i = 0; i < all_subscriptions.length; ++i) {
```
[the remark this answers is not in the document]
kneth: I used `i` for debugging, but you are right :-)

lib/extensions.js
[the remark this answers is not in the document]
kneth: True. We need to first check the type and then the value and give precise errors.

lib/extensions.js
kneth: Avoiding an empty transaction good practice.

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added `Realm.subscriptions()` to query active query-based sync subscriptions. This method is in beta and might change in future releases. ([#2052](https://github.com/realm/realm-js/issues/2052))
+* Added `Realm.unsubscribe()` to unsubscribe by name an active query-based sync subscriptions. This method is in beta and might change in future releases. ([#2052](https://github.com/realm/realm-js/issues/2052))
```
bmunkholm: "an" is singular, "subscriptions" plural...

docs/realm.js
```diff
@@ -331,6 +331,22 @@ class Realm {
...
+     * Get a list of subscriptions. THIS METHOD IS IN BETA AND MAY CHANGE IN FUTURE VERSIONS.
+     * @param {string} name - Optional parameter to query for either a specific name or pattern (using
+     *   cards `?` and `*`).
+     * @throws {Error} If `name` is not a string.
+     * @returns an array of objects of (`name`, `objectType`, `query`).
+     */
+    subscriptions(name) {}
+
+    /**
+     * Unsubscribe a name subscription. THIS METHOD IS IN BETA AND MAY CHANGE IN FUTURE VERSIONS.
```
bmunkholm: name -> named

src/js_sync.hpp
kneth: My concern is that for unnamed subscriptions `subscription.name` will be `undefined` while `Realm.subscriptions()` will return a list of objects where `name` is not `undefined` but a name constructed by object store. You can see the inconsistency on the test 😞 .

lib/extensions.js
```diff
@@ -379,6 +403,53 @@ module.exports = function(realmConstructor) {
                     return classPermissions[0];
                 }
             },
+
+            subscriptions(name) {
+                let all_subscriptions = this.objects('__ResultSets');
+                if (typeof(name) !== 'string') {
```
nirinchev: The docs mention that `name` is optional, but here it seems like we'll throw if it's not provided. Let's only check the type if it's not `undefined`.

cmelchior: I believe OS will auto-generate a name if one isn't provided. It is in the form of `[Type] Query`

kneth: Yeah, `name` is optional. Otherwise it will be hard to get *all* subscriptions.

PR #2064, opened 2018-10-10
Unit test for partial Realm schema updates

nirinchev:
## What, How & Why?

This adds a unit test to demonstrate an issue with partial Realm schema updates when opened in dynamic mode.

### DIFF
--- a/tests/js/realm-tests.js
+++ b/tests/js/realm-tests.js
@@ -32,6 +32,17 @@ function uuid() {
     });
 }
 
+function closeAfterUpload(realm) {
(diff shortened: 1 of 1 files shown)


kneth:
The schema listeners are called but they only see the schema defined in their configuration. I think this behaviour can be explain by https://github.com/realm/realm-object-store/blob/master/src/shared_realm.cpp#L535. If a Realm has a schema in the configuration, `m_dynamic_realm` is false, and only new columns are added but not new tables.

kneth:
@nirinchev Is it still relevant or can it be closed?

PR #2065, opened 2018-10-12
Restore support for opening query-based sync Realms with a dynamic schema

tgoyne:
Mostly just a proof of concept and needs some further work and tests. The basic idea here is to only add the permissions types to the schema if there is one (rather than forcing it to be in non-dynamic mode), and then call `update_schema()` on the temporary download Realm to solve the actual problem hit in #1976.

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -645,6 +645,18 @@ SharedRealm RealmClass<T>::create_shared_realm(ContextType ctx, realm::Realm::Co
         js_binding_context->m_defaults = std::move(defaults);
         js_binding_context->m_constructors = std::move(constructors);
     }
+#if REALM_ENABLE_SYNC
+    // For query-based Realms we need to register the constructors for the
+    // permissions types even if a schema isn't specified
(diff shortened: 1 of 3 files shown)


kneth:
@tgoyne I have made a few changes. You can find them in #2066. We need to get the permissions into the JS.

tgoyne:
This now passes tests so I guess it's ready for review.

src/js_realm.hpp
```diff
@@ -272,6 +274,9 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
         {"clearTestState", wrap<clear_test_state>},
         {"copyBundledRealmFiles", wrap<copy_bundled_realm_files>},
         {"deleteFile", wrap<delete_file>},
+ #if REALM_ENABLE_SYNC
+        {"_asyncOpen", wrap<async_open_realm>},
```
kneth: So we have `Realm.openAsync()` and `Realm._asyncOpen()`? It must be a sign that we should remove `openAsync()` very soon.

PR #2068, opened 2018-10-12
Update package.json

hyandell:
Pointing to the LICENSE file as Realm JS has mixed licensing

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "realm",
   "description": "Realm is a mobile database: an alternative to SQLite and key-value stores",
   "version": "2.18.0",
-  "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


hyandell:
Noting that NPM instructions are:

https://docs.npmjs.com/files/package.json

> If you are using a license that hasn't been assigned an SPDX identifier, or if you are using a custom license, use a string value like this one:
> 
> { "license" : "SEE LICENSE IN <filename>" }
> Then include a file named <filename> at the top level of the package.

Which does seem to lead to a bit of a mouthful given the file name. I'm not sure if the above is really freeform text or if the 'SEE LICENSE IN ' is parsed out sometimes.

hyandell:
Will close, git update my fork, and then recreate.

hyandell:
See #2069

PR #2106, opened 2018-11-12
Fix Typescript definition for Permission

cmelchior:
The `role` property was not correct in the `Permission` Typescript definition leading to compile errors of the form: `error TS2339: Property 'role' does not exist on type 'Permission'.`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,22 @@ x.x.x Release notes (yyyy-MM-dd)
  ### Internal
 * None.
 
+2.19.1 Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,22 @@ x.x.x Release notes (yyyy-MM-dd)
  ### Internal
 * None.
 
+2.19.1 Release notes (yyyy-MM-dd)
```
kneth: Please don't specify version number. `npm set-version` will do it during a release.

CHANGELOG.md
```diff
@@ -15,6 +15,22 @@ x.x.x Release notes (yyyy-MM-dd)
...
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* The Typescript definition for `Realm.Permissions.Permission` did not have the correct `role` property defined. Since 2.3.0.
```
kneth: Please paste the error message so users can see if they have been hit.

Issue #2107, opened 2018-11-13
new Buffer is deprecated in node 10.


nirinchev:
We should use `Buffer.from` in submit-analytics.js

PR #2108, opened 2018-11-13
Give the app a chance to disable analytics

nirinchev:
## What, How & Why?

1. Introduce 100 ms delay before submitting analytics - this is helpful to handle a case where setting the environment variable in code happens after realm is imported.
2. Add a `context` argument to distinguish between node.js and electron.
3. Removes `new Buffer` calls which are deprecated in node 10.

Fixes https://github.com/realm/realm-js/issues/2107.
## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/lib/submit-analytics.js
+++ b/lib/submit-analytics.js
@@ -34,7 +34,14 @@ if (isAnalyticsDisabled()) {
 } else {
     const os = require('os');
     const request = require('request');
-    const { machineIdSync } = require("node-machine-id");
+    const crypto = require('crypto');
+    const { machineId } = require("node-machine-id");
+
+    function sha256(data) {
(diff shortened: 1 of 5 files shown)


lib/submit-analytics.js
```diff
@@ -33,56 +33,8 @@ if (isAnalyticsDisabled()) {
...
-                return sha256(getDarwinIdentifier());
-            case 'linux':
-                return sha256(getLinuxIdentifier());
-            case 'windows':
-                return sha256(getWindowsIdentifier());
-            default:
-                return null;
-        }
-    }
+    const { machineIdSync } = require("node-machine-id");
```
kneth: What a nice simplification :smile:

PR #2116, opened 2018-11-15
Add  lower bound when refreshing tokens.

### LINKED ISSUE #2115: Token refresh appears to happen more often than necessary
@cmelchior noticed that with a standalone script - we issue unreasonably many refresh requests in certain cases. It's not clear how to repro that yet, but we should investigate.

cmelchior:
Fixes #2115 

This introduces a somewhat arbitrary lower bound for refreshing tokens. It is mostly selected to be low enough so you never hit it in practical use cases, and high enough so it doesn't completely spam the server log if you configure the server in a way that would cause immediate refreshes.

Potentially a safer way is to enforce a lower bound on the server side since this is where the TTL is defined. but since there is no contract for this with the server, having it here also makes sense.

In the worst case where the server sets a TTL below 10 seconds, then you will experience "gaps" in the synchronization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ x.x.x Release notes (yyyy-mm-dd)
 * None.
 
 ### Fixed
-* Tokens are refreshed ahead of time. If the lifetime of the token is lower than the threshold for refreshing it will cause the client to continously refresh, spamming the server with refresh requests. A lower bound of 10 seconds have been introduced. ([#2115](https://github.com/realm/realm-js/issues/2115), since 1.0.2)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+x.x.x Release notes (yyyy-mm-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* Tokens are refreshed ahead of time. If the lifetime of the token is lower than the threshold for refreshing it will cause the client to continously refresh, spamming the server with refresh requests. A lower bound of 10 seconds have been introduced. ([#2115](https://github.com/realm/realm-js/issues/2115), since 1.0.2)
```
```suggestion
* Tokens are refreshed ahead of time. If the lifetime of the token is lower than the threshold for refreshing it will cause the client to continously refresh, spamming the server with refresh requests. A lower bound of 10 seconds has been introduced. ([#2115](https://github.com/realm/realm-js/issues/2115), since v1.0.2)
```

Issue #2122, opened 2018-11-19
Expose `waitForUpload` and `waitForDownload`


nirinchev:
## Goals

We need the native `waitForUpload` and `waitForDownload` methods exposed either on the session or the Realm. Previously we had an internal `_waitForDownload` method on the `Realm` object but that is now gone and `_waitForUpload` is awkwardly placed on `Realm.Sync`. We should expose a consistent public API for these as using the byte-level progress notifications is overkill for the simple scenario where you want to ensure that everything has been uploaded/downloaded.

nirinchev:
Ideally the API should look something like this:

```ts
    class Session {
        // Other session methods

        waitForUpload(): Promise<void>;
        waitForDownload(): Promise<void>;
    }
```

nirinchev:
@roberhofer this is proving problematic for merging stage 1 of the service scalability effort. Please bring it up at the upcoming sdk meeting.

bmunkholm:
Perhaps @cmelchior can take a stab at this?

cmelchior:
Java already exposes this build on top of `Session::async_wait_fow_download_completion()`. They are called `SyncSession.downloadAllServerChanges()` or `SyncSession.uploadAllLocalChanges()`. We also have variants that accept a timeout. 

From what I recall using the blocking variant is dangerous in Sync as you are then effectively blocked behind a mutex with no way to cancel.

But I can take a look.

cmelchior:
@nirinchev One question though. Would a JS user expect the thread to be fully up to date with notifications triggered once `Session.downloadAllServerChanges()` completes? 

In Java we don't do this as the SyncSession isn't coupled to any particular thread, but I'm a bit unsure about the JS semantics since there is only 1 thread? I suspect that people would expect it to be up to date?

PR #2124, opened 2018-11-20
Add support for uploading. and downloading changes easily

cmelchior:
Closes https://github.com/realm/realm-js/issues/2122

This PR adds two new API's:
```
Realm.Sync.Session.uploadAllLocalChanges(timeout): Promise<void>
Realm.Sync.Session.downloadAllServerChanges(timeout): Promise<void>
```

These match the methods found in Realm Java.

TODO:
- [x] Fix tests
- [x] Changelog
- [x] Cleanup

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -272,7 +272,7 @@ module.exports = function(realmConstructor) {
             waitForCompletion(this, this._waitForUploadCompletion, timeout, `Uploading changes did not complete in ${timeout} ms.`);
         };
 
-        realmConstructor.Sync.prototype.downloadAllLocalChanges = function(timeout) {
+        realmConstructor.Sync.Session.prototype.downloadAllLocalChanges = function(timeout) {
             waitForCompletion(this, this._waitForDownloadCompletion, timeout, `Downloading changes did not complete ${timeout} ms.`);
         };
 
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -269,11 +269,11 @@ module.exports = function(realmConstructor) {
         }
 
         realmConstructor.Sync.Session.prototype.uploadAllLocalChanges = function(timeout) {
-            waitForCompletion(this, this._waitForUploadCompletion, timeout, `Uploading changes did not complete in ${timeout} ms.`);
+            return waitForCompletion(this, this._waitForUploadCompletion, timeout, `Uploading changes did not complete in ${timeout} ms.`);
         };
 
         realmConstructor.Sync.Session.prototype.downloadAllServerChanges = function(timeout) {
-            waitForCompletion(this, this._waitForDownloadCompletion, timeout, `Downloading changes did not complete ${timeout} ms.`);
+            return waitForCompletion(this, this._waitForDownloadCompletion, timeout, `Downloading changes did not complete ${timeout} ms.`);
         };
 
         // Keep these value in sync with subscription_state.hpp
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -78,7 +78,7 @@ function addSchemaIfNeeded(schemaList, schemaObj) {
 
 function waitForCompletion(session, fn, timeout, timeoutErrorMessage) {
     const waiter = new Promise((resolve, reject) => {
-        fn((error) => {
+        fn.call(session, (error) => {
             if (error === undefined) {
                 setTimeout(() => { resolve(); }, 1); // FIXME Check if thread is up to date?
             } else {
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -268,7 +268,7 @@ module.exports = function(realmConstructor) {
             }
         }
 
-        realmConstructor.Sync.prototype.uploadAllLocalChanges = function(timeout) {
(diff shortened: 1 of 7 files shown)


lib/extensions.js
```diff
@@ -244,6 +268,14 @@ module.exports = function(realmConstructor) {
             }
         }
 
+        realmConstructor.Sync.prototype.uploadAllLocalChanges = function(timeout) {
```
fealebenpae: Shouldn't this be 
```suggestion
        realmConstructor.Sync.Session.prototype.uploadAllLocalChanges = function(timeout) {
```
instead?

lib/extensions.js
```diff
@@ -244,6 +268,14 @@ module.exports = function(realmConstructor) {
...
+            waitForCompletion(this, this._waitForUploadCompletion, timeout, `Uploading changes did not complete in ${timeout} ms.`);
+        };
+
+        realmConstructor.Sync.prototype.downloadAllLocalChanges = function(timeout) {
```
fealebenpae: Ditto.

lib/extensions.js
cmelchior: Ups, yes

lib/index.d.ts
```diff
@@ -489,6 +489,9 @@ declare namespace Realm.Sync {
 
         resume(): void;
         pause(): void;
+
+        downloadAllServerChanges(): Promise<void>;
```
nirinchev: Missing the timeout argument

lib/extensions.js
```diff
@@ -76,6 +76,30 @@ function addSchemaIfNeeded(schemaList, schemaObj) {
...
+            if (error === undefined) {
+                setTimeout(() => { resolve(); }, 1); // FIXME Check if thread is up to date?
+            } else {
+                setTimeout(() => { reject(error); }, 1);
+            }
+        });
+    });
+    if (timeout === undefined) {
+        return waiter;
+    } else {
```
nirinchev: This `else` is redundant - there's a `return` in the if-branch.

lib/index.d.ts
nirinchev:
```suggestion
        downloadAllServerChanges(timeoutMs?: number): Promise<void>;
```

lib/index.d.ts
```diff
@@ -489,6 +489,9 @@ declare namespace Realm.Sync {
...
+        uploadAllLocalChanges(): Promise<void>;
```
nirinchev:
```suggestion
        uploadAllLocalChanges(timeoutMs?: number): Promise<void>;
```

cmelchior:
Ready for review

src/js_sync.hpp
```diff
@@ -679,6 +687,59 @@ void SessionClass<T>::override_server(ContextType ctx, ObjectType this_object, A
...
+                Object::set_property(protected_ctx, error_object, "message", Value::from_string(protected_ctx, error.message()));
+                Object::set_property(protected_ctx, error_object, "errorCode", Value::from_number(protected_ctx, error.value()));
+                callback_arguments[0] = error_object;
+            } else {
+                callback_arguments[0] = Value::from_undefined(ctx);
+            }
+            Function<T>::callback(protected_ctx, protected_callback, typename T::Object(), 1, callback_arguments);
+        });
+
+        completion_func = std::move(completion_handler);
```
kneth: Why do you need to move twice?

src/js_sync.hpp
```diff
@@ -679,6 +687,59 @@ void SessionClass<T>::override_server(ContextType ctx, ObjectType this_object, A
...
+        bool callback_registered;
+        switch(direction) {
+            case Upload:
+                callback_registered = session->wait_for_upload_completion(std::move(completion_func));
+                break;
+            case Download:
+                callback_registered = session->wait_for_download_completion(std::move(completion_func));
+                break;
+        }
+        if (!callback_registered) {
```
nirinchev: Wouldn't it be better to handle that in a `default` case rather than introducing the extra bool variable?

tests/js/session-tests.js
```diff
@@ -1271,4 +1271,106 @@ module.exports = {
...
+            Realm.Sync.User.login(AUTH_URL, Realm.Sync.Credentials.nickname("admin1", true))
+                .then((admin1) => {
+                    const admin1Config = admin1.createConfiguration({
+                        schema: [schema],
+                        sync:  {
+                            url: REALM_URL,
+                            fullSynchronization: true
+                        }
+                    });
+                    Realm.open(admin1Config).then((admin1Realm) => {
```
nirinchev: I would rather move these continuation to the top-level scope. I guess you needed to write it like that to access some of the variables, but I would rather just return them. Something like:

```
.then(( admin1) => {
    // ...
    return Realm.open(admin1Config).then(realm => {
        return { admin1Realm: realm, admin1 };
    });
})
.then(({ admin1Realm, admin1 }) => {
    // ...
});
```

src/js_sync.hpp
cmelchior: The problem is that that registering the listener can apparently fail. At least that is how the sync API is documented, so it isn't an extra case as such.

Issue #2131, opened 2018-11-26
permissions().findOrCreate advertised on compile time, but not available on runtime


bmunkholm:
@programaths commented on [Mon Nov 19 2018](https://github.com/realm/realm-object-server/issues/403)

## Goals

Add a `Realm` level permission

## Expected Results

Permission get added with role if non existant

## Actual Results

```
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.867026142Z error: [http] HTTP response: 863878cd-5eb3-4eba-9872-8a873f1150dd
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.867045308Z {}
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.874735997Z error: Internal server error: TypeError: globalRealm.permissions(...).findOrCreate is not a function
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.874758609Z     at Realm.globalRealm.write (/realm-object-server/src/providers/legacy-authprovider.ts:26:50)
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.874773629Z     at LegacyAuthProvider.<anonymous> (/realm-object-server/src/providers/legacy-authprovider.ts:25:17)
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.874778000Z     at Generator.next (<anonymous>)
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.874791857Z     at fulfilled (/realm-object-server/dist/providers/legacy-authprovider.js:4:58)
ros-server_1_7fe0db58f502 | 2018-11-19T16:32:57.874795477Z     at <anonymous>
```

## Steps to Reproduce

Try to set permissions that way:

```typescript
/**
   *
   * @param server
   * @param {Realm} realm
   * @param userId
   * @param {permissionCallback} fn
   */
  async function setPermission(server:Server,realm:string,userId,fn){
    let globalRealm =await server.openRealm(realm);
    globalRealm.write(()=>{
      let permission = globalRealm.permissions().findOrCreate(`__User:${userId}`);
      fn(permission);
    });
    await globalRealm.close();
  }
```

## Code Sample

See above

## Version of Realm and Tooling

- Realm Object Server Version: 3.14.0
- Flavor:
  - [ ] Developer
  - [ ] Professional
  - [x] Enterprise
[...]

PR #2151, opened 2018-11-27
Add better error message for wrong Realm types

cmelchior:
Some methods are only available for Query-based Realms. Previously they would fail with non-sensical error messages.

The type of Realm is now checked and a better error message is thrown.

This fixes the problems reported in https://github.com/realm/realm-js/issues/2131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Using methods only available for Query-based Realms now throw a better error message if called on the wrong Realm type.
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Using methods only available for Query-based Realms now throw a better error message if called on the wrong Realm type.
```
kneth: I don't think our documentation talks about "Realm type". Maybe it can be confused with "Realm property type"? Is "Realm file type" or "Realm instance type" a better term?

cmelchior: Realm file type sounds fine with me

Issue #2155, opened 2018-11-30
Cannot serialize + deserialize an admin token user


kraenhansen:
## Goals

Deserializing and serializing a user authenticated with an admin token.

This is the root cause of https://github.com/realm/realm-studio-private/issues/19.

## Expected Results

Serializing followed by deserializing should yield a user with the same property values.

## Actual Results

The `isAdminToken` property is forgotten when serializing and disregarded when deserializing.

## Steps to Reproduce & Code Sample

```javascript
// Create a user from an admin token
const user = Realm.Sync.User.login(
    'http://localhost:9080',
    Realm.Sync.Credentials.adminToken('...'),
);

// Notice how it has `isAdminToken == true`
console.log(user.isAdminToken); // prints true

// This gets forgotten when serializing the user
const serialized = user.serialize();
console.log(serialized.isAdminToken); // prints undefined

// Even if we fake it by setting the value it's disregarded when deserializing
serialized.isAdminToken = true;
const reserialized = Realm.Sync.User.deserialize(serialized);
console.log(reserialized.isAdminToken); // prints undefined
```

I created a branch with a failing test https://github.com/realm/realm-js/tree/kh/serialize-deserialize-admin-token

## Version of Realm and Tooling

- Realm JS SDK Version: 2.19.1
- Node or React Native: N/A
- Client OS & Version: N/A
- Which debugger for React Native: N/A

PR #2156, opened 2018-11-30
Fix admin token user (de)serialization

nirinchev:
## What, How & Why?

This closes https://github.com/realm/realm-js/issues/2155

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -290,6 +290,11 @@ declare namespace Realm.Sync {
         isAdmin: boolean;
     }
 
+    interface SerializedTokenUser {
+        server: string;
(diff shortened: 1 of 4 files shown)


lib/index.d.ts
```diff
@@ -290,6 +290,11 @@ declare namespace Realm.Sync {
         isAdmin: boolean;
     }
 
+    interface SerializedTokenUser {
+        server: string;
+        adminToken: string;
```
kraenhansen: Technically ... Isn't the admin token a an `accessToken` that never expires?

Issue #2041, opened 2018-09-24
Unable to find a matching configuration of project :realm: None of the consumable configurations have attributes.


Amir-M7:
[##]([url](url
<img width="869" alt="screen shot 2018-09-24 at 4 07 03 pm" src="https://user-images.githubusercontent.com/17439071/45952440-e8f3bf00-c013-11e8-8fce-6490ee61495f.png">
)) Steps to Reproduce
when i want to run project with  react-native run-android shod this error in terminal : 
Unable to find a matching configuration of project :realm: None of the consumable configurations have attributes.
and then build failed

## Version of Realm and Tooling

- Realm JS SDK Version: ? 3.7
- Node or React Native: ? 0.55.4
- Client OS & Version: ? macOS
- Which debugger for React Native: ?/None

rayjadore:
another possible probelm could be that realm was not install, so try installing it seperately if it doesnt install use nvm to change the version of node possibly v9 and try install it again

Mohit-gwg:
Hello everyone, I am facing this same error which is listed above, I tried everyone suggested solutions but it didn't work in my case, 
error: 
Could not resolve all task dependencies for configuration ':app:debugRuntimeClasspath'.
   > Could not resolve project :realm.
     Required by:
         project :app
      > Unable to find a matching configuration of project :realm:
          - None of the consumable configurations have attributes.

Issue #1995, opened 2018-08-27
Deprecation warning from ./gradlew: Task.leftShift(Closure) method has been deprecated


kraenhansen:
## Goals

Running `./gradlew` in an app using Realm without deprecation warnings.

## Expected Results

No warnings when running `./gradlew`

## Actual Results

After following https://realm.io/docs/javascript/latest/ I get this warning as the first line when running any `./gradlew` command:

```
The Task.leftShift(Closure) method has been deprecated and is scheduled to be removed in Gradle 5.0. Please use Task.doLast(Action) instead.
        at build_d9y9ztx4qpkqkvbhyukdt2t0y.run(/Users/.../node_modules/realm/android/build.gradle:77)
```

## Steps to Reproduce

1. Follow the instructions on initializing a React Native app with Realm JS: https://realm.io/docs/javascript/latest/
2. Change directory into `android` and run any `./gradlew` command.

## Version of Realm and Tooling

- Gradle: 3.5.1
- Groovy: 2.4.10
- Ant: Apache Ant(TM) version 1.9.6 compiled on June 29 2015
- JVM: 1.8.0_144 (Oracle Corporation 25.144-b01)
- Realm JS SDK Version: 2.15.3
- Node or React Native: React Native 0.56.0
- Client OS & Version: Mac OS X 10.13.6 x86_64
- Which debugger for React Native: N/A

jaggs6:
Any update on this issue now that Gradle 5.0 has been launched. Its a breaking the 5.0 gradle builds.

naviet2308:
<img width="752" alt="screen shot 2018-11-30 at 23 46 42" src="https://user-images.githubusercontent.com/8268750/49302722-73680b80-f4fa-11e8-9983-7676f479b181.png">
Help me please.

cmelchior:
A quick fix should be to replace

```
task buildReactNdkLib(dependsOn: [downloadJSCHeaders,prepareRealmCore,prepareOpenSSL_x86,prepareOpenSSL_arm], type: Exec) {
    inputs.file('src/main/jni')
    outputs.dir("$buildDir/realm-react-ndk/all")
```

with 

```
task buildReactNdkLib(dependsOn: [downloadJSCHeaders,prepareRealmCore,prepareOpenSSL_x86,prepareOpenSSL_arm], type: Exec) {
    inputs.files('src/main/jni')
    outputs.dir("$buildDir/realm-react-ndk/all")
```

[...]

PR #2159, opened 2018-12-01
Add support for Gradle 5.0

cmelchior:
Closes https://github.com/realm/realm-js/issues/1995
Closes https://github.com/realm/realm-js/issues/2041

This PR adds Gradle 5.0 support and removes a bunch of build warnings as well. However, in order to support later versions of Gradle, I had to change how dependencies work. Before,  the `:realm` Android module actually pulled in `react-native` as well. This has now been changed to a compileOnly dependency meaning that people must add `react-native` themselves. This is done anyway when calling `react-native init`, so I doubt that anyone will notice.

TODO
- [x] More testing with various combinations of Gradle (Tested with 4.4 and 5.0 on app projects)
- [x] Check that metrics still work correctly


package.json
```diff
@@ -106,7 +106,8 @@
     "rimraf": "^2.6.1",
     "semver": "^5.1.0",
     "shelljs": "^0.7.7",
-    "shx": "^0.2.2"
+    "shx": "^0.2.2",
+    "react-native": "0.57.7"
```
cmelchior: I need to add react-native to `devDependecies` in order to be able to buid `/react-native/android` independently. I don't assume this a problem? @kneth ?

kneth: It means that you will have to install React Native even when you only develop for node.js. But I think it is acceptable.

react-native/android/analytics_template
```diff
@@ -75,7 +75,7 @@ class SendAnalyticsTask extends DefaultTask {
             {
                "event": "%EVENT%",
                "properties": {
-                  "token": "%TOKEN%,
+                  "token": "%TOKEN%",
```
kneth: Good catch 😈

PR #2171, opened 2018-12-12
Upgrading core to v5.12.5 and sync to v3.14.3

kneth:
Upgrading to latest core and sync.

PR #2172, opened 2018-12-12
Don't use Nan::SetNamedPropertyHandler

tgoyne:
Nan's wrapper for these functions incorrectly asserts that the property name argument is a `v8::String`, while it's actually a `v8::Name` (a superclass of String) and depending on the whim of the optimizer it may be a different concrete subclass. This resulted in the realm-js-private tests (and some non-test code) crashing with the error "FATAL ERROR: v8::String::Cast Could not convert to string" when run with a debug build of the library.

Fortunately the last time this API changed was in node 0.12 so there's no actual need to use Nan's wrapper.


kneth:
Without #2171, this PR will hang on Windows. I have kill the CI job, and everything but Windows is passing.

Issue #1981, opened 2018-08-21
createConfiguration could use the path in the sync url


kraenhansen:
## Goals

When using `createConfiguration` to create a configuration that syncs to something else than the `/default` Realm it would be nice if the user had a way to specify the path of the Realm on the server.

## Expected Results

Either the `config.path` could be appended to the server URL or perhaps a second string argument to the method.

## Actual Results

Calling `createConfiguration` on a user always returns a configuration object with the `/default` path.

## Steps to Reproduce

1. Authenticate a user
2. Call `user.createConfiguration({ path: "whatever" })`, log the result and observe that it has `/default` as the realm path of the `sync.url`.

## Version of Realm and Tooling

- Realm JS SDK Version: 2.14.2
- Node or React Native: Node
- Client OS & Version: N/A
- Which debugger for React Native: N/A

nirinchev:
I believe `path` is the local Realm path. If you want to modify the url, you should do it by specifying `sync.url`:
```ts
user.createConfiguration({
  sync: {
    url: "realm://foobar.cloud.realm.io/whatever"
  }
});
```

I haven't tested if it works though :)

kraenhansen:
Yeah - but as the main reason for the `createConfiguration` is to translate the server URL on the user into a `sync.url` an override like that is pretty pointless :) In that case, I would just create the configuration manually.

The `path` is for local realms as is, but one of my two suggestions is to change that to specify the path on the server if we're using `createConfiguration` - as the local path is ignored when opening a synced realm anyway.

kraenhansen:
Ahh okay - it seems as it that was actually fixed: https://github.com/realm/realm-js/issues/1392 ... I just wasn't aware of that.

kraenhansen:
Anyway - this issue is still relevant.

I suggest that `createConfiguration` takes a second optional string argument which is the virtual path of the Realm that will be synced against, which defaults to `"/default"`.

nirinchev:
I think that would be very weird API as you'll have two ways of specifying the same thing. I would rather handle relative urls in the `sync` object:

```ts
user.createConfiguration({
  sync: {
    url: "/whatever"
  }
});
```

We can easily detect that and concatenate with the user url.

kraenhansen:
@nirinchev that would be fine by me too.

kraenhansen:
I just had a need for this again today :) Glad to see it's in active progress.

Issue #1980, opened 2018-08-21
createConfiguration adds a colon to the servers hostname


kraenhansen:
## Goals

A more predictable behaviour from the `createConfiguration` method.

## Expected Results

I would expect the URL to not contain a : if no port was provided in the URL used for authentication.

## Actual Results

When calling `createConfiguration` on a `User` that was authenticated against a server without a port (such as `https://whatever.cloud.realm.io` in this example) a colon (:) is appended to the hostname of the `sync.url`:

    url: 'realms://whatever.cloud.realm.io:/default',
                                         ☝️😕

## Steps to Reproduce

1. Authenticate a user against a server with the default HTTPs port not specified.
2. Call `user.createConfiguration()` and log the result.

## Version of Realm and Tooling

- Realm JS SDK Version: 2.14.2
- Node or React Native: Node
- Client OS & Version: N/A
- Which debugger for React Native: N/A

PR #2175, opened 2018-12-13
Fix automatic conversion of Auth to Realm URLs

cmelchior:
Closes https://github.com/realm/realm-js/issues/1980
Closes https://github.com/realm/realm-js/issues/1981

Fixes how we do automatic conversions of Auth URLs to Realm URLs. This also adds support for defining URL's as relative to the auth host.


lib/user-methods.js
```diff
@@ -62,6 +62,18 @@ function checkObjectTypes(obj, types) {
     }
 }
 
+function normalizeSyncUrl(authUrl, syncUrl) {
+    let parsedAuthUrl = new URL(authUrl);
+    let secure = (parsedAuthUrl.protocol === 'https:')?'s':'';
```
bmunkholm:
```suggestion
    let secure = (parsedAuthUrl.protocol === 'https:') ? 's' : '';
```

lib/user-methods.js
```diff
@@ -629,18 +641,16 @@ const instanceMethods = {
...
-        }
 
-        // Create default config
-        let url = new URL(this.server);
-        let secure = (url.protocol === 'https:')?'s':'';
-        let port = (url.port === undefined)?'9080':url.port;
-        let realmUrl = `realm${secure}://${url.hostname}:${port}/default`;
+            if (config.sync.url) {
+                config.sync.url = new URL(config.sync.url, {})
+
```
bmunkholm:
```suggestion
```

cmelchior:
Ready for review.
@kraenhansen I'm not missing any use cases from you, right?

lib/user-methods.js
```diff
@@ -62,6 +62,19 @@ function checkObjectTypes(obj, types) {
...
+function normalizeSyncUrl(authUrl, syncUrl) {
+    let parsedAuthUrl = new URL(authUrl);
+    let secure = (parsedAuthUrl.protocol === 'https:') ? 's' : '';
+    // Inherit ports from the Auth url
+    let port = (parsedAuthUrl.port === undefined) ? "" : `:${parsedAuthUrl.port}`;
+    let baseUrl = `realm${secure}://${parsedAuthUrl.hostname}${port}`;
+    if (syncUrl) {
+        return new URL(syncUrl, baseUrl, false).toString();
+    } else {
+        return baseUrl + "/default";
```
nirinchev: You can avoid the multiple returns if you supply a default value to the `syncUrl`. Something like:

```js
if (!syncUrl) {
    syncUrl = "/default";
}

return new URL(syncUrl, baseUrl, false).toString();
```

lib/user-methods.js
```diff
@@ -62,6 +62,19 @@ function checkObjectTypes(obj, types) {
     }
 }
 
+function normalizeSyncUrl(authUrl, syncUrl) {
+    let parsedAuthUrl = new URL(authUrl);
+    let secure = (parsedAuthUrl.protocol === 'https:') ? 's' : '';
```
nirinchev: This seems like an overly clever way to save a few letters 😛I think it'd be much more obvious if we were explicit about what this is and why we need it. Something like:

```js
const realmProtocol = parsedAuthUrl.protocol === "https:" ? "realms" : "realm";
```

lib/user-methods.js
```diff
@@ -62,6 +62,19 @@ function checkObjectTypes(obj, types) {
     }
 }
 
+function normalizeSyncUrl(authUrl, syncUrl) {
+    let parsedAuthUrl = new URL(authUrl);
```
nirinchev: These all should be `const` (although I know that throughout the codebase we have been very liberal with the usage of `let`).

lib/user-methods.js
```diff
@@ -62,6 +62,19 @@ function checkObjectTypes(obj, types) {
...
+    let secure = (parsedAuthUrl.protocol === 'https:') ? 's' : '';
+    // Inherit ports from the Auth url
+    let port = (parsedAuthUrl.port === undefined) ? "" : `:${parsedAuthUrl.port}`;
```
nirinchev: You can safely check the port for falsy-ness and save a few characters:

```suggestion
    let port = parsedAuthUrl.port ? `:${parsedAuthUrl.port}` : "";
```

lib/user-methods.js
```diff
@@ -629,18 +642,16 @@ const instanceMethods = {
...
             }
-        }
 
-        // Create default config
-        let url = new URL(this.server);
-        let secure = (url.protocol === 'https:')?'s':'';
-        let port = (url.port === undefined)?'9080':url.port;
-        let realmUrl = `realm${secure}://${url.hostname}:${port}/default`;
+            if (config.sync.url) {
+                config.sync.url = new URL(config.sync.url, {})
```
nirinchev: Why do we need to change the type here?

kraenhansen:
@cmelchior I was looking for how to use this in the documentation https://realm.io/docs/javascript/2.22.0/api/Realm.Sync.User.html. I'm afraid we forgot to update the docs with this change in behaviour.

Issue #1832, opened 2018-05-28
Bug: Should soft fail and continue on fallocate error on AWS Lambda


heri16:
## Goals
Trying to run realm-js on nodejs 8.10 AWS Lambda.

## Expected Results
Should be able to open a realm path on /tmp

## Actual Results
```
{
    "errorMessage": "posix_fallocate() failed: Operation not permitted",
    "errorType": "Error",
    "stackTrace": [
        "Function.open (/var/task/node_modules/realm/lib/extensions.js:72:47)",
        "exports.handler (/var/task/api/realm.js:45:25)"
    ]
}
```

## Steps to Reproduce
1. Create a serverless framework project.
1. npm install realm.
1. Deploy the function below: `sls deploy -f realm`
1. Try invoking the function below: `sls invoke -f realm --data "{'test': 'ok'}"`

## Code Sample
    'use strict'
    
    // See: https://realm.io/docs/javascript/latest/#getting-started
    
    const process = require('process')
    
    const debug = require('debug')('realm')
    const Realm = require('realm')
    
    const CarSchema = {
      name: 'Car',
      properties: {
        make: 'string',
        model: 'string',
        miles: { type: 'int', default: 0 },
      },
    }
    
    // Define new ES6 class
    class Person {
      get fullName() {
        return `${this.firstName} ${this.lastName}`
      }
    
      static get schema() {
        return {
          name: 'Person',
          properties: {
            name: 'string',
            birthday: 'date',
            cars: 'Car[]',
            picture: 'data?', // optional property
          },
        }
      }
    }
    
    exports.handler = async (event, context) => {
      console.log('Received event', JSON.stringify(event, 3)) // eslint-disable-line no-console
      console.log('Received context', JSON.stringify(context, 3)) // eslint-disable-line no-console
    
      let realm
      try {
        process.chdir('/tmp')
        realm = await Realm.open({
          path: '/tmp/cars.realm',
          inMemory: true,
          schema: [CarSchema, Person],
        })
      } catch (err) {
        console.error(err) // eslint-disable-line no-console
        throw err
      }
    
      // Create Realm objects and write to local storage
      realm.write(() => {
        const myCar = realm.create('Car', {
          make: 'Honda',
          model: 'Civic',
          miles: 1000,
        })
        myCar.miles += 20 // Update a property value
      })
    
      // Query Realm for all cars with a high mileage
      const cars = realm.objects('Car').filtered('miles > 1000')
    
      // Will return a Results object with our 1 car
[...]

PR #2182, opened 2018-12-18
Upgrading core doesn't help AWS Lambda

kneth:
After release v2.22.0-beta.1 we still observe `Operation not permitted` on AWS Lambda. It means that #1832 isn't solved. If you execute `fallocate -l 1M foobar.img` on AWS Lambda, the same error is observed. AWS support hasn't been able to help us yet.


kneth:
@bmunkholm Please review.

PR #2201, opened 2019-01-08
Updating CHANGELOG.md

kneth:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Fixed
 * Fixed building on Android. ([#2189](https://github.com/realm/realm-js/issues/2189), since v2.22.0-beta.2)
 * Fix an occasional crash due to an uncaught `realm::IncorrectThreadException` when a client reset error occurs. ([#2193]()https://github.com/realm/realm-js/pull/2193)
+* Fix a bug in the sync worker to use correct default values. Thanks to @radovanstevanovic. ([#2195](https://github.com/realm/realm-js/pull/2195), since v2.2.2)
```
bmunkholm: It's not clear for me what the end user would see before this fix?

kneth: I believe we set the value (see https://github.com/realm/realm-js-private/blob/f22b064b7f20047d4cf33b01ee4070a69a308b69/spec/notifier_tests.ts#L721) when we use `Realm.Worker`. We haven't seen any reports, and it is unlikely that any user has used it directly.

PR #2206, opened 2019-01-11
The new distribution form for Windows doesn't require SSH.

kneth:


fealebenpae:
You need to modify https://github.com/realm/realm-js/blob/master/scripts/download-realm.js to use the new path without the SHA (and delete the call to `git ls-remote`).

kneth:
@fealebenpae Thanks for point it out. I remove the usage of the method when releasing v2.22.0 but forgot to remove the method :-)

PR #2220, opened 2019-01-16
Upgrade to core v5.13.0 and sync v3.14.14

kneth:
* Upgrade to Realm Core v5.13.0
* Upgrade to Realm Sync v3.14.14
* Upgrade object store to head of `master`
* Update query language tutorial


docs/tutorials/query-language.md
```diff
@@ -66,7 +66,7 @@ Use parentheses `()` and the `&&`/`AND` and `||`/`OR` operators to compose queri
...
+Normally queries can be written with variable substitution so that the syntax of dates is taken care of by Realm. However, sometimes it fits better to compose a entirely string based query. In that case the syntax follows the formats `YYYY-MM-DD@HH:MM:SS:N` and `YYYY-MM-DDTHH:MM:SS:N` where the `:N` suffix specifies nanoseconds but can be omitted (defaulting to 0). If preferred, a programmatic format is also supported `TSS:NS` which is a literal `T` followed by seconds since epoch, and nanoseconds modifier. In both formats, negative nanoseconds are considered invalid syntax. Timestamps are not quoted like strings. Due to platform limitations, using the first syntax is not supported for dates pre epoch on windows and pre 1901 on other platforms.
```
bmunkholm:
```suggestion
Normally queries can be written with variable substitution so that the syntax of dates is taken care of by Realm. However, sometimes it fits better to compose a entirely string based query. In that case the syntax follows the formats `YYYY-MM-DD@HH:MM:SS:N` or `YYYY-MM-DDTHH:MM:SS:N` where the `:N` suffix specifies nanoseconds but can be omitted (defaulting to 0). If preferred, a programmatic format is also supporting `TSS:NS` which is a literal `T` followed by seconds since epoch, and nanoseconds modifier. In both formats, negative nanoseconds are considered invalid syntax. Timestamps are not quoted like strings. Due to platform limitations, using the first syntax is not supported for dates pre epoch on windows and pre 1901 on other platforms.
```

bmunkholm: just changed to words "supported" -> "supporting" and 
"and" to "or" around the two formats.

PR #2226, opened 2019-01-18
Add the notifier and adapter to the node bulid

tgoyne:
This pulls in https://github.com/realm/realm-object-store/pull/760, removes the files that are now in the object store, and updates the build system to actually build it.


tgoyne:
Tests are failing because the react and android projects need the path to json.hpp fixed. I'll get back to this later.

tgoyne:
Closing in favor of #2343.

PR #2231, opened 2019-01-22
Add support for proxy when downloading realm-sync-cocoa

ezjoke:
## What, How & Why?

I had trouble building for iOS due to network (proxy) issues. I tried the fix proposed here : https://realm.io/docs/javascript/latest/#cannot-download-realm-sync-cocoa but couldn't make it work.

I believe this patch fixes issues related to proxy by adding a proxy configuration to the node-fetch call for downloading the realm-sync-cocoa package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "decompress": "^4.2.0",
     "deepmerge": "2.1.0",
     "fs-extra": "^4.0.2",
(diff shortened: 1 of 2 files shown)


kneth:
@ezjoke Thanks for the contribution - and thanks for signing our CLA.

PR #2241, opened 2019-01-28
Reduce bundled size from 42KB to 24KB

lebedev:
## What, How & Why?
By using `source-map-explorer` with my production bundle I've noticed that `realm` includes `browser` folder, which makes it's bundled size 42KB total.

Current bundled realm-js

Metro Bundler of React Native lacks tree shaking feature, however it's possible to exclude `browser` folder from production bundles by making it's require conditional with `__DEV__`.

Bundled realm-js with this PR applied

I tested this patch on Android in production mode and in debug mode with Chrome debugger. It works fine.
I didn't test it on iOS, but it should works without any changes too.
And I didn't test with VSCode debugger. I haven't VSCode and can't test this patch with it. It could break if with VSCode debugger `__DEV__` global isn't set for some reason. But I assume that it should work as that global is set by React Native bundler, and not by VScode.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -110,7 +110,12 @@ switch(context) {
 
     case 'chromedebugger':
     case 'vscodedebugger':
-        realmConstructor = require('./browser').default; // (exported as ES6 module)
+        // This condition is for stripping "browser" folder from production bundles.
+        if (__DEV__) {
+            realmConstructor = require('./browser').default; // (exported as ES6 module)
+        } else {
+            throw new Error('Can\'t use debugger if __DEV__ isn\'t true.');
+        }
         break;
 }
 


angly-cat:
Another catch: this PR would break remote debugging if one disables "JS Dev Mode" in React Native Dev Settings on Android.

image

However, even without this PR I wasn't able to get remote debugging working with JS Dev Mode disabled. I kept getting `Realm failed to connect to the embedded debug server inside the app. If attempting to use Chrome debugging from a device, ensure the device is reachable on the same network as this machine.` with emulator from the same "network" and all ports properly forwarded/reversed.

Is it possible, though?

cmelchior:
Hi @angly-cat 

Sorry for the late reply. 
Unfortunately, the `browser` folder is required when you want to enable Remote Debugging. The reason is that this folder contains the required infrastructure to make React Native debugging work (it is rather complicated). 

It will be loaded if the environment is detected as the `chromedebugger` or the `vsdebugger`. If you don't enable debugging you can thus safely remove it, but otherwise, it is required for now.

React Native is working on a new architecture that will hopefully make it possible to remove it in the future, but there is no definite timeline from Facebook about when these changes might land.

angly-cat:
Hey, @cmelchior!

Thanks for the reply. I may have explained my reasoning badly.

The thing is if one builds a React Native app (that uses Realm) in release/production mode, where Remote Debugging can't be used by definition, `browser` folder from Realm unconditionally gets bundled in a JS bundle of the app, despite it will never be used, making it a dead weight.

By adding the condition there, it's possible to exclude `browser` folder from release/production bundles.

That condition wouldn't change behavior of Remote Debugging at all. It just allows that folder to be tree shaked when unused.

kneth:
@lebedev I have merged #2362 which I used to run your PR through our CI system.

PR #2243, opened 2019-01-31
Add support for ssl configuration to addListener

cmelchior:
Add public API support controlling SSL to the global notifier.

- [x] Merge https://github.com/realm/realm-js-private/pull/510
- [x] Changelog

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -555,11 +555,11 @@ declare namespace Realm.Sync {
     /**
      * @deprecated, to be removed in future versions
      */
-    function addListener(serverURL: string, adminUser: Realm.Sync.User, regex: string, name: RealmListenerEventName, changeCallback: (changeEvent: ChangeEvent) => void, sslConfiguration?: Realm.Sync.SSLConfiguration): void;
+    function addListener(serverURL: string, adminUser: Realm.Sync.User, regex: string, name: RealmListenerEventName, changeCallback: (changeEvent: ChangeEvent) => void): void;
     /**
      * @deprecated, to be removed in future versions
      */
-    function addListener(serverURL: string, adminUser: Realm.Sync.User, regex: string, name: RealmListenerEventName, changeCallback: (changeEvent: ChangeEvent) => Promise<void>, sslConfiguration?: Realm.Sync.SSLConfiguration): void;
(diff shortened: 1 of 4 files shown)


cmelchior:
Yeah. I had the same doubt as you about the API. Creating a new function with a single config object seems like a good approach and then we can deprecate the other method.

cmelchior:
@tgoyne This API does not appear to be used, and looking at the code I'm not sure it even works: https://github.com/realm/realm-js/blob/master/docs/sync.js#L141 Can it be safely removed?

tgoyne:
Yes, that function is used and works. There's tests for it in realm-js-private.

lib/notifier.js
```diff
@@ -256,10 +256,22 @@ class Listener {
...
+            // SSL Configuration not supported in the old API
+        }
+    }
+
     if (!listener) {
-        listener = new Listener(this, server, user);
+        listener = new Listener(this, config);
     }
-    return listener.add(regex, event, callback);
+    return listener.add(config.filterRegex, event, callback);
```
nirinchev: With the new API, wouldn't `user` and `regex` effectively be `event` and `callback`?

lib/index.d.ts
```diff
@@ -543,8 +543,25 @@ declare namespace Realm.Sync {
...
+    }
+
+    /**
+     * @deprecated, to be removed in future versions
+     */
+    function addListener(serverURL: string, adminUser: Realm.Sync.User, regex: string, name: RealmListenerEventName, changeCallback: (changeEvent: ChangeEvent) => void, sslConfiguration?: Realm.Sync.SSLConfiguration): void;
+    /**
+     * @deprecated, to be removed in future versions
+     */
+    function addListener(serverURL: string, adminUser: Realm.Sync.User, regex: string, name: RealmListenerEventName, changeCallback: (changeEvent: ChangeEvent) => Promise<void>, sslConfiguration?: Realm.Sync.SSLConfiguration): void;
```
nirinchev: If ssl config is not supported in the old API, should we remove it from the method signature?

lib/notifier.js
cmelchior: Yes, good catch 👍

PR #2227, opened 2019-01-18
Integration tests

kraenhansen:
## What, How & Why?

This PR fixes a large part of https://github.com/realm/realm-js/issues/1929 by adding an integration test harness to run a suite of tests in a list of environments:

- NodeJS (v10 on Mac and v10 + v8 on Linux)
- React Native on Android 
- React Native on iOS
- Electron in the main process (v4 on Linux)
- Electron in the renderer process (v4 on Linux)

Notice the [README.md](https://github.com/realm/realm-js/blob/kh/integration-tests/integration-tests/README.md) in the `integration-tests` directory.

Note: I've added a comment on https://github.com/realm/realm-js/issues/1929#issuecomment-460232059 naming what's missing (from my point of view) before we have a trustworthy integration test harness.

## ☑️ ToDos
* [x] Implement a test harness for the five (actually just three) environments.
* [x] Copy and adjust (some of) the [realm-tests.js](https://github.com/realm/realm-js/blob/master/tests/js/realm-tests.js) to the integration test suite.
* [x] Create scripts to run the integration tests on CI (using the JUnit XML reporter).
* [x] Add running the integration tests to the Jenkinsfile.
* [ ] 📝 Changelog entry
* [x] 📝 `Compatibility` label is updated or copied from previous entry
* [x] 🚦 Tests
* [x] 📝 Public documentation PR created or is not necessary
* [x] 💥 `Breaking` label has been applied or is not necessary

Issue #2216, opened 2019-01-15
Imperative API for schema manipulation


kraenhansen:
## Goals

To manipulate a Realms schema, Realm Studio currently edits the schema, closes and reopens the Realm with the mutated schema. It would be great if it could create, update and delete properties and classes with an imperative API on a Realm instance.

## Expected Results

I would like the following methods available on an instance of `Realm`:

### Methods for class manipulation
```typescript
function createClass(name: string, properties: Realm.PropertiesTypes, primaryKey?: string): void;
function updateClass(name: string, properties: Realm.PropertiesTypes): void; // ⚠ Might delete or modify existing properties
function deleteClass(name: string): void;
function renameClass(oldName: string, newName: string): void;
```

### Methods for property manipulation
```typescript
function createProperty(className: string, name: string, schema: Realm.ObjectSchemaProperty): void;
function updateProperty(className: string, name: string, schema: Realm.ObjectSchemaProperty): void;
function deleteProperty(className: string, name: string): void;
function renameProperty(className: string, oldName: string, newName: string): void;
```

bmunkholm:
@kraenhansen Is this done now?

kraenhansen:
I believe this can be considered done with #2251 - but we still need tests on what happens when the user calls `_updateSchema` with destructive schema changes and at some point we should consider making this API properly documented and public.

PR #2251, opened 2019-02-12
Imperative aka. dynamic schema manipulation

kraenhansen:
## What, How & Why?

This solves part of #2216: Creating object schemas and object schema properties.

Note: This PRs branch is based on https://github.com/realm/realm-js/pull/2227 - once that is merged this should be rebased from the updated master branch before being merged.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


src/js_realm.hpp
```diff
@@ -1285,5 +1292,137 @@ void RealmClass<T>::privileges(ContextType ctx, ObjectType this_object, Argument
 #endif
 }
 
+class InvalidNameException : public std::invalid_argument {
+    public:
+        InvalidNameException(const std::string& __s) : std::invalid_argument(__s) {}
```
tgoyne: `__s` isn't a legal variable name (it's reserved for the implementation and using it is UB).

tgoyne: There also doesn't appear to be any reason for this type to exist at all; there's no benefit to throwing a custom subclass of `invalid_argument` rather than just throwing `invalid_argument` unless there's something that'd want to catch that specific subclass, which will never be the case when the calling code is JS.

kraenhansen:
> It's not really obvious to me why any of this code needs to exist rather than just exposing `Realm::update_schema()` (which was originally intended to be part of the public API for JS). It seems like it'd be easier to just write the code which manipulates the schema objects in JS?

These are the first two of a couple of methods (see #2216) that I would like to expose. Renaming a class or property cannot easily be represented by a call to `update_schema`, but in principle I agree - it would be easier to implement an exposure of `update_schema` directly. Renaming could then be handled via a special method as a special case of schema manipulation.

I didn't know it was originally intended to be part of the public API, is there any reason it was put on hold, except time constraints?

kraenhansen:
@tgoyne I've updated the code to simply wrap the `update_schema` method - do you see any limitations that I need to explicitly check for regarding destructive schema changes & sync?

src/js_realm.hpp
kraenhansen: I'll remove that - no need for it now that I've implemented a simpler solution.

PR #2257, opened 2019-02-27
Prepare v2.24.0

kneth:


CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added a `_updateSchema` method on an Realm instance to perform schema manipulation. Specifically creating an object schema and a property on an existing object schema. (partly solving [#2216](https://github.com/realm/realm-js/issues/2216))
-* Add support for react-native 0.58 ([#2239](https://github.com/realm/realm-js/issues/2239)).
+* Add the method `Realm._updateSchema` to perform schema manipulation. Specifically creating an object schema and a property on an existing object schema. (partly solving [#2216](https://github.com/realm/realm-js/issues/2216))
```
cmelchior: Why are we mentioning a private method in the public changelog? Shouldn't it be moved to Internal?

kraenhansen: In my opinion it's too easy to confuse `Realm._updateSchema` for a static method on the `Realm` class instead of a method on an instance of `Realm` (which it is). That was why I formulated the original note as I did.

```suggestion
* Add the method `realm._updateSchema` to perform schema manipulation on an open Realm. Specifically creating an object schema and a property on an existing object schema are tested. It's undocumented as it's not fully tested and the API is subject to change. (partly solving [#2216](https://github.com/realm/realm-js/issues/2216))
```

CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
...
+* Add support for React Native v0.58. ([#2239](https://github.com/realm/realm-js/issues/2239))
 
 ### Fixed
 * Fixed an assertion failure after creating an object with a null int primary key in a synchronized Realm. ([#3227](https://github.com/realm/realm-core/pull/3227))
```
```suggestion
* Fixed an assertion failure when using a synchronized Realm and an object was created after another object was created with an int primary key of `null`. ([#3227](https://github.com/realm/realm-core/pull/3227))
```

PR #2260, opened 2019-02-28
TS definition and API doc for setSyncLogger

### LINKED ISSUE #2125: Document Realm.Sync.setSyncLogger
## Goals

To implement https://github.com/realm/realm-studio-private/issues/16 it would be great to get a way to intercept messages logged by the sync client.

## Expected Results

It would be great to be able to set a callback which would get called when the client wants to log a message, overriding the default behaviour of printing to the console:

```typescript
enum LogLevel {
    All = 'all'
    Trace = 'trace'
    Debug = 'debug'
    Detail = 'detail'
    Info = 'info'
    Warn = 'warn'
    Error = 'error'
    Fatal = 'fatal'
    Off = 'off'
};

Realm.Sync.setLogCallback((level: LogLevel, message: string) => {
    // Do whatever with the message ...
});

// Because `setLogCallback` was called no messages are logged to the console.
```

kneth:
This closes #2125 and https://github.com/realm/realm-js-private/issues/517

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* Add `Realm.Sync.setSyncLogger()` to TypeScript definitions and API documentation. ([#2125](https://github.com/realm/realm-js/issues/2125) and [realm-js-private#517](https://github.com/realm/realm-js-private/issues/517), since v2.0.0)
 
 ### Compatibility
 * Realm Object Server: 3.11.0 or later.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add `Realm.Sync.setLogger()` to capture the sync client's log. ([#2125](https://github.com/realm/realm-js/issues/2125) and [realm-js-private#517](https://github.com/realm/realm-js-private/issues/517))
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


kraenhansen:
Because the API has not been published or documented yet we should be able to easily change its name. It's only used [a handful of places](https://github.com/search?p=1&q=org%3Arealm+Realm.Sync.setSyncLogger&type=Code) in our own code. I would suggest that we simplify the naming before it goes out as a documented API:

```
Realm.Sync.setLogger()
```

nirinchev:
This doesn't need to be a breaking change. We can have a `setSyncLogger` and `setLogger` that do the same thing.

kraenhansen:
> This doesn't need to be a breaking change. We can have a `setSyncLogger` and `setLogger` that do the same thing.

That's also a possibility, but it would be a breaking change to first publish types for `setSyncLogger` and remove them again when renaming. And it's probably not the best idea to have two functions with slightly different names and same behaviour. We could leave an undocumented `setSyncLogger` (with a deprecation warning at runtime) which calls a documented `setLogger` - that would be a fine solution for me.

I'm afraid we're bikeshedding here. We could just merge this, but if we want to follow a process of implementing a fast solution and documenting it when we learn from using it, we should be willing to take the hit of adapting the fast solution to the stuff we're learned along the way. If not - we should just document and provide types for functions like this, right away.

cmelchior:
I agree with @kraenhansen ... having the method named `setLogger` just feels more natural and easier. It also means we can extend it in the future with logs from pure JS (which would be nice when debugging).

There is no reason to pollute the API with methods we intend to remove anyway, especially if they haven't been made public yet.

nirinchev:
We can't remove the method, but we can leave it undocumented and document just the new one.

kneth:
Required to be released together with https://github.com/realm/realm-js-private/pull/518

PR #2262, opened 2019-03-01
Add Realm.Sync.getExistingSessions() and Realm.Sync.SyncConfiguration._sessionStopPolicy

### LINKED ISSUE #2248: Expose a hasActiveSessions property
In ROS tests, we call `clearTestState` after almost all tests, but that's subject to timing-related issues as there might be some sessions that have not yet been inactivated. Currently, we work around it by waiting arbitrarily (100 ms) before calling the method. Instead, realm-js should expose a `hasActiveSessions` property which we can periodically poll to ensure we don't hit assertion failures.

Also, it would be useful to be able to specify the session stop policy - in tests, I would rather sessions are terminated immediately after the Realm is closed.

cmelchior:
Closes #2248 

TODO
- [x] Wait for https://github.com/realm/realm-object-store/pull/774 and redirect back to OS master

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -254,7 +254,7 @@ class Sync {
     static initiateClientReset(path) {}
 
     /**
-     * Returns `true` if Realm still has a references to any sync sessions regardless of their state.
+     * Returns `true` if Realm still has a reference to any sync sessions regardless of their state.
      * If `false` is returned it means that no sessions currently exists.
      */
     static hasExistingSessions() {}
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -255,7 +255,7 @@ class Sync {
 
     /**
      * Returns `true` if Realm still has a reference to any sync sessions regardless of their state.
-     * If `false` is returned it means that no sessions currently exists.
+     * If `false` is returned it means that no sessions currently exist.
      */
     static hasExistingSessions() {}
 }
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -257,7 +257,7 @@ class Sync {
      * Returns `true` if Realm still has a reference to any sync sessions regardless of their state.
      * If `false` is returned it means that no sessions currently exist.
      */
-    static hasExistingSessions() {}
+    static _hasExistingSessions() {}
 }
 
 /**
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -246,6 +246,12 @@ class Sync {
      * }
      */
     static initiateClientReset(path) {}
+
+    /**
+     * Returns `true` if Realm still has a references to any sync sessions regardless of their state.
+     * If `false` is returned it means that no sessions currently exists.
+     */
+    static hasExistingSessions() {}
 }
 
 /**
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -568,6 +568,7 @@ declare namespace Realm.Sync {
     function setLogLevel(logLevel: 'all' | 'trace' | 'debug' | 'detail' | 'info' | 'warn' | 'error' | 'fatal' | 'off'): void;
     function setUserAgent(userAgent: string): void;
     function initiateClientReset(path: string): void;
+    function hasExistingSessions(): boolean;
     function reconnect(): void;
 
     /**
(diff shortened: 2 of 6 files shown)


docs/sync.js
```diff
@@ -252,6 +252,12 @@ class Sync {
      * }
      */
     static initiateClientReset(path) {}
+
+    /**
+     * Returns `true` if Realm still has a references to any sync sessions regardless of their state.
```
```suggestion
     * Returns `true` if Realm still has a reference to any sync sessions regardless of their state.
```

docs/sync.js
```diff
@@ -252,6 +252,12 @@ class Sync {
...
+     * If `false` is returned it means that no sessions currently exists.
```
```suggestion
     * If `false` is returned it means that no sessions currently exist.
```

src/js_sync.hpp
```diff
@@ -1100,6 +1108,22 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
         config.sync_config->error_handler = std::move(error_handler);
         config.sync_config->is_partial = is_partial;
 
+        SyncSessionStopPolicy session_stop_policy = SyncSessionStopPolicy::AfterChangesUploaded;
+        ValueType session_stop_policy_value = Object::get_property(ctx, sync_config_object, "_sessionStopPolicy");
+        if (!Value::is_undefined(ctx, session_stop_policy_value)) {
+            std::string stop_session = Value::validated_to_string(ctx, session_stop_policy_value, "_sessionStopPolicy");
+            if (stop_session == std::string("immediately")) {
```
nirinchev: Not a C++ expert, but wouldn't this be more naturally written as a `switch` rather than a bunch of if-s?

bmunkholm: because in C++ you can't use switch on strings  :-)

docs/sync.js
```diff
@@ -252,6 +252,12 @@ class Sync {
      * }
      */
     static initiateClientReset(path) {}
+
+    /**
+     * Returns `true` if Realm still has a reference to any sync sessions regardless of their state.
+     * If `false` is returned it means that no sessions currently exist.
+     */
+    static hasExistingSessions() {}
```
bmunkholm: I know this was added to help us with tests. What's the use case for end users?

cmelchior: Very little. It is potentially helpful in a Client Reset situation or elsewhere where you want to be 100% sure the Realm is fully closed.

lib/index.d.ts
```diff
@@ -437,6 +437,11 @@ declare namespace Realm.Sync {
 
     type ErrorCallback = (session: Session, error: SyncError) => void;
     type SSLVerifyCallback = (sslVerifyObject: SSLVerifyObject) => boolean;
+    enum SessionStopPolicy {
```
bmunkholm: Why make a string enum and not a numeric?

src/js_sync.hpp
bmunkholm: But I don't get why a string enum is used to store a config value instead of a numeric enum, which would remove a number of failure options?

lib/index.d.ts
nirinchev: Because js doesn't have real enums 😑 This is useable only from typescript projects and people using plain js would be annoyed to use some magic numbers. We can emulate an enum for sure, but I guess it's not very idiomatic.

docs/sync.js
bmunkholm: I would consider making it a private API then. We should limit the API surface to only truly useful features. We can always make it public later if we think it's valuable.

src/js_sync.hpp
cmelchior: JS doesn't support enums, so historically in the JS API enum arguments are always parsed in as strings. You could argue we could create a "type-safe" enum object in JS, but it would only really help with auto-complete in editors. The compiler would still let you do awful things.

Personally, I think having enum objects in JS would be nice, but that would probably be something we need to consider as a general API enhancement. Note, that TypeScript do have an enum definition.

lib/index.d.ts
cmelchior: See https://github.com/realm/realm-js/pull/2262#discussion_r262000800

docs/sync.js
cmelchior: Sure, we can do that.

PR #2264, opened 2019-03-04
Fix broken user auth functions when running in electron

tgoyne:
And actually run the electron tests on CI. The test.sh/package.json changes are due to that it turned out to be only building from source when no release with the current version existed. This isn't an issue for the other targets because RN doesn't use a node-pre-gyp-built binary and node doesn't have to go through the pack/install dance to work around issues with symlinks.

`fetchWithTimeout()` depending on some non-standard XMLHttpRequest bits which happen to be present in react-native but not electron so it was making invalid requests.

### DIFF
--- a/lib/user-methods.js
+++ b/lib/user-methods.js
@@ -90,6 +90,9 @@ function fetchWithTimeout(input, init) {
                 url: xhr.responseURL,
                 headers: xhr.responseHeaders
             };
+            if (!options.headers) {
+                options.headers = {'content-type': xhr.getResponseHeader('content-type')};
+            }
             const body = 'response' in xhr ? xhr.response : xhr.responseText;
             resolve(new Response(body, options));
         };
@@ -97,7 +100,7 @@ function fetchWithTimeout(input, init) {
         xhr.ontimeout = () => reject(new TypeError('Network request failed'));
         xhr.open(request.method, request.url, true);
         request.headers.forEach((value, name) => xhr.setRequestHeader(name, value));
-        xhr.send(typeof request._bodyInit === 'undefined' ? null : request._bodyInit);
+        xhr.send(typeof request._bodyInit === 'undefined' ? init.body : request._bodyInit);
     });
 }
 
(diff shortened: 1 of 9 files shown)


scripts/test.sh
```diff
@@ -365,21 +369,22 @@ case "$TARGET" in
   stop_server
   ;;
 "electron")
-  npm install --no-save
+  npm ci
```
kraenhansen: Awesome! I didn't know that command :)

lib/user-methods.js
```diff
@@ -90,14 +90,17 @@ function fetchWithTimeout(input, init) {
                 url: xhr.responseURL,
                 headers: xhr.responseHeaders
             };
+            if (!options.headers) {
+                options.headers = {'content-type': xhr.getResponseHeader('content-type')};
+            }
             const body = 'response' in xhr ? xhr.response : xhr.responseText;
             resolve(new Response(body, options));
         };
         xhr.onerror = () => reject(new TypeError('Network request failed'));
```
nirinchev: It's not directly related to this PR, but it would be nice to flow the original error as we're going to print that if the promise is rejected and it'll be more informative than the vague `Network request failed`. Also, I'm not a js dev, so maybe I'm wrong, but throwing a TypeError feels a bit odd.

lib/user-methods.js
```diff
@@ -90,14 +90,17 @@ function fetchWithTimeout(input, init) {
...
         xhr.ontimeout = () => reject(new TypeError('Network request failed'));
```
nirinchev: Similarly, it would be nice to reject that with a more meaningful message like `Network request timed out after xxx ms`.

lib/user-methods.js
tgoyne: I have no idea why, but TypeError is actually what the spec says should be thrown.

Probably some historical thing where some library that originally implemented fetch() threw it for anything that didn't result in valid json and that got enshrined at the correct behavior.

scripts/test.sh
tgoyne: It's a fairly recent addition that makes it just install the versions from package-lock rather than rerunning dependency resolution. Seems to knock a few seconds off installation for us.

kraenhansen:
I still see a broken test on CI when building master:

```
1) UserTests : testLoginTowardsMisbehavingServer
   Error: 'Network request failed' does not equal expected value 'Could not authenticate: Realm Object Server didn't respond with valid JSON'
```

It seems that the test expects a different error message from the fetch client when communicating with a misbehaving ROS.

PR #2265, opened 2019-03-04
Credentials and users integration tests

kraenhansen:
## What, How & Why?

This moves the tests related to credentials and authentication.

## ☑️ ToDos
* [ ] Move more of the tests
* [ ] Remove the existing tests to minimize the maintenance burden.
* [ ] Fix "with-ros" to read the url of the server from the environment global and figure out a way to get that passed into the React Native app.
* [ ] 📝 Changelog entry
* [x] 🚦 Tests


integration-tests/tests/src/sync-user.js
```diff
@@ -0,0 +1,41 @@
...
+    it("defines the login function", () => {
+        expect(Realm.Sync);
+        expect(Realm.Sync.User);
+        expect(Realm.Sync.User.login).to.be.a("function");
+    });
+
+    withRos.it("can authenticate using the nickname provider", async function () {
+        const nickname = `realm-js-tests-${uuid()}`;
+        const credentials = Realm.Sync.Credentials.nickname(nickname);
+        expect(Realm.Sync.User.login).to.be.a("function");
```
nirinchev: This is tested in another test, so seems redundant here.

integration-tests/tests/src/utils/with-ros.js
```diff
@@ -0,0 +1,63 @@
...
+        'Define "REALM_OBJECT_SERVER_URL" environment variable to run tests that require ROS',
+    );
+}
+
+const ros = {
+    url: REALM_OBJECT_SERVER_URL,
+    createTestUser: () => {
+        return Realm.Sync.User.login(
+            REALM_OBJECT_SERVER_URL,
+            Realm.Sync.Credentials.nickname(`react-realm-context-tests-${uuid()}`),
```
nirinchev: This seems to have been copied from another library :)

kraenhansen:
> It's unrelated to this PR, but it would have been nice to use typescript for this new test framework :)

I agree ... I think I just cut that corner to get something working fast, but I think it would be worth the effort to get it typed. Especially now that we install the test suite from a packaged archive anyway (it started as symbolically linked packages without the need for a build/package-step).

kraenhansen:
Closing this as the credentials and authentication API will break in the near future.

PR #2275, opened 2019-03-07
Fix a flaky test

### LINKED ISSUE #2274: lib/user-methods.js's performFetch API is unstable across platforms
## Goals

We have to make a choice: Either the `performFetch` method of [lib/user-methods.js](https://github.com/realm/realm-js/blob/master/lib/user-methods.js#L107-L143) should adhere to the fetch API, which means that we should be able to [read header values from a response](https://github.com/realm/realm-js/blob/master/lib/user-methods.js#L376-L386) such as `response.headers.get('Content-Type')` or we should simplify the way we call the method making less assumptions on the shape of the response object (I don't recommend the latter).

In its current implementation `performFetch` performs very differently whether Realm JS is running on Node.js or in an environment with `fetch` available.

We currently have [tests failing](https://github.com/realm/realm-js/blob/master/tests/js/user-tests.js#L177-L187) on master due to this discrepancy in behaviour across platforms. 

- Realm JS SDK Version: 2.24.0
- Node or React Native: Both - but mostly React Native and Electron exhibits regressions.
- Client OS & Version: N/A

tgoyne:
The testLoginTowardsMisbehavingServer() would sometimes fail on CI due to failing to connect to the server at all rather than getting an invalid response from the server. Fix this by instead pointing at a local endpoint that won't give a valid response.

Fixes #2274.

### DIFF
--- a/tests/electron/app/main.js
+++ b/tests/electron/app/main.js
@@ -55,15 +55,11 @@ app.on("ready", () => {
     console.log("Running tests in the main process.");
     const jasmine = require("./jasmine.js").execute(filter);
     jasmine.onComplete((passed) => {
-      process.exit(passed ? 0 : -1);
+      console.log(`Testing completed with status ${passed ? 0 : -1}`)
     });
   } else if(runIn === "render") {
     console.log("Running tests in the render process.");
   } else {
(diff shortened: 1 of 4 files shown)


tgoyne:
The commit message explains why the changes in the commit were needed.

PR #2280, opened 2019-03-12
Removing Realm as dev dependency of the integration test suite

kraenhansen:
## What, How & Why?

This fixes an issue where integration tests suite would fail to install because Realm JS was not published. This happens when the integration test environments tries to pack up the integration test suite while the package.json of the Realm JS has a version that has not been published yet.

Instead of a dev dependency on Realm the test suite will `npm link` Realm at runtime when running the tests using `npm run dev`.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests


kraenhansen:
I'll have to scrap that as building the integration tests suite needs Realm installed to run the TypeScript transpilation to JavaScript.

PR #2286, opened 2019-03-15
An attempt at fixing the Electron integration tests

kraenhansen:
## What, How & Why?

This is an attempt at fixing the electron integration tests.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * File format: Generates Realms with format v9 (Reads and upgrades all previous formats)
 
 ### Internal
-* None.
+* Fixing the Electron integration tests. ([#2286](https://github.com/realm/realm-js/pull/2286))
 
 2.25.0 Release notes (2019-3-12)
 =============================================================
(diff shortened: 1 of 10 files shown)


kraenhansen:
@kneth not entirely - originally we had a build-time issue, which got fixed by upgrading to Electron 4.1.0. But then I experienced a runtime issue where requiring the Realm module would hang the process - I havn't identified the issue. It might be something in the way we build Realm for Linux in the test or alternatively an actual bug with Realm JS for Electron on Linux.

If we want to investigate this further, we could re-enable the test or run the docker container locally.

brightsider:
@kneth can you fix it?
node-pre-gyp WARN Using request for node-pre-gyp https download
node-pre-gyp WARN Tried to download(404): https://static.realm.io/node-pre-gyp/2.26.1/realm-v2.26.1-electron-v4.1-win32-x64.tar.gz
node-pre-gyp WARN Pre-built binaries not found for realm@2.26.1 and electron@4.1.4 (electron-v4.1 ABI, unknown) (falling back to source compile with node-gyp)

kneth:
@brightsider We currently don't build for Electron v4.1 but only v4.0. We are currently moving more large code chunks to this repository (#2343) and I would like to stabilize the code base first before making a public release. If you look closely, you can see that #2343 includes Electron v4.1 builds.

Issue #2252, opened 2019-02-12
Selected backlink closures for subscriptions


kneth:
See Epic: [realm/engineering#13](https://github.com/realm/engineering/issues/13)
Helps unblock a adapter performance issue.

cmelchior:
Fixed by #2296

PR #2296, opened 2019-03-21
Subscription Inclusions

ironage:
This is the beginning of https://github.com/realm/realm-js/issues/2252
It requires a updating to a version of sync which includes https://github.com/realm/realm-sync/pull/2892 and possibly object store https://github.com/realm/realm-object-store/pull/782

TODO: 
- [x] Upgrade to new core/sync/object store
- [x] Tests
- [x] Changelog update


kneth:
Great work - thanks!

dependencies.list
```diff
@@ -1,5 +1,5 @@
 PACKAGE_NAME=realm-js
 VERSION=2.26.0
-REALM_CORE_VERSION=5.14.0
-REALM_SYNC_VERSION=3.15.0
+REALM_CORE_VERSION=5.19.1
+REALM_SYNC_VERSION=4.4.1
```
kneth: Should we upgrade the Object Server too?

ironage:
I'm not sure why the node 10 tests are failing on CI. It appears that a node symbol is missing, but running locally with node 10.15.3 works for the same test on my machine.

tests/js/subscription-tests.js
```diff
@@ -250,7 +302,7 @@ module.exports = {
         if (!isNodeProccess) {
             return;
         }
-
+        console.log("beginning subscribeAndUpdateQuery() test")
```
tgoyne: leftover debug code

dependencies.list
```diff
@@ -1,5 +1,5 @@
 PACKAGE_NAME=realm-js
 VERSION=2.26.0
-REALM_CORE_VERSION=5.14.0
-REALM_SYNC_VERSION=3.15.0
-REALM_OBJECT_SERVER_VERSION=3.17.1
+REALM_CORE_VERSION=5.19.1
+REALM_SYNC_VERSION=4.4.2
```
tgoyne: Changelog needs to be updated with the changes from these core and sync versions.

docs/collection.js
```diff
@@ -158,9 +158,11 @@ class Collection {
      * });
      *
      * @param {string|Realm.Sync.SubscriptionOptions} options - an optional name or configuration object for the subscription.
-     *   The configuration object must contain `name` and the optional properties `update` and `timeToLive`.
+     *   The configuration object must contain `name` and the optional properties `update`, `timeToLive` and `inclusions`.
      *   If `update` is true the query on an existing subscription will be updated. If `timeToLive` is set, it indicates
-     *   to Realm for how long it must be kept alive, after which it will be automatically removed.
+     *   to Realm for how long it must be kept alive, after which it will be automatically removed. The `inclusions` property
+     *   is an array of property paths which specifies which linkingObjects relationships should be included in the subscription.
```
tgoyne: It doesn't appear to be just linkingObjects properties that can be included in this array.

src/js_results.hpp
```diff
@@ -196,7 +196,10 @@ typename T::Object ResultsClass<T>::create_filtered(ContextType ctx, const U &co
     NativeAccessor<T> accessor(ctx, realm, object_schema);
     query_builder::ArgumentConverter<ValueType, NativeAccessor<T>> converter(accessor, &args.value[1], args.count - 1);
     query_builder::apply_predicate(query, result.predicate, converter, mapping);
-    query_builder::apply_ordering(ordering, query.get_table(), result.ordering);
+    query_builder::apply_ordering(ordering, query.get_table(), result.ordering, mapping);
+    if (ordering.will_apply_include()) {
+        throw std::runtime_error("An INCLUDE clause is not supported on local queries, only on query based subscriptions.");
```
tgoyne: Does this really need to throw an error? It seems reasonable to just ignore it so that doing something like grabbing a query off a subscription object and passing it to filtered() always works.

docs/collection.js
tgoyne: `inclusions` is an incredibly vague name and seeing something like `.subscribe({inclusions: ['parents']})` wouldn't leave me with any clue about what it's doing.

tests/js/subscription-tests.js
```diff
@@ -328,4 +380,116 @@ module.exports = {
...
+            return;
+        }
+        return verifySubscriptionWithParents("@links.Parent.child");
+    },
+
+    testSubscribeToChildrenWithMalformedInclusion1() {
+        if (!isNodeProccess) {
+            return;
+        }
+        return new Promise((resolve, reject) => {
```
tgoyne: `return verifySubscriptionWithParents("something.wrong").then(() => throw new Error('subscription should have failed'), (err) => TestCase.assertEqual(err.message, "No property 'something' on object of type 'ObjectA'"));`

docs/collection.js
ironage: I'm struggling to come up with an intuitive name. How about `included_parents`, `included_reverse_relationships`, or `reverse_relationship_inclusions`? (cc: @kneth )

src/js_results.hpp
ironage: The restriction is entirely arbitrary. During the sync implementation (https://github.com/realm/realm-sync/pull/2892#issuecomment-476427632) it was recommended that the js binding throws if this syntax is attempted on a local query because it is probably an error if someone tries to do this. A counter argument might be that users want to copy out the subscriptions as listed in their query based sync Realms on the server and run them verbatim locally? I don't know if that's a strong enough argument against saying that any other use of the "INCLUDE" syntax is invalid locally. It is intended to be a server side operation only. I don't have a strong opinion on it, just slightly leaning towards it being an error in local js queries because there is no way the js binding would generate something like this internally and so it is most likely a user error and having a descriptive error message is probably better than silently doing nothing.

tgoyne: I mostly don't see any reason why someone would have an INCLUDE clause in their query other than grabbing it off a subscription. There's no reason for them to write one manually and we probably shouldn't even document it, so I don't think anyone would ever expect it to actually do something and then be surprised when it doesn't? As long as `.filtered('... INCLUDES (...)').subscribe()` actually does pass the includes clause to the subscription I don't think it really counts as silently doing nothing when the user has tried to ask for something. If it doesn't just happen to work then it's not worth making it work and should just throw, though.

docs/collection.js
kneth: Since we call the `linkingObjects` in the schema, maybe `includes_linkingObjects` is less confusing?

cmelchior: In JS naming conventions are camelCase, but trying to weave in the word linkingObjects seems beneficial IMO. I don't have a strong opinion on using `include/includes/included`, but e.g. `includeLinkingObjects` seems somewhat descriptive.

src/js_results.hpp
cmelchior: Note, that with new subscription management API's getting to the underlying query description is easier than it used to be. So I suspect that people might be copy/pasting it to e.g. Studio when debugging. I'm not sure if that is a strong enough argument for allowing it to run locally though.

cmelchior: In any case, it is easier to be more restrictive from the start, and enough people actually run into this exception we can consider removing it?

docs/collection.js
cmelchior: As I started on the Java API, what about `downloadLinkingObjects`?

ironage: Personally, I think "download" is confusing terminology to add because it is not a comprehensive description of what is happening, we want to include these relationships in the subscription and keep them up to date as opposed to downloading them in the traditional "once off" sense of the word. Maybe I'm overthinking it. But I like `includeLinkingObjects`, because I think it reads like those relationships are included in the subscription (ie: `.subscribe({name: "myName", includeLinkingObjects: ['parents']})`)

ironage: Side question, the recent updates of ttl and other parameters on the subscription parameters added a restriction that if the parameter to `subscribe` is an object, it must include a user defined name. This means that if users want to add a `includeLinkingObjects` parameter they  must also name the subscription. Is that an unnecessary restriction?

tgoyne: `includeLinkingObjects` seems decent to me.

nirinchev: I agree - all properties on the subscribe argument should be optional, including the name.

cmelchior: The reason I made `name` required is that `update` doesn't make any sense if you don't have a name, as you don't know what to update. Idon't know If that weights stronger than allowing fetching linking objects on an unnamed subscription.

In any case, we can make a runtime check for it. It was just nicer when it was in the type system.

nirinchev: Update without a name could probably work like Realm.create(update: true) for object without a primary key, or we could throw an error in that particular case 🤷‍♂️ I felt it's a bit annoying though as that meant you can't register an anonymous subscription that expires after some time.

cmelchior: Sure, I don't have a problem lifting that restriction 👍

ironage: Thanks for the input. I renamed to `includeLinkingObjects` and lifted the restriction requiring the name in the subscription options.

src/js_results.hpp
```diff
@@ -324,28 +329,65 @@ void ResultsClass<T>::subscribe(ContextType ctx, ObjectType this_object, Argumen
...
+                update = Value::validated_to_boolean(ctx, update_value, available_options[UPDATE]);
 
-            ValueType ttl_value = Object::get_property(ctx, options_object, "timeToLive");
+            ValueType ttl_value = Object::get_property(ctx, options_object, available_options[TTL]);
             if (!Value::is_undefined(ctx, ttl_value))
-                ttl = util::Optional<int64_t>(Value::validated_to_number(ctx, ttl_value, "timeToLive"));
+                ttl = util::Optional<int64_t>(Value::validated_to_number(ctx, ttl_value, available_options[TTL]));
+
+            ValueType user_includes = Object::get_property(ctx, options_object, available_options[INCLUSIONS]);
+            if (!Value::is_undefined(ctx, user_includes)) {
```
cmelchior: This section could use a few comments. There is a lot of things happening that is hard to follow IMO

src/js_results.hpp
ironage: I lifted the restriction, so now queries with "INCLUDE" will do the right thing and added tests for it. However, I agree we shouldn't document it as officially part of the query syntax as it might be confusing.

nirinchev:
The typescript definitions (index.d.ts) need to be updated to reflect the new property and also the lifted restriction for the `name`.

kneth:
Testing on node 10 is failing due to a missing symbol (`EVP_CIPHER_CTX_cleanup ` from OpenSSL). The usage was  introduced by https://github.com/realm/realm-core/pull/3241. We might have to do something to https://github.com/realm/realm-js-private/blob/master/Dockerfile#L32.

tgoyne:
https://github.com/realm/realm-js/pull/2343 fixes some openssl-related linking issues including with that specific symbol IIRC.

kneth:
Sounds like we should merge #2343 first :smile:

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+NOTE: This release is only compatible with Realm Object Server 3.21.0 or later.
+
+### Enhancements
+* Add an optional parameter to the `SubscriptionOptions`: `inclusions` which is an array of linkingObjects properties. This tells subscriptions to include objects linked through these relationships as well (links and lists are already included by default). ([#2296](https://github.com/realm/realm-js/pull/2296)
+
+### Fixed
+* Making a query that compares two integer properties could cause a segmentation fault in the server or x86 node apps. ([core PR #3253](https://github.com/realm/realm-core/issues/3253))
```
```suggestion
* Making a query that compares two integer properties could cause a segmentation fault in the server or x86 node apps. ([realm-core#3253](https://github.com/realm/realm-core/issues/3253))
```

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+* Fix an error in the calculation of the downloadable_bytes value supplied to the progress callback. (See sync version 4.0.0)
+* HTTP requests made by the Sync client now always include a Host: header, as required by HTTP/1.1, although its value will be empty if no value is specified by the application. (sync v4.2.0)
```
```suggestion
* HTTP requests made by the Sync client now always include a `Host: header`, as required by HTTP/1.1, although its value will be empty if no value is specified by the application. (sync v4.2.0)
```

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+* The server no longer rejects subscriptions based on queries with distinct and/or limit clauses. (sync 4.2.0)
+* A bug was fixed where if a user had canCreate but not canUpdate privileges on a class, the user would be able to create the object, but not actually set any meaningful values on that object, despite the rule that objects created within the same transaction can always be modified. (sync 4.2.0)
```
```suggestion
* A bug was fixed where if a user had `canCreate` but not `canUpdate` privileges on a class, the user would be able to create the object, but not actually set any meaningful values on that object, despite the rule that objects created within the same transaction can always be modified. (sync 4.2.0)
```

Issue #2297, opened 2019-03-21
Retry of authentication requests are broken


kraenhansen:
## Goals

Calling `Realm.Sync.User.login` against a server under heavy load.

## Expected Results

The authenticate request should be retried and eventually succeed.

## Actual Results

After the first retry the following runtime error is thrown:

```
TypeError: Cannot read property 'get' of undefined
    at request.then (.../node_modules/realm/lib/user-methods.js:345:50)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

I believe the root cause is that we're setting the `request` object to something that resolves to a user and not a response, when performing the retry at line 340 of lib/user-methods.js: https://github.com/realm/realm-js/blob/master/lib/user-methods.js#L337-L345

## Steps to Reproduce & Code Sample

Run this a couple of times against an otherwise busy server (with nickname auth provider enabled):

```javascript
const Realm = require("realm");

const SERVER_URL = "http://localhost:9080";

async function run() {
  const nicknames = new Array(100).fill();
  const runNumber = Math.floor(Math.random() * 10000);
  await Promise.all(nicknames.map(async (nickname, i) => {
    const credential = Realm.Sync.Credentials.nickname(`nickname-${runNumber}-${i}`);
    const user = await Realm.Sync.User.login(SERVER_URL, credential);
    console.log(`Authenticated as ${user.identity}`);
  }));
}

run().then(() => {
  process.exit();
},err => {
  console.error(err.stack);
  process.exit(1);
});
```

## Version of Realm and Tooling

- Realm JS SDK Version: 2.25.0
- Node or React Native: Node
- Client OS & Version: MacOS
- Which debugger for React Native: N/A

PR #2298, opened 2019-03-21
Update user-methods.js to fix authentication retrying

kraenhansen:
## What, How & Why?

This closes https://github.com/realm/realm-js/issues/2297.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests (only by patching and re-running a script that previously failed)


tgoyne:
What's the actual change here that fixes the bug? I don't see any functional differences other than changing `function(){}` to an arrow function and moving the catch to the end. The latter change I think is wrong; we only want to retry on network errors and not things like incorrect passwords.

tgoyne:
Oh, I see, two-arg then() doesn't work how I thought it did and this does the right thing.

PR #2301, opened 2019-03-25
Add support for field aliases.

cmelchior:
This is a prerequisite for https://github.com/realm/realm-js/issues/2266 so we can rename internal fields to match naming conventions in JS as well as adding custom getters/setters when exposing the managed subscriptions.

This PR adds the option of mapping a Javascript property to a differently named internal column name. Similar functionality is already present in Java and Cocoa.

Implementation details: It looks like the `KeyPathMapper` still allows queries on the original column. So, for now, we should probably just allow that.

API:
```
// Defining the schema
const schema = {
  publicName: { type: 'string', mapTo: 'internalName' },
}

// The schema object returns the mapped property
realm.schema[0].properties['publicName'].mappedTo == 'internalName'
````

TODO
- [x] Merge OS support: https://github.com/realm/realm-object-store/pull/783 and redirect to master

### CODE REVIEW

File: src/js_realm_object.hpp [hunk 1]
```diff
@@ -235,14 +236,16 @@ std::vector<String<T>> RealmObjectClass<T>::get_property_names(ContextType ctx,
     auto realm_object = get_internal<T, RealmObjectClass<T>>(object);
     auto &object_schema = realm_object->get_object_schema();
 
+    // When parsing schema objects from JS, we always ensure an alias is set. If the user didn't define
+    // one, the name property will be used as the alias.
```
ironage: Doesn't this disagree with the logic in object store validation which checks that there are no conflicts between aliases and names?

cmelchior: Yes, this comment isn't correct anymore. I'll update it.

tests/js/alias-tests.js
```diff
@@ -0,0 +1,120 @@
...
+
+        // Creating objects most use the alias
+        realm.create('ObjectA', {
+            otherName: 'Foo'
+        });
+
+        // Creating uses arrays still work
+        realm.create('ObjectA', ['Bar'])
+
+        // Using the internal name instead of the alias throws an exception.
```
ironage: I don't find this behaviour intuitive. To me an "alias" means another name (ie in addition to but optional). I think the original property should still be valid and useable. At least, I cannot think of why we would want to restrict this.

cmelchior: The problem is that if your print the content of an object you will suddenly get 2x the number of properties if we keep both. That feels really award to me. This is how it works in Java (and I believe Cocoa).

Maybe it is the name `alias` that should be changed?

ironage: What if we still only print the aliased name of a property if one exists, but still allow accessing the property by both the original name and the alias?

cmelchior: Yes, I guess we could do that. I'm not 100% sure what the use case would be though?

On the plus side, it grants you the flexibility of a "hidden" API still allowing you to use underlying names. On the negative side, you get an API that seems error prone as mixing "internal" and "public" names probably indicate a user error somewhere?

I'm guessing I just fail to see why you would want to define an alias and still use the internal name? The use case for linking objects where a bit more obvious due to the abstract nature fo defining them.

ironage: >  I'm not 100% sure what the use case would be though?

The way the tests showcase this feature, the user defines both the name and the alias of a property. This means that the user knows the name in the first place, and may as well refer to the property by its actual name. I might be misunderstanding how you intend to use this feature but here's a made up use case of general aliasing where users may want to use both the alias and the real property name: several classes have properties which have different names but serve a similar purpose, eg: `US_Resident.social_security_number` and `Canadian_Resident.social_insurance_number`. Throughout the code, these properties are referenced via the normal property name, except in a generic function `can_legally_work(resident)` which processes both class types and needs to refer to the property of both in a generic way to check if the id is null. In order to achieve this, both properties have an alias "legal_work_id" applied. In this way the generic code is able to check if the id is null on both class types without having specialised code per class type.

ironage: The scenario describes the need (or convenience) for general aliasing which I guess wasn't what you had in mind? What is an example of a use case where you would explicitly set up the name and alias of a property but consider it an error if the name was used?

cmelchior: I guess I was coming from this from a more statistically typed mindset. The use case I had in mind was e.g renaming fields to conform to naming practises in the language, e.g. the current use case is exposing fields in a subscription in a more idiomatic way. In my case, `createdAt` instead of `created_at` and `timeToLive` instead of `time_to_live`. 

 But I can see the argument for allowing both in a more dynamic language like JS. I'll add support for it.

kneth: To me, it is most intuitive to be able to use both.

cmelchior: After looking at my use case, I don't think that exposing the internal properties are a good idea.

My current use case:

1) Expose `__ResultSets` in a sensible and "safe" manner, including updating fields

Especially updating properties is a problem in Javascript as you cannot override setters. Basically, the pattern is to define a pseudo setter that manipulates a differently named property.

E.g.

```
        const ResultSets = function() {};
        ResultSets.schema = Object.freeze({
            name: '__ResultSets',
            properties: {
                name: { type: 'string', indexed: true, alias: '_name' },
                query: {type: 'string', alias: '_query'},
                matches_property: 'string',
                status: {type: 'int', alias: 'state' },
                error_message: { type: 'string', alias: 'error' },
                query_parse_counter: 'int',
                created_at: { type: 'date', alias: '_createdAt'},
                updated_at: { type: 'date', alias: '_updatedAt'},
                expires_at: { type: 'date', alias: '_expiresAt', optional: true },
                time_to_live: { type: 'int', alias: '_timeToLive', optional: true },
            }
        });
        Object.defineProperty(ResultSets.prototype, "query", {
            enumerable: true,
            get: function() {
                return this._query;
            }
            set: function(val) {
                this._query = validateAndSerializeQuery(val);
                this._updatedAt = new Date();
            }
        });
```

For this to work, I need the following

1) The ability to add a pseudo setter with a name not already used (by using aliases).
2) The ability to query on the pseudo-property. Possible since the internal name and the pseudo-property are the same.

Specifically, this approach means that it is _impossible_ to enumerate the internal names as well since it conflicts with the custom setters/getters. So by exposing both, it is impossible to create a nice wrapper on a managed object. 

Thoughts?

ironage: That's an interesting approach, but it does seem like you'd need to do some surprising things to get it to work. Wouldn't a more straightforward design be to provide the relevant methods on a subscription object? I don't think js users should be manipulating the `__ResultSets` directly? Ideally they should not know about that implementation and they should be able to get any relevant information from the subscription object?

cmelchior: The problem with a standalone subscription object is that it quickly falls apart when it represents a mutating `ResultSets` object. We saw that with the old permission system, where the API just got really weird, quickly.

As long as the subscriptions were static it was a bit more acceptable, but now with the query/timestamps that can change, we would need to do a lot of awkward things to manage subscription object that is immutable.

Just exposing the `__ResultSets`  means we can leverage all the nice things from Realm directly (like changelisteners). The only problem is that we need to mask the underlying implementation as well as add custom logic to setters. Unfortunately, that is rather hard in JS, which is what lead me to the above approach. 

If we supported JS proxies we could have wrapped the `__ResultSets` more nicely, but we don't fully support ES6 across all platforms yet. So, the above approach is as far as I know the best way to hide the underlying implementation.

An alternative could have been to have methods like `updateQuery()` and `updateTtl()`, but that wouldn't stop people from directly manipulating the properties. By renaming the properties to `_XXX` we don't stop them either, but by convention, these are now considered implementation details, and we can then add custom logic to the setters of the JS defined properties.

ironage: I'm not familiar with the history of the old permission system. I would still favour having `updateQuery()` and `updateTtl()` methods on a subscription object, if it is immutable, it could return an instance to an updated (but immutable) subscription. Maybe I don't fully understand the problems this would create, but it still seems the standard way to operate on subscriptions at a high level. (Also note that in the code sample you posted above, the object is frozen by `Object.freeze` and updates won't be triggered and I think setting the query the way you do will throw a type error?)

cmelchior: Note that `Object.freeze()` is only on the schema. Not the rest of the object.

The problem with wrapper objects vs. exposing live objects is three-fold:

1) It goes against everything else we designed Realm to do. Our API for normal model classes exposes "live" objects that are automatically kept up to date. If we believe this is the better approach why don't we do this for our own API as well? 

2) We create a disconnect in our API if some parts of it return static objects while others return "live" objects. It just makes it hard to use and reason about. 

3) Wrapping a live object makes it hard to manage when the underlying properties change. We saw this with the old permission API. Basically, you are forced to add change listeners inside the wrapper and copy out all objects every time something changes to avoid problems if the Realm is closed. You never know if you are looking at the "latest" state and filtering/changelisteners just become hard/impossible. Exposing live objects directly gives us all of this out of the box.

Basically, I don't see the problem in exposing `__ResultSets` directly. We cannot really claim "implementation details" anymore as people have real needs for interacting with it for updating the query or TTL.

Can you elaborate on why you would prefer `updateXXX()` methods on a static Subscription object instead of just exposing the `__ResultSets` ?

kneth: Will exposing `__ResultSets` make us sensitive to breaking changes in sync?

cmelchior: Depending on what type of breaking change you think about. Adding/renaming/removing fields would all be supported using aliases. 

More fundamental changes like removing it completely would obviously not, but we would need to change the wrapper API anyway if that happened.

ironage: Thanks for explaining, I think I am seeing things more from your perspective now. I do agree that exposing the raw __ResultSets will allow for the most "Realm-like" reactive behaviour. I think the risk is that people can circumvent our intended design, and change things at will like update the query without any real way to guarantee that the new timestamps are updated as well. If you can find a way to make this work with aliases then that sounds like a good way forward.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
+* Added support for defining property aliases in the schema using `name: { type: 'int', alias: 'aliasName' }`. In that case `name` will be  used in the Realm file, while the alias is used for reading/writing the property as well as querying it.
```
```suggestion
* Add support for defining property aliases in the schema using `name: { type: 'int', alias: 'aliasName' }`. In that case `name` will be  used in the Realm file, while the alias is used for reading/writing the property as well as querying it.
```

docs/realm.js
```diff
@@ -455,6 +455,9 @@ class Realm {
  *   This is not supported for `"list"` properties of object types and `"linkingObjects"` properties.
  * @property {boolean} [indexed] - Signals if this property should be indexed. Only supported for
  *   `"string"`, `"int"`, and `"bool"` properties.
+ * @property {string} [alias] - an alias in JavaScript for the underlying name in the Realm file. If set, this must be
+ *   used when reading/writing the property or making queries against it. Trying to use the internal name will throw
```
kneth: "internal name" might be a little unclear here. Maybe "underlying name" is better.

cmelchior:
Ready for final review. The CI failures do not seem relevant to this PR.

src/js_schema.hpp
```diff
@@ -386,6 +396,9 @@ typename T::Object Schema<T>::object_for_property(ContextType ctx, const Propert
     static const String optional_string = "optional";
     Object::set_property(ctx, object, optional_string, Value::from_boolean(ctx, is_nullable(property.type)));
 
+    static const String mapped_to_string =  "mappedTo";
```
cmelchior: Right now I use `mapTo` when defining the schema, but `mappedTo` is returned when calling `realm.schema`. Should these be the same?

kneth: Yes, I think so.

PR #2302, opened 2019-03-26
Adding a global notifier listener that triggers at startup

nhachicha:
## What, How & Why?
This closes https://github.com/realm/realm-js-private/issues/521 
see https://github.com/realm/realm-js-private/pull/522

This will allow the client to listen for a new event `startup`, which triggers before we open the Realm for sync.
```Javascript
var startupCallback = function (startupObject) {    
    console.log(`Realm at virtualPath := ${startupObject.path}`);
    // use the Realm
   let realm = startupObject.realm;
}

Realm.Sync.addListener(`realm://${SERVER_URL}`, adminUser, NOTIFIER_PATH, 'startup', startupCallback);
````

#### Using the `Realm.Worker` API
```Javascript
// worker.js
function onstartup(startupObject) {
     console.log(`Realm at virtualPath := ${startupObject.path}`);
    // use the Realm
   let realm = changes.realm;
}

function onchange(change) {
}

function onavailable(path) {
}

function ondelete(path) {
}

module.exports = {onchange, onavailable, ondelete, onstartup};

// server.js
await Realm.Sync.addListener(`realm://${SERVER_URL}`, adminUser, NOTIFIER_PATH, new Realm.Worker('worker.js'));
```

## ☑️ ToDos

* [x] 📝 Changelog entry

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* [Sync] Emit a `startup` event to the Sync event listener when a Realm matching the regex is detected upon the listener startup.
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -1037,8 +1037,7 @@ class Subscription {
  *    console.log(`Realm at ${path} deleted`);
  * }
  * 
-
- * module.exports = {onchange, oncavailable, ondelete};
+ * module.exports = {onstartup, onchange, oncavailable, ondelete};
  *
  * // server script
  * Realm.Sync.addListener(realmServerURL, adminUser, '.*', new Realm.Worker('my-worker'));
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -122,6 +122,10 @@ class Sync {
      *
      * Currently supported events:
      *
+     *  * `'startup'`: Emitted when the listener start. It returns the Realm(s) matching the filter
+     *    regex. A [ChangeEvent]{@link Realm.Sync.ChangeEvent} argument
+     *    containing the `path` and `realm` properties (other [ChangeEvent]{@link Realm.Sync.ChangeEvent} 
+     *    properties are N/A). 
      *  * `'available'`: Emitted whenever there is a new Realm which has a virtual
      *    path matching the filter regex, either due to the Realm being newly created
      *    or the listener being added. The virtual path (i.e. the portion of the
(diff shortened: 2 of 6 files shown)


lib/notifier.js
```diff
@@ -280,6 +280,42 @@ function addListener(server, user, regex, event, callback) {
...
+    const realmAvailableCallback = (realmPath, realmId) => {
+        const absolutePath = require('path').join(require('process').cwd(), 'realm-object-server/listener/realms', realmPath, `${realmId}.realm`);
+        if (require('fs').existsSync(absolutePath)) {
+            new Promise(async (resolve) => {
+                var realm = new Realm({
+                    path: absolutePath,
+                    sync: true
+                });        
+                await callback(realmPath, realm); // callback could be an async or sync function.  
+                realm.close(); // close the Realm to avoid realm::MismatchedConfigException realm already opened with different sync configurations.
```
tgoyne: If the user's callback is async then this is far too late to close the realm, as the GN will have already tried to open the realm with a different configuration before this is reached.

docs/sync.js
```diff
@@ -990,6 +998,10 @@ class Subscription {
  *
  * Currently supported events:
  *
+ *  * `'startup'`: Emitted when the listener start. It returns the Realm(s) matching the filter
```
tgoyne: This is not an accurate description of the startup event.

docs/sync.js
```diff
@@ -1009,19 +1021,23 @@ class Subscription {
  *
  * @example
  * // my-worker.js
- * function onchange(path) {
- *    console.log(`Realm created at ${path}`);
+ * function onstartup(path) {
+ *    console.log(`Realm at ${path} deleted`);
```
tgoyne: This sample is wrong in a bunch of places.

lib/notification-worker.js
```diff
@@ -60,6 +60,21 @@ process.on('message', (m) => {
             }
             process.send({change: m.change});
             break;
+        case 'startup':
+            if (impl['on' + m.message]) {
```
tgoyne: The `'on' + m.message` construction is used above because there are multiple messages being handled. That is not the case here.

lib/notification-worker.js
```diff
@@ -60,6 +60,21 @@ process.on('message', (m) => {
...
+                const startupRealm = Realm.Sync._deserializeStartup(m.change);
```
tgoyne: This variable is not a realm.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* [Sync] Emit a `startup` event to the Sync event listener when a Realm matching the regex is detected upon the listener startup.
```
```suggestion
* Emit a `startup` event to the Sync event listener when a Realm matching the regex is detected upon the listener startup.
```

docs/sync.js
```diff
@@ -122,6 +122,10 @@ class Sync {
      *
      * Currently supported events:
      *
+     *  * `'startup'`: Emitted when the listener start. It returns the Realm(s) matching the filter
```
```suggestion
     *  * `'startup'`: Emitted when the listener starts. It returns the Realm(s) matching the filter
```

lib/notifier.js
```diff
@@ -263,7 +325,7 @@ function addListener(server, user, regex, event, callback) {
 
     // If the old API is used, convert all arguments to the new API
     // If `event` is a Worker, only 4 arguments are provided
-    if (arguments.length === 4 || arguments.length === 5) {
+    if (arguments.length > 3) {
```
kneth: Why not use the original - and more precise - condition?

nhachicha: this is used mainly to detect old/new function signature, if the caller passes more arguments then they will be ignored.

PR #2303, opened 2019-03-26
Fix Xcode 10.2 'Control may reach end of non-void function' error

mohammadalijf:
## What, How & Why?

This Pull Request will resolve following build error on XCode 10.2

> Control may reach end of non-void function


mohammadalijf:
this resolves #2305

hyb175:
I may be completely wrong but seems to me that it's always better to have a `default` case to switch statements. In this case, even if we are not comfortable returning a value such as `null` or `unknown`, I would advocate for throwing an error of `invalid argument` or something along those lines. This way, at least any future update to the language shouldn't cause this error to happen again, and we can add a test case to make sure, if there is a new entry to the enum, that we should update this function.

mohammadalijf:
@hyb175 
when we are dealing with enums it's always better to cover all enum cases in switch than providing `default` case statements. `default` case should be used for dynamic values like strings or numbers. also if in future language get changes (or someone add a case to the enum) its better to get compile errors than run time errors (unless you have a 100% test cover).

timwangdev:
Unfortunately, the PR will break Xcode build below version 10.2

tbenr:
an `#ifdef` is needed here :-)

mohammadalijf:
@tbenr 
i'm a little bit confused. should i check __IPHONE_12_2 which is defined in `Availability.h` or i am looking at the wrong place? also should i check for mac os version?

something like this?

```c
#ifdef __IPHONE_12_2 || __MAC_10_14_4
    case kJSTypeSymbol: return "symbol";
#endif
```

tbenr:
@mohammadalijf good question, i didn't spent time on this, i was just suggesting a compile time define :-)
btw i don't see it related to mac version but more to target compile version \ sdk version.

mohammadalijf:
@tbenr 
according to [this](https://stackoverflow.com/questions/18406622/ifdef-else-endif-choose-ios-sdk-version-and-function) i guess i should use `Availability.h`.

also in `JSType` they have defined `kJSTypeSymbol` as follow
```c
kJSTypeSymbol API_AVAILABLE(macosx(10.14.4), ios(12.3))
```
so i also checked if `__MAC_10_14_4` is defined

mohammadalijf:
@timwangdev i have changed the code check for iOS and Mac target version but i don't have access to Xcode build below version 10.2. can you please verify the code works?

tgoyne:
There's another place where we switch on JSValueGetType(), in rpc.cpp, that'll probably need similar changes.

Bautista-Baiocchi-lora:
Fixed worked for me on RN 57.8 and Realm 2.25

kneth:
@mohammadalijf

I believe you need to handle the case `kJSTypeSymbol` in https://github.com/realm/realm-js/blob/master/src/rpc.cpp#L541 too.

StasDoskalenko:
@mohammadalijf @kneth hey guys! If it's possible, could you please implement this? It's a huge blocker. Thanks!

mohammadalijf:
@kneth 
Glad i could help. I have signed the CLA.
Also i have added a case for `kJSTypeSymbol` in `RPCServer::serialize_json_value(JSValueRef js_value)` but used a simple `break` to skip serializing a `Symbol` since i don't know if Realm is going to support symbols or how the framework will use JavaScript's `Symbol`.

mohammadalijf:
Hey @StasDoskalenko
Already on it ;)

bmunkholm:
An update to the changelog.md file is appreciated.

mohammadalijf:
@bmunkholm 
Sorry i have updated the changelog.md

tandat2209:
Sorry. But how can I use code from this PR?

rajivshah3:
@tandat2209 if you need a patch and can’t wait for the next release, you can use https://github.com/ds300/patch-package to apply this [patch](https://github.com/iotaledger/trinity-wallet/blob/40c9f3a6e306feb7cc791543f5d1ceba815da051/src/mobile/patches/realm%2B2.21.1.patch)

joshuapinter:
@tandat2209 Or simply use the `master` branch with `"realm": "realm/realm-js"`.

Samox:
Hello guys ! When using `"realm": "realm/realm-js"` the android folders disapears from the node_modules (hence breaking the android build). Is there a postinstall script to run so I can have the android folder back ?

kristoferpelchat:
@Samox I came across this and found that even the `realm/react-native` folder did not contain `android`. I pulled it down via `git clone` off [realm-js](https://github.com/realm/realm-js). I then proceeded to build and was getting some errors. Come to find out I don't need realm...so I stopped there and didn't proceed. I am not quite sure why `android` folder isn't coming down from the `npm install` when `"realm": "realm/realm-js#master"` is set in `package.json`. 🤷‍♂️

kneth:
The `android` folder is distributed under `realm/android` and not `realm/react-native/android`.

StasDoskalenko:
@kneth I believe it will be easier for you guys to release a patch version to the npm instead of supporting this thread 😅

kneth:
I have just released v2.26.1 with the fix.

rasom:
thanks, @kneth

PR #2307, opened 2019-03-26
Changing `Download Core` Xcode Build Phase Shell From `/bin/sh` to `/bin/bash` in RealmJS.

ttahmouch:
## What, How & Why?

We've been trying to migrate to a new CI environment that's pretty locked down. We were previously using `2.4.0`, but had an issue because we needed to proxy to access the public Internet in order to download from `static.realm.io`. I noticed that support for an HTTP proxy was added in a later minor version update. However, the `Download Core` phase in the Xcode project `RealmJS` relies on `/bin/sh` in the environment, and we need to be able to set `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` environment variables in our `bash` shell. We'd need to do it in a non-interactive, login shell.

I wasn't sure of the best approach to solve this problem. If the shell used in the Xcode project build phase for `Download Core` remains `/bin/sh`, we'll have no way to successfully `Download Core` in our CI environment. I don't know that it's fair to change it to `/bin/bash`, but I'm honestly not sure of the best approach.

What are your thoughts?


kraenhansen:
I wonder if it would be [more portable](https://stackoverflow.com/a/10383546/503899) to use `/usr/bin/env bash`?

ttahmouch:
If I understood it correctly, the Build Phase script just needed a way to ensure that `node` was installed in the shell because the Download Realm script is a `node` script. However, the Build Phase script previously just assumed `nvm` was available in the shell as it looked for the installation in the user's home directory.

I modified the script to install `nvm` locally in the project. It may be more in line with the original intent. 

Take a look in Xcode, and let me know what you think. The diff looks poor because Xcode stores the whole script as one line.

ttahmouch:
This may help as well.

image

ttahmouch:
@kraenhansen 

I totally get it if you didn't have time to take a look, but would you or someone else be able to help out? I'd really like to take care of the issue upstream if possible so that we can avoid maintaining a fork.

Let me know even if the team won't have time to take a look soon. I'd really appreciate it.

I was also trying to explain the changes to my teammate, and was thinking that it may be useful in this context too.

> The `Download Core` build phase runs its own shell, i.e., `/usr/bin/env bash -l`. In that shell, it tries to install `nvm` using the installation command which downloads the install script using `curl` and runs it using `bash`. In the main bash shell, it exports `PROFILE` as `/dev/null` so the `nvm` install script doesn't touch the `.bashrc`, `.bash_profile`, or `.profile` in the user's home directory. It exports `NVM_DIR` to install `nvm` in a folder called `nvm/` in the `node_modules/realm-js/src/`  directory. It then sources the local copy of `nvm` so the command is useable. Then it tells the local copy of `nvm` to install a local copy of `node@11.13.0` that resides in `node_modules/realm-js/src/nvm/`. Then it uses that local copy of node to `Download Core`.

ttahmouch:
Added an uninstallation of `nvm` and `node` from the local directory when they are no longer needed after running the `Download Core` script.

image

kraenhansen:
Are there good reasons for using nvm at all? We might as well download node directly from https://nodejs.org/dist/v10.15.3/node-v10.15.3-darwin-x64.tar.gz as we know this will be running on a mac (via xcode).

kneth:
@kraenhansen @ttahmouch 

Since this is only supposed to be used on macOS, we can use macOS feature: the purpose of `launchctl setenv` is to set environment variables for all future processes. So I can't see the need for this PR.

ttahmouch:
@kraenhansen 

It would be mostly for environment architecture and version management. Version management shouldn't matter in this case since we only need one version of `node` to run the `Download Core` script, and to my knowledge the version of `node` used doesn't matter. You could just download it directly.

However, it would probably be just as safe to assume that the environment already has a `node` installed in it as it would be to assume the environment already has an `nvm` installed in it if the version of `node` doesn't matter. Downloading it manually would mostly just be to ensure the build step doesn't fail just because of the lack of dependency, but that could also just be managed through documentation.

I'm personally not a fan of downloading anything in an Xcode build phase, but I suppose `Download Core` needs to happen for environment architecture reasons as well.

@kneth 

The original intent for this PR was because we needed to ensure `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` could be exported for the shell that runs this Xcode build phase. Our CI environment is not exposed to the public Internet, and we have to route the request to `static.realm.io` through our proxy in order to `Download Core`. However, the only way I could think to export those variables into the Xcode build phase shell would be to change the shell to `shellPath = "/usr/bin/env bash -l";`. We need a way to export environment variables from our `.bash_profile`.

The rest of the changes came in when we started discussing what implications of changing the shell to `bash` would be on `nvm`.

I can try using `launchctl setenv` in our CI environment, and see if the `/bin/sh` shell receives the environment variables. I'll let you know what happens.

Thank you both for your help.

ttahmouch:
It half works, @kneth . When reading online about which applications would have access to environment variables exported with `launchctl setenv`, it sounded like GUI and shell applications would be able to see them, but I `echo`'d the values to be sure if I ran a build with `react-native run-ios` versus using `Xcode` directly that I'd see the same values, and they only show if I build using `react-native run-ios`.

### Building using Xcode. Notice the whitespace where the variable values should be logged.
image

### Building using CLI.
```
info nvm is not compatible with the "npm_config_prefix" environment variable: currently set to "/Users/hzbnyy/.nvm/versions/node/v10.13.0"
Run `unset npm_config_prefix` to unset it.

info v8.15.0 is already installed.

info nvm is not compatible with the "npm_config_prefix" environment variable: currently set to "/Users/hzbnyy/.nvm/versions/node/v10.13.0"

info Run `unset npm_config_prefix` to unset it.

**info http://user:pass@naproxy.hostname.com:8080**

**info http://user:pass@naproxy.hostname.com:8080**

**info *.hostname.com**

info Resolved requirements: { SYNC_SERVER_FOLDER: 'sync',
  SYNC_ARCHIVE: 'realm-sync-cocoa-3.15.0.tar.gz',
  SYNC_ARCHIVE_ROOT: 'core' }

info Skipping lockfile check as --force is enabled

info Download url: https://static.realm.io/downloads/sync/realm-sync-cocoa-3.15.0.tar.gz

info Downloading realm-sync-cocoa-3.15.0.tar.gz
```

### Logging the environment variables.
image

kraenhansen:
> the Build Phase script just needed a way to ensure that `node` was installed in the shell because the Download Realm script is a `node` script.

That's correct and it also attempts to ensure that `node` is not a very old version (as the download script uses modern features of Node.js). The sourcing of `~/.nvm/nvm.sh` is an attempt to make it easier for people to control the version of node that the script is run with. We experienced that the system-installed version of node is often outdated and many developers use `nvm` or `n` to keep them updated, hence sourcing the ~/.nvm/nvm.sh if it existed seemed appropriate. Users could specify an updated version as their `default` alias, which would be used when running this script.

I don't believe sourcing scripts or installing NVM are needed if we use a `/usr/bin/env bash -l` shell: In that case we can rely on the `.bash_profile` to setup the right version of node.

> I'm personally not a fan of downloading anything in an Xcode build phase, but I suppose `Download Core` needs to happen for environment architecture reasons as well.

I agree - we should rely on the version of Node.js supplied by the user and instead check that they actually provided a version that is able to run the script. I just preferred downloading directly over installing NVM in the shell-script, but if we can remove the need to download anything all-together thats better.

> When reading online about which applications would have access to environment variables exported with launchctl setenv, it sounded like GUI and shell applications would be able to see them, but I echo'd the values to be sure if I ran a build with react-native run-ios versus using Xcode directly that I'd see the same values, and they only show if I build using react-native run-ios.

Are you by any change using iTerm to call `launchctl`? If so - you might be hitting this issue: https://gitlab.com/gnachman/iterm2/issues/7654 .. try doing it from a regular MacOS terminal instead - and make sure XCode is completely killed before / after `launchctl setenv` has been called.

---

I actually think using a `/usr/bin/env bash -l` shell would be the best option here. If we did the script could be a simple one-liner:

```bash
node ${SRCROOT}/../scripts/download-realm.js ios --sync
```

ttahmouch:
> Are you by any change using iTerm to call launchctl? If so - you might be hitting this issue: https://gitlab.com/gnachman/iterm2/issues/7654 .. try doing it from a regular MacOS terminal instead - and make sure XCode is completely killed before / after launchctl setenv has been called.

I did use `iTerm 2` to use `vim` to modify my `~/.bash_profile` to have 
```
launchctl setenv HTTPS_PROXY "http://..."
launchctl setenv HTTP_PROXY "http://..."
launchctl setenv NO_PROXY "..."
```

I then manually sourced the `~/.bash_profile` to ensure the environment variables were set. I even ran the `react-native run-ios` build that succeeded using `iTerm 2`. It was only the shell that ran within Xcode itself that failed. For some reason, the shell, that is presumably the `xcodebuild` CLI, wrapped by the `react-native` build works fine. I'm not sure why they'd be any different if invoked from the GUI versus from a separate shell.

I can't recall if I tried restarting Xcode.

kneth:
@ttahmouch You need to restart Xcode as `launchd` only has an effect for **new** processes. Moreover, if you fork a process, I believe that `launchd` is not involved (so a new terminal in iTerm or Terminal will not see the changes until you restart the application).

kneth:
I am closing as my [comments](https://github.com/realm/realm-js/pull/2307#issuecomment-480182831) has been addressed.

RichardFevrier:
Hello @kneth @kraenhansen 

Would it be easier to replace this behavior:
    1) behing sure nvm is available
    2) using npm to download node (some MB)
    3) using node to execute a curl & tar to finally download realm

by a shell script doing only the last step (rewrite 'download-realm.js' -> 'download-realm.sh') ?

Especially in cases where people like me don't rely on nvm to use a node environment.

PR #2363, opened 2019-05-07
Adding doc for localListenerRealms

nhachicha:


lib/index.d.ts
```diff
@@ -612,6 +612,15 @@ declare namespace Realm.Sync {
         Off,
     }
 
+    /**
+     * LocalRealm
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Sync.LocalRealm.html }
+     */
+    interface LocalRealm {
+        readonly path: string;
+        realm (): Realm;
```
nirinchev: Bit of bikeshedding here, but since this is a method, it should be a verb - e.g. `getRealm`. By changing it to a verb, we will also convey the fact that executing the method is what causes the Realm to be opened, as opposed to a noun which implies that the Realm is already opened and we can access it by following the property/field.

kneth: I see your point but it is consistent with other parts of the API. For example we have [`Realm.objects()`](https://github.com/realm/realm-js/blob/master/lib/index.d.ts#L878) which should then be `Realm.getObjects()`.

nirinchev: I agree, the js API are quite inconsistent, but you could argue that we have two types of naming when talking about realms - there's `RealmObject.realm`, which gives you a reference to the existing open Realm and is supposedly cheap. Then there's `Realm.open()`, which opens a fresh Realm and is supposedly expensive. I'd argue that `LocalRealm.realm()` is semantically closer to the latter (expensive and opens a fresh Realm rather than give you an existing one). 

In any case, this is .NET, so I don't feel strongly about the API matching my beliefs 😜

kneth: sounds like we can do a big clean up in the next major version.

Issue #2282, opened 2019-03-13
Support React Native 0.59


cmelchior:
React Native 0.59 is released: https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059

It, among other things, enables us to support 64 bit binaries on React Native Android, but it also seems there are more internal changes we need to handle.

Some headers appear to be gone or have changed: https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmReact.mm#L27. It is unclear exactly what is required to fix this.

64 bit support seems to require changes to https://github.com/realm/realm-js/blob/master/react-native/android/src/main/jni/Application.mk#L3 and https://github.com/realm/realm-js/blob/master/react-native/android/build.gradle#L230, but this does not need to be part of the same PR as 0.59 support

See also: https://github.com/realm/realm-js/issues/2281#issuecomment-472349008

tgoyne:
Arm64 support will require writing a new version of the [hook function](https://github.com/realm/realm-js/blob/master/src/android/jsc_override.cpp#L60) since that's inherently platform-specific.

tgoyne:
On iOS we seem to build fine with 0.59 with no changes but it crashes on startup.

lodev09:
Is there a timeline on this? trying to upgrade our app to RN 0.59.
Thanks!

ferrannp:
The issues is with Android 64 bits. If you try to install the 64 bits in your phone the app will crash with:

```
java.lang.UnsatisfiedLinkError: couldn't find DSO to load: librealmreact.so
```

cmelchior:
Realm doesn't support 64 binaries on React Native just yet. This is being tracked separately here https://github.com/realm/realm-js/issues/2221 but we plan to address this as soon as possible but I don't have an exact timeframe just yet.

lodev09:
@cmelchior I see, can you recommend a stable/latest RN version that realm fully supports? I see other issues related to compatibility as well e.g. RN 0.58

ferrannp:
@cmelchior thanks! We should probably unified it in one issue.

@lodev09 the latest react-native `0.59.x` works with the latest realm. The only thing is that you should build only 32 bits variants for Android as 64 bits one will crash.

lodev09:
@ferrannp I'm getting dependency issue during fresh install though e.g. `https-proxy-agent`. Do you recommend a workaround of adding those in my `package.json` instead? 

They are filed in #2259 and #2281

[...]

PR #2364, opened 2019-05-07
Add arm64-v8a build support [WIP] 

### LINKED ISSUE #2221: [React Native] Add 64 bit support for Android
From August 1, 2019 Google Play will not accept apps with 32 bit only `.so` files. Currently, we only compile our React Native (Android) `.so` for ARMv7 and x86, but we should in good time provide 64 bit `.so` files too.

React Native (Android) added support for ARM64-v8a and x86-64 recently: https://github.com/facebook/react-native/pull/18754. 

Reference: https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html

reime005:
I've been trying to make Realm 64bit compatible for my react native app, but apparently this did not work so far. I'm stuck currently. The message I get on app startup before it crashed is the following:

```
java.lang.UnsatisfiedLinkError: couldn't find DSO to load: librealmreact.so caused by: dlopen failed: cannot locate symbol "_ZN5realm16ColumnBaseSimple18replace_root_arrayENSt6__ndk110unique_ptrINS_5ArrayENS1_14default_deleteIS3_EEEE"
```

Maybe someone has an idea on how to proceed.

This is a Draft.

This closes #2221 #2282

### DIFF
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -237,7 +237,7 @@ def getNdkBuildFullPath() {
             null)
     }
 
-    checkNdkVersion(ndkBuildFullPath);
+    // checkNdkVersion(ndkBuildFullPath);
 
     return ndkBuildFullPath
 }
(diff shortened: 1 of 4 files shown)


reime005:
I think we can continue on PR #2366 and close this

PR #2368, opened 2019-05-10
Publish before testing

kneth:
After the move from `realm-js-private` to `realm-js` we need to update when we pre-gyp.


Jenkinsfile
```diff
@@ -375,7 +382,7 @@ def buildAndroid() {
   }
 }
 
-def publish(nodeVersions, dependencies, tag) {
+def publish(nodeVersions, electronVersions, dependencies, tag) {
```
cmelchior: Is electronVersions used at all?

kneth: Yes, there is a loop in the function where it unstash the Electron builds.

kneth:
I am merging since it is unrelated tests failing.

PR #2373, opened 2019-05-15
Improve performance and correctness of chrome debugging

tgoyne:
This does a few things:

1. It adds proper read isolation support to the RPC server. We didn't actually enforce the inteded MVCC semantics when using chrome debugging since we rely on autorefreshes only being able to happen via the event loop, which means that they can't happen in the middle of a user's function. However, that isn't the case with the RPC server, because there's *two* event loops involved, and an autorefresh could happen in between two RPC calls even though the JS running in chrome never did any async operations. This used to be mostly a non-issue due to the lack of threads, but with synchronized Realms background writes happen even in JS. Most users are unlikely to have ever noticed this, but it resulted in some of the tests which relied on read isolation working failing inconsistently.
   
   The fix for this was to add a `beforenotify` listener, which merely by existing forces the RPC server to block autorefreshes until the browser calls `/callbacks_poll`.
2. Add a caching layer to the browser library to cut down on RPC requests. Repeated reads on the same property without a call to a mutating function in between now just return the same value, and the values of primitive properties on objects are sent along with the object id to enable pre-caching them. The caching is currently very cautious and flushes the entire cache after any possibly-mutating function is called or after a refresh. There's probably a lot of room to improve this later.
3. Add some missing bits to the RPC wrapper to make query-based Realms actually work and enable the subscription tests.
4. Fix some assorted bugs in the tests and test support stuff.

The caching doesn't help our tests much (they mostly just change something, then read only exactly the thing which should have changed), but seems to help a *lot* for real apps. Testing with [Rocket.Chat's react native client](https://github.com/RocketChat/Rocket.Chat.ReactNative), opening a channel previously performed 13k requests and took 110 seconds to fully load. With the caching, doing the same thing performs 4k requests and takes 20 seconds. This is still pretty painfully slow (it takes around a second when not using chrome debugging), but at least is a pretty big step towards being usable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,16 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Improve performance when using Chrome Debugging with React Native by adding caching and reducing the number of RPC calls required. Read-heavy workflows are as much as 10x faster. Write-heavy workflows will see a much smaller improvement, but also had a smaller performance hit to begin with. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
+* Improve performance when using Chrome Debugging with React Native by adding caching and reducing the number of RPC calls required. Read-heavy workflows are as much as 10x faster. Write-heavy workflows will see a much smaller improvement, but also had a smaller performance hit to begin with. (Issue: [#491](https://github.com/realm/realm-js/issues/491), PR: [#2373](https://github.com/realm/realm-js/pull/2373)).
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* Opening a query-based Realm using `new Realm` did not automatically add the required types to the schema when running in Chrome, resulting in errors when trying to manage subscriptions. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
-* The Chrome debugger did not properly enforce read isolation, meaning that reading a property twice in a row could produce different values if another thread performed a write in between the reads. This was typically only relevant to synchronized Realms due to the lack of multithreading support in the supported Javascript environments. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
-* The RPC server for Chrome debugging would sometimes deadlock if a notification fired at the same time as a Realm function which takes a callback was called. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
+* Opening a query-based Realm using `new Realm` did not automatically add the required types to the schema when running in Chrome, resulting in errors when trying to manage subscriptions. (PR: [#2373](https://github.com/realm/realm-js/pull/2373), since v2.15.0).
+* The Chrome debugger did not properly enforce read isolation, meaning that reading a property twice in a row could produce different values if another thread performed a write in between the reads. This was typically only relevant to synchronized Realms due to the lack of multithreading support in the supported Javascript environments. (PR: [#2373](https://github.com/realm/realm-js/pull/2373), since v1.0.0).
+* The RPC server for Chrome debugging would sometimes deadlock if a notification fired at the same time as a Realm function which takes a callback was called. (PR: [#2373](https://github.com/realm/realm-js/pull/2373), since v1.0.0 in various forms).
 
 ### Compatibility
-* Realm Object Server: 3.11.0 or later.
+* Realm Object Server: 3.21.0 or later.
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
+* Opening a query-based Realm using `new Realm` did not automatically add the required types to the schema when running in Chrome, resulting in errors when trying to manage subscriptions. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
+* The Chrome debugger did not properly enforce read isolation, meaning that reading a property twice in a row could produce different values if another thread performed a write in between the reads. This was typically only relevant to synchronized Realms due to the lack of multithreading support in the supported Javascript environments. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
+* The RPC server for Chrome debugging would sometimes deadlock if a notification fired at the same time as a Realm function which takes a callback was called. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
+
+### Compatibility
+* Realm Object Server: 3.11.0 or later.
```
```suggestion
* Realm Object Server: 3.21.0 or later.
```

src/js_realm.hpp
```diff
@@ -67,7 +67,24 @@ class RealmClass;
 
 template<typename T>
 class RealmDelegate : public BindingContext {
-  public:
+private:
+    void did_change(std::vector<ObserverState> const&, std::vector<void*> const&, bool) override {
+        HANDLESCOPE
```
kneth: Why do you need `HANDLESCOPE`?

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* Improve performance when using Chrome Debugging with React Native by adding caching and reducing the number of RPC calls required. Read-heavy workflows are as much as 10x faster. Write-heavy workflows will see a much smaller improvement, but also had a smaller performance hit to begin with. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
```
kneth: We have https://github.com/realm/realm-js/issues/491, and even if this PR doesn't solve it all, I suggest that we mention the issue.

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
+* Opening a query-based Realm using `new Realm` did not automatically add the required types to the schema when running in Chrome, resulting in errors when trying to manage subscriptions. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
```
```suggestion
* Opening a query-based Realm using `new Realm` did not automatically add the required types to the schema when running in Chrome, resulting in errors when trying to manage subscriptions. ([#2373](https://github.com/realm/realm-js/pull/2373), since v2.15.0)
```

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+* The Chrome debugger did not properly enforce read isolation, meaning that reading a property twice in a row could produce different values if another thread performed a write in between the reads. This was typically only relevant to synchronized Realms due to the lack of multithreading support in the supported Javascript environments. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
+* The RPC server for Chrome debugging would sometimes deadlock if a notification fired at the same time as a Realm function which takes a callback was called. (PR: ([#2373](https://github.com/realm/realm-js/pull/2373))).
```
```suggestion
* The RPC server for Chrome debugging would sometimes deadlock if a notification fired at the same time as a Realm function which takes a callback was called. ([#2373](https://github.com/realm/realm-js/pull/2373), since v1.0.0)
```

kneth: I think sync support for RN was added in v1.0.0 (or some beta before that) - and this potential deadlock might have been there from the beginning.

tgoyne: The handling of callbacks has changed a few times over the years. This specific deadlock was something I introduced the last time I touched it, but the problematic scenario didn't actually work before that and would have just broken in a different way.

src/js_realm.hpp
tgoyne: notify() allocates a JS object, so node requires an explicit scope for our local reference. Putting the HANDLESCOPE inside notify() doesn't work for the schema change case.

bmunkholm:
@tgoyne Could you please add "(since ?)" after the Fixes?

Issue #2293, opened 2019-03-19
Add support for new async open (speed up initial download)


bmunkholm:
See realm/realm-object-store#778 and related PR's in Java and Cocoa (available from the Epic)

PR #2381, opened 2019-05-23
Add support for new improved async open API's

cmelchior:
Closes https://github.com/realm/realm-js/issues/2293

This PR adds a number of new API's:

```typescript
// Check if the underlying Realm file already exists
boolean Realm.exists(config)

// New API options:
interface SyncConfiguration {
    // ...
    newRealmFileBehavior?: OpenRealmBehaviorConfiguration
    existingRealmFileBehavior?: OpenRealmBehaviorConfiguration
}

interface OpenRealmBehaviorConfiguration {
    readonly type: OpenRealmBehaviorType
    readonly timeOut?: number;
    readonly timeOutBehavior?: OpenRealmTimeOutBehavior;
}

const enum OpenRealmBehaviorType {
    OpenImmediately = "openImmediately",
    DownloadBeforeOpen = "downloadBeforeOpen"
}

const enum OpenRealmTimeOutBehavior {
    ThrowException = "throwException",
    OpenLocalRealm = "openLocal"
}

// Usage from JS - Common scenario where you want to wait for downloading 
// the initial Realm, and afterwards just open the local Realm immediately.
// IMO: Something like this should become the new default in the next major release.
let config = {
    sync: 
        newRealmFileBehavior: { 
            type: 'downloadBeforeOpen',
            timeOut: 30*1000,
            timeOutBehavior: 'throwException'
        },
        existingRealmFileBehavior: {
           type: 'openLocal'
        }
    }
}

// Default options if no behaviour is defined (these mimic current behaviour in JS):
let config = {
    sync: 
        newRealmFileBehavior: { 
            type: 'downloadBeforeOpen',
            timeOut: Number.MAX_VALUE,
            timeOutBehavior: 'throwException' 
        },
        existingRealmFileBehavior: {
            type: 'downloadBeforeOpen'
            timeOut: Number.MAX_VALUE,
            timeOutBehavior: 'throwException'
        }
    }
}
````

Naming this is rather hard. Current thinking is trying to use a common prefix for the types associated with this, `OpenRealmBehavior<X>`. It is, unfortunately, a bit long, but hopefully descriptive. 

Some other names or concepts considered:

* `OpenRealmBehaviourOptions` : No precedence for using suffix `Options`
* `OpenRealmBehaviorConfiguration`: We have precedence for sub configuration blocks, e.g. `SSLConfiguration`
* `InitialDownloadBehavior|SubsequentDownloadBehavior`: Feels awkward as you can define to open Realm immediately without any download involved.
* `newRealmFileBehavior` vs `openNewRealmBehavior`: Seems more descriptive as the behaviour depends on the presence of files on the device, not if the Realm is new or not on the server

TODO
- [ ] Changelog
- x ] Merge ObjectStore PR and redirect to that.
- [x] Unit tests for `Realm.exists()`. Verify it works on all platforms: Node, Android, iOS
- [x] Implement AsyncTask for JS
- [ ] Is it `Behavior` or `Behaviour`?
- [x] Implement proper defaults for new/currentFile behaviors. 
- [x] Unit and integration tests for new behavior API's. Especially around cancellation. 
- [x] Iterate on naming for public API

### CODE REVIEW

**Comment by bmunkholm** [comment]
@cmelchior Is the TODO list above up to date?

**Comment by cmelchior** [comment]
@bmunkholm Yes

File: docs/sync.js [hunk 1]
```diff
@@ -107,6 +107,22 @@
...
+ * The default behavior settings if you want to open a synchronized Realm immediately and start working on it.
+ * If this is the first time you open the Realm, it will be empty while the server data is being downloaded
+ * in the background.
+ *
+ * @typedef {Object} Realm.Sync.openLocalRealmBehavior
+ */
+
+/**
+ * The default behavior settings if you want to fully synchronize a Realm before it is opened.
+ * If this takes more than 30 seconds, and exception will be thrown.
```
tgoyne: an exception

lib/extensions.js
```diff
@@ -106,38 +113,112 @@ module.exports = function(realmConstructor, context) {
...
-            return promise;
+            // Opens the synchronized Realm
+            if (openLocalRealmImmediately) {
+                return openLocalRealm(realmConstructor, config);
+            } else {
+                let asyncOpenTask;
+                let openRemoteRealmPromise = new Promise((resolve, reject) => {
+                    asyncOpenTask = realmConstructor._asyncOpen(config, (realm, error) => {
+                        if (error) {
+                            setTimeout(() => {
```
tgoyne: Pushing the `if (error)` check inside the timeout function cuts down on duplicated code.

lib/extensions.js
```diff
@@ -106,38 +113,112 @@ module.exports = function(realmConstructor, context) {
...
                 }
+            }
 
-                return promise;
-            };
-            return promise;
+            // Opens the synchronized Realm
+            if (openLocalRealmImmediately) {
+                return openLocalRealm(realmConstructor, config);
+            } else {
```
tgoyne: The else here just kinda pointlessly indents all the following code.

lib/extensions.js
```diff
@@ -106,38 +113,112 @@ module.exports = function(realmConstructor, context) {
...
+                    case 'openImmediately': 
+                        openLocalRealmImmediately = true;
+                        break;
+                    default:
+                        throw Error("Invalid type: '" + type + "'. Only 'syncBeforeOpen' and 'openImmediately' is allowed.")
+                }
+            }
+
+            // Configure timeout
+            let timeOut = 0x7FFFFFFF; // Max value allowed by setTimeOut()
```
tgoyne: This doesn't appear to really simplify the code as compared to just skipping the whole timeout thing if the timeout is undefined.

lib/extensions.js
```diff
@@ -106,38 +113,112 @@ module.exports = function(realmConstructor, context) {
...
+                        if (throwOnTimeOut) {
+                            reject(new Error(config.sync.url + " could not be downloaded in the allocated time: " + timeOut + " ms."));
+                        } else {
+                            return resolve(openLocalRealm(realmConstructor, config));
+                        }
+                    }, timeOut)
+                });
+
+                let openPromise = Promise.race([timeOutPromise, openRemoteRealmPromise]);
+                openPromise._canceled = false;
```
tgoyne: This doesn't appear to ever be read?

src/android/platform.cpp
```diff
@@ -108,4 +109,11 @@ namespace realm {
         std::string cmd = "rm " + path;
         system(cmd.c_str());
     }
+
+    bool file_exists(const std::string &path)
```
tgoyne: Core already has `File::exists()`

src/js_realm.hpp
```diff
@@ -874,61 +888,50 @@ void RealmClass<T>::async_open_realm(ContextType ctx, ObjectType this_object, Ar
...
-            return;
-        }
-
-        // Ensure that all of our metadata tables are properly initialized if
-        // this is a query-based sync Realm
-        if (config.sync_config->is_partial) {
-            realm->update_schema({}, 0);
+        // Since the callback is only ever invoked once, make sure to release reference to AsyncOpenTask, so internal
+        // resources like the RealmCoordinator are correctly released.
+        delete ptr;
```
tgoyne: This is invalid. `AsyncOpenTaskClass<T>` will also call `delete` on the pointer when it is destroyed. The correct way to do this is to call `set_internal(nullptr)`.

tests/js/open-behavior-tests.js
```diff
@@ -0,0 +1,579 @@
...
+                        newRealmFileBehavior: Realm.Sync.openLocalRealmBehavior
+                    }
+                });
+                TestCase.assertFalse(Realm.exists(config));
+                return Realm.open(config);
+            })
+            .then(realm => {
+                TestCase.assertTrue(realm.path !== undefined);
+                realm.close();
+                return new Promise((resolve) => {
```
tgoyne: why? None of these Promises do anything.

docs/async_open_task.js
```diff
@@ -0,0 +1,29 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+
+class AsyncOpenTask {
+
+    /**
+     * FIXME
+     */
+    cancel() { }
+
```
kneth: many empty lines

docs/realm.js
```diff
@@ -341,6 +341,13 @@ class Realm {
      */
     static deleteFile(config) { }
 
+    /**
+     * Checks if the Realm already exists on disk.
+     * @param {Realm~Configuration} config The configuration for the Realm.
+     * @throws {Error} if anything in the provided `config` is invalid.
+     */
+    static exists(config) { }
```
kneth: does it return something?

src/js_realm.hpp
```diff
@@ -401,6 +404,36 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
...
+        ObjectType config_object = Value::validated_to_object(ctx, value);
+        realm::Realm::Config config;
+
+        static const String path_string = "path";
+        ValueType path_value = Object::get_property(ctx, config_object, path_string);
+        if (!Value::is_undefined(ctx, path_value)) {
+            config.path = Value::validated_to_string(ctx, path_value, "path");
+        }
+        else {
+            #if REALM_ENABLE_SYNC
```
kneth: Please don't indent `#if` and `#endif`.

src/js_realm.hpp
```diff
@@ -874,61 +888,50 @@ void RealmClass<T>::async_open_realm(ContextType ctx, ObjectType this_object, Ar
...
-    }
-    catch (const RealmFileException& ex) {
-        handleRealmFileException(ctx, config, ex);
-    }
-
-    auto session = user->session_for_on_disk_path(realm->config().path);
-    EventLoopDispatcher<WaitHandler> wait_handler([=, config=std::move(config),
-                                                   defaults=std::move(defaults),
-                                                   constructors=std::move(constructors)](std::error_code error_code) mutable {
+    std::shared_ptr<AsyncOpenTask>* ptr = new std::shared_ptr<AsyncOpenTask>(Realm::get_synchronized_realm(config));
```
kneth: Why do you need to create a pointer to a shared pointer?

cmelchior: I don't remember the specifics, but I ran into issues when not doing this. Also, it seems that SharedRealm uses a similar pattern.

But I did refactor a few things after this, so I can try to remove it again.

tgoyne: It's a common silly-looking pattern in the JS stuff. the JS wrapper object stores a `unique_ptr<T>`, so we end up with a `unique_ptr<shared_ptr<Realm>>` when we actually want shared ownership.

cmelchior:
> type: 'syncBeforeOpen' is kinda weird. Maybe downloadBeforeOpen: true? timeOut should maybe be timeOutMilliseconds to make sure it's clear what the units are.

I renamed `syncBeforeOpen` to `downloadBeforeOpen`. After looking at a number of other JS API's they all seem to just accept milliseconds without any definition of the unit, so I would probably prefer to keep `timeOut` the way it is for now.

tests/js/test-utils.js
```diff
@@ -0,0 +1,30 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
kneth:
```suggestion
// Copyright 2019 Realm Inc.
```

rajivshah3:
Hi @cmelchior , is it possible to release a beta from `0a305a1`? We currently use our own version of `realm-core` to include https://github.com/realm/realm-core/pull/3293 but we're trying to upgrade to the latest version of `realm-js`. Upgrading to 2.28.1 would require us to rebuild `realm-core` with the Windows path fix again, so I tried `yarn add https://github.com/realm/realm-js#0a305a1` but that didn't work either since fetching the submodules (e.g. `GCDWebServers`) from a `postinstall` script seems to be failing silently (not sure why)

kneth:
@rajivshah3 I am hoping to release it this week (Wednesday or Thursday).

rajivshah3:
Great, thanks!

PR #2384, opened 2019-05-24
NaN is supported by core

kneth:
## What, How & Why?

For historical reason, `Nan` was considered an illegal value. But Realm Core supports the value for floating-point properties.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


CHANGELOG.md
```diff
@@ -4,8 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* Not-a-number (NaN) was wrongly assumed not to be an illegal values in Realm Core (thanks to @jsflax for discovering this).
```
```suggestion
* Not-a-number (NaN) was wrongly assumed to be an illegal values in Realm Core (thanks to @jsflax for discovering this).
```

src/jsc/jsc_value.hpp
```diff
@@ -177,8 +177,7 @@ inline double jsc::Value::to_number(JSContextRef ctx, const JSValueRef &value) {
         throw jsc::Exception(ctx, exception);
     }
     if (isnan(number)) {
-        throw std::invalid_argument(util::format("Value '%1' not convertible to a number.",
-                                                 (std::string)to_string(ctx, value)));
+        std::nan("");
```
fealebenpae: return?

bmunkholm:
```suggestion
        return std::nan("");
```

bmunkholm: Seems like we should add some tests that would have detected this?

tgoyne: Isn't `if (isnan(number) { return std::nan(""); } return number;` just a roundabout way of writing `return number`? When does the extra check do anything?

kneth: I think you are right. But `NaN` can be represented by multiple bit patterns in IEEE 754. Maybe I am just seeing ghosts where no ghost exist.

kneth: I think the RN tests will detect it. Locally, I only ran the test using node.js.

kneth:
It seems to work now expect when debugging in React Native. I believe the reason is than JSON doesn't support nan as a number, so we cannot move nan between the app on the device and the debugger in Chrome.

src/jsc/jsc_value.hpp
bmunkholm: I mean, `std::nan("")` doesn't do anything... You at least need to return the value.
But as TG writes, why not remove it altogether - unless you can show with a test that it actually does something?

kneth:
I don't think we will be able to support NaN before we support Flipper for RN debugging.

PR #2405, opened 2019-06-06
Fix ROS dependency in release notes

bmunkholm:
The template for release notes did not update the ROS dependency and subsequent release notes were wrong on that point.

I've updated the Releases (where the release notes were published) manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Compatibility
-* Realm Object Server: 3.11.0 or later.
+* Realm Object Server: 3.21.0 or later.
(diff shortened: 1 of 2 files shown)


kneth:
@bmunkholm Thank you. I have apparently missed that when I did the last releases :cry:.

bmunkholm:
so did the reviewers :-)

bmunkholm:
@kneth Are the unstable tests tracked? :

**************************************************
*                    Failures                    *
**************************************************

1) Adapter should collapse schema changes into the table creation
  - Notification not recieved thrown

kneth:
@bmunkholm We haven't looked much into the test situation recently (after the merge of `realm-js-private`). In particular React Native and Linux test environments seem to suffer from race conditions, and my hunch is that our test framework is to blame.

PR #2410, opened 2019-06-12
Migrate to http based permissions

nirinchev:
## What, How & Why?

This closes https://github.com/realm/realm-js-private/issues/531

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


docs/sync.js
```diff
@@ -720,6 +710,13 @@ class User {
      */
     invalidatePermissionOffer(permissionOfferOrToken) { }
 
+    /**
+     * Asynchronously retrieve the permission offers that this user has created by invoking
+     * {@link Realm#Sync#User#offerPermissions offerPermissions}.
+     * @returns {Promise} a collection with a collection of {@link PermissionOffer PermissionOffer} objects.
```
```suggestion
     * @returns {Promise} A promise that, upon completion, contains a collection of {@link PermissionOffer PermissionOffer} objects.
```

bmunkholm:
We should retarget this the 3.0 branch.

kneth:
The `3.0` branch already contains the commits from this PR.

kneth:
Part of #2411

lib/permission-api.js
```diff
@@ -189,64 +78,60 @@ module.exports = {
...
-      expiresAt,
-      realmUrl,
-      mayRead,
-      mayWrite,
-      mayManage
+    const options = {
+      method: 'POST',
+      headers: { Authorization: this.token },
+      body: {
+        expiresAt,
```
cmelchior: This looks wrong? Shouldn't this be a `"expiresAt": expiresAt`?

nirinchev: They are equivalent.

docs/sync.js
```diff
@@ -720,6 +710,13 @@ class User {
      */
     invalidatePermissionOffer(permissionOfferOrToken) { }
```
cmelchior: Shouldn't this be `invalidatePermissionsOffer` (note plural permissions). It seems to me that we are normally talking about multiple permissions when referring to the `Permission` object (my head hurts already). But to me, this means that these methods should also be plural, so `invalidatePermissionsOffer` and `getPermissionsOffers`

PR #2425, opened 2019-06-28
Add Node.js 12 and Electron 5

fealebenpae:
## What, How & Why?

Update for the v8 breaking changes in Node.js 12:
- Use the Nan equivalents of `v8::NamedProperty*Callback`
- `v8::Value::ToString()` now takes a `v8::Context`
- Replace `v8::String::WriteUtf8` with `Nan::Utf8String`

Because the latest released version of `node-pre-gyp` doesn't know about the Node.js 12 ABI we also add [`abi_crosswalk.json`](https://github.com/mapbox/node-pre-gyp/blob/333de584a839c0e16331043108ace2607598a44e/lib/util/abi_crosswalk.json) from their master branch.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


src/node/node_object.hpp
```diff
@@ -26,15 +26,7 @@ namespace js {
...
-    v8::Local<v8::String> node_key;
-
-    // If we have just plain ASCII, we can skip the conversion from UTF-8
-    if (std::all_of(key.data(), key.data() + key.size(), [](char c) { return c <= 127; })) {
-        node_key = v8::String::NewExternal(isolate, new v8::ExternalOneByteStringResourceImpl(key.data(), key.size()));
-    }
-    else {
-        node_key = v8::String::NewFromUtf8(isolate, key.data(), v8::String::kNormalString, key.size());
-    }
+    v8::Local<v8::String> node_key = v8::String::NewFromUtf8(isolate, key.data(), v8::String::kNormalString, key.size());
```
fealebenpae: `v8::ExternalOneByteStringResourceImp` is sadly missing in latest v8.

src/node/node_object.hpp
```diff
@@ -26,15 +26,7 @@ namespace js {
 template<>
 inline v8::Local<v8::Value> node::Object::get_property(v8::Isolate* isolate, const v8::Local<v8::Object> &object, StringData key) {
     Nan::TryCatch trycatch;
-    v8::Local<v8::String> node_key;
-
-    // If we have just plain ASCII, we can skip the conversion from UTF-8
-    if (std::all_of(key.data(), key.data() + key.size(), [](char c) { return c <= 127; })) {
-        node_key = v8::String::NewExternal(isolate, new v8::ExternalOneByteStringResourceImpl(key.data(), key.size()));
```
fealebenpae: `v8::ExternalOneByteStringResourceImp` is sadly missing in latest v8. Should we implement it ourselves, is the optimization worth it?

src/node/node_function.hpp
```diff
@@ -38,14 +38,6 @@ inline v8::Local<v8::Value> node::Function::call(v8::Isolate* isolate, const v8:
 
 template<>
 inline v8::Local<v8::Value> node::Function::callback(v8::Isolate* isolate, const v8::Local<v8::Function> &function, const v8::Local<v8::Object> &this_object, size_t argc, const v8::Local<v8::Value> arguments[]) {
-    if (!isolate->GetCallingContext().IsEmpty()) {
```
fealebenpae: v8 has removed `GetCallingContext()` because of tail calls. I don't know how to implement this optimization otherwise.

src/node/node_string.hpp
```diff
@@ -48,9 +48,9 @@ inline String<node::Types>::String(const v8::Local<v8::String> &s) {
     if (s.IsEmpty() || s->Length() == 0) {
         return;
     }
-    m_str.resize(s->Utf8Length());
```
fealebenpae: The `Utf8Length` and `WriteUtf8` APIs now take an `v8::Isolate` - should we `ifdef` the calls or do we not care about transforming the strings as optimally as possible here?

src/node/node_class.hpp
```diff
@@ -188,7 +188,7 @@ inline v8::Local<v8::FunctionTemplate> ObjectWrap<ClassType>::create_template()
     if (s_class.string_accessor.getter || s_class.index_accessor.getter || s_class.index_accessor.setter) {
         // Use our own wrapper for the setter since we want to throw for negative indices.
         auto &string_accessor = s_class.string_accessor;
-        instance_tpl->SetNamedPropertyHandler(string_accessor.getter ? string_accessor.getter : get_nonexistent_property, set_property, 0, 0, string_accessor.enumerator);
+		Nan::SetNamedPropertyHandler(instance_tpl, string_accessor.getter ? string_accessor.getter : get_nonexistent_property, set_property, 0, 0, string_accessor.enumerator);
```
tgoyne: I switched this away from Nan::SetNamedPropertyHandler in 9079cb9d78e3e50491cc196f34bc1d6e225ab6fa because the Nan wrapper is broken, and it doesn't appear to have been fixed since then.

src/node/node_object.hpp
tgoyne: This was a pretty big speedup for the mssql loader, and not because the loader was doing anything unusual.

src/node/node_class.hpp
fealebenpae: That's unfortunate - this API is completely gone in latest v8. We'll have to convert everything to use `Name` instead of `String` manually.

src/node/node_object.hpp
fealebenpae: Alright, I'll try to re-implement `v8::ExternalOneByteStringResourceImpl` - it's pretty simple so it should be okay.

src/node/node_string.hpp
kneth: We don't have any benchmark which can guide us here. I suggest that we leave it as it is.

Jenkinsfile
```diff
@@ -306,17 +306,17 @@ def buildMacOS(workerFunction) {
 
 def buildWindows(nodeVersion, arch) {
   return {
-    myNode('windows && nodejs && cph-windows-01') {
+    myNode('windows && nodejs') {
```
kneth: I can't remember what the problem was but builds on `cph-windows-01` was failing.

src/node/node_class.hpp
fealebenpae: What should we do in our `wrap`pers when we encounter a `Name` that isn't a `String`? Throw an error, stringify, do nothing?

tgoyne: The other thing it can get is `Symbol`. I think the correct thing to do there is return `undefined` as we don't have any Symbol-keyed properties?

fealebenpae:
@tgoyne can you take one last look, please? I think I covered all the cases where we were in danger of performance regressions.

src/node/node_object.hpp
```diff
@@ -104,8 +118,9 @@ inline std::vector<node::String> node::Object::get_property_names(v8::Isolate* i
     std::vector<node::String> names;
     names.reserve(count);
 
+	v8::Local<v8::Context> ctx = isolate->GetCurrentContext();
```
tgoyne: Indentation is off here

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
+### Breaking Changes
+* On Node.js and Electron, change notification callbacks will no longer be invoked directly by the `Realm.write()` call. Instead, they will be invoked asynchronously on a subsequent event loop tick.
```
tgoyne: This seems like a significant regression in functionality. Invoking the change callbacks synchronously makes it difficult to accidentally read the updated data before the callback fires. If the change callback does anything important then any code which runs in between them is very likely to be incorrect.

fealebenpae: I wasn't able to figure out a v8 API that allows us to detect whether the change notification callback fires because of a call to `write()` or because another thread/process committed a write - that is to say `v8::Isolate::GetCallingContext()` which we called in `node::Function::callback` is removed. I also checked n-api and there isn't anything that seems useful there either.

fealebenpae: Do you have any ideas how we can implement this functionality differently?

tgoyne: After digging into this more, the problem isn't that MakeCallback makes it async; it actually is still sync, but it runs microtasks before invoking the callback. This seems to be the correct thing to do when we're invoking a top-level callback (the tests hang in several places that look valid without it), but when we're invoking a callback from within a function called by JS it means that we process microtasks from within that function, which is definitely not something we want to do.

Newer versions of v8 don't seem to have anything comparable. GetCallingContext() walks the call stack looking for JS frames, but with tail calls there may not be any JS frames on the call stack even if we were called by JS.

I think maybe what we need is a node-specific version of `EventLoopSignal` that wraps the invocation of the callback in a `CallbackScope`. Doing something like the following would probably fix it:

```
{
    auto isolate = Isolate::GetCurrent();
    v8::HandleScope handle_scope(isolate);
    node::CallbackScope callback_scope(isolate);
    data.callback();
}
```

The problem with this is that it's a node dependency in object store, which maybe complicates things, and it processes microtasks more often than we currently do (not all event loop signals end up invoking a JS function), which may or may not matter.

An alternative would be to track if we're the top level of execution ourselves and do the current logic of maybe using MakeCallback based on that. A crude way of doing that would be to do something like:

```
// EventLoopSignal callback
{
    g_in_top_level_callback = true;
    data.callback();
    g_in_top_level_callback = false;
}

template<>
inline v8::Local<v8::Value> node::Function::callback(v8::Isolate* isolate, const v8::Local<v8::Function> &function, const v8::Local<v8::Object> &this_object, size_t argc, const v8::Local<v8::Value> arguments[]) {
    if (!g_in_top_level_callback) {
        // if there are any JavaScript frames on the stack below this one we don't need to
        // go through the trouble of calling MakeCallback. MakeCallback is only for when a
        // thread with no JavaScript frames on its stack needs to call into JavaScript, like in
        // an uv_async callback.
        return call(isolate, function, this_object, argc, arguments);
    }

    v8::TryCatch trycatch(isolate);
    
    auto recv = this_object.IsEmpty() ? isolate->GetCurrentContext()->Global() : this_object;
    g_in_top_level_callback = false;
    auto result = ::node::MakeCallback(isolate, recv, function, (int)argc, const_cast<v8::Local<v8::Value>*>(arguments));
    g_in_top_level_callback = true;

    if (trycatch.HasCaught()) {
        ::node::FatalException(isolate, trycatch);
    }
    return result;
}
```

Maybe there's a better way to do this.

PR #2454, opened 2019-07-12
react example - upgrade to rn 0.59

kneth:


kneth:
I am closing for now. We should rewrite our example add using a recent version of React Native.

PR #2475, opened 2019-08-09
[RN] Keep retrying in case of network timeouts

kneth:
## What, How & Why?
Network errors and timeout can indicate that the device is offline, and we should keep retrying. Currently the retrying is only done on React Native.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


kneth:
Doesn't solve the issue

PR #2483, opened 2019-08-20
[RJS-6] Bypass the sync proxy on Cloud

fealebenpae:
## What, How & Why?
See https://github.com/realm/realm-object-store/pull/824

I'm not sure how to add tests, because this relies on functionality that resides entirely on Cloud and not in ROS.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+X.Y.Z Release notes (TBD)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* None.
+
+### Compatibility
+* Realm Object Server: 3.21.0 or later.
+* APIs are backwards compatible with all previous release of realm in the 2.x.y series.
+* File format: Generates Realms with format v9 (Reads and upgrades all previous formats)
+
+### Internal
+* Add support for direct access to sync workers on Cloud, bypassing the Sync Proxy [RJS-6](https://jira.mongodb.org/browse/RJS-6)
+
 2.29.2 Release notes (2019-8-14)
 =============================================================
 ### Enhancements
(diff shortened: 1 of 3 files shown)


tgoyne:
Why isn't the functionality in ROS? We need a test server to develop the clients against which supports the full feature set of the production server (even if it's not in a way that's actually production quality). That doesn't have to be ROS, but if ROS is no longer suitable for that then it's pretty critical that we start working on a replacement.

kneth:
@tgoyne Can we extend the test server used in the data adapters?

tgoyne:
The adapter/notifier tests now always run against real ROS specifically because the test server kept having issues where it didn't behave like the real thing.

nirinchev:
The SDKs can mock http requests/responses if they want to run unit tests. For integration tests that rely on cloud functionality, it's very ambitious to try and run them locally as that would entail spinning up a local k8s cluster + eventually running stitch and mongodb on the same machine. Instead, for integration tests that run as part of RP jobs, we should provision cloud instance(s) and SDKs should ensure their tests use random users and random realms to avoid collisions between test runs. When developing new features against unreleased functionality, we can simply override the instance url to point to something in the local k8s cluster.

tgoyne:
"Just mock things and hope that you guess correctly about how the server will behave" is not a very good strategy for writing software that actually works. A test which verifies that the request handler and the mock agree on the structure of the data isn't testing anything useful or relevant.

Spinning up a local k8 cluster to test against isn't inherently something that should be difficult.

kneth:
@fealebenpae @tgoyne I suggest that we merge this, and work on find a generic way to test against a "local cloud".

PR #2503, opened 2019-08-30
fix error on refreshAdminToken and refreshAccessToken

blagoev:
This fixes RN apps when going offline to not show error dialogs while refreshing tokens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ X.Y.Z Release notes (TBD)
 * None.
 
 ### Fixed
-* None.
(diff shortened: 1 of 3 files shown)


nirinchev:
It's out of scope for this PR, but it'd be nice to expose a mechanism to set a custom logger for these errors so they don't end up on the console.

bmunkholm:
Need to update the changelog.

blagoev:
@nirinchev Indeed I was going to do that, but then realized it is getting out of scope since we don't have the native implementation for RN. So hence this solution of just fixing this only problem.

@bmunkholm Thanks, forgot about that. will do

PR #2473, opened 2019-08-09
Fix readme for env setup

blagoev:
updated readme for building on windows

PR #2504, opened 2019-08-30
Fix node win build debug

blagoev:


.vscode/launch.json
```diff
@@ -9,18 +9,33 @@
             "request": "launch",
             "name": "Debug LLDB",
             "program": "node",
+            "windows": {
+                "program": "C:\\Progra~1\\nodejs\\node.exe"
```
bmunkholm: Will node.exe always be placed here or could it be installed elsewhere?

README.md
```diff
@@ -34,18 +34,17 @@ In case you don't want to use the precompiled version on npm, you can build Real
...
+Clone RealmJS repository:
 
 ```
 git clone https://github.com/realm/realm-js.git
-```
-
-Then in the cloned directory:
-
-```
+cd relm-js
```
bmunkholm:
```suggestion
cd realm-js
```

README.md
```diff
@@ -61,8 +60,28 @@ Note: If you have cloned the repo previously make sure you remove your node_modu
 - `./gradlew publishAndroid`
 - The compiled version of the Android module is here: `<project-root>/android`
 
+Note: On Windows the RealmJS repo should be cloned with symlynks enabled
```
bmunkholm:
```suggestion
Note: On Windows the RealmJS repo should be cloned with symlinks enabled 
```

README.md
```diff
@@ -61,8 +60,28 @@ Note: If you have cloned the repo previously make sure you remove your node_modu
...
 - The compiled version of the Android module is here: `<project-root>/android`
 
+Note: On Windows the RealmJS repo should be cloned with symlynks enabled 
+```
+#run in elevated command prompt
+git clone -c core.symlinks=true https://github.com/realm/realm-js
+
+```
+
+or manually create the symlinks using directory junctions
```
bmunkholm: When would one want to do it manually instead of like above?

README.md
```diff
@@ -75,12 +94,29 @@ npm install --build-from-source=realm
          - Open an elevated command prompt (As Administrator)
 
             ```
-            npm install -g --production windows-build-tools
-            ```
-
+            npm install -g --production windows-build-tools --vs2015
+            ```  
+                 
+
```
bmunkholm:
```suggestion
```

README.md
```diff
@@ -75,12 +94,29 @@ npm install --build-from-source=realm
...
-
+            npm install -g --production windows-build-tools --vs2015
+            ```  
+                 
+             
     * Option 2: Manually install and configure
 
         - Check [node-gyp](https://github.com/nodejs/node-gyp) manual for custom installation procedure for Windows
+        
+   Install openssl libraries vith vcpkg
```
bmunkholm:
```suggestion
   Install openssl libraries with vcpkg
```

README.md
```diff
@@ -100,6 +136,10 @@ VSCode has good support for debugging JavaScript, but to work with C++ code, you
 
 To begin, you will need to build the node addon and prepare the test environment:
 
+```
```
bmunkholm: Why this?

README.md
bmunkholm: Shouldn't this be mentioned further up under the "Clone RealmJS repository:" section?

README.md
```diff
@@ -61,8 +60,28 @@ Note: If you have cloned the repo previously make sure you remove your node_modu
...
+mklink /j "src" "../../../../../src/"
+mklink /j "vendor" "../../../../../vendor"
+cd realm-js\tests\react-test-app\android\app\src\main
+mklink /j assets "../../../../../data"
+```
+
 ### Building for nodejs:
 Be sure you have python2.7 as the default python. 3.x won't work, and it's not enough to use `--python=python2.7` as parameter to npm.
+```
+brew install python@2
```
bmunkholm: Are we expecting everyone to use brew?

package.json
```diff
@@ -50,6 +50,8 @@
     "test": "scripts/test.sh",
     "package": "npm ci --build-from-source=realm && node-pre-gyp package",
     "install": "node-pre-gyp install --fallback-to-build",
+    "rebuild": "npm run isWin -s && SET REALMJS_USE_DEBUG_CORE=1 & npm run rebuild-changes || REALMJS_USE_DEBUG_CORE=1 && npm run rebuild-changes",
```
bmunkholm: Do you need the debug version of core?

README.md
blagoev: when one has the repo already cloned

README.md
blagoev: brew is already mentioned in the readme before this mention. 
https://github.com/realm/realm-js/blob/5cf0c6d1121e73553dbc4f0696a5c1f503f3e816/README.md#L121

package.json
blagoev: yes. the build fails without this

README.md
blagoev: yes. will fix

.vscode/launch.json
blagoev: This is the default install location on Windows. The problem is that by only having "program: node" VS Code finds the C:\Program Files\ path and VSCode seems to mess up and tries to run the given "program" from the "current dir\Files\nodejs\node.exe. So the build fails with cryptic message like path not found C:\realm-js\Files\nodejs\node.exe. This is a bug I guess is in VS Code where it fails to handle paths with spaces. Previous versions of VS Code were working fine with this. So here we encode the path with the shorthand 8.3 version to the default location. This way at least the fail message will be exact so the developer will know what is not found and where it was searched for. 

If you have any better suggestion will be glad to change this.

README.md
blagoev: thanks will fix

README.md
```diff
@@ -34,19 +34,35 @@ In case you don't want to use the precompiled version on npm, you can build Real
...
-Then in the cloned directory:
+Note: On Windows the RealmJS repo should be cloned with symlinks enabled 
+```
+#run in elevated command prompt
+git clone -c core.symlinks=true https://github.com/realm/realm-js
 
 ```
-git submodule update --init --recursive
+
+or manually create the symlinks using directory junctions
```
bmunkholm:
```suggestion
or manually create the symlinks using directory junctions if you already have the repo cloned.
```

README.md
bmunkholm: It's after. Then I guess we should mention it as prerequisites in the top? Or mention that's just an example and people can install it the way they want?

blagoev: What I meant is it was mentioned in here before I added this change.
I guess it is just an example. Should we explicitly state this is an example how to install python?

blagoev:
@bmunkholm I added homebrew explanations here https://github.com/realm/realm-js/pull/2473

README.md
```diff
@@ -63,6 +79,9 @@ Note: If you have cloned the repo previously make sure you remove your node_modu
 
 ### Building for nodejs:
 Be sure you have python2.7 as the default python. 3.x won't work, and it's not enough to use `--python=python2.7` as parameter to npm.
+```
+brew install python@2
```
bmunkholm:
```suggestion
brew install python@2     # or use any other means of installing 
```

scripts/test.sh
```diff
@@ -110,7 +110,9 @@ cleanup() {
 
   # Restore nvm state
   if [ -n "$nvm_old_default" ]; then
+    echo Restoring nvm default to $nvm_old_default
     nvm alias default $nvm_old_default
+    echo nvm default restored succesfuly
```
bmunkholm:
```suggestion
    echo nvm default restored successfully 
```

scripts/test.sh
```diff
@@ -340,7 +354,7 @@ case "$TARGET" in
   check_test_results ReactTests
   ;;
 "node")
-  npm run check-environment
+npm run check-environment
```
bmunkholm:
```suggestion
  npm run check-environment
```

tests/js/index.js
```diff
@@ -35,10 +35,11 @@ if (isNodeProcess && process.platform === 'win32') {
 }
 
 // catching segfaults during testing can help debugging
-if (isNodeProcess) {
-    const SegfaultHandler = node_require('segfault-handler');
-    SegfaultHandler.registerHandler("crash.log");
-}
+//uncomment to enable segfault handler
```
bmunkholm: Why don't we want to catch segfaults by default?

README.md
blagoev: This is fixed here 
https://github.com/realm/realm-js/pull/2473/commits/857afe7270d9070b6868850170d33785b788edb8

So I will resolve this one

tests/js/index.js
blagoev: Cause segfault-handler is not working right now for us and it outputs a lot of ... by default which is not useful at all

README.md
```diff
@@ -75,17 +93,33 @@ npm install --build-from-source=realm
          - Open an elevated command prompt (As Administrator)
 
             ```
-            npm install -g --production windows-build-tools
-            ```
-
+            npm install -g --production windows-build-tools --vs2015
+            ```  
+
```
bmunkholm:
```suggestion
```

README.md
```diff
@@ -75,17 +93,33 @@ npm install --build-from-source=realm
          - Open an elevated command prompt (As Administrator)
 
             ```
-            npm install -g --production windows-build-tools
-            ```
-
+            npm install -g --production windows-build-tools --vs2015
+            ```
```
bmunkholm:
```suggestion
            ```
```

README.md
```diff
@@ -23,46 +23,64 @@ The API reference is located at [realm.io/docs/javascript/latest/api/](https://r
...
-- The compiled version of the Android module is here: `<project-root>/android`
+* `cd react-native/android`
+* `./gradlew publishAndroid`
+* The compiled version of the Android module is here: `<project-root>/android`
 
 ### Building for nodejs:
 Be sure you have python2.7 as the default python. 3.x won't work, and it's not enough to use `--python=python2.7` as parameter to npm.
+```
+brew install python@2
+```
```
bmunkholm: For Windows should I do the `npm install --build-from-source=realm` before or after the additional steps described below?

blagoev: before

blagoev:
@bmunkholm please revert you last section edit.

blagoev:
also @kneth What do you think? Maybe we can keep it as @bmunkholm suggested? It's a matter of taste I think

bmunkholm:
@blagoev Unless you feel it's worse,  I'd prefer to just keep it. It is related to the general change as the whole section on what to do with Windows got longer and the section visually looked complex with all those indents and single bullets. Doing another PR for this seems unnecessary.

blagoev:
yes lets keep it. it's not worse. it's just me loosing track if this relates to windows section above or not. Anyway we need to have separate sections by platform going step by step over what is required and what are the prerequisites and the steps needed. Currently it's not clear enough and the person looking into it is coming from a platform standpoint and going on from there. I find the React Native framework getting started page to good example of that.

PR #2539, opened 2019-09-26
Add predicate function filter option to Adapter

fealebenpae:
## What, How & Why?
Make it possible for the Adapter to filter realms to observe with a predicate function instead of a regular expression. Needs https://github.com/realm/realm-object-store/pull/850 to be merged first.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -1124,7 +1124,7 @@ class Adapter {
 	 * @param {string} serverUrl - the sync server to listen to
 	 * @param {SyncUser} adminUser - an admin user obtained by calling {@linkcode Realm.Sync.User.login|User.login} with admin credentials.
 	 * @param {(string|Realm.Sync.Adapter~RealmWatchPredicate)} filter - a filter used to determine which changed Realms should be monitored -
-	 *  can be a regular expression string or a predicate function. Use `.*` to match all all Realms.
+	 *  can be a regular expression string or a predicate function. Use `'.*'` to match all Realms.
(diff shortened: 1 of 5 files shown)


tests/js/adapter-tests.ts
```diff
@@ -150,6 +150,30 @@ describe('Adapter', () => {
         tmpDir.removeCallback();
     });
 
+    it("predicate functions", async () => {
```
tgoyne: This should also test that non-matching paths aren't watched (currently the test would pass if the adapter unconditionally watched all realms and ignored the predicate).

docs/sync.js
```diff
@@ -1123,9 +1123,9 @@ class Adapter {
 	 * @param {string} localPath - the local path where realm files are stored
 	 * @param {string} serverUrl - the sync server to listen to
 	 * @param {SyncUser} adminUser - an admin user obtained by calling {@linkcode Realm.Sync.User.login|User.login} with admin credentials.
-	 * @param {string} regex - a regular expression used to determine which changed Realms should be monitored -
-	 *  use `.*` to match all all Realms
-	 * @param {function(realmPath)} changeCallback - called when a new transaction is available
+	 * @param {(string|Realm.Sync.Adapter~RealmWatchPredicate)} filter - a filtered used to determine which changed Realms should be monitored -
```
tgoyne: s/filtered/filter/

PR #2547, opened 2019-10-02
Fix #2294 null _NativeModules$Realm.debugHosts without remote debug

### LINKED ISSUE #2294: [Android] undefined is not an object (evaluating '_NativeModules$Realm.debugHosts')
## Goals

Simply trying to launch the app. After installing Realm Database it has worked and I could successfully write data to it. Since then, it just doesn't launch anymore. Tried to rebuild, clear caches, and really, tried everything I know of. Followed all steps from the docs and troubleshooting, and tried many solutions from the other related issues (#1150, #527, #762), but nothing seems to work.

## Expected Results

Launch the app normally (not using Realm on startup or anything like that, it only gets called upon touching a button).

## Actual Results

image

Above image is from an emulator.

Also tried to get a stacktrace from the Android Device Monitor, however, I am being unsuccessful.

## Steps to Reproduce

Install Realm normally, check all changes from `react-native link` are OK, write some data, and restart the app.

- Realm JS SDK Version: 2.25.0
- React Native: 0.57.4

borysstach:
## What, How & Why?
If some libraries in project use `atob` realm always run in chrome debugger mode.
It is not reliable check as is mentioned in StackOverflow commented above this check. ([link](https://stackoverflow.com/a/42839384/3090989))
Removing this fixed my issue similar to #2294

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -56,12 +56,6 @@ function getContext() {
             return 'chromedebugger';
         }
 
-        // Check if its in remote js debugging mode
-        // https://stackoverflow.com/a/42839384/3090989
-        if (typeof atob !== 'undefined') {
-            return 'chromedebugger';
-        }
-
         // Otherwise, we must be in a "normal" react native situation.
         // In that case, the Realm type should have been injected by the native code.
         // If it hasn't, the user likely forgot to run link.


kraenhansen:
Perhaps it would be better to check for the existence of the `__REMOTEDEV__` global (see https://stackoverflow.com/a/43461409) instead of removing the if-expression entirely?

borysstach:
@kraenhansen I checked this solution and in my project __REMOTEDEV__ is 'undefined' with Chrome remote debug on. I can add this as replacement, but im not convinced it will change anything.

kraenhansen:
We need a check to replace this - cannot simply delete the lines as they were addressing another issue.

I'll close this pull request in favour of https://github.com/realm/realm-js/pull/2681, which updates more of the logic determining the execution context.

Thank you @borysstach for finding and addressing this bug!

PR #2534, opened 2019-09-24
Migrate away from rnpm in package.json, use react-native.config.js instead

sam-drew:
## What, How & Why?

Add a valid react-native.config.js, and removed rnpm usage in package.json

This closes #2476 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 🚦 Tests
* [ ] 📝 Public documentation PR created or is not necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [ ] typescript definitions file is updated
* [ ] jsdoc files updated
* [ ] Chrome debug API is updated if API is available on React Native

PR #2555, opened 2019-10-09
Revert "Migrate away from rnpm in package.json, use react-native.config.js instead"

kneth:
Reverts realm/realm-js#2534


kneth:
It doesn't seem to have any impact on our testing.

PR #2567, opened 2019-10-17
Update user-methods.js to check if XMLHttpRequest is defined

### LINKED ISSUE #2406: XMLHttpRequest is not defined when Realm is used in the electron main process
Studio is using Realm mostly in the renderer process, but there are a few cases where we call some of the API in the main process - notably when handling links with custom protocol - these fire in the main process. To be able to hand off a user to the renderer process, we call `Realm.Sync.User.login`, but that fails with the following error:

```
[ELECTRON] ReferenceError: XMLHttpRequest is not defined
[ELECTRON]     at fetchWithTimeout (/Users/nikolairinchev/work/realm-studio/node_modules/realm/lib/user-methods.js:82:17)
[ELECTRON]     at next (/Users/nikolairinchev/work/realm-studio/node_modules/realm/lib/user-methods.js:146:9)
[ELECTRON]     at Promise (/Users/nikolairinchev/work/realm-studio/node_modules/realm/lib/user-methods.js:161:13)
[ELECTRON]     at new Promise (<anonymous>)
[ELECTRON]     at performFetch (/Users/nikolairinchev/work/realm-studio/node_modules/realm/lib/user-methods.js:159:16)
[ELECTRON]     at _authenticate (/Users/nikolairinchev/work/realm-studio/node_modules/realm/lib/user-methods.js:351:12)
[ELECTRON]     at performFetch.then (/Users/nikolairinchev/work/realm-studio/node_modules/realm/lib/user-methods.js:374:20)
[ELECTRON]     at process._tickCallback (internal/process/next_tick.js:68:7)
```

While I know the Realm API would not work in the main process, it's a bit surprising that purely js API fail. It would be great if we could support those without having to resort to workarounds.

kraenhansen:
## What, How & Why?

This checks if `XMLHttpRequest` is defined before using it.

This closes #2406 (which was already closed) and some of https://github.com/realm/realm-js/issues/2274#issuecomment-472598810

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
+* Fixed user methods (authentication etc) when running on the Electron main process, where `XMLHttpRequest` is `undefined`. ([#2274](https://github.com/realm/realm-js/issues/2274), since v2.24.0)
 * None.
 
 ### Compatibility
--- a/lib/user-methods.js
+++ b/lib/user-methods.js
@@ -107,7 +107,7 @@ function fetchWithTimeout(input, init) {
 // Perform a HTTP request, enqueuing it if too many requests are already in
 // progress to avoid hammering the server.
 const performFetch = (function() {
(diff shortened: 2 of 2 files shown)


kraenhansen:
Merging, since the failing tests seems more related to the harness than the actual changes to code.

PR #2574, opened 2019-10-21
Fix #2501 -- Made the remote_path relative

### LINKED ISSUE #2501: Support mirroring of binary
## Goals

We would like to be able to host/cache the Realm Binaries on a server inside the company. This should be possible according to the documentation of 'node-pre-gyp', unfortunately because of the way Realm is configured, only the host-part of the mirror is used. Because of company rules, we also need to add a context-root to the mirror URL.

Currently when setting the mirror-property 'realm_binary_host_mirror' to any URL with a path will be cut to just the protocol and host because Realm's configuration for 'remote_path' is an absolute path ("/node-pre-gyp/{version}/"). Changing it to "./node-pre-gyp/{version}/" will make node-pre-gyp accept the configured property.

FYI: The cutting of the path for absolute remote-path values is documented behavior in node-pre-gyp, so this is not a bug on their part!

## Expected Results

I expect my Realm-Binary to be downloaded from my mirror-host.

## Actual Results

Node aborts with an exception because it is trying to connect to the wrong host and the fall-back of rebuilding from source will not work because of proxy-constraints.

malice00:
## What, How & Why?

This closes # 2501

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,6 @@
     "module_name": "realm",
     "module_path": "./compiled/{node_abi}_{platform}_{arch}/",
     "host": "https://static.realm.io",
-    "remote_path": "/node-pre-gyp/{version}/"
+    "remote_path": "./node-pre-gyp/{version}/"
   }
 }


kneth:
@malice00 I have release v3.4.0-alpha.1 for you to test.

malice00:
Works as expected! Could this make it into (one of) the next release(s)?

kneth:
@malice00 Sorry for the late reply (bad excuses: meetings and a couple of days off). I hope to do a release by the end of the week.

PR #2586, opened 2019-11-01
Podspec

### LINKED ISSUE #2460: Incompatible with React Native Autolinking (React Native v0.60)
## Goals

Use the Realm.js with React Native Autolinking correctly.

## Expected Results

Don't show warning information.

Let the CocoaPods don't show warning after executing `pod install`.

## Actual Results

After the `pod install`, it shows the following warning:

```
[!] use_native_modules! skipped the react-native dependency 'realm'. No podspec file was found.
    - Check to see if there is an updated version that contains the necessary podspec file
    - Contact the library maintainers or send them a PR to add a podspec. The react-native-webview podspec is a good example of a package.json driven podspec. See
    https://github.com/react-native-community/react-native-webview/blob/master/react-native-webview.podspec
    - If necessary, you can disable autolinking for the dependency and link it manually. See
    https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#how-can-i-disable-autolinking-for-unsupported-library
```

## Steps to Reproduce

```bash
cd ${REACT_NATIVE_PROJECT_ROOT_PATH}
npm install --save realm
cd ios && pod install & cd ..
```

- Realm JS SDK Version: `realm-js 2.29.1`
- Node or React Native: `React Native v0.60.3`
- Client OS & Version: `iOS 12.2`
- Which debugger for React Native: `None`

kraenhansen:
## What, How & Why?

This closes #2460 by adding a Podspec file that can build Realm JS for React Native on iOS.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manual with a test app)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added a CocoaPod Podspec file, enabling [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) for React Native on iOS. ([#2586](https://github.com/realm/realm-js/pull/2586)).
+* Added a CocoaPod Podspec file, enabling [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) for React Native on iOS. ([#2586](https://github.com/realm/realm-js/pull/2586))
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added a CocoaPod Podspec file, enabling [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) for React Native on iOS. ([#2586](https://github.com/realm/realm-js/pull/2586)).
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * File format: Generates Realms with format v9 (Reads and upgrades all previous formats)
 
 ### Internal
-* None.
+* Fixed a couple of flaky tests.
+* Upgraded to Xcode 10.3 for building and testing.
 
 3.3.0 Release notes (2019-10-18)
 =============================================================
--- a/RealmJS.podspec
+++ b/RealmJS.podspec
@@ -6,14 +6,15 @@ package = JSON.parse(File.read(File.expand_path('package.json', __dir__)))
 git_path_local = File.expand_path('.git', __dir__)
 git_path = File.directory?(git_path_local) ? "file://#{git_path_local}" : 'https://github.com/realm/realm-js.git'
 
-node_modules_path = File.expand_path('../../node_modules', __dir__)
+app_path = File.expand_path('../..', __dir__)
+node_modules_path = File.expand_path('node_modules', app_path)
 react_native_module_path = File.expand_path('react-native', node_modules_path)
 if not File.directory?(react_native_module_path)
   raise "Couldn't find React Native node module (#{react_native_module_path}) - did you install `npm install react-native`?"
 end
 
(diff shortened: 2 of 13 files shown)


RealmJS.podspec
```diff
@@ -0,0 +1,97 @@
...
+
+app_path = File.expand_path('../..', __dir__)
+node_modules_path = File.expand_path('node_modules', app_path)
+react_native_module_path = File.expand_path('react-native', node_modules_path)
+if not File.directory?(react_native_module_path)
+  raise "Couldn't find React Native node module (#{react_native_module_path}) - did you install `npm install react-native`?"
+end
+
+Pod::Spec.new do |s|
+  s.name                   = "RealmJS"
```
kneth: Why not `Realm`?

RealmJS.podspec
```diff
@@ -0,0 +1,97 @@
...
+
+  s.source                 = { :git => git_path, :tag => "v#{s.version}", :submodules => true }
+
+  s.prepare_command        = <<-CMD
+    # Run `npm ci` to install the dependencies the download script needs, if not installed as a node_module
+    ([ -d '../../node_modules' ] || npm ci) && node ./scripts/download-realm.js ios --sync;
+    # Clean up from any previous builds
+    rm -rf include
+    # Build up the include directory
+    mkdir -p include include/jsc include/impl/apple include/util/apple include/sync/impl/apple
```
kneth: You can skip `include` - `-p` will create it.

RealmJS.podspec
```diff
@@ -0,0 +1,97 @@
...
+                             'src/object-store/src/util/apple/*.cpp',
+                             'react-native/ios/RealmReact/*.mm',
+                             'vendor/*.cpp'
+  
+  s.frameworks             = 'Security', 'JavaScriptCore'
+  s.library                = 'c++', 'z'
+  s.compiler_flags         = '-DREALM_HAVE_CONFIG -DREALM_ENABLE_SYNC'
+  s.header_mappings_dir    = 'include'
+  s.preserve_paths         = 'include', 'scripts'
+  s.pod_target_xcconfig    = { 'CC' => '$(PODS_TARGET_SRCROOT)/scripts/ccache-clang.sh',
```
kneth: I don't think we should require users to install ccache.

RealmJS.podspec
kraenhansen: This is primarily because this name ends up being the product name in the xcode project, which is referenced here https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmAnalytics.mm#L69. Choosing this name I didn't have to publish an update to RealmAnalytics.mm on GitHub.

RealmJS.podspec
kraenhansen: Yeah - makes sense :)

RealmJS.podspec
kraenhansen: It doesn't the https://github.com/realm/realm-js/blob/master/scripts/ccache-clang.sh bash scripts checks for the existence of ccache and falls back to clang (or clang++).

RealmJS.podspec
tgoyne: Realm is also already taken by the obj-c Pod.

RealmJS.podspec
```diff
@@ -0,0 +1,97 @@
+require 'json'
+
+package = JSON.parse(File.read(File.expand_path('package.json', __dir__)))
+
+# Building the podspec from a local .git repository when developing - GitHub otherwise
+git_path_local = File.expand_path('.git', __dir__)
+git_path = File.directory?(git_path_local) ? "file://#{git_path_local}" : 'https://github.com/realm/realm-js.git'
```
tgoyne: I don't think this does anything? The Spec source path is used to find out where to get the code when installing from the repository. Unless RN is doing something very strange it's not applicable to local installs at all, and normal usage of Cocoapods doesn't require changing it for local testing of pods.

kraenhansen: I realized this way too late in the process. I had been testing the pod with `pod spec lint`, but as React Native loads in the pod from its local path, I just need the podspec to pass `pod lib lint`. I've pushed a change to simplify this.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added a CocoaPod Podspec file, enabling [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) for React Native on iOS. ([#2586](https://github.com/realm/realm-js/pull/2586)).
```
```suggestion
* Added a CocoaPod Podspec file, enabling [autolinking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) for React Native on iOS. ([#2586](https://github.com/realm/realm-js/pull/2586))
```

PR #2612, opened 2019-11-21
Call ObjectStore::table_for_object_type() in outer loop

### LINKED ISSUE #2610: Optimize setup_aliases() and how it is used
One observation from profiling some application benchmarks is that setup_aliases() is taking a significant time. This is a surprising, and we should optimize the function.

kneth:
## What, How & Why?

A fall-out of a profiling of the data adapters benchmark, we observed that `setup_aliases()` and in particular `ObjectStore::table_for_object_type()` took much CPU time. This is a trivial optimization by moving the call from the inner to the outer loop.

This closes #2610 (https://jira.mongodb.org/browse/RJS-340)

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Internal
 * Upgraded the React Native integration tests app (now using RN v0.61.3). ([#2603](https://github.com/realm/realm-js/pull/2603))
+* Optimized some utility functions to improve performance. ([RJS-350](https://jira.mongodb.org/browse/RJS-340))
 
 3.4.2 Release notes (2019-11-14)
 =============================================================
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Internal
 * Upgraded the React Native integration tests app (now using RN v0.61.3). ([#2603](https://github.com/realm/realm-js/pull/2603))
+* Optimized some utility functions to improve performance. ([RJS-350](https://jira.mongodb.org/browse/RJS-340))
```
bmunkholm: Doesn't the performance affect any users? Move to public section then?

kneth:
Closing as I believe it will degrade the performance in the general case.

kneth:
@bmunkholm I didn't measure but all queries without links will be slower since there is a call to `ObjectStore::table_for_object_type()`. This call is expensive, and I need to find a way to minimize the number of calls to it (and avoid it if possible).

PR #2618, opened 2019-11-25
Call ObjectStore::table_for_object_type() only when needed

kneth:
## What, How & Why?

`ObjectStore::table_for_object_type()` is an expensive call, and we should minimize the number of calls.

This closes https://jira.mongodb.org/browse/RJS-340

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~

### DIFF
--- a/src/js_results.hpp
+++ b/src/js_results.hpp
@@ -160,17 +160,22 @@ inline void setup_aliases(parser::KeyPathMapping &mapping, const realm::SharedRe
 {
     const realm::Schema &schema = realm->schema();
     for (auto it = schema.begin(); it != schema.end(); ++it) {
+        TableRef table;
         for (const Property &property : it->computed_properties) {
             if (property.type == realm::PropertyType::LinkingObjects) {
+                if (!table) {
+                    table = ObjectStore::table_for_object_type(realm->read_group(), it->name);
+                }
                 auto target_object_schema = schema.find(property.object_type);
-                const TableRef table = ObjectStore::table_for_object_type(realm->read_group(), it->name);
                 std::string native_name = "@links." + target_object_schema->name + "." + property.link_origin_property_name;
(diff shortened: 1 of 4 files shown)


src/js_results.hpp
```diff
@@ -160,17 +160,22 @@ inline void setup_aliases(parser::KeyPathMapping &mapping, const realm::SharedRe
 {
     const realm::Schema &schema = realm->schema();
     for (auto it = schema.begin(); it != schema.end(); ++it) {
+        TableRef table;
```
bmunkholm: Init to null.

kneth: I don't think it is needed since `TableRef`'s default constructor is setting it to `null`.

src/js_results.hpp
```diff
@@ -160,17 +160,22 @@ inline void setup_aliases(parser::KeyPathMapping &mapping, const realm::SharedRe
...
         for (const Property &property : it->computed_properties) {
             if (property.type == realm::PropertyType::LinkingObjects) {
+                if (!table) {
```
cmelchior: Any reason this is still inside the loop of properties and not just defined where you define the `table` variable? It doesn't seem to change for each property?

kneth: `ObjectStore::table_for_object_type()` is an expensive call, and I only want to do it when needed (linking object property or a persisted property. The idea is to call it at most one time.

cmelchior: But we don't support schemas with no properties, right? So having the code twice just seems to reduce readability for no real benefit?

kneth: I see your point. Even in the case where a schema doesn't have any properties, we could do the call and avoid the branching. I mean, it is likely to be very rare that a schema doesn't have a property.

PR #2622, opened 2019-11-27
Update to Realm Sync v4.9.0

kneth:
## What, How & Why?
Realm Object Server and Realm Cloud will soon be upgraded, and we need to be on par wrt sync protocol.

This closes https://jira.mongodb.org/browse/RJS-350

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


CHANGELOG.md
```diff
@@ -17,6 +19,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added a test to verify that `Realm.close()` is idempotent.
 * Upgraded the React Native integration tests app (now using RN v0.61.3). ([#2603](https://github.com/realm/realm-js/pull/2603) & [#2609](https://github.com/realm/realm-js/pull/2609))
 * Upgraded Realm Object Store to commit be469eedfb573024839fd61d69e92933b9c1fc9e. ([RJS-349](https://jira.mongodb.org/browse/RJS-349))
+* Upgraded to Realm Sync from v4.8.3 to v4.9.0. ([RJS-350](https://jira.mongodb.org/browse/RJS-350))
```
kspangsege: Nitpicking: Remove `to` after `Upgraded` :-)

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
+NOTE: The synchronization protocol version has been bumped from 29 to 30. The server side remains compatible with protocol versions back to, and including 22. The client side remains compatible with protocol versions back to, and including 26 (Realm JavaScript v2.27.0).
```
bmunkholm: It's not clear to me what the user will gain with this info?
We have not communicated protocol versions in the past. There is no change in compatibility either.
So I would remove this.

What could be relevant to note perhaps is that we provide a better error message (that needs to be implemented in OS likely)? So this is not enough to gain anything from the upgrade.

kneth: We did mention protocol update in v2.27.0.

True, in order to get a better error message, we need to do something in object store.

bmunkholm:
@kneth Just to be clear, there is no urgency in getting the SDK's released.

CHANGELOG.md
kspangsege: But there is also very little risk. To the extent that there is significant risk in Sync 4.9.0, it is all on the server side.

I see no point in holding it back.

cc @geragray

bmunkholm: That's a mistake in 2.27.0. We shouldn't really do that as it's an implementation detail to us. What we need to tell users is which SDK versions are compatible with which ROS versions. Including an indirection (protocol version) is not helpful. We talked about this before. So we should only update the Docs that reflects this and is referenced in the release notes.

kneth: I'll remove it from the change log.

bmunkholm:
I assume we should mention what this improves?
Like "Improved error message in case of synchronization errors." - perhaps that could be more specific, but it shouldn't be too detailed.

kneth:
@bmunkholm:

> I assume we should mention what this improves?

Maybe something like:

"A new error state is added. When the state is entered, it is often when a client attempt to connect to a server after a long period of being offline."

In lack of better ideas, I suggest that we add it to the "Internal" section.

PR #2623, opened 2019-11-28
React Native iOS: Support comitting `ios/Pods` to a users git repository

### LINKED ISSUE #2617: build failed:  'realm/util/assert.hpp' file not found
## Goals
I want to build project successfully

## Expected Results

build success
## Actual Results
build failed:  'realm/util/assert.hpp' file not found

image

## Steps to Reproduce
npm install --save realm
pod install
build

- Realm JS SDK Version:  3.4.2
- Node or React Native:  Node 10.17.0 ,  React Native 0.60.5
- Client OS & Version:  iOS
- Which debugger for React Native: ?/None

kraenhansen:
## What, How & Why?

It seems there is no consensus in the CocoaPods community on whether or not to add the `Pods` directory (created when running `pod install`) to the git repository of a project: https://stackoverflow.com/questions/9446644/what-goes-into-your-gitignore-if-youre-using-cocoapods.

One advantage of adding it is that it simplifies the tools needed to build the app: I.e. CI servers and co-workers don't need to have the CocoaPod CLI installed. You can simply clone the repository, run `npm install` and build the project (although possibly running into issues as the JS code of Realm JS might not match the artifacts in the `ios/Pods` directory).

This PR changes the `RealmJS.podspec` to be more resilient towards this usage pattern, by: 
1. Using a [script_phase](https://guides.cocoapods.org/syntax/podspec.html#script_phases) in addition to the [prepare_command](https://guides.cocoapods.org/syntax/podspec.html#prepare_command).
2. Referencing `Headers/Public/React-Core/` relative to the `$(PODS_ROOT)` environment instead of producing an absolute path, which would otherwise end up in `ios/Pods/Target Support Files/RealmJS/RealmJS.xcconfig`.
3. Hiding the "RealmJS thinks the Podfile is (not) calling use_frameworks!" behind a `DEBUG_REALM_JS_PODSPEC` environment variable, to limit the confusion that this has caused.

This closes #2617 and it will most probably also fix an issue where `yarn add [another package]` resets the Realm package directory.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests


CONTRIBUTING.md
```diff
@@ -176,4 +176,7 @@ Note that it isn't possible to easily run a single unit test from Xcode. Instead
...
 
-The Podspec file expects to be located in the `./node_modules/realm/` folder of a React Native project with a `./ios/Podfile` file. The easiest way to verify that changes to the Podspec is not breaking installing the library, install Realm JS in a new React Native app, make the changes to the `./node_modules/realm/RealmJS.podspec` and run `pod lib lint --verbose` from the `./node_modules/realm/` directory. Consider adding `--no-clean` to prevent the CocoaPods CLI from deleting the temporary project created during linting. Once linting passes, try installing the pod and running the consuming app on an iOS device.
+The Podspec file expects to be located in the `./node_modules/realm/` folder of a React Native project with a `./ios/Podfile` file. The easiest way to verify that changes to the Podspec is not breaking installing the library, install Realm JS in a new React Native app, make the changes to the `./node_modules/realm/RealmJS.podspec` and build the app.
```
kneth: Should we add something like this to the documentation at https://realm.io?

kraenhansen: I thought about it, but I think this will be the case for 99% of all installments. It might be a little too detailed for the vast majority.

PR #2653, opened 2019-12-13
Core 6: Notifications update

ironage:
Upgrade to the object store PR https://github.com/realm/realm-object-store/pull/880, when merged, we should use the core-6 branch of object store again.

- ~~Removes the "oldModifications" and "newModifications" properties on the global notifier events, this might be a breaking change, depending on if these were documented or not?~~
- There is no guarantee about the ordering of keys reported by the global notifier. This was probably implied before, but now the keys are not reported in "queue" order any longer.
- Fixed behaviour of reported notifications on a clear/deleteAll event which was mostly working by accident in the previous core 6 implementation.
- Should theoretically speed up the global notifier reporting mechanism, but this is not measured.
- Notification changes now vend proper Realm Objects, instead of our internal object key representation which users couldn't do anything with.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


tgoyne:
The typescript definitions need to be updated to reflect the removal of newModifications/oldModifications. The multi-process notifier stuff also does things with those properties; at the minimum they need to be removed from the test support code (tests/js/support/notification-worker.ts).

kneth:
It is a breaking change. And we need to update the [API doc](https://realm.io/docs/javascript/latest/api/Realm.Results.html#addListener) too.

ironage:
I took a stab at the docs updates. Looks like this was a breaking change from the typescript definitions point of view, but not from the docs point of view?

I also just realized that I'm not sure if people can actually do anything with the keys reported in this way... is there anything in js that translates a core stable id into a user object primary key? Or should we be doing this translation for them when reporting keys?

ironage:
@kneth @tgoyne I don't see any way the raw object ids can be used through the public API. How about we report the changes as a collection of objects; inserted and modified as live into the new realm, and deleted into the old realm? I think the translation from object-id to object can happen entirely at the js binding level.

tgoyne:
We probably want a way to look up objects by key and get the key from an object in the public API, but even if we do add that changing the GN API to vend objects sounds like a good idea to me.

kneth:
The JS api has a private method to retrieve an object using the object id. I have been waiting for a final solution on the object id to surface (to align with MongoDB's object ids).

ironage:
Updated the global notifier to pass out objects (there's still one failing test to figure out). This depends on getting a new core released with https://github.com/realm/realm-core/pull/3528

tests/js/realm-tests.js
```diff
@@ -1529,10 +1534,15 @@ module.exports = {
     },
 
     testDisableFileFormatUpgrade: function() {
+        let config = { path: 'realm-bundle.realm' };
+        if (Realm.exists(config)) {
+            Realm.deleteFile(config);
+        }
         Realm.copyBundledRealmFiles();
+        TestCase.assertTrue(Realm.exists(config)))
```
kneth:
```suggestion
        TestCase.assertTrue(Realm.exists(config));
```

PR #2657, opened 2019-12-17
Update CHANGELOG.md

bmunkholm:
@kneth will you fill in the js versions that the beta aligns with?
And fix the dates of those beta releases? Could we please use a format that no one is in doubt about? Like "2019-Dec-24"


kneth:
I have incorporated this change into the `core-6` branch.

PR #2661, opened 2019-12-18
Add Realm.Object.globalKey() and Realm.objectForGlobalKey()

kneth:
## What, How & Why?

Core v6 swaps out the old row index with an object key. But in order to avoid confusion with MongoDB's object id, we can call Realm's object keys for global key.

This closes [RJS-383](https://jira.mongodb.org/browse/RJS-383)

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/object.js
+++ b/docs/object.js
@@ -91,10 +91,10 @@ class Object {
     removeAllListeners() { }
 
     /**
-     * Returns a unique identifier. For synced objects, the uniqueness is only guaranteed when the object has been
+     * Returns a unique and global key. For synced objects, the uniqueness is only guaranteed when the object has been
      * synchronized.
-     * @see {@link Realm#objectForObjectId objectForObjectId}
+     * @see {@link Realm#objectForGlobalKey objectForGlobalKey}
      * @since 4.0.0
      */
-    objectId() { }
+    globalKey() { }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ NOTE: Contains all previous releases up to v3.6.0.
 * None.
 
 ### Enhancements
-* None.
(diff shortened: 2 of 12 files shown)


docs/object.js
```diff
@@ -91,10 +91,10 @@ class Object {
...
     /**
-     * Returns a unique identifier. For synced objects, the uniqueness is only guaranteed when the object has been
+     * Returns a unique and global key. For synced objects, the uniqueness is only guaranteed when the object has been
      * synchronized.
-     * @see {@link Realm#objectForObjectId objectForObjectId}
+     * @see {@link Realm#objectForGlobalKey objectForGlobalKey}
      * @since 4.0.0
      */
-    objectId() { }
+    globalKey() { }
```
cmelchior: Maybe it is worth thinking a bit about the naming of this. See https://mongodb.slack.com/archives/CKW2HG3M0/p1576660133033600

Especially when it comes to communicating it to the outside.
From my understanding, `GlobalKey` is the sync variant that isn't used by non-synced Realms. In those cases, only `ObjKey` exists.

I agree that that using `objectId` should probably be avoided no matter what to avoid conflict with MongoDB and maybe `GlobalKey` is the word we want to expose to the outside. Or maybe we want to come up with another name altogether in order to not cause internal confusion as well.

- ObjectId (taken by Mongo)... Would have been much more accurate if it used `DocumentId` 😄 
- RealmId (new name returning either internal GlobalKey/ObjKey depending on context).
- GlobalKey (overloading the internal term GlobalKey as it sometimes will return a ObjKey)`
- ObjectKey (new name, although very close to `ObjKey` and `ObjectId`)
- Maybe something else?

kneth:
This is obviously too early to introduce. Let's get the new object id type into core first and maybe converge the many object ids we have :-)

PR #2667, opened 2019-12-20
Adding integration tests for iterators

kraenhansen:
## What, How & Why?

This adds integration tests verifying the correctness of various ways of iterating over `Results` (unfiltered & filtered), `lists` and `linkingObjects`.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/iterators.ts
+++ b/integration-tests/tests/src/iterators.ts
@@ -1,0 +1,198 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2019 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


integration-tests/tests/src/iterators.ts
```diff
@@ -0,0 +1,198 @@
...
+            realm.objects("SomeOtherClass");
+        }).throws("Object type 'SomeOtherClass' not found in schema.");
+    });
+
+    type CollectionCallback = () => Realm.Collection<any>;
+
+    function itCanIterate(
+        getCollection: CollectionCallback,
+        expectedNames: string[]
+    ) {
```
blagoev: function itCanIterate(
        getCollection: CollectionCallback,
        expectedNames: string[]) 
{

little and tiny code formatting note :)

kraenhansen: Not with the current `tslint` rules :)

blagoev: that sucks that our rules require us to do that.

kraenhansen: Personally I don't really care either way, as long as we have a standard that I can use automated tools to format my code into. Our [tslint.json](https://github.com/realm/realm-js/blob/master/integration-tests/tests/tslint.json) mostly use the recommended / latest options, in an effort to make us less special compared to other code-bases on the interwebs.

PR #2686, opened 2020-01-15
Use the node versions as specified (Linux)

kneth:


Jenkinsfile
```diff
@@ -110,9 +110,9 @@ stage('test') {
   for (def nodeVersion in nodeVersions) {
     parallelExecutors["macOS node ${nodeVersion} Debug"]   = testMacOS("node Debug ${nodeVersion}")
```
kraenhansen: Could / should the `testMacOS` have the same arguments as `testLinux`?

scripts/test.sh
```diff
@@ -353,9 +353,20 @@ case "$TARGET" in
 
   check_test_results ReactTests
   ;;
+
+"node-test-runners")
```
kraenhansen: Should this be "test-runners"?
Is this different from https://github.com/realm/realm-js/blob/master/scripts/test.sh#L397-L401?

kneth: I'll remove it and adjust `Jenkinsfile`.

kneth:
I am merging this after manually inspecting that the specified node version is actually used on Linux.

PR #2712, opened 2020-02-04
Enable building with Visual Studio 2019

fealebenpae:
This fixes some of the hardcoded gyp configuration that locked the build to Visual Studio 2017:
 - remove the hardcoded path to the Windows 8.1 SDK, and instead link against whatever version of the SDK node-gyp found in the Visual Studio installation
 - remove the hardcoded toolset version so gyp can build with newer versions of the MSVC compiler
 - fix the `_WIN32_WINNT` macro define to actually select Windows 8.1 as the minimum supported API level (previously it was Vista).

### DIFF
--- a/realm.gypi
+++ b/realm.gypi
@@ -204,13 +204,7 @@
           "libraries": [ "-lrealm-sync<(debug_library_suffix)" ],
           "conditions": [
               ["OS=='win'", {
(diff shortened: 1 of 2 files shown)


kneth:
@fealebenpae Thank you

Issue #2655, opened 2019-12-13
Change listeners on Results should always update React Native UI


kraenhansen:
## Goals

A callback registered as a listener for changes on a Realm JS `Results` instance should always be able to trigger updates of the React Native UI.

## Expected Results

From a React Native app, when registering a callback on a `Results` object and calling `this.setState` on a component, I would expect the UI to always update.

## Actual Results

- The UI only updates 4 out of 5 times when the change listener fires.
- The bug seems to disappear if `this.setState` is the first method called in the callback.
- It's possible to trigger an update of the UI by touching / clicking the simulator.

## Steps to Reproduce & Code Sample

Initialize a new React Native app:

    npx react-native init Issue2655 --directory realm-js-issue-2655

Copy in the files below, install the dependencies

    cd realm-js-issue-2655
    npm install

Create an instance of ROS and update `constants.js`.

In one terminal, start the updater (a node process changing a realm every second) - in another start the app on iOS

    npm run updater
    npm run ios

The app has two modes: "interval" where a timer will update the state every second and "realm" where a change listener will be registered and the bug is observed.

Use [the Safari developer tools](https://facebook.github.io/react-native/docs/debugging#safari-developer-tools) to attach to the apps JSContext and observe the issue in the timeline:

// TODO: Add images of the timeline and call-stacks when running in the two modes.

### package.json
```javascript
{
  "name": "realm-js-issue-2655",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "react-native run-android",
    "ios": "react-native run-ios",
    "start": "react-native start",
    "lint": "eslint .",
    "updater": "cd updater && node index.js"
  },
  "dependencies": {
    "react": "16.9.0",
    "react-native": "0.61.5",
    "realm": "^3.5.0"
  },
  "devDependencies": {
    "@babel/core": "^7.6.2",
    "@babel/runtime": "^7.6.2",
    "@react-native-community/eslint-config": "^0.0.5",
    "eslint": "^6.5.1",
    "metro-react-native-babel-preset": "^0.56.0"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

### App.js

```javascript
import React, {Component} from 'react';
import {Button, Text, View} from 'react-native';
import Realm from 'realm';

import {schema} from './schema';
import {NICKNAME, SERVER_URL} from './constants';

const SEQUENCE = 'abcdefghijklmn';

const styles = {
  mainView: {
    justifyContent: 'center',
[...]

PR #2733, opened 2020-02-21
Adding a failing test showing that Results listeners are called synchronously

kraenhansen:
## What, How & Why?
I have a theory that #2655 could be fixed by having change listener callbacks (any listener callbacks really) added to a task queue and picked up by the event loop instead of being called synchronously.

This PR adds a failing integration test verifying that change listeners are indeed called synchronously.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/index.ts
+++ b/integration-tests/tests/src/index.ts
@@ -16,24 +16,20 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-if (!global.Realm) {
-    throw new Error("Expected 'Realm' to be available as a global");
-}
+const EXPECTED_GLOBALS = {
+    Realm: "function",
+    title: "string",
+    fs: "object",
+    path: "object",
+    environment: "object",
+    setTimeout: "function",
+    clearTimeout: "function"
+};
 
-if (!global.title) {
-    throw new Error("Expected 'title' to be available as a global");
-}
-
-if (!global.fs) {
-    throw new Error("Expected 'fs' to be available as a global");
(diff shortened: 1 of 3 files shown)


tgoyne:
Calling change notification callbacks asynchronously would break anything which actually uses the change indices in the notification. A simple single-threaded example:

```
let results = realm.objects('Person');
await new Promise(resolve => {
    let first = false;
    results.addListener((_, changes) => {
        if (first) {
            first = false;;
            resolve();
            return;
        }

        for (let i of changes.insertions) {
            console.log(results[i]);
        }
    }
});

realm.write(() => realm.create('Person', {...}; );
realm.write(() => realm.delete(results[0]); };
```

Here the second call to `realm.write()` will synchronously deliver the notification about the insertion prior to invoking the write callback, resulting in the listener being able to log the newly inserted object. If the notification listener is instead invoked asynchronously, the newly inserted object will no longer exist by the time it is called, and it'll log `undefined`.

The JS environment being single-threaded eliminates a lot of the complicated cases that make synchronous delivery required for non-sync Realms, but the fact that sync makes writes on a background thread reintroduces all of them for sync Realms.

If the problem is that the listeners aren't being invoked in the context of a JS task queue, then EventLoopSignal should get a RN version that integrates with that rather than the lower-level CF runloop.

kraenhansen:
This was testing the v11 SDK and I don't see the need to rebase nor merge, given Thomas' comment above.

PR #2742, opened 2020-03-01
MongoDB Realm integration

kneth:
## What, How & Why?

This closes the node.js/React Native part of #2691.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* ~[ ] typescript definitions file is updated~ (see #2741)
* [ ] jsdoc files updated


packages/realm-network-transport/src/NetworkTransport.ts
```diff
@@ -0,0 +1,189 @@
...
+                    responseHeaders[key] = value;
+                });
+                return {
+                    statusCode: response.status,
+                    headers: responseHeaders,
+                    body: decodedBody
+                };
+            })
+            .then(r => handler.onSuccess(r))
+            .catch(e => handler.onError(e));
```
kraenhansen: If the two methods are bound to the handler, this could also be
```
.then(handler.onSuccess)
.catch(handler.onError);
```

PR #2753, opened 2020-03-13
Realm App importer

kraenhansen:
## What, How & Why?

This implements a CLI around the Stitch CLI to help work around a few of its shortcomings, most notably:
- It authenticates, creates an empty app, creates secrets (via HTTP) and then performs an import using the Stitch CLI.
- No need to authenticate and fetch the users group / project id.
- It implements three ways of getting the id of the newly created app (which is needed when communicating with it):
  1. the consuming test harness can use the package programically, instantiating the `RealmAppImporter` class and calling its `importApp` method, which returns a `Promise<{ appId: string }>`.
  2. the `--app-id-path` runtime option saves the app id to a file, which can be read by the test harness.
  3. the `--app-id-port` runtime option starts up a web-server on the specified port and serves the app id as a text response.

## Installing and running it

To try out the package, initialize an NPM package:

    npm init

and add the package from S3 as a dev-dependency:

    npm install https://static.realm.io/downloads/realm-app-importer/realm-app-importer-0.1.0.tgz --save-dev

To use it, call it directly via `node_modules/.bin/realm-app-importer` or  even better, just run

    npx realm-app-importer ./my-app-template

To ease the testing, I've included a two templates, which can be imported with

    npx realm-app-importer node_modules/realm-app-importer/templates/simple
    npx realm-app-importer node_modules/realm-app-importer/templates/with-secret

I imagine adding other templates here that we can share across the SDKs if we ever converge on one or more apps that we agree to test against.

To import secrets add a secrets.json to the template directory, containing a single JSON object with keys and string values:

```json
{
    "my-secret": "v3ry-s3cr3t"
}
```

Please note that I'm also adding [a README.md to the package repository](https://github.com/realm/realm-js/blob/kh/realm-app-importer/packages/realm-app-importer/README.md), containing more details.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually, locally)


packages/realm-app-importer/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "tslint": "../../node_modules/.bin/tslint -p .",
+      "prepack": "rm -rf dist && npm run build",
+      "start": "ts-node src/cli.ts"
+    },
+    "files": [
+      "dist",
+      "templates",
+      "realm-app-importer"
+    ],
+    "author": {
```
kneth: Please add `contributors` with your name in the array :smile:

packages/realm-app-importer/src/RealmAppImporter.ts
```diff
@@ -0,0 +1,246 @@
...
+        const body = JSON.stringify({ name, value });
+        const response = await fetch(url, {
+            method: "POST",
+            headers: {
+                Authorization: `Bearer ${this.accessToken}`,
+                "content-type": "application/json"
+            },
+            body
+        });
+        if (response.ok) {
```
kneth: `return response.ok;` maybe

packages/realm-app-importer/package.json
kraenhansen: 👍

packages/realm-app-importer/src/RealmAppImporter.ts
kraenhansen: I changed this to throw if the response is not OK.

Issue #2699, opened 2020-01-23
Merge Stitch and Realm SDKs - Realm Web - Phase 1


RealmBot:
## Tasks
* [x] Setup multi package repository.
* [x] Implement first draft of TypeScript interfaces shared across Realm JS and Realm Web.
* [x] Implement a shared network transport between Realm JS and Realm Web (realm-network-transport).
* [x] Implement basics of authentication (anon + user-pass credentials, authenticated network transport).
* [x] Setup the "realm-web-integration-tests" package using WebPack, Puppeteer & Mocha Remote.
* [x] Implement basics of functions (call a function with an authenticated user)
* [x] Use eslint instead of the deprecated tslint.
* [x] Implement a couple of credentials for authentication providers:
  * [x] Anonymous
  * [x] API Key
  * [x] Email & password (formerly known as username & password)
* [x] Implement standard service clients on top of functions:
  * [x] Remote MongoDB
  * [x] HTTP

Below tasks has been moved to move to "phase 2":

* [ ] Implement auth-provider clients:
  * [ ] Email & password (reset password, confirm email, etc.) _[implemented, but missing passing e2e tests]_
  * [ ] API keys (enable, disable, list, get, create, etc.) _[implemented, but missing passing e2e tests]_
* [ ] Implement more credentials for authentication providers:
  * [ ] Facebook
  * [ ] Google
  * [ ] Custom JWT
  * [ ] Apple Sign In
  * [ ] Functions
* [ ] Linking a current user with another identity
* [ ] Retrieve a users custom data _[implemented, but missing passing e2e tests]_
* [ ] Ensure xjson is used to serialize requests and parse responses where appropriate
* [ ] Persist logged in users in the browsers local storage.
* [ ] Revisit and align Remote MongoDB API with the Node.js driver.
* [ ] Upgrade from Mocha 5 to latest (and use "modules": "es6" in tsconfig.jsons)
* [ ] Add the the missing credentials types to app.d.ts.
* [ ] Implement fetching the location (baseURL) of where an app is running in the Stitch production environment.
* [ ] Send device information on authentication requests.

Details of above subtasks are described in this [doc](https://docs.google.com/document/d/1iD9ujY76BWVUOdN7rRH4XyPF9hHxx48RcGc-ZaQPyII/edit#heading=h.5zfqez3rx85o)

RealmBot:
➤ Kræn Hansen commented:

[...]

PR #2762, opened 2020-03-20
Realm Web: MongoDB & HTTP services

kraenhansen:
## What, How & Why?

This PR is adding remote MongoDB and HTTP services to Realm Web, closing more of #2699.
Please note: This has already been released internally as Realm Web 0.1.0 and for the tasks not checked below, I suggest we postpone this to a separate PR, once the API has stabilized.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen:
> looks alright. The code formatting is highlighting we should definitely increase the line length either in this or in a later PR.
> A notable example is this one
> 
> ```
> type NewDocument<T extends Document> = Realm.Services.RemoteMongoDB.NewDocument<
>     T
> >;
> ```

I agree - although you'll most likely be able to find a similar example once it's increased.

blagoev:
or maybe the formatter in this case is missing some settings. I can't say for sure. Since why not split the lines like this
```
type NewDocument<T extends Document> = 
      Realm.Services.RemoteMongoDB.NewDocument<T>;
```

packages/realm-web/src/services/RemoteMongoDBService.ts
```diff
@@ -0,0 +1,294 @@
...
+    return new RemoteMongoDBCollection<T>(
+        transport,
+        serviceName,
+        databaseName,
+        collectionName,
+    );
+}
+
+/**
+ * Creates an Remote MongoDB Database
```
kneth, line 261:
```suggestion
 * Creates a Remote MongoDB Database
```

packages/realm-web/src/services/RemoteMongoDBService.ts
```diff
@@ -0,0 +1,294 @@
...
+            null,
+            transport,
+            serviceName,
+            databaseName,
+        ),
+    };
+}
+
+/**
+ * Creates an Remote MongoDB Service
```
kneth, line 285:
```suggestion
 * Creates a Remote MongoDB Service
```

lib/app.d.ts
```diff
@@ -124,10 +129,15 @@ declare namespace Realm {
         readonly id: string;
 
         /**
-         * Use this to call functions defined on the MongoDB Realm server.
+         * Use this to call functions within by the MongoDB Realm app.
```
kraenhansen, line 132:
```suggestion
         * Use this to call functions defined by the MongoDB Realm app.
```

Issue #2778, opened 2020-03-30
[v5] Can't use realmObject.keys() with TypeScript


sintylapse:
## Goals
Using `realmObject.keys()` causes TypeScript error `Property 'keys' does not exist on type 'Object'.`. It might be related to https://eslint.org/docs/rules/no-prototype-builtins so we could use keys like this 
```js
Realm.Object.prototype.keys.call(realmObject)
```
but the same error appears in ts `Property 'keys' does not exist on type 'Object'.`

It seems new methods should be added to prototype definition

## Expected Results
No ts errors

## Actual Results
`Property 'keys' does not exist on type 'Object'.`

## Version of Realm and Tooling
- Realm JS SDK Version: 5.0.2
- Node or React Native: 61.5

blagoev:
@sintylapse It should be fixed in Realm JS 5.0.3. We added TypeScript definitions for these methods

sintylapse:
@blagoev yes it works now, thanks

PR #2786, opened 2020-04-01
update ts definition of the Realm Object

blagoev:
fixes https://github.com/realm/realm-js/issues/2778

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -138,6 +138,16 @@ declare namespace Realm {
      */
     interface Object {
         /**
+         * @returns returns an array of a given Realm Object property names
+         */
+        keys(): string[];
+
+        /**
+         * @returns Returns an array of key/value pairs of the given Realm Object properties
+         */
+        entries(): [string, any][];
+
+        /**
          * @returns boolean
          */
         isValid(): boolean;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -143,7 +143,7 @@ declare namespace Realm {
         keys(): string[];
 
         /**
-         * @returns Returns an array of key/value pairs of the given Realm Object properties
+         * @returns Returns an array of key/value pairs of a given Realm Object properties
          */
         entries(): [string, any][];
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -137,6 +137,16 @@ declare namespace Realm {
      * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Object.html }
      */
     interface Object {
+        /**
+         * @returns returns an array of a given Realm Object property names
```
kraenhansen, line 141:
```suggestion
         * @returns An array of the names of the object's properties.
```

lib/index.d.ts
```diff
@@ -137,6 +137,16 @@ declare namespace Realm {
...
+         */
+        keys(): string[];
+
+        /**
+         * @returns Returns an array of key/value pairs of the given Realm Object properties
```
kraenhansen:
```suggestion
         * @returns An array of key/value pairs of the object's properties.
```

blagoev:
@kneth I didn't notice Kraen already did a review. Consider ignoring

lib/index.d.ts
blagoev: I tried to mimic this one. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys
Also from the TypeScript definitions 
/**
     * Returns the names of the enumerable string properties and methods of an object.
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    keys(o: object): string[];

I think these are an example of good explanations but I can change it to your suggestion if you still think it's better

kraenhansen: My main objection is the double "@returns returns" and since its an instance method on a `Realm.Object` I think it is better to write "the object" instead of "a given Realm Object", because - it's not given by the user as with `Object.keys()`

blagoev: good point.

Issue #2763, opened 2020-03-20
[5.0.0] Android returns Results with length 1 but no keys or values


bimusiek:
## Goals
I am trying to fetch some objects from database.

## Expected Results
Getting 1 object with all properties.

## Actual Results
I get Results of length 1 but when getting the object via index 0, it returns no properties.

## Steps to Reproduce
Run the code on Android

## Code Sample
```
const results = realm
        .objects<ImporterAuthenticationModel>(ImporterAuthenticationModel.schema.name)
        .filtered('importerId = $0', importer.importer.id);

      if (results) {
        console.log(results.length);
        console.log(results[0].id);
      }
```
Output is:
```
1
undefined
```

Schema:
```
export class ImporterAuthenticationModel extends BaseModel {
  public static schema = {
    ...BaseModel.schema,
    name: 'ImporterAuthenticationModel',
    primaryKey: 'uniqueId',
    properties: {
      ...BaseModel.schema.properties,
      uniqueId: 'string',
      id: 'string',
      title: 'string',
      subTitle: 'string?',
      imageUrl: 'string?',
      importerId: 'string',
      additionalDataString: 'string?',
      rawStatus: { type: 'int', default: ImporterAuthenticationStatus.AUTHENTICATED },
      expiresAt: 'int'
    }
  };
```
## Version of Realm and Tooling

- Realm JS SDK Version: 5.0.0
- Node or React Native: React Native
- Client OS & Version: Android 9PKQ1.181121.001 (Mi 9 Lite)
- Which debugger for React Native: React Native Debugger (but it also fails if using production build from FireBase)

<img width="511" alt="Screenshot 2020-03-20 17 12 16" src="https://user-images.githubusercontent.com/1025588/77183797-30ef8700-6acf-11ea-99e3-7f2a0787328c.png">

<img width="259" alt="Screenshot 2020-03-20 17 12 22" src="https://user-images.githubusercontent.com/1025588/77183811-35b43b00-6acf-11ea-8f43-e074f90dc1cb.png">
 
<img width="472" alt="Screenshot 2020-03-20 17 13 49" src="https://user-images.githubusercontent.com/1025588/77183864-406ed000-6acf-11ea-9093-c24e968ea579.png">

<img width="354" alt="Screenshot 2020-03-20 17 13 55" src="https://user-images.githubusercontent.com/1025588/77183889-4795de00-6acf-11ea-94c7-28ae332abbd4.png">

bimusiek:
[...]

PR #2793, opened 2020-04-02
Watch-mode for the React Native integration tests

kraenhansen:
## What, How & Why?

When working on #2763 I wrote some code that I think could benefit us when debugging on React Native in the future. Specifically this adds a watch-mode to the React Native integration tests - documented in the React Native environment's README.md:

> To run the tests in watch mode, prepend the `--watch` runtime option when starting the tests:
> 
> On Android
> 
>     npm run test/android -- --watch
> 
> On iOS
> 
>     npm run test/ios -- --watch
> 
> This will keep the harness, metro server and mocha-remote servers running and connected to the device. When hot reloading (from an update to Realm JS, the tests or the app itself) the app will re-connect and rerun the tests.
> 
> While in watch mode, you can rebuild, repackage and reinstall the tests by running
> 
>     npm run update-tests

This also adds a simple test that retrieves an object from the database and access its properties.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests


integration-tests/environments/react-native/harness/runner.js
```diff
@@ -16,162 +16,212 @@
...
-const { timeout, TimeoutError } = require("promise-timeout");
+const {MochaRemoteServer} = require('mocha-remote-server');
+const {timeout, TimeoutError} = require('promise-timeout');
+const yargs = require('yargs');
+const path = require('path');
 
-const rn = require("./react-native-cli");
-const android = require("./android-cli");
-const xcode = require("./xcode-cli");
+const rn = require('./react-native-cli');
```
blagoev, line 24: not to be picky but why we need this change :)

integration-tests/environments/react-native/harness/runner.js
```diff
@@ -16,162 +16,212 @@
...
-        xcode.simctl.boot(deviceId);
-        console.log('Simulator is booting');
-        xcode.simctl.bootstatus(deviceId);
-        console.log("Simulator is ready 🚀");
+  if (platform === 'android') {
+    const devices = android.adb.devices();
+    const activeDevices = devices.filter(({state}) => state === 'device');
+    if (activeDevices.length === 0) {
+      throw new Error(
+        'Missing an active device: Attach a device via USB or start an emulator',
```
blagoev, line 40: possibly move this line up ?

integration-tests/environments/react-native/src/App.js
```diff
@@ -24,132 +24,152 @@
...
-        margin: 10,
-    },
-    details: {
-        fontSize: 14,
-        textAlign: "center",
-        margin: 10,
-    },
+  container: {
+    flex: 1,
+    justifyContent: 'center',
```
blagoev, line 161: is this some tool that is causing this

integration-tests/environments/react-native/harness/runner.js
kraenhansen: With the way the package is currently setup, it is [extending the react native eslint config](https://github.com/realm/realm-js/blob/master/integration-tests/environments/react-native/.eslintrc.js#L3), which [uses single-quotes for strings](https://github.com/facebook/react-native/blob/master/packages/eslint-config-react-native-community/index.js#L261): We could change this to extend from our eslintrc (located in the root of the repository), to keep a consistent style across all our code.

This change was introduced simply because my editor is setup to format on save in compliance with the configuration in the package folder.

integration-tests/environments/react-native/src/App.js
kraenhansen: ESlint and [Prettier](https://prettier.io/) I suggest we make this more explicit and align the tools across, once we have multiple (sub) packages relying on it.

blagoev: The thing is we might not be able to align all tools to work exactly the same on all platforms and environments we have. That's why having automating on save is problematic since it will toggle the formatting depending on which platform the change is done. I think in practice this will not work and will be an inconvenience for us.

integration-tests/environments/react-native/harness/runner.js
```diff
@@ -144,34 +171,60 @@ async function runApp(platform, junitFilePath) {
...
-    if (!platform) {
-        throw new Error("Expected a platform runtime argument");
+    } else {
+        // Run tests with a 5 minute timeout
+        return timeout(
+            new Promise(resolve => {
+                console.log("Running tests 🏃‍");
+                server.run(resolve);
+            }),
+            60000 * 5,
```
blagoev, line 183: does it make sense to have this as constant somewhere on top ?

PR #2794, opened 2020-04-02
Improving eslint and adding prettier

kraenhansen:
## What, How & Why?

This is an attempt at improving the readability and therefore maintainability of our codebase by introducing a single coherent code-style throughout the project.

I suggest that we introduce [Prettier](https://prettier.io/) which I've used on Realm Studio and other codebases with great success. It basically introduces strict rules for indention and wraps code nicely to conform to these rules, when the linter is running in "fix" mode (from a terminal or when a developer with the eslint plugin setup in their IDE formats on save).

I suggest deviate from the defaults by setting print width to 120 (instead of the 80 default) and stick to the default 2 space indention that prettier suggests.

This PR configures eslint (was sort-of already there) and prettier on the project and most of its sub-packages, but it doesn't run the fix (as that would introduce changes to 16064 lines of code).

I suggest we keep this PR as a draft until we have codebase in less flux, rebase this on top of the master branch when that happens, apply the fix (running `npm run lint -- --fix`) and commit the changes across the entire project.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


integration-tests/tests/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "tsc -p .",
     "start": "ts-mocha --project ./tsconfig.json --require src/utils/segfault-handler.js --require src/utils/inject-globals.js --watch-extensions ts,js --watch src/index.ts",
     "prepack": "npm run build",
-    "lint": "tslint -p ."
+    "lint": "../../node_modules/.bin/eslint . --ext .js,.ts"
```
steffenagger, line 11: Could this be handles in `.eslintrc.json` through files, instead of `--ext .js,.ts` here? Then each project defines if linting should be done for additional `jsx`/`tsx` extensions.

.prettierrc.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
```
steffenagger, line 1: I would suggest removing this config, and move it to `.eslint.json`. Then running prettier through ESLint as a plugin. This way we have a single tool for linting/formatting (ESLint formats as well with `--fix`, and if the two have conflicting rules or runs in parallel, "fun" stuff can happen).

It's also easier to setup editors. vscode example:
1) install [ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) 
2) Add the following to `settings.json`:
```json
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  }
```

kraenhansen: I though this was the recommended way of configuring Prettier, even if invoked via ESLint: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration

steffenagger: It is, but afaik it's with the soul focus of letting tools pick up that prettier is "active" in the project, for IDE extensions/tooling. This is also the reason for me to suggest moving it - as running eslint & prettier simultaneously will produce unwanted results (in case you have both prettier & eslint extensions installed).
Might not be an issue.

steffenagger:
120 char with is a bit much for my 13" screen, could we meet at the middle and say 100?

kraenhansen:
This got superseded by https://github.com/realm/realm-js/pull/3850

PR #2808, opened 2020-04-13
remove aws from Jenkins

blagoev:


fealebenpae:
This needs https://github.com/realm/realm-js/pull/2748/files#diff-3254677a7917c6c01f55212f86c57fbf to work around the git issue.

blagoev:
@fealebenpae Jenkins script in master already has it

fealebenpae:
@blagoev I'm seeing the exact same git error in the build logs, however: `fatal: unable to look up current user in the passwd file: no such user`.

blagoev:
this is in master https://github.com/realm/realm-js/blob/master/Jenkinsfile#L21
I don't know what's going on. Is the git version up-to-date. Maybe we can git config in the test.sh?

blagoev:
@fealebenpae I think the culprit is the git version `git version 1.8.3.1`

fealebenpae:
@blagoev I don't think the environment defined in the Jenkinsfile trickles down to Docker containers. I recommend defining those variables directly inside the Dockerfile like I did.

blagoev:
@fealebenpae You meann like this https://github.com/realm/realm-js/blob/master/Dockerfile#L6

fealebenpae:
Ah, so you're right - it is already there. I should have looked, sorry.

blagoev:
no worries. I will try fixing this not to waste your time. Thanks for the help

PR #2838, opened 2020-04-23
use prebuild libraries for openssl.

blagoev:
use prebuild libs for openssl to not build openssl locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,12 +19,12 @@ RUN yum -y install \
  && yum clean all \
  \
   # TODO: install openssl in /usr/local
- && curl -SL https://www.openssl.org/source/openssl-1.1.1b.tar.gz | tar -zxC / \
- && cd /openssl-1.1.1b \
- && ./config --prefix=/usr \
- && make && make install \
- && rm -rf /openssl-1.1.1b \
- && cd /tmp \
+#  && curl -SL https://www.openssl.org/source/openssl-1.1.1b.tar.gz | tar -zxC / \
+#  && cd /openssl-1.1.1b \
+#  && ./config --prefix=/usr \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -19,12 +19,12 @@ RUN yum -y install \
  && yum clean all \
  \
   # TODO: install openssl in /usr/local
- && curl -SL https://www.openssl.org/source/openssl-1.1.1b.tar.gz | tar -zxC / \
- && cd /openssl-1.1.1b \
- && ./config --prefix=/usr \
- && make && make install \
- && rm -rf /openssl-1.1.1b \
- && cd /tmp \
+#  && curl -SL https://www.openssl.org/source/openssl-1.1.1b.tar.gz | tar -zxC / \
```
kneth, line 22: I guess we can remove this part?

scripts/download-realm.js
```diff
@@ -164,6 +167,7 @@ function getCoreRequirements(dependencies, options, required = {}) {
         case 'linux':
             required.CORE_SERVER_FOLDER += `/linux/${flavor}`;
             required.CORE_ARCHIVE = `realm-core-${flavor}-v${dependencies.REALM_CORE_VERSION}-Linux-devel.tar.gz`;
+            required.openssl = "https://static.realm.io/downloads/openssl/1.1.1b/Linux/x86_64/openssl.tgz";
```
kneth, line 170: Maybe `openssl` -> `OPENSSL` to be consistent with the rest.

blagoev: 👍

Dockerfile
blagoev: 👍

PR #2842, opened 2020-04-24
fix integration tests

blagoev:
Fixing integration tests


kraenhansen:
In general I think we should migrate away from installing (or calling CI) in the NPM / package scripts, as this makes the runtime longer if we need to execute more than one of these in sequence. I suggest moving the `npm ci` to the Jenkinsfile if possible.

blagoev:
Yes indeed, although `npm ci` is called anyways from the outside package. What this is doing is trying to fix the existing integration tests runs on CI which seems does not have the dependencies installed on this specific execution path. In general the integration tests scripts in this package need some refactoring. Especially the postinstall scripts.

kraenhansen:
> Especially the postinstall scripts.

What scripts are you referring to, specifically?

blagoev:
the integration-tests scripts are full of `npm ci` and `npm install` in an tangled scripts. Also they have a `postinstall` script which runs every time on install and does not work on Windows. We should refactor them to do nothing on install and require an `npm install` from the user

integration-tests/environments/electron/.npmrc
```diff
@@ -1,3 +1,4 @@
 target=4.1.0
 runtime=electron
 disturl=https://atom.io/download/electron
+package-lock=false
```
kneth, line 4: And this was the trick :smile:

blagoev: yep :)

kraenhansen:
I believe removing the package locks will make it harder for us to reproduce issues between runs on CI and locally, because without them there is no guarantee that the dependencies of the environments will be the same. I think this merge should be reverted as it seems to me that it's fixing one issue and introducing another, (admittedly less obvious) issue.

We should be able to get the CI into the state that it was in before this became an issue or find an alternative solution that doesn't remove our ability to use package locks.

blagoev:
The issue was how the packages are structured, installed and executed. This change is a fix that fixes mostly this. It was a pure luck this was working before. While removing package-locks seems like a step backwards lets ask ourselves why the opportunity exists to do exactly that,  instruct npm not to generate package-lock files. This is exactly for cases like this. As these packages are tightly coupled to each other and have dependencies of tgz files which are prebuild on the exact previous build step. package-locks files are not designed for cases like these and are getting in the way of how these packages were envisioned to be tested. The change introduced here is of fixes to make the current model run without many changes and refactoring of the infrastructure. 

One example of why these npm packages were not alright is the fact that one of the test code paths was doing `npm run build` which did `tsc -p`, but it never got installed and was failing on CI. if done locally this issue was not there. 

As as we have plan to refactor the regular realm-js tests in the near future, we can take the time to refactor these tests as well. 

I consider the fix good enough for now, especially since we are with limited capacity and are on tight schedule. And also because we are not deprived to fix any potential issues of not using package-locks should they present themselves since this is internal code to the project.

PR #2861, opened 2020-05-06
docs and tests for query primitive lists

ironage:
This adds some minimal docs and tests for the query primitive lists feature added to the core query parser (https://github.com/realm/realm-core/pull/3693).

### DIFF
--- a/docs/tutorials/query-language.md
+++ b/docs/tutorials/query-language.md
@@ -119,8 +119,7 @@ let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $frien
 ### Queries on lists of primitives
 
 The query syntax for a list of primitive values is mostly the same as querying a list of objects via links or lists.
-There are however, some small differences. To illustrate, let's construct a movie database, where each movie has an anyonomous star rating, and a list of string tags.
-Note that a normal list of objects should be used to model more complex relationships, such as if the start or tags should belong to a specific user.
+However, there are some small differences. To illustrate, let's construct a movie database, where each movie has an anonymous star rating and a list of string tags.
 
 ```JS
 const MovieSchema = {
@@ -147,7 +146,7 @@ realm.write(() => {
 ```
 
 Just like with lists of objects, we can use aggregates: `.@count`, `.@avg`, `.@min`, `.@max`, `.@sum`.
-Collection operators (`ANY`, `ALL`, `NONE`) are also available and if nothing is specified then `ANY` is implied.
+Collection operators (`ANY`, `ALL`, `NONE`) are also available. If nothing is specified, then `ANY` is implied.
 Let's look at some example queries from the movie database:
 
 ```JS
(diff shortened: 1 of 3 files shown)


docs/tutorials/query-language.md
```diff
@@ -116,6 +116,66 @@ Example:
...
+realm.objects('Movie').filtered('NONE tags CONTAINS " "')
+// Find movies that have an average rating of more than 4 stars
+realm.objects('Movie').filtered('ratings.@avg >= 4')
+// Find movies that do not have any one star ratings
+realm.objects('Movie').filtered('ALL ratings > 1')
+// Find movies that have a tag that is also the title (multi-property comparison of the same types is allowed)
+realm.objects('Movie').filtered('ANY tags = name')
+```
+
+There is one unique operation on lists of primives which is `.length`. This operator will compare the length of
```
kneth, line 168:
```suggestion
There is one unique operation on lists of primitives which is `.length`. This operator will compare the length of
```

kneth, line 168: Unique in what sense? Compared to lists in general?

docs/tutorials/query-language.md
```diff
@@ -116,6 +116,66 @@ Example:
 let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = "SF").@count > 0');
 ```
 
+### Queries on lists of primitives
+
+The query syntax for a list of primitive values is mostly the same as querying a list of objects via links or lists.
+There are however, some small differences. To illustrate, let's construct a movie database, where each movie has an anyonomous star rating, and a list of string tags.
+Note that a normal list of objects should be used to model more complex relationships, such as if the start or tags should belong to a specific user.
```
bmunkholm, line 123: What do you mean with "start ... belong to a user"? 
```suggestion
Note that a normal list of objects should be used to model more complex relationships, such as if the tags should belong to a specific user.
```

bmunkholm: That can probably be left out.

docs/tutorials/query-language.md
```diff
@@ -116,6 +116,66 @@ Example:
 let teens = realm.objects('Contact').filtered('SUBQUERY(friends, $friend, $friend.age > 21 AND $friend.city = "SF").@count > 0');
 ```
 
+### Queries on lists of primitives
+
+The query syntax for a list of primitive values is mostly the same as querying a list of objects via links or lists.
+There are however, some small differences. To illustrate, let's construct a movie database, where each movie has an anyonomous star rating, and a list of string tags.
```
cbush, line 122:
```suggestion
However, there are some small differences. To illustrate, let's construct a movie database, where each movie has an anonymous star rating and a list of string tags.
```

docs/tutorials/query-language.md
```diff
@@ -116,6 +116,66 @@ Example:
...
+  let m1 = realm.create('Movie', {
+    name: 'Inception',
+    ratings: [3, 5, 3, 4, 5, 5],
+    tags: ['dream', 'science fiction', 'thriller'],
+  })
+});
+```
+
+Just like with lists of objects, we can use aggregates: `.@count`, `.@avg`, `.@min`, `.@max`, `.@sum`.
+Collection operators (`ANY`, `ALL`, `NONE`) are also available and if nothing is specified then `ANY` is implied.
```
cbush, line 150:
```suggestion
Collection operators (`ANY`, `ALL`, `NONE`) are also available. If nothing is specified, then `ANY` is implied.
```

docs/tutorials/query-language.md
ironage: yes compared to lists, I'll clarify

docs/tutorials/query-language.md
ironage: ok I'll remove this

ironage:
Docs comments are addressed, and the added tests now pass locally.

PR #2862, opened 2020-05-06
Refix integration tests

kraenhansen:
### History
We've recently experienced a regression with package-lock integrity checks failed for our integration tests environments.
This was fixed with #2842 by removing the environments package locks.
I was personally opposed to this fix since it in my opinion introduces another substantial downside, that we won't be able to reproduce failures between our local setup and CI, since the packages will essentially install the latest compatible version of any package allowed by the semantic constraints in the environments package.json files. In particular, we would always be running our tests against latest React Native (if we didn't pin the version in the package.json file).

The root cause of the issue is [a change we merged back](https://github.com/realm/realm-js/pull/2603) in november of last year, which adds the "realm" and "realm-integration-tests" packages as path dependencies to packed archives. This introduced an issue where the integrity of the archive produced on CI wouldn't match the hash in the package-lock.json.
The workaround was a "preinstall" scripts that would install the archive and update the hashes in the lock before the actual install (or ci) ran.
This worked until [I accidentally changed the preinstall script to run with `--no-save`](https://github.com/realm/realm-js/pull/2793) because - well I forgot why I made it this way in the first place ...

### Why would the archives be dependencies in the first place?
Because a lot of tools, including [electron-builder](https://www.npmjs.com/package/electron-builder), CocoaPods `pod install` and the React Native Android Gradle files all depend on native modules being listed in dependencies for them to correctly bundle or install the packages.

### This PR
This replaces #2842 as an attempt to fix the original issue as well as addressing the concern around removing package-locks from the environment packages.
The idea is to introduce a shared script (peer-deps-as-deps.js) which will temporarily modify an environments package.json by moving its peer dependencies into its dependencies.
This script can be invoked with another command as a runtime argument that it will run and restore the package.json after it exits (or gets interrupted).

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -158,10 +158,6 @@ def nodeIntegrationTests(nodeVersion, platform) {
   unstash "pre-gyp-${platform}-${nodeVersion}"
   sh "./scripts/nvm-wrapper.sh ${nodeVersion} ./scripts/pack-with-pre-gyp.sh"
 
-  dir('integration-tests/tests') {
-    sh "../../scripts/nvm-wrapper.sh ${nodeVersion} npm ci"
-  }
-
   dir('integration-tests') {
     // Renaming the package to avoid having to specify version in the apps package.json
     sh 'mv realm-*.tgz realm.tgz'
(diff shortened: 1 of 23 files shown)


kraenhansen:
This got replaced by https://github.com/realm/realm-js/pull/3690

PR #2874, opened 2020-05-11
Fix OpenSSL linking on Windows

### LINKED ISSUE #2560: Realm crashes on Windows, when using Node version higher than 8.x
## Goals
With Node.js 8, getting closer to becoming unsupported (https://nodejs.org/en/about/releases/), I'm trying to get Realm to work with newer versions of Node under Windows.

## Expected Results
Install and work properly on Windows OS.

## Actual Results
Crashes whenever you try to connect to Realm.
(code ELIFECYCLE errno 3221225477)

## Steps to Reproduce
1. Install Node.js version, higher than 8.x (ex. 10.16.3 LTS).
2. Install Realm.
3. Try opening any realm.

- Realm JS SDK Version: 3.2.0
- Node or React Native: 10.16.3
- Client OS & Version: Windows 10 Pro 64 bit

┆Attachments: <a href="https://jira.mongodb.org/secure/attachment/238987/github-2560.stacktrace.txt">github-2560.stacktrace.txt</a>

fealebenpae:
This closes #2560

The Visual Studio linker ends up mixing up OpenSSL imports from the static OpenSSL library we link against and from the OpenSSL symbols that the `node.lib` import library exports. Because Node.js 10 and up ships an OpenSSL version that's incompatible with the one we link against when their symbols mix up we get crashes. This is verifiable by running `dumpbin /exports realm.node` and inspecting the symbols imported from `node.exe` - they contain some OpenSSL functions, which is not desired.

The fix is twofold: we "hack" gyp so that it rearranges the linker input in the generated project files, making sure that the OpenSSL static library we link comes before the Node.js import library, and we also instruct the linker to "whole archive" link the OpenSSL static library in order to make sure that all its symbols make it into the final executable, trumping any OpenSSL symbols that the Node.js import library might export.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* Fixed a crash on Windows and Node.js 10+ when using Sync over HTTPS. ([#2560](https://github.com/realm/realm-js/issues/2560))
 
 ### Compatibility
 * Realm Object Server: 3.23.1 or later.
--- a/realm.gypi
+++ b/realm.gypi
@@ -236,11 +236,23 @@
           ["OS=='win'", {
             "conditions": [
               ["target_arch=='ia32'", {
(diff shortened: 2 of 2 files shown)


kneth:
@fealebenpae Thanks!

PR #2881, opened 2020-05-14
User customData support

RedBeard0531:
## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -46,6 +46,7 @@
 #include "results.hpp"
 #include "shared_realm.hpp"
 #include "thread_safe_reference.hpp"
+#include "util/scheduler.hpp"
 
 #include <realm/disable_sync_to_disk.hpp>
 #include <realm/global_key.hpp>
@@ -295,6 +296,7 @@ public:
     static void realm_file_exists(ContextType, ObjectType, Arguments &, ReturnValue &);
 
     static void create_user_agent_description(ContextType, ObjectType, Arguments &, ReturnValue &);
+    static void bson_parse_json(ContextType, ObjectType, Arguments &, ReturnValue &);
 
     // static properties
     static void get_default_path(ContextType, ObjectType, ReturnValue &);
@@ -309,6 +311,7 @@ public:
         {"deleteFile", wrap<delete_file>},
         {"exists", wrap<realm_file_exists>},
         {"_createUserAgentDescription", wrap<create_user_agent_description>},
+        {"_bsonParseJsonForTest", wrap<bson_parse_json>},
 #if REALM_ENABLE_SYNC
         {"_asyncOpen", wrap<async_open_realm>},
 #endif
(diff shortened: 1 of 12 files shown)


tests/js/bson-tests.ts
```diff
@@ -0,0 +1,63 @@
+'use strict';
```
kraenhansen, line 1: Any reason this is a `ts` file instead of a `js`?
If these tests are expected to pass in [the React Native application](https://github.com/realm/realm-js/tree/master/tests/react-test-app), I don't think we can use a typescript source file.

But I might be wrong?

src/js_user.hpp
```diff
@@ -146,6 +148,15 @@ void UserClass<T>::get_state(ContextType ctx, ObjectType object, ReturnValue &re
     }
 }
 
+template<typename T>
+void UserClass<T>::get_custom_data(ContextType ctx, ObjectType object, ReturnValue &return_value) {
+    auto custom_data = get_internal<T, UserClass<T>>(object)->get()->custom_data();
+    if (!custom_data)
+        return_value.set_null();
+
+    return return_value.set(js::Value<T>::from_bson(ctx, *custom_data));
```
kneth, line 157: Why is `return` needed here?

src/js_types.hpp
```diff
@@ -488,9 +507,92 @@ inline typename T::Value Value<T>::from_mixed(typename T::Context ctx, const uti
...
+        // XXX this is wrong. Should not be creating a JS Date!
+        // See https://github.com/realm/realm-object-store/pull/1005
+        return from_timestamp(ctx, value.operator Timestamp());
+    case Type::Datetime:
+        // TODO adjust after https://github.com/realm/realm-object-store/pull/1005
+        return from_timestamp(ctx, Timestamp(value.operator bson::Datetime().seconds_since_epoch, 0));
+    case Type::String:
+        return from_string(ctx, value.operator const std::string&());
+    case Type::Binary: {
+        // XXX Should this be realm binary or bson binary?
```
kneth, line 559: We should align with the other SDKs.

src/js_user.hpp
RedBeard0531: It isn't. I must have typed it out of habit when typing the line to return something 🙂

tests/js/bson-tests.ts
RedBeard0531: No real reason. Converted to plain js.

src/js_types.hpp
RedBeard0531: Seems like Cocoa is using native and Java is using BSON. I'm going with the BSON since that gives us a place to put the binary subtype once it is exposed from object store. Users can still get to the js-native buffer by adding `.buffer` to `user.custom_data.my_binary_field`.

Also, if I'm reading this right, we already are inconsistent between engines. v8 uses an ArrayBuffer for binary fields, while jsc uses Uint8Array. This will use Uint8Array everywhere.

bmunkholm:
I assume the "this is the initial state" is no longer true :-)

blagoev:
the description is outdated

src/js_user.hpp
RedBeard0531: But a `return` *is* needed above.  😄 

It may be a good general convention to just always do `return return_value.set_foo(...);` even if the `return` is redundant. Then the js return also acts as a c++ return, reducing some of the impedance mismatch.

RedBeard0531:
> 
> 
> Shouldn't this new property be called `customData` instead? We don't use snake case anywhere else in the JS API, AFAIK.

@fealebenpae good pint. Fixed.

src/js_realm.hpp
```diff
@@ -309,6 +311,7 @@ class RealmClass : public ClassDefinition<T, SharedRealm, ObservableClass<T>> {
         {"deleteFile", wrap<delete_file>},
         {"exists", wrap<realm_file_exists>},
         {"_createUserAgentDescription", wrap<create_user_agent_description>},
+        {"_bsonParseJsonForTest", wrap<bson_parse_json>},
```
kraenhansen, line 314: Any reason we couldn't use the [bson](https://www.npmjs.com/package/bson) package for that?

PR #2894, opened 2020-05-27
Include scheduler.hpp to be able to build without Realm Sync

kneth:
## What, How & Why?

Building without Realm Sync ( `npm install --build-from-source --realm_enable_sync=0`) would fail with the error `incomplete type 'realm::util::Scheduler' named in nested name specifier`.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


blagoev:
I suggest we use this one instead https://github.com/realm/realm-js/pull/2905

blagoev:
this is already in

PR #2897, opened 2020-05-27
Fix ObjectId and Decimal128

kneth:
## What, How & Why?

Fixes two calls to JS functions.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/node/node_value.hpp
+++ b/src/node/node_value.hpp
@@ -18,7 +18,6 @@
 
 #pragma once
 
-#include <iostream>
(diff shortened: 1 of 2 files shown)


src/node/node_value.hpp
```diff
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include <iostream>
```
ironage, line 21: left over from debugging? :)

PR #2898, opened 2020-05-27
App config platform

kneth:
## What, How & Why?

Adds platform parameters to app config.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


lib/extensions.js
```diff
@@ -470,6 +470,32 @@ module.exports = function(realmConstructor, context) {
                     return "RealmJS/Unknown"
                 }
             },
+
+            _createPlatformDescription() {
+                try {
+                    const sdk_version = require("../package.json").version;
+                    if (typeof navigator !== "undefined" && navigator.product === "ReactNative") {
```
blagoev, line 477: could we move this check to a function inside `lib/index.js` and expose it so both here and our `getContext` can use the one and the same function for checking if this environment is RN or node. Otherwise I suspect we will forget to fix this when things change and we will fix only the getContext implementation

tests/js/app-tests.js
```diff
@@ -158,15 +156,18 @@ module.exports = {
             }
         };
         Realm.deleteFile(realmConfig);
-
+        console.log("KAT 1");
```
blagoev, line 159: debugging leftover ?

btw, in bulgarian this means "Traffic Police 1" 😀

tests/js/app-tests.js
```diff
@@ -158,15 +156,18 @@ module.exports = {
...
+        console.log("KAT 2");
         realm.write(() => {
             realm.deleteAll();
         });
+        console.log("KAT 3");
         realm.write(() => {
-            realm.create("Dog", { "_id": new ObjectId('0000002a9a7969d24bea4cf5'), name: "King" });
+            console.log("KAT 4");
+            realm.create("Dog", { "_id": new ObjectId("0000002a9a7969d24bea4cf5"), name: "King" });
             console.log('FISK 1');
```
blagoev, line 169: is this is a left over from previous debugging

tests/js/app-tests.js
kneth: Indeed so - thanks.

lib/extensions.js
blagoev: also I think even better probably call the new function from native code directly instead of having another one in the lib/extensions.js

kneth: I put it next to the old user agent method but I see your point.

kneth: I have moved function.

tests/js/app-tests.js
kneth: Clean up a bit :-)

PR #2900, opened 2020-05-28
Add support for calling server functions

RedBeard0531:
Note that I cherry-picked @kneth's platform parameters fix, so **ignore the first commit**. Before merging I'll rebase on top of something with that fix and remove the cherry-picked commit.

The tests are using the functions defined in https://github.com/realm/realm-java/tree/v10/tools/sync_test_server/app_config/functions. We should probably get these into a common stitch app (possibly in the object-store repo so it can use the app for testing as well) so that we don't all have to reinvent the wheel.

Does this need a chrome debug API change? It didn't look like other methods on `User` had shimming.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [X] jsdoc files updated

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -422,6 +422,30 @@ class User {
     get customData() { }
 
     /**
+     * Calls the named server function as this user.
+     * @param {string} name - name of the function to call
+     * @param {any[]} args - list of arguments to pass
+     */
+    call_function(name, args) { }
+    
+    /**
+     * Convenience wrapper around `call_function(name, [args])`
+     *
+     * @example
+     * // These are all equivalent:
+     * await user.call_function("do_thing", [a1, a2, a3]);
+     * await user.functions.do_thing(a1, a2, a3);
+     * await user.functions["do_thing"](a1, a2, a3);
+     *
+     * @example
+     * // It it legal to store the functions as first-class values:
+     * const do_thing = user.functions.do_thing;
+     * await do_thing(a1);
+     * await do_thing(a2);
+     */   
+    get functions() { }
+
+    /**
      * Creates the configuration object required to open a synchronized Realm.
      *
      * @param {Realm.PartialConfiguration} config - optional parameters that should override any default settings.
--- a/src/js_types.hpp
+++ b/src/js_types.hpp
@@ -163,6 +163,7 @@ struct Value {
     static ObjectType to_object(ContextType, const ValueType &);
     static String<T> to_string(ContextType, const ValueType &);
     static OwnedBinaryData to_binary(ContextType, ValueType);
+    static bson::Bson to_bson(ContextType, ValueType);
 
 
 #define VALIDATED(return_t, type) \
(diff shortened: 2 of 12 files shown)


src/js_user.hpp
```diff
@@ -254,14 +256,43 @@ void UserClass<T>::link_credentials(ContextType ctx, ObjectType this_object, Arg
         }
 
         ValueType callback_arguments[2];
-        callback_arguments[0] = create_object<T, UserClass<T>>(ctx, new User<T>(shared_user, user->m_app));
+        callback_arguments[0] = create_object<T, UserClass<T>>(protected_ctx, new User<T>(shared_user, user->m_app));
```
kneth, line 259: Well spotted.

docs/sync.js
```diff
@@ -421,6 +421,30 @@ class User {
      */
     get customData() { }
 
+    /**
+     * Calls the named function as this user.
```
bmunkholm, line 425: Perhaps expand this description to say it's a remote function defined in MongoDB Realm Cloud?

kneth:
> Does this need a chrome debug API change? It didn't look like other methods on User had shimming.

At least we need to add all methods to `lib/browser/user.js` (the array in the call `createMethods`).

src/js_types.hpp
```diff
@@ -550,13 +572,20 @@ inline typename T::Value Value<T>::from_bson(typename T::Context ctx, const bson
     case Type::Int32:
         // All int32 values can be precisely represented as a double
         return from_number(ctx, double(value.operator int32_t()));
-    case Type::Int64:
-        // int64 needs special handling.
-        // TODO: Consider using plain js numbers for values within [-2^52, 2^52).
+    case Type::Int64: {
+        // int64 needs special handling. Stitch uses it for all intish numbers, even 1.0, so we map
```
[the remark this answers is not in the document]
RedBeard0531: I'll just call it "the server"

docs/sync.js
RedBeard0531: Does `Calls the named server function as this user.` work for you? Using the full name just reads like ad copy in the middle of the docs 🙂

bmunkholm: yeah - "server" at least let's the user know it's not just a local function :-)

src/js_user.hpp
```diff
@@ -254,14 +256,43 @@ void UserClass<T>::link_credentials(ContextType ctx, ObjectType this_object, Arg
...
+    args.validate_count(3);
+    auto user = get_internal<T, UserClass<T>>(ctx, this_object);
+
+    auto name = Value::validated_to_string(ctx, args[0], "name");
+    auto call_args_js = Value::validated_to_array(ctx, args[1], "args");
+    auto callback = Value::validated_to_function(ctx, args[2], "callback");
+    auto call_args_bson = Value::to_bson(ctx, call_args_js);
+
+    user->m_app->call_function(
+        *user,
```
blagoev, line 278: Probably make sense to validate user is not nullptr since get_internal will return that if `this` is not a valid js user instance.

tests/js/app-tests.js
```diff
@@ -132,15 +132,18 @@ module.exports = {
             }
         };
         Realm.deleteFile(realmConfig);
-
+        console.log("KAT 1");
```
blagoev, line 135: these `console.logs` are debugging leftovers from another branch

tests/js/asserts.js
```diff
@@ -140,17 +140,23 @@ module.exports = {
     },
 
     assertThrows: function(func, errorMessage, depth) {
-        let caught = false;
         try {
             func();
         }
         catch (e) {
-            caught = true;
+            return e
```
blagoev, line 147: good refactor

src/js_user.hpp
blagoev: actually checked it. for node it will throw `invalid object` and for JSC it will return `nullptr`

blagoev: ok ignore this one. resolving

tests/js/app-tests.js
RedBeard0531: Yeah, that was the commit to ignore since I'm going to remove it during rebase 🙂

blagoev: 👍

PR #2912, opened 2020-06-03
RPi support for realm-js

RedBeard0531:
## ☑️ ToDos

* [x] 📝 Changelog entry
* ~~[ ] 📝 `Compatibility` label is updated or copied from previous entry~~
* ~~[ ] 🚦 Tests~~
* ~~[ ] 📝 Public documentation PR created or is not necessary~~
* ~~[ ] 💥 `Breaking` label has been applied or is not necessary~~

* ~~[ ] typescript definitions file is updated~~
* ~~[ ] jsdoc files updated~~
* ~~[ ] Chrome debug API is updated if API is available on React Native~~


RedBeard0531:
After discussing with @kneth, I backed out the Jenkinsfile changes and am putting this up for review without CI integration in order to get it into beta-1. I'll work on getting it built in CI for beta-2. That means that for now users will need to compile it locally when running `npm install`. I've confirmed that this works automatically on my RPi4.

PR #2919, opened 2020-06-05
handle error responses

ironage:
We should let object store handle the responses, there is logic there already to parse valid json and show the error. Intercepting these here means we would need special handling in the js onError code to parse the json or pass through the entire response like this: `"{"error":"cannot find app using Client App ID 'smurf'"}"`
TODO:
- [x] increment the version of realm-network-transport
- [x] publish this
- [x] upgrade js dependency on v10

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -595,7 +595,7 @@ def testMacOS(target, postStep = null) {
     node('osx_vegas') {
       withEnv(['DEVELOPER_DIR=/Applications/Xcode-11.2.app/Contents/Developer',
                'REALM_SET_NVM_ALIAS=1',
-               'DISABLE_REALM_SYNC=1']) {
+               'REALM_DISABLE_SYNC_TESTS=1']) {
         doInside('./scripts/test.sh', target, postStep)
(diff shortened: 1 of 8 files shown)


ironage:
@kneth if this makes sense, can you publish?

kraenhansen:
I need to remember to update Realm Web after this change has been merged.

kneth:
I think the consequence is that even when the callbacks from object store have no error defined, we might be in an error situation.

ironage:
This change combined with https://github.com/realm/realm-js/pull/2901/commits/e7485f477b94c9038f3b4a617cf9dd5b797b2eef will handle both a network error from the transport as well as errors generated from valid json returned from the stitch server. If the object-store code fails to parse the response body, it will return "http error code considered fatal" just based on the response code.

kneth:
@ironage In that case, I am all for it.

ironage:
I don't have access to the npm user to publish this, can someone do it for me?

ironage:
I've published realm-network-transport 0.3.0

PR #2925, opened 2020-06-06
build with sync but allow CI to exclude sync tests

ironage:
build with sync on mac, but don't run the tests there because we don't have docker to spin up a local server


blagoev:
the fix is good. merging it on James behalf

PR #2929, opened 2020-06-06
Allow Decimal128 to be null

kneth:
## What, How & Why?

Stupid mistake not to allow nulls.

## ☑️ ToDos

* ~[ ] 📝 Changelog entry~
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~

### DIFF
--- a/src/node/node_value.hpp
+++ b/src/node/node_value.hpp
@@ -242,6 +242,10 @@ template<>
 inline Napi::Value node::Value::from_decimal128(Napi::Env env, const Decimal128& number) {
 	Napi::EscapableHandleScope scope(env);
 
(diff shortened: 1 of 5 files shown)


src/node/node_value.hpp
```diff
@@ -242,6 +242,10 @@ template<>
 inline Napi::Value node::Value::from_decimal128(Napi::Env env, const Decimal128& number) {
 	Napi::EscapableHandleScope scope(env);
 
+	if (number.is_null()) {
```
blagoev, line 245: could we use Napi::Is_Null here

blagoev: ignore

PR #2939, opened 2020-06-08
Fix Realm.App constructor's TS def

kneth:
Closes https://github.com/realm/realm-js/issues/2940

### DIFF
--- a/types/app.d.ts
+++ b/types/app.d.ts
@@ -26,7 +26,7 @@ declare namespace Realm {
     // See https://stackoverflow.com/a/51114250 on why we're importing the BSON types like this
     type ObjectId = import("bson").ObjectId;
     type Binary = import("bson").Binary;
-    
+
     namespace Credentials {
         /**
          * Payload sent when authenticating using the [Email/Password Provider](https://docs.mongodb.com/stitch/authentication/userpass/).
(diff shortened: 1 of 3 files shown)


types/app.d.ts
```diff
@@ -24,9 +24,8 @@
 
 declare namespace Realm {
     // See https://stackoverflow.com/a/51114250 on why we're importing the BSON types like this
-    type ObjectId = import("bson").ObjectId;
```
kraenhansen, line 27: Any reason this is getting removed? It's used by `services.d.ts` and `auth-providers.d.ts`.

kraenhansen: We might want to move it to line 27 of index.d.ts if we want to use it there as well.

types/app.d.ts
```diff
@@ -123,6 +122,9 @@ declare namespace Realm {
      * A MongoDB Realm App.
      */
     class App<FunctionsFactoryType extends object = DefaultFunctionsFactory> {
+
+        constructor(AppConstructor);
```
kraenhansen, line 126: Shouldn't this be taking an `AppConfiguration` and not the `AppConstructor` type here? Did you test this manually?

nirinchev:
What's the status of this one?

kneth:
Superseded by #3056

PR #2941, opened 2020-06-08
Add RemoteMongoClient

RedBeard0531:
## ☑️ ToDos

* [X] 📝 Changelog entry
* [X] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -555,6 +555,20 @@ class User {
      * await do_thing(a2);
      */
     get functions() { }
+
+    /**
+     * Returns a connection to the MongoDB service.
+     *
+     * @example
+     * let blueWidgets = user.remoteMongoClient('myClusterName')
+     *                       .db('myDb')
+     *                       .collection('widgets')
+     *                       .find({color: 'blue'});
+     *
+     * @param {string} serviceName
+     * @returns {Realm~RemoteMongoDB}
+     */
+    remoteMongoClient(serviceName) { }
 }
 
 /**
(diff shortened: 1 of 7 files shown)


lib/mongo_client.js
```diff
@@ -0,0 +1,211 @@
+////////////////////////////////////////////////////////////////////////////
```
RedBeard0531, line 1: This is a JS-ified copy of https://github.com/realm/realm-js/blob/v10/packages/realm-web/src/services/RemoteMongoDBService.ts with only the RemoteMongoDBCollection type and without the factory functions. The constructor was lightly modified to accept the `functions` object directly rather than constructing it itself.

lib/user.js
```diff
@@ -31,9 +33,23 @@ const instanceMethods = {
         });
     },
 
-    callFunction(name, args) {
+    callFunction(name, args, service = undefined) {
```
RedBeard0531, line 36: Added the optional `service` param to support `remoteMongoClient()`. Also now using `cleanArgs()` preprocessing similar taken from https://github.com/realm/realm-js/blob/11f690217253a095cf605779d006f0be9f7cb6a1/packages/realm-web/src/FunctionsFactory.ts#L51-L68

lib/user.js
```diff
@@ -55,17 +71,41 @@ const instanceMethods = {
...
+                            user._functionsOnService(serviceName),
+                            dbName,
+                            collName);
+                    },
                 };
             },
+        };
+    },
+
+    _functionsOnService(service) {
```
RedBeard0531, line 90: This is basically the same as the old `get functions()` getter but now takes a `service` and uses the property name filtering logic taken from https://github.com/realm/realm-js/blob/11f690217253a095cf605779d006f0be9f7cb6a1/packages/realm-web/src/FunctionsFactory.ts#L149-L158

lib/user.js
```diff
@@ -31,9 +33,23 @@ const instanceMethods = {
...
+                    }
+                }
+            }
+            return args;
+        }
+        args = cleanArgs(args);
+
         return new Promise((resolve, reject) => {
-            this._callFunction(name, args, (result, error) => {
+            this._callFunction(name, args, service, (result, error) => {
```
blagoev, line 52: this does not resolve

blagoev: ah.. ignore. the github ui have failed me

lib/user.js
```diff
@@ -55,17 +71,41 @@ const instanceMethods = {
...
 
-    get functions() {
+    remoteMongoClient(serviceName) {
         const user = this;
-        return new Proxy({}, {
-            get(target, name) {
-                return (...args) => {
-                    return user.callFunction(name, args);
+        return {
+            db(dbName) {
```
blagoev, line 77: Is this calling a function or returning an object?

src/js_user.hpp
```diff
@@ -242,18 +242,23 @@ void UserClass<T>::link_credentials(ContextType ctx, ObjectType this_object, Arg
 
 template<typename T>
 void UserClass<T>::call_function(ContextType ctx, ObjectType this_object, Arguments& args, ReturnValue &) {
-    args.validate_count(3);
+    args.validate_count(4);
```
blagoev, line 245: are all arguments required. Isn't something optional

lib/user.js
RedBeard0531: I have a patch locally that makes this just `return promisify(cb => this._callFunction(name, args, service, cb));` so you won't have to worry about that in the future :)

src/js_user.hpp
```diff
@@ -242,18 +242,23 @@ void UserClass<T>::link_credentials(ContextType ctx, ObjectType this_object, Arg
...
     auto user = get_internal<T, UserClass<T>>(ctx, this_object);
 
     auto name = Value::validated_to_string(ctx, args[0], "name");
     auto call_args_js = Value::validated_to_array(ctx, args[1], "args");
-    auto callback = Value::validated_to_function(ctx, args[2], "callback");
+    auto service = Value::is_undefined(ctx, args[2])
```
blagoev, line 250: if service is optional isn't it better to be last parameter. Or this is just because to follow JS convention where callback is last

src/js_user.hpp
RedBeard0531: Optional in the user-facing JS API. Not optional in the internal C++ API. Well semi-optional since it accepts an explicit `undefined`.

src/js_user.hpp
RedBeard0531: I generally try to follow the callback-last API principle since it makes calling code clearer. And in this case the only caller always explicitly provides all arguments, one of them is just frequently the value `undefined`.

lib/user.js
RedBeard0531: `remoteMongoClient()` is returning an object with a single method `db()`, which returns and object with the single method `collection()` which returns a RemoteMongoDBCollection object.

It may be clearer if you see a usage example: https://github.com/realm/realm-js/blob/d13d14eb38c7b27ad7b3c5fd2ceb6d479d2f704e/tests/js/user-tests.js#L158-L165

tests/js/user-tests.js
```diff
@@ -150,6 +150,21 @@ module.exports = {
     TestCase.assertEqual(err.message, "function not found: 'error'");
   },
 
+  async testRemoteMongoClient() {
+    let app = new Realm.App(appConfig);
```
kraenhansen, line 154: Not really testing much, but I guess that's fine for now, we should reuse the Realm Web integration tests long term.

tests/js/user-tests.js
```diff
@@ -150,6 +150,21 @@ module.exports = {
...
+    let credentials = Realm.Credentials.anonymous();
+    let user = await app.logIn(credentials);
+
+    let mongo = user.remoteMongoClient('BackingDB');
+    let collection = mongo.db('test').collection('testRemoteMongoClient');
+
+    await collection.deleteMany({});
+    await collection.insertOne({hello: "world"});
```
kraenhansen, line 162: Is this relying on a cleanup defined elsewhere or does this require a manual reset of the database?

lib/user.js
blagoev: got it

src/js_user.hpp
blagoev: 👍

tests/js/user-tests.js
RedBeard0531: Yeah, the tests were just to ensure that everything was hooked up right and working sanely. Agree that we should one have one comprehensive test long term.

src/js_user.hpp
blagoev: 👍

tests/js/user-tests.js
RedBeard0531: Line 161 does cleanup so the rest of the test always runs in a clean state.

kraenhansen: 🤦 Never mind ...

RedBeard0531:
I added jsdoc documentation for this, mostly copied and modified from the .d.ts file. Let me know if you are happy with the result. Or if you would prefer me to open a separate PR for that rather than adding it to this one.

docs/sync.js
```diff
@@ -762,3 +776,208 @@ class Worker {
...
+     * See [Tutorial: Project Fields to Return from Query](https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/).
+     * @param {object} [options.sort] The order in which to return matching documents.
+     * @returns {object} The document or null if nothing matched.
+     */
+    async findOne(filter, options) { }
+
+    /**
+     * Finds a document which matches the provided query and performs the desired update to individual fields.
+     *
+     * @param {object} filter An filter applied to narrow down the results.
```
bmunkholm, line 843:
```suggestion
     * @param {object} filter A filter applied to narrow down the results.
```

docs/sync.js
```diff
@@ -762,3 +776,208 @@ class Worker {
...
+     * @param {boolean} [options.returnNewDocument=false] if true, indicates that the action should return
+     * the document in its updated form instead of its original, pre-update form.
+     * @returns {?object} The document (before or after modification) or null if nothing matched.
+     */
+    async findOneAndUpdate(filter, update, options) { }
+
+    /**
+     * Finds a document which matches the provided filter and replaces it with a new document.
+     *
+     * @param {object} filter An filter applied to narrow down the results.
```
bmunkholm, line 860:
```suggestion
     * @param {object} filter A filter applied to narrow down the results.
```

docs/sync.js
```diff
@@ -762,3 +776,208 @@ class Worker {
...
+     * @param {boolean} [options.returnNewDocument=false] if true, indicates that the action should return
+     * the document in its updated form instead of its original, pre-update form.
+     * @returns {?object} The document (before or after modification) or null if nothing matched.
+     */
+    async findOneAndReplace(filter, replacement, options) { }
+
+    /**
+     * Finds a document which matches the provided filter and deletes it
+     *
+     * @param {object} filter An filter applied to narrow down the results.
```
bmunkholm, line 877:
```suggestion
     * @param {object} filter A filter applied to narrow down the results.
```

docs/sync.js
```diff
@@ -762,3 +776,208 @@ class Worker {
...
+     * @typedef {object} Realm.RemoteMongoDBCollection~UpdateResult Result of updating documents
+     * @property {number} matchedCount The number of documents that matched the filter.
+     * @property {number} modifedCount The number of documents matched by the query.
+     * @property [upsertedId] The identifier of the inserted document if an upsert took place.
+     */
+
+    /**
+     * Updates a single document matching the provided filter in this collection.
+     *
+     * @param {object} filter An filter applied to narrow down the results.
```
bmunkholm, line 963:
```suggestion
     * @param {object} filter A filter applied to narrow down the results.
```

docs/sync.js
```diff
@@ -762,3 +776,208 @@ class Worker {
...
+     * @param {boolean} [options.upsert=false] if true, indicates that MongoDB should insert a new document that matches the
+     * query filter when the query does not match any existing documents in the collection.
+     * @returns {Realm.RemoteMongoDBCollection~UpdateResult}
+     */
+    async updateOne(filter, update, options) { }
+
+    /**
+     * Updates multiple documents matching the provided filter in this collection.
+     *
+     * @param {object} filter An filter applied to narrow down the results.
```
bmunkholm, line 975:
```suggestion
     * @param {object} filter A filter applied to narrow down the results.
```

PR #2947, opened 2020-06-08
beef up the user tests

ironage:
Now including testing the email type users by way of the auto register function that is already installed to the test app.

### DIFF
--- a/tests/js/user-tests.js
+++ b/tests/js/user-tests.js
@@ -83,31 +83,51 @@ function randomVerifiableEmail() {
     return `realm_tests_do_autoverify_${Utils.uuid()}_@test.com`;
 }
 
+function randomNonVerifiableEmail() {
+  // according to the custom register function, emails will not register if they don't contain "realm_tests_do_autoverify"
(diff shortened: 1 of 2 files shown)


tests/js/user-tests.js
```diff
@@ -150,107 +222,95 @@ module.exports = {
...
+    assertIsSameUser(all[user1.identity], user1);
+    TestCase.assertFalse(user2.isLoggedIn);
+    TestCase.assertTrue(user1.isLoggedIn);
+    TestCase.assertArrayLength(Object.keys(all), 2, "still holds references to both users");
+
+    await user1.logOut();
+    all = app.allUsers();
+    TestCase.assertFalse(user1.isLoggedIn);
+    TestCase.assertFalse(user2.isLoggedIn);
+    TestCase.assertArrayLength(Object.keys(all), 2, "still holds references to both users"); // FIXME: is this actually expected?
```
kneth, line 314: I think it is expected - for now. I agree it is counter-intuitive that anonymous users are not removed when they log out. 

See also https://github.com/realm/realm-object-store/issues/1031

PR #2950, opened 2020-06-09
Merge Realm JS and Realm Web types

kraenhansen:
## What, How & Why?

The purpose of this PR is to merge duplicated type definitions (such as `App`, `User`, `UserProfile`) across ´types/index.d.ts` and the other files in that directory.

I propose the following changes to the Realm JS types:
- `User` related changes:
  - Rename `identity` on `User` to `id` (to align with other SDKs and avoid confusion from authentication provider `identities` returned when fetching the user profile).
  - Rename `token` to `accessToken` (to make it more obvious what token this is) and add the `refreshToken` as a property (this is up for debate - but it makes it easier to test the User implementation).
  - Change `customData` to use a generic type defined by `User` and the `App` (like we do for `FunctionsFactoryType`) to allow users of the API to define types for this custom data, once (when creating the app instance).
  - Change the return type of `logOut` to `Promise<void> | void`, since it sends a request to invalidate the refresh token.
- `App` related changes:
  - Change the `allUsers` from a method to a getter-property returning a list instead of a `UserMap`.
  - Change the `currentUser` from a method to a getter-property.
  - Have `removeUser` return a `Promise<void>` instead of `Promise<User>`, since that user would have limited use anyway.
- `Auth` related changes:
  - Rename from `API*` to `Api`.
  - Prepend `Auth` to the class names (`Realm.Auth.ApiKeyAuth` and `Realm.Auth.EmailPasswordAuth`).
  - A lot of changes to the naming of methods in the provider clients (see commits for details).
  - Added a `callResetPasswordFunction` which was part of the existing Stitch SDK.

I propose the following changest to the Realm Web Types:
- `App` related changes:
  - Remove `string` as a valid argument on methods called with a user parameter (it was ment to enable users to pass in a user ID instead of a user object).
- `UserProfile` related changes:
  - Rename `pictureURL` to `pictureUrl`.
- `Auth` related changes:
  - Rename `ApiKeyProvider` to `ApiKeyAuth`.
  - Rename `EmailPasswordProvider` to `EmailPasswordAuth`.
  - Rename `get` and `list` to `fetch` and `fetchAll` respectively, on the `ApiKeyProvider`.
  - Rename `resendConfirmation` to `resendConfirmationEmail` on the `EmailPasswordProvider`.

These changes lead to some simplifications of the way user state is managed by the app, which also found its way into this PR.


packages/realm-web/src/User.ts
```diff
@@ -240,17 +201,37 @@ export class User implements Realm.User {
...
         this._state = UserState.LoggedOut;
     }
+
+    /** @inheritdoc */
+    public async linkCredentials(credentials: Realm.Credentials) {
+        throw new Error("Not yet implemented");
+    }
+
+    public async refreshCustomData() {
+        await this.refreshAccessToken();
```
kneth, line 226: So in reality, an exception is thrown, right?

types/app.d.ts
```diff
@@ -194,34 +201,99 @@ declare namespace Realm {
...
+        readonly accessToken: string | null;
 
         /**
          * The refresh token used when requesting a new access token.
          */
-        refreshToken: string | null;
+        readonly refreshToken: string | null;
+
+        /**
+         * You can store arbitrary data about your application users in a MongoDB collection and configure Stitch to automatically expose each user’s data in a field of their user object.
```
kneth, line 232:
```suggestion
         * You can store arbitrary data about your application users in a MongoDB collection and configure MongoDB Realm to automatically expose each user’s data in a field of their user object.
```

packages/realm-web/src/User.ts
kraenhansen: Yeah, for now :)

blagoev:
what does this means 

> Change the return type of logOut to Promise<void> | void, since it sends a request to invalidate the refresh token.
I find it problematic to have the return type differ. Sometimes returning Promise and sometimes void.  When would this return void. Or I am not reading this right?

kraenhansen:
> I find it problematic to have the return type differ. Sometimes returning Promise and sometimes void. When would this return void. Or I am not reading this right?

Ideally this would just be `Promise<void>` and you're right, it's confusing.

It's because Realm Web sends a request (like the legacy SDKs) but that doesn't seem to be implemented in the way Object Store logs out a user - or it's implemented in a way that allows it to fail silently.

I propose we change it to make Realm JS return a resolved `Promise<void>` (even though it's just a synchronous call underneath) and create an issue in object store tracking that it should send the request to invalidate the refresh token. Would that be better?

blagoev:
exactly what I was thinking. if its a synchronous call in some cases this should return a resolved Promise.

packages/realm-web/src/User.ts
```diff
@@ -153,13 +114,7 @@ export class User implements Realm.User {
     private _state: Realm.UserState;
     private transport: AuthenticatedTransport;
```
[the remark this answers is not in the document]
kraenhansen: The private fields with an `_` prefix is only because the class also has a public getter overloading the name. It could be called `internalState` or something else. But I actually think this specific fields goes away in one of the later PRs.

With regards to JS private fields. It's valuable to keep an eye on, but I don't think [the adoption in browsers](https://caniuse.com/#feat=mdn-javascript_classes_private_class_fields) are good enough for us to rely on. And I don't know the state of polyfils for this?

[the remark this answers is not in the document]
steffenagger: I think TS handles that, but I'm not sure (as in I haven't tested it) https://github.com/microsoft/TypeScript/pull/30829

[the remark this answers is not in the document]
kraenhansen: Oh ...

packages/realm-web/src/App.ts
```diff
@@ -122,36 +124,23 @@ export class App<
...
-                // Remove the user from the stack
-                const [handle] = this.users.splice(index, 1);
-                // Insert the user in the beginning of the stack
-                this.users.splice(0, 0, handle);
-            } else {
-                throw new Error("The user was not logged into this app");
-            }
+    public switchUser(nextUser: User<FunctionsFactoryType, CustomDataType>) {
+        const index = this.users.findIndex(u => u === nextUser);
+        if (index >= 0) {
```
kneth, line 137: `index !== -1` might be more precise

packages/realm-web/src/App.ts
```diff
@@ -162,45 +151,25 @@ export class App<
...
-        // If active - log out the user
-        if (user.state === UserState.Active) {
-            await this.logOut(user);
-        }
-        // Set the state of the user
-        controller.setState(UserState.Removed);
+    public async removeUser(user: User<FunctionsFactoryType, CustomDataType>) {
         // Remove the user from the list of users
-        const index = this.users.findIndex(({ user: u }) => u === user);
+        const index = this.users.findIndex(u => u === user);
```
kneth, line 169: Will we always find a user?

types/app.d.ts
```diff
@@ -194,34 +190,116 @@ declare namespace Realm {
...
+
+        /**
+         * Log out the user.
+         * 
+         * @returns A promise that resolves once the user has been logged out of the app.
+         */
+        logOut(): Promise<void>;
+
+        /**
+         * Link the user with a new identity represented by another set of credentials.
```
kneth, line 251:
```suggestion
         * Link the user with an identity represented by another set of credentials.
```

types/app.d.ts
```diff
@@ -194,34 +190,116 @@ declare namespace Realm {
...
+         * 
+         * @param credentials The credentials to use when linking.
+         */
+        linkCredentials(credentials: Credentials): Promise<void>;
+
+        /**
+         * Call a remote MongoDB Realm function by it's name.
```
kneth, line 258:
```suggestion
         * Call a remote MongoDB Realm function by its name.
```

packages/realm-web/src/App.ts
kraenhansen: If not, `findIndex` will return `-1`.

PR #2954, opened 2020-06-09
RPi support in Jenkins

RedBeard0531:
## What, How & Why?

This adds building of binary releases for RPi. Because of a compatability issue, it removes building on node 11.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


kneth (approved):
I suggest that we also update https://github.com/realm/realm-js/blob/v10/README.md#getting-started

Issue #2948, opened 2020-06-09
Translate type names


RealmBot:
Object store returns `object id` and `decimal`, and we what to see `objectId` and `decimal128`.

Moreover, object store will also use these names in error messages.

blagoev:
fixed by #2974

PR #2974, opened 2020-06-15
use correct JS type name for Decimal128 and ObjectID

blagoev:
use correct JS type name for Decimal128 and ObjectID

### DIFF
--- a/src/js_realm.cpp
+++ b/src/js_realm.cpp
@@ -106,7 +106,7 @@ std::string TypeErrorException::type_string(Property const& prop)
             ret = "decimal128";
             break;
         case PropertyType::ObjectId:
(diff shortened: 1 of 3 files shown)


kraenhansen:
I searched for "decimal128" across the codebase and found a couple of places that I think we need to update, if the change to `Decimal128` (with a capital D) is the intended change:
- https://github.com/realm/realm-js/blob/v10/src/js_realm.cpp#L106
- https://github.com/realm/realm-js/blob/v10/src/js_schema.hpp#L108-L113

blagoev:
fixes https://github.com/realm/realm-js/issues/2948

bmunkholm:
Missing note in changelog @blagoev ?

PR #2918, opened 2020-06-04
v10.0.0 beta.1

kneth:

PR #2988, opened 2020-06-18
Adds missing syncconfig.error typescript definitions

nirinchev:
## What, How & Why?
Noticed that error was missing on the sync configuration typings.

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ NOTE: This version bumps the Realm file format to version 11. It is not possible
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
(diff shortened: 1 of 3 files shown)


kraenhansen:
It looks to me that this property (and the `ErrorCallback` type) was removed with https://github.com/realm/realm-js/pull/2918.
@kneth do you have any context as to why this was removed?

@nirinchev I suggest we also [add back the `ErrorCallback` type](https://github.com/realm/realm-js/pull/2918/files#diff-88bec6beae84369c0376b56b3bb88fe1L422), as it has dependencies on it:

```typescript
type ErrorCallback = (session: Session, error: SyncError) => void;
```

nirinchev:
What are those dependencies?

kraenhansen:
I allowed myself to add it back in.
For one, Studio is dependent on it - and I think I've used it elsewhere in other tools too.

It is an exported type, so anything really could be using it, hence it would be breaking to remove it.

PR #2990, opened 2020-06-19
Realm Web: Persist and retrieve users

### LINKED ISSUE #2964: Realm Web: Persist and retrieve active users
Currently, when the web-browser is refreshed the currently logged-in user is forgotten.
This needs to change by:
1. Persisting the users access and refresh tokens in the browsers local-storage whenever these are known or gets updated.
2. Retrieve stored credentials when the library gets loaded.

kraenhansen:
## What, How & Why?

This closes #2964 by adding:
- A storage abstraction and three specializations:
  - `LocalStorage` the browsers local storage.
  - `MemoryStorage` ephemeral storage used when running from Node.js (during unit tests).
  - `PrefixedStorage` wrapping another `Storage`, adding a prefix to the key before getting or setting values, helps keep storage local and avoid repeating the prefixed key in code.
- Storing a users access & refresh tokens after successful log in.
- Storing a users profile after successful refresh.
- Restoring the list of users and their data (id, access & refresh tokens and profile) when an instance of `App` gets constructed.

Since localStorage is shared across tabs of the same origin and the current implementation allows for multiple `App` instances with the same `id` a couple of considerations had to be made when adding this feature.

### Merge users when logged in across multiple tabs
Consider the following activity from two app instances with the same ID:
1. App-1 logs in "Alice"
2. App-2 logs in "Bob" (in a different tab)
3. App-1 logs in "Charlie"
4. All tabs are closed and the user returns the next day
5. App-3 restores users.

In this case both "Alice", "Bob" and "Charlie" must be logged in, "Charlie" being the most recent (and therefore current user).

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web-integration-tests/src/app.test.ts
+++ b/packages/realm-web-integration-tests/src/app.test.ts
@@ -20,12 +20,22 @@ import { expect } from "chai";
 
 import { App, Credentials } from "realm-web";
 
-import { createApp } from "./utils";
+import { createApp, clearStorage } from "./utils";
+
+async function logOutAllUsers(app: App<object>) {
+    for (const user of app.allUsers) {
+        await user.logOut();
+    }
+}
 
 describe("App#constructor", () => {
     it("constructs", () => {
         const app = new App("default-app-id");
         expect(app).to.be.instanceOf(App);
+    });
+
+    afterEach(() => {
+        clearStorage();
     });
 
     it("can login a user", async () => {
--- a/packages/realm-web-integration-tests/src/app.test.ts
+++ b/packages/realm-web-integration-tests/src/app.test.ts
@@ -18,9 +18,11 @@
 
 import { expect } from "chai";
 
-import { App, Credentials, User } from "realm-web";
+import { App, Credentials, User, createDefaultStorage } from "realm-web";
 
-import { createApp, clearStorage } from "./utils";
+import { createApp } from "./utils";
+
+const storage = createDefaultStorage();
 
 describe("App#constructor", () => {
     it("constructs", () => {
@@ -29,7 +31,7 @@ describe("App#constructor", () => {
     });
 
     afterEach(() => {
-        clearStorage();
+        storage.clear();
     });
 
     it("can login a user", async () => {
@@ -79,7 +81,6 @@ describe("App#constructor", () => {
             const credentials = Credentials.anonymous();
             user = await app.logIn(credentials);
             expect(typeof user.id).equals("string");
-            console.log(window.localStorage);
         }
         // Recreate the app and expect the user to be restored
         {
--- a/packages/realm-web-integration-tests/src/app.test.ts
+++ b/packages/realm-web-integration-tests/src/app.test.ts
@@ -18,15 +18,9 @@
 
 import { expect } from "chai";
 
-import { App, Credentials } from "realm-web";
+import { App, Credentials, User } from "realm-web";
 
 import { createApp, clearStorage } from "./utils";
-
-async function logOutAllUsers(app: App<object>) {
-    for (const user of app.allUsers) {
-        await user.logOut();
-    }
-}
 
 describe("App#constructor", () => {
     it("constructs", () => {
@@ -76,5 +70,28 @@ describe("App#constructor", () => {
         expect(app.allUsers).deep.equals([user2, user1]);
         await app.removeUser(user1);
         expect(app.allUsers).deep.equals([user2]);
+    });
+
+    it("restores a user", async () => {
+        let user: User<object>;
+        {
+            const app = createApp();
+            const credentials = Credentials.anonymous();
+            user = await app.logIn(credentials);
+            expect(typeof user.id).equals("string");
+            console.log(window.localStorage);
+        }
+        // Recreate the app and expect the user to be restored
+        {
+            const app = createApp();
+            expect(app.allUsers.length).equals(1);
+            expect(app.currentUser).instanceOf(User);
+            expect(app.currentUser?.id).equals(user.id);
+            expect(app.currentUser?.profile).deep.equals(user.profile);
+            expect(app.currentUser?.accessToken).equals(user.accessToken);
+            expect(app.currentUser?.refreshToken).deep.equals(
+                user.refreshToken,
+            );
+        }
     });
 });
--- a/packages/realm-web-integration-tests/src/utils.ts
+++ b/packages/realm-web-integration-tests/src/utils.ts
@@ -22,7 +22,9 @@ import { App } from "realm-web";
 declare const APP_ID: string;
 declare const BASE_URL: string;
 
-export function createApp<FunctionsFactoryType extends object>() {
+export function createApp<
+    FunctionsFactoryType extends object = Realm.DefaultFunctionsFactory
+>() {
     if (typeof APP_ID !== "string") {
         throw new Error("Expected a global APP_ID");
     }
--- a/packages/realm-web-integration-tests/src/utils.ts
+++ b/packages/realm-web-integration-tests/src/utils.ts
@@ -34,3 +34,9 @@ export function createApp<FunctionsFactoryType extends object>() {
         baseUrl: BASE_URL,
     });
 }
+
+export function clearStorage() {
+    if (window.localStorage) {
+        localStorage.clear();
+    }
+}
--- a/packages/realm-web-integration-tests/src/utils.ts
+++ b/packages/realm-web-integration-tests/src/utils.ts
@@ -36,9 +36,3 @@ export function createApp<
         baseUrl: BASE_URL,
     });
 }
-
-export function clearStorage() {
-    if (window.localStorage) {
-        localStorage.clear();
-    }
-}
--- a/packages/realm-web/CHANGELOG.md
+++ b/packages/realm-web/CHANGELOG.md
@@ -1,3 +1,15 @@
+?.?.? Release notes (2020-??-??)
+=============================================================
+
+### Enhancements
+* Users are now persisted across refreshes and browser tabs (using the browsers local-storage). ([#2990](https://github.com/realm/realm-js/pull/2990))
+
+### Fixed
+* None.
+
+### Internal
+* None
+
 0.4.0 Release notes (2020-6-11)
 =============================================================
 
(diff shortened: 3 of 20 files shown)


packages/realm-web-integration-tests/src/app.test.ts
```diff
@@ -67,4 +71,27 @@ describe("App#constructor", () => {
...
     });
+
+    it("restores a user", async () => {
+        let user: User<object>;
+        {
+            const app = createApp();
+            const credentials = Credentials.anonymous();
+            user = await app.logIn(credentials);
+            expect(typeof user.id).equals("string");
+            console.log(window.localStorage);
```
kneth, line 82: Left-over from debugging?

packages/realm-web-integration-tests/src/utils.ts
```diff
@@ -34,3 +36,9 @@ export function createApp<FunctionsFactoryType extends object>() {
         baseUrl: BASE_URL,
     });
 }
+
+export function clearStorage() {
+    if (window.localStorage) {
+        localStorage.clear();
```
kneth, line 42: Why not `window.localStorage.clear()`?

packages/realm-web/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+?.?.? Release notes (2020-??-??)
+=============================================================
+
+### Enhancements
+* Users are now persisted across refreshes and browser tabs (using the browsers local-storage). ([#2990](https://github.com/realm/realm-js/pull/2990))
```
kneth, line 5:
```suggestion
* Users are now persisted across refreshes and browser tabs (using the browser's local-storage). ([#2990](https://github.com/realm/realm-js/pull/2990))
```

packages/realm-web/src/UserStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+     * @returns Refresh token.
+     */
+    get refreshToken() {
+        return this.get(REFRESH_TOKEN_STORAGE_KEY);
+    }
+
+    /**
+     * Set the refresh token in storage.
+     *
+     * @param value Refresh token.
```
kneth, line 74: `null` is not documented

packages/realm-web/src/UserStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+     * @returns Access token.
+     */
+    get accessToken() {
+        return this.get(ACCESS_TOKEN_STORAGE_KEY);
+    }
+
+    /**
+     * Set the access token in storage.
+     *
+     * @param value Access token.
```
kneth, line 52: `null` is not documented

packages/realm-web/src/UserStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+     * @returns User profile.
+     */
+    get profile() {
+        const value = this.get(PROFILE_STORAGE_KEY);
+        if (value) {
+            const profile = new UserProfile();
+            // Patch in the values
+            Object.assign(profile, JSON.parse(value));
+            return profile;
+        }
```
kneth, line 96: Explicit return here for readability?

packages/realm-web/src/storage/PrefixedStorage.ts
```diff
@@ -0,0 +1,65 @@
...
+     * @param keyPart The part of the key to prefix when performing operations.
+     */
+    constructor(storage: Storage, keyPart: string) {
+        this.storage = storage;
+        this.keyPart = keyPart;
+    }
+
+    /** @inheritdoc */
+    public get(key: string) {
+        return this.storage.get(this.keyPart + ":" + key);
```
kneth, line 48: Template string for consistent style?

packages/realm-web-integration-tests/src/app.test.ts
kraenhansen: Yep - thanks.

packages/realm-web-integration-tests/src/utils.ts
kraenhansen: Because `window` is the global in this context.
But this code has another issue. Since it should also be able to run from Node.js, getting `localStorage` on `window` (which will be `undefined`) throws. I've refactored the Storage instance to also have a clear() method on it.

packages/realm-web/src/storage/PrefixedStorage.ts
kraenhansen: I agree.
I refactored this to use a constant and will therefore stick with good ol' string concatenation.

packages/realm-web/src/UserStorage.ts
kraenhansen: I would like to leave it at this 👍

packages/realm-web-integration-tests/src/utils.ts
```diff
@@ -34,3 +36,9 @@ export function createApp<FunctionsFactoryType extends object>() {
         baseUrl: BASE_URL,
     });
 }
+
+export function clearStorage() {
+    if (window.localStorage) {
```
steffenagger: in other placed `window` is also checked.

packages/realm-web/src/storage/PrefixedStorage.ts
steffenagger: I'm not sure I follow, how does a constant affect this?
`this.keyPart + ":" + key` vs. `` `${this.keyPart}:${key}` ``

packages/realm-web/src/User.ts
```diff
@@ -107,26 +109,54 @@ export class User<
...
+    }
+
     private _id: string;
     private _accessToken: string | null;
     private _refreshToken: string | null;
-    private _profile: Realm.UserProfile | undefined;
-    private _state: Realm.UserState;
+    private _profile: UserProfile | undefined;
     private transport: AuthenticatedTransport;
+    private storage: UserStorage;
```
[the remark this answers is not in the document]
kraenhansen: It's only the private fields that also have public getters.

packages/realm-web/src/storage/PrefixedStorage.ts
kraenhansen: With the latest commit, this is now `this.keyPart + PrefixedStorage.PART_SEPARATOR + prefix` which I believe looks cleaner without the template literals.

packages/realm-web/src/App.test.ts
```diff
@@ -282,4 +292,163 @@ describe("App", () => {
         expect(app.services).keys(["mongodb", "http"]);
         expect(typeof app.services.mongodb).equals("function");
     });
+
+    it("hydrates users from storage", () => {
+        const storage = new MemoryStorage();
+        const transport = new MockNetworkTransport([]);
+
+        // Fill data into the storage that can be hydrated
+        const appStorage = storage.prefix("app(default-app-id)");
```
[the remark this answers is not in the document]
kraenhansen: Perhaps `prefixed` would be a better name?

packages/realm-web-integration-tests/src/utils.ts
kraenhansen: Agree - this was a bug. It's removed now.

packages/realm-web/src/App.test.ts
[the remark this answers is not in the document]
kraenhansen: Or perhaps `extend` or `cloneWithPrefix`?

Although - the docs clearly states that it will:
> Create a new store prefixed with a part of the key.

packages/realm-web/src/App.test.ts
```diff
@@ -282,4 +292,163 @@ describe("App", () => {
...
+        expect(alice.id).equals("alices-id");
+        expect(alice.accessToken).equals("alices-access-token");
+        expect(alice.refreshToken).equals("alices-refresh-token");
+        expect(alice.profile.firstName).equals("Alice");
+
+        const bob = app.allUsers[1];
+        expect(bob.id).equals("bobs-id");
+        expect(bob.accessToken).equals("bobs-access-token");
+        expect(bob.refreshToken).equals("bobs-refresh-token");
+    });
```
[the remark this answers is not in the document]
kraenhansen: Actually - its not optional, but it would throw if requested, because we explicitly ask for it not to fetch the profile when we login (second argument of the `logIn` method).

packages/realm-web-integration-tests/src/app.test.ts
```diff
@@ -67,4 +73,26 @@ describe("App#constructor", () => {
...
+            const app = createApp();
+            const credentials = Credentials.anonymous();
+            user = await app.logIn(credentials);
+            expect(typeof user.id).equals("string");
+        }
+        // Recreate the app and expect the user to be restored
+        {
+            const app = createApp();
+            expect(app.allUsers.length).equals(1);
+            expect(app.currentUser).instanceOf(User);
```
steffenagger, line 89: There's only an `instanceOf` check when it's been restored.

packages/realm-web/src/App.test.ts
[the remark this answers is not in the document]
steffenagger: You're right, I just commented as the first thing I thought it did, was not the case.
Normally I would also suggest another name, but couldn't really come up with a better one...

packages/realm-web/src/App.test.ts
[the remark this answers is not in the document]
steffenagger: Fair enough, do we test that it throws?

packages/realm-web/src/User.ts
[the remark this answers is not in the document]
steffenagger: Ok, but might be miss leading if used as good ol' JS.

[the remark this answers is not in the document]
kraenhansen: Agree - it's only used for the private and internal parts of the API. All written in and consumed from TypeScript.

packages/realm-web/src/App.test.ts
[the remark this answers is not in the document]
kraenhansen: Yep 🤞  https://github.com/realm/realm-js/blob/kh/web/persist-users/packages/realm-web/src/User.test.ts#L82-L84

PR #2994, opened 2020-06-22
Realm Web: Refactored `AuthenticatedTransport`

kraenhansen:
## What, How & Why?

This simplifies the `Transport` interface, by removing the optional `user` argument and instead returning an `AuthenticatedTransport` when prefixing a `AuthenticatedTransport`. This change is purely internal and has no effects on the API users are consuming.

Note: This PR should be rebased onto `v10` once the parent branch (`kh/web/persist-users`) gets merged.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web/src/transports/AuthenticatedTransport.test.ts
+++ b/packages/realm-web/src/transports/AuthenticatedTransport.test.ts
@@ -84,4 +84,30 @@ describe("AuthenticatedTransport", () => {
             },
         ]);
     });
+
+    it("returns an AuthenticatedTransport when prefixed", async () => {
+        const baseTransport = new MockTransport([{}]);
+        const transport = new AuthenticatedTransport(baseTransport, {
+            currentUser: { accessToken: "my-access-token" } as Realm.User,
+        });
+        const prefixedTransport = transport.prefix("/prefixed-path");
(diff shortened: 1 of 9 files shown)


packages/realm-web/src/transports/AuthenticatedTransport.test.ts
```diff
@@ -84,4 +84,30 @@ describe("AuthenticatedTransport", () => {
...
+            path: "/w00t",
+        });
+        // Expect something of the request
+        expect(baseTransport.requests).deep.equals([
+            {
+                method: "GET",
+                url: "http://localhost:1337/prefixed-path/w00t",
+                headers: {
+                    Authorization: "Bearer my-access-token",
+                    Accept: "application/json",
```
steffenagger, line 107: You could spread `DEFAULT_HEADERS` here, so changes wont break this testing that they gets set. (and only checks that the `Authorization` header is added)

kraenhansen: I think something like https://github.com/realm/realm-js/blob/kh/web/fetch-location-metadata/packages/realm-web/src/App.test.ts#L212 covers this.

PR #2995, opened 2020-06-22
Implement Realm.User.push()

RedBeard0531:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~~[ ] 📝 `Compatibility` label is updated or copied from previous entry~~
* [X] 🚦 Tests
* ~~[ ] 📝 Public documentation PR created or is not necessary~~
* ~~[ ] 💥 `Breaking` label has been applied or is not necessary~~

* [X] jsdoc files updated
* [X] Chrome debug API is updated if API is available on React Native

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ NOTE: This version bumps the Realm file format to version 11. It is not possible
 
 ### Enhancements
 * Added RemoteMongoClient functionality to `Realm.User`
+* Added Push functionality to `Realm.User`
 
 ### Fixed
 * Added missing `SyncConfiguration.error` field in the typescript definitions.
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,0 +1,47 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2020 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+"use strict";
+
+
+module.exports = {
+    /**
+     * Helper to wrap callback-taking C++ function into a Promise-returning JS function
+     * @example
+     * // floop() is a wrapper method on a type with a _floop C++ method.
+     * function floop(how, why) {
+     *   return promisify(cb => this._floop(how, why, cb));
+     * }
+     */
+    promisify(func) {
+        return new Promise((resolve, reject) => {
+            func((...cbargs) => {
+                if (cbargs.length < 1 || cbargs.length > 2)
+                    throw Error(`invalid cbargs length ${cbargs.length}`)
+                let error = cbargs[cbargs.length-1];
+                if (error) {
+                    reject(error);
+                } else if (cbargs.length == 2) {
+                    resolve(cbargs[0]);
+                } else {
+                    resolve();
+                }
+            });
+        });
+    },
+}
--- a/src/js_app_credentials.hpp
+++ b/src/js_app_credentials.hpp
@@ -140,8 +140,11 @@ template<typename T>
 void CredentialsClass<T>::function(ContextType ctx, ObjectType this_object, Arguments& arguments, ReturnValue& return_value) {
     arguments.validate_count(1);
     const std::string payload_json = Value::validated_to_string(ctx, arguments[0], "payload");
+    const auto payload_bson = bson::parse(payload_json);
+    if (payload_bson.type() != bson::Bson::Type::Document) 
+        throw std::invalid_argument("payload must be a json object");
 
-    auto credentials = realm::app::AppCredentials::function(payload_json);
+    auto credentials = realm::app::AppCredentials::function(payload_bson.operator const bson::BsonDocument&());
     return_value.set(create_object<T, CredentialsClass<T>>(ctx, new app::AppCredentials(credentials)));
 }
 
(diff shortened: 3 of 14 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ NOTE: This version bumps the Realm file format to version 11. It is not possible
 
 ### Enhancements
 * Added RemoteMongoClient functionality to `Realm.User`
+* Added Push functionality to `Realm.User`
```
bmunkholm, line 9: Perhaps we can elaborate a bit more what this is (in end user terms). (I can see we also missed to do that with RemoteMongoClient.

src/js_user.hpp
```diff
@@ -303,5 +307,54 @@ void UserClass<T>::refresh_custom_data(ContextType ctx, ObjectType this_object,
...
+                !error ? Value::from_undefined(ctx) : Object::create_obj(ctx, {
+                    {"message", Value::from_string(ctx, error->message)},
+                    {"code", Value::from_number(ctx, error->error_code.value())},
+                }),
+            });
+        }));
+}
+
+template<typename T>
+void UserClass<T>::push_deregister(ContextType ctx, ObjectType this_object, Arguments& args, ReturnValue &return_value) {
```
bmunkholm, line 336: What's the difference to to "register" function above? They look identical.

RedBeard0531: 😬 Yikes! That would be a bug. It is supposed to be calling `deregister_device()` on like 343.

CHANGELOG.md
RedBeard0531: Unfortunately, I don't actually know what "Push" does in end user terms. This sounds like a good thing to have our Docs team work on so that they can ensure that release notes use consistent terminology with the rest of the documentation.

kraenhansen:
I don't have a good suggestion on how to resolve this, but since Realm Web does not currently have an implementation of the push service we would have an issue if Realm Web gets released with these types.

In principle we should strive to have feature parity between platforms, but I remember having a discussion with @jsflax about the (lack of) need for this service in Realm Web (since it runs in the more ephemeral browser, than an app installed on the users phone):
- If we choose to implement the push service for Realm Web, we could hold back this PR and merge the implementation in Realm Web into this branch before it gets merged to `v10`.
- If we choose not to implement the push service for Realm Web, we need to ensure this property on the `User` is properly merged with https://github.com/realm/realm-js/pull/2950 such that it will only be exported from Realm JS. And we should probably update the JSdocs to specify that this is not supported in "Realm Web".

src/js_user.hpp
bmunkholm: So seems like the tests needs an update as well then ;-)
Isn't it possible to reuse some of that code so it's clear where there is supposed to be a difference :-)

RedBeard0531: The API for `deregisterDevice` changed upsteam in object-store, so I'll add an update to the submodule to this PR so we use the new API https://github.com/realm/realm-object-store/commit/1f3290a840a114cc21df78f2fb5fdc722fe1cd7d#diff-b215f5f5ce1cec59d075f06fbab4510e

> Isn't it possible to reuse some of that code so it's clear where there is supposed to be a difference

Probably. I have some ideas on refactors that can clean these repetitive methods up a bit, but so far I've only done the `promisify()` part on the JS side. I'll probably make a separate PR to clean up the CPP side at some point.

src/js_app_credentials.hpp
```diff
@@ -140,8 +140,11 @@ template<typename T>
 void CredentialsClass<T>::function(ContextType ctx, ObjectType this_object, Arguments& arguments, ReturnValue& return_value) {
     arguments.validate_count(1);
     const std::string payload_json = Value::validated_to_string(ctx, arguments[0], "payload");
+    const auto payload_bson = bson::parse(payload_json);
+    if (payload_bson.type() != bson::Bson::Type::Document) 
+        throw std::invalid_argument("payload must be a json object");
 
-    auto credentials = realm::app::AppCredentials::function(payload_json);
+    auto credentials = realm::app::AppCredentials::function(payload_bson.operator const bson::BsonDocument&());
```
RedBeard0531, line 147: Upstream object-store changed the parameter type.

lib/utils.js
```diff
@@ -0,0 +1,47 @@
...
+     * @example
+     * // floop() is a wrapper method on a type with a _floop C++ method.
+     * function floop(how, why) {
+     *   return promisify(cb => this._floop(how, why, cb));
+     * }
+     */
+    promisify(func) {
+        return new Promise((resolve, reject) => {
+            func((...cbargs) => {
+                if (cbargs.length < 1 || cbargs.length > 2)
```
kneth, line 34: So only 1 or 2 arguments? Maybe `[1, 2].includes(cbargs.length)`?

lib/utils.js
```diff
@@ -0,0 +1,47 @@
...
+module.exports = {
+    /**
+     * Helper to wrap callback-taking C++ function into a Promise-returning JS function
+     * @example
+     * // floop() is a wrapper method on a type with a _floop C++ method.
+     * function floop(how, why) {
+     *   return promisify(cb => this._floop(how, why, cb));
+     * }
+     */
+    promisify(func) {
```
kneth, line 31: Nice simplification!!

src/js_network_transport.hpp
```diff
@@ -216,7 +216,7 @@ struct JavaScriptNetworkTransport : public JavaScriptNetworkTransportWrapper<T>
             case app::HttpMethod::get:   return "GET";
             case app::HttpMethod::put:   return "PUT";
             case app::HttpMethod::post:  return "POST";
-            case app::HttpMethod::del:   return "DEL";
+            case app::HttpMethod::del:   return "DELETE";
```
kneth, line 219: Oops, my mistake - sorry.

types/index.d.ts
```diff
@@ -371,6 +371,10 @@ declare namespace Realm {
         linkCredentials(credentials: Credentials): Promise<void>;
         callFunction(name: string, args: any[]): Promise<any>;
         refreshCustomData(): Promise<Object>;
+        push(serviceName: string): {
```
kraenhansen, line 374: @RedBeard0531 Going over this code when merging it with my changes (kh/v10-merged-types) I realize that we might want to have this `serviceName` be optional and provide a default name.

PR #3004, opened 2020-06-26
Add RemoteMongoDBCollection.watch()

RedBeard0531:
## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


lib/mongo_client.js
```diff
@@ -206,6 +207,32 @@ class RemoteMongoDBCollection {
...
+    async* watch({ids = undefined, filter = undefined} = {}) {
+        let args = {
+            database: this.databaseName,
+            collection: this.collectionName,
+            ids,
+            filter,
+        };
+
+        const request = this.user._makeStreamingRequest("watch", this.serviceName, this.user._cleanArgs([args]));
+        const reply = await (new DefaultNetworkTransport().fetch(request));
```
kraenhansen, line 220: The network transport is ment to be reused throughout the lifetime of the runtime. Do you see a way to avoid instantiating it upon every call to this function?

lib/mongo_client.js
```diff
@@ -206,6 +207,32 @@ class RemoteMongoDBCollection {
             upsert: options.upsert,
         });
     }
+
+    async* watch({ids = undefined, filter = undefined} = {}) {
```
kraenhansen, line 211: This is the first generator function in Realm JS. Is there a different way of achieving this without introducing the use of a new ES feature to the users? I'm thinking ex the observer pattern (event emitters) are used extensively throughout the API.

lib/mongo_client.js
RedBeard0531: I could, but isn't it really cheap to construct? It looks like after the first call of the constructor, all other constructor calls don't really do anything.

lib/mongo_client.js
RedBeard0531: I was specifically *avoiding* using an event emitter-style API here. They have the significant problem with all push-based APIs that the consumer has no way of controlling when they are ready for the next event, and therefore no way of providing backpressure. Using async iterables/iterators (whether via a generator or directly implementing the `@@asyncIterator` protocol) uses a pull-style API that puts the client in control of when they want the next event. This lets you do things like this, and ensure that events are processed in order:

```js
for await (let event of collection.watch()) {
    await notifyServerOfEvent(event);
}
```

That is not possible to implement correctly if we were using the event emitter pattern. This is a general case of pull-style iteration being more flexible and generally preferable to push-style. Note that it is trivial to wrap a pull-style iterator into a push-style one, but impossible to go back. That means that if we only provided a emitter API, developers *wouldn't be able* to use `for await`-style iteration. If they want to use event emitters, this is all it takes:

```js
function onWatchEvent(watch, handler) {
  (async() => {
      for await (let event of watch)) handler(event);
  )().onError(e => handler(undefined, e));
}```

We could of course offer that helper as a separate function, but I'm not sure when you would ever prefer to use that rather than just directly consuming the async iterator.

File: lib/user.js [hunk 3]
```diff
@@ -77,6 +65,22 @@ const instanceMethods = {
         };
     },
 
+    get functions() {
```
RedBeard0531, line 68: These are unmodified, just moved to be above the internal helpers.

src/js_network_transport.hpp
```diff
@@ -180,30 +180,28 @@ struct JavaScriptNetworkTransport : public JavaScriptNetworkTransportWrapper<T>
...
-        // Create a JavaScript version of the request
-        ObjectType request_object = Object::create_empty(m_ctx);
-        Object::set_property(m_ctx, request_object, "method", Value::from_string(m_ctx, fromHttpMethod(request.method)));
-        Object::set_property(m_ctx, request_object, "url", Value::from_string(m_ctx, request.url));
-        Object::set_property(m_ctx, request_object, "timeoutMs", Value::from_number(m_ctx, request.timeout_ms));
-        if (!request.body.empty()) {
-            Object::set_property(m_ctx, request_object, "body", Value::from_string(m_ctx, request.body));
-        }
-        ObjectType headers_object = Object::create_empty(m_ctx);
+    static ObjectType makeRequest(ContextType ctx, const app::Request& request) {
```
RedBeard0531, line 183: This was extracted and cleaned up from the `send_request_to_server` method, so that it could be reused for `watch()` requests which just need to be able to get the POJsO version of the `app::Request`.

RedBeard0531:
Moving out of Draft status.

There is likely an issue on RN, since there are [multiple](https://github.com/facebook/react-native/issues/9629) [bugs](https://github.com/facebook/react-native/issues/12912] around their implementation of the `fetch()` API being broken for streaming usages. Discussed with @kneth and decided the best action is to merge without proper RN support and then we can fill in a working networking implementation for RN in a later commit.

The RN debug hooks are not complete, and I will probably need some help getting the new `WatchStream` type exposed, but since RN won't be supported by this commit anyway, I don't think that needs to be a blocker.

types/app.d.ts
```diff
@@ -26,6 +26,8 @@ declare namespace Realm {
     // See https://stackoverflow.com/a/51114250 on why we're importing the BSON types like this
     type ObjectId = import("bson").ObjectId;
     type Binary = import("bson").Binary;
+    type Long = import("bson").Binary;
```
kneth, line 29: Why import `Binary` three times under different names?

RedBeard0531: Typo in last commit. Fixed by https://github.com/realm/realm-js/pull/3004/commits/76aae2b4718388600b898ec8d13202119db8089e

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 NOTE: Support for syncing with realm.cloud.io and/or Realm Object Server has been replaced with support for syncing with MongoDB Realm Cloud.
 
-NOTE: This version bumps the Realm file format to version 11. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v1.0.0, might not be upgradeable. Only [Realm Studio 10.0.0](https://github.com/realm/realm-studio/releases/tag/v10.0.0-beta.1) or later will be able to open the new file format.
+NOTE: This version bumps the Realm file format to version 11. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v10.0.0, might not be upgradeable. Only [Realm Studio 10.0.0](https://github.com/realm/realm-studio/releases/tag/v10.0.0-beta.1) or later will be able to open the new file format.
```
bmunkholm, line 5: @RedBeard0531 that fix is not correct!

bmunkholm: @kneth

kneth: True, it should say v1.0.0 (= Realm Core v2.3.0 = file format v6).

RedBeard0531: Sorry about that. Fixed in https://github.com/realm/realm-js/commit/c877b02054d450d85f76c3c046bac7c65cd7ef58.

Our version numbers are really confusing...

PR #3013, opened 2020-06-30
JSON serialization of Realm.Object & Realm.Collection

steffenagger:
#### TS declarations (used in integration-tests)
- [x] Limit input in `create<T>(...)` to match a model (using @nirinchev's `ExtractPropertyNamesOfType`©) **notion of any implications this could introduce, is welcome!**.
- [x] Allow input to contain `Array<T>` in `RealmPartialModel<T>` instead of a defined `Realm.Collection<T>` in `create<T>(...)`.
- [x] Update `objects<T>(...)` to accommodate the underlying logic & different responses.
- [x] Allow extending `Realm.Object` for class models.

#### `toJSON` on `Realm.Object` & `Realm.Collection` (this includes `Realm.List` & `Realm.Results`)
- [x] Implement solution with object references (this introduces a cache to not return new objects).
- [x] Ensure `TypeError: Converting circular structure to JSON` is thrown for circular structures, not `RangeError: Maximum call stack size exceeded`.
- [x] Implement & expose replacer for circular structures. \*

\* _I've included a combined key of the table-type & `_objectId()` as unique `$refId` in `Realm.JsonSerializationReplacer`, for a (visual) reference in serialized JSON. This is only used in case a primaryKey is not defined._

#### Integration Tests
- [x] Write test covering `toJSON` implementations for both interface/schema & class model usage.

#### Changelog
- [x] Update `CHANGELOG.md`

### DIFF
--- a/integration-tests/tests/src/serialization.ts
+++ b/integration-tests/tests/src/serialization.ts
@@ -1,3 +1,21 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2020 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
 import { expect } from "chai";
 
 import {
--- a/integration-tests/tests/src/serialization.ts
+++ b/integration-tests/tests/src/serialization.ts
@@ -1,0 +1,89 @@
+import { expect } from "chai";
+
+import { IPerson, PersonSchema } from "./schemas/person-and-dogs";
+import * as expectedCircularStructure from "./structures/circular-result.json"
+
+describe("Serialization", () => {
+    it("JsonSerializationReplacer is exposed on the Realm constructor", () => {
+        expect(typeof Realm.JsonSerializationReplacer).equal('function');
+        expect(Realm.JsonSerializationReplacer.length).equal(2);
+    })
+
+    it("Realm.Object has 'toJSON' implemented", () => {
+        const realm = new Realm({ schema: [PersonSchema] });
+
+        let john!: IPerson & Realm.Object;
+
+        realm.write(() => {
+            john = realm.create<IPerson>("Person", {
+                name: "John Doe",
+                age: 42
+            });
+
+            john.friends.push(john)
+        });
+
+        expect(typeof john.toJSON).equals('function');
+        expect(john.toJSON().objectSchema).equals(undefined);
+        expect(john.toJSON().friends).instanceOf(Array);
+    });
+
+    it("Realm.Results (Collection) has 'toJSON' implemented", () => {
+        const realm = new Realm({ schema: [PersonSchema] });
+
+        let john!: IPerson & Realm.Object;
+
+        realm.write(() => {
+            john = realm.create<IPerson>("Person", {
+                name: "John Doe",
+                age: 42
+            });
+        });
+
+        const persons = realm.objects<IPerson>("Person");
+
+        expect(typeof persons.toJSON).equals('function');
+        expect(persons.toJSON()).instanceOf(Array);
+
+        const first = persons.toJSON()[0];
+        expect(first.objectSchema).equals(undefined);
+        expect(first.friends).instanceOf(Array);
+    });
+
+    it("Realm.Results can be serialized with circular references", () => {
+        const realm = new Realm({ schema: [PersonSchema] });
+
+        let john!: IPerson & Realm.Object;
+        let jane!: IPerson & Realm.Object;
+        let tony!: IPerson & Realm.Object;
+
+        realm.write(() => {
+            john = realm.create<IPerson>("Person", {
+                name: "John Doe",
+                age: 42
+            });
+            jane = realm.create<IPerson>("Person", {
+                name: "Jane Doe",
+                age: 40
+            });
+            tony = realm.create<IPerson>("Person", {
+                name: "Tony Doe",
+                age: 35
+            });
+
+            // ensure circular references
+            john.friends.push(jane);
+            john.friends.push(tony);
+
+            jane.friends.push(tony);
+            jane.friends.push(john);
+            
+            tony.friends.push(jane);
+        });
+
+        const persons = realm.objects<IPerson>("Person");
+
+        expect(JSON.stringify(persons, Realm.JsonSerializationReplacer))
+            .equal(JSON.stringify(expectedCircularStructure));
+    });
+});
--- a/integration-tests/tests/src/serialization.ts
+++ b/integration-tests/tests/src/serialization.ts
@@ -1,23 +1,50 @@
 import { expect } from "chai";
 
-import { IPerson, PersonSchema, DogSchema, Person, Dog } from "./schemas/person-and-dogs";
-import * as circularCollectionResult from "./structures/circular-collection-result.json"
+import {
+    Dog as DogWithId,
+    DogSchema as DogSchemaWithId,
+    IPerson as IPersonWithId,
+    Person as PersonWithId,
+    PersonSchema as PersonSchemaWithId
+} from "./schemas/person-and-dog-with-primary-ids";
+import {
+    Dog,
+    DogSchema,
+    IPerson,
+    Person,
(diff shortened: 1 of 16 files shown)


lib/extensions.js
```diff
@@ -111,37 +111,39 @@ module.exports = function(realmConstructor, context) {
...
+            const result = { $realmId: id };
+            cache.set(id, result);
+
+            const moveAsSerializable = (cache, key, from, to) => {
+                // TODO: check that we're dealing with a Realm type, before passing in cache?
+                to[key] = from[key].toJSON ? from[key].toJSON(undefined, cache) : this[key];
+            };
+
+            Object.keys(this).forEach(k => moveAsSerializable(cache, k, this, result));
+            Object.keys(Object.getPrototypeOf(this)).forEach(k => moveAsSerializable(cache, k, this, result));
```
kraenhansen, line 144: Probably nit picking here, but it would make it a bit more DRY.
```suggestion
            Object.keys(this)
                .concat(Object.keys(Object.getPrototypeOf(this)));
                .forEach(k => moveAsSerializable(cache, k, this, result));
```

lib/extensions.js
```diff
@@ -776,4 +778,34 @@ module.exports = function(realmConstructor, context) {
...
+            return function(_, value) {
+                if (typeof value !== "object") {
+                    return value;
+                }
+
+                const parent = this;
+                if (seen.length > 0) {
+                    const pos = seen.indexOf(parent);
+                    if (pos !== -1) {
+                        seen.splice(pos + 1)
```
kraenhansen, line 795:
```suggestion
                        seen.splice(pos + 1);
```

lib/extensions.js
```diff
@@ -776,4 +778,34 @@ module.exports = function(realmConstructor, context) {
...
+                    } else {
+                        seen.push(parent)
```
kraenhansen, line 797:
```suggestion
                        seen.push(parent);
```

lib/extensions.js
```diff
@@ -776,4 +778,34 @@ module.exports = function(realmConstructor, context) {
...
+                    }
+
+                    if (seen.includes(value)) {
+                        return value.$realmId ? { $ref: value.$realmId } : "[Circular reference]";
+                    }
+                } else {
+                    seen.push(value)
```
kraenhansen, line 804:
```suggestion
                    seen.push(value);
```

😉

lib/extensions.js
kraenhansen, line 797:
```suggestion
                        seen.push(parent);
```

lib/extensions.js
kraenhansen, line 804:
```suggestion
                    seen.push(value);
```

lib/index.d.ts
```diff
@@ -817,6 +828,14 @@ interface NamedSubscription {
     timeToLive: number;
 }
 
+type ExtractPropertyNamesOfType<T, S> = {
+    [K in keyof T]: T[K] extends S ? K : never
+}[keyof T];
```
kraenhansen, line 833: I for one (obviously 🤔 ) know what it does - but for future maintainers of the code, we might need a small explanation of this piece of black magic.

lib/index.d.ts
```diff
@@ -817,6 +828,14 @@ interface NamedSubscription {
...
+
+type RealmPartialModel<T> =
```
kraenhansen, line 835: I don't know if `Model` is the right word here - we're already using `ObjectSchema` and `Class` elsewhere.

nirinchev, line 835: I expect to see license fees once that gets merged ;p

steffenagger: Of course! And I imagine they'll just go directly out of my salary... 😁

lib/extensions.js
```diff
@@ -776,4 +778,34 @@ module.exports = function(realmConstructor, context) {
...
+                    } else {
+                        seen.push(parent);
+                    }
+
+                    if (seen.includes(value)) {
+                        return value.$realmId ? { $ref: value.$realmId } : "[Circular reference]";
+                    }
+                } else {
+                    seen.push(value);
+                }
```
nirinchev, line 805: This section of the code definitely needs comments - I'm having a very hard time wrapping my head around it and confirming it's behaving as expected, especially considering it's modifying some external array.

integration-tests/tests/src/serialization.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        const persons = realm.objects<IPerson>("Person");
+
+        expect(typeof persons.toJSON).equals('function');
+        expect(persons.toJSON()).instanceOf(Array);
+
+        const first = persons.toJSON()[0];
+        expect(first.objectSchema).equals(undefined);
+        expect(first.friends).instanceOf(Array);
+    });
```
kraenhansen, line 51: It would be great with a text which explicitly tests the behaviour `toJSON` with circular references.
Specifically I would expect `john` to be the same object (by reference) if he is befriended with himself (or via other person).

lib/extensions.js
```diff
@@ -111,37 +111,39 @@ module.exports = function(realmConstructor, context) {
...
+            if (existing) {
+                return existing;
             }
 
-            return Object.entries(result);
+            // create new result, and store in cache
+            const result = { $realmId: id };
+            cache.set(id, result);
+
+            const moveAsSerializable = (cache, key, from, to) => {
```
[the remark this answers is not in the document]
steffenagger: True. Those were moved, thanks.

lib/extensions.js
steffenagger: Yes definitely, it's on the todo - maybe the draft/WIP was opened a bit early, just to show the overall approach.

lib/extensions.js
steffenagger: I think we should add `"semi": [1, "always"]` to the eslint config then.

lib/extensions.js
nirinchev: The PR description was longer than the PR itself, so of course I completely skipped it 😅

lib/extensions.js
steffenagger: I like it. It also opens for having `moveAsSerializable` back in as an anonymous function.

lib/extensions.js
```diff
@@ -111,37 +111,39 @@ module.exports = function(realmConstructor, context) {
...
-            return Object.keys(this).concat(Object.keys(Object.getPrototypeOf(this)));
-        },
-
-        writable: true,
-        configurable: true,
-        enumerable: false
-    });
+    Object.defineProperty(realmConstructor.Object.prototype, "toJSON", {
+        value: function (key, cache = new Map()) {
+            const id = this._objectId();
```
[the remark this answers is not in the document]
steffenagger: Oh... good to know, and back to the drawing board. Think I have a solution though.

lib/extensions.js
```diff
@@ -111,37 +111,39 @@ module.exports = function(realmConstructor, context) {
...
-        value: function () {
-            return Object.keys(this).concat(Object.keys(Object.getPrototypeOf(this)));
-        },
-
-        writable: true,
-        configurable: true,
-        enumerable: false
-    });
+    Object.defineProperty(realmConstructor.Object.prototype, "toJSON", {
+        value: function (key, cache = new Map()) {
```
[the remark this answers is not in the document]
steffenagger: It's not.

lib/extensions.js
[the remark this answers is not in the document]
steffenagger: Wouldn't the new `const id = this.constructor.name + this._objectId();` fix this?

integration-tests/tests/src/serialization.ts
steffenagger: included in the updated tests.

lib/extensions.js
steffenagger: Added comments.

blagoev:
Is this a draft still

lib/index.d.ts
steffenagger: Comments added

blagoev (approved):

lib/index.d.ts
steffenagger: This has now changed name to `RealmInsertionModel` I personally think it's quite descriptive and a commonly used name. I don't the either Schema or Class fits here, but could use `RealmInsertionObject`?
This naming is already in v10, so I suggest we revisit the naming if it's confusing to users?

integration-tests/tests/src/serialization.ts
```diff
@@ -0,0 +1,288 @@
+import { expect } from "chai";
```
kneth, line 1: Please add copyright header.

kneth (approved):

KalAvroniev:
Still getting `RangeError: Maximum call stack size exceeded`.
It happens on records with this schema when using toJSON
`{
  "name": "tag",
  "primaryKey": "uuid",
  "properties": {
    "tags": {
      "type": "list",
      "objectType": "tag"
    }
  }
}`

Issue #2967, opened 2020-06-15
Realm Web: Retrieve and refresh a users custom data


kraenhansen:
Users of MongoDB Realm can specify custom data (stored in the JWT access token), we need to expose this on the `User` and allow for this to be refreshed (by refreshing the token).

Issue #2965, opened 2020-06-15
Realm Web: Refresh access tokens


kraenhansen:
Refresh access token:
- [x] when requests fail with a 4xx response code.
- [ ] ahead of time, by registering a timer and clearing this when it's no longer needed (upon logout).

PR #3019, opened 2020-07-01
Realm Web: More credentials

### LINKED ISSUE #2963: Realm Web: Implement missing credentials (Facebook, Google, JWT, Apple, Functions)
Implement more credentials for authentication providers:
  * [ ] Facebook
  * [ ] Google
  * [ ] Custom JWT
  * [ ] Apple Sign In
  * [ ] Functions

kraenhansen:
## What, How & Why?

This closes #2963 by adding implementation and integration tests for all (except Apple ID).
This also closes the retrieving part of #2967 (refreshing will be solved with #2965).

Note: This should be rebased on `v10` once the current base has been merged.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (except for `User#customData` and Apple ID credentials)

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-web-integration-tests/README.md
+++ b/packages/realm-web-integration-tests/README.md
@@ -1,9 +1,66 @@
 # Realm Web integration tests
 
+For ultimate flexability, the tests are written to be able to run from both a Node.js and a browser.
+
 ## Running the tests
 
 Ensure you have access to a running instance of MongoDB Realm.
 
 To run it locally, ensure you have Docker and your Realm AWS credentials installed and run
 
-    docker run --rm -i -t --publish 9090:9090 012067661104.dkr.ecr.eu-west-1.amazonaws.com/ci/mongodb-realm-images:test_server-0ed2349a36352666402d0fb2e8763ac67731768c-race
+```
+docker run --rm -i -t --publish 9090:9090 012067661104.dkr.ecr.eu-west-1.amazonaws.com/ci/mongodb-realm-images:test_server-0ed2349a36352666402d0fb2e8763ac67731768c-race
+```
+
+## Running the tests from Node.js
+
+To enable a short feedback loop the integration tests can be run from Node.js (via the Mocha CLI):
+
+```
+npm start
+```
+
+This will start the tests with Mocha in --watch mode.
+To narrow in and run a single test, use the --grep runtime argument:
+
+```
+npm start -- --grep Credentials
+```
+
+## Running the tests from a browser
+
+To start the harness, opening a headless Chromium browser (via Puppeteer), run:
+
+```
+npm test
+```
+
+To enable a long timeout and open developer tools, run this with the DEV_TOOLS environment variable.
+
+```
+DEV_TOOLS=1 npm test
+```
+
+## Running the OAuth integration tests
+
+Currently, these are only available when running in a browser.
+
+Before running tests which depends on a specific authentication provider:
+1. Run once to import the app,
+2. Go to http://localhost:9090 (or whereever your MongoDB Realm server is listening)
+3. Setup the appropreate authentication providers.
+
+To enable testing credentials that require interaction, run with DEV_TOOLS and TEST_CREDENTIALS set to a comma-separated list of credentials.
+
+```
+DEV_TOOLS=1 TEST_CREDENTIALS=anonymous,email-password,google npm test
+```
+
+## Additional environment variables
+
+To skip importing the app or use a different server, specify one or more of the following environment variables:
+
+- `MDB_REALM_APP_ID`
+- `MDB_REALM_BASE_URL`
+- `MDB_REALM_USERNAME`
+- `MDB_REALM_PASSWORD`
--- a/packages/realm-web-integration-tests/README.md
+++ b/packages/realm-web-integration-tests/README.md
@@ -1,6 +1,6 @@
 # Realm Web integration tests
 
-For ultimate flexability, the tests are written to be able to run from both a Node.js and a browser.
+For ultimate flexability, the tests are written to be able to run from both Node.js and a browser.
 
 ## Running the tests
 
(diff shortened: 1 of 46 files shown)


packages/realm-web-integration-tests/README.md
```diff
@@ -1,9 +1,66 @@
 # Realm Web integration tests
 
+For ultimate flexability, the tests are written to be able to run from both a Node.js and a browser.
```
kneth, line 3:
```suggestion
For ultimate flexability, the tests are written to be able to run from both Node.js and a browser.
```

packages/realm-web-integration-tests/README.md
```diff
@@ -1,9 +1,66 @@
...
+```
+
+## Running the OAuth integration tests
+
+Currently, these are only available when running in a browser.
+
+Before running tests which depends on a specific authentication provider:
+1. Run once to import the app,
+2. Go to http://localhost:9090 (or whereever your MongoDB Realm server is listening)
+3. Setup the appropreate authentication providers.
```
kneth, line 51:
```suggestion
3. Setup the appropriate authentication providers.
```

packages/realm-web-integration-tests/tsconfig.json
```diff
@@ -1,10 +1,11 @@
 {
 	"compilerOptions": {
 		"strict": true,
+		"sourceMap": true,
 		"module": "commonjs",
-		"target": "es6",
+		"target": "ES2019",
```
kneth, line 6: Is this too modern? Do browsers and node.js support it?

packages/realm-web/src/App.ts
```diff
@@ -107,6 +108,11 @@ export class App<
      */
     private _location: Promise<AppLocation> | undefined;
 
+    /**
+     * An helper used to open new and close current windows.
```
kneth, line 112:
```suggestion
     * A helper used to open new and close current window.
```

packages/realm-web/src/index.ts
```diff
@@ -36,6 +38,25 @@ export function app(id: string) {
     }
 }
 
+/**
+ * Handle an OAuth 2.0 redirect
+ *
+ * @param location An optional location to use (defaults to the windows current location)
```
kneth, line 44:
```suggestion
 * @param location An optional location to use (defaults to the windows current location).
```

packages/realm-web/src/storage/MemoryStorage.ts
```diff
@@ -60,5 +67,22 @@ export class MemoryStorage implements Storage {
...
+    public addListener(listener: StorageChangeListner) {
+        return this.listeners.add(listener);
+    }
+
+    /** @inheritdoc */
+    public removeListener(listener: StorageChangeListner) {
+        return this.listeners.delete(listener);
+    }
+
+    /** Fire the  */
```
kneth, line 84: fire the what?

packages/realm-web/src/utils/base64.ts
```diff
@@ -0,0 +1,49 @@
...
+        return Buffer.from(data, "utf-8").toString("base64");
+    } else {
+        throw new Error("Failed to determine the platform specific decoder");
+    }
+}
+
+/**
+ * Decode a string of base64 as text
+ *
+ * @param data The string of base64 encoded text
```
kneth, line 38:
```suggestion
 * @param data The string of base64 encoded text.
```

packages/realm-web/src/utils/string.ts
```diff
@@ -0,0 +1,62 @@
...
+    [key: string]: string | number | boolean;
+}) {
+    return Object.entries(params)
+        .map(([k, v]) => [k, encodeURIComponent(v)])
+        .map(([k, v]) => `${k}=${v}`)
+        .join("&");
+}
+
+/**
+ * Decodes a query string into an object
```
kneth, line 50:
```suggestion
 * Decodes a query string into an object.
```

packages/realm-web/src/utils/string.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param str The query string to decode.
+ * @returns The decoded query string
```
kneth, line 53:
```suggestion
 * @returns The decoded query string.
```

packages/realm-web-integration-tests/tsconfig.json
kraenhansen: Thought about it, but I decided that I expect most users to pass the code exported by our package through their own bundler (Webpack / Parcel) invoking Babel to transpile it to their desired target. We should consider lowering the language version for the upcoming IIFE bundle.

kraenhansen: What do you think @steffenagger?

steffenagger: What are the benefits of a higher target? Unless we want to expose features only found in "ES2019", I think we should go as low as makes sense, otherwise aren't we just limiting developers?
I skipped this one, as it was related to tests (I say "skipped", as I saw it, but haven't gotten to finish the review yet).

kraenhansen: This is in the integration tests, but I'm also increasing this for the main Realm Web package.
The benefit is a better developer experience (for me at least) and potentially better runtime performance as the browser implementation of the runtime would be used instead of TypeScript (or other) polyfills in situations where our users would actually be able to target a newer version of ES.

My thinking is - a user can always lower this version with Babel, if they run into problems, but they cannot increase the target if their environment allows it - if we've made the decision of targeting es6.

kraenhansen: I made this change to enable the use of `Object.fromEntries`, which is supported by 83.8% of users globally https://caniuse.com/#search=fromEntries and since I expect users of Realm Web to transpiling via a bundler, I don't think this will be an issue.

kneth: Let's leave it as it is.

PR #3020, opened 2020-07-01
Realm Web: Refresh access token

kraenhansen:
## What, How & Why?

This fixes the first part of #2965 (refreshing tokens upon the first 401 response.

Note: This should be rebased on `v10` once the current base has been merged.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-network-transport/src/NetworkTransport.ts
+++ b/packages/realm-network-transport/src/NetworkTransport.ts
@@ -17,48 +17,12 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import { MongoDBRealmError } from "./MongoDBRealmError";
+import { NetworkTransport, Request, ResponseHandler, Headers } from "./types";
 
 declare const process: any;
 declare const require: ((id: string) => any) | undefined;
 
 const isNodeProcess = typeof process === "object";
-
-export type Method = "GET" | "POST" | "DELETE" | "PUT";
-
-export type Headers = { [name: string]: string };
-
-export interface Request<RequestBody> {
-    method: Method;
-    url: string;
-    timeoutMs?: number;
-    headers?: Headers;
-    body?: RequestBody | string;
-}
-
-export interface Response {
-    statusCode: number;
-    headers: Headers;
-    body: string;
-}
-
-export type SuccessCallback = (response: Response) => void;
-
-export type ErrorCallback = (err: Error) => void;
-
-export interface ResponseHandler {
-    onSuccess: SuccessCallback;
-    onError: ErrorCallback;
-}
-
-export interface NetworkTransport {
-    fetchAndParse<RequestBody extends any, ResponseBody extends any>(
-        request: Request<RequestBody>,
-    ): Promise<ResponseBody>;
-    fetchWithCallbacks<RequestBody extends any>(
-        request: Request<RequestBody>,
-        handler: ResponseHandler,
-    ): void;
-}
 
 export class DefaultNetworkTransport implements NetworkTransport {
     public static fetch: typeof fetch;
@@ -128,6 +92,8 @@ export class DefaultNetworkTransport implements NetworkTransport {
                 contentType.startsWith("application/json")
             ) {
                 throw new MongoDBRealmError(
+                    request.method,
+                    request.url,
                     response.status,
                     response.statusText,
                     await response.json(),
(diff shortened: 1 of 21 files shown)


packages/realm-web/src/App.test.ts
```diff
@@ -371,6 +367,108 @@ describe("App", () => {
...
+            invalidSessionError,
+            invalidSessionError,
+            {
+                user_id: "bobs-id",
+                access_token: "second-access-token",
+                refresh_token: "very-refreshing",
+            },
+            { bar: "baz" },
+        ]);
+        // Login with an aanonymous user
```
kneth, line 396:
```suggestion
        // Login with an anonymous user
```

packages/realm-web/src/User.ts
```diff
@@ -264,6 +265,23 @@ export class User<
...
+            headers: {
+                Authorization: `Bearer ${this._refreshToken}`,
+            },
+        });
+        const { access_token: accessToken } = response;
+        if (typeof accessToken === "string") {
+            this._accessToken = accessToken;
+            this.storage.accessToken = accessToken;
+        } else {
+            throw new Error("Expected a 'access_token' in the response");
```
kneth, line 281:
```suggestion
            throw new Error("Expected an 'access_token' in the response");
```

packages/realm-web/src/transports/AuthenticatedTransport.ts
```diff
@@ -59,19 +60,37 @@ export class AuthenticatedTransport implements Transport {
...
+                    ...this.buildAuthorizationHeader(user),
+                    ...request.headers,
+                },
+            });
+        } catch (err) {
+            if (
+                user &&
+                retries === 0 &&
+                err instanceof MongoDBRealmError &&
+                err.statusCode === 401
```
steffenagger, line 85: Could there be a case where the endpoint returns `403` (forbidden) from an outdated token - if a user has been granted new/extended access, but the client has not updated its access token yet?
(I guess this depends on how much of the access rights are stored in the JWT).

packages/realm-network-transport/src/types.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface ResponseHandler {
+    onSuccess: SuccessCallback;
+    onError: ErrorCallback;
+}
+
+export interface NetworkTransport {
+    fetchAndParse<RequestBody extends any, ResponseBody extends any>(
+        request: Request<RequestBody>,
+    ): Promise<ResponseBody>;
+    fetchWithCallbacks<RequestBody extends any>(
```
steffenagger, line 50: Just wondering why fetch with parsing is promise-based, and fetch without parsing (or limited parsing) is callback-based?

packages/realm-web/src/App.test.ts
```diff
@@ -371,6 +367,108 @@ describe("App", () => {
...
+        const credentials = Credentials.anonymous();
+        await app.logIn(credentials, false);
+        // Send a request (which will fail)
+        try {
```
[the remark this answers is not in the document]
kraenhansen: Yeah I agree, I did have a look at it when writing this. I didn't like the way they're loading in that plugin.

packages/realm-web/src/User.ts
```diff
@@ -264,6 +265,23 @@ export class User<
...
+            method: "POST",
+            path: "/auth/session",
+            headers: {
+                Authorization: `Bearer ${this._refreshToken}`,
+            },
+        });
+        const { access_token: accessToken } = response;
+        if (typeof accessToken === "string") {
+            this._accessToken = accessToken;
+            this.storage.accessToken = accessToken;
```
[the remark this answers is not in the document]
kraenhansen: `this.storage` persists it to local storage. It's very frequently used and I don't want to hit localStorage on every request to the server, so the user stores its token as a property as well, which is hydrated from the storage at its convenience.

packages/realm-network-transport/src/NetworkTransport.ts
```diff
@@ -124,6 +88,8 @@ export class DefaultNetworkTransport implements NetworkTransport {
                 contentType.startsWith("application/json")
             ) {
                 throw new MongoDBRealmError(
+                    request.method,
+                    request.url,
                     response.status,
                     response.statusText,
                     await response.json(),
```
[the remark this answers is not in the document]
kraenhansen: Is what awaited?

packages/realm-web/src/transports/AuthenticatedTransport.ts
kraenhansen: I don't think so ... it seems to me only 401 are returned when a refresh of the access token would resolve the issue.

packages/realm-network-transport/src/types.ts
kraenhansen: Because when Realm JS is using this, its easier to consume from JS if a `handler` can be passed when invoked instead of having to call methods on the returned promise. Makes sense?

kraenhansen:
I just published realm-network-transport v0.6.0 off this branch.

PR #3022, opened 2020-07-01
bump electron version in tests

blagoev:


Jenkinsfile
```diff
@@ -10,7 +10,7 @@ nodeTestVersion = nodeVersions[0]
 
 //Changing electron versions for testing requires upgrading the spectron dependency in tests/electron/package.json to a specific version.
 //For more see https://www.npmjs.com/package/spectron
-def electronVersions = ['8.1.1', '7.2.3']
+def electronVersions = ['9.0.5', '7.2.3']
```
kneth, line 13: Should we test 8.1.1 too?

blagoev: no we should be testing only the latest. For 7.2.3 I think we added that since studio was using it, so its kept here

blagoev:
this is outdated

PR #3030, opened 2020-07-03
Log out using App::log_out()

kneth:
## What, How & Why?

But only do we log out locally, we also need to log out on the server.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


tests/js/session-tests.js
```diff
@@ -172,7 +172,7 @@ module.exports = {
                 TestCase.assertEqual(session.config.url, config.sync.url);
                 TestCase.assertEqual(session.config.user.identity, config.sync.user.identity);
                 TestCase.assertEqual(session.state, 'active');
-                user.logOut();
+                return user.logOut();
```
kraenhansen, line 175: Would it be too big of an unrelated change to make this function an `async function` and awaiting the `user` and then awaiting the `user.logOut()` afterwards? That will reduce the nesting caused by `.then` and the code calling this won't get the result of `logOut` returned (which it shouldn't need).

tests/js/encryption-tests.js
```diff
@@ -89,7 +89,7 @@ module.exports = {
                     partitionValue: "LoLo"
                 }
             });
-            user.logOut(); // FIXME: clearTestState() doesn't clean up enough and Realm.Sync.User.current might not work
+            return user.logOut(); // FIXME: clearTestState() doesn't clean up enough and Realm.Sync.User.current might not work
```
kraenhansen, line 92: Same comment as in the `session-tests.js`.

tests/js/session-tests.js
kneth: It will require some work since we are spawning a new process (at least, it's more than a 10 minutes hack :smile:)

kraenhansen: What do you mean by "since we are spawning a new process"?

PR #3031, opened 2020-07-04
Fix network transport threading

blagoev:
Always invoke the EventLoopDispatcher when making network requests using the NetworkTransport

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,23 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+NOTE: Support for syncing with realm.cloud.io and/or Realm Object Server has been replaced with support for syncing with MongoDB Realm Cloud.
+
+NOTE: This version bumps the Realm file format to version 11. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v1.0.0, might not be upgradeable. Only [Realm Studio 10.0.0](https://github.com/realm/realm-studio/releases/tag/v10.0.0-beta.1) or later will be able to open the new file format.
+
+### Enhancements
+* None
+
+### Fixed
+* Fix `401 Unauthorized` when using anonymous login resulting in a `Fatal error in v8::HandleScope::CreateHandle()`
+
+### Compatibility
+* MongoDB Realm Cloud.
+* APIs are backwards compatible with all previous releases of Realm JavaScript in the 10.x.y series.
+* File format: generates Realms with format v11 (reads and upgrades file format v5 or later).
+
+### Internal
+* None
+
 10.0.0-beta.7 Release notes (2020-6-26)
 =============================================================
 NOTE: Support for syncing with realm.cloud.io and/or Realm Object Server has been replaced with support for syncing with MongoDB Realm Cloud.
--- a/src/js_network_transport.hpp
+++ b/src/js_network_transport.hpp
@@ -157,16 +157,7 @@ void ResponseHandlerClass<T>::on_error(ContextType ctx, ObjectType this_object, 
 
 
 template<typename T>
-struct JavaScriptNetworkTransportWrapper : public app::GenericNetworkTransport {
-    using ContextType = typename T::Context;
-
-    JavaScriptNetworkTransportWrapper(ContextType ctx) : GenericNetworkTransport() {};
-
-    void send_request_to_server(const app::Request request, std::function<void(const app::Response)> completion_callback) override {};
-};
-
-template<typename T>
-struct JavaScriptNetworkTransport : public JavaScriptNetworkTransportWrapper<T> {
+struct JavaScriptNetworkTransport : public app::GenericNetworkTransport {
     using ContextType = typename T::Context;
     using FunctionType = typename T::Function;
     using ObjectType = typename T::Object;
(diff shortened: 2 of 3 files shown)


bmunkholm:
Which issue does this fix? changelog entry?

blagoev:
@bmunkholm : This is a Jira issue. HELP-16804. An update to that issue was added on July 04 2020 11:23:51 GMT

CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+NOTE: Support for syncing with realm.cloud.io and/or Realm Object Server has been replaced with support for syncing with MongoDB Realm Cloud.
+
+NOTE: This version bumps the Realm file format to version 11. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v1.0.0, might not be upgradeable. Only [Realm Studio 10.0.0](https://github.com/realm/realm-studio/releases/tag/v10.0.0-beta.1) or later will be able to open the new file format.
```
RedBeard0531, line 5: Is this supposed to say `v10.0.0` rather than `v1.0.0`?

src/js_network_transport.hpp
```diff
@@ -175,41 +166,58 @@ struct JavaScriptNetworkTransport : public JavaScriptNetworkTransportWrapper<T>
...
+        send_request_handler(m_ctx, request, completion_callback);
     }
 
 private:
     ContextType m_ctx;
-    ObjectType realm_constructor;
-    ValueType network_transport;
+
+    static realm::util::EventLoopDispatcher<SendRequestHandler> get_send_request_handler() {
+        static auto dispatcher = realm::util::EventLoopDispatcher([&]
```
RedBeard0531, line 187: Is this lambda actually capturing anything? If it was, it would be unsafe, but I don't think it is. It may be good to make this say `[]` rather than `[&]`.

src/js_network_transport.hpp
```diff
@@ -175,41 +166,58 @@ struct JavaScriptNetworkTransport : public JavaScriptNetworkTransportWrapper<T>
...
-            response_handler,
-        };
+    static void init(ContextType ctx) {
+        //this initializes the EventLoopDispatcher on the main JS thread
+        get_send_request_handler();
+    }
 
-        Object::call_method(m_ctx, Value::to_object(m_ctx, network_transport), "fetchWithCallbacks", 2, arguments);
+    void send_request_to_server(const app::Request request, std::function<void(const app::Response)> completion_callback) override {
+        auto send_request_handler = get_send_request_handler();
```
RedBeard0531, line 179: You don't actually need to name this. You can just do `get_send_request_handler()(m_ctx, request, completion_callback)`

src/js_network_transport.hpp
```diff
@@ -175,41 +166,58 @@ struct JavaScriptNetworkTransport : public JavaScriptNetworkTransportWrapper<T>
...
+        send_request_handler(m_ctx, request, completion_callback);
     }
 
 private:
     ContextType m_ctx;
-    ObjectType realm_constructor;
-    ValueType network_transport;
+
+    static realm::util::EventLoopDispatcher<SendRequestHandler> get_send_request_handler() {
```
RedBeard0531, line 186: Should this return `realm::util::EventLoopDispatcher<SendRequestHandler>&` so that it doesn't need to make a copy on every call?

src/js_network_transport.hpp
blagoev: :+1 will fix

src/js_network_transport.hpp
blagoev: Yeah I don't like doing that. It's a personal preference of mine to have a local variable on such invocations. Eases debugging etc.

src/js_network_transport.hpp
blagoev: yes. good point. will fix

src/js_network_transport.hpp
RedBeard0531: btw, my rule of thumb for lambdas is to always just use `[&]` for lamdas that never outlive their local scope, eg, those that are passed to algorithms, and to use fully explicit captures like `[var1, var2, var3=expr]` for lambdas that do outlive their local scope. This does mean that it basically never makes sense to use `[=]` since the cases where you care about the lifetime of the captures are also cases where you usually want to be explicit about them. For example, that helps avoid the hard-to-debug issues where `[=]` lambdas picked up variables that were supposed to be `Protected`.

src/js_network_transport.hpp
RedBeard0531: Sorry I didn't realized this earlier. If you do this, you should make that `auto&` or `auto&&` to avoid making a copy. This is part of the reason I prefer an expression-oriented style that only introduces named vars where needed. But it is fine if you would prefer to keep using the variable, as long as it uses a reference here.

RedBeard0531: I changed this to `auto&` as part of resolving the conflict in my watch() PR when rebasing. https://github.com/realm/realm-js/pull/3004/files#diff-174249016341759699d0ec6689002912

blagoev: good catch, thanks

src/js_network_transport.hpp
blagoev: @RedBeard0531 hey Mathias, thanks for the suggested best practice. Really appreciate this comment.

PR #3034, opened 2020-07-07
toJSON correction matching master

steffenagger:


blagoev:
we should have a changelog entry as well

PR #3039, opened 2020-07-07
Factor out common code for handling callbacks into wrapper functions

RedBeard0531:
## What, How & Why?
This gets us closer to the ideal of each C++ function only expressing its core business logic without excess noise.

## ☑️ ToDos

Since this is just an internal refactor, I don't think any of the default ToDos really apply. Let me know if you disagree.

* ~[ ] 📝 Changelog entry~
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~

### DIFF
--- a/src/js_types.hpp
+++ b/src/js_types.hpp
@@ -32,7 +32,9 @@
 #include <realm/util/base64.hpp>
 #include <realm/mixed.hpp>
 
-#include <object-store/src/util/bson/bson.hpp>
+#include "object-store/src/util/bson/bson.hpp"
+#include "object-store/src/util/event_loop_dispatcher.hpp"
+#include "object-store/src/sync/generic_network_transport.hpp"
 
 #if defined(__GNUC__) && !(defined(DEBUG) && DEBUG)
 # define REALM_JS_INLINE inline __attribute__((always_inline))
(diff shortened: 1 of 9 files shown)


src/js_types.hpp
```diff
@@ -221,6 +223,24 @@ struct Function {
...
     }
+
+    /**
+     * These wrap a js callback into a C++ callback that takes an Optional<AppError> and possibly a result.
+     * The versions that accept a result use the converter argument to convert the C++ result into a js result.
+     * The returned callbacks have the following signatures:
+     *
+     * wrap_void_callback         - void(const util::Optional<app::AppError>& error)
+     * wrap_callback_error_first  - void(const util::Optional<app::AppError>& error, auto&& result)
+     * wrap_callback_result_first - void(auto&& result, const util::Optional<app::AppError>& error)
```
kneth, line 234: Why don't we fix it in object store?

RedBeard0531: I didn't want to break compile for everyone in order to make the callback argument order consistent. At least not without discussing it with the other sdk teams. I'll ping the slack channel and see what everyone thinks. Do you mind going with this for now, and if we fix obj store, we can eliminate one of the `wrap_callback_foo_first` functions in a later commit?

blagoev:
@tgoyne Could you share your view on this. Why not use a single instance of the EventLoopDispatcher in all places. Note that this PR is stepping on this change https://github.com/realm/realm-js/pull/3031.

In short we made a single instance of the EventLoopDispatcher to be used for all network requests in #3031 . Then this PR cleans up a lot of code using the fact that all callbacks will be invoked correctly on the main thread.

The question is why not just have a global EventLoopDispatcher somewhere and use that in similar fashion of this PR. Do you see any issues with that?

thanks

tgoyne:
EventLoopDispatcher's design probably doesn't really make much sense any more. It was originally written as a simple adapter to wrap some pre-existing frameworks without having to redesign anything. Once you're actually reworking the code using it, there's probably a better abstraction that could be found.

blagoev:
thanks. I think we could do that later. Not to expand the churn and the meaning of this PR.

PR #3044, opened 2020-07-08
JSON serialization of Realm.Object & Realm.Collection (incl. polished TS declarations)

### LINKED ISSUE #2997: Tweak existing TS declarations for creation.
The current type checks for the create functionality is very loose, and offer close to no help for developers.

This will be an ongoing task.

steffenagger:
## JSON serialization of `Realm.Object` & `Realm.Collection` & a rework of TS declarations

This PR is a port of https://github.com/realm/realm-js/pull/3013 to the `v10` branch.

These changes fixes #2997 & drastically simplifies https://github.com/realm/realm-studio/issues/1312

#### TS declarations (used in integration-tests)
- [x] Limit input in `create<T>(...)` to match a model (using @nirinchev's `ExtractPropertyNamesOfType`©) **notion of any implications this could introduce, is welcome!**.
- [x] Allow input to contain `Array<T>` in `RealmPartialModel<T>` instead of a defined `Realm.Collection<T>` in `create<T>(...)`.
- [x] Update `objects<T>(...)` to accommodate the underlying logic & different responses.
- [x] Allow extending `Realm.Object` for class models.

#### `toJSON` on `Realm.Object` & `Realm.Collection` (this includes `Realm.List` & `Realm.Results`)
- [x] Implement solution with object references (this introduces a cache to not return new objects).
- [x] Ensure `TypeError: Converting circular structure to JSON` is thrown for circular structures, not `RangeError: Maximum call stack size exceeded`.
- [x] Implement & expose replacer for circular structures. \*

\* _I've included a combined key of the table-type & `_objectId()` as unique `$refId` in `Realm.JsonSerializationReplacer`, for a (visual) reference in serialized JSON. This is only used in case a `primaryKey` is not defined._

#### Integration Tests
- [x] Write test covering `toJSON` implementations for both interface/schema & class model usage.

#### Changelog
- [x] Update `CHANGELOG.md`

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -493,16 +493,16 @@ module.exports = function(realmConstructor, context) {
                 if (seen.length > 0) {
                     const pos = seen.indexOf(parent);
                     if (pos !== -1) {
-                        seen.splice(pos + 1)
+                        seen.splice(pos + 1);
                     } else {
-                        seen.push(parent)
+                        seen.push(parent);
                     }
 
                     if (seen.includes(value)) {
                         return value.$realmId ? { $ref: value.$realmId } : "[Circular reference]";
                     }
                 } else {
-                    seen.push(value)
+                    seen.push(value);
                 }
 
                 return value;
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -100,17 +100,17 @@ module.exports = function(realmConstructor, context) {
                 return existing;
             }
 
-            // create new result, and store in cache
-            const result = { $realmId: id };
+            const result = {};
             cache.set(id, result);
 
-            const moveAsSerializable = (cache, key, from, to) => {
-                // TODO: check that we're dealing with a Realm type, before passing in cache?
-                to[key] = from[key].toJSON ? from[key].toJSON(undefined, cache) : this[key];
-            };
+            Object.defineProperty(result, "$refId", { value: id, configurable: true });
 
-            Object.keys(this).forEach(k => moveAsSerializable(cache, k, this, result));
-            Object.keys(Object.getPrototypeOf(this)).forEach(k => moveAsSerializable(cache, k, this, result));
+            Object.keys(this)
+                .concat(Object.keys(Object.getPrototypeOf(this)))
+                .forEach(k => {
+                    // TODO: check that we're dealing with a Realm type, before passing in cache?
+                    result[k] = this[k].toJSON ? this[k].toJSON(undefined, cache) : this[k];
+                });
 
             return result;
         },
@@ -485,12 +485,19 @@ module.exports = function(realmConstructor, context) {
             const seen = [];
 
             return function(_, value) {
-                if (typeof value !== "object") {
+                if (value === null || typeof value !== "object") {
                     return value;
                 }
 
                 const parent = this;
-                if (seen.length > 0) {
+
+                if (value.$refId) {
+                    Object.defineProperty(value, "$refId", { enumerable: true })
+                }
+
+                if (!seen.length) {
+                    seen.push(value);
+                } else {
                     const pos = seen.indexOf(parent);
                     if (pos !== -1) {
                         seen.splice(pos + 1);
@@ -499,14 +506,12 @@ module.exports = function(realmConstructor, context) {
                     }
 
                     if (seen.includes(value)) {
-                        return value.$realmId ? { $ref: value.$realmId } : "[Circular reference]";
+                        return value.$refId ? { $ref: value.$refId } : "[Circular reference]";
                     }
-                } else {
-                    seen.push(value);
                 }
 
                 return value;
-            }
+            };
         }
-    })
-}
+    });
+};
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -91,7 +91,7 @@ module.exports = function(realmConstructor, context) {
     });
 
     Object.defineProperty(realmConstructor.Object.prototype, "toJSON", {
-        value: function (key, cache = new Map()) {
+        value: function (_, cache = new Map()) {
             // Construct a reference-id of table-name & objectId
             const id = this.constructor.name + this._objectId();
 
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -94,17 +94,20 @@ module.exports = function(realmConstructor, context) {
         value: function (key, cache = new Map()) {
             const id = this._objectId();
 
-            // check if currect objectId has already matched
+            // Check if current objectId has already processed, to keep object references the same.
             const existing = cache.get(id);
             if (existing) {
                 return existing;
             }
 
+            // Create new result, and store in cache.
             const result = {};
             cache.set(id, result);
 
+            // Add '_objectId' as a non-enumerable prop '$refId', for later exposure though e.g. Realm.JsonSerializationReplacer.
             Object.defineProperty(result, "$refId", { value: id, configurable: true });
 
+            // Move all enumerable keys to result, triggering any specific toJSON implementation in the process.
             Object.keys(this)
                 .concat(Object.keys(Object.getPrototypeOf(this)))
                 .forEach(k => {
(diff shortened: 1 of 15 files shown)


lib/extensions.js
```diff
@@ -80,37 +80,43 @@ module.exports = function(realmConstructor, context) {
...
         configurable: true,
         enumerable: false
     });
 
-    Object.defineProperty(realmConstructor.Object.prototype, "keys", {
-        value: function () {
-            return Object.keys(this).concat(Object.keys(Object.getPrototypeOf(this)));
-        },
+    Object.defineProperty(realmConstructor.Object.prototype, "toJSON", {
+        value: function (_, cache = new Map()) {
```
blagoev, line 94: Why do we have `cache` in the argument list. Is this intended to be passed in by the user and whats the idea. What can the user achieve with a custom map?

lib/extensions.js
```diff
@@ -80,37 +80,43 @@ module.exports = function(realmConstructor, context) {
...
+            // Construct a reference-id of table-name & objectId
+            const id = this.constructor.name + this._objectId();
```
blagoev, line 96: we are using the `objectId` but in the generated json we call it `$refId` and `$ref`. Doesn't it make sense to use the same naming like `$obj` and '$objId'. Also why not name this with underscore and full property name `_objectId` and `_object`. It will be more consistent with the actual `_objectId` every Realm Object has.

lib/extensions.js
```diff
@@ -80,37 +80,43 @@ module.exports = function(realmConstructor, context) {
...
+            cache.set(id, result);
 
-    Object.defineProperty(realmConstructor.Object.prototype, "entries", {
-        value: function () {
-            let result = {};
-            for (const key in this) {
-                result[key] = this[key];
-            }
+            // Add the generated reference-id as a non-enumerable prop '$refId', for later exposure though e.g. Realm.JsonSerializationReplacer.
+            Object.defineProperty(result, "$refId", { value: id, configurable: true });
```
blagoev, line 109: does it need to be configurable?

blagoev: yeah. ignore this

lib/extensions.js
```diff
@@ -477,4 +483,48 @@ module.exports = function(realmConstructor, context) {
...
+                const parent = this;
+
+                // Expose the non-enumerable prop $refId for circular serialization, if it exists.
+                if (value.$refId) {
+                    Object.defineProperty(value, "$refId", { enumerable: true })
+                }
+
+                if (!seen.length) {
+                    // If we haven't seen anything yet, we push the current value (root element/array).
+                    seen.push(value);
```
blagoev, line 507: you could do `return value;` here and avoid the big nested `else` after it. The `return early` style ;). Of course that's not always possible but here it seems it will be alright.

lib/extensions.js
```diff
@@ -80,37 +80,43 @@ module.exports = function(realmConstructor, context) {
...
     realmConstructor._networkTransport = new DefaultNetworkTransport();
 
-    Object.defineProperty(realmConstructor.Object.prototype, "toJSON", {
-        value: function () {
-            const result = {}
-            Object.keys(this).forEach(p => result[p] = this[p]);
-            Object.keys(Object.getPrototypeOf(this)).forEach(p => result[p] = this[p]);
-            return result;
+    Object.defineProperty(realmConstructor.Collection.prototype, "toJSON", {
+        value: function (_, cache = new Map()) {
```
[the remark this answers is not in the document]
steffenagger: `_` is the key where the current element resides in its parent (an empty string in case of the root element), and is supplied as first argument by `JSON.stringify` (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior)).

The underscore convention is used to clarify to the reader that the argument is not used, and I changed it because of a question in the v6 draft PR: https://github.com/realm/realm-js/pull/3013#discussion_r447591166
The convention is also used in [lib/notifier.js](https://github.com/realm/realm-js/blob/sa/v10/realm-tojson/lib/notifier.js#L239).

I could add a comment to the function, stating what it is?
Something like this:
```
// The first unused argument is supplied by JSON.stringify, and is the key where the current element resides in its parent
// MDN docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior.
```

lib/extensions.js
steffenagger: `cache` has a default value (the `cache` is initialized if not supplied, and passed on internally), and thus is not required by the user (or `JSON.stringify` for that matter).
It's in the argument list for the `Realm.Object` & `Realm.Collection` implementations to share a cache for the current execution only.
`toJSON()` is a built-in functionality in JS. It's not intended to be called by the user, but can be (and should be able to be) called without arguments.

lib/extensions.js
steffenagger: `$refId` is a combination of the object-type/table-name & `objectId`, because the `objectId` apparently only is unique with in the table (see @kneth's comment here: https://github.com/realm/realm-js/pull/3013).
It's is only included in the JSON serialization to indicate where circular references already exists. By not only consisting of `objectId` I named it to something matching the current context (JSON serialization).

A property with a prefixed underscore indicates a private property (in JS), which should (most likely) not be serialized.
In Realm `_objectId` is a function, which are not serializable, so I would argue that `_objectId` showing up (in an extended state) as a string, would be more confusing.

lib/extensions.js
```diff
@@ -80,37 +80,43 @@ module.exports = function(realmConstructor, context) {
...
-            }
+            // Add the generated reference-id as a non-enumerable prop '$refId', for later exposure though e.g. Realm.JsonSerializationReplacer.
+            Object.defineProperty(result, "$refId", { value: id, configurable: true });
+
+            // Move all enumerable keys to result, triggering any specific toJSON implementation in the process.
+            Object.keys(this)
+                .concat(Object.keys(Object.getPrototypeOf(this)))
+                .forEach(key => {
+                    // TODO: check that we're dealing with a Realm type, before passing in cache?
+                    result[key] = this[key].toJSON ? this[key].toJSON(key, cache) : this[key];
```
kraenhansen, line 116:
```suggestion
                    const value = this[key];
                    result[key] = value instanceof Realm.Collection || value instanceof Realm.Object
                        ? value.toJSON(key, cache)
                        : value;
```

lib/extensions.js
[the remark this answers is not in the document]
nirinchev: I prefer the discard syntax because it immediately tells the reader that we're not using that argument rather than having them read through the entire function to figure out what's happening with the `key` argument.

lib/extensions.js
nirinchev: Wouldn't that be violating the contract of `toJSON`? If an object implements `toJSON`, shouldn't we call it recursively?

steffenagger: I think we're fine. We call it recursively to ensure we can determine circular structures. I can test it.

kraenhansen: It depends on the interpretation of the spec of `toJSON`. I havn't been able to find a better spec for the `toJSON` implementation than https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior. Although its not stated explicitly `JSON.stringify` will call `toJSON` recursively on all the values of the object being returned from this method. If we for-example returned an object with a `Date` as a value, that date will be converted to a string by its `toJSON` even if we didn't explicitly called `toJSON` on it.

steffenagger: I tried @kraenhansen's suggestion and `Date`s are still serialized to iso-strings & `ObjectId`s to hex-strings 👍.
It accommodates my own TODO (that was left in...), so I'll add this check.

kraenhansen:
I'll be moving this forward while Steffen is 🍹 🏖️

lib/extensions.js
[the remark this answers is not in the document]
blagoev: @nirinchev  why would you need to "figure out what's happening with the key argument" when reading this function. @steffenagger I know why is that used but would you think that this is a good code doWork(x, (_, _, _, value)) { }.
Lets leave it as it this since it seems its a personal preference. I still think this is not readable though. As we all know **Programs are meant to be read by humans and only incidentally for computers to execute**  :)

lib/extensions.js
blagoev: sure, but why is it in the arguments list. Can it be declared inside the function body? Or I am missing something on this one.

lib/extensions.js
blagoev: sure thanks.

lib/extensions.js
steffenagger: `Realm.Collection` & `Realm.Object` call `toJSON` on each other, and has to build up the cache together, for that execution only. This was the most elegant way I could think of.

blagoev: I guess I want to ask in that case, could we make this to default to null/undefined and instantiate it only where its used externally.

blagoev: ok ignore this. I am resolving this comment

lib/extensions.js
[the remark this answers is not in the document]
nirinchev: Well, if there's an argument in a function, you expect it to be used. If it isn't, that's likely a mistake and a reader of the code might wonder where was this argument supposed to be used. If something is not used, it shouldn't exist in the first place. In this case, we don't have control over it, so I think the right thing to do is to tell the code reader that whatever that first argument is, we don't care about it.

hudon:
correct me if I'm wrong, but I think this PR breaks when you're updating a record using `create(...UpdateMode.all)` with a partial record, like:
```
interface Record {
  id: number;
  foo: number;
  bar: string;
}
const partOfModel: Partial<Record> = { id: 1, foo: 3 };
realm.create<Record>('record', partOfModel, UpdateMode.All);
```

TypeScript will throw an error saying `Argument of type 'Partial<Record>' is not assignable to parameter of type 'RealmInsertionModel<Record>'`

FYI @blagoev @steffenagger

steffenagger:
@hudon I suspect you're right - and thanks for the example.
The thing that springs to mind is if using `UpdateMode.Never` we would still like to require a full model / warn about missing props... I'll have a look!

PR #3052, opened 2020-07-11
Realm Web: Fixing duplication of user ids

kraenhansen:
## What, How & Why?

This closes an issue occurring when logging in with the same user multiple times:
The list of all users would grow and the user object wouldn't be reused.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


packages/realm-web/src/User.ts
```diff
@@ -252,12 +243,27 @@ export class User<
...
-        // Forget the refresh token
-        this._refreshToken = null;
-        this.storage.refreshToken = null;
+        // Forget the access and refresh token
+        this.accessToken = null;
+        this.refreshToken = null;
+    }
+
+    /**
+     * Authenticate and retrieve the access and refresh token.
```
kneth, line 252:
```suggestion
     * Authenticate and retrieve the access and refresh tokens.
```

PR #3059, opened 2020-07-13
throw more meaningful message when parsing properties

blagoev:


src/js_schema.hpp
```diff
@@ -278,7 +278,15 @@ ObjectSchema Schema<T>::parse_object_schema(ContextType ctx, ObjectType object_s
         for (uint32_t i = 0; i < length; i++) {
             ObjectType property_object = Object::validated_get_object(ctx, properties_object, i);
             std::string property_name = Object::validated_get_string(ctx, property_object, name_string);
-            Property property = parse_property(ctx, property_object, object_schema.name, std::move(property_name), object_defaults);
+            Property property;
+            try {
+                property = parse_property(ctx, property_object, object_schema, property_name, object_defaults);
```
bmunkholm, line 283: Is there a reason to not fix this inside parse_property()?

blagoev: What is your concern exactly? The meaning of this is to handle all parsing errors at once.

kneth: I think the trick here is that we want to catch `invalid_argument` only while `logic_error` should bubble up to the user. If we want to put a `try`/`catch` around every `validtated_to_X` in `parse_property()`, I believe it would end up being unreadable.

blagoev: exactly

bmunkholm: My thought was just to avoid the code duplication.
Either by 1) wrapping the entire body of parse_property() in this try/catch or 2) make another function that does that and call that instead.

blagoev: I could see that being beneficial in general but these catch statements happen in a single if/else block and I prefer to have the simpler code inside it. I don't want to to it inside parse_property since this will mean altering the parse_property behavior which might not be desired in other places.

PR #3061, opened 2020-07-14
Minor type bugs related to Credentials and MongoDB Collection watch

kraenhansen:
## What, How & Why?

* Fixes a few issues with the MongoDB collection watch types and the credentials types.
* Adds a stub method in Realm Web, throwing if a user calls `watch` on a collection.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


types/services.d.ts
```diff
@@ -213,6 +209,205 @@ declare namespace Realm {
...
+                to: DocumentNamespace;
+            } & BaseChangeEvent;
+
+            /**
+             * Occurs when a database is dropped.
+             */
+            type DropDatabaseEvent = {
+                /** The type of operation which was performed on the document. */
+                operationType: OperationType.DropDatabase;
+                /** The namespace (database and collection) the collection that got dropped. */
```
RedBeard0531, line 375: This seems incorrect.

types/services.d.ts
```diff
@@ -213,6 +209,205 @@ declare namespace Realm {
...
+                ns: Omit<DocumentNamespace, "coll">;
+            } & BaseChangeEvent;
+
+            /**
+             * Occurs when a database is dropped.
```
RedBeard0531, line 380: I think you meant to put something else here.

types/services.d.ts
```diff
@@ -213,6 +209,205 @@ declare namespace Realm {
...
+             */
+            type InvalidateEvent = BaseChangeEvent;
+
+            /**
+             * Represents a change event communicated via a MongoDB change stream.
+             * This type of stream always includes a fullDocument for update events, and also includes the change event ID and namespace of the event as returned by MongoDB.
```
RedBeard0531, line 386: Note that your type doesn't have `ns` in all cases, so I don't think this comment is correct. I think I'd just drop this whole sentence. I'd also add `@see https://docs.mongodb.com/manual/reference/change-events/`

packages/realm-web/src/services/MongoDBService.ts
```diff
@@ -229,6 +229,11 @@ class MongoDBCollection<T extends Document>
             upsert: options.upsert,
         });
     }
+
+    /** @inheritdoc */
+    watch(): AsyncGenerator<any> {
```
RedBeard0531, line 234: Do you want to throw the `async*` on here now and put in a defaulted arg? I don't think it makes a huge difference either way, it just seems worth making the signature correct now, even if you aren't implementing yet.

types/app.d.ts
```diff
@@ -223,14 +269,18 @@ declare namespace Realm {
         /**
          * All authenticated users.
          */
-        readonly allUsers: Readonly<User<FunctionsFactoryType, CustomDataType>[]>;
+        readonly allUsers: Readonly<
```
RedBeard0531, line 272: Do you actually need the `Readonly<>` here in addition to the `readonly` keyword?

types/services.d.ts
```diff
@@ -213,6 +209,205 @@ declare namespace Realm {
              */
             type AggregatePipelineStage = object;
 
+            /**
+             * An operation performed on a document.
+             */
+            enum OperationType {
```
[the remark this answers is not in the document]
kraenhansen: No - I think you're right here. I just copied the enum from the legacy SDK.

types/app.d.ts
kraenhansen: The first warns if the user assigns to `allUsers` the `Readonly` warns if the user assigns to any of the elements of the list.

packages/realm-web/src/services/MongoDBService.ts
kraenhansen: Adding the options argument, but adding the `async*` yields a eslint warinng (so I'll leave that out for now):
> This generator function does not have 'yield'. `eslint(require-yield)`

types/services.d.ts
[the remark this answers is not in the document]
kraenhansen: One advantage of an `enum` is that it enables us to document the semantics of the various values.

types/app.d.ts
```diff
@@ -168,6 +176,24 @@ declare namespace Realm {
             password: string,
         ): Credentials<Credentials.EmailPasswordPayload>;
 
+        /**
+         * Creates credentials that logs in using the [Custom Function Provider](https://docs.mongodb.com/realm/authentication/custom-function/).
```
kneth, line 180: Disclaimer: I am not a native English speaker.

`that logs` -> `to log`

types/services.d.ts
```diff
@@ -213,6 +209,205 @@ declare namespace Realm {
...
+                /** The type of operation which was performed on the document. */
+                operationType: OperationType.Delete;
+                /** The namespace (database and collection) which the document got deleted from. */
+                ns: DocumentNamespace;
+                /** A document that contains the _id of the deleted document. */
+                documentKey: DocumentKey<T["_id"]>;
+            } & BaseChangeEvent;
+
+            /**
+             * Occurs when a collection is dropped from a database.
```
kneth, line 347: Maybe `Emitted` instead of `Occurs`

types/app.d.ts
kraenhansen: I changed it to "authenticates".

types/services.d.ts
kraenhansen: Technically, they won't be emitted. Since @RedBeard0531 choose to go for a generator function over an event emitter pattern a change events gets "yielded" from the generator.

types/app.d.ts
RedBeard0531: I think those are all correct, but expressing slightly different aspects of the idea. I would probably phrase it as "for logging in", but none of these are really wrong.

More importantly, though, I think we need to change the `Creates credentials` part. With a lower-case, non-code-font "credentials", it implies that you are making new credentials, ie making a new account, or generating a new API key, but I don't think that is what this does. Maybe we want something like "Returns a `Credentials` object for logging in using...", possibly without the word "object" which is somewhat redundant. Or we could just leave the word "credentials" out entirely, since we don't need to repeat the signature in prose, and just say "Factory for logging in using ..."

types/services.d.ts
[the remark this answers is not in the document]
kraenhansen: The rename and drop operations are not obvious in my opinion.
I don't think those comments will find its way into an editor or a jsdoc generated site.

types/app.d.ts
RedBeard0531: "credentials which authenticates" at the very least should be changed to "credentials which authenticate" (no `s`).

kraenhansen: I updated this to:
> Factory for `Credentials` which authenticate using the ...

types/services.d.ts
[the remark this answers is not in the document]
RedBeard0531: Lets make this a `const enum` so that it can only be used in ways that completely compile away. We don't want people to be doing operations that assume that there is some `OperationType` object.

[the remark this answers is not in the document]
RedBeard0531: Actually, it looks like the semantics for enums aren't quite right ([Playground Link](https://www.typescriptlang.org/play/#code/MYewdgzgLgBApmArgWxgDRgbxgQxgXhgCIciYBfAKABs5YAPALnQONIG4YB6LmAFQCeABzgwA5CSJiYASwgwwIWDggQZAczA4ARrRhQQ+4aLFoxAOgAUAJgDM16wEogA)). Maybe we should go back to union type of string literals and put the documentation for each kind in the docstring for the type (or just refer people to the mongo docs)?

types/services.d.ts
```diff
@@ -213,6 +209,205 @@ declare namespace Realm {
...
+                operationType: OperationType.Insert;
+                /** The namespace (database and collection) the document got inserted into. */
+                ns: DocumentNamespace;
+                /** A document that contains the _id of the inserted document. */
+                documentKey: DocumentKey<T["_id"]>;
+                /** The new document created by the operation */
+                fullDocument: T;
+            } & BaseChangeEvent;
+
+            type UpdateEvent<T extends Document> = {
```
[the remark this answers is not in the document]
RedBeard0531: ping

types/services.d.ts
```diff
@@ -213,6 +209,206 @@ declare namespace Realm {
...
+                to: DocumentNamespace;
+            } & BaseChangeEvent;
+
+            /**
+             * Occurs when a database is dropped.
+             */
+            type DropDatabaseEvent = {
+                /** The type of operation which was performed on the document. */
+                operationType: OperationType.DropDatabase;
+                /** The namespace (specifying the database name) of the database that got dropped. */
```
RedBeard0531, line 374: `specifying only the database name`

types/services.d.ts
```diff
@@ -213,6 +209,206 @@ declare namespace Realm {
...
+            type WatchOptions<IdType> =
+                | {}
+                | {
+                      /** Filter events. */
+                      filter: Filter;
+                  }
+                | {
+                      /** A list of document ids to watch. */
+                      ids: IdType[];
+                  };
```
RedBeard0531, line 410: Turns out this doesn't actually work at preventing mixing `filter` and `ids` since you are always allowed to have extra fields (and `|{}` even means that any object can be passed, oops!) In order to disallow mixing `filter` and `ids` we would need to add `filter: never` and/or `ids: never` to each of the 3 alternatives. But that still leaves the type a bit of a pain to work with. Maybe it is best to just document the restriction and use a simple `{filter?: Filter, ids?: IdType[]}` definition.

kraenhansen: That is unexpected - I'll have to investigate further.

RedBeard0531: Function overloads seem to have the desired semantics: [Playground Link](https://www.typescriptlang.org/play/?target=5#code/PTAEEsDtQEwOgC4GcBQAzArpAxg8B7aNACgEMAnAcyQH4AuUAb1AA97RIMBbAIwFNyoAL4BKBp14CA3Oiy4CRMlSQNmAT3Hd+g0ZsnkZKEBGjguABwA2iVJhx5CoEsxYBeAAwAaUGo-DQrkyiTKDkfAgY5NAsoADUPlLCKOjEIjLOounELgwAjMJpKep5BYbOrHneGqD5maDG2PgW4JZ8oALk+ORAA)

kraenhansen: Okay - it makes sense, I believe you're right that function overloading is the best way to express this.
It does however have the drawback that it seems I need to provide the doc comment three times.

types/services.d.ts
```diff
@@ -269,11 +264,11 @@ declare namespace Realm {
             /**
              * A base change event containing the properties which apply across operation types.
              */
-            type BaseChangeEvent = {
+            type BaseChangeEvent<T extends OperationType> = {
```
RedBeard0531, line 267: Maybe `s/T/OpType/` so that it is clear that it isn't a type (in the usual sense) but one of the string values?

types/services.d.ts
```diff
@@ -566,9 +536,39 @@ declare namespace Realm {
...
                  */
-                watch(
-                    options?: WatchOptions<T["_id"]>,
-                ): AsyncGenerator<ChangeEvent<T>>;
+                watch(options?: {}): AsyncGenerator<ChangeEvent<T>>;
+
+                /**
+                 * Creates an asynchronous change stream to monitor this collection for changes.
+                 *
+                 * By default, yields all change events for this collection. You may specify at most one of
```
RedBeard0531, line 544: I'm not sure if it is better to make the documentation cover all overloads (as you are doing here) or be specific to the overload selected? It seems like tsserver only shows the one overload at a time, so maybe this is best when writing code, even if the overload-specific docs are better for things like tooltips/hover text. 🤷🏻‍♂️ Have you seen an idiom among other projects?

PR #3064, opened 2020-07-14
Realm Web: Reject login when window closes

kraenhansen:
## What, How & Why?

This adds an enhancement of rejecting the promised returned by `logIn` if the user closes the window before the OAuth flow completes.

This is based off `kh/realm-web-iife`, it should be rebased onto `v10` once that PR merges.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually in an app)

### DIFF
--- a/packages/realm-web/src/OAuth2Helper.ts
+++ b/packages/realm-web/src/OAuth2Helper.ts
@@ -191,7 +191,7 @@ export class OAuth2Helper {
         // Return a promise that resolves when the  gets known
         return new Promise((resolve, reject) => {
             let redirectWindow: Window | null = null;
-            let windowClosedInterval = 0;
+            let windowClosedInterval: TimerHandle = 0;
 
             const handleStorageUpdate = () => {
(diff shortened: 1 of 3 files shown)


packages/realm-web/src/OAuth2Helper.ts
```diff
@@ -206,10 +212,22 @@ export class OAuth2Helper {
                     }
                 }
             };
+
             // Add a listener to the state storage, awaiting an update to the secret
             stateStorage.addListener(handleStorageUpdate);
             // Open up a window
             redirectWindow = this.openWindow(url);
+            // No using a const, because we need the two listeners to reference each other when removing the other.
+            windowClosedInterval = setInterval(() => {
```
kraenhansen, line 221: I'm using `setInterval` to poll the `window.closed` property, because apparently adding an event listener on the "close" event of on the newly created window is a cross-origin violation.

PR #3071, opened 2020-07-17
Realm Web: Publishing an IIFE bundle

### LINKED ISSUE #2966: Realm Web: Publish an "iife" bundle
To make Realm Web consumable via a simple <script /> tag, we should publish a self-executing function ("iife").

From the [Rollup.js guide](https://rollupjs.org/guide/en/#using-output-plugins):
> This format wraps the code so that it can be consumed via a script tag in the browser while avoiding unwanted interactions with other code. As we have an export, we need to provide the name of a global variable that will be created by our bundle so that other code can access our export via this variable.

We could publish using https://unpkg.com/

kraenhansen:
## What, How & Why?

This closes #2966 and refactors the rollup configuration and code for both realm-network-transport and realm-web to avoid having runtime checks for the environment. I also added a (minor) breaking change in renaming a helper method on the namespace used to get a cached app instance from an app id.

This should be rebased onto `v10` once the parent branch merges.
## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


packages/realm-network-transport/src/types.ts
```diff
@@ -52,3 +52,202 @@ export interface NetworkTransport {
         handler: ResponseHandler,
     ): void;
 }
+
```
kraenhansen, line 55: All of these types are more or less copy-pasted from node_modules/typescript/lib/lib.dom.d.ts.

packages/realm-network-transport/src/types.ts
```diff
@@ -52,3 +52,202 @@ export interface NetworkTransport {
...
+type FetchHeadersInit = FetchHeaders | string[][] | Record<string, string>;
+type FetchRequestCache = unknown;
+type FetchRequestCredentials = unknown;
+type FetchRequestMode = unknown;
+type FetchRequestRedirect = unknown;
+type FetchReferrerPolicy = unknown;
+type FetchRequestDestination = unknown;
+
+interface FetchBody {
+    readonly body: unknown | null;
```
RedBeard0531, line 96: This needs to be the correct ReadableStream type for `watch()`

packages/realm-network-transport/types/index.d.ts
```diff
@@ -0,0 +1,6 @@
+// Timer related stuff
```
RedBeard0531, line 1: We really shouldn't be declaring this stuff ourselves.

packages/realm-network-transport/tsconfig.json
```diff
@@ -3,19 +3,24 @@
...
 		"types": [
 			"mocha",
 			"chai",
-			"express",
 			"node"
 		],
 		"lib": [
-			"ES2015",
-			"DOM"
+			"ES2015"
```
RedBeard0531, line 17: Why not ES2018? Isn't that the default for the current target?

Also, we should leave trailing comments on lists (in files that support it like these) so that we don't get diffspam.

kraenhansen: It's just because we're not actively using features newer than ES2015 - if we did want to use those, we could simply bump this.

packages/realm-network-transport/types/index.d.ts
kraenhansen: I think we do.
I'm failing to find a good dependency declaring types of the intersection of the Node.js and browser / DOM runtime.
I'll be happy to change this if you can find an alternative that doesn't define globals that are not available in the other.

packages/realm-web/src/OAuth2Helper.test.ts
```diff
@@ -41,7 +41,7 @@ describe("OAuth2Helper", () => {
                             userAuth: "our-little-secret",
                         }),
                     );
-                });
+                }, 10);
```
RedBeard0531, line 44: Why 10? It seems like omitting a number here would be closest to the old code.

packages/realm-web/src/index.ts
```diff
@@ -38,26 +36,11 @@ export function app(id: string) {
...
-        // Ensure calling this never throws: It should not interrupt a users flow.
-        console.warn(err);
-    }
-}
-
 export { App };
 export { Credentials } from "./Credentials";
 export { User, UserState } from "./User";
-export { createDefaultStorage } from "./storage";
+export * from "./environment";
```
RedBeard0531, line 42: Isn't the environment stuff internal implementation details? Do we really want to be exporting it like this?

packages/realm-network-transport/src/types.ts
kraenhansen: I suggest we'll leave this "unknown" here and allow the consumer of this provide [an instanceof type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#instanceof-type-guards) when consuming this API.

packages/realm-web/src/index.ts
```diff
@@ -38,26 +36,11 @@ export function app(id: string) {
...
+
+// Export a selection of the bson types
+import { ObjectId } from "bson";
+export const BSON = { ObjectId };
```
RedBeard0531, line 46: Why not export the full `BSON`? It seems confusing to introduce 2 different BSONs

packages/realm-web/src/node/index.ts
```diff
@@ -0,0 +1,61 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import { setEnvironment, Environment } from "../environment";
+import { MemoryStorage } from "../storage";
+
+const environment: Environment = {
+    console,
+    defaultStorage: new MemoryStorage(),
+    openWindow: url => {
+        console.log(`Please open this URL: ${url}`);
```
RedBeard0531, line 44: Should we use something like https://github.com/sindresorhus/open? Or maybe do something with a headless browser?

If not, should this just throw since it isn't able to fulfill the contract of opening a window?

packages/realm-web/src/OAuth2Helper.test.ts
kraenhansen: I honestly can't remember if there was a good reason that this was added, I'll revert.

packages/realm-web/src/index.ts
kraenhansen: It's exposed primarily for [the integration tests to hook into](https://github.com/realm/realm-js/blob/kh/realm-web-iife/packages/realm-web-integration-tests/src/app.test.ts#L40). I don't think we'll need to include this in any generated API docs.

packages/realm-web/src/index.ts
kraenhansen: Technically it won't be 2 different BSONs, when importing from "bson" it will return the same module object as any other API which will be importing "bson".

I choose to go with the single type, because that was the only type the MongoDB service expected, but now that this has changed we might actually choose to leave this re-export out entirely?

packages/realm-web/src/node/index.ts
kraenhansen: I would only really expect this to be called from integration tests when they're running from node.
In that case I don't know what the expected behaviour would be - I can see how it would be annoying to get UI spawned when running on CI for example.
I believe it does fulfill the contract. `openWindow` is allowed to return `null` if no windows handle was produced.

packages/realm-web/src/index.ts
kraenhansen: I figured out that this export turns out to be the cause of a ///-refrence to node in out index.d.ts, so I'll remove this for now and add it if someone requests it.

kraenhansen:
> My initial take is that this is making the code worse because it is making us responsible for maintaining type decls that we use to just get automatically. Are these changes actually required to make an IIFE bundle?

I'm unable to find a package / repository which declares the intersection of the node.js and browser runtimes, if you can help me find that I too think we should adopt that over defining these types ourselves.

Ideally this should have been a part of [the types](http://npmjs.com/package/@types/isomorphic-fetch) for [isomorphic-fetch](http://npmjs.com/package/isomorphic-fetch), but they too have a ///-slash reference to the Node.js types which brings in the entire Node.js runtime.

packages/realm-web/src/environment.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Window } from "./OAuth2Helper";
+
+/**
+ * A simplified interface for the console.
+ * This should reflect the lowest common denominator across environemnts.
+ */
+interface Console {
+    /**
+     *
+     */
```
RedBeard0531, line 29: ???

packages/realm-web/src/environment.ts
```diff
@@ -0,0 +1,80 @@
...
+    defaultStorage: Storage;
+
+    /**
+     * Open a browser window.
+     */
+    openWindow: (url: string) => Window | null;
+};
+
+let environment: Environment | null = null;
```
RedBeard0531, line 57: I'm not sure it is better to export the getter+setter vs just exporting this variable.

packages/realm-web/types/simplified-node/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+// Our version of Node.js types are very restricted
+// This file will get resolved when realm's dependency on bson resolves a dependency on @types/node.
+// We need this file to prevent the NodeJS globals to be accessable from within the test source files.
+
+/**
+ * The simples buffer we can come up with.
```
RedBeard0531, line 24: `simplest`

packages/realm-network-transport/src/types.ts
```diff
@@ -52,3 +52,122 @@ export interface NetworkTransport {
...
+export type AbortController = {
+    /**
+     * The prototype of an instance is the class.
+     */
+    prototype: AbortController;
+
+    /**
+     * Constructs an AbortController.
+     */
+    new (): AbortController;
```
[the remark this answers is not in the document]
kraenhansen: With the current (recommended) configuration of Prettier, yes.

packages/realm-network-transport/src/types.ts
```diff
@@ -52,3 +52,122 @@ export interface NetworkTransport {
...
     ): void;
 }
+
+// AbortController
+
+type AbortSignal = any;
+
+/** A controller object that allows you to abort one or more DOM requests as and when desired. */
+
+export type AbortController = {
```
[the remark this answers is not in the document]
kraenhansen: I agree that this is a little stupid. The `abort-controller` package does not ship with types and there are no types available via a "definitely typed" @types package, so ...
I put the types/abort-controller/index.d.ts in, but it's not really a strong guarantee that this will match the package's implementation.

packages/realm-web/package.json
```diff
@@ -2,14 +2,18 @@
...
   "version": "0.7.0",
   "description": "Authenticate and communicate with the MongoDB Realm platform, from your web-browser",
-  "main": "dist/bundle.cjs.js",
-  "module": "dist/bundle.es.js",
-  "types": "dist/bundle.d.ts",
+  "main": "./dist/bundle.cjs.js",
+  "module": "./dist/bundle.es.js",
+  "types": "./dist/bundle.d.ts",
+  "browser": {
+    "./dist/bundle.cjs.js": "./dist/bundle.dom.cjs.js",
```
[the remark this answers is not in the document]
kraenhansen: I don't think this has been properly standardized, but it's described via https://gist.github.com/defunctzombie/4339901/49493836fb873ddaa4b8a7aa0ef2352119f69211 and it's implemented by most major bundlers to my knowledge.

packages/realm-web/src/environment.ts
kraenhansen: I want to ensure no code is accessing this if it has not been initialized, so to prevent a strange runtime error I expose it via a function that throws a meaningful error if that invariant is violated.

packages/realm-network-transport/types/index.d.ts
kraenhansen: Nope.
They are [included in the compilation](https://github.com/realm/realm-js/blob/kh/realm-web-iife/packages/realm-network-transport/tsconfig.json#L22) and as long as that file is never imported into the dependency tree, they will not end up in the bundle.

packages/realm-network-transport/src/NetworkTransport.ts
```diff
@@ -36,38 +38,15 @@ export class DefaultNetworkTransport implements NetworkTransport {
...
-                const nodeRequire = require;
-                DefaultNetworkTransport.AbortController = nodeRequire(
-                    "abort-controller",
-                );
-            } else {
-                throw new Error(
-                    "DefaultNetworkTransport.AbortController must be set before it's used",
-                );
-            }
+            throw new Error(
```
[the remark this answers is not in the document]
kraenhansen: We could create an `InternalError` which extends `Error` but in my experience, it's not really that common.
I did it with `MongoDBRealmError` because I want to propagate the statusCode, etc., but for an error thrown at runtime if the bundle is malformed, I wouldn't spend the time dealing with this - since no users will ever see this error 🤞

packages/realm-web/src/environment.ts
kraenhansen: I updated these, but also moved them out of the environment and onto `environment.d.ts` since they are available as globals, it seems strange to include them in the environment object like this.

kraenhansen:
@RedBeard0531 I believe I've address your comments above.
I realize that this PR is now including a fairly large restructuring of the packages, but I believe this is needed to ensure the IIFE bundle doesn't contain code which is not intended for a browser. This was a little less critical when shipping code intended primarily for bundlers (like Parcel or Webpack) since they have ways to polyfill the node.js runtime.

packages/realm-web/src/index.ts
RedBeard0531: Maybe I don't quite understand how this works exactly. When someone imports realm-web into a TS-using project, would they ever see the "real" types that you are using here, or would they only see the "fake" ones from the realm-js/types directory?

packages/realm-network-transport/types/index.d.ts
RedBeard0531: I'm still not convinced that this is a good idea. However, I also don't want to block our progress. How about we move forward with this for now, but after everyone is back from vacation, we discuss as a full team and decide how we feel about this? Does that work for you?

packages/realm-network-transport/src/NetworkTransport.ts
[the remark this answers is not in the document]
RedBeard0531: I was half asking so I knew what to do for a similar situation in my code. I guess I'll just throw an Error as well.

packages/realm-network-transport/src/types.ts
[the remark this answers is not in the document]
RedBeard0531: Is there any advantage over just using `any`? It seems like either way we are just telling the compiler "trust me, I know what I'm doing", this is just less explicit about doing that.

packages/realm-web/src/environment.ts
RedBeard0531: Doesn't seem like a huge improvement over just getting the null error, but 🤷‍♂️

packages/realm-web/types/environment.d.ts
```diff
@@ -0,0 +1,55 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+/**
+ * A simplified interface for the console.
```
RedBeard0531, line 20: Is there a reason github isn't highlighting this file? Do we need to do something different for it to know what type it is?

PR #3082, opened 2020-07-21
Allow a string argument when constructing an App

### LINKED ISSUE #3081: Construct an App from a string
With the current implementation of the app constructor, no exception is thrown if the argument is not an object (leading to the app id becoming an empty string).

We agreed that we would provide an overload (shorthand) for the constructor, where users would be able to call it with an id, instead of the entire configuration object, like so:

```typescript
const app = new Realm.App("my-awesome-app-id");
```

* We're [missing a branch in this code](https://github.com/realm/realm-js/blob/v10/src/js_app.hpp#L107-L142) to account for users calling the function with a string and an else clause throwing an error for unexpected types of the argument.
* The JS docs also need to be updated to reflect this (TypeScript types are updated).

kraenhansen:
## What, How & Why?

This closes #3081 by adding a branch to the check if the argument is a string.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ NOTE: Support for syncing with realm.cloud.io and/or Realm Object Server has bee
 NOTE: This version bumps the Realm file format to version 11. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v1.0.0, might not be upgradeable. Only [Realm Studio 10.0.0](https://github.com/realm/realm-studio/releases/tag/v10.0.0-beta.1) or later will be able to open the new file format.
 
 ### Enhancements
-* None.
+* Adding the possibility of passing a string argument when constructing an App. ([#3082](https://github.com/realm/realm-js/pull/3082))
 
 ### Fixed
 * A crash when calling `user.apiKeys.fetchAll()`
--- a/tests/js/app-tests.js
+++ b/tests/js/app-tests.js
@@ -53,6 +53,28 @@ module.exports = {
         TestCase.assertInstanceOf(app, Realm.App);
     },
 
+    testNewAppFromString() {
+        let app = new Realm.App(config.id);
+        TestCase.assertInstanceOf(app, Realm.App);
+        TestCase.assertEqual(app.id, config.id);
+    },
+
+    testNewAppFromUndefined() {
+        const error = TestCase.assertThrows(() =>  new Realm.App());
+        TestCase.assertEqual(
+            error.message,
+            'Invalid arguments: 1 expected, but 0 supplied.',
+        );
+    },
+
+    testNewAppFromOther() {
+        const error = TestCase.assertThrows(() => new Realm.App(1234));
+        TestCase.assertEqual(
+            error.message,
+            'Expected either a configuration object or an app id string.',
+        );
+    },
+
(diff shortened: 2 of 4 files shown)


tests/js/app-tests.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+"use strict";
```
RedBeard0531, line 1: Why did all of the strings change from `'` to `"`?

tests/js/app-tests.js
```diff
@@ -53,14 +53,40 @@ module.exports = {
...
     },
 
+    testNewAppFromString() {
+        let app = new Realm.App(config.id);
+        TestCase.assertInstanceOf(app, Realm.App);
+        TestCase.assertEqual(app.id, config.id);
+    },
+
+    testNewAppFromUndefined() {
+        const error = TestCase.assertThrows(() => {
```
RedBeard0531, line 63: Why use the statement form `() => { stmt; }` rather than expression form `() => expr` here?

tests/js/app-tests.js
```diff
@@ -53,14 +53,40 @@ module.exports = {
...
+            new Realm.App();
+        });
+        TestCase.assertEqual(
+            error.message,
+            "Invalid arguments: 1 expected, but 0 supplied.",
+        );
+    },
+
+    testNewAppFromOther() {
+        const error = TestCase.assertThrows(() => {
```
RedBeard0531, line 73: Ditto

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ NOTE: Support for syncing with realm.cloud.io and/or Realm Object Server has bee
 NOTE: This version bumps the Realm file format to version 11. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v1.0.0, might not be upgradeable. Only [Realm Studio 10.0.0](https://github.com/realm/realm-studio/releases/tag/v10.0.0-beta.1) or later will be able to open the new file format.
 
 ### Enhancements
-* None.
+* Adding the possibility of passing a string argument when constructing an App. ([#3082](https://github.com/realm/realm-js/pull/3082))
```
RedBeard0531, line 8: I thought GFM auto-linkified things like `#3082` (eg #3082). I think it does something fancier than a plain link, so it may be better to let it do its thing than manually linkify it here.

RedBeard0531: Also, it may be better to link to the issue resolved (#3081) rather than the PR resolving it.

tests/js/app-tests.js
kraenhansen: This happened because [the project defines an .eslintrc.json prescribing double quotes](https://github.com/realm/realm-js/blob/master/.eslintrc.json#L19) and I happen to be running VSCode with the eslint plugin formatting on saving.

But - I'll revert and save without formatting and leave the job of fixing this across the codebase (or updating the eslint.json) to a later time.

CHANGELOG.md
kraenhansen: I'm primarily just following past usage here.
Also, historically we've pasted the top part of the changelog when publishing a release on Slack, where these links translate better than if they're just plain `#3082` text.

PR #3088, opened 2020-07-24
Realm Web: Linking credentials

### LINKED ISSUE #2968: Realm Web: Link a user with other credentials
Users can have multiple identities at multiple authentication providers, we should allow users to link there `User` with another set of credentials.

When implementing `linkWithRedirect` we should use a different header than "X-Stitch-Location" which was used in the previous Stitch SDKs https://github.com/mongodb/stitch-js-sdk/blob/310f0bd5af80f818cdfbc3caf1ae29ffa8e9c7cf/packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts#L150-L184
(https://jira.mongodb.org/browse/RJS-606)

kraenhansen:
## What, How & Why?

This closes #2968.
It should be rebased onto `v10` once the current base has merged.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


kraenhansen:
I believe this is mostly missing a changelog entry and an integration test.

kraenhansen:
Turns out I needed to do a little more refactoring 🙄

packages/realm-web/src/App.ts
```diff
@@ -247,6 +258,38 @@ export class App<
...
+                    ({ hostname }) => {
+                        if (typeof hostname !== "string") {
+                            throw new Error(
+                                "Expected response to contain a 'hostname'",
+                            );
+                        } else {
+                            return hostname;
+                        }
+                    },
+                    err => {
```
steffenagger, line 283: Personal opinion: I prefer the rejection-handler to be in it's own `catch` chain. (If you disagree, just resolve this one)

packages/realm-web/src/Fetcher.ts
```diff
@@ -235,34 +234,24 @@ export class Fetcher {
...
-            throw new Error("Expected response to contain a 'hostname'");
-        }
-        return hostname;
+    public get appRoute() {
+        return routes.api().app(this.appId);
     }
 
     /**
-     * @returns The path of the app route.
+     * @returns The location URL of the app.
```
steffenagger, line 244: Maybe indicate that it's returned in the form of a promise?

packages/realm-web/src/utils/string.ts
```diff
@@ -35,15 +35,29 @@ export function generateRandomString(length: number, alphabet: string) {
...
+        ([k, v]) => typeof v !== "undefined",
+    ) as Array<[string, string | number | boolean]>;
+    // Determine if a prefixed "?" is appropreate
+    const prefix = prefixed && entriesWithValue.length > 0 ? "?" : "";
+    // Transform keys and values to a query string
+    return (
+        prefix +
+        entriesWithValue
+            .map(([k, v]) => [k, encodeURIComponent(v)])
+            .map(([k, v]) => `${k}=${v}`)
```
steffenagger, line 58: Any reason to map twice? Seems like this could, just as easily & readable , be done in one.

steffenagger (commented):
I'm a bit rusty in OAuth, but LGTM.

packages/realm-web-integration-tests/src/user.test.ts
```diff
@@ -51,4 +51,26 @@ describe("User", () => {
...
+
+        const now = new Date();
+        const nonce = now.getTime();
+        const email = `dinesh-${nonce}@testing.mongodb.com`;
+        const password = "v3ry-s3cret";
+        try {
+            await app.emailPasswordAuth.registerUser(email, password);
+        } catch (err) {
+            // Allow the user to already be registered in the app
+            expect(err.message).contains("name already in use", err.message);
```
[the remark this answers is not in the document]
kraenhansen: No, you're right - this is a left over from before I added a "nonce" to the email.

packages/realm-web/src/App.ts
kraenhansen: I guess it depends on what I want to achieve. The semantics are not exactly equivalent.
But ... In this case I probably do want to reset the location if the response doesn't contain a hostname.

packages/realm-web/src/utils/string.ts
kraenhansen: I agree.

packages/realm-web/src/App.ts
steffenagger: The current rejection-handler (2nd arg in the `then`) will also catch a rejection/throw from `fetchJSON`, and reset `_locationUrl`, if that's what you are trying to avoid?

```js
Promise.reject(new Error('NOPE'))
    .then(() => { console.log('THEN') }, (err) => { console.log('NESTED:', err) });
// outputs: NESTED: Error: NOPE
```

Just FYI.

kraenhansen: Yea - I get that 🙂
What I wanted to get across is that I should probably use catch here to ensure the `_locationUrl` is reset *also* for the case where the response is missing the hostname. That case would not had hit the 2nd argument callback in the original code.

steffenagger (approved):

PR #3093, opened 2020-07-28
collection change callback ts declaration fix

steffenagger:
`CollectionChangeSet` edited & used in `CollectionChangeCallback` (looks like a merge slip).

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -26,19 +26,11 @@ declare namespace Realm {
     interface CollectionChangeSet {
         insertions: number[];
         deletions: number[];
-        modifications: number[];
         newModifications: number[];
         oldModifications: number[];
     }
 
-    interface ObjectChanges {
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -26,19 +26,11 @@ declare namespace Realm {
     interface CollectionChangeSet {
         insertions: number[];
         deletions: number[];
-        modifications: number[];
         newModifications: number[];
         oldModifications: number[];
     }
 
-    interface ObjectChanges {
-        insertions: Object[];
```
blagoev, line 35: I see these removed. aren't they needed since these are arrays of Objects and the others are arrays of number?

steffenagger: Those are the old declarations (now known as `CollectionChangeSet `), I suspect a merge slip has occurred.

PR #3096, opened 2020-07-29
use local archive instead of downloading it. Eases debugging js tests

blagoev:
This will extract an already available archive for sync and core if available in user's home directory

### DIFF
--- a/scripts/download-realm.js
+++ b/scripts/download-realm.js
@@ -86,7 +86,7 @@ function copyLocal(archive, destination) {
 
 function download(serverFolder, archive, destination) {
     if (copyLocal(archive, destination)) {
-        return;
+        return Promise.resolve();
     }
      
     const url = `https://static.realm.io/downloads/${serverFolder}/${archive}`;
--- a/scripts/download-realm.js
+++ b/scripts/download-realm.js
@@ -72,7 +72,23 @@ function printProgress(input, totalBytes, archive) {
     });
 }
 
+function copyLocal(archive, destination) {
+    const archivePath = `${process.env.HOME}/${archive}`;
+    const exists = fs.existsSync(archivePath);
+    if (exists) {
+        console.log(`Skipping download for ${archive}. Local archive available at ${archivePath}`);
+        fs.copySync(archivePath, destination);
+        return true;
+    }
+
+    return false;
+}
+
 function download(serverFolder, archive, destination) {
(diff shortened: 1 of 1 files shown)


scripts/download-realm.js
```diff
@@ -72,7 +72,23 @@ function printProgress(input, totalBytes, archive) {
     });
 }
 
+function copyLocal(archive, destination) {
+    const archivePath = `${process.env.HOME}/${archive}`;
```
steffenagger, line 76: `archive` is randomized on each run, so `copyLocal` will only work if `REALM_DOWNLOAD_CORE_TEMP_DIR` environment variable is set (witch it is for testing I guess). But you could just use `REALM_DOWNLOAD_CORE_TEMP_DIR` instead, that might be a bit clearer.

scripts/download-realm.js
```diff
@@ -72,7 +72,23 @@ function printProgress(input, totalBytes, archive) {
...
+        fs.copySync(archivePath, destination);
+        return true;
+    }
+
+    return false;
+}
+
 function download(serverFolder, archive, destination) {
+    if (copyLocal(archive, destination)) {
+        return;
```
steffenagger, line 89: A lot is happening in this file, but not returning a promise would break the promise-chain in the `acquire` function: https://github.com/realm/realm-js/blob/f6638c322e4f94d0bfb30f7db5b0a5311c50b513/scripts/download-realm.js#L152-L159

Would probably be safer to return `Promise.resolve()` instead of `undefined`.

But isn't this something `shouldSkipAcquire` combined with a lock-file should handle?

blagoev: I was looking into the other PR. yes indeed the result should be a resolved promise. pushed a fix for that

scripts/download-realm.js
blagoev: no this is for changing the temp dir. copyLocal is to use an pre downloaded binary to skip downloading it when testing.

scripts/download-realm.js
blagoev: shouldSkipAcquire is when you have already downloaded it once. This means that you should really have the required files in the required places to use it. 

this change allows you to have the tgz downloaded manually to skip downloading the tgz for debugging

scripts/download-realm.js
steffenagger: Ok.

PR #3097, opened 2020-07-29
V10 use local archive

blagoev:
a v10 version of https://github.com/realm/realm-js/pull/3096

### DIFF
--- a/scripts/download-realm.js
+++ b/scripts/download-realm.js
@@ -86,7 +86,7 @@ function copyLocal(archive, destination) {
 
 function download(serverFolder, archive, destination) {
     if (copyLocal(archive, destination)) {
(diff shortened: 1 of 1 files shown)


scripts/download-realm.js
```diff
@@ -72,7 +72,23 @@ function printProgress(input, totalBytes, archive) {
...
+        fs.copySync(archivePath, destination);
+        return true;
+    }
+
+    return false;
+}
+
 function download(serverFolder, archive, destination) {
+    if (copyLocal(archive, destination)) {
+        return;
```
steffenagger: This should return a promise (Just a reminder, same as https://github.com/realm/realm-js/pull/3096).

PR #3103, opened 2020-07-30
Allow realm object inheritance from Realm.Object in TS

### LINKED ISSUE #1226: Object is not a constructor on Typescript import * from "realm"
Version:
- react-native: 0.47.1
- realm: 1.10.3

When use:

```javascript
import * as Realm from "realm"

new Realm({
	schema: [{ name: 'Dog', properties: { name: 'string' } }]
})

// or
Realm.open({
	schema: [{ name: 'Dog', properties: { name: 'string' } }]
})
.then((realm) => {...})
```

Will show error in red screen:

```text
Object is not a constructor (evaluating 'new Realm({
    schema: [{ name: 'Dog', properties: { name: 'string' } }]
})')

// or
Object is not a constructor (evaluating 'new _this(config)')
```
Change import to this type is no problem:
```javascript
import Realm from "realm"
```

But lose type intelligence and get ts error:

```text
Module ''realm'' has no default export.
```

Similar problems in stackoverflow
[Cannot open Realm using typescript 2.4.1 + react native 0.46.1 because Realm is not a constructor](https://stackoverflow.com/questions/45112877/cannot-open-realm-using-typescript-2-4-1-react-native-0-46-1-because-realm-is)

Thanks

steffenagger:
Currently extending `Realm.Object` like so:
```ts
class Foo extends Realm.Object {}
```

Results in this TS compile error:
```zsh
Type '{ readonly prototype: Object; }' is not a constructor function type.
```

(Even with `"allowSyntheticDefaultImports": true` in tsconfig.json, which previously has fixed the problem, see: https://github.com/realm/realm-js/issues/1226)

This closes #1226 - again.


blagoev:
@steffenagger We need a changelog entry with the error we are fixing.

PR #3104, opened 2020-07-30
Realm Web: Unbreaking types

kraenhansen:
## What, How & Why?

A small issue was introduced when merging https://github.com/realm/realm-js/pull/3071, which this PR will resolve.
Merge at will if you need the Realm Web project to build again :)

## ☑️ ToDos
* [ ] 📝 ~Changelog entry~
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web/src/dom/index.ts
+++ b/packages/realm-web/src/dom/index.ts
@@ -16,6 +16,10 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+declare global {
+    type TimerHandle = ReturnType<typeof setTimeout>;
+}
+
 export * from "../index";
 
 import { setEnvironment, Environment } from "../environment";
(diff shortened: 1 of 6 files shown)


packages/realm-web/src/dom/index.ts
```diff
@@ -16,6 +16,10 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+declare global {
+    type TimerHandle = ReturnType<typeof setTimeout>;
```
steffenagger, line 20: This could also just be `number` right? I'm mostly asking as `TimerHandle` is used both in relations to `setTimeout` & `setInterval`, but we use the return type of _one_ of them.

And in `environment.d.ts` TimerHandle is declared as such:
```ts
// Timer related stuff
type TimerHandle = any;
```
Is this intentional?

kraenhansen: Yeah - I could probably just be `number` here, I just mirrored the code I used to infer the type for node.
I would expect the return type of `setTimeout` and `setInterval` to be the same in any relevant runtime. Also, if that turns out not to be the case, any code trying to assign the result of `setInterval` to a `TimerHandle` would fail at compile time.

PR #3108, opened 2020-08-03
update to latest os in master and latest sync (5.0.14) and core (6.0.10)

blagoev:
update to latest os in master and latest sync (5.0.14) and core (6.0.10)


dependencies.list
```diff
@@ -1,5 +1,5 @@
 PACKAGE_NAME=realm-js
 VERSION=6.0.3
-REALM_CORE_VERSION=6.0.9
-REALM_SYNC_VERSION=5.0.8
+REALM_CORE_VERSION=6.0.10
```
steffenagger, line 3: Any reason not to go all the way to `6.0.14`?
(https://github.com/realm/realm-core/tree/v6.0.14)

blagoev: no we should. it was not announced thought

PR #3124, opened 2020-08-12
Use a bash script for building (fixes Xcode builds), detect and use n…

blagoev:
Use a bash script for RN on iOS instead of commands directly.
Detect nvm when node is not available at the current $PATH. Required for building from inside XCode where the shell used is not interactive.

### DIFF
--- a/scripts/xcode-download-realm.sh
+++ b/scripts/xcode-download-realm.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -e
+
+#enable for debugging
+# set -x
+
+export NODE_COMMAND="node"
+
+if ! which node > /dev/null; then
+  echo "warning: node is not found in PATH. Looking for nvm in $HOME/.nvm"
+  [ ! -d "$HOME/.nvm" ] && echo "error: node or nvm not found." && exit 1;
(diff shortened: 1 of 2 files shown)


scripts/xcode-download-realm.sh
```diff
@@ -0,0 +1,27 @@
...
+  [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+  export NODE_CURRENT_VERSION=$(nvm current);  
+  echo "warning: nvm found. Using node version: $NODE_CURRENT_VERSION"
+  # set -x
+  
+  #use nvm to invoke the current node
+  export NODE_COMMAND="nvm run $NODE_CURRENT_VERSION"
+fi
+
+$NODE_COMMAND ${PODS_TARGET_SRCROOT}/scripts/download-realm.js ios --sync
```
kneth, line 25:
```suggestion
$NODE_COMMAND "${PODS_TARGET_SRCROOT}/scripts/download-realm.js" ios --sync
```

blagoev: fixed

PR #3131, opened 2020-08-13
Check for standalone embedded objects

kneth:
## What, How & Why?

It is not allowed to create standalone object for tables marked as embedded. A check is added, and an exception is thrown.

This closes https://jira.mongodb.org/browse/RJS-636 (private ticket).

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


tests/js/list-tests.js
```diff
@@ -1475,6 +1475,28 @@ module.exports = {
             TestCase.assertEqual(addresses[i]["street"], streets[i]);
         }
 
+        // creating standalone embedded object is not allowed
+        let addrs1 = owners[0].addresses;
```
blagoev, line 1479: shouldn't this be a separate test.

blagoev: the logic from line 1478 onwards

tests/js/list-tests.js
```diff
@@ -1475,6 +1475,28 @@ module.exports = {
...
+        realm.write(() => {
+            TestCase.assertThrows(() => {
+                realm.create(schemas.AddressSchema.name, { street: "Njalsgade", city: "Islands Brygge" });
+            });
+
+            TestCase.assertThrows(() => {
+                addrs1.push({ street: "Njalsgade", city: "Islands Brygge" });
```
steffenagger, line 1486: We're not allowed to push to an existing user's list of addresses either?

tests/js/list-tests.js
kneth: Sounds as a good idea

tests/js/list-tests.js
kneth: The above implies that you create an embedded object. Standalone or free-floating embedded objects are not allowed. Moreover, you cannot move an embedded object (removing it from it parent implies that it becomes free-floating).

kneth: See "Semantics" in https://github.com/realm/realm-java/pull/6730

steffenagger: But wouldn't this be equal to:
```java
managedObject.list.add(EmbeddedPerson()) // Will automatically copy the object as embedded
```
 (line from description @ https://github.com/realm/realm-java/pull/6730)

kneth:
@blagoev @steffenagger I have changed it to a draft as I have discovered things we need cover before merging.

blagoev:
LGTM for merge after the comments are resolved

tests/js/list-tests.js
steffenagger: Fine as a temp crash fix.

kneth: We will reintroduce (fix might be a better word) the ability to push to a list later.

kneth:
Tests pass locally (with and without sync)

PR #3134, opened 2020-08-14
proxy fix for internal calls

steffenagger:
This fix binds `callFunction` to `FunctionsFactory`.

A different approach would be to define `callFunction` as a property containing an arrow-function:
```ts
callFunction = async (name, ...args) => {
  // ...
}
```


kraenhansen:
I think a test for this would be great too.

steffenagger:
@kraenhansen added test for `callFunction`

PR #3136, opened 2020-08-14
v10: Fix create performance in RN on iOS

### LINKED ISSUE #2845: Object creating takes very long in ios (react native)
recently updated from 3.6.4 to 5.0.3.
object creating takes much longer that it used to on both simulators and iphones + blocks ui.

## Expected Results
quick creating as it used to be

## Actual Results
average object creation takes about 400ms and blocks the ui

## Code Sample

	postRealm.write(() => {
		if (cleanAll) {
			let allPosts = postRealm.objects('post24').filtered('projectId = "' + projectId + '"');
			let allLastUpdateTS = postRealm.objects('lastUpdateTS').filtered('projectId = "' + projectId + '" AND type = "post"');
			postRealm.delete(allLastUpdateTS);
			postRealm.delete(allPosts); 
		}
		
		preparedPostArray.forEach(post => {
			if (post && post.id) 
				postRealm.create('post24', post, true);
			else {
				console.warn('post missing ID'); 
			}
		});
	
		if (projectId && ignoreTimestamp)
			postRealm.create('lastUpdateTS', { id: projectId + '_post', lastUpdateTS: newLastUpdateTS, projectId, type: 'post' }, true)
	});	

## Versions
- Realm JS SDK Version: 5.0.3
- React Native: 0.61.5
- Client OS & Version: macOs Catalina 10.15.4

blagoev:
fix object create performance regression in RN on iOS
Fixes #2845


blagoev:
Yes, as it turned out in order for the JSC to return a set `JSValueRef exception` on a `JSxxx` call it needs to capture the stack trace which is time consuming operation on iOS apparently. We don't need the exception in this case since we are walking up the prototype chain until no parent.

Issue #3141, opened 2020-08-16
Bug | args.map is not a function


omerts:
Hey, 
Realm version: v10.0.0-beta9

Seems like there is a serious bug around the lib/browser/rpc.js file, many of the auth methods are expecting an args array param, while they are receiving scalar values, which cause an `args.map is not a function` error.
For example (lib/browser/rpc.js line 103: 
```javascript
export function _emailPasswordRPC (args) {
  args = args.map(arg => serialize(null, arg))
  const result = sendRequest('_emailPassword', { arguments: args })
  return deserialize(undefined, result)
}
```

Calling function is lib/browser/credentials.js line 37:

``` javascript
static emailPassword(email, password) {
  return _emailPasswordRPC(email, password);
}
```
so rpc function should either be updated to `...args` or calling function should pass an array.

kraenhansen:
Thanks for reporting this @omerts - this is surely seems like an oversight on our part.

blagoev:
Thanks for reporting. This is fixed by https://github.com/realm/realm-js/pull/3149 and will be part of our next release

PR #3149, opened 2020-08-19
Fix rpc functions

blagoev:
fixes https://github.com/realm/realm-js/issues/3141

### DIFF
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -88,38 +88,37 @@ export function asyncOpenRealm(id, config, callback) {
     }));
 }
 
-export function _anonymousRPC(args) {
-    args = args.map((arg) => serialize(null, arg));
-    const result = sendRequest('_anonymous', { arguments: args });
+export function _anonymousRPC() {
+    const result = sendRequest('_anonymous', { arguments: undefined });
     return deserialize(undefined, result);
 }
 
-export function _appleRPC(args) {
-    args = args.map((arg) => serialize(null, arg));
+export function _appleRPC(token) {
+    const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
     const result = sendRequest('_apple', { arguments: args });
     return deserialize(undefined, result);
 }
 
-export function _emailPasswordRPC(args) {
-    args = args.map((arg) => serialize(null, arg));
+export function _emailPasswordRPC(email, password) {
+    const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
     const result = sendRequest('_emailPassword', { arguments: args });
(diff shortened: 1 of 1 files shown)


kneth (approved):

lib/browser/rpc.js
```diff
@@ -88,38 +88,37 @@ export function asyncOpenRealm(id, config, callback) {
...
-    args = args.map((arg) => serialize(null, arg));
+export function _appleRPC(token) {
+    const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
     const result = sendRequest('_apple', { arguments: args });
     return deserialize(undefined, result);
 }
 
-export function _emailPasswordRPC(args) {
-    args = args.map((arg) => serialize(null, arg));
+export function _emailPasswordRPC(email, password) {
```
kraenhansen, line 102: I suppose these positional arguments will become unused variables. Are there any reason for naming them explicitly instead of simply removing them?

lib/browser/rpc.js
```diff
@@ -88,38 +88,37 @@ export function asyncOpenRealm(id, config, callback) {
...
-    const result = sendRequest('_anonymous', { arguments: args });
+export function _anonymousRPC() {
+    const result = sendRequest('_anonymous', { arguments: undefined });
     return deserialize(undefined, result);
 }
 
-export function _appleRPC(args) {
-    args = args.map((arg) => serialize(null, arg));
+export function _appleRPC(token) {
+    const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
```
steffenagger, line 97: Could also be achieved with:
```js
export function _appleRPC(...args) {
    const result = sendRequest('_apple', {
        arguments: args.map((arg) => serialize(null, arg))
    });
    return deserialize(undefined, result);
```
The arguments are already named in the public methods.

steffenagger (approved):
Works, but could also use rest parameter.

lib/browser/rpc.js
blagoev: thanks, but I don't like this. any method regardless if public or not should name the arguments. Maybe I am to protective, but would you do that if this was TypeScript.

lib/browser/rpc.js
blagoev: any method regardless if public or not should name the arguments.

lib/browser/rpc.js
steffenagger: Yes I would, but also just ES6+. `...args` declares that everything passed to the function is (potentially) used. (In TS you would further declare the types allowed in the args array).
I'm not sure I agree that named arguments is more protective, in this case, as the function disregards them by utilizing `arguments` instead.

PR #3157, opened 2020-08-24
EJSON serialization fix

steffenagger:
EJSON serialization done in canonical mode (non-relaxed).

### DIFF
--- a/packages/realm-web/src/utils/ejson.ts
+++ b/packages/realm-web/src/utils/ejson.ts
@@ -18,6 +18,10 @@
 
 import { EJSON } from "bson";
 
+const SERIALIZATION_OPTIONS = {
+    relaxed: false, // Ensure Canonical mode
+};
+
 /**
  * Serialize an object containing BSON types into extended-JSON.
  *
@@ -25,7 +29,7 @@ import { EJSON } from "bson";
  * @returns The document in extended-JSON format.
  */
 export function serialize(obj: object): object {
-    return EJSON.serialize(obj);
+    return EJSON.serialize(obj, SERIALIZATION_OPTIONS);
 }
 
 /**
--- a/packages/realm-web/CHANGELOG.md
+++ b/packages/realm-web/CHANGELOG.md
@@ -5,7 +5,7 @@
 * None
 
 ### Fixed
-* None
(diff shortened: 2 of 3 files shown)


packages/realm-web/src/utils/ejson.ts
```diff
@@ -34,9 +38,9 @@ export function serialize(obj: object): object {
  * @param obj The object or array of objects in extended-JSON format.
  * @returns The object or array of objects with inflated BSON types.
  */
-export function deserialize<R extends any>(obj: object | object[]): R {
+export function deserialize<R extends any>(obj: object | object[]): R | R[] {
     if (Array.isArray(obj)) {
-        return obj.map(doc => EJSON.deserialize(doc)) as any;
+        return obj.map(doc => EJSON.deserialize(doc)) as R[];
```
kraenhansen, line 43: I don't like this change. With this it's no longer possible to lock down the return type to not return an array.
It was the intent that if an array was returned, `R` needed to be an array and the consumer of the API was responsible for providing the correct return type (which was potentially an array of something).

kraenhansen (changes requested):

packages/realm-web/src/utils/ejson.ts
```diff
@@ -34,9 +38,9 @@ export function serialize(obj: object): object {
  * @param obj The object or array of objects in extended-JSON format.
  * @returns The object or array of objects with inflated BSON types.
  */
-export function deserialize<R extends any>(obj: object | object[]): R {
+export function deserialize<R extends any>(obj: object | object[]): R | R[] {
```
kraenhansen, line 41: With this it's no longer possible to lock down the return type to not return an array.
It was the intent that if an array was returned, `R` needed to be an array and the consumer of the API was responsible for providing the correct return type (which was potentially an array of something).

kraenhansen, line 41: With this it's no longer possible to lock down the return type to not return an array.
It was the intent that if an array was returned, `R` needed to be an array and the consumer of the API was responsible for providing the correct return type (which was potentially an array of something).

kraenhansen (changes requested):

kraenhansen (approved):

PR #3161, opened 2020-08-26
Realm Web: Fixing install races & clean-up Lerna packages

kraenhansen:
## What, How & Why?

While pair-programming with @steffenagger we made a few changes in an attempt to clean-up and make the Lerna multi-package structure more resilient.

It introduces a three changes:

### 1. Removing the "file path" dependency on shared development dependencies
The "file path" dependency from the Lerna packages on TypeScript and ESLint was introduced to limit confusion from using different versions of these tools in the sub-packages (and potentially saving a few MB on developers machines by reusing packages). However it turns out NPM+Lerna doesn't play nice with this, as mutations are made to a package directory while installing it as a path dependency. Hence, installing it from multiple sub-packages concurrently resulted in non-deterministic errors while bootstrapping the Lerna project, which grew in frequency as more packages were added.

### 2. It removes the root package (`realm`) as a (Lerna) sub-package (in /packages)
Originally, this was most likely introduced as a means of depending on the realm-network-package before it was released on NPM. One issue of having it a lerna sub-package is that the root package needs to be installed first to get access to the Lerna CLI and then bootstrapping will re-install the root package. Another issue is that we cannot register a "prepare" script that bootstraps sub-packages, if the root package is a sub-package itself, since that would result in an endless install loop.

### 3. Adding a "prepare" script bootstrapping Lerna sub-packages (in /packages) when the root package installs
To make it easier for developers to get the sub-packages installed, this PR adds a "prepare" script (which runs after local "npm install", but not when a user of the package installs it, as postinstall would) to the root project which bootstraps the sub-packages. This addition is debatable, but it seems like a valuable addition to me.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "lint": "eslint",
     "test": "scripts/test.sh",
     "package": "npm ci --build-from-source=realm && node-pre-gyp package",
+    "prepare": "lerna bootstrap",
     "install": "node-pre-gyp install --fallback-to-build -j 8",
     "rebuild": "npm run isWin -s && SET REALMJS_USE_DEBUG_CORE=1 && npm run rebuild-changes || REALMJS_USE_DEBUG_CORE=1 && npm run rebuild-changes",
     "build": "npm run isWin -s && SET REALMJS_USE_DEBUG_CORE=1 && npm run build-changes || REALMJS_USE_DEBUG_CORE=1 && npm run build-changes",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "lint": "eslint",
     "test": "scripts/test.sh",
     "package": "npm ci --build-from-source=realm && node-pre-gyp package",
-    "prepare": "lerna bootstrap",
     "install": "node-pre-gyp install --fallback-to-build -j 8",
     "rebuild": "npm run isWin -s && SET REALMJS_USE_DEBUG_CORE=1 && npm run rebuild-changes || REALMJS_USE_DEBUG_CORE=1 && npm run rebuild-changes",
     "build": "npm run isWin -s && SET REALMJS_USE_DEBUG_CORE=1 && npm run build-changes || REALMJS_USE_DEBUG_CORE=1 && npm run build-changes",
--- a/packages/realm-web/package.json
+++ b/packages/realm-web/package.json
@@ -60,7 +60,7 @@
     "@types/node": "^13.7.6",
     "abort-controller": "^3.0.0",
     "chai": "^4.2.0",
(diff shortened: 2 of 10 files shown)


steffenagger (approved):
So boiled down this results in `npm i` (`npm install` for the verbose), installing dependencies for realm-js (root), and all /packages, and do so without race conditions, while avoiding weird updates to `package-lock.json`'s. (right?)

This seemed a lot less brittle (than `npm i && npx lerna bootstrap --concurrency=1`, and a bit easier to "remember") when we were trying it out. Only thing is - I don't know if `prepare` has any side effects?

_(integration-tests also needs some inclusion in this, but that will be another time)_

kneth (approved):

kneth:
> Originally, this was most likely introduced as a means of depending on the realm-network-package before it was released on NPM.

I believe you are right - adding `.` as a package was a workaround to get things to work before the packages were stabilized (in particular `realm-network-transport`).

package.json
```diff
@@ -52,6 +52,7 @@
     "lint": "eslint",
     "test": "scripts/test.sh",
     "package": "npm ci --build-from-source=realm && node-pre-gyp package",
+    "prepare": "lerna bootstrap",
```
blagoev, line 55: this reflects on the Realm package. I would prefer we don't introduce such changes until we have time to fix our repo with our long term decision. This task was dropped in the Q3 planning.

blagoev (changes requested):
I would prefer not to make this change until when we have planned our repo discussions and tasks on repo restructure for the long term

package.json
kraenhansen:
```suggestion
```

packages/realm-web/package.json
```diff
@@ -43,7 +43,8 @@
   "dependencies": {
     "@types/bson": "^4.0.0",
     "bson": "4.0.0",
-    "js-base64": "^2.6.3"
+    "js-base64": "^2.6.3",
+    "realm-network-transport": "^0.6.0"
```
blagoev, line 47: should this be updated to 0.7

blagoev (approved):

packages/realm-web/package.json
kraenhansen: This is super strange - I thought that was what I did. And it should remain a development dependency, since it'll get bundled in. Thanks for catching this!

kraenhansen: I'm confusing myself here ... it needs to be v0.7.0 now that https://github.com/realm/realm-js/pull/3163 merged - I'll rebase and fix this.

PR #3188, opened 2020-09-02
Fix building if in a path with a space in it

richardgroves:
The build phase calling the download-realm.js script failed to wrap the path in "s

## What, How & Why?
Wrap a path in "s so it builds without error when the project is in a path with a space in it.


kneth:
@richardgroves 

PR #3205, opened 2020-09-09
repopulate partitionValue on syncSession (v10)

### LINKED ISSUE #3155: MongoDB realm - realm.syncSession.config does not return entire config
## Goals

Trying to retrieve the current partitionValue from a sync session configuration on a realm instance.

## Expected Results

`realm.syncSession.config.partitionValue` to return the value of the partition key. 

## Actual Results

returns `undefined`

## Steps to Reproduce

run `realm.syncSession.config.partitionValue` on a synced realm.

- Realm JS SDK Version: 10.0.0-beta.9
- Node: v12.13.0
- Client OS & Version: osx 10.13.6

steffenagger:
This closes #3155

Besides _"repopulate partitionValue on syncSession"_, this PR adds checks for `partitionValue`:
- string: must not be empty.
- number: must be a whole number `>= Number.MIN_SAFE_INTEGER` and `<= Number.MAX_SAFE_INTEGER`.

In particular I'm worried I've overlooked something in regards to changes to: src/js_types.hpp:
When converting from Int64 to a number, we switch to Long. Doing so we have a `max_precise_double`, but this doesn't align with `Number.MAX_SAFE_INTEGER`, which is defined as `Math.pow(2, 53) - 1`.
```c++
constexpr static int64_t max_precise_double = 1ll << 52;
```

### NOTE: Missing CHANGELOG entry.

### DIFF
--- a/src/js_sync.hpp
+++ b/src/js_sync.hpp
@@ -62,11 +62,18 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
         bson::Bson partition_bson;
         if (Value<T>::is_string(ctx, partition_value_value)) {
             std::string pv = Value<T>::validated_to_string(ctx, partition_value_value);
+            if (pv.length() == 0) {
+                throw std::runtime_error("partitionValue of type 'string' cannot be an empty string.");
+            }
             partition_bson = bson::Bson(pv);
         }
         else if (Value<T>::is_number(ctx, partition_value_value)) {
             auto pv = Value<T>::validated_to_number(ctx, partition_value_value);
-            partition_bson = bson::Bson(static_cast<int64_t>(pv));
+            auto pvi = static_cast<int64_t>(pv);
+            if (pv != pvi || pvi < 0) {
+                throw std::runtime_error("partitionValue of type 'number' must be a whole positive number.");
+            }
+            partition_bson = bson::Bson(pvi);
         }
         else if (Value<T>::is_object_id(ctx, partition_value_value)) {
             auto pv = Value<T>::validated_to_object_id(ctx, partition_value_value);
@@ -88,6 +95,7 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
 
 
 using WeakSession = std::weak_ptr<realm::SyncSession>;
+using bson::Bson;
 
 template<typename T>
 class SessionClass : public ClassDefinition<T, WeakSession> {
--- a/src/js_sync.hpp
+++ b/src/js_sync.hpp
@@ -321,6 +321,8 @@ void SessionClass<T>::get_config(ContextType ctx, ObjectType object, ReturnValue
         ObjectType config = Object::create_empty(ctx);
         Object::set_property(ctx, config, "user", create_object<T, UserClass<T>>(ctx, new User<T>(session->config().user, nullptr))); // FIXME: nullptr is not an app object
         // TODO: add app id
+        bson::Bson partition_value_bson = bson::parse(session->config().partition_value);
+        Object::set_property(ctx, config, "partitionValue", Value::from_bson(ctx, partition_value_bson));
         if (auto dispatcher = session->config().error_handler.template target<util::EventLoopDispatcher<SyncSessionErrorHandler>>()) {
             if (auto handler = dispatcher->func().template target<SyncSessionErrorHandlerFunctor<T>>()) {
                 Object::set_property(ctx, config, "error", handler->func());
--- a/src/js_sync.hpp
+++ b/src/js_sync.hpp
@@ -18,6 +18,8 @@
 
 #pragma once
 
+#include <math.h>
+
 #include "js_class.hpp"
 #include "js_collection.hpp"
 #include "js_app.hpp"
@@ -69,10 +71,10 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
         }
         else if (Value<T>::is_number(ctx, partition_value_value)) {
             auto pv = Value<T>::validated_to_number(ctx, partition_value_value);
-            auto pvi = static_cast<int64_t>(pv);
-            if (pv != pvi || pvi < 0 || pvi > JS_MAX_SAFE_INTEGER) {
-                throw std::runtime_error("partitionValue of type 'number' must be a whole number > 0 and <= Number.MAX_SAFE_INTEGER.");
+            if (pv < 0 || pv > JS_MAX_SAFE_INTEGER || fmod(pv, 1) != 0) {
+                throw std::runtime_error("partitionValue of type 'number' must be a whole number >= 0 and <= Number.MAX_SAFE_INTEGER.");
             }
+            auto pvi = static_cast<int64_t>(pv);
             partition_bson = bson::Bson(pvi);
         }
         else if (Value<T>::is_object_id(ctx, partition_value_value)) {
--- a/src/js_sync.hpp
+++ b/src/js_sync.hpp
@@ -70,12 +70,13 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
             partition_bson = bson::Bson(pv);
         }
         else if (Value<T>::is_number(ctx, partition_value_value)) {
-            auto pv = Value<T>::validated_to_number(ctx, partition_value_value);
-            if (pv < 0.0 || pv > JS_MAX_SAFE_INTEGER || fmod(pv, 1.0) > 0.0) {
+            double pv = Value<T>::validated_to_number(ctx, partition_value_value);
+            double integerPart;
+            double fractionalPart = modf(pv, &integerPart);
+            if (fractionalPart > 0.0 || pv < 0.0 || pv > JS_MAX_SAFE_INTEGER) {
                 throw std::runtime_error("partitionValue of type 'number' must be a non-negative integer <= Number.MAX_SAFE_INTEGER.");
             }
-            auto pvi = static_cast<int64_t>(pv);
-            partition_bson = bson::Bson(pvi);
+            partition_bson = bson::Bson(static_cast<int64_t>(integerPart));
         }
         else if (Value<T>::is_object_id(ctx, partition_value_value)) {
             auto pv = Value<T>::validated_to_object_id(ctx, partition_value_value);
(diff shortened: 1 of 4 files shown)


src/js_sync.hpp
```diff
@@ -62,11 +64,18 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
...
+            if (pv.length() == 0) {
+                throw std::runtime_error("partitionValue of type 'string' cannot be an empty string.");
+            }
             partition_bson = bson::Bson(pv);
         }
         else if (Value<T>::is_number(ctx, partition_value_value)) {
             auto pv = Value<T>::validated_to_number(ctx, partition_value_value);
-            partition_bson = bson::Bson(static_cast<int64_t>(pv));
+            if (pv < 0 || pv > JS_MAX_SAFE_INTEGER || fmod(pv, 1) != 0) {
+                throw std::runtime_error("partitionValue of type 'number' must be a whole number >= 0 and <= Number.MAX_SAFE_INTEGER.");
```
kneth, line 75: I don't think "whole number" is a well-defined mathematical term. Maybe "natural number", "non-negative integer" or just "integer".

tests/js/session-tests.js
```diff
@@ -829,20 +830,57 @@ module.exports = {
...
     },
 
-    testNullPartitionValue() {
-        const app = new Realm.App(appConfig);
-        const credentials = Realm.Credentials.anonymous();
+    async testAcceptedPartitionValueTypes() {
+        const testPartitionValues = [
+            Utils.genPartition(), // string
+            0,
+            Math.floor(Math.random() * Number.MAX_SAFE_INTEGER),
```
kneth, line 837: In general we should avoid random test input. It will make it much harder to reproduce a failing test.

kneth (approved):

src/js_sync.hpp
steffenagger: https://dictionary.cambridge.org/dictionary/english/whole-number 🤷‍♂️
Then I would suggests `"partitionValue of type 'number' must be a non-negative integer <= Number.MAX_SAFE_INTEGER."`?

tests/js/session-tests.js
steffenagger: This was mostly to not only test the extremes, I'll replace it with a static "random" small number & a large number.

src/js_sync.hpp
fronck: I don't think the `fmod(pv, 1) != 0` guard is safe across compilers and architectures.  It may perform a division (is the user has't done so already), which is not guaranteed to return exact integers, even if the maths say it should.

I think a safer approach would be something like 

```
            auto pv = Value<T>::validated_to_number(ctx, partition_value_value);
            decltype(pv) fraction;
            partition_bson = bson::Bson(static_cast<int64_t>(pv));
            if (pv < 0 || pv > JS_MAX_SAFE_INTEGER || modf(pv, *fraction) > std::numeric_limits<decltype(pv)>::epsilon(); {
               ...
```

tests/js/session-tests.js
fronck: ..  Or, alternatively, pre-generate a set of 40 random numbers, add the the extremes to the list and test on those..

fronck:
LGTM if the `fmod()` comment is considered.

tests/js/session-tests.js
steffenagger: That was the idea, although where you thought 40, I thought 2 - I'll try to find something mid way.

src/js_types.hpp
```diff
@@ -616,8 +619,7 @@ inline typename T::Value Value<T>::from_bson(typename T::Context ctx, const bson
         // it to a plain js number if it is in the range where it can be done precisely, otherwise
         // we map to the bson.Long type which preserves the value, but is harder to use.
         const auto i64_val = value.operator int64_t();
-        constexpr static int64_t max_precise_double = 1ll << 52; // 52 bits mantissa + implicit leading 1 bit.
-        if (-max_precise_double <= i64_val && i64_val <= max_precise_double)
+        if (-JS_MAX_SAFE_INTEGER <= i64_val && i64_val <= JS_MAX_SAFE_INTEGER)
```
fronck, line 622: Does JS have an actual definitoin of "min safe integer"  (these are usually something like `-JS_MAX_SAFE_INTEGER - 1`)?

fronck (approved):
LGTM if comments are addressed.

src/js_sync.hpp
steffenagger: Thanks @fronck that's just what I needed. I'll test it out & update.

src/js_types.hpp
steffenagger: Yes `-Number.MAX_SAFE_INTEGER === Number.MIN_SAFE_INTEGER` => `true`.

fronck: Great

src/js_sync.hpp
steffenagger: @fronck I've updated to the solution we discussed elsewhere.

steffenagger: @cesarvr the last "stringification" you mention is into a bson/ejson format, so the type is preserved at that point.

src/js_sync.hpp
```diff
@@ -62,11 +64,18 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
...
             std::string pv = Value<T>::validated_to_string(ctx, partition_value_value);
+            if (pv.length() == 0) {
+                throw std::runtime_error("partitionValue of type 'string' may not be an empty string.");
+            }
             partition_bson = bson::Bson(pv);
         }
         else if (Value<T>::is_number(ctx, partition_value_value)) {
             auto pv = Value<T>::validated_to_number(ctx, partition_value_value);
-            partition_bson = bson::Bson(static_cast<int64_t>(pv));
+            if (pv < 0.0 || pv > JS_MAX_SAFE_INTEGER || fmod(pv, 1.0) > 0.0) {
```
blagoev, line 74: why do we need to validate the partition value to be an integer.

src/js_sync.hpp
```diff
@@ -62,11 +64,18 @@ static std::string partition_value_bson_to_string(typename T::Context ctx, typen
...
+                throw std::runtime_error("partitionValue of type 'number' must be a non-negative integer <= Number.MAX_SAFE_INTEGER.");
+            }
+            auto pvi = static_cast<int64_t>(pv);
```
blagoev, line 77: I would use `modf` and cast the integral part instead of casting pv. 
http://www.cplusplus.com/reference/cmath/modf/
```
double integerPart;
fractionalPart = modf(value, &integerPart);
```
and compare using the `fractionalPart > 0.0` and cast the `integerPart`

src/js_sync.hpp
```diff
@@ -321,6 +330,8 @@ void SessionClass<T>::get_config(ContextType ctx, ObjectType object, ReturnValue
         ObjectType config = Object::create_empty(ctx);
         Object::set_property(ctx, config, "user", create_object<T, UserClass<T>>(ctx, new User<T>(session->config().user, nullptr))); // FIXME: nullptr is not an app object
         // TODO: add app id
+        bson::Bson partition_value_bson = bson::parse(session->config().partition_value);
```
blagoev, line 333: why do we do a roundtrip to/from Bson. Can't we just use the `partition_value` and set it to `Object::set_property(ctx, config, "partitionValue"`

blagoev (commented):
I have some comments before I LGTM

src/js_sync.hpp
steffenagger: As far as I'm informed, Sync doesn't support other numbers.

src/js_sync.hpp
steffenagger: `session->config().partition_value` is a serialized bson/ejson string (I believe this is a common interface for all SDKs to OS/Sync (?)), we could store the original value somewhere - but after going through the different approaches with @kneth, I went with parsing the value used by Sync.

src/js_sync.hpp
steffenagger: @fronck also suggested `modf`. I've updated to this way instead, but wouldn't the result be the same?

src/js_sync.hpp
blagoev: 👍

src/js_sync.hpp
blagoev: 👍

blagoev (approved):

src/js_sync.hpp
cesarvr: Hey @steffenagger that's ok, just wanted to understand that decimal removal validation when the value that comes is a ``number``.  

If you just want only whole numbers you can cast to ``int64_t`` and the number will lost decimal information and you just need to validate for the numerical constraint. 

```cpp
auto number = static_cast<int64_t>(pv)
if (pv < 0 || pv > JS_MAX_SAFE_INTEGER) {
 //throw exception.... 
}
partition_bson = bson::Bson(number);
```

If throwing exception on a decimal is a must then I would suggest: 

```cpp 
  //#include <cmath>
 //  for pv = 10.5;                                                                    
 if (pv < 0 || pv > JS_MAX_SAFE_INTEGER || std::floor(pv) != pv ) { // 10 != 10.5 
  //throw exception.... 
 }
 partition_bson = bson::Bson(pv);
```
I my opinion is simpler way to achieve that goal.

Cheers.

steffenagger: Of course, ask away 🙂 (I'll answer to the best of my knowledge).

The exception is a must, as we want to inform the users if they are supplying an unsupported number-format (it would otherwise be completely hidden that we converted the inputted `10.5` to `10`).

The  `int64_t` cast is there to trigger the `Bson` constructor into creating a `Long`  for Sync (not a `Double`).

I do like the simpler `std::floor(pv) != pv` check, @blagoev & @fronck any comments on this?

cesarvr (approved):

src/js_sync.hpp
steffenagger: Seems like my information was wrong, will update...

src/js_sync.hpp
cesarvr: Nice, thanks for the update. 👍

kneth (approved):

PR #3211, opened 2020-09-11
Disallow querying embedded objects

kneth:
## What, How & Why?

The original scope document states that it is not possible to query embedded objects directly.

This closes https://jira.mongodb.org/browse/RJS-763

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


kraenhansen:
I believe this will require UI changes in Studio.

kneth:
> This is only when querying the "root collection" of a type of embedded objects, right? 

Yes, that is correct.

> I believe this will require UI changes in Studio

I guess you will not be able to show the table alone but will always have embedded objects as part of a parent object.

kneth:
Tests (including sync) pass locally on MacOS.

blagoev (approved):

PR #3212, opened 2020-09-11
Document sync logger more carefully

kneth:
## What, How & Why?

This closes https://jira.mongodb.org/browse/RJS-672

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -168,8 +168,16 @@ class Sync {
     static reconnect() { }
 
     /**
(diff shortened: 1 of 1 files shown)


steffenagger (commented):
In general I think we should use `const` and not `let` in examples.

With that said.. If the `Realm.Sync.setLogLevel` can only be set before opening the Ream, shouldn't it be a part of the config, and set internally?

cbush (approved):

Issue #2982, opened 2020-06-16
use single binary for every major nodejs version


RealmBot:
Build and reuse the realm library for every nodejs version

RealmBot:
➤ Lyubomir Blagoev commented:

PR: https://github.com/realm/realm-js/pull/3218

PR #3218, opened 2020-09-14
use single binary for Node.js and Electron.

blagoev:
Publish single Node.js N-API binary for every Realm version.

fixes https://github.com/realm/realm-js/issues/2982


dependencies.list
```diff
@@ -1,5 +1,6 @@
 PACKAGE_NAME=realm-js
-VERSION=6.1.1
+VERSION=6.1.2-test
 REALM_CORE_VERSION=6.0.25
 REALM_SYNC_VERSION=5.0.22
 REALM_OBJECT_SERVER_VERSION=3.23.1
+NAPI_VERSION=4
```
kneth, line 6: Don't we need to document which node versions, we no longer support?

kneth (approved):

PR #3235, opened 2020-09-18
Fix type/objectType for lists of ObjectId/Decimal128

steffenagger:
*Issue:*
Lists of `ObjectId` or `Decimal128` still returns `'object id'` & `'decimal'` instead of `'objectId'` & `'decimal128'`, when checking `type`/`objectType`.

*Solution:*
A util method `local_string_for_property_type` has been introduced, wrapping OS `string_for_property_type` - this is now used instead of `string_for_property_type`.
Tests expanded to include checks for `objectType`.

### DIFF
--- a/src/js_schema.hpp
+++ b/src/js_schema.hpp
@@ -145,7 +145,7 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
             prop.object_type = "";
         }
         else if (prop.object_type == "objectId") {
-            prop.type |= PropertyType::Decimal | PropertyType::Array;
+            prop.type |= PropertyType::ObjectId | PropertyType::Array;
             prop.object_type = "";
         }
         else {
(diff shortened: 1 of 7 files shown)


blagoev (approved):

src/js_schema.hpp
```diff
@@ -145,7 +145,7 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
             prop.object_type = "";
         }
         else if (prop.object_type == "objectId") {
-            prop.type |= PropertyType::Decimal | PropertyType::Array;
+            prop.type |= PropertyType::ObjectId | PropertyType::Array;
```
kneth, line 148: Good catch!

kneth (approved):

kneth:
Please add an entry in `CHANGELOG.md`: fixed a bug where `.type` is incorrect for some property types.

PR #3239, opened 2020-09-20
Realm Web: Fixing linking with OAuth2 credentials

### LINKED ISSUE #3238: Realm Web: OAuth 2.0 credentials doesn't link correctly
## Goals

When linking an anonymous user with an OAuth 2.0 authentication provider, the id stored in the access token changes, but the `id` property on the user object in the SDK doesn't.

I suspect this is due to a broken implementation of credential linking in combination with the OAuth 2.0 authentication flow.

## Expected Results

The `id` stored on the user should match the id stored in the access token.

## Actual Results

The `id` stored on the user has the value associated with the anonymous user before linking.

## Steps to Reproduce

1. Setup a Realm app with anonymous auth and google auth providers enabled.
2. Create an anonymous
3. Link the user with Google credentials
4. Observe how the id in the access token no longer matches that of the user object.

## Code Sample

Here's a failing test (leaving extractIdFromToken as an exercise parsing JWTs).

```javascript
const anonymousCredentials = Credentials.anonymous();
const user = await app.logIn(anonymousCredentials);
const redirectUrl = window.location.origin + "/google-callback";
const googleCredentials = Credentials.google(redirectUrl);
await user.linkCredentials(googleCredentials);
const idFromToken = extractIdFromToken(user.accessToken);
assert(user.id).equals(idFromToken);
```

- Realm JS SDK Version: Realm Web 0.8.1
- Node or React Native: Web
- Client OS & Version: N/A

kraenhansen:
## What, How & Why?

This closes #3238 by ensuring the redirect URL is fetched and linking performed with the correct user's token.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web-integration-tests/src/index.ts
+++ b/packages/realm-web-integration-tests/src/index.ts
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import { MochaRemoteClient } from "mocha-remote-client";
-import Realm from "realm-web";
+import * as Realm from "realm-web";
 
 if (location.pathname.endsWith("-callback")) {
     console.log("This is the callback from Google OAuth 2.0 flow");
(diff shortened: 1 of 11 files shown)


kraenhansen:
@steffenagger please review in retrospect.

PR #3243, opened 2020-09-21
Fix: premature TS deprecation warning

steffenagger:
## What, How & Why?
When using `create<T>(...)` without the 3rd optional argument, the deprecation warning for `create<T>(..., update?: boolean)` currently triggers.
This change just reorders the TS declarations for create<T>(...), to not trigger the deprecation warning prematurely.

Quite visual in the lastest vscode:

* [ ] jsdoc files updated


kraenhansen (approved):
Havn't tested it myself, but the change seems legit! :)

PR #3090, opened 2020-07-24
Initial impl of realm-web watch()

RedBeard0531:
This isn't ready for review yet, but it is mostly functional. I'm mostly putting this up for visibility in case someone needs it while I'm away, or something happens to my laptop

## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 🚦 Tests
* [ ] 📝 Public documentation PR created or is not necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [ ] typescript definitions file is updated
* [ ] jsdoc files updated
* [ ] Chrome debug API is updated if API is available on React Native

PR #3247, opened 2020-09-21
Realm Web: Adding watch to MongoDB collections

### LINKED ISSUE #2969: Realm Web: Implement "watch" method on the (remote) MongoDB Client
This was initially left out to simplify the network-transport, but we need the (remote) MongoDB Client to be able to watch a query for changes.

kraenhansen:
## What, How & Why?

This is a (somewhat brutal) takeover of #3090 in an attempt to close #2969 ..

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
+++ b/packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
@@ -72,7 +72,10 @@ describe("Remote MongoDB", () => {
 
     it("can find documents", async () => {
         const collection = getCollection();
-        const result = await collection.find({}, { limit: 10 });
+        const result = await collection.find(
+            { name: { $exists: true } },
+            { limit: 10 },
+        );
         if (result.length > 0) {
             const [firstDocument] = result;
             expect(typeof firstDocument._id).equal("object");
@@ -320,7 +323,7 @@ describe("Remote MongoDB", () => {
         expect(typeof result.upsertedId).equals("object");
     });
 
-    it("can watch changes correctly",async function() {
+    it("can watch changes correctly", async function () {
         this.timeout(10_000);
         this.slow(2_000);
 
@@ -330,50 +333,60 @@ describe("Remote MongoDB", () => {
 
         await collection.deleteMany({});
 
-        const sleep = async (time: number) => new Promise(resolve => setInterval(resolve, time));
-        const str = 'use some odd chars to force weird encoding %\n\r\n\\????>>>>';
+        const sleep = async (time: number) =>
+            new Promise(resolve => setInterval(resolve, time));
+        const str =
+            "use some odd chars to force weird encoding %\n\r\n\\????>>>>";
         await Promise.all([
             (async () => {
                 // There is a race with creating the watch() streams, since they won't
-                // see inserts from before they are created. 
+                // see inserts from before they are created.
                 // Wait 500ms (490+10) before first insert to try to avoid it.
                 await sleep(490);
                 for (let i = 0; i < 10; i++) {
                     await sleep(10);
-                    await collection.insertOne({_id: i, hello: "world", str});
+                    await collection.insertOne({ _id: i, hello: "world", str });
                 }
-                await collection.insertOne({_id: 'done', done: true}); // break other sides out of loop
+                await collection.insertOne({ _id: "done", done: true }); // break other sides out of loop
             })(),
+            // Watch any event
             (async () => {
                 let expected = 0;
-                for await (let event of collection.watch() as any) {
-                    if (event.fullDocument.done)
-                        break;
+                for await (const event of collection.watch() as any) {
+                    if (event.fullDocument.done) break;
                     expect(event.fullDocument._id).equals(expected++);
                 }
                 expect(expected).equals(10);
             })(),
+            // Watch using the filter option
             (async () => {
-                const filter = {$or:[
-                    {'fullDocument._id': 3, 'fullDocument.str': str},
-                    {'fullDocument.done': true},
-                ]}
+                const filter = {
+                    $or: [
+                        { "fullDocument._id": 3, "fullDocument.str": str },
+                        { "fullDocument.done": true },
+                    ],
+                };
                 let seenIt = false;
-                for await (let event of collection.watch({filter}) as any) {
-                    if (event.fullDocument.done)
-                        break;
-                    expect(event.fullDocument._id).equals(3);
-                    seenIt = true;
+                for await (const event of collection.watch({ filter })) {
+                    if (event.operationType === "insert") {
+                        if (event.fullDocument.done) break;
+                        expect(event.fullDocument._id).equals(3);
+                        seenIt = true;
+                    }
                 }
                 expect(seenIt, "seenIt for filter");
             })(),
+            // Watch using the ids option
             (async () => {
                 let seenIt = false;
-                for await (let event of collection.watch({ids: [5, 'done']}) as any) {
-                    if (event.fullDocument.done)
-                        break;
-                    expect(event.fullDocument._id).equal(5);
-                    seenIt = true;
+                for await (const event of collection.watch({
+                    ids: [5, "done"],
+                })) {
+                    if (event.operationType === "insert") {
+                        if (event.fullDocument.done) break;
+                        expect(event.fullDocument._id).equal(5);
+                        seenIt = true;
+                    }
                 }
                 expect(seenIt, "seenIt for ids");
             })(),
@@ -391,5 +404,5 @@ describe("Remote MongoDB", () => {
         if (err.code != 401)
             throw err;
             */
-  });
+    });
 });
--- a/packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
+++ b/packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
@@ -352,9 +352,15 @@ describe("Remote MongoDB", () => {
             // Watch any event
             (async () => {
                 let expected = 0;
-                for await (const event of collection.watch() as any) {
-                    if (event.fullDocument.done) break;
-                    expect(event.fullDocument._id).equals(expected++);
+                for await (const event of collection.watch()) {
+                    if (event.operationType === "insert") {
+                        if (event.fullDocument.done) break;
+                        expect(event.fullDocument._id).equals(expected++);
+                    } else {
+                        throw new Error(
+                            `Unexpected ${event.operationType} event`,
+                        );
+                    }
                 }
                 expect(expected).equals(10);
             })(),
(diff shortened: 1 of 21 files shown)


packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
```diff
@@ -319,4 +322,101 @@ describe("Remote MongoDB", () => {
...
+            // Watch using the filter option
+            (async () => {
+                const filter = {
+                    $or: [
+                        { "fullDocument._id": 3, "fullDocument.str": str },
+                        { "fullDocument.done": true },
+                    ],
+                };
+                let seenIt = false;
+                for await (const event of collection.watch({ filter })) {
```
kraenhansen, line 376: This is using the async generator API which might be too exotic or not a good fit with a React app - see https://jira.mongodb.org/browse/RJS-462

packages/realm-web/src/FunctionsFactory.ts
```diff
@@ -180,4 +182,34 @@ export class FunctionsFactory {
...
     }
+
+    /**
+     * Call a remote function by it's name.
+     *
+     * @param name Name of the remote function.
+     * @param args Arguments to pass to the remote function.
+     * @returns A promise of the value returned when executing the remote function.
+     */
+    public callFunctionStreaming(
```
steffenagger, line 193: I'm guessing this only works if the function supports streaming? If so, maybe rename to `callStreamableFunction`?

Streaming functions are then not supported through the proxied function calls, correct? We probably need to document that somewhere.

packages/realm-web/src/User.ts
```diff
@@ -346,6 +347,10 @@ export class User<
         throw new Error("Not yet implemented");
     }
 
+    public remoteMongoClient(serviceName: string): Realm.Services.MongoDB {
```
steffenagger, line 350: Wasn't `remoteMongoClient` renamed to `mongoClient` at some point?

packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
blagoev: +1

blagoev: This is tedious to use in non async functions

blagoev: but I like how @RedBeard0531 remembers to use these features, cause for me this does not even exist :)

blagoev (commented):
I would consider using an event emitter in place of the async generator

packages/realm-web/src/WatchStream.ts
```diff
@@ -0,0 +1,270 @@
...
+                if (typeof errorCode != "string") return;
+                if (typeof error != "string") return;
+                // XXX in realm-js, object-store will error if the error_code is not one of the known
+                // error code enum values.
+                this._error = { err: errorCode, message: error };
+            } catch {
+                return; // Use the default state.
+            }
+        } else {
+            // Ignore other event types
```
steffenagger, line 186: Does this comment need it own else clause?

packages/realm-web/src/WatchStream.ts
```diff
@@ -0,0 +1,270 @@
...
+            // line is ignored (even if field is id or retry).
+        }
+    }
+
+    feedSse(sse: ServerSentEvent) {
+        this.assertState(WatchStreamState.NEED_DATA);
+        const firstPercentIndex = sse.data.indexOf("%");
+        if (firstPercentIndex != -1) {
+            // For some reason, the stich server decided to add percent-encoding for '%', '\n', and '\r' to its
+            // event-stream replies. But it isn't real urlencoding, since most characters pass through, so we can't use
```
steffenagger, line 115: Did we notify the server team, this sounds like a partial double encoding...

packages/realm-web/src/WatchStream.ts
```diff
@@ -0,0 +1,270 @@
...
+                start = percent + encoded.length;
+            }
+
+            sse.data = buffer;
+        }
+
+        if (
+            sse.eventType === undefined ||
+            sse.eventType.length == 0 ||
+            sse.eventType == "message"
```
steffenagger, line 150: Any reason not just to check by:
```ts
!sse.eventType || sse.eventType === "message"
```
... for example: would `null` imply a special meaning?

types/app.d.ts
```diff
@@ -451,6 +451,17 @@ declare namespace Realm {
...
+        /**
+         * Returns a connection to the MongoDB service.
+         *
+         * @example
+         * let blueWidgets = user.remoteMongoClient('myClusterName')
+         *                       .db('myDb')
+         *                       .collection('widgets')
+         *                       .find({color: 'blue'});
+         */
+        remoteMongoClient(serviceName: string): Realm.Services.MongoDB;
```
steffenagger, line 464: I think this was renamed to `mongoClient`?

steffenagger (commented):
I think your worries around using async generator API is well founded.

packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
kraenhansen: Personally I feel we should release as is, to get the feature out (users have been waiting for this quite a while) and then actively pursue feedback on this API.

I don't think it would be hard to change this into an event emitter, but it would have to be done for Realm JS as well, which would take a bit of time too.

kraenhansen: > Does it add anything over an event emitter? (I see/share your scepticism).

If users are just iterating the events as they come in from the server, I do think it's a simpler API to consume (once you know and understand the syntax).

> How does one unsubscribe / unwatch?

Thats a good question - I don't see any code explicitly closing the request to the server.
Reading both https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function* it's unclear to me how the caller of a generator would signal to the implementer of the generator that they're done using it.

If it's not possible to stop consuming a watch, that's for sure a significant downside to this API.

packages/realm-web/src/FunctionsFactory.ts
kraenhansen: It's mainly here for the watch support.
I'm unsure but believe that it's not possible to change the response content-type of a MongoDB Realm Function, it's always `application/json`, which would be a prerequisite for custom streaming functions defined by the users of MongoDB Realm.

packages/realm-web/src/User.ts
kraenhansen: Yep - good catch. I saw it and forgot all about it :D

packages/realm-web/src/FunctionsFactory.ts
steffenagger: Ok, I'm just worried users will expect `callFunctionStreaming` to be a streaming implementation of `callFunction`, usable in the same way. And this would be exposed on `user.functions`, right?

kraenhansen: Right ..

packages/realm-web/src/WatchStream.ts
kraenhansen: No clue, but I made the change and the tests are green 🟢

packages/realm-web/src/WatchStream.ts
kraenhansen: I havn't ... I agree that if there's a problem, we inform them to have it solved instead of just working around it.

packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts
nirinchev: Here's a blog post that proposes an approach for cancelling iteration precisely in relation to SSEs: http://seg.phault.net/blog/2018/03/async-iterators-cancellation/

bmunkholm: What's the alternative API and how long would it take to change it?
Are you all in agreement the API is too exotic or should we list up the pro/con for alternatives?

kraenhansen: These are some of our alternative, off the top of my head:

# Alternatives

## 0. Keep the `AsyncIterator` as is.

Whatever we do, I believe it might make sense to keep this API around for expert users. I do see the advantages of the pull-based iteration which this API enables and I think it could be valuable for an advanced user of the library.

## 1. A simple listener pattern, like the Node.js driver `ChangeStream` API

One alternative is the simplified listener pattern which the Node.js driver is currently exposing: https://mongodb.github.io/node-mongodb-native/3.6/api/ChangeStream.html

This is probably my favorite, since it's simpler to implement and it mimics the Node.js driver enabling transfer of knowledge between drivers and SDKs.

```javascript
const stream = collection.watch();
stream.onNext(change => {
  // Do whatever with the change event ...
});
// After some time, when we're done with the stream
stream.close()
```

## 2. A full-on listener pattern through an API similar to (but perhaps a bit more limited than) [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter) with separate event names per operation type

```javascript
const emitter = collection.watch();
// Register callbacks to react to changes
// Could also be the more verbose `emitter.addListener("insert", event => {`
emitter.on("insert", event => {
  const document = event.fullDocument;
  // Stop listening conditionally based on some value in a document inserted in the collection
  if (document.stop === true) {
    emitter.close();
  }
});
// Register a callback to react to updates
emitter.on("update", event => {
  const document = event.fullDocument;
  // Stop listening conditionally based on some value in a document inserted in the collection
  if (document.stop === true) {
    emitter.close();
  }
});
// Register a callback to react to errors
// This includes errors from establishing the connection as well as errors parsing the event stream, etc.
emitter.on("error", console.error);
// Potentially a way to register a callback to react to any type of event
emitter.on(event => {
  if (event.operationType === "insert") {
    const document = event.fullDocument;
    // Do whatever with the document ...
  }
});
// After some time - close the emitter when we don't care about events anymore
emitter.close();
```

# Alternatives for extending the API later.

If we choose to add an event emitting variant to this API, I see three options

## 0) Provide a `mode` to watch

Users could provide an additional optional option when calling `collection.watch`.

This is probably my favorite, since we would still support an event emitter and it wouldn't feel as "second class" as option 1 (below).

### Using an Async iterator

Probably the default to limit breaking changes and nudge users into using this pattern.

```javascript
for await (const event of collection.watch({ filter, mode: "iterator" })) {
  // Do whatever with event
}
```

### Using an Event Emitter

```javascript
const emitter = collection.watch({ filter, mode: "emitter" });
// Register a listener
emitter.on(event => {
  // Do whatever with event
});
// When you're done, call close
emitter.close();
```

## 1) Adding a method with a different name

The easiest in terms of development time would be to add a future alternative as a separate method.
Ideas for names could be: `collection.watchEmitting()`, `collection.watchWithEmitter()`, etc.


## 2) Break the API

Simply break the API, only returning an event emitter and abandoning the async iterator all together.

blagoev (approved):

steffenagger (approved):
I'm still sceptical, but ok, let our users try it out & decide..

Issue #3254, opened 2020-09-23
Copying same object multiple time while writing JSON that contains multiple nested objects


iamprahlad:
## Goals
Writing JSON that contains multiple nested objects

## Expected Results
```js
{
    id: '12324',
    age: 22,
    name: 'Pankaj',
    phones: [{ model: 'Redmi 5A', price: 60000 }, { model: 'Nokia Lumia X2', price: 5000 }, { model: 'Nokia A1', price: 8000 }],
    car: { model: 'i20', price: 500000, launchDate: new Date(1600028986548) },
    address: {
        addressLocality: "Khatima",
        addressRegion: "Uk",
        postalCode: "262308",
        streetAddress: "Hill View Colony "
    }
}
```

## Actual Results
```js
{
    address: { model: "Redmi 5A", price: 60000, $refId: "RealmObject{0000-0000}" },
    age: 22,
    car: { model: "Redmi 5A", price: 60000, $refId: "RealmObject{0000-0000}" },
    id: "12324",
    name: "Pankaj",
    phones: [
        { model: "Redmi 5A", price: 60000, $refId: "RealmObject{0000-0000}" },
        { model: "Nokia Lumia X2", price: 5000, $refId: "RealmObject{0000-0001}" },
        { model: "Nokia A1", price: 8000, $refId: "RealmObject{0000-0002}" }
    ]
}
```

## Code Sample
```js
const PersonSchema = {
    name: 'Person',
    primaryKey: 'id',
    properties: {
        id: 'string',
        age: 'int',
        name: 'string',
        phones: 'Phone[]',
        car: 'Car',
        address: 'Address'
    }
}

const PhoneSchema = {
    name: 'Phone',
    properties: {
        model: 'string',
        price: 'int'
    }
}

const CarSchema = {
    name: 'Car',
    properties: {
        model: 'string',
        price: 'int',
        launchDate: 'date',
    }
}

const addressSchema = {
    name: 'Address',
    properties: {
        streetAddress: 'string?',
        postalCode: 'string?',
        addressRegion: 'string?',
        addressLocality: 'string?'
    }
}

const dummyPersonData = {
    id: '12324',
    age: 22,
    name: 'Pankaj',
    phones: [{ model: 'Redmi 5A', price: 60000 }, { model: 'Nokia Lumia X2', price: 5000 }, { model: 'Nokia A1', price: 8000 }],
    car: { model: 'i20', price: 500000, launchDate: new Date(1600028986548) },
    address: {
[...]

PR #3258, opened 2020-09-23
bugfix: toJSON internal cache ids resulting in wrong data (v10)

steffenagger:
Relying on `this.constructor.name` was a bad idea. This fix switches to using `objectSchema.name` instead.

Related issue #3254

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
+* Fixed and issue with `toJSON` where data from a different objects could be serialized. ([#3254](https://github.com/realm/realm-js/issues/3254))
```
kneth, line 7: "Fixed and issue" --> "Fixed an issue"
"a different objects" --> "a different object"

Do we know which version the detect was introduced?

kneth (approved):

PR #3260, opened 2020-09-24
Xcode Archive

### LINKED ISSUE #3076: xCode Archive fails
### What are you trying to achieve? 

Publish my app to app store

### What did you expect to happen? 

Archive my app using xcode.

### What happened instead?
Build failed 

I have installed realm in my React Native project and it works well in Android. Although, in iOS, despite I can run the application in the simulator, when I proceed to archive it to publish to the app store, it throws this error:

```
Aligned deallocation function of type 'void (void *, std::align_val_t) noexcept' is only available on iOS 11 or newer
```
This happen in different files: alloc.hpp to mention one of them.

I tried changing the iOS target version but it didn't work.

Any suggestions?

Thanks in advance.

### What are steps we can follow to reproduce this issue? 

Add realm as a dependency and try yo archive the project using xcode.

- Realm JS SDK Version: 10.0.0-beta.7?
- Node: 12.2.0
- React Native: 0.62
- Client OS & Version: MacOS Catalina 10.15.5

steffenagger:
'-fno-aligned-new' added to podspec as C++ flag for for armv7

(is the wording in my added comment correct?)

This closes #3076

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Fixed
 * Fixed `create<T>(...)` deprecation warning. ([#3243](https://github.com/realm/realm-js/pull/3243))
 * Throw error when `deleteRealmIfMigrationNeeded` is requested on a synced realm (incompatible options) ([#3245](https://github.com/realm/realm-js/pull/3245))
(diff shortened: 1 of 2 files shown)


kneth (approved):
Please add an entry in `CHANGELOG.md`

blagoev (approved):

Issue #3110, opened 2020-08-05
TypeError: Reflect.construct requires the first argument to be a constructor


uendar:
I am getin this error when i run : react-native run-android
First the app opens correctly and after 3 sec it shows me this error.

I have never had this before. I am using last realm version 6.0.4
Even with older versions is the same issue.

I have lot of schemas that i import in an index file :

```
const Realm = require('realm');

class CountriesSchema extends Realm.Object {}

CountriesSchema.schema = {
    name: 'Countries',
    primaryKey: 'passportId',
    properties: {
        passportId: {type: 'string'},
        nationName: {type:'string', optional: true, default:''},
        code: {type:'string', optional: true, default:''},

    }
};

export default CountriesSchema;
```

```

import Countries from './CountriesDataBase/CountriesSchema';

realmInstance = new Realm({
            encryptionKey: getKey(),
            schema: [User]
})
```

As i said there are lot of schemas but i follow the same rule

blagoev:
Please fill out our issue template in full so we can see what the error is. Currently the issue does not contain any error.

blagoev:
My bad the error is in the title. 
Could you fill out the issue template so we know which version and platforms this happens.

thanks

rdamborsky:
I'm having same issue. I've seen it when I tried to upgrade to RN 62 few weeks ago (and so rolled back). Recently I upgraded to RN 63 and have it again.

## Goals

Use classes with instance methods for my models, while connecting to realm.

## Expected Results

Realm should open, data should be fetched correctly and instance methods work on instances of the models returned by realm.

## Actual Results

Getting error in terminal:

**TypeError: Reflect.construct requires the first argument be a constructor**

## Steps to Reproduce

Running an example below and using class instead of schema, in Realm.open.

## Code Sample

```js
class Base extends Realm.Object {
  debug() {
    log.debug('Area debug info', [`id=<${this.id}>`, `label=<${this.label}>`]);
  }
}

class AreaPlain extends Base {
  // @observable label;
}

const area = new AreaPlain();
area.id = 'all';
area.label = 'All areas';

Realm.open([Area]);
```

## Version of Realm and Tooling

- Realm JS SDK Version: 6.0.4 (tried with 6.0.2 and 6.0.3 as well)
- React Native: 0.63.2
- Client OS & Version: Android 29
- Which debugger for React Native: None

[...]

PR #3261, opened 2020-09-24
fix RN inheriting objects transpiled code  support

blagoev:
this fixes https://github.com/realm/realm-js/issues/3110

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -95,6 +95,12 @@ function getRealmConstructor(context) {
 
             return nodeRequire(binding_path).Realm;
         case 'reactnative':
+            //switch how babel transpiled code creates children objects.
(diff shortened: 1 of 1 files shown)


kneth (approved):
LGTM.

Please add an entry to `CHANGELOG.md`: Fixed inheritance when transpiling with Babel (or something like this).

steffenagger (commented):
I think we should update the Class Model schema in out tests to use the class syntax:
https://github.com/realm/realm-js/blob/master/tests/js/schemas.js#L30-L49
Or maybe I could ask: would that have caught this issue?

blagoev:
probably yes. We need definitely improve our class syntax support in future

PR #3262, opened 2020-09-24
fix RN inheriting objects transpiled code  support

### LINKED ISSUE #3110: TypeError: Reflect.construct requires the first argument to be a constructor

blagoev:
this fixes #3110 for v10

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -95,6 +95,12 @@ function getRealmConstructor(context) {
 
             return nodeRequire(binding_path).Realm;
         case 'reactnative':
+            //switch how babel transpiled code creates children objects.
+            //Inheriting from Realm.Object with class syntax does not support using Reflect.construct the way babel transpiles it.
+            //Defining Reflect.construct.sham makes the transpiled code use different standart mechanism for inheriting. (Function.apply with setPrototypeOf)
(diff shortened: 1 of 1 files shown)


kneth (approved):
LGTM

See my comment in #3261.

steffenagger (commented):
I think we should update the Class Model schema in out tests to use the class syntax:
https://github.com/realm/realm-js/blob/v10/tests/js/schemas.js#L44-L63
Or maybe I could ask: would that have caught this issue?

PR #3266, opened 2020-09-25
Fix for TS create declarations when updating (v10)

steffenagger:
### Main goal
Initiated based on this comment: https://github.com/realm/realm-js/pull/3044#issuecomment-698617303 - v6 PR can be found here: https://github.com/realm/realm-js/pull/3271

Allow for partial models when using `Realm.UpdateMode.All` or `Realm.UpdateMode.Modified`

**Create declarations has been updated to:**
```ts
/** ... */
create<T>(type: string, properties: RealmInsertionModel<T>, mode?: Realm.UpdateMode.Never): T & Realm.Object;
create<T>(type: string, properties: Partial<T> | Partial<RealmInsertionModel<T>>, mode: Realm.UpdateMode.All | Realm.UpdateMode.Modified): T & Realm.Object;

/** ... */
create<T extends Realm.Object>(type: {new(...arg: any[]): T; }, properties: RealmInsertionModel<T>, mode?: Realm.UpdateMode.Never): T;
create<T extends Realm.Object>(type: {new(...arg: any[]): T; }, properties: Partial<T> | Partial<RealmInsertionModel<T>>, mode: Realm.UpdateMode.All | Realm.UpdateMode.Modified): T;
```
Deprecated declarations has been removed after agreement with @kneth 
Any better way of achieving the above?

### Additional changes

I extended integration-tests for object creation/reading to include update-tests & test for Class Models (these additions only works in combination with https://github.com/realm/realm-js/pull/3262 - now rebased on `v10` efter merge).

And fixed an issue with serialization tests (so it now works in different environments - the previous error message check was a bit too specific, as described [here (MDN)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value)).

### Update

**objectForPrimaryKey<T> updated to:**
(to align with the existing declarations for `objects<T>(...)` - tests for this added)
```ts
/** ... */
objectForPrimaryKey<T>(type: string, key: number | string | Realm.ObjectId): (T & Realm.Object) | undefined;

/** ... */
objectForPrimaryKey<T extends Realm.Object>(type: {new(...arg: any[]): T; }, key: number | string | Realm.ObjectId): T | undefined;
```

The above change limits the use of `realm.objectForPrimaryKey(Person)` to only accept Class Model which extends `Realm.Object` (just like `objects<T>(...)` is currently declared).

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -571,11 +571,18 @@ declare class Realm {
     deleteAll(): void;
 
     /**
-     * @param  {string|Realm.ObjectType|Function} type
+     * @param  {string} type
      * @param  {number|string|ObjectId} key
      * @returns {T | undefined}
      */
-    objectForPrimaryKey<T>(type: string | Realm.ObjectType | Function, key: number | string | Realm.ObjectId): T & Realm.Object | undefined;
+    objectForPrimaryKey<T>(type: string, key: number | string | Realm.ObjectId): (T & Realm.Object) | undefined;
+
+    /**
+     * @param  {Class} type
+     * @param  {number|string|ObjectId} key
+     * @returns {T | undefined}
+     */
+    objectForPrimaryKey<T extends Realm.Object>(type: {new(...arg: any[]): T; }, key: number | string | Realm.ObjectId): T | undefined;
 
     /**
      * @param  {string} type
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -542,7 +542,8 @@ declare class Realm {
      * @param  {Realm.UpdateMode} mode? If not provided, `Realm.UpdateMode.Never` is used.
      * @returns T & Realm.Object
      */
-    create<T>(type: string, properties: RealmInsertionModel<T>, mode?: Realm.UpdateMode): T & Realm.Object
+    create<T>(type: string, properties: RealmInsertionModel<T>, mode?: Realm.UpdateMode.Never): T & Realm.Object;
+    create<T>(type: string, properties: Partial<RealmInsertionModel<T>>, mode: Realm.UpdateMode.All | Realm.UpdateMode.Modified): T & Realm.Object;
 
     /**
      * @param  {Class} type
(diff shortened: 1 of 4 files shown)


kneth (approved):

blagoev (approved):

blagoev (changes requested):
maybe we can improve a bit

types/index.d.ts
```diff
@@ -542,35 +542,17 @@ declare class Realm {
      * @param  {Realm.UpdateMode} mode? If not provided, `Realm.UpdateMode.Never` is used.
      * @returns T & Realm.Object
      */
-    create<T>(type: string, properties: RealmInsertionModel<T>, mode?: Realm.UpdateMode): T & Realm.Object
+    create<T>(type: string, properties: RealmInsertionModel<T>, mode?: Realm.UpdateMode.Never): T & Realm.Object;
```
blagoev, line 545: Why not remove the overloads and use this instead 
create<T>(type: string, properties: RealmInsertionModel<T>, mode: Realm.UpdateMode = Realm.UpdateMode.Never): T & Realm.Object;

In TS default-initialized parameters at the end are treated as optional

steffenagger: I had the same thought, and it would simplify the declaration - but default initializers are only allowed in implementations, not (pure) declarations 🤷‍♂️ - I'm not entirely sure why this is a limit... as you'll need to know the underlying implementation anyway, when writing the declarations..?

TS error says:
```
A parameter initializer is only allowed in a function or constructor implementation.
```

blagoev: 🤦 got it.  
But then this is a good reminder that  we don't need the initializer since this is just a type check. We could just write this 
`create<T>(type: string, properties: Partial<T> | Partial<RealmInsertionModel<T>> | RealmInsertionModel<T>, mode?: Realm.UpdateMode): T & Realm.Object;`

steffenagger: Yeah, but this would allow inserts of `Partial<T>` for `Realm.UpdateMode.Never`. This is where we want TS to help the user uphold their own model definitions (basically mandatory/optional props) - that's why I kept the two separate.

blagoev: I see.

blagoev (approved):

PR #3269, opened 2020-09-28
Realm Web: Making session refresh and deletion more robust

kraenhansen:
## What, How & Why?

**As is**: Failure during access token / session refresh (which could happen due to the session or user already being deleted from the server), would fail an authenticated request with the failure from refreshing the token.
**To be**: Merging this PR, if the request refreshing the access token fails with a 401 HTTP status code, we expect the refresh token to have been invalidated and we therefore don't expect a retry of the request to solve anything - the only thing we can do to make progress is logging out the user.

This also fixes an issue where session deletion upon logging out a user wouldn't make the user forget their token.

The semantics of logging out or removing a user might change in the future, but this is an attempt to make the current implementation a bit more robust and usable.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web/CHANGELOG.md
+++ b/packages/realm-web/CHANGELOG.md
@@ -2,10 +2,10 @@
 =============================================================
 
 ### Enhancements
-* None
+* Changing the behaviour when refreshing an access token fails. With this change, if the refresh token cannot be used to refresh an access token, the user is logged out. ([#3269](https://github.com/realm/realm-js/pull/3269))
 
 ### Fixed
-* None
+* Fixed forgetting the users access and refresh tokens, even if the request to delete the session fails. ([#3269](https://github.com/realm/realm-js/pull/3269))
 
 ### Internal
 * None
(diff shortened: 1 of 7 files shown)


packages/realm-web/src/tests/App.test.ts
```diff
@@ -489,6 +478,66 @@ describe("App", () => {
...
+            throw new Error("Expected the request to fail");
+        } catch (err) {
+            expect(err).instanceOf(MongoDBRealmError);
+            if (err instanceof MongoDBRealmError) {
+                expect(err.message).equals(
+                    // The last failure is from failing to delete the session at logout
+                    "Request failed (DELETE http://localhost:1337/api/client/v2.0/auth/session): invalid session (status 401)",
+                );
+            }
+        }
```
steffenagger, line 528: Maybe we should consider adding [chai-as-promised](https://www.npmjs.com/package/chai-as-promised)?

packages/realm-web/src/tests/User.test.ts
```diff
@@ -89,6 +90,37 @@ describe("User", () => {
...
+            id: "some-user-id",
+            accessToken: "deadbeef",
+            refreshToken: "very-refreshing",
+        });
+        // Log out the user
+        let errorThrown;
+        await user.logOut().then(null, err => {
+            errorThrown = err;
+        });
+        expect(errorThrown).instanceOf(MongoDBRealmError);
```
steffenagger, line 120: Instead of mixing `await` and `then` (`catch`) chaining (which is a bit weird to read, for me at least), I would suggest:

```suggestion
        try {
            await user.logOut();
            assert.fail("Logout should fail");
        } catch (err) {
            expect(err).instanceOf(MongoDBRealmError);
        }
```

steffenagger (approved):
LGTM
(see comments / suggestions)

packages/realm-web/src/User.ts
```diff
@@ -252,16 +252,19 @@ export class User<
...
+                await this.fetcher.fetchJSON({
+                    method: "DELETE",
+                    path: routes.api().auth().session().path,
+                    tokenType: "refresh",
+                });
+            }
+        } finally {
+            // Forget the access and refresh token
+            this.accessToken = null;
+            this.refreshToken = null;
```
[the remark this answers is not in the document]
kraenhansen: Sending the request to delete the session needs the refresh token on the user object, so I don't think that would be ideal. But perhaps we could null the access token right away 🤔

packages/realm-web/src/tests/App.test.ts
kraenhansen: Yeah - I considered adding it. I don't like they way it gets registered into chai, but I might be able to overcome that if I keep repeating this pattern.

packages/realm-web/CHANGELOG.md
```diff
@@ -2,10 +2,10 @@
 =============================================================
 
 ### Enhancements
-* None
+* Changing the behaviour when refreshing an access token fails. With this change, if the refresh token cannot be used to refresh an access token, the user is logged out. ([#3269](https://github.com/realm/realm-js/pull/3269))
 
 ### Fixed
-* None
+* Fixed forgetting the users access and refresh tokens, even if the request to delete the session fails. ([#3269](https://github.com/realm/realm-js/pull/3269))
```
kneth, line 8:
```suggestion
* Fixed forgetting the user's access and refresh tokens, even if the request to delete the session fails. ([#3269](https://github.com/realm/realm-js/pull/3269))
```

kneth (approved):
LGTM

packages/realm-web/src/User.ts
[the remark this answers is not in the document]
steffenagger: Good point...

PR #3273, opened 2020-09-29
fix syntax error when loaded in Electron renderer process

blagoev:
fix syntax error when loaded in Electron renderer process


steffenagger (approved):
LGTM

PR #3280, opened 2020-10-01
Core 6.1.3 & Sync 5.0.28 update

steffenagger:


kneth (approved):

bmunkholm:
Ideally we want all Fixes to mention when the bug was introduced (in the SDK) through "since x.y.z". I see a few fixes here that doesn't have that info.
It's also not entirely clear to me if the bugs discovered in .NET and Cocoa are relevant for JS? They likely are, but then we are missing tests for them I assume :-)

bmunkholm (approved):
There are still 3 other (previous bugfixes) that don't have "since". It's not critical, let's just ensure we get it in going forward.

steffenagger:
> It's also not entirely clear to me if the bugs discovered in .NET and Cocoa are relevant for JS? They likely are, but then we are missing tests for them I assume :-)

As they're fixed in Core/Sync I assume it's a shared problem, and that tests for these cases have been introduced in Core/Sync respectively?

> There are still 3 other (previous bugfixes) that don't have "since". It's not critical, let's just ensure we get it in going forward.

Updated the 2 remaining related to core/sync - the last one is hard (for me) to say, I think it might also depend on the version of RN consuming Realm?

PR #3281, opened 2020-10-01
Rename Realm.App.getCachedApp to Realm.App.getApp

kneth:
## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* [x] jsdoc files updated


kraenhansen:
Do we have any tests exercising this?
I don't think this method have [the same semantics as in Realm Web](https://github.com/realm/realm-js/blob/v10/packages/realm-web/src/index.ts#L29-L37).

What happens if a user calls `getApp` before instantiating an app elsewhere with that particular id?

```javascript
import { App } from "realm";
const app = App.getApp("my-awesome-app");
```

kneth:
> Do we have any tests exercising this?

Not at the moment.

> Ex: What happens if a user calls `getApp` before instantiating an app elsewhere with that particular id?

`null` will be returned. I agree that it will be valuable to create the app if it hasn't been created.

kneth:
@kraenhansen I have changed the behaviour to mimic Realm Web.

lib/app.js
```diff
@@ -19,6 +19,7 @@
 "use strict";
 
 const {promisify} = require("./utils.js");
+const { realpath } = require("fs");
```
kraenhansen, line 22: This looks like an auto-complete error? :)

Issue #3268, opened 2020-09-25
JWT metadata missing on User object


rvanmil:
## Goals

I have specified metadata fields in the custom JWT authentication settings, but these fields do not appear on the user object after logging in.
I am working with `realm-web` version `0.9.0`

## Expected Results

When using the Stitch SDK these fields could be found on the `user.profile.data` object (`mongodb-stitch-browser-sdk` version `4.8.0`).

## Actual Results

The `user.profile.data` object does not exist. See screenshots.

<img width="888" alt="realmjwt1" src="https://user-images.githubusercontent.com/991226/94289215-ad6c9200-ff58-11ea-8418-404e498f59e3.png">
<img width="1181" alt="realmjwt2" src="https://user-images.githubusercontent.com/991226/94289223-b1001900-ff58-11ea-9b52-0d07ea37c263.png">

## Steps to Reproduce

Use custom JWT authentication with metadata fields. Login in the browser and check the logged in user object.

## Code Sample

```
const loggedInUser = await realmApp.logIn(Realm.Credentials.jwt(jwtString))
console.log(loggedInUser.profile.data) // undefined
```

## Version of Realm and Tooling

- Realm JS SDK Version: realm-web 0.9.0
- Node or React Native: Web
- Client OS & Version: Chrome 85 on macOS 10.15.7
- Which debugger for React Native: None

rvanmil:
The HTTP request to https://eu-west-1.aws.stitch.mongodb.com/api/client/v2.0/auth/profile does return the missing data fields, so I think this is a problem in the code which parses that response.

nirinchev:
Shouldn't that go into [`user.customData`](https://github.com/realm/realm-js/blob/v10/types/app.d.ts#L383-L388)?

rvanmil:
I don’t think so, that is a different feature.

kraenhansen:
I can see on the screenshot that you're using a slash in the key path when setting up metadata fields. Are you sure those shouldn't be dots? [See dot notation](https://docs.mongodb.com/manual/core/document/#dot-notation).

That being said, there might very well be an issue here we need to investigate. Can you confirm the data is not present on the `customData` property? Are you using other SDKs where you're observing a different behavior?

rvanmil:
I've done some investigating by comparing the `realm-web` source to the `mongodb-stitch-browser-sdk` source and I have found the cause of this issue.

[...]

PR #3302, opened 2020-10-06
Realm Web: Add custom JWT metadata to UserProfile data property

rvanmil:
## What, How & Why?
Fixes realm/realm-js#3268
Any properties from the `data` property on the `/profile` HTTP response which are not specified in the `DATA_MAPPING` keys, will be added to the `UserProfile.data` property. This results in the same behaviour as the old Stitch browser SDK.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-web/src/UserProfile.ts
+++ b/packages/realm-web/src/UserProfile.ts
@@ -94,6 +94,9 @@ export class UserProfile implements Realm.UserProfile {
     public readonly maxAge?: string;
 
     /** @inheritdoc */
+    public readonly data?: any;
+
+    /** @inheritdoc */
     public readonly type: Realm.UserType = UserType.Normal;
 
     /** @inheritdoc */
@@ -134,6 +137,17 @@ export class UserProfile implements Realm.UserProfile {
                     ) {
                         this[propertyName] = value;
                     }
+                }
+                // Add remaining data to `data` property
+                const metadata = Object.entries(data).reduce((acc, entry) => {
+                    if (DATA_MAPPING[entry[0] as DataKey]) {
+                        return acc;
+                    }
+                    acc[entry[0]] = entry[1];
+                    return acc;
+                }, {} as Record<string, unknown>);
+                if (Object.keys(metadata).length > 0) {
+                    this.data = metadata;
                 }
             } else {
                 throw new Error("Expected 'data' in the response body");
--- a/packages/realm-web/src/tests/User.test.ts
+++ b/packages/realm-web/src/tests/User.test.ts
@@ -131,6 +131,7 @@ describe("User", () => {
                 {
                     data: {
                         first_name: "John",
+                        metadata: "JWT metadata",
                     },
                     identities: [],
                     type: "normal",
@@ -150,6 +151,9 @@ describe("User", () => {
             identities: [],
             type: UserType.Normal,
             firstName: "John",
+            data: {
+                metadata: "JWT metadata",
+            },
         });
     });
 


kneth:
@rvanmil 

rvanmil:
Done 👍

kraenhansen:
@rvanmil have you built and tested this with your own app?

kraenhansen:
Looking at the implementation in the old Stitch SDK it looks like all data (including the keys in `DATA_MAPPING`) are exposed via the public `data` property: https://github.com/mongodb/stitch-js-sdk/blob/master/packages/core/sdk/src/auth/internal/StitchUserProfileImpl.ts#L45

nirinchev:
Should this be implemented across all SDKs? cc @jsflax

nirinchev:
The discussion in https://github.com/realm/realm-js/issues/3268 points that this is different from custom data - I haven't had a chance to test it yet though.

jsflax:
Unless I'm mistaken, it's just prefetched custom data. We could expose it on the profile like we did previously. I don't really have a strong opinion though.

rvanmil:
> The discussion in #3268 points that this is different from custom data - I haven't had a chance to test it yet though.

It is. The data we access from the `user.profile.data` property is actually from the Auth0 JWT token we use to log users in. The metadata mapping feature on the Realm backend allows us to copy this data onto the Realm user context.

This is something we use extensively on both the frontend and in Realm functions and rules, for example checking a user's permissions (which are defined in Auth0).

nirinchev:
@jsflax I did test it and confirmed @rvanmil's findings - metadata fields from JWTs get sent down along with the rest of the profile info, so we need to expose a dictionary-like lookup on UserProfile to allow users to retrieve their custom fields. I filed https://github.com/realm/realm-object-store/issues/1123 to track it for non-web SDKs.

rvanmil:
@kraenhansen is there anything I can do to move this PR forward and get it into a release? We'd love to move away from the Stitch web SDK towards the Realm web SDK but this is a blocking issue for us.

kraenhansen:
@rvanmil I'm really sorry this has stalled. Especially since I encouraged you to submit this PR.
When adding a feature to the SDK we need to align across multiple SDKs (to avoid the breaking changes which we would need to align retroactively) and this has proven a bit more difficult (than it probably needs to be) in this particular case.
I'll keep you updated and hopefully we can get this merged and released soon 🤞

rvanmil:
No problem @kraenhansen thanks for the update 😊

rvanmil:
@kraenhansen any chance you could share an ETA for this? We've got a couple of projects still running on the almost a year old `mongodb-stitch-browser-sdk` and it would be great if we could migrate those towards the Realm Web SDK this month or january. This issue is the only thing that's blocking the upgrades. Let me know if there's anything I can do to help 😊

kraenhansen:
@rvanmil okay, I've got some good news and some bad news.

I just had a meeting with @jsflax and @nirinchev yesterday, to discuss this issue and this is where we ended up:
The extra properties provided by authentication providers will be exposed on the `profile` property of a `User` instance.
Your original issue and this PR spawned a greater discussion and for that I am very grateful! I did however find that the change we needed was a bit more involved for me to reuse your work here, so I went ahead and created a different PR adding this (and a few other things): https://github.com/realm/realm-js/pull/3481 ... and with this, I will close this PR.

Thanks again for your attention, patience and time!
I hope you haven't been completely discouraged from proposing changes to our SDKs in the future 🙏

rvanmil:
@kraenhansen thanks and no problem at all! I'm glad this led to a greater discussion and an improvement to the SDK. Looking forward to the release including your PR 😊

PR #3305, opened 2020-10-07
Update to Realm Core v6.1.4 and Realm Sync v5.0.29

kneth:


fronck (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* If you make a case insensitive query on an indexed string property, it may fail in a way that results in a `No such key` exception if an object that matches the query is deleted. ([realm/realm-cocoa#6830](https://github.com/realm/realm-cocoa/issues/6830), since v6.0.0)
```
steffenagger, line 7: It's quite wordy, maybe something like:
```suggestion
* A case-insensitive query, on an indexed string property, could throw a `"No such key"` when the query matched a deleted object. ([realm/realm-cocoa#6830](https://github.com/realm/realm-cocoa/issues/6830), since v6.0.0)
```

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+* If a property both has a change in nullability and is set as the new primary key property, migration can throw a `No Such Column` exception. ([#3270](https://github.com/realm/realm-js/issues/3270), since v6.0.0)
```
steffenagger, line 8: Past tense?
 
```suggestion
* A schema migration could throw a `"No Such Column"` if a property changed from optional to primary key. ([#3270](https://github.com/realm/realm-js/issues/3270), since v6.0.0)
```

steffenagger (commented):
I think we should keep descriptions of "Fixed" in past tense - and as concise as possible.

fronck (approved):

PR #3307, opened 2020-10-07
open-behaviour-tests (v10)

steffenagger:
## What, How & Why?

This PR currently:
- Re-introduces the `open-behavior-tests` (re-written to `async/await` for non-cancelable promises).
- Moves TS declarations of `openLocalRealmBehavior` & `downloadBeforeOpenBehavior` to `Realm.App.Sync` (from `Realm`).

Please see any `// NOTE:` in the changes.

https://github.com/realm/realm-js/issues/3339 has been created to track documentation.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests: Add a unit and preferably also an integration test to avoid regressions.
* [x] Run prettier on the source code changed.

### DIFF
--- a/tests/js/open-behavior-tests.js
+++ b/tests/js/open-behavior-tests.js
@@ -140,7 +140,7 @@ module.exports = {
 
         const realm = await Realm.open(config);
 
-        // TODO: Not quite sure what we're testing here?
+        // NOTE: Not quite sure what we're testing here?
         TestCase.assertTrue(realm.empty);
 
         realm.close();
@@ -522,7 +522,7 @@ module.exports = {
 
         const openPromise = new Promise((resolve, reject) => {
             const promise = Realm.open(config);
-            // TODO: could this potentially trigger before canceling?
+            // NOTE: could this potentially trigger before canceling?
             promise.progress(() => {
                 reject("Progress listener called");
             });
--- a/tests/js/open-behavior-tests.js
+++ b/tests/js/open-behavior-tests.js
@@ -26,7 +26,6 @@ const TestCase = require("./asserts");
 const schemas = require("./schemas");
 const Utils = require("./test-utils");
 const AppConfig = require("./support/testConfig");
-const user = require("../../lib/user");
 
 const APP_CONFIG = AppConfig.integrationAppConfig;
 
@@ -36,14 +35,7 @@ class TestError extends Error {
     }
 }
 
-const createSyncConfig = (sync, clearLocalFile = true) => {
-    const config = { schema: [schemas.DogForSync], sync };
-    if (clearLocalFile) {
-        // By default clean any previous test file.
-        Realm.deleteFile(config);
-    }
-    return config;
-}
+// NOTE: these tests can be simplified once we can create multiple anonymous users.
 
 module.exports = {
     testNewFile_openLocal: async function() {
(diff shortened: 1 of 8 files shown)


tests/js/test-utils.js
```diff
@@ -18,19 +18,43 @@
...
+}
+
+function randomVerifiableEmail() {
+    // according to the custom register function, emails will register if they contain "realm_tests_do_autoverify"
+    return `realm_tests_do_autoverify_${uuid()}_@test.com`;
+}
+
+async function getRegisteredEmailPassCredentials(app) {
+    if (!app) {
+      throw new Error("No app supplied to 'getRegisteredEmailPassCredentials'");
```
blagoev, line 44: maybe "Undefined argument app"

blagoev: considering this is for testing it does not matter much

tests/js/open-behavior-tests.js
```diff
@@ -16,463 +16,567 @@
...
+                user,
+                partitionValue,
+                _sessionStopPolicy: "immediately",
+                newRealmFileBehavior: { type: "downloadBeforeOpen" }
+            }
+        };
+
+        const realm = await Realm.open(config);
+
+        // NOTE: Not quite sure what we're testing here?
```
blagoev, line 143: new file should have no objects. Maybe just remove this comment cause the test was written like that before. Its in a context with the next tests which is `existingFile`

blagoev (approved):

tests/js/test-utils.js
steffenagger: I was trying to make the error message as explicit as possible - it's only in the case _we_ accidentally call `Utils. getRegisteredEmailPassCredentials()` (without args).
Maybe we should introduce a `TestSetupError`, and throw an error of that type in these cases? (Would be quite easy to spot when writing tests).

tests/js/open-behavior-tests.js
steffenagger: Can't be in the context of the next test, that's done with a different user (so wont be the same realm-file, and the next test ensures an existing file before testing - this is what confused me about this test).
I only left my comment in to get a clue what's going on (will remove before merge) :)

tests/js/test-utils.js
```diff
@@ -18,19 +18,43 @@
...
+function genPartition() {
+    return "xxxxxxxxxxxx4xxxyxxxxxxxxxxxxxxx".replace(/[xy]/g, function(c) {
+        let r = Math.random() * 16 | 0, v = c == "x" ? r : (r & 0x3 | 0x8);
+        return v.toString(16);
+    });
+}
+
+function randomVerifiableEmail() {
+    // according to the custom register function, emails will register if they contain "realm_tests_do_autoverify"
+    return `realm_tests_do_autoverify_${uuid()}_@test.com`;
```
kraenhansen, line 39: Just in case someone runs this against a server which is capable of sending emails, I think we should use a sub-domain of mongodb.com, in Realm Web I use:

```typescript
const now = new Date();
const nonce = now.getTime();
const email = `gilfoil-${nonce}@testing.mongodb.com`;
```

kraenhansen (approved):
Looks legitimate although I havn't gone though the tests in detail. Some tests on this is better than none.

tests/js/test-utils.js
steffenagger: good point.

PR #3315, opened 2020-10-08
Improving the building realm-js for react-native Android from scratch experience.

cesarvr:
## What, How & Why?

It was difficult to compile Android out-of-the-box so here I make three modification that hopefully will make things easier when trying to compile from scratch for react-native Android. 

First change is the error messaging when running ``./gradlew publishAndroid``, in case you don't have the NDK installed you get this error: 

```No such property: sdkHandler for class: com.android.build.gradle.LibraryPlugin```

I added some failure checking and change the message to: 

```
Unable to locate your NDK configuration. ¿Have you forgot to install Android NDK? 
More info: https://developer.android.com/studio/projects/install-ndk
```

Second change: I have added an additional way to get information about the NDK by reading the ```local.properties``` and extracting from there (if any) the information about the NDK path. This make it easier to use in Android Studio.

And third and last was to bump to ```Android-21``` from ```Android-9``` otherwise the user will get some cryptic errors (it won't be able to find the toolchain for ```Android-9``` SDK) if you are trying to use the latest NDK (```21.3.6528147```) to compile.

### DIFF
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -189,7 +189,7 @@ def findNdkBuildFullPath() {
         ndkDir = properties.getProperty('ndk.dir')
         if(ndkDir != null)
             return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
-    }catch(FileNotFoundException e){
+    } catch(FileNotFoundException e) {
         println("[Warning]: local.properties not found ")
         println("Error: ", e)
     }
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -269,7 +269,7 @@ def checkForHeadersFile(folder){
     File[] files = new File("$projectDir/../../$folder").listFiles();
 
     if (files == null || files.length == 0)
-        throw new GradleException("Didn't found any headers in the \\$folder folder \n" +
+        throw new GradleException("Didn't found any headers in the /$folder folder \n" +
         "Please you go to the root folder (realm-js) and do: \n git submodule update --init --recursive")
 }
 
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -202,7 +202,7 @@ def findNdkBuildFullPath() {
         if (ndkDir) {
             return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
         }
-    }catch(MissingPropertyException e){
+    } catch(MissingPropertyException e) {
         throw new GradleException("Unable to locate your NDK configuration. Did you forget to install the Android NDK?\n" +
                 "more info: https://developer.android.com/studio/projects/install-ndk \n")
     }
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -187,10 +187,7 @@ def findNdkBuildFullPath() {
 
     if(ndkDir != null)
         return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
-
-    // If not there we try the next approach...
-    
-
+        
     try {
         ndkDir = android.hasProperty('plugin') ? android.plugin.ndkFolder :
                 plugins.getPlugin('com.android.library').sdkHandler.getNdkFolder()
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -270,7 +270,7 @@ def checkForHeadersFile(folder){
 
     if (files == null || files.length == 0)
         throw new GradleException("Didn't found any headers in the /$folder folder \n" +
-        "Please you go to the root folder (realm-js) and do: \n git submodule update --init --recursive")
+        "Please make sure you go to the root folder (realm-js) and do: \n git submodule update --init --recursive")
 }
 
 task buildReactNdkLib(dependsOn: [prepareRealmCore, prepareOpenSSL_x86, prepareOpenSSL_x86_64, prepareOpenSSL_arm, prepareOpenSSL_arm_64], type: Exec) {
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -318,10 +318,6 @@ android {
     lintOptions {
         abortOnError false
     }
-    compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-    }
-    buildToolsVersion '28.0.3'
 }
 
 task publishAndroid(dependsOn: [generateVersionClass, packageReactNdkLibs], type: Sync) {
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -182,7 +182,6 @@ def findNdkBuildFullPath() {
     def ndkDir = null
 
     try {
-        println("wtf?")
         // Reading [ndk.dir] from local.properties
         Properties properties = new Properties()
         properties.load(project.rootProject.file('local.properties').newDataInputStream())
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -187,8 +187,9 @@ def findNdkBuildFullPath() {
         properties.load(project.rootProject.file('local.properties').newDataInputStream())
 
         ndkDir = properties.getProperty('ndk.dir')
-        if(ndkDir != null)
+        if (ndkDir != null) {
             return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+        }
     } catch(FileNotFoundException e) {
         println("[Warning]: local.properties not found ")
         println("Error: ", e)
--- a/react-native/android/build.gradle
+++ b/react-native/android/build.gradle
@@ -182,18 +182,21 @@ def findNdkBuildFullPath() {
     def ndkDir = null
 
     try {
+        println("wtf?")
         // Reading [ndk.dir] from local.properties
         Properties properties = new Properties()
         properties.load(project.rootProject.file('local.properties').newDataInputStream())
 
-        properties.getProperty('ndk.dir')
+        ndkDir = properties.getProperty('ndk.dir')
         if(ndkDir != null)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: react-native/android/src/main/jni/Application.mk [hunk 1]
```diff
@@ -1,7 +1,7 @@
 APP_BUILD_SCRIPT := Android.mk
 
 APP_ABI := armeabi-v7a x86 x86_64 arm64-v8a
-APP_PLATFORM := android-9
+APP_PLATFORM := android-21
```
kneth, line 4: I am a bit concerned with this bump. RN is currently at 16 (https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/jni/Application.mk#L9). On the other hand, 21 is Lollipop and we will support roughly 95 % of all active devices.

react-native/android/build.gradle
```diff
@@ -177,11 +178,31 @@ def findNdkBuildFullPath() {
...
+
+    try {
+        ndkDir = android.hasProperty('plugin') ? android.plugin.ndkFolder :
+                plugins.getPlugin('com.android.library').sdkHandler.getNdkFolder()
+
+        if (ndkDir) {
+            return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+        }
+    }catch(MissingPropertyException e){
+        throw new GradleException("Unable to locate your NDK configuration. ¿Have you forgot to install Android NDK? \n" +
```
kneth, line 202: `¿` - Welcome to Spain :-)

kneth (commented):
I believe it is still important to stay in sync with Realm Core and Realm Sync when it comes to the NDK version.

react-native/android/build.gradle
steffenagger: ¡I love it! 🙂

react-native/android/src/main/jni/Application.mk
cesarvr: Screenshot 2020-10-09 at 10 08 31

Cool, I've tested with ```Android-16``` and seems to works will move it to that. 👍

react-native/android/build.gradle
cesarvr: ;-)

react-native/android/src/main/jni/Application.mk
blagoev: yes RN 0.59 should be able to run on android 9, so we should not be bumping this version.

cesarvr: Cool, then we need to change the documentation on the Readme.md to the correct NDK version or provide some workaround.

react-native/android/src/main/jni/Application.mk
```diff
@@ -1,7 +1,7 @@
 APP_BUILD_SCRIPT := Android.mk
 
 APP_ABI := armeabi-v7a x86 x86_64 arm64-v8a
-APP_PLATFORM := android-9
+APP_PLATFORM := android-16
```
blagoev, line 4: there should be no need to bump this version

blagoev (changes requested):

blagoev:
@cesarvr I don't know why it fails for you but there should be no need to bump the version from Android-9

blagoev:
ok the problem is you are using newer version of the ndk. You should be using NDK 21.0

cesarvr:
> ok the problem is you are using newer version of the ndk. You should be using NDK 21.0



> ok the problem is you are using newer version of the ndk. You should be using NDK 21.0

Thanks for the tip will try with that one and ```Android-9``` and see how it goes

react-native/android/build.gradle
kraenhansen, line 202:
```suggestion
        throw new GradleException("Unable to locate your NDK configuration. Did you forget to install the Android NDK?\n" +
```

kraenhansen (changes requested):

react-native/android/src/main/jni/Application.mk
cesarvr: You where right it works with the NDK ```21.0``` I move it back to ```android-9```.

react-native/android/build.gradle
```diff
@@ -177,11 +178,31 @@ def findNdkBuildFullPath() {
...
+    Properties properties = new Properties()
+    properties.load(project.rootProject.file('local.properties').newDataInputStream())
+
+    def ndkDir = properties.getProperty('ndk.dir')
+
+    if(ndkDir != null)
         return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+
+    // If not there we try the next approach...
+
```
blagoev, line 192: too much empty lines. I would pretty format the code here

blagoev (approved):

react-native/android/src/main/jni/Application.mk
blagoev: Yep, maybe we need to explicitly state this in the readme to be 21.0

cesarvr (commented):
In case the user have forgot to do: ``git submodule update --init --recursive`` and run the ./gradlew publishAndroid we get this generic error:  

> Process 'command ‘$ANDROID_HOME/ndk/21.0.6113669/ndk-build'' finished with non-zero exit value 2

I added some functionality detect this error case and show something more informative: 

> Didn't found any headers in the /src/object-store folder
  Please you go to the root folder (realm-js) and do:

This message reduce make the user/us more productive.

react-native/android/build.gradle
```diff
@@ -242,7 +267,19 @@ def getNdkBuildFullPath() {
     return ndkBuildFullPath
 }
 
+def checkForHeadersFile(folder){
+    File[] files = new File("$projectDir/../../$folder").listFiles();
+
+    if (files == null || files.length == 0)
+        throw new GradleException("Didn't found any headers in the /$folder folder \n" +
+        "Please make sure you go to the root folder (realm-js) and do: \n git submodule update --init --recursive")
```
steffenagger, line 275: Just a suggestion to streamline wording..
```suggestion
        throw new GradleException("Unable to locate any headers in the '/$folder' folder.\n" +
        "Please make sure you go to the root folder (realm-js) and do:\n" +
        "git submodule update --init --recursive")
```

react-native/android/build.gradle
```diff
@@ -177,11 +178,35 @@ def findNdkBuildFullPath() {
         def ndkDir = System.getenv('ANDROID_NDK_HOME')
         return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
     }
-    def ndkDir = android.hasProperty('plugin') ? android.plugin.ndkFolder :
-            plugins.getPlugin('com.android.library').sdkHandler.getNdkFolder()
-    if (ndkDir) {
-        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+
+    def ndkDir = null
```
steffenagger, line 182: Could this be defined within the following try/catches? (as in the other cases)

steffenagger (approved):
LGTM
(check comments)

blagoev (approved):

react-native/android/build.gradle
```diff
@@ -177,11 +178,35 @@ def findNdkBuildFullPath() {
...
+
+    try {
+
+        ndkDir = android.hasProperty('plugin') ? android.plugin.ndkFolder :
+                plugins.getPlugin('com.android.library').sdkHandler.getNdkFolder()
+
+        if (ndkDir) {
+            return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+        }
+    }catch(MissingPropertyException e){
```
kraenhansen, line 205: What line would you expect throw this error?

```suggestion
    } catch(MissingPropertyException e) {
```

react-native/android/build.gradle
```diff
@@ -177,11 +178,35 @@ def findNdkBuildFullPath() {
...
+
+    try {
+        // Reading [ndk.dir] from local.properties
+        Properties properties = new Properties()
+        properties.load(project.rootProject.file('local.properties').newDataInputStream())
+
+        ndkDir = properties.getProperty('ndk.dir')
+        if(ndkDir != null)
+            return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+    }catch(FileNotFoundException e){
```
kraenhansen, line 192:
```suggestion
    } catch(FileNotFoundException e) {
```
Just trying to keep the code-style consistent across the commit.

react-native/android/build.gradle
```diff
@@ -177,11 +178,35 @@ def findNdkBuildFullPath() {
...
+    def ndkDir = null
+
+    try {
+        // Reading [ndk.dir] from local.properties
+        Properties properties = new Properties()
+        properties.load(project.rootProject.file('local.properties').newDataInputStream())
+
+        ndkDir = properties.getProperty('ndk.dir')
+        if(ndkDir != null)
+            return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
```
kraenhansen, line 191:
```suggestion
        if (ndkDir != null) {
            return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
        }
```

Just trying to keep the code-style consistent across the commit.

kraenhansen (changes requested):
Minor suggestions, feel free to judge if they apply - if not, just merge.

react-native/android/build.gradle
cesarvr: [This line ](https://github.com/realm/realm-js/blob/master/react-native/android/build.gradle#L181
) will throw a error like this: 

```
No such property: sdkHandler for class: com.android.build.gradle.LibraryPlugin
```

kneth (approved):

fronck (approved):
LGTM.
Disclamer:  I'm not very familiar with Gradle.

PR #3322, opened 2020-10-09
[v6] Cache Realm instances

kneth:
## What, How & Why?

The introduction of Realm Core v6, Realm Object Store changed disabled the caching of Realm instances.

**Currently**: https://github.com/realm/realm-object-store/blob/301642fe90212c379f550656a7234f41db158ddf/src/shared_realm.hpp#L235
**Pre Core v6**: https://github.com/realm/realm-object-store/blob/5d8bfcac9804962945d47530f372c6f8dee77416/src/shared_realm.hpp#L234

By disabling caching, user might see growing Realm files when they forgot to call `Realm.close()`.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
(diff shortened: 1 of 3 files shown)


kraenhansen (approved):
Interesting change 👀 
I'm not sure I 💯 % get it and why it would affect the return value of `compact`, but I believe you've tested this.

kneth:
> why it would affect the return value of compact

You cannot compact an open Realm. When the two instances are the same, closing one, closes the second one.

blagoev (approved):

PR #3336, opened 2020-10-13
realm-web workflow fix (v10)

steffenagger:
attempt to limit publish-realm-web to only trigger on created tags of format: 'realm-web-v*'


steffenagger:
Fixed elsewhere.

Issue #3236, opened 2020-09-18
Realm.login() will not run again after refreshing React Native app


matt2legit:
## Goals

Be able to successfully call `Realm.login()` again after refreshing app.

## Expected Results

Login should run and return the user information like the first time the app loads

## Actual Results

Nothing happens, no errors shown in the console.  Realm logs don't show the request is being made.

When I fire up the React Native Debugger, I did see an error about `Realm.Credentials.function(payload);` not being able to be created.  I believe some of the functionality for function auth  is missing in the `lib/browser/credentials.js` and `lib/browser/rpc.js`.  We are using function auth, however, I received the same behavior of the original issue using anonymous and other auth types.

## Steps to Reproduce

Launch app from Xcode in simulator or on iOS device.  
Run Login (works beautifully).
Refresh the react native bundle from the package server.
Try to login again (nothing happens)
## Version of Realm and Tooling

- Realm JS SDK Version: 10.0.0-beta.12
- Node or React Native: React Native 0.62.2
- Client OS & Version: iOS
- Which debugger for React Native: Chrome

neopit:
I have the same issue.
I have to kill the app and relogin

RealmBot:
➤ Chris Bush commented:

Encountered this myself. Some investigation revealed that (on iOS at least) the RunLoopScheduler is stale because its m_runloop is still pointing to the original thread, whereas the JS executor is actually running on a different thread after an app reload. This causes the scheduler to never be able to schedule requests, so the requests to log in are never heard from again. Please DM me for more details.

blagoev:
this is fixed in https://github.com/realm/realm-js/pull/3340

PR #3340, opened 2020-10-13
recreate EventLoopDispatcher in NetworkTransport on hot reload in RN

blagoev:
This fixes NetworkTransport EvenetLoopDispatcher to be recreated each time a new NetworkTransport is created. This is needed on RN for hot reload which creates a new main thread each time.
this fixes https://github.com/realm/realm-js/issues/3236

### DIFF
--- a/src/js_network_transport.hpp
+++ b/src/js_network_transport.hpp
@@ -168,7 +168,10 @@ struct JavaScriptNetworkTransport : public app::GenericNetworkTransport {
 
     using SendRequestHandler = void(ContextType m_ctx, const app::Request request, std::function<void(const app::Response)> completion_callback);
 
-    JavaScriptNetworkTransport(ContextType ctx) : m_ctx(ctx) {};
+    JavaScriptNetworkTransport(ContextType ctx) : m_ctx(ctx),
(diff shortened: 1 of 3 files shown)


src/js_network_transport.hpp
```diff
@@ -168,7 +168,10 @@ struct JavaScriptNetworkTransport : public app::GenericNetworkTransport {
 
     using SendRequestHandler = void(ContextType m_ctx, const app::Request request, std::function<void(const app::Response)> completion_callback);
 
-    JavaScriptNetworkTransport(ContextType ctx) : m_ctx(ctx) {};
+    JavaScriptNetworkTransport(ContextType ctx) : m_ctx(ctx),
+        m_dispatcher { JavaScriptNetworkTransport::send_request_to_server_impl }
```
tgoyne, line 172: The syntax here is misleading; the spaces around { } make it look like a block rather than a member intialization.

tgoyne (approved):
Need a changelog entry.

src/js_network_transport.hpp
blagoev: 👍  will fix

PR #3341, opened 2020-10-13
Realm Web: Running lint, build and tests for PRs

kraenhansen:
## What, How & Why?

This adds running lint, build and tests for PRs to code that affects Realm Web.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/packages/realm-web/src/tests/WatchStream.test.ts
+++ b/packages/realm-web/src/tests/WatchStream.test.ts
@@ -134,13 +134,8 @@ describe("WatchStream", () => {
                     });
                     assert.deepEqual(ws.state, WatchStreamState.HAVE_ERROR);
                     assert.instanceOf(ws.error, WatchError);
-                    if (true) {
-                        // XXX behavior in realm-web (no list of errors)
-                        assert.deepEqual(ws.error?.code, "WhoKnows");
-                    } else {
-                        // behavior in realm-js (objstore has a list of known errors and this isn't one of them).
-                        assert.deepEqual(ws.error?.code, "unknown");
-                    }
+                    // FIXME: Behavior in realm-web (no list of errors)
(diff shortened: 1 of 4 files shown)


packages/realm-web/src/tests/WatchStream.test.ts
```diff
@@ -692,14 +687,8 @@ describe("WatchStream", () => {
             // This is a deviation from the spec. We do not support the legacy macOS < 10 CR-only newlines.
             // The server does not generate them, and there would be some overhead to supporting them.
             ws.feedBuffer(toBuffer('event: message\rdata: {"a": 1}\r\r'));
-            if (true) {
-                // This is what we do.
-                assert.deepEqual(ws.state, WatchStreamState.NEED_DATA);
-            } else {
-                // This is what we would do if following the spec.
```
steffenagger, line 699: So we're not following the spec?

steffenagger (commented):
Changes LGTM, but check comment..
Also, som linting issues according to the Test Action.

packages/realm-web/src/tests/WatchStream.test.ts
kraenhansen: I think the "spec" referred to here is the "server-sent events" / "event source" specification. The comment above states that the server won't generate these and since we're only implementing towards this particular server, I think it's safe enough to leave it at this.

@RedBeard0531 would know the details of this comment, but for now I just wanted to remove the `if (true)` since that triggered a lint error.

kraenhansen:
Since the lint warnings are not introduced by this PR (but just surface them), I believe it's fine to merge this.

steffenagger (approved):

PR #3345, opened 2020-10-14
Realm Web: Integration test on PRs

kraenhansen:
## What, How & Why?

As a follow-up to https://github.com/realm/realm-js/pull/3341 - this is an attempt at running the Realm Web integration tests on PRs.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/packages/realm-web-integration-tests/harness/github.ts
+++ b/packages/realm-web-integration-tests/harness/github.ts
@@ -64,7 +64,7 @@ async function startServer() {
             // Resolve the promise when the server gets ready
             output += chunk.toString();
             // If it contains the magic string
-            if (output.match("Starting Stitch... Done")) {
+            if (output.includes("Starting Stitch... Done")) {
                 // Stop listening for output
                 serverProcess.stdout.off("data", handleOutput);
                 resolve();
--- a/packages/realm-web-integration-tests/harness/github.ts
+++ b/packages/realm-web-integration-tests/harness/github.ts
@@ -1,0 +1,102 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2020 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


packages/realm-web-integration-tests/harness/github.ts
```diff
@@ -0,0 +1,102 @@
...
+        /**
+         * Handle standard output by looking for a started "Stitch" service.
+         */
+        function handleOutput(chunk: Buffer) {
+            // Write to the regular STDOUT
+            process.stdout.write(chunk);
+            // Resolve the promise when the server gets ready
+            output += chunk.toString();
+            // If it contains the magic string
+            if (output.match("Starting Stitch... Done")) {
```
steffenagger, line 67: Instead of using regex-match, maybe use `output.includes("Starting Stitch... Done")`?

packages/realm-web-integration-tests/harness/github.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+async function startRunAndStop() {
+    await startServer();
+    await run();
+}
+
+startRunAndStop().then(
+    () => {
+        process.exit();
```
steffenagger, line 96: `0` is default, but `process.exit(0)` is used elsewhere.

packages/realm-web-integration-tests/harness/index.ts
```diff
@@ -40,16 +41,35 @@ const EXPECTED_ISSUES = [
     /was set with `SameSite=None` but without `Secure`/,
     // User / "refresh invalid access tokens" is posting an invalid token
     "Failed to load resource: the server responded with a status of 401 (Unauthorized)",
+    // Closing the watch streams, might yield this error (we're doing that three times)
+    "Failed to load resource: net::ERR_FAILED",
+    "Failed to load resource: net::ERR_FAILED",
+    "Failed to load resource: net::ERR_FAILED",
```
steffenagger, line 47: Just curious, would the user have these thrown?

steffenagger (approved):
LGTM (minor comments)

packages/realm-web-integration-tests/harness/index.ts
kraenhansen: Nope, they are not thrown - just logged to the console when the watch connection is closed and to be honest, I still don't really understand why (status code is 200 OK). I think it's perhaps the way we signal that we want to close (we're calling [`cancel`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/cancel) on the response body stream), which should be ok:
> The cancel() method of the ReadableStreamDefaultReader interface cancels the stream, signaling a loss of interest in the stream by a consumer. The supplied reason argument will be given to the underlying source, which may or may not use it.

PR #3350, opened 2020-10-16
Added Sync logging support for react-native iOS and Android.

cesarvr:
## What, How & Why?

Adding sync logging support for react-native iOS/Android. (you can see sync-logs in JS land)

I've consolidate the sync logging-behaviour into one class (less code to maintain). 

Bonus for Android: This PR offers a translation between the realm logs levels to Android log levels. (Integration with Android Studio Logcat).

## ☑️ ToDos

* [x] 📝 Changelog entry
~~* [ ] 📝 `Compatibility` label is updated or copied from previous entry~~
* [x ] 🚦 Tests
~~* [ ] 📝 Public documentation PR created or is not necessary~~

* [x] jsdoc files updated
~~* [ ] Chrome debug API is updated if API is available on React Native~~


src/common/js_logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+#if REALM_ANDROID
+    #include <android/log.h>
+#endif
+
+namespace realm {
+namespace common {
+
+using logger_fn = std::function<void(std::string, std::string)>;
+using log_level = realm::util::Logger::Level;
+using log_entry = std::pair<std::string, std::string>;
```
kraenhansen, line 32: I'm not particularly well versed in C++ code styles, but I would expect these types to be PascalCased.

src/common/js_logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+        { log_level::warn, "warn"},
+        { log_level::error, "error"},
+        { log_level::fatal, "fatal"},
+        { log_level::off,   "off"},
+    };
+
+
+    #if REALM_ANDROID
+    std::map <log_level, android_LogPriority> map_android_log_level {
+        { log_level::all, ANDROID_LOG_VERBOSE},
```
kraenhansen, line 96: I would never expect a message to be logged with the "all" or "off" log level, it seems to me like these are only relevant when setting the log-level. Would it make sense to leave them out of this map?

src/common/js_logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+    /*
+       Log levels available.
+       More info on (realm-core) realm/util/logger.hpp
+    
+       [ all, trace, debug, detail, info, warn, error, fatal, off ]
+    */
+
+    std::map <log_level, std::string> map_level {
+        { log_level::all, "all"},
+        { log_level::info, "info"},
```
kraenhansen, line 83: I would prefer if these were in the "right" order and "info" would be moved between "detail" and "warn".

kraenhansen (changes requested):
Why did you ~cross over~ updating the docs and types in the description of this PR? I suggest that you just leave them unchecked if you havn't done it or check the once that has "or is not necessary" written on them.

I would suspect this change also needs to update [the TypeScript definition](https://github.com/realm/realm-js/blob/v10/types/index.d.ts#L430) and [the JSDoc](https://github.com/realm/realm-js/blob/v10/docs/sync.js#L213) for `setLogger`.

cesarvr (commented):
> Why did you ~cross over~ updating the docs and types in the description of this PR? I suggest that you just leave them unchecked if you havn't done it or check the once that has "or is not necessary" written on them.
> 
> I would suspect this change also needs to update [the TypeScript definition](https://github.com/realm/realm-js/blob/v10/types/index.d.ts#L430) and [the JSDoc](https://github.com/realm/realm-js/blob/v10/docs/sync.js#L213) for `setLogger`.


Thanks for the heads up totally forgot we support TS. 👍

src/common/js_logger.hpp
cesarvr: Who knows, I think it makes no harm to offer those possibilities for whom might need that someday.

binding.gyp
```diff
@@ -29,8 +29,7 @@
             "src/js_email_password_auth.hpp",
             "src/js_api_key_auth.hpp",
             "src/js_auth.hpp",
-            "src/node/sync_logger.cpp",
-            "src/node/sync_logger.hpp",
+            "common/js_logger.hpp"
```
kneth, line 32: Is the path correct? Shouldn't it be `src/common/js_logger.hpp`?

See also my comment below.

src/common/js_logger.hpp
```diff
@@ -0,0 +1,130 @@
...
+#pragma once
+#include <iostream>
+#include "sync/sync_manager.hpp"
+
+#if REALM_ANDROID
+    #include <android/log.h>
+#endif
+
+namespace realm {
+namespace common {
```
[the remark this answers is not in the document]
cesarvr: Don't know :/, now that you mention I think the ``js_`` part in ``js_logger.hpp`` is wrong because that class doesn't have any JS behaviour inside, it just a logging delegator. 

We can call it just ``logger.hpp`` and maintain the namespace, so we can put there classes that provide similar services (reusable classes that encapsulate things that has nothing to do with JS but are consumed by JS classes) ?

binding.gyp
cesarvr: it works on my machine : ). 

I can rename that if you want, it work both ways.

src/common/js_logger.hpp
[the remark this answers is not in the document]
kneth: Good point calling it `logger.hpp`. Since we only have one file in the `realm::common`, I don't think we should create a new folder for it.

kneth (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Added descriptive errors for `partitionValue` of unsupported formats & ranges.
+* Added Sync logging support for react-native iOS and Android.
```
```suggestion
* Added Sync logging support for react-native iOS and Android. ([#2491](https://github.com/realm/realm-js/issues/2491)
```

types/index.d.ts
```diff
@@ -430,7 +430,7 @@ declare namespace Realm {
         function getAllSyncSessions(user: Realm.User): [Realm.App.Sync.Session];
         function getSyncSession(user: Realm.User, partitionValue: string|number|ObjectId|null) : Realm.App.Sync.Session;
         function setLogLevel(app: App, logLevel: LogLevel): void;
-        function setLogger(app: App, callback: (level: NumericLogLevel, message: string) => void): void;
+        function setLogger(app: App, callback: (level: string, message: string) => void): void;
```
kneth, line 433: The API doc in `docs/sync.js` requires an update too.

kneth (approved):
I have only a minor comment which should be straight-forward to address.

types/index.d.ts
cesarvr: I think I make that amendment on [line 213](https://github.com/realm/realm-js/blob/022f5c2c49157e349320fd713de2eea01f11b655/docs/sync.js#L213-L214), if I'm not mistaken.

kraenhansen, line 433: Did you notice that the `LogLevel` type is already defined? I suggest using that instead of `string`.
```suggestion
        function setLogger(app: App, callback: (level: LogLevel, message: string) => void): void;
```

kraenhansen (changes requested):

src/logger.hpp
```diff
@@ -0,0 +1,130 @@
...
+protected:
+    void do_log(LoggerLevel level, std::string message){
+
+        std::lock_guard<std::mutex> lock(m_mutex); 
+        auto LoggerEntry = std::make_pair(map_level[level], message);
+        m_log_queue.push(LoggerEntry);
+
+        #if REALM_ANDROID
+        auto android_log_level = map_android_log_level[level]; 
+        __android_log_print(android_log_level,
```
blagoev, line 59: This does not seem write. Why is logging done here inplace.

blagoev (changes requested):
This implementation logs in place which is not what the feature should be doing. The logger should invoke a JS function which is registered by the user as a callback

bmunkholm:
If this is breaking the API then we can't just merge into v10.

src/logger.hpp
cesarvr: I think it's a good idea to match the behaviour observed on iOS, where the user can see the sync logs in they stdout there. Also I think it do no harm as its integrated with Android and the user can choose to opt-out by setting the log flag to off.

blagoev: currently there is not logging support for RN. This PR should be adding it. I don't know what you mean by "behaviour observed on iOS" but logging inplace is not what we want to do. 
As I have written above, this PR should implement a way to invoke a JS function which the developer has registered to be called as a callback in RN

docs/sync.js
```diff
@@ -210,8 +210,7 @@ class Sync {
     /**
      * A callback passed to `Realm.App.Sync.setLogger` when instrumenting the Realm Sync client with a custom logger.
      * @callback Realm.App.Sync~logCallback
-     * @param {number} level The level of the log entry between 0 and 8 inclusively.
-     * Use this as an index into `['all', 'trace', 'debug', 'detail', 'info', 'warn', 'error', 'fatal', 'off']` to get the name of the level.
+     * @param {string} level The level of the log which can be ['all', 'trace', 'debug', 'detail', 'info', 'warn', 'error', 'fatal', 'off'].
```
kneth, line 213: The callback get the log level as a number.

src/js_sync.hpp
```diff
@@ -734,16 +731,30 @@ void SyncClass<T>::set_sync_log_level(ContextType ctx, ObjectType this_object, A
...
-    app->sync_manager()->set_logger_factory(*new realm::node::SyncLoggerFactory(ctx, callback_fn));
+
+    Protected<typename T::GlobalContext> protected_ctx(Context<T>::get_global_context(ctx));
+    Protected<FunctionType> protected_callback(ctx, callback_fn);
+
+    logger_factory->logs([=](auto level, auto message){
+        HANDLESCOPE(protected_ctx)
+
+        ValueType arguments[2] = {
+            Value::from_string(protected_ctx, level),
```
kneth, line 749: Should be `from_number()`

src/logger.hpp
kneth: The idea is that we give the user the option to capture the sync client's log using a callback. If the user decide to use `console.log()`, it will end up in the android device's log.

kneth:
@kraenhansen 

> If this is breaking the API then we can't just merge into v10.

Why do you consider it a breaking change?

kraenhansen:
> Why do you consider it a breaking change?
@kneth because of the change to the types? It looks to me that @cesarvr changed the implementation to call the callback with a string as first argument? If that's not the case the types should be updated to reflect that and the breaking label removed.

blagoev:
I second @kraenhansen @bmunkholm. This feature should not be introducing a breaking change.

Issue #3331, opened 2020-10-10
to-many Relationship returning Circular object when ids are them same


gustavo-dev:
I currently have two schemas on my app: 

**Playlist**

```typescript
export default class Playlist extends Realm.Object {
    static schema = {
        name: 'Playlist',
        primaryKey: 'id',
        properties: {
            id: {type: 'int', indexed: true},
            title: 'string',
            songs: 'Song[]',
            author: 'string',
            isDownloaded: {type: 'bool', default: false},
        },
    }
}

```

**Song**

```typescript
export default class Song extends Realm.Object {
    static schema = {
        name: 'Song',
        primaryKey: 'id',
        properties: {
            id: {type: 'int', indexed: true},
            title: 'string',
            url: 'string',
            artwork: 'string',
            artist: 'string',
            stringId: 'string',
            slowed: 'bool',
        },
    }
}

```

## Goals
Display a `Playlist`'s  songs

## Expected Results
Songs being displayed correctly

## Actual Results
`Song` with the same id from the current `Playlist` is returned as Circular Object, that returns the playlist itself when accessed, after calling `toJSON()`

![image](https://user-images.githubusercontent.com/42685889/95659469-b4b6a280-0af7-11eb-8565-1ee963a53a9d.png)

When I console log this object before calling `toJSON()`, the `Song` is shown correctly
When I console log this object after I called `this.realm(Playlist).toJSON()`, it returns the playlist itself

```typescript
// '8' is the index of this Circular Object
console.log(realm.objects(Playlist)[0].songs[8]) // Works fine

console.log(realm.objects(Playlist).toJSON()[0].songs[8]) // Returns the `Playlist` itself, instead of the `Song`
```

## Steps to Reproduce
Create a `Playlist` containing an id equals to one of it's `Song`s.
Call `toJSON()`

## Code Sample
```typescript
const playlist = realm.objects(Playlist).toJSON()[0]
console.log(playlist)
```

## Version of Realm and Tooling
* Realm JS SDK Version: 6.1.3
* Node or React Native: 4.12.0
* Client OS & Version: Win 10
* Which debugger for React Native: None

steffenagger:
Error now located, expect a fix in the next release.

isnifer:
@gustavo-dev just a quick note — you don't have to write to `id` property, since it's a primary key and indexed by default :)

PR #3355, opened 2020-10-20
(v6) toJSON special cases

steffenagger:
## What, How & Why?
Added special cases for `Realm.Object.toJSON`:
- ignores all functions & internals
- converts `ArrayBuffer`s to base64 strings (currently these outputs as `{}`).
- *Added:* fix for https://github.com/realm/realm-js/issues/3331

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -136,9 +136,8 @@ module.exports = function(realmConstructor, context) {
         value: function (_, cache = new Map()) {
             // Construct a reference-id of table-name & primaryKey if it exists, or fall back to objectId.
             const { name: schemaName, primaryKey } = this.objectSchema();
-            const id = primaryKey
-                ? serializedPrimaryKeyValue(this[primaryKey])
-                 : schemaName + this._objectId();
+            const id = schemaName + "#" +
+                (primaryKey ? serializedPrimaryKeyValue(this[primaryKey]) : this._objectId());
 
             // Check if current objectId has already processed, to keep object references the same.
             const existing = cache.get(id);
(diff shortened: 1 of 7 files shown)


lib/extensions.js
```diff
@@ -162,10 +162,23 @@ module.exports = function(realmConstructor, context) {
             Object.keys(this)
                 .concat(Object.keys(Object.getPrototypeOf(this)))
                 .forEach(key => {
-                    const value = this[key];
-                    result[key] = value instanceof realmConstructor.Object || value instanceof realmConstructor.Collection
-                        ? value.toJSON(key, cache)
-                        : value;
+                    const org = this[key];
```
kraenhansen, line 165: I personally prefer not to abbreviate local variables like this - I'll let you decide if you want this suggestion.

```suggestion
                    const original = this[key];
```

lib/extensions.js
```diff
@@ -162,10 +162,23 @@ module.exports = function(realmConstructor, context) {
...
+
+                    if (org instanceof realmConstructor.Object || org instanceof realmConstructor.Collection) {
+                        // recursively trigger `toJSON` for Realm instances with the same cache.
+                        val = org.toJSON(key, cache);
+                    } else if (org instanceof ArrayBuffer) {
+                        // convert "data" to base64 strings.
+                        val = Buffer.from(org).toString("base64");
+                    }
+
+                    result[key] = val !== undefined ? val : org;
```
kraenhansen, line 181: I think this change could make us a bit more robust to the global being swapped (as some Jest environments turns out to do).

```suggestion
                    result[key] = typeof val !== "undefined" ? val : org;
```

kraenhansen (changes requested):

lib/extensions.js
steffenagger: Agreed. After creating the PR I gave it a workover.

kraenhansen (approved):

kneth (approved):
An entry in `CHANGELOG.md` would be nice 😃

kneth (approved):

blagoev (approved):

PR #3402, opened 2020-11-11
Implement S3 Webhook Analytics Metrics Integration

fronck:
## What, How & Why?

* Added capability to report analytics metrics to S3/webhook as well as MixPanel
* Refactored analytics submission code
* Removed dependency on deprecated request node module

(Internal task)

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

~* [ ] jsdoc files updated~


fronck:
Open question:  This implementation (and previous it seems) submit metric when being built on the CI.  Is that intentional?

lib/submit-analytics.js
```diff
@@ -23,6 +23,77 @@
...
+}
+
+class webhookProdDetails {
+    /**
+     * Path and credentials required to submit analytics through the webhook (production mode).
+     */
+    constructor() {
+        this.addressPrefix =
+                "https://webhooks.mongodb-realm.com/api/client/v2.0/app/realmsdkmetrics-zmhtm/service/metric_webhook/incoming_webhook/metric?data=";
+        this.addressSuffix = "&ip=1";    }
```
kraenhansen, line 40:
```suggestion
        this.urlPrefix =
                "https://webhooks.mongodb-realm.com/api/client/v2.0/app/realmsdkmetrics-zmhtm/service/metric_webhook/incoming_webhook/metric?ip=1&data=";
    }
```

lib/submit-analytics.js
```diff
@@ -23,6 +23,77 @@
...
 'use strict';
 
+let utils = require('./utils');
+
+const require_method = require;
+function node_require(module) {
+    return require_method(module);
+}
+
+class webhookProdDetails {
```
kraenhansen, line 33: Since these classes have no state, I suggest using a more (pure) functional approach here. There's little benefit from complicating this with a class in my opinion.

lib/submit-analytics.js
```diff
@@ -23,6 +23,77 @@
...
+                "https://webhooks.mongodb-realm.com/api/client/v2.0/app/realmsdkmetrics-zmhtm/service/metric_webhook/incoming_webhook/metric?data=";
+        this.addressSuffix = "&ip=1";    }
+
+    /**
+     * Constructs the full URL that will submit analytics to the webhook.
+     * @param  {Object} payload Information that will be submitted through the webhook.
+     * @returns {string} Complete analytics submission URL
+     */
+    buildRequest(payload) {
+        const request = this.addressPrefix + Buffer.from(JSON.stringify(payload.webHook), 'utf8').toString('base64') + this.addressSuffix;
```
kraenhansen, line 48: I would probably break this into local variables with appropriate names to increase readability.
```suggestion
        const jsonPayload = JSON.stringify(payload.webHook);
        const base64Payload = Buffer.from(jsonData, 'utf8').toString('base64');
        const request = this.addressPrefix + base64Payload + this.addressSuffix;
```

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                    });
+                return;
+            });
+
+            response.on('error', error => {
+                reject(
+                    {
+                        statusCode:  response.statusCode,
+                        statusMessage: response.statusMessage,
+                        errorMessage: error,
```
kraenhansen, line 139: Isn't `error` an instance of `Error` here and not a `string` which is implied by the `errorMessage` name?

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                    });
+                return;
```
kraenhansen, line 141: No need to return here - blocks automatically and implicitly `return undefined`.
```suggestion
                return;
```

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+function fetchPlatformData(context, eventName) {
     const os = require('os');
-    const request = require('request');
     const { machineId } = require("node-machine-id");
+    const environment = utils.getContext();
+
+    return new Promise((resolve, reject) => {
+        machineId()
+        .catch((e) => {
+            reject('Analytics submission failed');
```
kraenhansen, line 161: It's bad practice to reject a promise with a string for the same reason that one shouldn't throw a string. Although this is technically possible in JS, it won't have have a stack present, making debugging an error much more complicated.
```suggestion
            const err = new Error('Analytics submission failed');
            reject(err);
```

kraenhansen: If you moved this `.catch` after your call to `.then` any errors thrown from computing the payload would also be caught by this block and return a meaningful error.

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                        'OS Type': os.platform(),
+                        'OS Version': os.release(),
+                        'Node.js versions': process.versions
+                    }
+                }
+            };
+
+            resolve(payload);
+            return;
+        });
```
kraenhansen, line 207:
```suggestion
        }, reject);
```

If you don't move the `.catch` call down here, you need to pass `reject` into this call to `.then` or you risk an unhandled promise rejection.

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
-            machineId()
-                .catch() // Ignore errors
-                .then((identifier) => {
-                    if (!identifier) {
-                        identifier = sha256('unknown');
-                    }
+            fetchPlatformData(context, eventName)
+            .catch((e) => {
+                reject("Failed to fetch platform data  [Error: " + e + "]");
+                return;
```
kraenhansen, line 230: No need to return here and throw instances of `Error` instead of strings:

```suggestion
                const err = new Error(`Failed to fetch platform data: ${e.message}`);
                reject(err);
```

The code above assumes that lower levels are throwing instances of `Error` (to be able to read out the `message`);

I would however ask you to consider if wrapping this error is actually needed?

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                dispatchAnalytics(requestFactory, payload)
+                .catch((e) => {
+                    reject(e);
+                    return;
+                })
+                .then(() => {
+                    resolve('');
+                    return;
                 });
+            });
```
kraenhansen, line 251: Again, this might result in an unhandled promise rejection.

```suggestion
            }, reject);
```

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+}
+
+/**
+ * Generate and submit analytics data to servers.
+ * @param  {[webhookProdDetails|webhookStageDetails|mixpanelDetails]} requestFactory Submission URL generator
+ * @param  {Object} context contents of the app's package.json file
+ * @param  {string} eventName Custom tag that signifies what the app was doing when the analytics were submitted
+ * @returns {Promise} Promise that resolves when analytics data has been submitted
+ */
+async function submitAnalytics(requestFactory, context, eventName) {
```
kraenhansen, line 218: I think this function could be greatly simplified if it was simply awaiting a promise which resolved after 100ms.

```
await new Promise(resolve => setTimeout(resolve, 100));
```

If you did that, then you can `await` the result of the call to `fetchPlatformData` and avoid using the `.then` and `.catch` methods, which are the "older" and more complicated way to program with promises.

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+
+        .then((identifier) => {
+            if (!identifier) {
+                identifier = sha256('unknown');
+            }
 
-    module.exports = function(eventName, context) {
-        // Submit analytics after some time - we do this to give the application a chance to disable analytics.
+            // payloads for webhook and MixPanel differ slightly
+            const payload = {
```
kraenhansen, line 171:
```suggestion
            const payloads = {
```

It feels more appropriate to pluralize this, since we're basically building two separate objects here.

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
-                            'OS Version': os.release(),
-                            'Node.js versions': process.versions
-                        }
-                    };
-
-                    request(`https://api.mixpanel.com/track/?data=${Buffer.from(JSON.stringify(payload), 'utf8').toString('base64')}&ip=1`,
-                    () => { /* Analytics failed. Do nothing. */ });
+                if (versionTags[0] < 10) {
+                    // on v6.x, we submit both MixPanel and S3 analytics
+                    // MixPanel analytics make no distinction between production and staging modes
```
kraenhansen, line 238: I don't think I understand this comment.
This piece of code is about to be merged into v10 (and should probably be rebased onto `master`), but the version of Realm JS would never be lower than 10? Except if you're preparing this to be copied into the `v6` branch already with this piece of code?

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                    dispatchAnalytics(new mixpanelDetails(), payload);
```
kraenhansen, line 239: You would need to either await the promise returned here or call `.then(resolve, reject)` to avoid an unhandled rejected promise.

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                }
+
+                dispatchAnalytics(requestFactory, payload)
+                .catch((e) => {
+                    reject(e);
+                    return;
+                })
+                .then(() => {
+                    resolve('');
```
kraenhansen, line 248: Little need to resolve to an empty string over `undefined` here.

```suggestion
                    resolve();
```

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+                if (versionTags[0] < 10) {
+                    // on v6.x, we submit both MixPanel and S3 analytics
+                    // MixPanel analytics make no distinction between production and staging modes
+                    dispatchAnalytics(new mixpanelDetails(), payload);
+                }
+
+                dispatchAnalytics(requestFactory, payload)
+                .catch((e) => {
+                    reject(e);
+                    return;
```
kraenhansen, line 245: No need for a `return here` since it's the last statment of the block.

```suggestion
```

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+    return submitAnalytics(webhookStageDetails, context, eventName);
+}
+
+module.exports.fetchPlatformData = fetchPlatformData;
+module.exports.webhookProdDetails = webhookProdDetails;
+module.exports.webhookStageDetails = webhookStageDetails;
+module.exports.submitAnalytics = submitAnalytics;
+module.exports.submitProductionAnalytics = submitProductionAnalytics;
+module.exports.submitStageAnalytics = submitStageAnalytics;
+module.exports.isAnalyticsDisabled = isAnalyticsDisabled;
```
kraenhansen, line 283: Since these all have the same names, it is simpler to assign the `module.exports` object instead:

```suggestion
module.exports = {
  fetchPlatformData,
  webhookProdDetails,
  webhookStageDetails,
  submitAnalytics,
  submitProductionAnalytics,
  submitStageAnalytics,
  isAnalyticsDisabled,
};
```

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+
+if (require.main === module) {
+    module.submitProductionAnalytics('Install');
```
kraenhansen, line 286: Little need to call this off the `module` global, since it's already a local variable in scope here.

```suggestion
    submitProductionAnalytics('Install');
```

kraenhansen, line 286: Either warn the user about this or simply "swallow" this error. We need to catch a rejected promise never the less, or it might end up as an unhandled promise rejection.

```suggestion
    module.submitProductionAnalytics('Install').catch(err => {
        console.warn(`Failed to submit analytics: ${err.message}`);
    });
```

package.json
```diff
@@ -96,7 +96,6 @@
     "progress": "^2.0.3",
     "prop-types": "^15.6.2",
     "realm-network-transport": "^0.7.0",
-    "request": "^2.88.0",
```
kraenhansen, line 99: Did you run `npm install`? I would expect the package-lock.json to update as well after this change.

tests/js/session-tests.js
```diff
@@ -929,5 +929,31 @@ module.exports = {
...
+    },
+
+    async testAnalyticsSubmission() {
+        const context = node_require('realm/package.json');
+        const analytics = node_require('realm/lib/submit-analytics');
+
+        const payload = await analytics.fetchPlatformData(context, 'TestEvent')
+        .catch((e) => {
+            return Promise.reject("Failed to fetch platform data  [Error: " + e + "]");
+        });
```
kraenhansen, line 941: I would have left this `.catch` out, since you're awaiting the result of the call to `fetchPlatformData` the test will fail if the promise returned gets rejected. The only reason that you would want to add this `catch` would be to "extend" the message, but that probably makes little sense in a test, since you don't expect this to fail at all.

```suggestion
        const payload = await analytics.fetchPlatformData(context, 'TestEvent');
```

tests/js/session-tests.js
```diff
@@ -929,5 +929,31 @@ module.exports = {
...
+        TestCase.assertType(payload.webHook.properties.Binding, 'string');
+        TestCase.assertDefined(payload.mixPanel);
+        TestCase.assertType(payload.mixPanel.event, 'string');
+        TestCase.assertDefined(payload.mixPanel.properties);
+        TestCase.assertType(payload.mixPanel.properties.Binding, 'string');
+
+        await analytics.submitStageAnalytics('TestEvent')
+        .catch((e) => {
+            return Promise.reject("Failed to submit webhook analytics  [" + e + "]")
+        });
```
kraenhansen, line 955: Again, since we're not expecting this to happen in the test and this would only benefit our developer in debugging a future issue, I would probably just leave this `.catch` out.
```suggestion
        await analytics.submitStageAnalytics('TestEvent');
```

tests/js/session-tests.js
```diff
@@ -929,5 +929,31 @@ module.exports = {
...
+
+        return Promise.resolve();
```
kraenhansen, line 957: No need for this - the function is `async` so it always implicitly returns a resolved promise of `undefined` when it reaches this point of execution.
```suggestion
```

kraenhansen (changes requested):
I have a few requests for changes, let's perhaps have a face-2-face conversation around these?

lib/index.js
```diff
@@ -18,75 +18,17 @@
...
-    throw new Error("Unknown execution context");
-}
-
 function getRealmConstructor(context) {
     switch(context) {
         case 'node.js':
         case 'electron':
-            nodeRequire('./submit-analytics')('Run', context);
+            var analytics = nodeRequire('./submit-analytics');
+            analytics.submitProductionAnalytics('Run');
```
kneth, line 31: We only submit analytics for node/electron and not React Native?

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+ * Collect analytics data from the runtime system
+ * @param  {Object} context contents of the app's package.json file
+ * @param  {string} eventName Custom tag that signifies what the app was doing when the analytics were submitted
+ * @returns {Object} Analytics payload
+ */
+function fetchPlatformData(context, eventName) {
     const os = require('os');
-    const request = require('request');
     const { machineId } = require("node-machine-id");
+    const environment = utils.getContext();
```
kneth, line 156: I find it mildly confusing that `getContext()` returns an `environment` which is different from the function's first argument (`context`).

lib/submit-analytics.js
```diff
@@ -33,50 +104,184 @@ function sha256(data) {
...
+            const payload = {
+                'webHook': {
+                    'event': eventName,
+                    'properties': {
+                        'token': 'aab85907a13e1ff44a95be539d9942a9',
+                        'distinct_id': identifier,
+                        'Anonymized Machine Identifier': identifier,
+                        'Anonymized Application ID': sha256(__dirname),
+                        'Binding': 'javascript',
+                        'Target': environment,
```
steffenagger, line 180: I think I suggested using the existing context determination 😬 but I didn't think about install-time.
During `"Install"`, I think `"Target"` will always be `"node.js"`..?

steffenagger (changes requested):
I think `environment` will always be "node.js" for install-analytics.

lib/index.js
fronck: As per extensive discussions, it will be kept the way it is for now.
Possible reimplementation to be discussed in the future.

lib/submit-analytics.js
fronck: As per request from PM, it will be kept as-is for now.  It reflects how we submit analytics to MixPanel.

lib/submit-analytics.js
```diff
@@ -23,6 +23,77 @@
...
+function node_require(module) {
+    return require_method(module);
+}
+
+class webhookProdDetails {
+    /**
+     * Path and credentials required to submit analytics through the webhook (production mode).
+     */
+    constructor() {
+        this.addressPrefix =
```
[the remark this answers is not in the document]
fronck: I've changed it.

lib/submit-analytics.js
fronck: Actually, this error is mostly for debugging purposes.  In production, it should fail silently.  Would it be more appropriate to resolve() the promise in the .catch()?

lib/submit-analytics.js
fronck: +1

lib/submit-analytics.js
fronck: `.catch` has been duly moved.

lib/submit-analytics.js
fronck: Yes, the exact same code will be merged into v6, which will report both to S3 and MixPanel.

lib/submit-analytics.js
fronck: Fair enough.  I've renamed the function to `getEnvironment()`

lib/submit-analytics.js
fronck: Done

package.json
fronck: package-lock.json added.  Thanks :)

kneth (approved):

kneth:
@fronck Do we need to:

* backport it to v6.x?
* add similar submission to [`RealmAnalytics.mm](https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmAnalytics.mm) and [`analytics_template](https://github.com/realm/realm-js/blob/master/react-native/android/analytics_template)?

fronck:
@kneth 
>     * backport it to v6.x?

Yes, there's a ticket for that.

>     * add similar submission to [`RealmAnalytics.mm`](https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmAnalytics.mm) and [`analytics_template`](https://github.com/realm/realm-js/blob/master/react-native/android/analytics_template)?

Yes.  I'll create a new ticket for that as well.

steffenagger (commented):
General question, when or what events are we suppose to track?

I see that `submitProductionAnalytics('Run')` is triggered for node & electron.
`submitProductionAnalytics('Install')` is run if `require.main === module` - when are we triggering that? (I'm assuming that this is to run directly ex.: `node ./submit-analytics.js`).

lib/submit-analytics.js
```diff
@@ -33,50 +103,173 @@ function sha256(data) {
...
-if (isAnalyticsDisabled()) {
-    module.exports = function(){};
-} else {
+/**
+ * Send collected analytics data to Realm's servers over HTTPS
+ * @param  {[WebhookProdDetails|WebhookStageDetails|MixpanelDetails]} requestFactory Submission URL generator
+ * @param  {Object} payload analytics info
+ */
+async function dispatchAnalytics(requestFactory, payload) {
+    const https = require('https');
```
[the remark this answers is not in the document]
fronck: I think that's a possibly issue that we will have to tackle in an upcoming rework of the metrics reporting process across SDKs.

lib/submit-analytics.js
```diff
@@ -33,50 +103,173 @@ function sha256(data) {
...
+
+    return new Promise((resolve, reject) => {
+        const webhookRequest = new requestFactory().buildRequest(payload);
+        var req = https.get(webhookRequest, function (result) {
+            result.on('data', (data) => {
+                // we don't log the data that is returned -- we're just interested in the header
```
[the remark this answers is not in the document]
fronck: What we really want, is the status codes that tell us whether the `GET` operation was successful.
Those are captured in the `req.on('response')` clause.

fronck (commented):
> General question, when or what events are we suppose to track?

The short answer to that, currently, is "the same as the old implementation, just to both MixPanel and S3".
Aligning SDKs on exactly what to report, and when, is a future project.

lib/submit-analytics.js
```diff
@@ -33,50 +103,173 @@ function sha256(data) {
...
 
-    if (require.main === module) {
-        module.exports('Install');
+    const payload = await fetchPlatformData(context, eventName)
+    const versionTags = payload['mixPanel']['properties']['Version'].split('.');
+
+    if (versionTags[0] < 10) {
+        // on v6.x, we submit both MixPanel and S3 analytics
+        // MixPanel analytics make no distinction between production and staging modes
+        await dispatchAnalytics(new MixpanelDetails(), payload);
```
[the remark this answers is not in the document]
fronck: I have un-exported `submitAnalytics`, as it was only really exposed for debugging purposes.  Thanks.

lib/submit-analytics.js
```diff
@@ -33,50 +103,173 @@ function sha256(data) {
...
+                resolve(
+                    {
+                        statusCode:  response.statusCode,
+                        statusMessage: response.statusMessage,
+                        errorMessage: '',
+                    });
+                return;
+            });
+
+            response.on('error', error => {
```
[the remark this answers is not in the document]
fronck: You're right.  Good `catch` (pun intended), thanks.  :)

lib/submit-analytics.js
```diff
@@ -23,6 +23,76 @@
...
+const nodeRequire = require;
+
+class WebhookProdDetails {
+    /**
+     * Path and credentials required to submit analytics through the webhook (production mode).
+     */
+    constructor() {
+        this.urlPrefix =
+                "https://webhooks.mongodb-realm.com/api/client/v2.0/app/realmsdkmetrics-zmhtm/service/metric_webhook/incoming_webhook/metric?ip=1&data=";
+        this.urlSuffix = "&ip=1";
```
kraenhansen, line 38: This seems redundant now that `?ip=1` is already in the `urlPrefix`?

```suggestion
```

lib/submit-analytics.js
```diff
@@ -23,6 +23,76 @@
...
+    }
+
+    /**
+     * Constructs the full URL that will submit analytics to the webhook.
+     * @param  {Object} payload Information that will be submitted through the webhook.
+     * @returns {string} Complete analytics submission URL
+     */
+    buildRequest(payload) {
+        const request = this.urlPrefix + Buffer.from(JSON.stringify(payload.webHook), 'utf8').toString('base64') + this.urlSuffix;
```
kraenhansen, line 47: I see little need for a suffix since url query parameters can be considered unordered.
```suggestion
        const request = this.urlPrefix + Buffer.from(JSON.stringify(payload.webHook), 'utf8').toString('base64');
```

lib/submit-analytics.js
```diff
@@ -23,6 +23,76 @@
...
+};
+
+class WebhookStageDetails {
+    /**
+     * Path and credentials required to submit analytics through the webhook (staging mode).
+     */
+    constructor() {
+        this.urlPrefix =
+                "https://webhooks.mongodb-realm.com/api/client/v2.0/app/realmsdkmetrics-zmhtm/service/metric_webhook/incoming_webhook/metric-stage?data=";
+        this.urlSuffix = "&ip=1";
```
kraenhansen, line 60:
```suggestion
                "https://webhooks.mongodb-realm.com/api/client/v2.0/app/realmsdkmetrics-zmhtm/service/metric_webhook/incoming_webhook/metric-stage?ip=1&data=";
```

lib/submit-analytics.js
```diff
@@ -33,50 +103,173 @@ function sha256(data) {
...
+    });
+}
+
+/**
+ * Collect analytics data from the runtime system
+ * @param  {Object} context contents of the app's package.json file
+ * @param  {string} eventName Custom tag that signifies what the app was doing when the analytics were submitted
+ * @returns {Object} Analytics payload
+ */
+function fetchPlatformData(context, eventName) {
```
kraenhansen, line 152: I suggest making this an async function and `await` the promise returned by calling `machineId()`

```suggestion
async function fetchPlatformData(context, eventName) {
```

lib/utils.js
```diff
@@ -116,7 +116,7 @@ module.exports = {
     getVersions() {
         const packageJson = require("../package.json");
         const packageVersion = packageJson.version;
-        const context = this.getContext();
+        const context = this.getEnvironment();
```
kraenhansen, line 119: If we rename this to environment, we should be consistent in the naming throughout the code.
```suggestion
        const environment = this.getEnvironment();
```

lib/index.js
```diff
@@ -58,10 +63,10 @@ function getRealmConstructor(context) {
     }
 }
 
-const context = utils.getContext();
-const realmConstructor = getRealmConstructor(context);
+const environment = utils.getEnvironment();
+const realmConstructor = getRealmConstructor(environment);
 
-require("./extensions")(realmConstructor, context);
+require("./extensions")(realmConstructor, environment);
```
kraenhansen, line 69: Perhaps [the local variable on the other side of this call](https://github.com/realm/realm-js/blob/ff/metrics_s3/lib/extensions.js#L68) should be renamed from `context` to `environment` as well, to keep naming consistent?

kraenhansen (changes requested):

lib/submit-analytics.js
fronck: Fixed in #3524 .

lib/submit-analytics.js
fronck: Fixed in #3524 .

lib/submit-analytics.js
fronck: Fixed in #3524 .

lib/submit-analytics.js
fronck: Fixed in #3524 .

PR #3408, opened 2020-11-13
Cesarvr/sync logger documentation

cesarvr:
Revert back the breaking change ``<string>`` to ``<int>`` to set log levels in Javascript, improve the readability of ``the set_log_level`` member function and added some high level documentation. 

> Also the formatting here is ``clang-format`` using the default ``google`` style which I used as provisionally until we agree on a proper style configuration.


kneth (approved):
LGTM

contrib/sync_logging.md
```diff
@@ -0,0 +1,15 @@
...
+
+```js
+   Realm.App.Sync.setLogLevel()
+   Realm.App.Sync.setLogger()
+```
+
+These two functions are the entry point to configure the sync behaviour which is handled natively. Then in C++ land, we use the ``logger.hpp`` class to isolate our JS engine (``js_sync.hpp``) from the other two frameworks we depend on which are ``object-store`` and ``realm-core``.  
+
+
+This is important because we don't want changes on ``realm-core`` or ``object-store`` (related to the logger) to break other parts of our JS engine and leave us clueless, we want to look at this class for anything logging related.
```
fronck, line 15: I don't think the phrasing "leave us clueless" is very helpful in public-facing documentation.  Can you elaborate on what we will be missing?

src/js_sync.hpp
```diff
@@ -724,26 +721,37 @@ void SyncClass<T>::set_sync_log_level(ContextType ctx, ObjectType this_object, A
...
+            Value::from_string(protected_ctx, message)
+        };
+
+        Function::callback(protected_ctx, protected_callback, typename T::Object(), 2, arguments);
+    };
+
+    common::Logger logger;
+    auto sync_logger = logger.build_sync_logger(std::move(show_logs));
+    app->sync_manager()->set_logger_factory( *sync_logger );
 }
```
fronck, line 754: [Code Quality]  I believe we agreed on documenting functions that we added or substantially refactored in comments.  I'd suggest at least documenting inputs, outputs, and purpose of the function.

src/logger.hpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+private:
+    std::queue<Entry> m_log_queue;
+    Delegated loggerDelegate;
+    std::shared_ptr<realm::util::Scheduler> m_scheduler = realm::util::Scheduler::make_default();
+    std::mutex m_mutex;
+};
+
+class LoggerFactory : public realm::SyncLoggerFactory {
```
fronck, line 111: [Nitpick]  It seems somewhat confusing that `LoggerFactory` would be a subclass of `SyncLoggerFactory`.  I suggest perhaps ranaming `LoggerFactory` to `JSSyncLoggerFactory`?

src/logger.hpp
```diff
@@ -0,0 +1,164 @@
...
+        logger->delegate(std::move(logs_fn));
+
+        return logger;
+    }
+
+private:
+    Delegated logs_fn;
+};
+
+class Logger {
```
fronck, line 132: [Nitpick]  If you decide to rename `LoggerFactory`, probably this class should be renamed to reflect the new name as well.

src/logger.hpp
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        throw std::runtime_error("Bad log level");
+    }
+
+    LoggerFactory* build_sync_logger(Delegated&& log_fn)
+    {
+        return new LoggerFactory{std::move(log_fn)};
+    }
+};
```
fronck, line 161: Is there any particular reason why the two functions above aren't `static`?  That would eliminate the need to instantiate a `Logger` class to retrieve a `LoggerFactory`.

fronck (approved):
Please see attached comments.
For suggestion purposes only - they aren't merge blocking.

src/logger.hpp
cesarvr: yeah but the class doesn't do any JS work, so I think it could be misleading ? I think putting the ``Delegate`` suffix might be overkill as well ?

src/js_sync.hpp
```diff
@@ -724,26 +721,35 @@ void SyncClass<T>::set_sync_log_level(ContextType ctx, ObjectType this_object, A
...
+
+        ValueType arguments[2] = {
+            Value::from_number(protected_ctx, level),
+            Value::from_string(protected_ctx, message)
+        };
+
+        Function::callback(protected_ctx, protected_callback, typename T::Object(), 2, arguments); 
+    };
+
+    auto sync_logger = common::logger::Logger::build_sync_logger(show_logs);
```
fronck, line 750: This bit here seems to result in a memory leak.
`sync_logger` is a pointer.  The `SyncManager` stores this pointer, but doesn't take ownership over it.  And since `SyncClass<T>` is an N-API interface class, it is rather static, and can't really keep ownership of the pointer either.  Thus, no-one will deallocate it.
Can anyone with more insight into `SyncManager` shed some light on how to handle this issue (e.g., @kneth)?

fealebenpae: Good catch, @fronck. I made https://github.com/realm/realm-core/issues/4117 to track this.

Issue #3361, opened 2020-10-26
Must first create RPC session with a valid host


Muzzamil75:
URGENT

Getting this error when debugging from chrome. Its been a while but issue is not resolved till now.

kneth:
@Muzzamil75 Thank you for reporting.

kraenhansen:
I'm actively working on this, but I'm unable to provide a good ETA since all the work needed is not entirely clear to me at this point in time. I appreciate the patience and will update here as I approach a solution.

charleycesar:
Thanks I'm waiting for this solution.

kraenhansen:
I'm closing this as a duplicate of https://github.com/realm/realm-js/issues/3358 - please follow that issue instead of this.

Issue #2455, opened 2019-07-12
Support Hermes engine


ferrannp:
## Goals

Make [Hermes](https://hermesengine.dev/) engine to work with Realm. Hermes comes from react-native 0.60.2 and significantly improves app performance on Android.

##  Steps to Reproduce

Following https://facebook.github.io/react-native/docs/hermes. if `enableHermes: true`, you will get the following:

```java
java.lang.UnsatisfiedLinkError: couldn't find DSO to load: librealmreact.so caused by: dlopen failed: library "libjsc.so" not found
        at com.facebook.soloader.SoLoader.doLoadLibraryBySoName(SoLoader.java:738)
        at com.facebook.soloader.SoLoader.loadLibraryBySoName(SoLoader.java:591)
        at com.facebook.soloader.SoLoader.loadLibrary(SoLoader.java:529)
        at com.facebook.soloader.SoLoader.loadLibrary(SoLoader.java:484)
        at io.realm.react.RealmReactModule.<clinit>(RealmReactModule.java:56)
        at io.realm.react.RealmReactPackage.createNativeModules(RealmReactPackage.java:31)
```

With `enableHermes: false`, everything works as expected.

## Version of Realm and Tooling

- Realm JS SDK Version: 2.28.1
- React Native: 0.60.3

cbstauss:
I'm having the same issue. I'm wondering if there is something looking for "libjsc.so" that can't find it because hermes replaces "libjsc.so" with "libhermes.so". I've been looking through the SoLoader documentation and the Realm documentation to see if I can find anything.

fealebenpae:
realm-js uses JavaScriptCore APIs directly on React Native to implement its JS API so right now it cannot run on Hermes. We're planning on migrating to the new API in the coming months.

ferrannp:
Yeah I was pretty sure this was not an easy change. Glad to hear it is planned to happen @fealebenpae. Thank you!

paschaldev:
+1

Spent over 48hours debugging this issue. When I disabled hermes, it started working.

dominicsavb:
can't wait to implement realm with hermes...

matinzd:
We got stuck migrating project to hermes :((

raphaeleidus:
@fealebenpae firstly, thank you for all the hard work creating this amazing software.
Do you have any rough idea as to when Hermes support might arrive? 

[...]

PR #3411, opened 2020-11-18
Fixing chrome debugging on v10

### LINKED ISSUE #3358: Enabling React Native Debug mode causes Error: Must first create RPC session with a valid host from rpc.js:314
## Goals
Enable React Native **Debug** mode in order to troubleshoot console errors, set breakpoints, profile app, and monitor network requests.

## Expected Results
Mobile app connects to React Native Debugger successfully and I'm able to set breakpoints,  profile app, and monitor network requests, view console errors

## Actual Results
* Mobile App crashes with the following error

> Error: Must first create RPC session with a valid host from rpc.js:314

:19

Mobile App

### Sentry Stacktrace
```jsx
Error: Must first create RPC session with a valid host
  File "app:///rpc.js", col 18, in sendRequest
  File "app:///rpc.js", col 41, in setTimeout$argument_0
  File "app:///JSTimers.js", col 17, in _allocateCallback$argument_0
  File "app:///JSTimers.js", col 6, in _callTimer
  File "app:///JSTimers.js", col 6, in callTimers
  at <unknown>(app:///RNDebuggerWorker.js:2:134503)
```

## Steps to Reproduce

1. Run `yarn android` to build react native application and install on device
1. Shake device to reveal developer menu and select `Debug` . This is successful if the user is not logged in

   image

1. Login to account which causes realm database to be initialized
1. An exception occurs on mobile app

```
![https://user-images.githubusercontent.com/222459/96701643-bea69400-1388-11eb-911e-9ee8323b2795.png](https://user-images.githubusercontent.com/222459/96701643-bea69400-1388-11eb-911e-9ee8323b2795.png)
```

    The crash occurs when the following is called **`Realm.schemaVersion(Realm.defaultPath)`**     **`Realm.defaultPath`** displays an error when inspected using React Native debugger. See video below```
  image
```

## Code Sample
<!---
Please provide a code sample or test case that highlights the issue.
If relevant, include your model definitions.
For larger code samples, links to external gists/repositories are preferred.
Full projects that we can compile and run ourselves are ideal!
-->

```jsx
export const runMigration = ({currentSchema}: {currentSchema: any}) => {
  if (!schemas.length) {
    return {schemaVersion: 0};
  }

  **let nextSchemaIndex = Realm.schemaVersion(Realm.defaultPath);**
  schemas[schemas.length - 1].schema = currentSchema;

  // If -1, it means this is a new Realm file, so no migration is needed
  if (nextSchemaIndex !== -1) {
    while (nextSchemaIndex < schemas.length) {
      const migratedRealm = new Realm(schemas[nextSchemaIndex++]);
      migratedRealm.close();
    }
  }

  return {schemaVersion: schemas[schemas.length - 1].schemaVersion};
};
```

* Realm JS SDK Version: 10.0.1
* Node or React Native: 0.63.3
* Client OS & Version: OnePlus 6 A6003, Android 10
* Which debugger for React Native: React Native Debugger

kraenhansen:
## What, How & Why?

This closes #3358 and #3361 by
1. Fixing the implementation of the RPC stub for `App`
2. Introducing a `RPCNetworkTransport` implementation of the `GenericNetworkTransport` provided by object store. This struct is statically provided a function (assigned by the `RPCServer` when a session gets created), which calls back into the browser to execute fetch requests, ultimately on behalf of object store.
3. Bonus: Introduced a better serialization+deserialization of `Error` instances to propagate the stack when an error is thrown during an RPC call.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -165,11 +165,11 @@ export function callSyncFunction(name, args) {
 }
 
 export function callMethod(realmId, id, name, args) {
-    if (args && Array.isArray(args)) {
-        args = args.map((arg) => serialize(realmId, arg));
+    if (!Array.isArray(args)) {
+        throw new Error("Expected an array of arguments");
     }
-
-    let result = sendRequest("call_method", { realmId, id, name, arguments: args });
+    const serializedArgs = args.map((arg) => serialize(realmId, arg));
+    const result = sendRequest("call_method", { realmId, id, name, arguments: serializedArgs });
     return deserialize(realmId, result);
 }
 
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -65,6 +65,14 @@ function beforeNotify(realm) {
     invalidateCache(realm[keys.realm]);
 }
 
+
+export function createSession(host) {
+    sessionId = sendRequest("create_session", null, host);
+    sessionHost = host;
+
+    return sessionId;
+}
+
 export function createRealm(args) {
     if (args) {
         args = args.map((arg) => serialize(null, arg));
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -19,7 +19,6 @@
 'use strict';
 
 import * as base64 from './base64';
-import * as util from './util';
 import { keys, objectTypes } from './constants';
 import { invalidateCache } from './cache';
 
@@ -96,6 +95,19 @@ export function asyncOpenRealm(id, config, callback) {
                     realm.addListener('beforenotify', beforeNotify);
                 }
                 callback(realm, error);
+            })
+        ]
+    }));
+}
+
+export function _logInRPC(id, creds, callback) {
+    return deserialize(undefined, sendRequest("call_method", {
+        id,
+        name: "_logIn",
+        arguments: [
+            serialize(null, creds),
+            serialize(null, (user, error) => {
+                callback(user, error);
             })
         ]
     }));
--- a/lib/browser/rpc.js
+++ b/lib/browser/rpc.js
@@ -16,11 +16,11 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-'use strict';
+"use strict";
 
-import * as base64 from './base64';
-import { keys, objectTypes } from './constants';
-import { invalidateCache } from './cache';
+import * as base64 from "./base64";
+import { keys, objectTypes } from "./constants";
+import { invalidateCache } from "./cache";
 
 const { id: idKey, realm: _realmKey } = keys;
 let registeredCallbacks = [];
@@ -45,6 +45,7 @@ if (XMLHttpRequest.__proto__ != global.XMLHttpRequestEventTarget) {
 registerTypeConverter(objectTypes.DATA, (_, { value }) => base64.decode(value));
 registerTypeConverter(objectTypes.DATE, (_, { value }) => new Date(value));
 registerTypeConverter(objectTypes.DICT, deserializeDict);
+registerTypeConverter(objectTypes.ERROR, deserializeError);
 registerTypeConverter(objectTypes.FUNCTION, deserializeFunction);
 
 export function registerTypeConverter(type, handler) {
@@ -65,9 +66,14 @@ function beforeNotify(realm) {
 }
 
 
-export function createSession(host) {
+export function createSession(host, { performFetch }) {
     sessionHost = host;
-    sessionId = sendRequest("create_session", null, host);
+
+    sessionId = sendRequest(
+        "create_session",
+        { fetch: serialize(undefined, performFetch) },
+        host,
+    );
 
     return sessionId;
 }
@@ -77,7 +83,7 @@ export function createRealm(args) {
         args = args.map((arg) => serialize(null, arg));
     }
 
-    return sendRequest('create_realm', { arguments: args, beforeNotify: serialize(null, beforeNotify) });
+    return sendRequest("create_realm", { arguments: args, beforeNotify: serialize(null, beforeNotify) });
 }
 
 export function createAppRPC(config) {
@@ -85,14 +91,14 @@ export function createAppRPC(config) {
 }
 
 export function asyncOpenRealm(id, config, callback) {
-    return deserialize(undefined, sendRequest('call_method', {
+    return deserialize(undefined, sendRequest("call_method", {
         id,
-        name: '_asyncOpen',
+        name: "_asyncOpen",
         arguments: [
             serialize(null, config),
             serialize(null, (realm, error) => {
                 if (realm) {
-                    realm.addListener('beforenotify', beforeNotify);
+                    realm.addListener("beforenotify", beforeNotify);
                 }
                 callback(realm, error);
             })
@@ -100,88 +106,75 @@ export function asyncOpenRealm(id, config, callback) {
     }));
 }
 
-export function _logInRPC(id, creds, callback) {
-    return deserialize(undefined, sendRequest("call_method", {
-        id,
-        name: "_logIn",
-        arguments: [
-            serialize(null, creds),
-            serialize(null, (user, error) => {
-                callback(user, error);
-            })
-        ]
-    }));
-}
-
 export function _anonymousRPC() {
-    const result = sendRequest('_anonymous', { arguments: undefined });
+    const result = sendRequest("_anonymous", { arguments: undefined });
     return deserialize(undefined, result);
 }
 
 export function _appleRPC(token) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_apple', { arguments: args });
+    const result = sendRequest("_apple", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function _emailPasswordRPC(email, password) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_emailPassword', { arguments: args });
+    const result = sendRequest("_emailPassword", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function _facebookRPC(token) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_facebook', { arguments: args });
+    const result = sendRequest("_facebook", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function _functionRPC(payload) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_function', { arguments: args });
+    const result = sendRequest("_function", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function _googleRPC(authCode) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_google', { arguments: args });
+    const result = sendRequest("_google", { arguments: args });
     return deserialize(undefined, result);
 }
 export function _userApiKeyRPC(user_key) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_userAPIKey', { arguments: args });
+    const result = sendRequest("_userAPIKey", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function _serverApiKeyRPC(server_key) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_serverAPIKey', { arguments: args });
+    const result = sendRequest("_serverAPIKey", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function _jwtRPC(token) {
     const args = Array.prototype.map.call(arguments, (arg) => serialize(null, arg));
-    const result = sendRequest('_jwt', { arguments: args });
+    const result = sendRequest("_jwt", { arguments: args });
     return deserialize(undefined, result);
 }
 
 export function callSyncFunction(name, args) {
     args = (args || []).map((arg) => serialize(null, arg));
-    let result = sendRequest('call_sync_function', { name, arguments: args });
+    let result = sendRequest("call_sync_function", { name, arguments: args });
     return deserialize(undefined, result);
 }
 
 export function callMethod(realmId, id, name, args) {
-    if (args) {
+    if (args && Array.isArray(args)) {
         args = args.map((arg) => serialize(realmId, arg));
     }
 
-    let result = sendRequest('call_method', { realmId, id, name, arguments: args });
+    let result = sendRequest("call_method", { realmId, id, name, arguments: args });
     return deserialize(realmId, result);
 }
 
 export function getObject(realmId, id, name) {
-    let result = sendRequest('get_object', { realmId, id, name });
+    let result = sendRequest("get_object", { realmId, id, name });
     if (!result) {
         return result;
     }
@@ -192,17 +185,17 @@ export function getObject(realmId, id, name) {
 }
 
 export function getProperty(realmId, id, name) {
-    let result = sendRequest('get_property', { realmId, id, name });
+    let result = sendRequest("get_property", { realmId, id, name });
     return deserialize(realmId, result);
 }
 
 export function setProperty(realmId, id, name, value) {
     value = serialize(realmId, value);
-    sendRequest('set_property', { realmId, id, name, value });
+    sendRequest("set_property", { realmId, id, name, value });
 }
 
 export function clearTestState() {
-    sendRequest('clear_test_state');
+    sendRequest("clear_test_state");
 
     // Clear all registered callbacks that are specific to this session.
     registeredCallbacks = registeredCallbacks.filter(cb => Reflect.has(cb, persistentCallback));
@@ -214,13 +207,13 @@ function registerCallback(callback) {
 }
 
 function serialize(realmId, value) {
-    if (typeof value == 'undefined') {
+    if (typeof value == "undefined") {
         return { type: objectTypes.UNDEFINED };
     }
-    if (typeof value == 'function') {
+    if (typeof value == "function") {
         return { type: objectTypes.FUNCTION, value: registerCallback(value) };
     }
-    if (!value || typeof value != 'object') {
+    if (!value || typeof value != "object") {
         return { value: value };
     }
 
@@ -273,6 +266,13 @@ function deserializeDict(realmId, info) {
     return object;
 }
 
+function deserializeError(realmId, info) {
+    const { message, stack } = info.error;
+    const err =  new Error(message.value);
+    err.stack = stack.value;
+    return err;
+}
+
 function deserializeFunction(realmId, info) {
     return registeredCallbacks[info.value];
 }
@@ -283,8 +283,8 @@ function makeRequest(url, data) {
 
     // The global __debug__ object is provided by Visual Studio Code.
     if (global.__debug__) {
-        let request = global.__debug__.require('sync-request');
-        let response = request('POST', url, {
+        let request = global.__debug__.require("sync-request");
+        let response = request("POST", url, {
             body: JSON.stringify(data),
             headers: {
                 "Content-Type": "text/plain;charset=UTF-8"
@@ -292,12 +292,12 @@ function makeRequest(url, data) {
         });
 
         statusCode = response.statusCode;
-        responseText = response.body.toString('utf-8');
+        responseText = response.body.toString("utf-8");
     } else {
         let body = JSON.stringify(data);
         let request = new XMLHttpRequest();
 
-        request.open('POST', url, false);
+        request.open("POST", url, false);
         request.send(body);
 
         statusCode = request.status;
@@ -320,7 +320,7 @@ function deserialize_json_value(value) {
     for (let index = 0; index < value.keys.length; index++) {
         var propName = value.keys[index];
         var propValue = value.values[index];
-        if (propValue.type && propValue.type == 'dict') {
+        if (propValue.type && propValue.type == "dict") {
             result[propName] = deserialize_json_value(propValue);
         }
         else {
@@ -335,19 +335,19 @@ function sendRequest(command, data, host = sessionHost) {
     clearTimeout(pollTimeoutId);
     try {
         if (!host) {
-            throw new Error('Must first create RPC session with a valid host');
+            throw new Error("Must first create RPC session with a valid host");
         }
 
         data = Object.assign({}, data, sessionId ? { sessionId } : null);
 
-        let url = 'http://' + host + '/' + command;
+        let url = "http://" + host + "/" + command;
         let response = makeRequest(url, data);
         let callback = response && response.callback;
 
         // Reset the callback poll interval to 10ms every time we either hit a
         // callback or call any other method, and double it each time we poll
         // for callbacks and get nothing until it's over a second.
-        if (callback || command !== 'callbacks_poll') {
+        if (callback || command !== "callbacks_poll") {
             pollTimeout = 10;
         }
         else if (pollTimeout < 1000) {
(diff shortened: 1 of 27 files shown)


Anzormumladze:
are fucking kidding me ?

kraenhansen:
> are fucking kidding me ?

What more specifically are you referring to?

Anzormumladze:
> > are fucking kidding me ?
> 
> Most likely not ...
> What more specifically are you referring to?

we are waiting for this fix for 3 years.......

kraenhansen:
> we are waiting for this fix for 3 years.......

Most likely not, the issues I'm fixing with this PR was introduced with Realm JS v10.0.0, which was released 14th of October, so its been more like a month (which I agree is also a long wait). I suspect you're looking for a fix to an issue that might not have been identified if you've been waiting 3 years - have you created an issue on GitHub that you can link to?

Anzormumladze:
> > we are waiting for this fix for 3 years.......
> 
> Most likely not, since the issues I'm fixing with this PR was introduced with Realm JS v10.0.0, which was released 14th of October. So it has been more like a month (which I agree is also a long wait). I suspect you're looking for a fix to an issue that might not have been identified if you've been waiting 3 years - have you created an issue on GitHub that you can link to?

https://stackoverflow.com/questions/44016885/error-after-installing-realm-must-first-create-rpc-session-with-a-valid-host

kraenhansen:
> https://stackoverflow.com/questions/44016885/error-after-installing-realm-must-first-create-rpc-session-with-a-valid-host

This is a post on stackoverflow and not an issue here on GitHub.
It's referring to Realm JS v1.3.0 which was released more than three years ago. If you're still using v1.3.0, I would highly recommend upgrading to the latest version (v6 for now, if you're heavily relying on the remote chrome debugger).

Anzormumladze:
> > https://stackoverflow.com/questions/44016885/error-after-installing-realm-must-first-create-rpc-session-with-a-valid-host
> 
> This is a post on stackoverflow and not an issue here on GitHub.
> It's referring to Realm JS v1.3.0 which was released more than three years ago. If you're still using v1.3.0, I would highly recommend upgrading to the latest version (v6 for now, if you're heavily relying on the remote chrome debugger).

dude why you arguing with me, this issue exists on every version of realm js, go code something

geragray:
@Anzormumladze I understand your frustration when it comes to debugging with JS SDK. You are right, this is a well know issue (explained in detail here https://github.com/realm/realm-js/issues/491). We are hoping to fix it by investing in https://github.com/realm/realm-js/issues/2455. Please subscribe. 

I would also like to point out that leaving derogatory comments is against our code of conduct https://realm.io/conduct/. Please refrain from using foul language in the future.

kevinrodriguez-io:
@kraenhansen You're doing an amazing work to get this out. Tell me if I can help with something, looking forward to fix this issue for our developer team.

Anzormumladze:
> @Anzormumladze I understand your frustration when it comes to debugging with JS SDK. You are right, this is a well know issue (explained in detail here #491). We are hoping to fix it by investing in #2455. Please subscribe.
> 
> I would also like to point out that leaving derogatory comments is against our code of conduct https://realm.io/conduct/. Please refrain from using foul language in the future.

thanks for answering me back, I deleted all my comment and looking forward to seeing good news

lib/browser/rpc.js
```diff
@@ -366,14 +382,14 @@ function sendRequest(command, data, host = sessionHost) {
                     error = `Unknown callback id: ${callback}`
                 }
             } catch (e) {
-                error = e.message || ('' + e);
+                error = e.message || ("" + e);
```
kneth, line 385:
```suggestion
                error = e.message || (`${e}`);
```

src/js_app.hpp
```diff
@@ -160,18 +171,23 @@ void AppClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments
...
 
-    auto realm_constructor = js::Value<T>::validated_to_object(ctx, js::Object<T>::get_global(ctx, "Realm"));
     std::string user_agent_binding_info;
     auto user_agent_function = js::Object<T>::get_property(ctx, realm_constructor, "_createUserAgentDescription");
     if (js::Value<T>::is_function(ctx, user_agent_function)) {
         auto result = js::Function<T>::call(ctx, js::Value<T>::to_function(ctx, user_agent_function), realm_constructor, 0, nullptr);
         user_agent_binding_info = js::Value<T>::validated_to_string(ctx, result);
     }
+    else {
+        // FIXME: When debugging RN apps, _createUserAgentDescription cannot be found
```
kneth, line 188: We will have to create a callback to the browser to get it to work. On the other hand, while debugging, it is not so important.

src/js_app.hpp
```diff
@@ -160,18 +171,23 @@ void AppClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments
...
+        user_agent_binding_info = "N/A";
```
kneth, line 189:
```suggestion
        user_agent_binding_info = "RN debugging";
```

src/js_app.hpp
```diff
@@ -185,6 +201,12 @@ void AppClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments
         config.platform_version = js::Value<T>::validated_to_string(ctx, Object::get_property(ctx, result_object, platform_version_name));
         config.sdk_version = js::Value<T>::validated_to_string(ctx, Object::get_property(ctx, result_object, sdk_version_name));
     }
+    else {
+        // FIXME: When debugging RN apps, _createPlatformDescription() cannot be found
+        config.platform = "N/A";
```
kneth, line 206:
```suggestion
        config.platform = "RN debugging";
```

src/js_app_credentials.hpp
```diff
@@ -63,6 +63,12 @@ class CredentialsClass : public ClassDefinition<T, realm::app::AppCredentials> {
         {"serverApiKey",     wrap<server_api_key>},
         {"jwt",              wrap<jwt>},
     };
+    
+    static void get_payload(ContextType, ObjectType, ReturnValue &);
+    
+    PropertyMap<T> const properties = {
+        {"payload", {wrap<get_payload>, nullptr}},
```
kneth, line 70: Should it be added to the API doc and TS defs?

kneth (commented):
I think you need to add the new network transport to `Android.mk`.

lib/browser/rpc.js
kraenhansen: This was an automatic change from applying the eslint rules of the project root folder. I would rather not spent too much time on editing this any futher :)

src/js_app.hpp
kraenhansen: That's a great point - I overlooked this.

src/js_app_credentials.hpp
kraenhansen: Great catch. It's [already in the TS defs](https://github.com/realm/realm-js/blob/master/types/app.d.ts#L135) and I'll add it to the API docs.

kraenhansen:
> I think you need to add the new network transport to Android.mk.

If that was the case, I'm wondering why the integration tests are passing?

I think it doesn't have to be added since it's a header and those are taken care of on a pr directory level here: https://github.com/realm/realm-js/blob/master/react-native/android/src/main/jni/Android.mk#L95

src/js_app.hpp
kraenhansen: For context, this was how the previous react native SDK reported its platform:

https://github.com/mongodb/stitch-js-sdk/blob/310f0bd5af80f818cdfbc3caf1ae29ffa8e9c7cf/packages/react-native/core/src/core/auth/internal/StitchAuthImpl.ts#L141-L143

src/rpc.cpp
```diff
@@ -484,7 +527,9 @@ RPCServer::RPCServer() {
         }
 
         JSObjectRef credentials_object = (JSObjectRef)jsc::Function::call(m_context, email_password_method, arg_count, arg_values);
-        return (json){{"result", serialize_json_value(credentials_object)}};
+        
+        auto serialized_result = serialize_json_value(credentials_object);
```
blagoev, line 531: consider `auto&` or reverting this change

src/js_app.hpp
kraenhansen: I understand that this would be the case from a practical level, but if we had a bug in our extension code the `_createUserAgentDescription` method might not be assigned and then I think it's misleading to send "RN debugging".

src/js_app.hpp
kraenhansen: Same as [the other comment ](https://github.com/realm/realm-js/pull/3411#discussion_r529467419). I would suggest setting this to `"js-unknown"` to follow the pattern used by the legacy SDK and at the same time signal that the platform isn't known at this point. It'll most likely be react native chrome debugging, but we're checking for the existence of a function and not actually that we're running in that environment.

lib/browser/rpc.js
```diff
@@ -90,74 +107,74 @@ export function asyncOpenRealm(id, config, callback) {
...
 export function callSyncFunction(name, args) {
     args = (args || []).map((arg) => serialize(null, arg));
-    let result = sendRequest('call_sync_function', { name, arguments: args });
+    let result = sendRequest("call_sync_function", { name, arguments: args });
     return deserialize(undefined, result);
 }
 
 export function callMethod(realmId, id, name, args) {
-    if (args) {
+    if (args && Array.isArray(args)) {
```
blagoev, line 168: this should be a argument validation instead of silencing the error

lib/browser/rpc.js
blagoev: well there is the `commit suggestion` button :)

src/js_app.hpp
```diff
@@ -160,18 +171,23 @@ void AppClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments
         throw std::runtime_error("Expected either a configuration object or an app id string.");
     }
 
-    Protected<typename T::GlobalContext> protected_ctx(Context<T>::get_global_context(ctx));
-    config.transport_generator = [=] {
-        return std::make_unique<NetworkTransport>(protected_ctx);
+    auto realm_constructor = js::Value<T>::validated_to_object(ctx, js::Object<T>::get_global(ctx, "Realm"));
+    
+    Protected<typename T::GlobalContext> protected_ctx(Context::get_global_context(ctx));
+    config.transport_generator = [protected_ctx] {
```
blagoev, line 177: We should be using this syntax everywhere we have to protect the context 
https://github.com/realm/realm-js/blob/master/src/js_types.hpp#L708
this helps with making sure we don't introduce unintentional bugs cause of the two ctx names existing in the same scope

src/js_app.hpp
```diff
@@ -160,18 +171,23 @@ void AppClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments
...
 
-    auto realm_constructor = js::Value<T>::validated_to_object(ctx, js::Object<T>::get_global(ctx, "Realm"));
     std::string user_agent_binding_info;
     auto user_agent_function = js::Object<T>::get_property(ctx, realm_constructor, "_createUserAgentDescription");
     if (js::Value<T>::is_function(ctx, user_agent_function)) {
         auto result = js::Function<T>::call(ctx, js::Value<T>::to_function(ctx, user_agent_function), realm_constructor, 0, nullptr);
         user_agent_binding_info = js::Value<T>::validated_to_string(ctx, result);
     }
+    else {
+        // The `_createUserAgentDescription` method is undefined when chrome debugging a React Native app
```
blagoev, line 188: this function needs to accommodate that it might get executed in RN environment and we should call it here

src/jsc/jsc_rpc_network_transport.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace rpc {
+
+/*
+struct RPCFetchRef {
+    static JSValueRef value;
+};
+*/
+
+/**
+ * Provides an implementation of GenericNetworkTransport for use when the library is loaded in a runtime which doesn't provide the APIs required to perform network requests directly.
```
blagoev, line 36: I don't see the need to implement this. we should be able to use the JSNetworkTransport.

src/js_types.hpp
```diff
@@ -126,6 +126,7 @@ struct Value {
     static bool is_boolean(ContextType, const ValueType &);
     static bool is_constructor(ContextType, const ValueType &);
     static bool is_date(ContextType, const ValueType &);
+    static bool is_error(ContextType, const ValueType &);
```
blagoev, line 129: whenever we change the abstraction layer we need to handle it both for Nodejs and RN

src/jsc/jsc_value.hpp
```diff
@@ -85,6 +85,12 @@ inline bool jsc::Value::is_constructor(JSContextRef ctx, const JSValueRef &value
     return JSValueIsObject(ctx, value) && JSObjectIsConstructor(ctx, (JSObjectRef)value);
 }
 
+template<>
+inline bool jsc::Value::is_error(JSContextRef ctx, const JSValueRef &value) {
```
blagoev, line 89: whenever we change the abstraction layer we need to handle it both for Nodejs and RN

blagoev (changes requested):

src/js_app.hpp
nirinchev: I think it's a bit pessimistic to optimize for the case where we have bugs in our code. We should instead assume that the code works as intended and provide the most relevant UX to users.

kraenhansen: > I think it's a bit pessimistic to optimize for the case where we have bugs in our code.

I disagree.

The suggested change is too specific to the situations that might be causing the function to be undefined. In my opinion, the "right" solution would be to provide an authoritative runtime check which determines is we're running in the chrome debugger mode, but I honestly see very little value in that, since these values likely wont be relevant to the developer through the MongoDB Realm admin dashboard.

src/RealmJS.xcodeproj/project.pbxproj
```diff
@@ -727,6 +728,7 @@
 		F6874A441CAD2ACD00EEEE36 /* JSC */ = {
 			isa = PBXGroup;
 			children = (
+				DC0FCC4C256817E80028429B /* jsc_rpc_network_transport.hpp */,
```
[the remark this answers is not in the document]
kraenhansen: As [I wrote @kneth above](https://github.com/realm/realm-js/pull/3411#issuecomment-732062134), this is a header file and I therefore don't think that's needed.

src/rpc.cpp
```diff
@@ -248,6 +259,23 @@ RPCServer::RPCServer() {
         setIncludesNativeCallStack(m_context, false);
     }
 
+    m_requests["/create_session"] = [this](const json dict) {
+        RJSInitializeInContext(m_context);
```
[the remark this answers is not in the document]
kraenhansen: I honestly don't know, this is simply copying in the implementation that got removed by accident, from https://github.com/realm/realm-js/blob/v6/src/rpc.cpp#L252-L265.

src/jsc/jsc_value.hpp
kraenhansen: Thanks for catching this - I thought about it and forgot about it :)

src/js_app.hpp
```diff
@@ -36,12 +36,16 @@ using SharedUser = std::shared_ptr<realm::SyncUser>;
...
+template<typename T>
+using NetworkTransportFactory = std::function<std::unique_ptr<app::GenericNetworkTransport>(typename T::Context)>;
+
 template<typename T>
 class AppClass : public ClassDefinition<T, SharedApp> {
     using ContextType = typename T::Context;
     using FunctionType = typename T::Function;
     using ObjectType = typename T::Object;
     using ValueType = typename T::Value;
+    using Context = js::Context<T>;
```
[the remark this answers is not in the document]
kraenhansen: It's used here https://github.com/realm/realm-js/blob/kh/rn-create-session-continued/src/js_app.hpp#L176

src/rpc.cpp
kraenhansen: Is this to prevent a (potentially expensive) move?

kraenhansen: I'll revert, this was added while debugging.

src/js_app.hpp
kraenhansen: I don't understand this comment ☝️  Can I get you to rephrase it?

src/jsc/jsc_rpc_network_transport.hpp
kraenhansen: Can you provide an explanation to this? What makes you think it would be better to use the `JSNetworkTransport`? That is basically just a wrapper around some JavaScript code which won't be loaded in the case of us running in chrome debugger mode. I see no advantage in trying to polyfill the `_networkTransport` property from C++.

lib/browser/rpc.js
kneth: Just leave as it is - `"" + e` is just so much last year :smile:

src/js_app.hpp
kneth: In the past, a sync session should be opened without setting `user_agent_binding_info` so we silently skipped the call when the RN debugger was active: https://github.com/realm/realm-js/blob/v6/src/js_sync.hpp#L62-L65

src/js_app.hpp
nirinchev: I agree that an unambiguous check if the code is running in a debugger is better, but don't agree with the rest. I don't think a user will easily make the connection that "js-unknown" means a debugger, so is likely to be left with the impression that either our code or theirs is broken and doesn't correctly report the platform.

If this function is only unrefined in a debugger, then using that as a check, even if semantically incorrect, is better than providing a broken user experience.

kneth: In pre-v10 we skipped setting it as object store didn't require it. With v10, it is now a requirement. You can say that we are doing more or less the same as before.

kneth (approved):

src/RealmJS.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
blagoev: oh its a `hpp`. its alright then

lib/browser/rpc.js
blagoev: seriously, why not click the `commit suggestion` button

src/rpc.cpp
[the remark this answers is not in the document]
blagoev: I see. I was comparing against the previous version. If this was removed in v10 then it was a big accident indeed

lib/browser/rpc.js
```diff
@@ -334,20 +358,12 @@ function sendRequest(command, data, host = sessionHost) {
...
-                    // Remove the type prefix from the error message (e.g. "Error: ").
-                    responeMessage = response.message.replace(/^[a-z]+: /i, '');
-                }
-
-                const exceptionToReport = new Error(responeMessage);
-                Object.assign(exceptionToReport, responseError);
-                throw exceptionToReport;
+            if (error && typeof error === "string") {
+                error = error.replace(/^[a-z]+: /i, "");
+            } else if (error.type && error.type === "error") {
```
blagoev, line 363: Just to confirm we are sure we don't have this case anymore?
`error.type && error.type === 'dict'`

src/jsc/jsc_rpc_network_transport.hpp
blagoev: The way I see it is we have JavaScriptNetworkTransport which is a generic class to do network requests. It should be possible to use it instead of defining a new one for the debug context. The way it works is it just needs a `global.Realm._networkTransport.fetchWithCallbacks` in order to work.

kraenhansen: That was my initial thought as well. However it turned out to be way easier to implement an alternative specialisation of `GenericNetworkTransport` than providing a polyfill for `fetchWithCallbacks` implemented in C++. I'm sure you would have reached the same conclusion. From your comment, I don't see sufficient value in changing this approach and I'll leave it at this.

lib/browser/rpc.js
kraenhansen: We should definitely refrain from throwing anything other than instances of `Error` (or a subclass thereof).
If we do,  that will proceed to line 369 - so an error will be thrown client side, although that will probably be with a strange message:

    Error: [object Object]

blagoev: aren't we throwing an instance of Error in the old implementation. Seems we are
https://github.com/realm/realm-js/pull/3411/files#diff-1aeca8a83cdc5d4f7bf6859366830e44c8acbd0269a540bef9194be798436902L348
I wonder if the case with this check is still valid
`error.type && error.type === 'dict'`

kraenhansen: > aren't we throwing an instance of Error in the old implementation.

We are, but that would be serialized to a 'dict' which deserialized to a regular object on the client side, which would then be wrapped in an Error.

> We should definitely refrain from throwing anything other than instances of Error (or a subclass thereof).

This comment was to the native code. It's the only way I can see the `error.type` would be "dict" after this PR.

src/rpc.cpp
blagoev: ok resolving

src/js_app.hpp
blagoev: I see this is fixed. resolving

lib/browser/rpc.js
blagoev: ok. resolving

blagoev (approved):
approving the changes since my comments are already worked on

src/js_app.hpp
Muzzamil75: keep up the good work guys. Anxiously waiting for this bug fix ! @kneth  @kraenhansen

cesarvr (approved):
LGTM

kraenhansen:
I've added static members to the `AppClass` and exposed a `_setVersions` static method on it to set it from the outside as well as over the RPC bridge. I considered calling the `set_versions` method from `rpc.cpp`, but it turned out more complicated than setting the static members on the `AppClass` directly.

kneth (approved):

kraenhansen:
I found (and solved a bug) where fetching would cause a "EXC_BAD_ACCESS" (the function got called in a released context).

src/rpc.cpp
```diff
@@ -624,6 +624,9 @@ RPCServer::~RPCServer() {
     m_objects.clear();
     m_callbacks.clear();
 
+    // Clear the Object Store App cache, to prevent instances from using the context which is going to be released.
+    app::App::clear_cached_apps();
```
kraenhansen, line 628: I am a bit unsure if this is the right place to call it.
I'm unaware of another place in our code that gets called when we're "unloading" our library, but we could clear the object store app cache when we initialize instead? https://github.com/realm/realm-js/blob/master/src/jsc/jsc_init.cpp#L42

kraenhansen: @blagoev @kneth ☝️ what do you think about that?

kneth: I imagine the issue is when you reload the app, right? If so, I think it make sense to clear the cache as you shut down the RPC server.

kraenhansen: Correct. If it's not cleared before / after a reload object store might be reusing app instances that reference a context which has been garbage collected.

kneth: :+1:

kneth:
LGTM - merge at will!

cesarvr (approved):

fronck (commented):
In general:  I think we should make an effort to document functions -- at least new ones -- with e.g., Doxygen-style headers to improve the general developer-friendlyness of the codebase.


Otherwise, good job; LGTM.

mendesbarreto:
Thank you all for this work!!!

saivarunk:
When can we expect this commit/version to be published on npm?

kraenhansen:
@saivarunk it's out in [v10.0.2](https://github.com/realm/realm-js/releases/tag/v10.0.2).

PR #3416, opened 2020-11-20
Fix a bug where shouldCompactOnLaunch would lead to a crash 

kneth:
# What, How & Why?

The `shouldCompactOnLaunch` is specified for synced Realm, the callback has to be dispatched on the correct thread. If not, a fatal error can happen.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


fronck (approved):
LGTM

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
     friend class RealmClass<T>;
 };
 
+
+template<typename T>
+class ShouldCompactOnLaunchFunctor {
```
blagoev, line 218: We could refactor this in a generic functor class that enables us to use it in anywhere we need to invoke the callback on main thread and wait for the result. This implementation is too specific and next time we need it we will need to do it again.

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
...
+            should_compact_on_launch= m_should_compact_on_launch;
+        }
+
+        return should_compact_on_launch;
+    }
+
+    static void main_loop_handler(ShouldCompactOnLaunchFunctor<T>* this_object,
+                                  const uint64_t total_bytes,
+                                  const uint64_t used_bytes) {
+        const Protected<typename T::GlobalContext>& ctx = this_object->m_ctx;
```
blagoev, line 249: not that it matters much but the context should be already protected here

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
...
+        HANDLESCOPE(ctx);
+
+        const int argc = 2;
+        typename T::Value arguments[argc] = { Value<T>::from_number(ctx, total_bytes), Value<T>::from_number(ctx, used_bytes) };
+        typename T::Value ret_val = Function<T>::callback(ctx, this_object->m_func, typename T::Object(), argc, arguments);
+        bool ret_val_bool = Value<T>::to_boolean(ctx, ret_val);
```
blagoev, line 255: should be `validated_to_boolean` imo

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
...
+        this_object->m_cond_var->notify_one();
+    };
+
+private:
+    const Protected<typename T::GlobalContext> m_ctx;
+    const Protected<typename T::Function> m_func;
+    util::EventLoopDispatcher<void(ShouldCompactOnLaunchFunctor<T>* this_object,
+                                   uint64_t total_bytes,
+                                   uint64_t used_bytes)> m_event_loop_dispatcher;
+    bool m_should_compact_on_launch_callback_done = false;
```
blagoev, line 272: this should not be needed if we simplify the logic. the meaning of `m_should_compact_on_launch_callback_done`  is actually  set when the sync primitive is signaled

blagoev (changes requested):

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Fixed
 * Fixed an issue in `toJSON()`, in combination with primaryKeys, where data from another table could be returned. ([#3331](https://github.com/realm/realm-js/issues/3331), since v10.0.0)
 * Fixed an issue in `toJSON()` where `data` would output as `{}`, it now returns the data base64 encoded. ([#3356](https://github.com/realm/realm-js/pull/3356), since v10.0.0)
+* Fixed a bug where setting `shouldCompactOnLaunch` would lead to a crash with the error `Fatal error in HandleScope::HandleScope` (on node.js).
```
bmunkholm, line 9: Issue reference? Since?

kneth: We only have a HELP ticket in Jira. Since: either v10.0.0 or v1.0.0 (when sync was introduced).

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
     friend class RealmClass<T>;
 };
 
+
+template<typename T>
+class ShouldCompactOnLaunchFunctor {
+public:
+    ShouldCompactOnLaunchFunctor(typename T::Context ctx, typename T::Function should_compact_on_launch_func)
+    : m_ctx(Context<T>::get_global_context(ctx))
```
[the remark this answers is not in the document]
kneth: `m_ctx` has type `Protected<typename T::GlobalContext>`.

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
...
+    , m_func(ctx, should_compact_on_launch_func)
+    , m_event_loop_dispatcher {ShouldCompactOnLaunchFunctor<T>::main_loop_handler}
+    , m_mutex{new std::mutex}
+    , m_cond_var{new std::condition_variable}
+    {
+    }
+
+    bool operator ()(const uint64_t total_bytes, const uint64_t used_bytes) {
```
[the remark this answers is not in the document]
kneth: The function which is dispatched to the event loop in `EventLoopDispatcher` cannot return a value (the signature is `std::function<void(Args...)>`), and we use a shared variable between the threads.

cesarvr (approved):

src/js_realm.hpp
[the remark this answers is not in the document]
blagoev: resolving this

src/js_realm.hpp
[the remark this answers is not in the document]
blagoev: `main_loop_handler` can set a variable with the result and unlock the mutex. There is not need of two variables since there are only two threads involved and one of them is main which will be waiting for the result. this means there will be no chance of spurious wakeup. but you  can decide to keep the `m_should_compact_on_launch_callback_done` if you preferer.

so the pseudo code should be 
main thread
```
m_event_loop_dispatcher()
{
    std::unique_lock<std::mutex> lock(*m_mutex);
    m_cond_var->wait()
    return m_should_compact_on_launch;
}
```

and the loop_handler should be
```
Function::call
m_should_compact_on_launch = Value<T>::validated_to_boolean(ctx, ret_val);
this_object->m_should_compact_on_launch_callback_done = true;
this_object->m_cond_var->notify_one();
```

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
...
 
+
+template<typename T>
+class ShouldCompactOnLaunchFunctor {
+public:
+    ShouldCompactOnLaunchFunctor(typename T::Context ctx, typename T::Function should_compact_on_launch_func)
+    : m_ctx(Context<T>::get_global_context(ctx))
+    , m_func(ctx, should_compact_on_launch_func)
+    , m_event_loop_dispatcher {ShouldCompactOnLaunchFunctor<T>::main_loop_handler}
+    , m_mutex{new std::mutex}
```
blagoev, line 224: Could remove the shared pointer in this context. 
It means only main is responsible for destructing the mutex which is fine since main will wait for the result first and only then will complete

blagoev: So just declare these members on the class and they will be initialized for you

src/js_realm.hpp
```diff
@@ -213,6 +213,68 @@ class RealmDelegate : public BindingContext {
...
+    }
+
+    bool operator ()(const uint64_t total_bytes, const uint64_t used_bytes) {
+        {
+            std::lock_guard<std::mutex> lock {*m_mutex};
+            m_should_compact_on_launch_callback_done = false;
+        }
+
+        m_event_loop_dispatcher(this, total_bytes, used_bytes);
+        bool should_compact_on_launch = false;
```
blagoev, line 236: this local variable is not needed you could remove it to clean up the code

CHANGELOG.md
bmunkholm: Then list v1.0.0

src/js_realm.hpp
blagoev: these are required actually so the Functor can be moved.

src/js_realm.hpp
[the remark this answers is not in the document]
blagoev: we agreed to commit as is and later refactor this class

src/js_realm.hpp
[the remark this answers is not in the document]
blagoev: this was handled

src/js_realm.hpp
blagoev: this was handled

PR #3420, opened 2020-11-24
Making getApp return an App instead of Realm.App

kraenhansen:
## What, How & Why?

A fix to a slight inconvenience I discovered while reproducing a different bug.
Basically the `App.getApp` static method was returning instances of type `Realm.App` which resulted in the `app.currentUser` type being incompatible with the package's concrete `User` type (which extends the `Realm.User` type).

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web-integration-tests/src/app.test.ts
+++ b/packages/realm-web-integration-tests/src/app.test.ts
@@ -25,13 +25,23 @@ import { createApp } from "./utils";
 const environment = getEnvironment();
 
 describe("App#constructor", () => {
+    afterEach(() => {
+        environment.defaultStorage.clear();
+    });
+
     it("constructs", () => {
         const app = new App("default-app-id");
         expect(app).to.be.instanceOf(App);
     });
 
-    afterEach(() => {
-        environment.defaultStorage.clear();
+    it("construct singletons", () => {
+        const app1 = App.getApp("default-app-id");
+        const app2 = App.getApp("default-app-id");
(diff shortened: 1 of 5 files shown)


packages/realm-web/src/App.ts
```diff
@@ -62,7 +62,7 @@ export class App<
     /**
      * A map of app instances returned from calling getApp.
      */
-    private static appCache: { [id: string]: Realm.App } = {};
```
kraenhansen, line 65: This was the fix.

packages/realm-web-integration-tests/test/mocha.opts
```diff
@@ -1,2 +1,2 @@
 --watch-extensions ts
---require ts-node/register/transpile-only
+--require ts-node/register
```
kraenhansen, line 2: I switched to this, to get type errors when running the tests.

packages/realm-web-integration-tests/src/app.test.ts
```diff
@@ -25,13 +25,23 @@ import { createApp } from "./utils";
 const environment = getEnvironment();
 
 describe("App#constructor", () => {
+    afterEach(() => {
+        environment.defaultStorage.clear();
+    });
```
kraenhansen, line 30: I just moved this up a bit.

steffenagger (approved):

PR #3434, opened 2020-12-01
{Monorepo} Fixes the release build on Windows

blagoev:
fix the windows release build. 
The MSVC compiler is optimizing the initialization of static members of template class out of sequence which causes access violations on startup. This change makes sure all static members are initialized before they are used.

### DIFF
--- a/src/node/node_class.hpp
+++ b/src/node/node_class.hpp
@@ -195,8 +195,6 @@ class ObjectWrap {
 	static bool has_native_method(const std::string& name);
 
   private:
-    //These wrapper methods exists to force the compiler to initialize these class template static members in sequence before they are used
-	//https://stackoverflow.com/questions/42743048/c-template-the-static-member-in-a-global-object-is-not-initialized (https://stackoverflow.com/a/42743520)
 	static ClassType& get_class();
 	static auto& get_nativeMethods();
 	static auto& get_schemaObjectTypes();
@@ -267,7 +265,6 @@ static inline std::vector<napi_value> napi_get_arguments(const Napi::CallbackInf
 	return arguments;
 }
 
-//This forces initialization of the static member in sequence
 template<typename ClassType>
 inline ClassType& ObjectWrap<ClassType>::get_class() {
 	static ClassType s_class;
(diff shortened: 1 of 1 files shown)


src/node/node_class.hpp
```diff
@@ -266,15 +267,25 @@ static inline std::vector<napi_value> napi_get_arguments(const Napi::CallbackInf
 	return arguments;
 }
 
-// The static class variable must be defined as well.
+//This forces initialization of the static member in sequence
```
RedBeard0531, line 270: Pedantically, the technique you are using does the opposite of what the comment implies. It *avoids* forcing the members to be initialized (at the normal point), and instead defers initialization until the first time any of these functions are called.

I'd just delete the comment since I don't think it adds much. Really, you should probably comment when you *don't* use this technique to explain why it is OK, rather than commenting when you do 🙂

src/node/node_class.hpp
```diff
@@ -1086,6 +1101,8 @@ inline std::vector<Napi::PropertyDescriptor> ObjectWrap<ClassType>::create_napi_
 template<typename ClassType>
 Napi::Object ObjectWrap<ClassType>::create_instance_by_schema(Napi::Env env, Napi::Function& constructor, const realm::ObjectSchema& schema, Internal* internal) {
 	Napi::EscapableHandleScope scope(env);
+	auto& s_schemaObjectTypes = get_schemaObjectTypes();
+	ClassType& s_class = get_class();
```
RedBeard0531, line 1105: Not a big deal, but it seems odd to sometimes use `auto` for one of these and the actual type for the other. Maybe just always use auto?

src/node/node_class.hpp
```diff
@@ -195,7 +195,11 @@ class ObjectWrap {
 	static bool has_native_method(const std::string& name);
 
   private:
-    static ClassType s_class;
+    //These wrapper methods exists to force the compiler to initialize these class template static members in sequence before they are used
```
RedBeard0531, line 198: This comment isn't correctly aligned, and it has the same issue I describe below. I'm also not sure it is helpful.

RedBeard0531 (approved):
Seems like the correct fix to me. But at least fix the comment indentation before pushing.

src/node/node_class.hpp
blagoev: thanks. It guess it forces the compiler rather, to defer the initialization. But maybe this comment is confusing. will remove it

src/node/node_class.hpp
blagoev: will fix

src/node/node_class.hpp
blagoev: will remove it

Issue #2799, opened 2020-04-07
Fix Realm's React Native debugger infrastructure to support class syntax


blagoev:
- Realm JS SDK Version: 5.0
- Which debugger for React Native: Chrome

mariasiuvlad:
Any updates on this?

anotherleon:
it still happens on 6.1.2

PR #3447, opened 2020-12-08
Update RealmReact.mm

OwenRay:
## What, How & Why?
Old session where not properly disposed of when reloading the app, that's why we can clear the old sessions just before starting the rpc service (because the rpc service _is_ restarted)

This closes realm/realm-js#2799 ???

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/react-native/ios/RealmReact/RealmReact.mm
+++ b/react-native/ios/RealmReact/RealmReact.mm
@@ -214,6 +214,7 @@ RCT_REMAP_METHOD(emit, emitEvent:(NSString *)eventName withObject:(id)object) {
 }
 
 - (void)startRPC {
+    realm::_impl::RealmCoordinator::clear_all_caches();
     [GCDWebServer setLogLevel:3];
     _webServer = [[GCDWebServer alloc] init];
     _rpcServer = std::make_unique<RPCServer>();


kneth:
@OwenRay 

kraenhansen:
When working on #3411 I had to clear the Object Store app cache: I choose to put it on the RPCServer shutdown / destructor.
See https://github.com/realm/realm-js/pull/3411/files#r535234990.

Do you see a need for this to be specifically on startup?
If not, I think it would be better to move it to the RPCServers destructor, since:
- It impact the startup time negatively (I know it's very small work - but it is taking a lock)
- It would be aligned with the PR I just merged 🙂

kraenhansen:
Going over this again I realized that we are in-fact calling the `clear_all_caches` before instantiating Realm JS into a new context: https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmReact.mm#L290-L291

I'm left wondering why we would need this addition when we're already calling this elsewhere.

Did you verify that this does in-fact solve #2799 or another issue?

kraenhansen:
@OwenRay have you had a chance to look into my comment above?

PR #3449, opened 2020-12-08
Update to Realm Sync v10.1.4

kneth:


CHANGELOG.md
```diff
@@ -5,14 +5,16 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * Fixing the construction of Realm instances declararing the schema, when running in the React Native Chrome debugging mode, by removing it. Note: This is not considered a breaking change, since this behaviour was never documented. ([#3442](https://github.com/realm/realm-js/pull/3442), since v10.0.0)
+* Fixed a bug that would prevent eventual consistency during conflict resolution. Affected clients would experience data divergence and potentially consistency errors as a result if they experienced conflict resolution between cycles of Create-Erase-Create for objects with primary keys.
 
 ### Compatibility
-* MongoDB Realm Cloud.
+* MongoDB Realm Cloud version 20201202 or later.
```
bmunkholm, line 11: true. But no one externally has a chance to affect which version they use. But perhaps for internal recording it's good to have?

bmunkholm: So maybe we can leave that to the internal note you have?

bmunkholm (approved):

scripts/changelog-header.sh
```diff
@@ -14,7 +14,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Compatibility
-* MongoDB Realm Cloud.
+* MongoDB Realm Cloud version 20201202 or later.
```
bmunkholm, line 17: same here - we can leave this out as it's not something users care about.

Issue #3454, opened 2020-12-09
Support BSON types when debugging using the chrome debugger


kraenhansen:
In the current state, creating an object with a property (_id in the error message below) of type BSON ObjectId will fail when the app is running in Chrome debugging mode, with the following error message:

> ClassName._id must be of type 'objectId', got 'object' ([object Object])

The current implementation is depending on a `_bson`, `_Decimal128` and `_ObjectId` constants being patched onto the `Realm` constructor in a couple of places:
- [js_types.hpp / create_bson_type](https://github.com/realm/realm-js/blob/master/src/js_types.hpp#L540-L546)
- [js_types.hpp / to_bson](https://github.com/realm/realm-js/blob/master/src/js_types.hpp#L693-L705)
- [jsc_value.cpp / from_decimal128](https://github.com/realm/realm-js/blob/master/src/jsc/jsc_value.cpp#L68-L86)
- [jsc_value.cpp / from_object_id](https://github.com/realm/realm-js/blob/master/src/jsc/jsc_value.cpp#L89-L100)

When reading a property of type ObjectId, users will also experience:

> Error: _ObjectId must be of type 'constructor', got (undefined)

This constant is never set when running in chrome debugging mode.

Issue #3206, opened 2020-09-10
Realm in Debug Mode kills React Native on iOS and Android


Shumuu:
Using Chrome Debugger and using sorted or filtered on Realm Collection kills the Application

## Goals
Using the Chrome Debugger with Realm and sorting/filtering Realm Objects

## Expected Results
App not crashing ...

## Actual Results
App crashes as soon as the Debugger connects and the filtered or sorted action happens.

```
2020-09-10 11:43:16.821 297-5036/? E/android.hardware.audio@6.0-impl.ranchu: WriteThread::doWriteImpl:226: pcm_write failed with Operation not permitted
2020-09-10 11:43:16.834 9701-9765/com.realmtest E/unknown:ReactContextBaseJavaModule: Unhandled SoftException
    java.lang.RuntimeException: Catalyst Instance has already disappeared: requested by WebSocketModule
        at com.facebook.react.bridge.ReactContextBaseJavaModule.getReactApplicationContextIfActiveOrWarn(ReactContextBaseJavaModule.java:67)
        at com.facebook.react.modules.websocket.WebSocketModule.sendEvent(WebSocketModule.java:62)
        at com.facebook.react.modules.websocket.WebSocketModule.access$100(WebSocketModule.java:40)
        at com.facebook.react.modules.websocket.WebSocketModule$1.onMessage(WebSocketModule.java:190)
        at okhttp3.internal.ws.RealWebSocket.onReadMessage(RealWebSocket.java:323)
        at okhttp3.internal.ws.WebSocketReader.readMessageFrame(WebSocketReader.java:219)
        at okhttp3.internal.ws.WebSocketReader.processNextFrame(WebSocketReader.java:105)
        at okhttp3.internal.ws.RealWebSocket.loopReader(RealWebSocket.java:274)
        at okhttp3.internal.ws.RealWebSocket$2.onResponse(RealWebSocket.java:214)
        at okhttp3.RealCall$AsyncCall.execute(RealCall.java:206)
        at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:923)
[...]

PR #3456, opened 2020-12-09
Fix debugging synced realms (which uses BSON types)

### LINKED ISSUE #3452: Syncing realm whist debugging throws XMLHttpRequest failure
## Goals
To be able to debug a react native application using realm sync application using chrome debugger using android simulator

## Expected Results

The remote data under the specified partition should be pulled to the client application

## Actual Results

When opening a realm with the following config

```
{"sync":{"user":{},"partitionValue":"opUnit=7523&tourDate=2020-12-07&tourName=MA44","existingRealmFileBehavior":{"type":"openImmediately"}}}
```
with the debugger attached. The app seams to function for a while until the listener that is bound like below is invoked

```
Realm.open(config).then(projectRealm => {
      realmRef.current = projectRealm;
      const syncParcels = projectRealm.objects('Parcel');
      syncParcels.addListener(() => {
        setData({ parcels: [...syncParcels], syncComplete: true });
      });
    });

```
then in the chrome console the following error is thrown

```
DOMException: Failed to execute 'send' on 'XMLHttpRequest': Failed to load 'http://localhost:8083/get_property'.
    at makeRequest (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:113621:18)
    at sendRequest (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:113675:22)
    at Object.getProperty (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:113460:18)
    at Object.getProperty (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:113182:33)
    at Object.get (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:113022:21)
    at Object.value (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:115458:29)
    at Function.from (<anonymous>)
    at _iterableToArray (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:2858:88)
    at _toConsumableArray (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:2842:38)
    at Proxy.eval (http://10.0.2.2:8081/src/providers/tour-parcels-provider.bundle?platform=android&dev=true&minify=false&modulesOnly=true&runModule=false&shallow=true:61:54)
```

The data that is trying to be sent to the rpc service when the error is thrown is:

```
{"realmId":130823838580256,"id":89,"name":"0","sessionId":1}
```

There is one record in atlas under the partitonValue `opUnit=7523&tourDate=2020-12-07&tourName=MA44` that trying to be synced to the react-native app.

## Steps to Reproduce

You should be able to replicate this issue on your sample app https://github.com/mongodb-university/realm-tutorial-react-native as our app follows a similar pattern.

- Realm JS SDK Version: `10.1.0`
- Node or React Native:  react-native version `0.63.3`
- Client OS & Version: Android 29 Simulator
- Which debugger for React Native: Both react native debugger and chrome debugger yield the same result.

Thanks

Andy

kraenhansen:
## What, How & Why?

This closes #3452  and #3454 by serializing BSON types and allowing EJSON to be used for property values when creating objects and assigning values (on React Native only). We might want to consider making this a more general feature to avoid diverging implementations on Node.js and React Native and allow for easy object creation from EJSON (useful when dumping data and restoring it via EJSON serialization).

## ☑️ ToDos
* [x] Extend the RPC server to send ObjectId and Decimal128 values as EJSON.
* [x] Extend the RPC client to deserialize ObjectId and Decimal128 values using the "bson" package.
* [x] Extend the RPC client to serialize ObjectId and Decimal128 values when calling functions (create) and setting values.
* [x] Extend the RPC server to passthrough EJSON values, when deserializing.
* [x] Change the `jsc::Value::is_decimal128` to look for a `$numberDecimal` property as an alternative.
* [x] Change the `jsc::Value::is_object_id` to look for a `$oid` property as an alternative.
* [x] Change the `jsc::Value::to_decimal128` to use the `$numberDecimal` property as an alternative.
* [x] Change the `jsc::Value::to_object_id` to use the `$oid` property as an alternative.
* [ ] Investigate the impact and an alternative solution to the missing `_bson` constant for the `Object<T>::create_bson_type` and `Value<T>::to_bson`
* [x] Ensure the `get_property` RPC command works, when the cache is bypassed and the property is accessed directly. I think it wouldn't, since getting a property using the regular accessors depends on the missing `_ObjectId` and `_Decimal128` properties on the `Realm` constructor.
* [x] 📝 Changelog entry
* [ ] 🚦 Tests: Only manually and locally and the chrome debugging on Android is broken by [a seemingly unrelated issue](https://github.com/realm/realm-js/issues/3206).

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* Fixing creating objects and assigning properties of BSON Decimal128 and ObjectId types, when running in React Native Chrome debugging mode. ([#3452](https://github.com/realm/realm-js/issues/3452) & [#3454](https://github.com/realm/realm-js/issues/3454), since v10.0.0)
-* Fixed a crash that would happen if the app did a network request after the app got refreshed during development and the Chrome debugging mode was disabled. ([#3457](https://github.com/realm/realm-js/issues/3457), since v10.0.2)
+* Fixed creating objects and assigning properties of BSON Decimal128 and ObjectId types, when running in React Native Chrome debugging mode. ([#3452](https://github.com/realm/realm-js/issues/3452) & [#3454](https://github.com/realm/realm-js/issues/3454), since v10.0.0)
+* Fixed a crash that would happen if the app did a network request after the app got refreshed during development and the Chrome debugging mode was disabled. NOTE: Because of [#3206](https://github.com/realm/realm-js/issues/3206) the fix has not been implemented on Android. ([#3457](https://github.com/realm/realm-js/issues/3457), since v10.0.2)
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
 * Fixing creating objects and assigning properties of BSON Decimal128 and ObjectId types, when running in React Native Chrome debugging mode. ([#3452](https://github.com/realm/realm-js/issues/3452) & [#3454](https://github.com/realm/realm-js/issues/3454), since v10.0.0)
+* Fixed a crash that would happen if the app did a network request after the app got refreshed during development and the Chrome debugging mode was disabled. ([#3457](https://github.com/realm/realm-js/issues/3457), since v10.0.2)
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* Fixing creating objects and assigning properties of BSON Decimal128 and ObjectId types, when running in React Native Chrome debugging mode. ([#3452](https://github.com/realm/realm-js/issues/3452) & [#3454](https://github.com/realm/realm-js/issues/3454), since v10.0.0)
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/src/jsc/jsc_value.hpp
+++ b/src/jsc/jsc_value.hpp
@@ -141,7 +141,7 @@ inline bool is_bson_type(JSContextRef ctx, const JSValueRef &value, std::string 
         throw jsc::Exception(ctx, error);
     }
 
-    JSValueRef bsonType = JSObjectGetProperty(ctx, object, JSStringCreateWithUTF8CString("_bsontype"), &error);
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* Fixing creating objects and assigning properties of BSON Decimal128 and ObjectId types, when running in React Native Chrome debugging mode. ([#3452](https://github.com/realm/realm-js/issues/3452) & [#3454](https://github.com/realm/realm-js/issues/3454), since v10.0.0)
```
fronck, line 8: *fixed

src/jsc/jsc_value.hpp
```diff
@@ -150,25 +150,47 @@ inline bool is_bson_type(JSContextRef ctx, const JSValueRef &value, std::string
...
+        throw jsc::Exception(ctx, error);
+    }
+
+    return JSValueIsUndefined(ctx, property) == false;
 }
 
 template<>
 inline bool jsc::Value::is_decimal128(JSContextRef ctx, const JSValueRef &value) {
-    return is_bson_type(ctx, value, "Decimal128");
+    return is_bson_type(ctx, value, "Decimal128") || is_ejson_type(ctx, value, "$numberDecimal");
```
fronck, line 188: This approach will execute the guards in ln. 167-175 / 133+142 twice on EJSON types, which seems unnecessary.  Is there a way to avoid this?
(Also, to reduce code duplications, consider if it would be practical to move this guards into a separate function.

src/jsc/jsc_value.hpp
```diff
@@ -150,25 +150,47 @@ inline bool is_bson_type(JSContextRef ctx, const JSValueRef &value, std::string
...
 }
 
 template<>
 inline bool jsc::Value::is_object_id(JSContextRef ctx, const JSValueRef &value) {
-    return is_bson_type(ctx, value, "ObjectID");
+    return is_bson_type(ctx, value, "ObjectID") || is_ejson_type(ctx, value, "$oid");
```
fronck, line 193: Same comment as above.

fronck (approved):
A few comments to consider, otherwise LGTM.

Good work.

src/rpc.cpp
```diff
@@ -236,6 +237,20 @@ static json read_object_properties(Object& object) {
                 }
                 break;
             }
+            case PropertyType::Decimal: {
```
kneth, line 240: `{` not needed.

src/rpc.cpp
```diff
@@ -236,6 +237,20 @@ static json read_object_properties(Object& object) {
...
+                cache[property.name] = {
+                    {"type", RealmObjectTypesEJSON},
+                    {"value", {"$numberDecimal", obj.get<Decimal>(property.column_key).to_string()}},
+                };
+                break;
+            }
+            case PropertyType::ObjectId: {
```
kneth, line 247: `{` not needed

kneth (approved):
Only minor comments.

src/jsc/jsc_value.hpp
kraenhansen: I agree this is not efficient.
Since I'm pretty much just building on the existing pattern, I would like to keep this as is and create an issue to track this for the future.

PR #3466, opened 2020-12-10
Cmake RN Android

blagoev:
Cmake changes for Realm JS RN Android

The mixed type changes are disabled for RN. 
There is a `//FIXME: //MIXED:` comment on every place where RN code is switched.

This builds the android binaries ahead of time and packages them in the npm package


scripts/prepublish.js
```diff
@@ -26,11 +26,11 @@ const dependencies = ini(fs.readFileSync(path.resolve(__dirname, '../dependencie
 console.log(`Core version: ${dependencies.REALM_CORE_VERSION}`);
 console.log(`Sync version: ${dependencies.REALM_SYNC_VERSION}`);
 
-if ('REALM_BUILD_ANDROID' in process.env) {
+if ('REALM_BUILD_ANDROID_PACKAGE' in process.env) {
```
kraenhansen, line 29: For a long time I've been contemplating replacing this with a per package config: https://docs.npmjs.com/cli/v6/using-npm/config#per-package-config-settings

I wonder if now is a good time/place, if we're renaming this anyway?

package.json
```diff
@@ -61,23 +61,13 @@
...
     "isWin": "node -p \"if (process.platform == 'win32') { process.exit(0); } else { process.exit(-1); }\"",
-    "prealias:win:apply": "git config alias.apply-win-symlink-aliases \"!f() { $(scripts/git-win-symlink-aliases); }; f\"",
-    "alias:win:apply": "git apply-win-symlink-aliases",
-    "postalias:win:apply": "git config --unset alias.apply-win-symlink-aliases",
-    "alias:win:revert": "git config alias.rm-symlink \"\" && git config alias.rm-symlinks \"\" && git config alias.checkout-symlinks \"\"",
-    "win-fix-symlinks": "npm run isWin -s && npm run alias:win:apply && git checkout-symlinks && git rm-symlinks || echo . ",
-    "win-revert-symlinks": "npm run isWin -s &&  npm run alias:win:apply && git checkout-symlinks && npm run alias:win:revert || echo .",
-    "prereact-tests-android": "npm run isWin -s && npm run win-fix-symlinks || echo . ",
-    "react-tests-android": "node scripts/react-tests-android.js || npm run postreact-tests-android",
-    "postreact-tests-android": "npm run win-revert-symlinks",
```
kraenhansen, line 80: Great to see these go! 🎉  👍

kraenhansen:
I see some removal of Realm Object Server related stuff too (and I like it 😄 )

blagoev:
@kraenhansen  let's review when this is ready

scripts/prepublish.js
blagoev: I don't want to diverge too much. I am fixing some stuff I see is outdated or not clear. This name was conflicting with the new concept of building for android that we do now before publishing hence I renamed it. 
But while I am doing these changes I see a lot of things that need to be improved. Probably this one as well

package.json
blagoev: no body uses these currently.

Dockerfile
```diff
@@ -21,10 +21,7 @@ RUN yum -y install \
 RUN mkdir -p $NVM_DIR \
  && curl -s https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | bash \
  && . $NVM_DIR/nvm.sh \
- && nvm install 10 \
  && nvm install 12 \
```
kneth, line 24: Version 14 is LTS now - should we use that version instead?

Jenkinsfile
```diff
@@ -389,17 +388,16 @@ def buildWindows(nodeVersion, arch) {
 
 def inAndroidContainer(workerFunction) {
   return {
-    myNode('android') {
+    myNode('docker-cph-03') {
```
kneth, line 391: Why can we only use `docker-cph-03`?

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
kneth, line 3: `2016` -> `2021`

src/node/node_class.hpp
```diff
@@ -266,7 +266,7 @@ static inline std::vector<napi_value> napi_get_arguments(const Napi::CallbackInf
 }
 
 template<typename ClassType>
-inline ClassType& ObjectWrap<ClassType>::get_class() {
+inline auto& ObjectWrap<ClassType>::get_class() {
```
kneth, line 269: I am a bit curious: why change to `auto`?

tests/react-test-app/android/app/build.gradle
```diff
@@ -88,7 +88,7 @@ android {
 
     defaultConfig {
         applicationId "io.realm.react.testapp"
-        minSdkVersion 16
+        minSdkVersion 19
```
kneth, line 91: I believe that React Native 0.60.0 has `minSdkVersion` set to 16. If we go with 19 (Android 4.4), we need to document it somewhere.

Dockerfile.android
```diff
@@ -37,12 +42,12 @@ RUN dpkg --add-architecture i386 && \
...
 # Jenkins will run this image as user 1001, let's ensure that user exists and has a home directory
 RUN adduser --uid 1001 --disabled-password --gecos "" jenkins
-# Install the SDK and NDK as jenkins
+# Install the SDK and NDK
 USER jenkins
 
 # Install the Android SDK and NDK
 ENV ANDROID_HOME /home/jenkins/android-sdk
-ENV PATH ${PATH}:${ANDROID_HOME}/tools:${ANDROID_HOME}/tools/bin:${ANDROID_HOME}/platform-tools
+ENV PATH ${PATH}:${ANDROID_HOME}/emulator:${ANDROID_HOME}/tools:${ANDROID_HOME}/tools/bin:${ANDROID_HOME}/platform-tools
```
kraenhansen, line 50: Now that we're touching and adding to this, you might want to refactor this to no use the deprecated `ANDROID_HOME` environment variable. See https://developer.android.com/studio/command-line/variables#android_sdk_root.

scripts/test.sh
```diff
@@ -333,49 +333,70 @@ case "$TARGET" in
   RUN_STITCH_IN_FORGROUND=true
   start_server
   ;;
-"react-tests-android")
+"test-android")
+  npm ci --ignore-scripts
```
kraenhansen, line 337: This might depend on our common understanding of the purpose of the `test.sh` script:
- If it's only intended for use on CI, this chance is okay, since we control how its called from the Jenkinsfile.
- If it's supposed to be used locally, I don't think we should use `npm install` and especially not `npm ci` here, since it makes each iteration on tests / bugs take much longer to validate. Is there a less demanding script that we can call to get the intended behaviour?

src/jsc/jsc_rpc_network_transport.hpp
```diff
@@ -18,7 +18,7 @@
 
 #pragma once
 
-#include "sync/generic_network_transport.hpp"
+#include "realm/object-store/sync/generic_network_transport.hpp"
```
kraenhansen, line 21: Is this a pattern that we need to repeat elsewhere?

tests/electron/package.json
```diff
@@ -8,6 +8,7 @@
     "test": "electron --js-flags=\"--expose-gc\" --no-sandbox --enable-logging ./app/main.js"
   },
   "devDependencies": {
+    "cmake-js": "^6.1.0",
```
kraenhansen, line 11: Is this really needed? I hope we don't need users to install "cmake-js" to consume the package?

tests/js/list-tests.js
```diff
@@ -32,6 +32,9 @@ const DATE1 = new Date(1);
 const DATE2 = new Date(2);
 const DATE3 = new Date(3);
 
+const isNodeProcess = typeof process === 'object' && process + '' === '[object process]';
+const isElectronProcess = typeof process === 'object' && process.versions && process.versions.electron;
```
kraenhansen, line 36: I'm surprised to see that we haven't already moved these to `/tests/js/test-utils.js`.
Not saying that should be a part of this PR - just a note 🙂

tests/js/object-tests.js
```diff
@@ -554,46 +554,55 @@ module.exports = {
...
-    async testNotification() {
-        const realm = new Realm({schema: [schemas.StringOnly]});
+    testNotification: async function () {
+        const realm = new Realm({ schema: [schemas.StringOnly] });
 
+        let calls = 0;
+        let resolve;
+        let reject;
+        
+        let promise = new Promise((res, rej) => { resolve = res, reject = rej });
```
kraenhansen, line 564: I don't like this pattern. Why not simply include the following (lines 565-592) in the callback passed to the `Promise` constructor?

tests/js/object-tests.js
```diff
@@ -554,46 +554,55 @@ module.exports = {
         });
     },
 
-    async testNotification() {
-        const realm = new Realm({schema: [schemas.StringOnly]});
+    testNotification: async function () {
```
kraenhansen, line 557: I would really love to see this test get rewritten instead of adding on to the pattern of assigning and reassigning the resolve and reject functions to upper scope local variables. It should be pretty straight forward to rewrite into three `new Promise` calls each registering a listener. IMO we're adding to the technical debt here.

tests/react-test-app/index.android.js
```diff
@@ -88,18 +98,28 @@ async function runTests() {
...
     // write the unit tests reports
     try {
-        await RNFS.writeFile('/sdcard/tests.xml', xmlString, 'utf8');
+        const resultsFile = RNFS.DocumentDirectoryPath + '/tests.xml';
+        await RNFS.writeFile(resultsFile, xmlString, 'utf8');
+        console.log(`Test results file path ${resultsFile}`);
+        console.log("Test Results ==============================================");
+        const lines = xmlString.split(/\r?\n/);
+        //lines.forEach(line => console.log(line));
+        lines.forEach(line => split(line).forEach(str => console.log(str)));
```
kraenhansen, line 107: IMO it is not obvious why we need to chop up the output here, I believe a comment would help.

tests/react-test-app/package.json
```diff
@@ -15,18 +16,24 @@
...
   },
   "devDependencies": {
     "babel-preset-react-native": "4.0.0",
     "eslint": "^4.18.2",
     "fs-extra": "^4.0.3",
     "install-local": "^1.0.0",
     "invariant": "^2.2.2",
-    "metro-react-native-babel-preset": "^0.45.0"
+    "metro-react-native-babel-preset": "^0.45.0",
+    "cmake-js": "^6.1.0"
```
kraenhansen, line 29: Again, I believe we're doing something wrong if our users are forced to install this peer dependency.

tests/react-test-app/run-android.sh
```diff
@@ -5,35 +5,14 @@ set -e
...
-adb install app/build/outputs/apk/release/app-release.apk
-
-echo "Starting the Main Activity"
-adb shell am start -n io.realm.react.testapp/.MainActivity
+echo "Running test application"
+npx react-native run-android
 
-popd
+# granting permissions after react-native run-android since the package needs to be installed
+echo "Granting permission to write to sdcard"
```
kraenhansen, line 17: Might be worth mentioning the need for this? Is it to write the XML output of running the test?

tests/react-test-app/tests/index.js
```diff
@@ -73,6 +73,7 @@ export async function runTest(suiteName, testName) {
     await RealmTests.runTest(suiteName, 'beforeEach');
 
     try {
+        console.log("calling RealmTests.runTest");
```
kraenhansen, line 76: Looks like a leftover.

```suggestion
```

kraenhansen (changes requested):
I would have liked a few more comments on the approach and changes proposed in the description of this PR. Missing it makes it way harder to approach such a big PR.

I have a few questions related to the restructuring too:
- Isn't it an anti-pattern to have an `include` directory as a sub-folder of `src`? (thinking about the JSC headers that got moved to `src/jsc/include`.
- Do we build the librealm.so ahead of time or is this now building on the developers machine?

examples/ReactExample/run-android.sh
```diff
@@ -6,7 +6,7 @@ set -e
 PATH="/opt/android-sdk-linux/platform-tools:$PATH"
 
 # Inform the prepublish script to build Android modules.
-REALM_BUILD_ANDROID=1 npm install realm
+REALM_BUILD_ANDROID_PACKAGE=1 npm install realm
```
steffenagger, line 9: The comment for this line states: `# Inform the prepublish script to build Android modules.`.
But `scripts/prepublish.js` has been deleted & the `prepublish` npm script removed - so is this env-setting still needed?

Jenkinsfile
```diff
@@ -517,12 +520,26 @@ def doDockerInside(script, target, postStep = null) {
   }
 }
 
+def runEmulator() {
+  sh """yes '\n' | avdmanager create avd -n CIRJSEmulator -k 'system-images;android-29;default;x86' --force"""
```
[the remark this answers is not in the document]
fronck: Wouldn't that be to ensure that this command never fails?  It will always evaluate to success, no matter whether `avdmanager ...` fails.

Dockerfile
blagoev: node 10 is EOL in April 2021 we should bump this at that time imo.

Dockerfile.android
blagoev: This might be worth it in a follow up PR.

Jenkinsfile
blagoev: that's the current code Java uses. I can not say for sure

src/node/node_class.hpp
blagoev: mostly for consistency. See this comment https://github.com/realm/realm-js/pull/3434#discussion_r544312411

Jenkinsfile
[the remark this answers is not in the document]
blagoev: that's a directly copy/pasted from what Java team uses.

Jenkinsfile
```diff
@@ -517,12 +520,26 @@ def doDockerInside(script, target, postStep = null) {
...
+  sh "adb start-server" // https://stackoverflow.com/questions/56198290/problems-with-adb-exe
```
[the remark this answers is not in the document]
blagoev: direct copy from what Java team Jenkins file.

Jenkinsfile
```diff
@@ -517,12 +520,26 @@ def doDockerInside(script, target, postStep = null) {
...
+  // Need to go to ANDROID_HOME due to https://askubuntu.com/questions/1005944/emulator-avd-does-not-launch-the-virtual-device
+  sh "cd \$ANDROID_HOME/tools && emulator -avd CIRJSEmulator -no-boot-anim -no-window -wipe-data -noaudio -partition-size 4098 -gpu swiftshader_indirect &"
```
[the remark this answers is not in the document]
blagoev: direct copy from what Java team Jenkins file. I don't see the need to change anything on that if that is working

integration-tests/environments/electron/package.json
```diff
@@ -10,14 +10,15 @@
   "private": true,
   "main": "app/main.js",
   "scripts": {
-    "preinstall": "npm install ../../realm-integration-tests.tgz ../../realm.tgz --no-save",
+    "preinstall": "npm install ../../realm-integration-tests.tgz ../../realm.tgz --no-save --ignore-scripts",
```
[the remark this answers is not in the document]
blagoev: I know for sure. The intent here is to skip native build which is already done in the "prepare integration step"

examples/ReactExample/run-android.sh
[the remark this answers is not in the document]
blagoev: Add clarity to the meaning. `REALM_BUILD_ANDROID` is too vague. Adding `_PACKAGE` is clear that it builds the realm package and no native binaries. Previously meant building native binaries as well. 

Actually this is a good find. This flag is not needed any more. I will remove it.

integration-tests/environments/node/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Realm JS tests running in a Node.js environment",
   "private": true,
   "scripts": {
-    "preinstall": "npm install ../../realm-integration-tests.tgz ../../realm.tgz --no-save",
+    "preinstall": "npm install ../../realm-integration-tests.tgz ../../realm.tgz --no-save --ignore-scripts",
```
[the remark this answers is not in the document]
blagoev: see comment above

package.json
```diff
@@ -56,28 +55,17 @@
...
     "build-changes": "cmake-js build --debug",
     "rebuild-changes": "cmake-js rebuild --debug && cd tests && npm install",
-    "prepublish": "echo prepublishing && node scripts/prepublish.js",
     "eslint": "npm install --ignore-scripts && npm run lint .",
     "license-check": "npm install && license-checker --exclude \"MIT,ISC,BSD,Apache-2.0,BSD-2-Clause,BSD-3-Clause,WTFPL,Unlicense,(MIT AND CC-BY-3.0)\" | node scripts/handle-license-check.js",
     "jsdoc:clean": "rimraf ./docs/output",
     "jsdoc": "npm install --ignore-scripts && npm run jsdoc:clean && jsdoc -u docs/tutorials -p package.json -c docs/conf.json",
-    "start-ros": "./scripts/download-object-server.sh && node ./scripts/test-ros-server.js",
-    "prenode-tests": "npm install --build-from-source=realm && cd tests && npm install",
+    "prenode-tests": "npm install && cd tests && npm install",
```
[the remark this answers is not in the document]
blagoev: this change is not considering if `npm install` should be done here or not it removes `--build-from-source=realm`. 
But its a good find since this change needs to be `--build-from-source`

react-native/android/build.gradle
```diff
@@ -5,461 +5,92 @@ buildscript {
...
-
-        licenses {
-            license {
-                name 'The Apache Software License, Version 2.0'
-                url 'http://www.apache.org/licenses/LICENSE-2.0.txt'
-                distribution 'repo'
-            }
-        }
-    }
+def getNpmVersion() {
```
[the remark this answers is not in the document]
blagoev: agreed

Jenkinsfile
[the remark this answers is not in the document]
steffenagger: It just pipes in "yes /n" as the input to a cli-question, right?

react-native/android/src/main/java/io/realm/react/RealmReactModule.java
```diff
@@ -53,7 +53,7 @@
     private final AssetManager assetManager;
 
     static {
-        SoLoader.loadLibrary("realmreact");
+        SoLoader.loadLibrary("realm");
```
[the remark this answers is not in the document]
blagoev: highly unlikely imo given the use case . Java uses librealm-jni.so for example.

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+const path = require('path');
+const exec = require('child_process').execFileSync;
+
+//simple validation of current directory.
+const rnDir = path.resolve(process.cwd(), "react-native");
+if (!fs.existsSync(rnDir)) {
+    throw new Error("This script needs to be run at the root dir of the project");
+}
+
+const copyOutputPath = path.resolve(process.cwd(), "react-native", "android", "src", "main", "jniLibs");
```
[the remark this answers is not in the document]
blagoev: that's highly personal preference isn't it? I prefer to be more explicit about it using the `process.cwd()` signifying the starting point. Reading `path.resolve(`react-native/android...` without the starting point requires more cognitive time to remember this works from the current directory. Also this is follows a pattern used on multiple places throughout the codebase.

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+
+if (!"ANDROID_NDK" in process.env) {
+    throw Error("ANDROID_NDK environment variable not set");
+}
+const ndkPath = process.env["ANDROID_NDK"];
+
+const sdkPath = getAndroidSdkPath();
+const cmakePath = getCmakePath(sdkPath);
+
+const buildPath = path.resolve(process.cwd(), 'build-realm-android');
```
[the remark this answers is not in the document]
blagoev: same as above

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+'use strict';
+const fs = require('fs');
+const { arch } = require('os');
+const path = require('path');
+const exec = require('child_process').execFileSync;
+
+//simple validation of current directory.
+const rnDir = path.resolve(process.cwd(), "react-native");
```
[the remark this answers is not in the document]
blagoev: see comment above

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+        `-DANDROID_NDK=${ndkPath}`,
+        `-DANDROID_ABI=${arch}`,
+        `-DCMAKE_MAKE_PROGRAM=${sdkPath}/cmake/3.10.2.4988404/bin/ninja`,
+        `-DCMAKE_TOOLCHAIN_FILE=${ndkPath}/build/cmake/android.toolchain.cmake`,
+        "-DANDROID_TOOLCHAIN=clang",
+        "-DANDROID_NATIVE_API_LEVEL=16",
+        `-DCMAKE_BUILD_TYPE=${options.buildType}`,
+        "-DANDROID_STL=c++_static",
+        `-DJSC_ROOT_DIR=${jscDir}`,
+        path.resolve(process.cwd())
```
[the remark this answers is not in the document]
blagoev: good point. this code was refactored down to this path only which can be removed.

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+    args = ["--build", "."];
+    exec(cmakePath, args, { cwd: archBuildDir, stdio: 'inherit' });
+
+    copyOutput(arch, archBuildDir);
+}
+
+generateVersionFile();
+
+function generateVersionFile() {
+    const targetFile = path.resolve(process.cwd(), "react-native", "android", "src", "main", "java", "io", "realm", "react", "Version.java");
```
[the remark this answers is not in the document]
blagoev: same ^

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+public class Version {
+    public static final String VERSION = "${version}";
+}
+`;
+
+    fs.writeFileSync(targetFile, versionFileContents);
+}
+
+function getVersion() {
+    const depencenciesListFile = path.resolve(process.cwd(), "dependencies.list");
```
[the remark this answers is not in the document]
blagoev: same ^

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+    const version = versionValue.split("=")[1];
+    if (!version) {
+        throw new Error("Realm version not found. Invalid version value in dependencies.list file");
+    }
+
+    return version;
+}
+
+function copyOutput(arch, buildDir) {
+    const outFile = path.resolve(buildDir, "src", "android", "libs", arch, "librealm.so");
```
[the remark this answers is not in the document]
blagoev: I don't agree with this one. mix matching path styles is error prone and not readable

scripts/build-android.js
```diff
@@ -0,0 +1,188 @@
...
+    //create a build dir per architecture
+    const archBuildDir = path.resolve(buildPath, arch);
+    if (!fs.existsSync(archBuildDir)) {
+        fs.mkdirSync(archBuildDir);
+    }
+
+    let args = [
+        cmakePath,
+        "-GNinja",
+        `-DANDROID_NDK=${ndkPath}`,
```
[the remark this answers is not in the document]
blagoev: I don't see an issue.

scripts/test.sh
blagoev: This is used to run the android scripts locally and on ci. This PR targets improving on the situation where we didn't have android tests running at all. I would suggest creating an issue and doing a following PR for changes which allow for better develop iterations. Which might take into account our future plans on moving away from this script to something better.

src/jsc/jsc_rpc_network_transport.hpp
blagoev: I don't understand the question? Do you see an issue here?

tests/electron/package.json
blagoev: This is for our tests only. Users will not need to do this

tests/js/object-tests.js
blagoev: This test does strange things for sure. I just fixed the test. It's not a goal to refactor it, although I have debated myself if it would make sense to do that. I decided against it in this PR since it will get out of scope quickly

tests/js/object-tests.js
blagoev: this PR adds nothing to the test. only fixes it to run successfully on CI. We can definitely improve it but this should be don in another PR.

tests/react-test-app/package.json
blagoev: this is our test app. users are not forced to install it.

tests/react-test-app/run-android.sh
blagoev: isn't it obvious. There is a comment and a echo output.

Jenkinsfile
```diff
@@ -417,21 +415,26 @@ def inAndroidContainer(workerFunction) {
...
-    sh 'npm pack'
-    stash includes: 'realm-*.tgz', name: 'android'
+  return {
+    myNode('docker') {
+      unstash 'source'
+      def image = buildDockerEnv('ci/realm-js:android-build', '-f Dockerfile.android')
+      image.inside('-e HOME=/tmp') {
+        sh "./scripts/nvm-wrapper.sh ${nodeTestVersion} npm ci --ignore-scripts" //using --ignore-scripts to skip building for node
+        sh "./scripts/nvm-wrapper.sh ${nodeTestVersion} node scripts/build-android.js"
+        sh "./scripts/nvm-wrapper.sh ${nodeTestVersion} npm pack ."
```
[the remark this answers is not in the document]
blagoev: this will create only one block in Jenkins and combine the shell commands. We are doing it elsewhere in the Jenkins script as well and I suspect it will not improve much on the performance.

Jenkinsfile
```diff
@@ -141,43 +138,43 @@ stage('test') {
...
+        stash includes: 'realm-integration-tests.tgz', name: 'integration-tests-tgz'
+      }
+    }
+  )
+}
+
+stage('integration tests') {
+  parallel(
+    'Node.js v10 on Mac':       buildMacOS { nodeIntegrationTests(nodeTestVersion, it) },
+    'Node.js v10 on Linux':     buildLinux { nodeIntegrationTests(nodeTestVersion, it) },
```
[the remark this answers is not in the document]
blagoev: that's just how it was defined before. I will remove v10 from the name

examples/ReactExample/run-android.sh
blagoev: good catch. removed it

blagoev:
@kraenhansen I have updated the PR description.

blagoev:
> Isn't it an anti-pattern to have an include directory as a sub-folder of src? (thinking about the JSC headers that got moved to src/jsc/include

it's not really. most of the files in `src` are include files

Jenkinsfile
kraenhansen: If we don't know why we're changing this, I would rather that we leave it as is until we know for sure. Alternatively reach our to Java to understand why they feel the need to do this.

Dockerfile.android
kraenhansen: I remember now that I already added an issue tracking this https://github.com/realm/realm-js/issues/3477#issue-767565143

tests/react-test-app/android/app/build.gradle
blagoev: this is our test app. why do you think it needs document somewhere? 
The minSDKVersion change here does not mean we are deprecating Android versions

Jenkinsfile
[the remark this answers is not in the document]
kraenhansen: `yes '\n' |` will pipe a continuous stream of new line characters into STDIN of the `avdmanager` process, presumably to ensure that all prompts are agreed to or something. The reason I'm asking is that I think there's other standard runtime parameters that one can use to achieve this.

I don't think "because javas does it" qualifies as sufficient reason to add this in. This is working around some unknown issue that we as maintainers of the Jenkinsfile need to be aware of. If we cannot find another way to achieve the goal, the least we can do to our fellow developer is to add a note on the origin of this.

I suggest removing it and investigate any issues that might or might not arise from it and document the workaround.

Jenkinsfile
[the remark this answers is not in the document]
kraenhansen: We're duplicating technical debt here. I see plenty of reason to change this.

Jenkinsfile
blagoev: we know exactly why we are changing this. This is the only node which is able to run Android emulator and Android tests currently on our Jenkins infrastructure. Leaving it as it is is not an option

Jenkinsfile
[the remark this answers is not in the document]
blagoev: yes

package.json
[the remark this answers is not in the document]
kraenhansen: Okay - I recognize you're not introducing this and it can be addressed in a followup PR.

Jenkinsfile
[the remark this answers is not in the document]
blagoev: Don't agree. We are making sure we are running exactly the same as the other team is. Trying to outsmart and refactor something on the first go will create technical debt. We need to make sure we are running our Android tests as soon as possible which is what this PR does. But can we refactor things in the future - yes

scripts/build-android.js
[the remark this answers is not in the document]
kraenhansen: In my opinion, this is an example of not using the API as intended.
The `path.resolve` behaviour is as it is to reduce boilerplate code.
I understand that there might be other (more verbose) ways to write the same code, there always will be, but to someone familiar with the Node.js `path` APIs, being this explicit serves little purpose in my opinion.

Jenkinsfile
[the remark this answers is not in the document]
blagoev: I don't understand the issue here.

scripts/build-android.js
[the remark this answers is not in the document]
blagoev: I don't want to get into heated discussion about personal style preferences. There are multiple ways of doing this and I have used the one I like more. I agree you suggestion will work as well. Let's keep the PR comments productive

tests/react-test-app/android/app/build.gradle
kneth: For the example, it is fine.

kneth (approved):

scripts/build-android.js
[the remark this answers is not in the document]
kraenhansen: Did you verify that a space in the path does not break this?

tests/react-test-app/run-android.sh
kraenhansen: It's not obvious why this is needed at all?

kraenhansen:
Do you know how much this will increase our package size, if at all?

blagoev:
it will be around 20mb

scripts/build-android.js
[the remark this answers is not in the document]
blagoev: as I said above, there is no issue with spaces in the path.

scripts/build.bat
```diff
@@ -0,0 +1,15 @@
+cmake.exe ^
+    -GNinja ^
+    -DANDROID_NDK=%ANDROID_NDK% ^
```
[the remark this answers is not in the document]
blagoev: no

tests/react-test-app/run-android.sh
blagoev: this says literally "granting permission to write to sdcard". I think its obvious that the application needs to have this permission

steffenagger (approved):
lgtm

tests/react-test-app/run-android.sh
kraenhansen: But why do we need permission to write to the sdcard, would be my question when I read this code 2 months from now.

blagoev: that's in the code. if code is refactored to require sdcard permission or not in the future this permission and this script may stay not causing any issues. Also this permission is present in the manifest file but we are not writing comments there as well. 
In my view there is little value in adding another comment next to these two lines.

the first comment says why we are doing that after app install and the `echo` writes what it does for logging. Its perfect imo. If we say "Giving permission to write to sdcard so we can write our test output file there" means we are setting code behavior there. Then if code is refactored later to also write something else we need to remember to change this comment as well (which is in another file and so on). In my view that's not needed and is of little value

blagoev: I added a comment there. I still think that adds little value.

react-native/android/src/main/java/io/realm/react/RealmReactModule.java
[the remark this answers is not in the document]
kraenhansen: 🤞 I guess

src/jsc/jsc_rpc_network_transport.hpp
kraenhansen: No, but i was wondering if there was a reason this is the only place you're making the change. I can look into this myself afterwards. Just curious.

tests/electron/package.json
kraenhansen: I need to spent more time understanding the need for this, it smells like an undocumented peer dependency to me.

src/jsc/jsc_rpc_network_transport.hpp
blagoev: that's cause this file changed its location in the monorepo

scripts/test.sh
kraenhansen: Can't we just move this line out into the Jenkinsfile? Then I would be happy 👍

kraenhansen (commented):
- I do think adding code that has been blindly copied from the Java team's Jenkinsfile is an issue. We're maintaining this and should therefore be able to reason about it.
- I still feel the use of `path.resolve` doesn't comply with [the examples in the documentation](https://nodejs.org/api/path.html#path_path_resolve_paths) and is unnecessarily verbose.

However, I don't object against this PR being merged, as it contains a lot of changes that we need to proceed.

blagoev:
@kraenhansen 

> I do think adding code that has been blindly copied from the Java team's Jenkinsfile is an issue. We're maintaining this and should therefore be able to reason about it.

Copying Java teams code for the Android Jenkins infrastructure is no issue since its working for them and for us. And its not `blindly copied`. I don't agree with this statement. The copied scripts have been tested for our project and we are able to reason about them. If you have any questions about the scripts that are part of this PR feel free to ask I will be happy to break it down for you.

> I still feel the use of path.resolve doesn't comply with the examples in the documentation and is unnecessarily verbose.
I can't agree with you assessment about `path.resolve`  usage. Also don't quite understand what you mean by `doesn't comply with the examples.`
`path.resolve` in the node docs you linked is documented to accept a sequence of path arguments. It's `path.resolve` responsibility to stitch them together using the correct path separator.  Examples are just examples there is no better and worst way of invoking this function. To me your comment is hankering for issues. 
As I said its highly personal. For me being explicit is of more value when it comes to build scripts, build paths etc. This is also reflected in the logging this PR introduces. The more explicit the logs the better way to understand why a debug script is not working locally and on CI.

kraenhansen (approved):

scripts/test.sh
blagoev: that would break the local runs of this script. Unfortunately we have this script which tries to provide both local and CI runs. Maybe our next solution to this would be more robust. 
What I have chosen is to do is do all scripts in JavaScript. This is universal for us since we have Node everywhere. So it should run on every OS for example unlike the shell differences between Linux and Mac that we experience currently. And is more easily debugged and fixed by everyone when needed. Hence my `build-android.js` here. Also my opinion is that the Jenkins script should only be invoking these JS scripts the same way we are running them locally. This way we will not find ourselves in the situation to have big discrepancies between how Jenkins and local scripts are run.

PR #3479, opened 2020-12-16
adding cmake-js documentation

cesarvr:
Some documentation for using `cmake` more efficiently.

### DIFF
--- a/contrib/how-to-build.md
+++ b/contrib/how-to-build.md
@@ -51,6 +51,6 @@ cmake-js build
 In order to debug the binary we need to tell the compiler to add debug symbols into it:
 
 ```sh
-cmake-js <build/rebuild> --debug
+npx cmake-js <build/rebuild> --debug
 ```
 > Now you can use tools like lldb to debug the native module.
--- a/contrib/how-to-build.md
+++ b/contrib/how-to-build.md
@@ -26,7 +26,7 @@ git submodule update --init --recursive
 Now you should get [cmake-js](https://www.npmjs.com/package/cmake-js) which is a [node-gyp](https://github.com/nodejs/node-gyp) replacement:
 
 ```sh
-  npm install cmake-js -g
+  npm install
 ```
 
 ## Compiling From Scratch
--- a/contrib/how-to-build.md
+++ b/contrib/how-to-build.md
@@ -1,0 +1,56 @@
+# How To Build With `cmake`
+
+To build this project you can still use `npm install`, but here some advice to improve your development experience:
+
+
+## Getting Started
+
+First clone the project:
+```sh
+git clone git@github.com:realm/realm-js.git
+
+# The only branch supporting cmake at the moment is develop
+git checkout develop
+```
+
+Pull the submodules:
+
+```sh
+git submodule update --init --recursive
+```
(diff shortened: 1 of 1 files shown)


contrib/how-to-build.md
```diff
@@ -0,0 +1,56 @@
...
+
+```sh
+git submodule update --init --recursive
+```
+> If this folder `src/object-store` exist then do a `git clean -xfd`.
+
+
+
+## Configuration
+Now you should get [cmake-js](https://www.npmjs.com/package/cmake-js) which is a [node-gyp](https://github.com/nodejs/node-gyp) replacement:
```
kneth, line 26: `cmake-js` is a dev dependency so `npm install` should install it.

Of course, if you want to run `cmake-js` manually, it might be easier to install it globally.

kneth (approved):

contrib/how-to-build.md
```diff
@@ -0,0 +1,56 @@
...
+
+```sh
+  npm install cmake-js -g
```
kraenhansen, line 29: In general I would advice against globally installed packages, since these are hard to keep updated and we would be running different versions locally and on CI.

Instead use dev-dependencies (as Kenneth mentions, it is a dev-dependency now) and the `npx cmake-js` command to run its CLI.

```suggestion
  npm install
```

contrib/how-to-build.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## Compiling From Scratch
+
+To build the project from scratch you can use the `rebuild` command:
+
+```sh
+cmake-js rebuild
```
kraenhansen, line 37:
```suggestion
npx cmake-js rebuild
```

contrib/how-to-build.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## Fast Iteration
+
+To improve the development speed, you should avoid recompiling the whole project each time you make new changes, to avoid that you can pass the `build` argument:
+
+```sh
+cmake-js build
```
kraenhansen, line 45:
```suggestion
npx cmake-js build 
```

contrib/how-to-build.md
```diff
@@ -0,0 +1,56 @@
...
+cmake-js build 
+```
+> Now `cmake` should detect changes in the workspace and compile only the necessary dependencies.
+
+## Debug
+
+In order to debug the binary we need to tell the compiler to add debug symbols into it:
+
+```sh
+cmake-js <build/rebuild> --debug
```
kraenhansen, line 54:
```suggestion
npx cmake-js <build/rebuild> --debug
```

kraenhansen (changes requested):
Seems like a valuable add. Let's keep it updated as develop gets merged into master.
I have a few minor suggestions on how to invoke the `cmake-js` CLI.

Did you notice the `package.json` also [added invoking the `cmake-js` CLI in its scripts](https://github.com/realm/realm-js/blob/develop/package.json#L57-L58)?
That way `npx cmake-js build --debug` could also be invoked with `npm run build-changes`.

fronck (approved):
LGTM

PR #3482, opened 2020-12-18
Datatype: UUID

### LINKED ISSUE #3244: API for data type UUID
## Schema

A schema with a UUID property:

```js
const Person = {
  name: “Person”,
  properties: {
    id: “uuid”,
    name: “string”
  }
};
```

## Creating objects with UUID

The BSON library does not have a class for UUID. Currently users of the MongoDB node.js driver are using the uuid-mongodb package. Using the 3rd party package, creating an object will look like:

```js
const MUUID = require('uuid-mongodb');
realm.write(() => {
  realm.create(“Person”, { id: MUUID.v4(), name: “John Doe” });
});
```

When a proper UUID class is added to the BSON library, we imagine that we should use the following instead:

```js
const UUID = require('bson').UUID;
realm.write(() => {
  realm.create(“Person”, { id: new UUID(), name: “John Doe” });
});
```

The above API for creating a new object will be consistent with the API for both ObjectId and Decimal128 (also from the BSON library).

kneth:
## What, How & Why?

Implementation of data type `UUID`.

This closes #3244, #3387, #3388

## ☑️ ToDos
* [x] Schema parsing
* [x] Accessors (node)
* [x] Integrate this type into Mixed #[3501](https://github.com/realm/realm-js/pull/3501).
* [x] Accessors (jsc) - **NOT TESTED ON IOS**
* [x] Switch to BSON's UUID class - **Referencing a pre-released GH commit from an open PR**
* [x] Test: schema parsing - required property
* [x] Test: schema parsing - optional property
* [x] Test: create object with UUID required property
* [x] Test: create object with optional UUID property
* [x] Test: query object
* [x] Test: lists
* [x] Test: query list
* [x] Test: UUID as primary keys
* [x] Test: UUID as partitionKey (**NOT** complete roundtrips, only SDK -> Core/OS -> Sync)
* [x] Test: UUID roundtrip through Sync (cloud)
  * [x] as primary key 
  * [x] as mandatory property
  * [x] as optional property
  * [x] as list ("uuid[]")
* [x] typescript definitions file is updated **(this awaits partitionKey)**
* [ ] jsdoc files updated
* [ ] 📝 Changelog entry

## Additional changes
* Tests & Integration-tests now uses the bson-package bundled with Realm


integration-tests/tests/src/schemas/person-and-dog-with-object-ids.ts
```diff
@@ -19,10 +19,9 @@
 /* tslint:disable max-classes-per-file */
 
 import Realm from "realm";
-import { ObjectId } from "bson";
 
 export interface IPerson {
-    _id: ObjectId;
+    _id: Realm.BSON.ObjectId;
```
takameyer, line 24: Would it be possible to make `ObjectId` a named import from `realm`?
Then we could import it so:
`import Realm, { ObjectId } from "realm";`

src/js_types.hpp
```diff
@@ -326,7 +330,7 @@ struct Object {
     }
     static ValueType call_method(ContextType ctx, const ObjectType &object, const String<T> &name, const std::initializer_list<ValueType> &arguments) {
         return call_method(ctx, object, name, (uint32_t)arguments.size(), arguments.begin());
-    }
+        }
```
takameyer, line 333:
```suggestion
    }
```
remove tab

src/js_types.hpp
```diff
@@ -565,13 +569,6 @@ inline typename T::Value Object<T>::create_from_optional_app_error(ContextType c
     return create_from_app_error(ctx, *error);
 }
 
-
-
-template<typename T>
-inline typename T::Value Value<T>::from_uuid(typename T::Context ctx, const UUID& value) {
-    throw std::runtime_error("'UUID' type support is not implemented yet");
-}
-
```
takameyer, line 574: Shouldn't this function be implemented?
On line 588 there is a runtime_error for Type::Uuid.  I think this is where `from_uuid` needs to be implemented.

steffenagger: It's depending on environment, so resides in `src/jsc/jsc_value.hpp` & `src/node/node_value.hpp` (this was inserted as a placeholder).

integration-tests/tests/src/schemas/person-and-dog-with-object-ids.ts
steffenagger: I hope that's a goal for the future. Not really possible atm.

takameyer: Me too.  That would definitely be more ergonomic.

kneth (commented):
:+1:

It is good to see so many tests.

tests/js/query-tests.js
```diff
@@ -19,12 +19,11 @@
...
-var Decimal128 = require('bson').Decimal128;
-
-var Realm = require('realm');
-var TestCase = require('./asserts');
-var testCases = require('./query-tests.json');
-var schemas = require('./schemas');
+var Realm = require("realm");
+var TestCase = require("./asserts");
+var testCases = require("./query-tests.json");
+var schemas = require("./schemas");
```
takameyer, line 25:
```suggestion
const Realm = require("realm");
const TestCase = require("./asserts");
const testCases = require("./query-tests.json");
const schemas = require("./schemas");
```

tests/js/uuid-sync-tests.js
```diff
@@ -0,0 +1,131 @@
...
+
+        const appConfig = AppConfig.integrationAppConfig;
+        const app = new Realm.App(appConfig);
+        const credentials = Realm.Credentials.anonymous();
+
+        const user = await app.logIn(credentials);
+        const config = {
+            sync: {
+                user,
+                partitionValue: "UUID",
```
takameyer, line 50: Could this partitionValue just be any string?  "UUID" has no special meaning here, correct?

steffenagger: correct, it just seemed to fit the theme :)

takameyer (approved):

cesarvr (approved):
LGTM

package.json
```diff
@@ -70,7 +70,7 @@
   },
   "dependencies": {
     "bindings": "^1.5.0",
-    "bson": "^4.0.3",
+    "bson": "github:steffenagger/js-bson#uuid-realm",
```
kraenhansen, line 73: Is this the final value or would this update before this PR is merged?

steffenagger: Ideally this links to a released bson-version, but I'm not sure that will happen in time.

src/jsc/jsc_value.cpp
```diff
@@ -99,6 +99,20 @@ JSValueRef jsc::Value::from_object_id(JSContextRef ctx, const ObjectId& value)
...
 
+template<>
+JSValueRef jsc::Value::from_uuid(JSContextRef ctx, const UUID& value)
+{
+    static jsc::String s_realm = "Realm";
+    static jsc::String s_uuid = "_UUID";
+
+    JSObjectRef global_object = JSContextGetGlobalObject(ctx);
+    JSObjectRef realm_constructor = jsc::Object::validated_get_constructor(ctx, global_object, s_realm);
+    JSObjectRef uuid_constructor = jsc::Object::validated_get_constructor(ctx, realm_constructor, s_uuid);
```
kraenhansen, line 110: This won't work when the app is in Chrome debugging mode, since the `_UUID` won't be available on the `Realm` constructor. This is a problem for the other BSON types as well.

steffenagger: Yeah, this implementation is suffering the same limitations as `ObjectId` currently sees. I've planned to look at it tomorrow with @takameyer.

cesarvr (approved):
LGTM 👍

PR #3501, opened 2021-01-13
Cesarvr/mixed implementation

cesarvr:
### Mixed Type

It provide integration between the realm::Mixed type into our Realm JS SDK. 

#### Usage
As mentioned before the Mixed type basically allows the user to delegate the type inference at runtime to Realm-JS, so instead of defining the types in the schema like this: 

```js
const Schema = {
    name: 'Test',
    properties: {
     a: 'int',
     b: 'string', 
     c: 'bool'
    }
}
```

Which require the user to remember the types, with the new **Mixed type** this is not longer necessary because this type deduct the type automatically, you can use the below schema interchangeably with the one above: 

```js
const Schema = {
    name: 'Test',
    properties: {
     a: 'mixed',
     b: 'mixed', 
     c: 'mixed'
    }
}
```

This in combination with Dictionaries can facilitate a more agile way of development where the users are not constrained by initial schema rules or mistakes. 

### Performance

To make sure we don’t degrade the user experience, we tested the performance by creating two tables with 2 Million rows using a wide variety of types, then we use the fixed schema as the reference for the performance. 

```xml
Mixed 207 seconds
Schema 175 seconds
```
> We found out that the Mixed type is just 18% slower than the non-Mixed.

##### Profiling

To understand this slowdown we can take a look at the profiling information: 

We expend the majority of CPU time is spend deducting types and parsing complex types as we can see in the figure below: 

Screenshot 2021-01-13 at 13 59 09

Another explanation for the performance degradation vs the fixed schema option is the time spent by the CPU dealing with the Mixed type in Realm Core: 

Screenshot 2021-01-13 at 14 03 25

Those are the points where we need to focus for further speed improvements. 

This closes # [RJS-544](https://jira.mongodb.org/browse/RJS-544)

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


bmunkholm:
Just a comment to the motivation. It's not (originally) intended to defer type decisions but to allow storing mixed values when there is a need for that.

cesarvr:
> Just a comment to the motivation. It's not (originally) intended to defer type decisions but to allow storing mixed values when there is a need for that.

Thanks for the comment @bmunkholm, but I think I used the word *inference*, instead of defer, meaning that the framework will figure out the type for you. 

And of course the side effect of that it will store everything you trow at it.

bmunkholm:
> This in combination with Dictionaries can facilitate a more agile way of development where the users are not constrained by initial schema rules or mistakes.

It doesn't have any effect to the feature itself (that's all good), I'm merely pointing out that above is not really the motivation for this feature or how we will promote the use of it. It shouldn't make you "lazy" or defer decisions about your data types you want.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
+* Adding Mixed types [RJS-544](https://jira.mongodb.org/browse/RJS-544).
```
kneth, line 8: Should go as an enhancement.

src/common/type_deduction.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
kneth, line 1: Please add our generic copyright notice (see any `.hpp` file).

src/js_mixed.hpp
```diff
@@ -0,0 +1,170 @@
+#include <common/type_deduction.hpp>
```
kneth, line 1: Please add our copyright notice.

src/js_mixed.hpp
```diff
@@ -0,0 +1,170 @@
...
+        {DataType::type_String, new HandleString<Context, Value, Utils>},
+        {DataType::type_Int, new HandleNumber<Context, Value, Utils, Int>},
+        {DataType::type_Float, new HandleNumber<Context, Value, Utils, Float>},
+        {DataType::type_Double,
+         new HandleNumber<Context, Value, Utils, Double>},
+        {DataType::type_Bool, new HandleBoolean<Context, Value, Utils>},
+        {DataType::type_Decimal, new HandleDecimal128<Context, Value, Utils>},
+        {DataType::type_ObjectId, new HandleObjectID<Context, Value, Utils>},
+        {DataType::type_Binary, new HandleBinary<Context, Value, Utils>},
+        {DataType::type_Timestamp, new HandleTimeStamp<Context, Value, Utils>},
```
kneth, line 132: Please add a todo item in #3482 so we remember to add UUID support.

src/js_object_accessor.hpp
```diff
@@ -375,8 +379,8 @@ struct Unbox<JSEngine, BinaryData> {
 
 template<typename JSEngine>
 struct Unbox<JSEngine, Mixed> {
-    static Mixed call(NativeAccessor<JSEngine> *ctx, typename JSEngine::Value const& value, realm::CreatePolicy, ObjKey) {
-        throw std::runtime_error("'Mixed' type support is not implemented yet");
+    static Mixed call(NativeAccessor<JSEngine> *native_accessor, typename JSEngine::Value const& value, realm::CreatePolicy, ObjKey) {
```
kneth, line 382: We should rename `ctx` in the other `Unbox` instances too. But should go in a separate PR.

src/js_schema.hpp
```diff
@@ -87,6 +88,9 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
     if (type == "bool") {
         prop.type |= PropertyType::Bool;
     }
+    else if (type == "mixed") {
```
kneth, line 91: I believe `Mixed` is always optional (and you shouldn't explicitly set it). Should we add a check or leave it to object store?

src/js_schema.hpp
```diff
@@ -95,8 +99,7 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
     }
     else if (type == "double") {
         prop.type |= PropertyType::Double;
-    }
-    else if (type == "string") {
+    }else if (type == "string") {
```
kneth, line 102: Please revert this change - or let `clang-format` do it :smile:

src/js_types.hpp
```diff
@@ -30,12 +30,14 @@
...
 #include <realm/util/optional.hpp>
 #include <realm/util/base64.hpp>
-#include <realm/mixed.hpp>
+
 
 #include <realm/object-store/util/bson/bson.hpp>
 #include <realm/object-store/util/event_loop_dispatcher.hpp>
 #include <realm/object-store/sync/generic_network_transport.hpp>
 
+//#include "js_mixed.hpp"
```
kneth, line 39: Can it be removed?

src/js_types.hpp
```diff
@@ -155,7 +157,7 @@ struct Value {
     static ValueType from_nonnull_binary(ContextType, BinaryData);
     static ValueType from_undefined(ContextType);
     static ValueType from_timestamp(ContextType, Timestamp);
-    static ValueType from_mixed(ContextType, const Mixed &);
+    //static ValueType from_mixed(ContextType, const Mixed &);
```
kneth, line 160: Can it be removed?

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
```
kneth, line 1: Please add copyright notice.

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,92 @@
...
+    static bool is_decimal128(Value &value) {
+        return TypeDeductionImpl::is_bson_type(value, "Decimal128");
+    }
+
+    static bool is_object_id(Value &value) {
+        return TypeDeductionImpl::is_bson_type(value, "ObjectID");
+    }
+
+    static std::string realm_typeof(DataType value) {
+        std::map<DataType, std::string> realm_typeof = {
```
kneth, line 33: If it can be a `std::array`, line 41 will have constant time complexity.

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,92 @@
...
+            {DataType::type_Float, "Float"},
+            {DataType::type_Double, "Double"},
+            {DataType::type_Decimal, "Decimal128"},
+            {DataType::type_Bool, "Boolean"},
+            {DataType::type_ObjectId, "ObjectId"}};
+        return realm_typeof[value];
+    }
+
+    static DataType typeof(Value &value) {
+        if (value.IsNull()) {
```
kneth, line 45: The returned type here and in line 61 is a bit surprising to me. Can you explain why `null` and `undefined` is a `TypedLink`?

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,92 @@
...
+            return DataType::type_ObjectId;
+        }
+        if (value.IsObject()) {
+            return DataType::type_Link;
+        }
+
+        return DataType::type_Link;
+    }
+
+    static bool is_boolean(Value &value) { return value.IsBoolean(); }
```
kneth, line 80: Should these functions be `inline`d?

src/common/type_deduction.hpp
```diff
@@ -0,0 +1,37 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
fronck, line 3: *2021 ;)

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <common/type_deduction.hpp>
```
fronck, line 21: Would you not need the `#if REALM_PLATFORM_NODE` guard here as well?

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+template <typename Context, typename Value>
+class MixedWrapper {
+   public:
+    virtual Mixed wrap(Context, Value const &) = 0;
+    virtual Value unwrap(Context, Mixed) = 0;
+};
+
+template <typename Context, typename Value, typename Utils>
+class HandleString : public MixedWrapper<Context, Value> {
```
fronck, line 38: I think these classes would be better named Mixed* (e.g., MixedString) rather than Handle*.  Just to make it clear that they deal with Mixed values only.

nirinchev:
Another note on the description of the PR: you can't use those two schemas interchangeably because:
1. You'll need a migration for a local Realm.
2. That'll be a destructive change for a synchronized Realm which is not allowed.

docs/realm.js
```diff
@@ -472,6 +472,7 @@ class Realm {
  * @typedef Realm~PropertyType
  * @type {("bool"|"int"|"float"|"double"|"string"|"decimal128"|"objectId"|"date"|"data"|"list"|"linkingObjects"|"<ObjectType>")}
  *
+ * @property {Mixed} "mixed" - Property value that allow any of the following types (`"bool","int","float","double","string","decimal128","objectId","date","data"`), this type is nullable by default.
```
nirinchev, line 475: Mixed can also hold objects, not only primitives.

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+   private:
+    // we need this <cache> to keep the value life long enough to get into the
```
nirinchev, line 40:
```suggestion
    // we need this <cache> to keep the value alive long enough to get into the
```

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+    Value unwrap(Context context, Mixed mixed) {
+        return Utils::from_object_id(context, mixed.get<ObjectId>());
+    }
+};
+
+template <typename Context, typename Value, typename Utils>
+class HandleBinary : public MixedWrapper<Context, Value> {
+    Mixed wrap(Context context, Value const &value) {
+        auto owned_binary_data = Utils::to_binary(context, value);
```
nirinchev, line 105: `BinaryData` and `StringData` are very similar in Core - they don't hold a reference to their contents. So it's surprising that you need the cache for the string case, but not in the binary case.

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+        return Utils::from_binary(context, mixed.get<BinaryData>());
+    }
+};
+
+template <typename Context, typename Value, typename Utils>
+class HandleTimeStamp : public MixedWrapper<Context, Value> {
+    Mixed wrap(Context context, Value const &value) {
+        auto date = Utils::to_date(context, value);
+
+        double milliseconds = Utils::to_number(context, date);
```
nirinchev, line 119: [q] Isn't it better to `std::move` the `date` (and the `ts` at the end) variables here?

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+template <typename JavascriptEngine>
+class TypeMixed {
+   private:
+    using Context = typename JavascriptEngine::Context;
+    using Value = typename JavascriptEngine::Value;
+    using Utils = js::Value<JavascriptEngine>;
+    using Strategy = MixedWrapper<Context, Value> *;
+
+    std::map<DataType, Strategy> strategies = {
```
nirinchev, line 140: This seems to be missing a strategy for object or is that handled elsewhere?

src/js_mixed.hpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    Value wrap(Context context, Mixed mixed) {
+        auto strategy = get_strategy(mixed.get_type());
+
+        if (strategy == nullptr) {
+            throw std::runtime_error(
+                "The " + TypeDeduction::realm_typeof(mixed.get_type()) +
+                " value is not supported for the mixed type.");
+        }
```
nirinchev, line 169: If this is hit, it would be a bug in the SDK, wouldn't it? There should be no type in `Mixed` that can't be wrapped to `Value`.

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        if (value.IsBoolean()) {
+            return type_Bool;
+        }
+        if (value.IsDate()) {
+            return type_Timestamp;
+        }
+        if (value.IsUndefined()) {
+            return type_TypedLink;
+        }
```
nirinchev, line 77: Why is undefined returned as typedlink?

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,107 @@
...
+        if (TypeDeductionImpl::is_decimal128(value)) {
+            return type_Decimal;
+        }
+        if (TypeDeductionImpl::is_object_id(value)) {
+            return type_ObjectId;
+        }
+        if (value.IsObject()) {
+            return type_Link;
+        }
```
nirinchev, line 90: Shouldn't this be `TypedLink`?

src/node/node_type_deduction.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+        return type_Link;
```
nirinchev, line 92: This seems like a weird default - perhaps hitting this should be an error or return `type_Invalid`?

fronck (changes requested):
A few nitpicks about the explicitness of the Mixed handler types, and a possible memory leak we need to look at. 

Also, some tests are needed.

Otherwise, good work.

docs/realm.js
cesarvr: Yes that's correct but that's something we plan to support in the next release.

src/js_mixed.hpp
cesarvr: Good catch, thanks for the heads up 👍

src/js_mixed.hpp
cesarvr: Yes that's correct but that's something we plan to support in the next release.

src/js_mixed.hpp
cesarvr: That's a good observation, but even in that unlikely case it would be nice to fail gracefully.

nirinchev: If we did want to fail gracefully, I would suggest rewording the error message to something that makes it clear to the developer that this is a bug on our end and they should file a ticket.

PR #3511, opened 2021-01-18
Hotfix: async realm open exception crash

### LINKED ISSUE #3414: If an error occurs during Realm.open, the app crashes.
In case of an error during Realm.open, the app crashes instead of rejecting the promise with the error.

Reported issue: [https://github.com/realm/realm-js/issues/3409](https://github.com/realm/realm-js/issues/3409)

Error caught caught through Flipper:

```sh
Ignored the notification for the crash 
{notificationID: "1", callstack: "/Volumes/Android/buildbot/src/android/ndk-release-r21/external/libcxx/../../external/libcxxabi/src/abort_message.cpp:72: abort_message: assertion "terminating with uncaught exception of type realm::InvalidAdditiveSchemaChangeException: The following changes cannot be made in additive-only schema mode:


* Property 'wishlist._id' has been made required.
* Property 'wishlist._partition' has been made required.
* Property 'wishlist.category' has been made required.
* Property 'wishlist.code' has been made required.
* Property 'wishlist.dateCreated' has been made required.
* Property 'wishlist.dateModified' has been made required.
* Property 'wishlist.description' has been made required.
* Property 'wishlist.listItems' has been changed from 'array<wishlist_listItems>' to 'array<items>'.
* Property 'wishlist.name' has been made required.
* Property 'wishlist.owner' has been made required.
* Property 'wishlist.status' has been made required." failed
Fatal signal 6 (SIGABRT), code ~~6 (SI_TKILL) in tid 2162 (mqt_js), pid 2083 (com.lysts)", name: "/Volumes/Android/buildbot/src/android/ndk~~release-r21/external/libcxx/../../external/libcxxabi/src/abort_message.cpp:72: abort_message: assertion "terminating with uncaught exception of type realm::InvalidAdditiveSchemaChangeException: The following changes cannot be made in additive-only schema mode:", reason: "Cannot figure out the cause", date: Thu Nov 19 2020 08:12:29 GMT+0100 (Central European Standard Time)}
```

steffenagger:
## What, How & Why?
App crashed if a native error was thrown during `Realm.open(...)` - (fix by: @RedBeard0531).

This closes #3414 

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/tests/js/session-tests.js
+++ b/tests/js/session-tests.js
@@ -177,6 +177,39 @@ module.exports = {
             });
     },
 
+    testRealmOpenWithDestructiveSchemaUpdate() {
+        if (!isNodeProcess) {
+            return;
+        }
+
+        const partition = Utils.genPartition();
+        const expectedObjectsCount = 3;
+
+        let user, config;
+        let credentials = Realm.Credentials.anonymous();
+        let app = new Realm.App(appConfig);
+        return runOutOfProcess(__dirname + '/download-api-helper.js', appConfig.id, appConfig.url, partition, REALM_MODULE_PATH)
+            .then(() => { return app.logIn(credentials) })
+            .then(u => {
+                user = u;
+                config = getSyncConfiguration(u, partition);
+                return Realm.open(config)
+            }).then(realm => {
+                realm.close();
+                // change the 'breed' property from 'string?' to 'string'
+                config.schema[0].properties.breed = "string";
+            }).then(() => {
+                return Realm.open(config);
+            }).then((realm) => {
+                // this should never happen (a better way of doing this?)
+                TestCase.assertNull(realm, "UNEXPECTED: Realm.open() with destructive schema update resolved.");
+            }).catch((err) => {
+                TestCase.assertDefined(err);
+                TestCase.assertDefined(err.message);
+                TestCase.assertTrue(/^The following changes cannot be made in additive-only schema mode:/.test(err.message));
+            });
+    },
+
     testRealmOpenWithExistingLocalRealm() {
         if (!platformSupported) {
             return;
--- a/tests/js/session-tests.js
+++ b/tests/js/session-tests.js
@@ -177,37 +177,30 @@ module.exports = {
             });
     },
 
-    testRealmOpenWithDestructiveSchemaUpdate() {
+    async testRealmOpenWithDestructiveSchemaUpdate() {
         if (!isNodeProcess) {
             return;
         }
 
         const partition = Utils.genPartition();
-        const expectedObjectsCount = 3;
 
-        let user, config;
-        let credentials = Realm.Credentials.anonymous();
-        let app = new Realm.App(appConfig);
-        return runOutOfProcess(__dirname + '/download-api-helper.js', appConfig.id, appConfig.url, partition, REALM_MODULE_PATH)
-            .then(() => { return app.logIn(credentials) })
-            .then(u => {
-                user = u;
-                config = getSyncConfiguration(u, partition);
-                return Realm.open(config)
-            }).then(realm => {
-                realm.close();
-                // change the 'breed' property from 'string?' to 'string'
-                config.schema[0].properties.breed = "string";
-            }).then(() => {
-                return Realm.open(config);
-            }).then((realm) => {
-                // this should never happen (a better way of doing this?)
-                TestCase.assertNull(realm, "UNEXPECTED: Realm.open() with destructive schema update resolved.");
-            }).catch((err) => {
-                TestCase.assertDefined(err);
-                TestCase.assertDefined(err.message);
-                TestCase.assertTrue(/^The following changes cannot be made in additive-only schema mode:/.test(err.message));
-            });
+        await runOutOfProcess(__dirname + "/download-api-helper.js", appConfig.id, appConfig.url, partition, REALM_MODULE_PATH);
+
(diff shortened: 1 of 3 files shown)


tests/js/session-tests.js
```diff
@@ -177,6 +177,39 @@ module.exports = {
...
+            return;
+        }
+
+        const partition = Utils.genPartition();
+        const expectedObjectsCount = 3;
+
+        let user, config;
+        let credentials = Realm.Credentials.anonymous();
+        let app = new Realm.App(appConfig);
+        return runOutOfProcess(__dirname + '/download-api-helper.js', appConfig.id, appConfig.url, partition, REALM_MODULE_PATH)
```
RedBeard0531, line 191: This could be personal style, but I think this would be clearer/simpler using `await` style:

```js
    async testRealmOpenWithDestructiveSchemaUpdate() {
        if (!isNodeProcess) {
            return;
        }

        const partition = Utils.genPartition();
        const expectedObjectsCount = 3;

        let credentials = Realm.Credentials.anonymous();
        let app = new Realm.App(appConfig);
        await runOutOfProcess(__dirname + '/download-api-helper.js', appConfig.id, appConfig.url, partition, REALM_MODULE_PATH);
        let user = await app.logIn(credentials);
        let config = getSyncConfiguration(user, partition);
        let realm = await Realm.open(config)
        realm.close();
        // change the 'breed' property from 'string?' to 'string'
        config.schema[0].properties.breed = "string";
        await TestCase.assertThrowsAsyncContaining(
            async() => await Realm.open(config), // This crashed in bug #3414.
            'The following changes cannot be made in additive-only schema mode:');
    },
```

This version still crashes without the fix and passes with it.

tests/js/session-tests.js
```diff
@@ -177,6 +177,39 @@ module.exports = {
...
+                config = getSyncConfiguration(u, partition);
+                return Realm.open(config)
+            }).then(realm => {
+                realm.close();
+                // change the 'breed' property from 'string?' to 'string'
+                config.schema[0].properties.breed = "string";
+            }).then(() => {
+                return Realm.open(config);
+            }).then((realm) => {
+                // this should never happen (a better way of doing this?)
```
RedBeard0531, line 204: See `TestCase.assertThrowsAsyncContaining` above. Perhaps that (and `assertThrowsAsync`) should be modified to check if the argument is already a promise and not call it, then you could just pass `Realm.open(...)` without wrapping in an extra lambda, but even without that change, I think it is an improvement. Also, IIUC, your version would still pass even if it failed on line 196, even though it is supposed to only be checking for a failure on line 202. You *could* solve that by putting the `.catch()` on line 202 directly on the `Realm.open()` call rather than at the top-level chain, but I think using the helper (and doing the await transform) is a better solution. (Technically that would also catch an immediate error in Realm.open() before it returns the promise, but I think that is OK)

tests/js/session-tests.js
steffenagger: If so, we have the same preferences 🙂
I kept this format (promise-chaining), as it's used in all other tests in the `session-tests.js` file.

steffenagger: Updated.

tests/js/session-tests.js
```diff
@@ -177,6 +177,32 @@ module.exports = {
...
+
+        const partition = Utils.genPartition();
+
+        await runOutOfProcess(__dirname + "/download-api-helper.js", appConfig.id, appConfig.url, partition, REALM_MODULE_PATH);
+
+        const app = new Realm.App(appConfig);
+        const user = await app.logIn(Realm.Credentials.anonymous());
+        const config = getSyncConfiguration(user, partition);
+
+        throw new Error('Steffen threw a new Error()');
```
RedBeard0531, line 193: 💣

blagoev: `throw new Error('Steffen threw a new Error()');`
Lets keep it more generic and meaningful?

steffenagger: OOOPS, that was a test NOT supposed to go in. Thanks for catching that!
(giant facepalm, sorry guys)

kneth (approved):

tests/js/session-tests.js
blagoev: oh no worries. I do this all the time so I can find test pieces I need to remove before merging.

cesarvr (approved):
👍

tests/js/session-tests.js
steffenagger: Well the dumbest part is that it was to showcase something to @RedBeard0531 in a conversation on slack... and 2 mins later he finds this genius line.. here 🙄

(also, normally @steffenagger wouldn't reference @steffenagger in the 3rd person, it was just for the end output 😬)

tholhubner:
Is this live on npm? I am having this kind of issue and wanted to inquire on this ticket since this seems to be the problem I am running into. I know there has been a release since this was merged, but I am still getting a crash on open. Can open a new issue if needed.

steffenagger:
@tholhubner yes the fix was release a couple of days ago as part of [realm@10.1.4](https://github.com/realm/realm-js/blob/master/CHANGELOG.md#1014-release-notes-2021-1-27).

So if you are still seeing an app crash during `Realm.open(...)` with the latest version (`v10.1.4` ATTOW), then yes, please open a new issue (with steps to reproduce etc.) - thanks.

PR #3512, opened 2021-01-18
API doc: fix logger example

kneth:
* Fix the logger examples in the API doc
* Update  the JSDoc template with a wider menu (see https://github.com/realm/realm-jsdoc/pull/4)

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -184,15 +184,14 @@ class Sync {
      * but before opening any Realms.
      *
      * @param {Realm.App} app - The Realm app.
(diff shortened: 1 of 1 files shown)


kraenhansen (approved):
It would have been a nicer API to have this directly on the `App` instance I think, but if this match the implementation its good to get the docs updated.

The changes looks reasonable :+1:

kneth:
@kraenhansen 

> it would have been a nicer API to have this directly on the `App` instance

We could at it to `App` and deprecate the existing API (and remove it at next major version).

PR #3515, opened 2021-01-20
Fix Array Buffer for NodeJS

cesarvr:
## What, How & Why?

This commit will fix the binary data type storage for NodeJS. 

This closes # [RJS-1018](https://jira.mongodb.org/browse/RJS-1018).

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


tests/js/binary-data-tests.js
```diff
@@ -0,0 +1,97 @@
...
+        let ab_view = new Uint8Array(data.a)
+
+        for(let i=0; i<view.byteLength; i++ ){
+            let p1 = view[i]
+            let p2 = ab_view[i]
+            TestCase.assertEqual(p1, p2, 'buffers should be equals');
+        }
+    },
+
+    testABNodeBuffer() {
```
kneth, line 78: If the test only works on node, I suggest that you add 

```
if (!isNodeProcess) {
    return;
}
```
at the top. `isNodeProcess` can be defined as `const isNodeProcess = typeof process === 'object' && process + '' === '[object process]';`

tests/js/index.js
```diff
@@ -45,6 +45,8 @@ var TESTS = {
     EncryptionTests: require("./encryption-tests"),
     AliasTests: require("./alias-tests"),
     BsonTests: require("./bson-tests"),
+    ArrayBuffer: require("./binary-data-tests")
```
kneth, line 48: Maybe `ArrayBufferTests` to be aligned with the lines above.

tests/js/index.js
```diff
@@ -45,6 +45,8 @@ var TESTS = {
...
+    //MixedTests: require("./mixed-tests"), comming soon
```
kneth, line 49: Can be removed.

kneth (commented):
The issue mentions that the bug is also onserved in v10.1.3, so I suggest that you target `master` instead of `develop`.

Please add an entry in `CHANGELOG.md` with a link to the Github issue.

src/node/node_value.hpp
```diff
@@ -213,26 +213,21 @@ inline double node::Value::to_number(Napi::Env env, const Napi::Value& value) {
...
-		int64_t byteOffset = value.As<Napi::Object>().Get("byteOffset").As<Napi::Number>();
-		Napi::ArrayBuffer arrayBuffer = value.As<Napi::Object>().Get("buffer").As<Napi::ArrayBuffer>();
-		return make_owned_binary_data(static_cast<char*>(arrayBuffer.Data()) + byteOffset, byteLength);
-	}
-	else {
-		throw std::runtime_error("Can only convert Buffer, ArrayBuffer, and ArrayBufferView objects to binary");
-	}
+    if(value.IsBuffer()) {
+        auto buffer = value.As<Napi::Buffer<char>>();
+        return OwnedBinaryData((const char *)buffer.Data(), buffer.Length() );
```
fronck, line 218: [code style]  In newer versions of C++, it's better to use `static_cast<const char *>` here.
It will help the compiler tell you when something might be unsafe.

Same for the two cases below.

tests/js/array-buffer-tests.js
```diff
@@ -0,0 +1,107 @@
...
+        for(let i=0; i<view.byteLength; i++ ){
+            let p1 = view[i]
+            let p2 = binary_view[i]
+
+            TestCase.assertEqual(p1, p2, 'buffers should be equals');
+        }
+    },
+
+    testABuffer() {
+        if(!isNodeProcess) {
```
fronck, line 57: [question]  As discussed at the meeting, is it intentional that this test is disabled outside node.js?  (As opposed to `testABNodeBuffer`)?

fronck (changes requested):
I'd suggest adhering to the newer C++ standards of typecasting.

Otherwise, thumbs up.  Good catch.

src/node/node_value.hpp
cesarvr: Good Catch 👍

tests/js/array-buffer-tests.js
cesarvr: Yeah that's a lapsus will fix that. Thanks!.

fronck (approved):
Looks good to me 👍

kneth (commented):
:+1: once CI is green :smile:

kneth:
I am closing the PR as we have merged `master` to `develop` instead.

PR #3531, opened 2021-01-27
Rephrasing API doc on client reset

kneth:
## What, How & Why?

The API doc and TS defs for client reset are inaccurate.

Notice: freezing JSdoc to v3.6.3.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * App crashed if a native error was thrown during `Realm.open(...)` ([#3414](https://github.com/realm/realm-js/issues/3414), since v10.0.0)
 * Fixed an issue in Node.js, where utilizing an ArrayBuffer for setting a binary property, would mangle the data. ([#3518](https://github.com/realm/realm-js/issues/3518))
 * Fixed an issue where scripts may hang rather than executing after all code has executed. ([#3525](https://github.com/realm/realm-js/issues/3518))
+* Fixed TS definitions for `Realm.ErrorCallback`.
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * App crashed if a native error was thrown during `Realm.open(...)` ([#3414](https://github.com/realm/realm-js/issues/3414), since v10.0.0)
 * Fixed an issue in Node.js, where utilizing an ArrayBuffer for setting a binary property, would mangle the data. ([#3518](https://github.com/realm/realm-js/issues/3518))
 * Fixed an issue where scripts may hang rather than executing after all code has executed. ([#3525](https://github.com/realm/realm-js/issues/3518))
-* Fixed TS definitions for `Realm.ErrorCallback`.
+* Fixed TS declarations for `Realm.ErrorCallback`.
 
 ### Compatibility
 * MongoDB Realm Cloud.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * App crashed if a native error was thrown during `Realm.open(...)` ([#3414](https://github.com/realm/realm-js/issues/3414), since v10.0.0)
 * Fixed an issue in Node.js, where utilizing an ArrayBuffer for setting a binary property, would mangle the data. ([#3518](https://github.com/realm/realm-js/issues/3518))
 * Fixed an issue where scripts may hang rather than executing after all code has executed. ([#3525](https://github.com/realm/realm-js/issues/3518))
+* Fixed TS definitions for `Realm.ErrorCallback`.
```
steffenagger, line 7: Haven't noticed this before, but it seems we're mixing the wording "TS definitions" & "TS declarations" ([typescriptlang.org](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) uses "declaration(s)") - should we streamline this? (maybe just going forward).

types/index.d.ts
```diff
@@ -343,7 +343,13 @@ declare namespace Realm {
         code: number;
     }
 
-    type ErrorCallback = (session: App.Sync.Session, error: SyncError) => void;
+    interface ClientResetError {
+        name: string;
```
RedBeard0531, line 347: I think you want `name: "ClientReset";` here. Ideally you could make 339 be something like `name: string - "ClientReset";` so that it would know that that type is everything else, but I don't see a way to express that in the TS type system. @kraenhansen any ideas?

Also `readOnly: true`;

docs/sync.js
```diff
@@ -40,7 +40,8 @@
  * @property {Realm.User} user - A {@link Realm.User} object obtained by calling `Realm.App.logIn`.
  * @property {string|number|BSON.ObjectId|null} partitionValue - The value of the partition key.
  * @property {function} [error] - A callback function which is called in error situations.
- *    The `error` callback can take up to five optional arguments: `name`, `message`, `isFatal`,
+ *    The errpr callback can take up to two arguments: `session` and `syncErr`. If `syncError.name == "ClientReset"`, `syncError.path` and `syncError.config` are set
+ *    and `syncError.readOnly` is true. Otherwise, `syncError` can have up to five properties: `name`, `message`, `isFatal`,
  *    `category`, and `code`.
```
steffenagger, line 45: Wording suggestion (incl. typos):
```suggestion
 *    The error callback is called with two arguments: `session` and `syncError`.
 *    If `syncError.name === "ClientReset"`, `syncError` will contain the properties `path`, `config` & `readOnly === true`.
 *    Otherwise, `syncError` contains the properties: `name`, `message`, `isFatal`, `code` & `category` (`category` is optional).
```

(Not sure if this suggestion uphold the char-length, it's a bit hard to see in a gh-suggestion)

types/index.d.ts
```diff
@@ -343,7 +343,13 @@ declare namespace Realm {
...
+        path: string;
+        config: SyncConfiguration;
+    }
```
steffenagger, line 350: I think this would be more correct? And should make it easier to determine if it's a a `ClientResetError` or a generic `SyncError`.
```suggestion
    interface ClientResetError {
        name: "ClientReset";
        path: string;
        config: SyncConfiguration;
        readOnly: true;
    }
```

docs/sync.js
kneth: The generated HTML seems file

CHANGELOG.md
kneth: I'll change it.

kneth:
@steffenagger @RedBeard0531 Thanks for reviewing.

Issue #3536, opened 2021-01-27
JS: A non-empty ArrayBuffer, BufferView or Buffer is expected.


kraenhansen:
## Goals

Setting a non-optional data property to an empty buffer.

## Expected Results & Code Sample

I would expect the following test to pass:

```
const Realm = require("realm");
const realm = new Realm({
  schema: [{
    name: "Test",
    properties: { blob: 'data?' },
  }],
});

realm.write(() => {
  realm.create("Test", { blob: new ArrayBuffer(0) })
});
```

## Actual Results

This error is thrown:

```
Error: A non-empty ArrayBuffer, BufferView or Buffer is expected.
```

## Steps to Reproduce

Save the code above as `index.js` in an empty directory and run

```
npm init -y && npm install realm && node index.js
```

## Version of Realm and Tooling

- Realm JS SDK Version: 10.1.4
- Node or React Native: Node 
- Client OS & Version: v12.20.0

PR #3546, opened 2021-01-29
This should fix the array-buffer problem by inserting a empty-string …

cesarvr:
## What, How & Why?
Fixing a problem with the way we handle empty array-buffer for fields that are nullable. 

This closes https://github.com/realm/realm-js/issues/3536

## ☑️ ToDos

~~* [ ] 📝 Changelog entry~~
~~* [ ] 📝 `Compatibility` label is updated or copied from previous entry~~
* [x] 🚦 Tests
~~* [ ] 📝 Public documentation PR created or is not necessary~~

~~* [ ] typescript definitions file is updated~~
~~* [ ] jsdoc files updated~~
~~* [ ] Chrome debug API is updated if API is available on React Native~~

### DIFF
--- a/tests/js/array-buffer-tests.js
+++ b/tests/js/array-buffer-tests.js
@@ -127,10 +127,13 @@ module.exports = {
             return;
         }
 
-        SingleSchema.properties.a = 'data'
+        SingleSchema.properties.a = 'data?'
         let realm = new Realm({schema: [SingleSchema]})
 
-        TestCase.assertThrowsException(() => realm.write(()=> realm.create(SingleSchema.name, { a:new ArrayBuffer() } )), new Error("A non-empty ArrayBuffer, BufferView or Buffer is expected.") )
+        //should not throw.
+        realm.write(()=> realm.create(SingleSchema.name, { a:new ArrayBuffer() } ))
+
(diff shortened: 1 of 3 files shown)


tests/js/array-buffer-tests.js
```diff
@@ -127,10 +127,13 @@ module.exports = {
...
 
-        SingleSchema.properties.a = 'data'
+        SingleSchema.properties.a = 'data?'
         let realm = new Realm({schema: [SingleSchema]})
 
-        TestCase.assertThrowsException(() => realm.write(()=> realm.create(SingleSchema.name, { a:new ArrayBuffer() } )), new Error("A non-empty ArrayBuffer, BufferView or Buffer is expected.") )
+        //should not throw.
+        realm.write(()=> realm.create(SingleSchema.name, { a:new ArrayBuffer() } ))
+
+       // TestCase.assertThrowsException(() => realm.write(()=> realm.create(SingleSchema.name, { a:new ArrayBuffer() } )), new Error("A non-empty ArrayBuffer, BufferView or Buffer is expected.") )
```
kraenhansen, line 136:
```suggestion
        realm.write(()=> realm.create(SingleSchema.name, { a: new ArrayBuffer() } ));
```

kraenhansen (changes requested):
Looks legit, I just suggest to remove the comments. And remember to add an entry in the changelog (you can use the checkmarks in the description to remember this, instead of ~striking them~ 🙂 )

PR #3549, opened 2021-01-29
Web: Fail app hydration gracefully and forget refresh token on 401s

kraenhansen:
## What, How & Why?

This closes two untracked issues, making it easier to use an app instance if storage gets corrupted.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


packages/realm-web/src/tests/User.test.ts
```diff
@@ -154,6 +173,25 @@ describe("User", () => {
         expect(user.profile).deep.equals({ firstName: "John" });
     });
 
+    it("can refresh access token", async () => {
```
kraenhansen, line 176: This test was in here twice .. looked like a copy-paste error.

packages/realm-web/src/tests/Fetcher.test.ts
```diff
@@ -107,37 +107,4 @@ describe("Fetcher", () => {
             },
         ]);
     });
-
-    it("allows overwriting headers", async () => {
```
kraenhansen, line 111: This test was in here twice .. looked like a copy-paste error.

packages/realm-web/src/App.ts
```diff
@@ -173,7 +173,15 @@ export class App<
...
         // Hydrate the app state from storage
-        this.hydrate();
+        try {
+            this.hydrate();
+        } catch (err) {
+            // The storage was corrupted
+            this.storage.clear();
+            // A failed hydration shouldn't throw and break the app experience
+            // Since this is "just" persisted state that unfortunately got corrupted or partially lost
+            console.warn("Realm app hydration failed:", err.message);
```
kraenhansen, line 183: This is the major change, we're logging a warning instead of throwing an error (that the user cannot do anything about anyway).

packages/realm-web/src/Fetcher.ts
```diff
@@ -269,24 +269,21 @@ export class Fetcher implements LocationUrlContext {
...
             } else {
+                if (
+                    user &&
+                    response.status === 401 &&
+                    tokenType === "refresh"
+                ) {
+                    // A 401 error while using the refresh token indicates the token has an issue.
+                    // Reset the tokens to prevent a lock.
+                    user.accessToken = null;
+                    user.refreshToken = null;
```
kraenhansen, line 285: The only way to resolve a 401 from a broken refresh request is to forget the tokens.

packages/realm-web/src/App.ts
steffenagger: Is the `this.storage.clear()` call needed? It seems like the corrupted value is remove already [AppStorage.getUserIds()](https://github.com/realm/realm-js/blob/master/packages/realm-web/src/AppStorage.ts#L41-L57).

packages/realm-web/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+=============================================================
+
+### Breaking Changes
+* None
+
+### Enhancements
+* None
+
+### Fixed
+* Fixed failures while hydrating an app, such that  ([#3549](https://github.com/realm/realm-js/pull/3549))
```
steffenagger, line 11:
```suggestion
* A more robust implementation of app-hydration. Errors are now logged instead of thrown. ([#3549](https://github.com/realm/realm-js/pull/3549))
```

packages/realm-web/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+* Fixed a lock happening if refresh access token or custom user data resulted failed. Tokens is now forgotten any request using the refresh token results in a 401 response. ([#3549](https://github.com/realm/realm-js/pull/3549))
```
steffenagger, line 12: This sentence is not clear (I'm not sure I made it that much better...).
```suggestion
* Fixed a potential lock, when refreshing an access token or fetching custom user data, failed due to 401/unauthorized. An unauthorized response, during token-refresh, will now clear all the users' tokens. ([#3549](https://github.com/realm/realm-js/pull/3549))
```

packages/realm-web/src/App.ts
kraenhansen: That only clears the list of user ids, not the storage related to the entire app.

kraenhansen: I believe it is, this is more broad as it will handle any corruption to the storage, including errors thrown while dehydrating a user instance: https://github.com/realm/realm-js/blob/master/packages/realm-web/src/User.ts#L134-L142. The error I was specifically trying to address is the `providerType` property which was added recently and users authenticated with a previous version of the SDK won't be able to be dehydrated from storage.

I could however, most likely remove the `catch` and `this.remove(USER_IDS_STORAGE_KEY);` from `AppStorage.getUserIds()` 🤔

kraenhansen: I removed the `this.remove` from `AppStorage.getUserIds`.

steffenagger (approved):

PR #3550, opened 2021-01-31
Fix type definition of sync in configuration

jie17:
## What, How & Why?
The type of `sync` in `Configuration` should be `boolean | SyncConfiguration`

- [x] typescript definitions file is updated

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -121,7 +121,7 @@ declare namespace Realm {
         inMemory?: boolean;
         schema?: (ObjectClass | ObjectSchema)[];
         schemaVersion?: number;
-        sync?: SyncConfiguration;
+        sync?: true | SyncConfiguration;
         deleteRealmIfMigrationNeeded?: boolean;
         disableFormatUpgrade?: boolean;
     }


types/index.d.ts
```diff
@@ -121,7 +121,7 @@ declare namespace Realm {
         inMemory?: boolean;
         schema?: (ObjectClass | ObjectSchema)[];
         schemaVersion?: number;
-        sync?: SyncConfiguration;
+        sync?: boolean | SyncConfiguration;
```
kraenhansen, line 124: Technically `false` shouldn't be a valid value.

```suggestion
        sync?: true | SyncConfiguration;
```

kraenhansen:
We've been reluctant to add this as an officially supported part of the API and with this change.
If we merge this, we should also update [the corresponding JSDoc](https://github.com/realm/realm-js/blob/master/docs/realm.js#L388).

jie17:
Thank you for reviewing. So it's kind of a private API. I'm wondering if there is a better way to support both main and rendering processes in Electron applications because currently it is recommended in the docs.

PR #3556, opened 2021-02-04
Merge logger

kneth:
Merging #3408 to `master`


src/logger.hpp
```diff
@@ -0,0 +1,171 @@
...
+    void delegate(Delegated& delegate)
+    {
+        m_scheduler->set_notify_callback([this, delegate] {
+            while (!m_log_queue.empty()) {
+                /*
+                 *  Extracting and delegating log entries.
+                 */
+                Entry entry;
+                {
+                    std::lock_guard<std::mutex> lock(m_mutex);
```
blagoev, line 83: this is prone to lock contention. locking too much in order retrieve every single log entry.
Instead we can swap the `m_log_queue` backing store like was implemented previously thus no locking is required when calling the log callback
https://github.com/realm/realm-js/blob/v6/src/node/sync_logger.cpp#L75-L79

blagoev (changes requested):

src/logger.hpp
kneth: The code has been updated

src/logger.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+class SyncLoggerDelegator : public realm::util::RootLogger {
+   public:
+    void delegate(Delegated& delegate) {
+        m_scheduler->set_notify_callback([this, delegate] {
+            while (!m_log_queue.empty()) {
+                /*
+                 *  Extracting and delegating log entries.
+                 */
+                std::queue<Entry> popped;
```
blagoev, line 84: the swap should be done outside the while loop. then the iteration should be over the `popped` queue.

blagoev (changes requested):

blagoev:
this was merged with a Red flag. The correct fix is here https://github.com/realm/realm-js/pull/3565

PR #3560, opened 2021-02-05
Clarified `Realm.App.removeUser()` method description in typedoc

mongomoe:
## What, How & Why?
Following a MongoDB slack [discussion](https://mongodb.slack.com/archives/CUSNYAJ4R/p1612473658124000), I updated the description to the method `Realm.App.removeUser()`. In the master branch this method has the description, "Removes the user from MongoDB Realm". This can be confusing because readers may assume the user is deleted from the MongoDB Realm Server as well as the device. I altered the description in this PR  to be "Removes the user from the client". This clarifies that the user is only removed from the user's device.

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -125,7 +125,7 @@
     switchUser(user) { }
 
     /**
-     * Removes the user from MongoDB Realm.
+     * Removes the user from the client.
      *
      * @param {Realm.User} user - The user to remove.
      * @returns {Promise<void>}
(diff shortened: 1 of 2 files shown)


kraenhansen:
We're also publishing Typescript types with the package: https://github.com/realm/realm-js/blob/0d6b98fe0e25404a9dc77553a5819e916be3bb0c/types/app.d.ts#L345 perhaps it makes sense to align these as well?

> Logs out and removes a user from the app.

Here "app" might also be a bit ambiguous.

kraenhansen (approved):
Looks great - thanks 👍

PR #3584, opened 2021-02-12
[v6 hotfix] Upcoming webkit update correctly checks "configurable" in property-descriptors 

### LINKED ISSUE #3557: Crash on iOS 14.5 beta: Attempting to change configurable attribute of unconfigurable property
## Expected Results

Not to crash.

## Actual Results

Crashed on iOS 14.5 beta. All other iOS versions are fine.

If we don't use Realm in JS, then everything is fine. But if we add one line to import Realm:

`import Realm from 'realm';`

It crashed with the following error: Attempting to change configurable attribute of unconfigurable property.

And crashed when running the following Object.defineProperties() in `lib/extensions.js` of realm-js:

```
Object.defineProperties(realmConstructor, getOwnPropertyDescriptors({
    _extendQueryBasedSchema(schema) {
        addSchemaIfNeeded(schema, realmConstructor.Permissions.Class);
        addSchemaIfNeeded(schema, realmConstructor.Permissions.Permission);
        addSchemaIfNeeded(schema, realmConstructor.Permissions.Realm);
        addSchemaIfNeeded(schema, realmConstructor.Permissions.Role);
        addSchemaIfNeeded(schema, realmConstructor.Permissions.User);
        addSchemaIfNeeded(schema, realmConstructor.Subscription.ResultSets);
    },

    // Creates the user agent description for the JS binding itself. Users must specify the application
    // user agent using Realm.Sync.setUserAgent(...)
    _createUserAgentDescription() {
        // Detect if in ReactNative (running on a phone) or in a Node.js environment
        // Credit: https://stackoverflow.com/questions/39468022/how-do-i-know-if-my-code-is-running-as-react-native
        try {
            var userAgent = "RealmJS/";
            userAgent = userAgent + require('../package.json').version + " (" + context + ", ";
            if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
                // Running on ReactNative
                const Platform = require('react-native').Platform;
                userAgent += Platform.OS + ", v" + Platform.Version;
            } else {
                // Running on a normal machine
                userAgent += process.version;
            }
            return userAgent += ")";
        } catch (e) {
            return "RealmJS/Unknown"
        }
    },
}));
```

Maybe due to some change in iOS 14.5 JavaScript core?

Simulator Screen Shot - iPhone 8 - 2021-02-04 at 17 03 48

## Steps to Reproduce

1.Setup a RN project and import realm-js
2.Run the App on iOS 14.5 beta.

Not sure whether it's related to my RN version (0.59). But it all worked fine until the recently released iOS 14.5 beta.
Tried Relam JS 5.0.5 and 6.1.5, both has the same issue.

## Code Sample

`import Realm from 'realm';`

- Realm JS SDK Version: 5.0.5 and 6.1.5
- Node or React Native: RN 0.59.10
- Client OS & Version: iOS 14.5 beta

steffenagger:
## What, How & Why?

This webkit change is currently only encountered through the bundled simulators in Xcode 12.5 beta, where the app will break with the following error: `"Attempting to change configurable attribute of unconfigurable property"`

As the property descriptor `configurable` has previously been ignored by webkit, we now avoid setting it.

This closes #3557

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 * Fixed an issue where creating an object after file format upgrade may fail with error message `Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()`. ([realm/realm-core#4295](https://github.com/realm/realm-core/issues/4295), since v6.0.0)
+* Due to an upcoming WebKit update (currently accessible through Xcode beta simulators), apps would throw `"Attempting to change configurable attribute of unconfigurable property"` at runtime ([#3557](https://github.com/realm/realm-js/issues/3557))
 
 ### Compatibility
 * Realm Object Server: 3.23.1 or later.
--- a/src/jsc/jsc_class.hpp
+++ b/src/jsc/jsc_class.hpp
@@ -325,7 +325,7 @@ inline std::vector<JSStaticFunction> ObjectWrap<ClassType>::get_methods(const Me
     std::vector<JSStaticFunction> functions;
     functions.reserve(methods.size() + 1);
 
-    JSPropertyAttributes attributes = kJSPropertyAttributeReadOnly | kJSPropertyAttributeDontEnum | kJSPropertyAttributeDontDelete;
+    JSPropertyAttributes attributes = kJSPropertyAttributeReadOnly | kJSPropertyAttributeDontEnum;
     size_t index = 0;
 
     for (auto &pair : methods) {


takameyer (approved):

kraenhansen (approved):
It would be nicer to fix the root cause of the error (that we're overriding the properties from JS): I would love if we created an issue tracking that technical debt as well.

kneth (approved):

steffenagger:
> It would be nicer to fix the root cause of the error (that we're overriding the properties from JS): I would love if we created an issue tracking that technical debt as well.

@kraenhansen agreed, but we're not doing this in `v10`, and now iOS/WebKit wont let us, going forward.
Though there is definitely a point to be made, and example here, for testing with upcoming versions as well (GH actions exists to allow for this: https://github.com/marketplace/actions/setup-xcode-version)

PR #3596, opened 2021-02-17
Support for http proxy

kneth:


fealebenpae:
Unfortunately the Sync client only supports HTTP proxies (we stubbed out HTTPS support, but never got around to implementing it). I suggest sticking with HTTP only for now.

PR #3597, opened 2021-02-18
avoid instantiating extra schema on instance creation

finnschiermer:
Fixes a performance bug.

We unintentionally created (and subsequently destroyed) an object schema for each object instance created. For larger schema instances this has substantial overhead.

Perhaps also illustrates why Linus Thorvalds prefers C over C++ for performance sensitive work.

## ☑️ ToDos

* [ ] 📝 Changelog entry

### DIFF
--- a/src/js_realm_object.hpp
+++ b/src/js_realm_object.hpp
@@ -131,8 +131,8 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
     static String prototype_string = "prototype";
 
     auto delegate = get_delegate<T>(realm_object.realm().get());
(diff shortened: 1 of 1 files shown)


simonask (approved):
~~Should this PR target `master`?~~

Never mind, I don't know how realm-js prefers to do things.

kraenhansen:
> Should this PR target master?

I think you're right, we would want this on master first and then cherry-picked (or via a separate PR) into `develop`.

The `develop` branch is currently used for the work related to new datatypes (mixed, map, set, dict, etc.).

finnschiermer:
I'll open a new one against master then.

PR #3598, opened 2021-02-18
avoid creating (and destroying) schema for each object

finnschiermer:
Fixes a performance bug.

We unintentionally created (and subsequently destroyed) an object schema for each object instance created. For larger schema instances this has substantial overhead.

Perhaps also illustrates why Linus Thorvalds prefers C over C++ for performance sensitive work.

Replaces https://github.com/realm/realm-js/pull/3597

## ☑️ ToDos
* [ ] 📝 Changelog entry

### DIFF
--- a/src/js_realm_object.hpp
+++ b/src/js_realm_object.hpp
@@ -131,8 +131,8 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
     static String prototype_string = "prototype";
 
     auto delegate = get_delegate<T>(realm_object.realm().get());
-    auto schema = realm_object.get_object_schema();
-    auto name = schema.name;
+    auto& schema = realm_object.get_object_schema();
+    auto& name = schema.name;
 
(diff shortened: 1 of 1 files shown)


blagoev (approved):

kraenhansen (approved):
Great find! It would be great if we could include a measurement of the (potential) performance increase in the changelog 👍

finnschiermer:
@kraenhansen the impact depends on the size of the schema. Creating a Realm object by reading it from the realm is roughly constant overhead. Creating a schema, however, depends on the size of that schema. I've got a case where half the overhead of the JS SDK was in copying schemas.

bmunkholm:
@blagoev Please don't merge without release note updates.

blagoev:
yes, sure, my bad. I was helping @finnschiermer who apparently does not have rights to merge in rjs repo. 
@finnschiermer could you please add some internal improvements notes in the changelog.md

PR #3609, opened 2021-02-23
Andrew/fix list push segfault

takameyer:
## What, How & Why?
```
    constructor(description) {
        super()
        this.id = `${new ObjectId()}`
        this.done = false;
    }

    static schema = {
        name: "TodoItem",
        properties: {
            id: "string",
            description: "string",
            deadline: "date?"
        },
        primaryKey: "id"
    }
}

    constructor(name ) {
        super()
        this.id = `${new ObjectId()}`
        this.name = name;
        this.items = []
    }

    static schema = {
        name: "TodoList",
        properties: {
            id: "string",
            name: "string",
            items: "TodoItem[]"
        },
        primaryKey: "id"
    }
}

realm.write(() => {

})
```

The problem is that realm was assuming that if the object is of type Realm.Object that it has already been created.  This assumption was causing a null pointer to be dereferenced which had caused the segmentation fault.  The logic now checks for null and falls through to creating the object if it has not been created.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


src/js_object_accessor.hpp
```diff
@@ -428,13 +428,16 @@ struct Unbox<JSEngine, Obj> {
...
         if (js::Object<JSEngine>::template is_instance<RealmObjectClass<JSEngine>>(ctx->m_ctx, object)) {
             auto realm_object = get_internal<JSEngine, RealmObjectClass<JSEngine>>(ctx->m_ctx, object);
-            if (realm_object->realm() == ctx->m_realm) {
-                return realm_object->obj();
-            }
-
-            bool updating = policy.copy && policy.update;
-            if (!updating && !policy.create) {
-                throw std::runtime_error("Realm object is from another Realm");
+            if (realm_object){
```
fronck, line 431:
```suggestion
            if (realm_object) {
```

src/js_object_accessor.hpp
```diff
@@ -428,13 +428,16 @@ struct Unbox<JSEngine, Obj> {
...
+                if (realm_object->realm() == ctx->m_realm) {
+                    return realm_object->obj();
+                }
+                else{
```
fronck, line 435:
```suggestion
                else {
```

takameyer:
Closing and making a new branch based on master instead of develop.

Issue #2469, opened 2019-08-01
Typescript Initialisation of List properties


thekevinbrown:
## Goals
Say I've got a schema like this (in Typescript):

```typescript
export class Stuff {
	static schema: ObjectSchema = {
		name: 'Stuff',
		primaryKey: 'id',
		properties: {
			id: 'string',
			text: 'string[]',
		},
	};

	id: string;
	text: List<string>;
}
```

This is a compiler error because `id` and `text` aren't optional and aren't instantiated in the constructor.

But in this case, what would the constructor actually look like? This is the best I can guess:

```typescript
constructor() {
	this.id = 'item-1';
	this.text = new List<string>();
}
```

but

```
Cannot use 'new' with an expression whose type lacks a call or construct signature.ts(2351)
```

## Expected Results
How do we correctly initialise objects with required lists so the Typescript types match up to the actual object? The goal is to be able to do this:

```typescript
const stuff = new Stuff();
stuff.text.push('text here');
```

so if we don't initialise `text` with anything then it won't be there and we'll be calling `push` on `undefined`.

What should the constructor for this model look like?

## Actual Results
Can't figure out how to correctly construct an object with Linking Objects or List property.

## Related Stack Overflow
I didn't ask this question on stack overflow because [this question is already there and hasn't been answered for a month](https://stackoverflow.com/questions/56700854/realm-object-relationships-using-typescript) that didn't get an answer.

## Version of Realm and Tooling
* Realm JS SDK Version: `2.29.1`
* Typescript Version: `3.5.3`
* Node or React Native: Doesn't matter
* Client OS & Version: Doesn't matter
* Which debugger for React Native: Doesn't matter

kneth:
I don't think our TS support is so well-developed.

Maybe you can get some ideas from https://stackoverflow.com/questions/46558215/how-to-resolve-error-ts2351-cannot-use-new-with-an-expression-whose-type-la/46558897.

thekevinbrown:
This doesn’t have anything to do with the import statement unfortunately.

nirinchev:
Can you try marking them as public? Or is it possible you're instructing the compiler to be overly aggressive? In one of our projects, we're using classes with non-optional properties without a constructor and that works just fine. It looks something like:

```ts

export class PermissionOffer extends BaseRealmClass {
    public static schema: Realm.ObjectSchema = {
        name: "PermissionOffer",
        primaryKey: "id",
        properties: {
            id: "string",
[...]

PR #3610, opened 2021-02-23
Fix segmentation fault when pushing a class object based on Realm.Object to a list

takameyer:
## What, How & Why?
```
    constructor(description) {
        super()
        this.id = `${new ObjectId()}`
        this.done = false;
    }

    static schema = {
        name: "TodoItem",
        properties: {
            id: "string",
            description: "string",
            deadline: "date?"
        },
        primaryKey: "id"
    }
}

    constructor(name ) {
        super()
        this.id = `${new ObjectId()}`
        this.name = name;
        this.items = []
    }

    static schema = {
        name: "TodoList",
        properties: {
            id: "string",
            name: "string",
            items: "TodoItem[]"
        },
        primaryKey: "id"
    }
}

realm.write(() => {

})
```

The problem is that realm was assuming that if the object is of type Realm.Object that it has already been created.  This assumption was causing a null pointer to be dereferenced which had caused the segmentation fault.  The logic now checks for null and throws an exception.
    
Also attempting to create an object with an unmanged instance of Realm.Object will throw an exception. This is to avoid future attempts to create which fail, as the prototype has been changed.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] ~📝 `Compatibility` label is updated or copied from previous entry~
* [x] 🚦 Tests

* [ ] ~jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* There seems to be a few issues regarding class support in realm-js. We are currently coming up with strategies to better support this in the future.  In the meantime, the following fixes have been applied to help avoid crashes and failures.
+  * When creating a class that extends Realm.Object and pushing the instantiated object to a list, a segmentation fault would occur.  This has been fixed by a null check and throwing an exception.
+  * Creating an object from an instance of Realm.Object that was manually constructed (detached from Realm) would fail the second time.  Now we throw an meaningful exception the first time.
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/src/js_object_accessor.hpp
+++ b/src/js_object_accessor.hpp
@@ -360,6 +360,9 @@ struct Unbox<JSEngine, Obj> {
         auto object = Value::validated_to_object(ctx->m_ctx, value);
         if (js::Object<JSEngine>::template is_instance<RealmObjectClass<JSEngine>>(ctx->m_ctx, object)) {
             auto realm_object = get_internal<JSEngine, RealmObjectClass<JSEngine>>(ctx->m_ctx, object);
+            if (!realm_object) {
+                throw std::runtime_error("Cannot reference a detached instance of Realm.Object");
+            }
             if (realm_object->realm() == ctx->m_realm) {
                 return realm_object->obj();
             }
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -1026,6 +1026,13 @@ void RealmClass<T>::create(ContextType ctx, ObjectType this_object, Arguments &a
         object = Schema<T>::dict_for_property_array(ctx, object_schema, object);
     }
 
+    if (Object::template is_instance<RealmObjectClass<T>>(ctx, object)) {
+        auto realm_object = get_internal<T, RealmObjectClass<T>>(ctx, object);
+        if (!realm_object) {
+            throw std::runtime_error("Cannot create an object from a detached Realm.Object instance");
+        }
+    }
+
     NativeAccessor accessor(ctx, realm, object_schema);
     auto realm_object = realm::Object::create<ValueType>(accessor, realm, object_schema, object, policy);
     return_value.set(RealmObjectClass<T>::create_instance(ctx, std::move(realm_object)));
--- a/tests/js/list-tests.js
+++ b/tests/js/list-tests.js
@@ -1601,4 +1601,61 @@ module.exports = {
         });
         realm.close();
     },
+
+    testClassObjectCreation: function() {
+        class TodoItem extends Realm.Object {
+            constructor(description) {
+                super()
+                this.id = new ObjectId()
+                this.description = description;
+                this.done = false;
+            }
+        }
+
+        TodoItem.schema =  {
+            name: "TodoItem",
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* When creating a class that extends Realm.Object and pushing the instantiated object to a list, a segmentation fault would occur.  This has been fixed by a null check and afterwards creating the object.
```
kneth, line 8: Do it also solve https://github.com/realm/realm-js/issues/2469? If so, we should add a link to it.

kneth (approved):

kraenhansen:
Let's have a face 2 face conversation around this, as I don't think this is the right approach.
I agree that passing an instance of a class inheriting from `Realm.Object` shouldn't crash the app.
At the same time, I think it should throw a meaningful error instead of allowing object creation from this.

In my understanding of the current implementation (and limitations) of our "class-based schema" support, we do not support users to explicitly call the schema class's constructor, hence it's undefined behaviour and not something we should be providing work arounds for. I also agree that we should continue improving the class based support, ideally to the point where explicitly calling the class constructor is supported, but this requires more work than this PR suggests.

CHANGELOG.md
takameyer: Unfortunately it does not.

fronck (approved):
LGTM

cesarvr (approved):
LGTM

takameyer:
@kraenhansen and I decided it would make more sense to throw an exception instead of implicitly creating the object on push.  We also added an exception when someone tries to use the class constructor to create an object, as this breaks the second time.

In the future we will rethink class support.

kneth (approved):

kraenhansen (approved):

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* When creating a class that extends Realm.Object and pushing the instantiated object to a list, a segmentation fault would occur.  This has been fixed by a null check and throwing an exception.
+* Creating an object from an instance of Realm.Object that was manually constructed (detached from Realm) would fail the second time.  Now we throw an meaningful exception the first time.
```
steffenagger, line 9: Should we mention that these are temporary solutions?

steffenagger (approved):
It's definitely better to have consistent behaviour & exceptions instead of crashes, but should we word this as temporary fixes?

CHANGELOG.md
takameyer: That's a valid point

takameyer: I added a parent bullet point to explain our current situation and that these are temporary fixes to avoid crashes.

src/js_object_accessor.hpp
```diff
@@ -360,6 +360,9 @@ struct Unbox<JSEngine, Obj> {
         auto object = Value::validated_to_object(ctx->m_ctx, value);
         if (js::Object<JSEngine>::template is_instance<RealmObjectClass<JSEngine>>(ctx->m_ctx, object)) {
             auto realm_object = get_internal<JSEngine, RealmObjectClass<JSEngine>>(ctx->m_ctx, object);
+            if(!realm_object){
```
fronck, line 363:
```suggestion
            if (!realm_object) {
```

src/js_realm.hpp
```diff
@@ -1026,6 +1026,13 @@ void RealmClass<T>::create(ContextType ctx, ObjectType this_object, Arguments &a
         object = Schema<T>::dict_for_property_array(ctx, object_schema, object);
     }
 
+    if (Object::template is_instance<RealmObjectClass<T>>(ctx, object)) {
+        auto realm_object = get_internal<T, RealmObjectClass<T>>(ctx, object);
+        if(!realm_object){
```
fronck, line 1031:
```suggestion
        if (!realm_object) {
```

tests/js/list-tests.js
```diff
@@ -1601,4 +1601,61 @@ module.exports = {
...
+        class TodoItem extends Realm.Object {
+            constructor(description) {
+                super()
+                this.id = new ObjectId()
+                this.description = description;
+                this.done = false;
+            }
+        }
+
+        TodoItem.schema =  {
```
fronck, line 1615: The tests usually put their schemas in `schemas.js`..  Is there any specific reason why these don't?

fronck (approved):
Just a few nitpicks and a general question.

Other than that, LGTM

tests/js/list-tests.js
takameyer: Well I would perhaps like to have a conversation about this.  I find in tests it is best to not put such things in separate files, as they rarely are reused and it makes the reader of the test have a nicer time following what is happening.  So I guess that's more my personal style and opinion.

fronck: I don't have a strong opinion on the subject myself..
@kneth ?

kneth: We are using the same schemas across tests, so we avoid code duplication with `schemas.js`.

takameyer: We are going to create a new issue for this where we will address refactor the tests and write a style guide.

PR #3619, opened 2021-03-01
Use monorepo releases

kneth:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


src/js_sync.hpp
```diff
@@ -780,7 +780,7 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
     if (Value::is_boolean(ctx, sync_config_value)) {
         config.force_sync_history = Value::to_boolean(ctx, sync_config_value);
         if (config.force_sync_history) {
-            config.schema_mode = SchemaMode::Additive;
+            config.schema_mode = SchemaMode::AdditiveDiscovered;
```
tgoyne, line 783: JS probably wants AdditiveExplicit. AdditiveDiscovered is for when the schema is auto-discovered from classes so that merely defining an unused embeddedobject subclass doesn't break things.

tests/js/object-tests.js
```diff
@@ -567,6 +568,7 @@ module.exports = {
 
         obj.addListener((obj, changes) => {
             calls++;
+            console.log(`FISK ${calls}: ${JSON.stringify(changes)}`);
```
tgoyne, line 571: I don't know if it's the source of problems, but one thing I notice about this test is that it seems like `obj` could be garbage collected immediately after the call to `realm.delete()` which would result in the notifier being destroyed.

kneth:
#3635 is the way forward (building everything from source).

PR #3624, opened 2021-03-04
Fix segmentation fault when pushing a class object based on Realm.Object to a list [v6 backport]

takameyer:
## What, How & Why?
```
    constructor(description) {
        super()
        this.id = `${new ObjectId()}`
        this.done = false;
    }

    static schema = {
        name: "TodoItem",
        properties: {
            id: "string",
            description: "string",
            deadline: "date?"
        },
        primaryKey: "id"
    }
}

    constructor(name ) {
        super()
        this.id = `${new ObjectId()}`
        this.name = name;
        this.items = []
    }

    static schema = {
        name: "TodoList",
        properties: {
            id: "string",
            name: "string",
            items: "TodoItem[]"
        },
        primaryKey: "id"
    }
}

realm.write(() => {

})
```

The problem is that realm was assuming that if the object is of type Realm.Object that it has already been created.  This assumption was causing a null pointer to be dereferenced which had caused the segmentation fault.  The logic now checks for null and throws an exception.
    
Also attempting to create an object with an unmanged instance of Realm.Object will throw an exception. This is to avoid future attempts to create which fail, as the prototype has been changed.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] ~📝 `Compatibility` label is updated or copied from previous entry~
* [x] 🚦 Tests

* [ ] ~jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,25 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
+* There seems to be a few issues regarding class support in realm-js. We are currently coming up with strategies to better support this in the future.  In the meantime, the following fixes have been applied to help avoid crashes and failures.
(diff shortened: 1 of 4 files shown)


fronck (approved):
LGTM :+1:

PR #3626, opened 2021-03-05
Update scripts in package.json

kraenhansen:
## What, How & Why?

Calling `cmake-js rebuild` on install makes it impossible to install from a `npm pack`'ed version into another package, since `cmake-js` is a `devDependency`.

Removing `cd tests && npm install` since we should assume a developer has already installed the `test` directory before running the tests.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests


takameyer (approved):

package.json
```diff
@@ -50,17 +50,16 @@
...
     "rebuild": "npm run rebuild-changes",
     "build": "npm run build-changes",
     "build-changes": "cmake-js build --debug",
-    "rebuild-changes": "cmake-js rebuild --debug && cd tests && npm install",
+    "rebuild-changes": "cmake-js rebuild --debug",
     "eslint": "npm install --ignore-scripts && npm run lint .",
     "license-check": "npm install && license-checker --exclude \"MIT,ISC,BSD,Apache-2.0,BSD-2-Clause,BSD-3-Clause,WTFPL,Unlicense,(MIT AND CC-BY-3.0)\" | node scripts/handle-license-check.js",
     "jsdoc:clean": "rimraf ./docs/output",
     "jsdoc": "npm install --ignore-scripts && npm run jsdoc:clean && jsdoc -u docs/tutorials -p package.json -c docs/conf.json",
-    "prenode-tests": "npm install --build-from-source && cd tests && npm install",
```
steffenagger, line 62: So now you'd have to `(cd test && npm i)`, before calling `npm run node-tests`..
Does our CI know that?

steffenagger (commented):
I agree that a `rebuild` script, shouldn't install dependencies some something unrelated... but a bit unsure about removing the pre-hook to `node-tests`.

And something unrelated to these changes: If we have a `node-tests` script, why don't vi have `android-tests`, `ios-tests` & `electron-tests`?

package.json
kraenhansen: Seem like the tests are passing, they didn't use this script.

kraenhansen:
I'm a bit in doubt if we want to move the `cmake-js` package to become a production dependency instead of simply a development dependency.

package.json
steffenagger: Valid point 🤦‍♂️

steffenagger:
> I'm a bit in doubt if we want to move the `cmake-js` package to become a production dependency instead of simply a development dependency.

Yeah, how does fallback to "build from source" currently work?

takameyer:
> I'm a bit in doubt if we want to move the cmake-js package to become a production dependency instead of simply a development dependency.

Is it possible to install the dev-dependencies on fallback (including cmake-js) and then start building?

kraenhansen:
> > I'm a bit in doubt if we want to move the `cmake-js` package to become a production dependency instead of simply a development dependency.
> 
> Yeah, how does fallback to "build from source" currently work?

It doesn't 😢  If the prebuilt binary isn't found, it simply fails due to the missing `cmake-js` dependency .. but then again, even with that dependency, a user is also required to have `cmake` installed on their system. So, perhaps we should wrap this in something that tries to run the `cmake-js` command and fails more gracefully with instructions on how to install the proper dependencies.

steffenagger:
> Perhaps we should wrap this in something that tries to run the `cmake-js` command and fails more gracefully with instructions on how to install the proper dependencies.

This sounds like a good approach to me.

fealebenpae:
I don't think that the case where someone installs the `realm` package in their project, or a package that depends on `realm`, and they need to compile from source should be a supported one. For all our supported Node.js platforms we should have prebuilt binaries on our CDN for `prebuilt-install` to fetch. If it's an unreleased package or when building from source (no binaries for Node.js were uploaded to our CDN) then we can just say that `cmake-js` and `CMake` itself are implicit dependencies that the user should install on their own.

kraenhansen:
I'll close this for now and open another PR with an attempt to clean up the scripts at a later point (preferably when / if we move our `realm` package into `/packages`).

PR #3631, opened 2021-03-08
Dictionaries V1

cesarvr:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


src/common/deleteme/collection.hpp
```diff
@@ -0,0 +1,8 @@
+//
```
fronck, line 1: [Question]  This file is in a directory called "deleteme".  Is it supposed to be checked in?

src/node/node_type_deduction.hpp
```diff
@@ -52,14 +70,19 @@ struct TypeDeductionImpl {
         return TypeDeductionImpl::is_bson_type(value, "ObjectID");
     }
 
-    static std::string to_javascript(types::Type value)
+    static bool realm_type_exist(std::string type){
```
fronck, line 73:
```suggestion
    static bool realm_type_exist(std::string const &type){
```

src/node/node_type_deduction.hpp
```diff
@@ -52,14 +70,19 @@ struct TypeDeductionImpl {
...
+        static auto realm_types = reverse_deduction_types_map();
+        return realm_types.find(type) == realm_types.end();
+    }
+
+    static types::Type realm_type(std::string type){
```
fronck, line 78:
```suggestion
    static types::Type realm_type(std::string const &type){
```

src/node/node_type_deduction.hpp
fronck: In general, it's good practice to `const` function arguments when possible.  It allows calling functions with `const` arguments to call this function.  It allows you to pass by reference, possibly avoiding a copy.

src/test/js_dictionary.cpp
```diff
@@ -0,0 +1,11 @@
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+#include <vector>
+#include "js_dictionary.hpp"
+
+using Catch::Matchers::Contains;
+
+TEST_CASE( "Testing DictionarySchema#Constructor" ) {
+    realm::js::DictionarySchema schm {"string{}"};
```
fronck, line 9: [Questions]  Is the postfix for `dictionary` not supposed to be `<>`?  I'm using `{}` for `set` at the moment.

src/test/test.sh
```diff
@@ -0,0 +1,10 @@
+g++ -std=c++17 -Wall -I catch.hpp -I ../ -I ../../vendor/realm-core/src/realm/object-store/ \
```
fronck, line 1: Please add some comments about what this script does.  It's not obvious to me at least.
Also, should it be moved to the main `scripts/test.sh` as an parameter?

src/test/test.sh
```diff
@@ -0,0 +1,10 @@
...
+    -I ../../vendor/realm-core/src/ \
+    -o testing js_dictionary.cpp \
+    core/src/realm/librealm-dbg.a \
+    core/src/realm/sync/librealm-server-dbg.a \
+    core/src/realm/sync/librealm-sync-dbg.a \
+    && 
+
+./testing  
+# missing realm::ConstLstIf
```
fronck, line 10: [Question]  Is this comment relevant here?

src/node/node_type_deduction.hpp
```diff
@@ -25,8 +29,22 @@
...
+class TypeDeductionImpl {
+private:
     using Value = const Napi::Value;
+    static std::map<types::Type, std::string> realm_types;
+
+    static auto reverse_deduction_types_map(){
+        std::map<std::string, types::Type> ret;
+        for(auto& [type, key] : realm_types ){
+            ret[key] = type;
+            std::transform(key.begin(), key.end(), key.begin(), [](unsigned char c){ return std::tolower(c); });
```
fronck, line 41:
```suggestion
            std::transform(key.begin(), key.end(), key.begin(), [](static_cast<unsigned char>(c)){ return std::tolower(c); });
```

src/js_schema.hpp
```diff
@@ -85,6 +85,13 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
         type = type.substr(0, type.size() - 1);
     }
 
+    DictionarySchema dictionary {type};
+
+    if(dictionary.is_dictionary()){
+        prop.type |= dictionary.schema();
+        return;
+    }
```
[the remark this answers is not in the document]
fronck: [Suggestion]  Adding a guard here to make sure that `DictionarySchema dictionary {type}` isn't executed on every invocation of `parse_property_type`.  It's a rather costly operation..

fronck (changes requested):
:+1:, but a few comments for your consideration.

src/test/js_dictionary.cpp
kneth: Our dictionaries look very much like plain JavaScript objects (key is a string). For example `{ name: "Arthur", age: 42 }` - so maybe `{}` is a better fit for dictionaries.

src/dictionary/methods/listeners.hpp
```diff
@@ -0,0 +1,77 @@
...
+        add_javascript_function("addListener", add_listener(object), object);
+        add_javascript_function("removeListener", remove_listener(object),
+                                object);
+        add_javascript_function("removeAllListeners",
+                                remove_all_listeners(object), object);
+    }
+
+    template <typename JavascriptPlainObject>
+    auto add_listener(JavascriptPlainObject* object) {
+        return [=](const Napi::CallbackInfo& info) {
```
kneth, line 42: By using `js::Function<VM>` you can support any engine.

src/dictionary/methods/callbacks.hpp
```diff
@@ -0,0 +1,76 @@
+//
+// Created by Cesar Valdez on 07/03/2021.
```
bmunkholm, line 2: replace w standard header

src/dictionary/methods/listeners.hpp
```diff
@@ -0,0 +1,77 @@
+//
+// Created by Cesar Valdez on 07/03/2021.
```
bmunkholm, line 2: replace with standard header.

src/js_schema.hpp
```diff
@@ -85,6 +85,13 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
         type = type.substr(0, type.size() - 1);
     }
 
+    DictionarySchema dictionary {type};
+
+    if(dictionary.is_dictionary()){
```
bmunkholm, line 90: Use spacing similar to the rest of the code. After "if", before "{".

src/dictionary/methods/callbacks.hpp
cesarvr: My name looks nice there, was joking good catch.

src/dictionary/methods/listeners.hpp
cesarvr: I would like to do it, but that [Napi::CallbackInfo](https://github.com/nodejs/node-addon-api/blob/main/doc/callbackinfo.md) is not a function, is some kind of object injected into functions. I would suggest waiting for the port to JSC and we re-assess.

fronck (approved):

src/js_schema.hpp
kneth: Please ignore code formatting - we plan to apply `clang-format` using Realm Core's configuration when we merge `develop` to `master` (most likely: immediately after the merge).

cesarvr:
# Added Support For JavaScript Core (JSC)

## Added
- New methods remove/put. This way users can update / remove keys from their dictionary.
- Object mutation to reflect the contents of the dict.
- Added support for JavaScript Core API (react-native).
- Added Catch2 framework (simple implementation using just the headers).
- Added a [test bed to run our C++ test](https://github.com/realm/realm-js/blob/5264c05d090e6158b33b5bda7e77b2a32c017e5d/src/test/README.md) against JSC.

src/dictionary/methods/accessors.hpp
```diff
@@ -0,0 +1,59 @@
...
+template <typename T>
+struct DictionaryGetterSetter{
+    IOCollection *collection;
+    using Value = js::Value<T>;
+    using JSMixedAPI = TypeMixed<T>;
+
+    void set(accessor::Arguments args){
+        auto context = args.context;
+        auto key = args.property_name.c_str();
+        try{
```
kneth, line 36: Is this needed? I mean, won't the `wrap` of `RealmObjectClass<T>::set_property()` catch it (https://github.com/realm/realm-js/blob/master/src/js_realm_object.hpp#L89)?

src/js_object_accessor.hpp
```diff
@@ -235,7 +242,7 @@ class NativeAccessor {
     }
 
     bool is_same_dictionary(realm::object_store::Dictionary const& dictionary, ValueType const& value) const {
-        throw std::runtime_error("'Dictionary' type support is not implemented yet");
+        return false;
```
kneth, line 245: Is there no way we can compare to dictionaries?

src/test/catch.cpp
```diff
@@ -0,0 +1,9430 @@
+//              Copyright Catch2 Authors
```
kneth, line 1: Is it possible to reuse the `Catch2` version in core: https://github.com/realm/realm-core/tree/master/external? If no, I would prefer to include it as a git submodule.

src/test/test.sh
```diff
@@ -0,0 +1,12 @@
+g++ -g -std=c++17 -Wall -I catch.hpp -I test_bed.hpp -I ../ \
```
kneth, line 1: This seems to be very MacOS specific. Of course, we are mostly using MacOS but I think it would be better to have a proper CMake configuration.

src/CMakeLists.txt
```diff
@@ -1,16 +0,0 @@
-add_library(realm-js-shared OBJECT
```
kneth, line 1: Why is this file deleted?

src/common/object/jsc/object.hpp
```diff
@@ -0,0 +1,230 @@
...
+class JavascriptObject {
+   private:
+    JSClassDefinition _class;
+    JSContextRef context;
+    JSObjectRef object{nullptr};
+    std::vector<JSStaticFunction> methods;
+    std::vector<std::string> accessors;
+    PrivateStore *private_object;
+
+    static std::string to_string(JSContextRef context, JSStringRef value) {
```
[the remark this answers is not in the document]
cesarvr: Because that one is not reusable, unless you pull the full global variable in the form of template type. 

In order to make this module self contain and testable I/we need to avoid pulling global variables in disguise like the one necessary to reuse the function you mention.

src/common/object/node/methods.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "common/collection.hpp"
+#include "common/object/observer.hpp"
+
+#pragma once
+
+auto NodeCallbackWrapper(const Napi::CallbackInfo &values) {
+    return [&](int index) -> Napi::Value { return values[index]; };
+}
+namespace method {
```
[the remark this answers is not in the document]
cesarvr: I think templates aren't meant to solve the problem I'm trying to solve here, which is to target/optimise for different architectures. 

I try to keep a homogeneous type across nodejs and JSC.

src/common/object/strategies.hpp
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include <iostream>
+
+#include "common/object/interfaces.hpp"
+
+#pragma once
```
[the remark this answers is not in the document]
cesarvr: I think that's unnecessary because those modules above should contain the same protection. If they don't we want to know.

src/js_object_accessor.hpp
cesarvr: We can but we don't need to, or at least in the current implementation. Can you suggest a use case for this ?

src/test/catch.cpp
cesarvr: It's possible, but I didn't have time to read the the extensive documentation in realm-core explaining the way they use it. 

Also pressed by time have to go with something simple and that solve our problem.

src/test/test.sh
cesarvr: Because it should work for us, edge cases running Linux are welcome to contribute with a PR :).

src/CMakeLists.txt
cesarvr: That was my ide that did that by accident will delete it if not necessary :)

src/dictionary/methods/accessors.hpp
cesarvr: Yes but it will make the module untestable.

Leaving us two choices tested code or perpetuating a bad practice with the use of global variables.

src/common/object/error_handling.hpp
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+#pragma once
+
+#if REALM_PLATFORM_NODE
```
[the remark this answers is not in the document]
cesarvr: Sorry that's a dead module will delete it.

src/common/js_plain_object.hpp
```diff
@@ -0,0 +1,136 @@
...
+    using Object = js::Object<VM>;
+    using ObjectType = typename VM::Object;
+    using ContextType = typename VM::Context;
+
+    ContextType context;
+    bool waiting_for_notifications{false};
+
+    std::unique_ptr<Builder> builder;
+    std::unique_ptr<Collection> collection;
+    std::vector<Subscriber*> subscribers;
```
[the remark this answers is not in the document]
cesarvr: Good catch, 👍 I would fix that soon.

Would run some instrumentation to catch others lurking in the dark :)

src/common/js_plain_object.hpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    void subscribe(Subscriber* subscriber) {
+        subscribers.push_back(subscriber);
+        watch_collection();
+    }
+
+    void remove_subscription(const Subscriber* subscriber) {
+        int index = -1;
+        for (auto const& candidate : subscribers) {
```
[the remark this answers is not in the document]
cesarvr: I try that the problem is that it makes a transformation in place leaving empty spaces and providing the wrong size for the vector at least in my experience.

I would double check this as I'm all in to make sure that's the case.

src/dictionary/methods/functions.hpp
```diff
@@ -0,0 +1,102 @@
...
+        arguments.observer->unsubscribe_all();
+    }
+
+    static void put(method::Arguments arguments) {
+        std::string error_msg =  "This method cannot be empty.";
+        auto context = arguments.context;
+        IOCollection *collection = arguments.collection;
+
+        object_keys(context, arguments.get(0, error_msg),
+                    [=](std::string& key, auto object){
```
fronck, line 77:
```suggestion
                    [=](std::string& key, auto object) {
```

src/dictionary/methods/subscriber.hpp
```diff
@@ -0,0 +1,93 @@
...
+    PFunction fn;
+    PGlobalContext context;
+
+    DictionaryChangesSubscriber(ContextType &_context, FunctionType &_fn)
+        : fn{_context, _fn},
+          context{Context<T>::get_global_context(_context)} {}
+
+    DictionaryChangesSubscriber(DictionaryChangesSubscriber&& _subscriber){}
+
+    DictionaryChangesSubscriber(DictionaryChangesSubscriber& _subscriber){}
```
fronck, line 44:
```suggestion
    DictionaryChangesSubscriber(DictionaryChangesSubscriber&& _subscriber) {}

    DictionaryChangesSubscriber(DictionaryChangesSubscriber& _subscriber) {}
```

src/dictionary/methods/subscriber.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+    template <typename Collection>
+    auto build_array(Collection &collection) const {
+        std::vector<ValueType> values;
+        for (auto mixed_item : collection) {
+            values.push_back(
```
fronck, line 50:
```suggestion
            values.emplace_back(
```

src/dictionary/methods/subscriber.hpp
```diff
@@ -0,0 +1,93 @@
...
+    bool equals(std::unique_ptr<Subscriber>& rhs) const{
+        return (callback() == rhs->callback());
+    }
+
+    void notify(ObjectType object, DictionaryChangeSet& change_set) {
+        ValueType arguments[]{object, build_changeset_object(change_set)};
+        Function<T>::callback(context, fn, object, 2,
+                              arguments);
+    }
+    ~DictionaryChangesSubscriber(){}
```
fronck, line 88:
```suggestion
    ~DictionaryChangesSubscriber() {}
```

src/js_types.hpp
```diff
@@ -24,6 +24,7 @@
 #include <string>
 #include <vector>
 #include <sstream>
+#include <iostream>
```
fronck, line 27: Is this include required, or a leftover from debugging?

src/test/README.md
```diff
@@ -0,0 +1,142 @@
+# How To Test C++ And JS
+
+Usually to test against Javascript Core (JSC) we have to boot up an Android emulator, with all its dependencies (SDK, NDK, etc.) and then
+go through the suffering of having to create/load a react-native dummy app. Usually this way of working will require at between `20-40G` of space and lots of patience.
```
fronck, line 4:
```suggestion
go through the process of having to create/load a react-native dummy app. Usually this way of working will require at between `20-40G` of space and lots of patience.
```

src/test/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+To avoid all that pain we should target JSC directly and avoid the expensive middlemen for all code that aims directly to JSC (roughly 90%) and while 
+doing it, we can enjoy the added benefit of also unit test our C++ codebase and improve the overall quality of our SDK. 
+
+>As a *proof of concept* this folder includes Dictionary tests against [JavaScriptCore](https://developer.apple.com/documentation/javascriptcore).
+
+### API's
```
fronck, line 11:
```suggestion
### APIs
```

src/test/README.md
```diff
@@ -0,0 +1,142 @@
+# How To Test C++ And JS
+
+Usually to test against Javascript Core (JSC) we have to boot up an Android emulator, with all its dependencies (SDK, NDK, etc.) and then
+go through the suffering of having to create/load a react-native dummy app. Usually this way of working will require at between `20-40G` of space and lots of patience.
+
+To avoid all that pain we should target JSC directly and avoid the expensive middlemen for all code that aims directly to JSC (roughly 90%) and while
```
fronck, line 6:
```suggestion
To avoid all that complexity we should target JSC directly and avoid the expensive middlemen for all code that aims directly to JSC (roughly 90%) and while 
```

fronck: Isn't this something that your implementation already does?  ("should" makes it sound like we've taken a shortcut in our implementation that we shouldn't have)

src/test/README.md
```diff
@@ -0,0 +1,142 @@
...
+    /* We test if this is correct by sending our string to C++ land. */
+    assert_enumerate(json);
+   ```  
+> Let's call this file `main.js`.
+
+Now we can write a test like this:
+
+```cpp
+  /*
+   *  Previous boilerplate should be here...
```
fronck, line 112:
```suggestion
   *  Previous boilerplate would have been here...
```

fronck (changes requested):
Please see the inline comments.

On a personal note -- in the interest of improving code quality -- I'd really like to see some Doxygen documentation of, at least, the major functionality.  This would also make the code logic easier to follow during code review.

src/common/object/node/methods.hpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+namespace accessor{
+    struct Arguments{
+        Napi::Env context;
+        std::string property_name;
+        Napi::Value value;
+
+        void throw_error(std::string&& message){
+            throw Napi::Error::New(context, message);
```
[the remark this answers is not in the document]
cesarvr: They are, but only if I use the global variables that will make the code very hard to test. To make this piece of work testable I decided to implement/replicate some functionalities in the form of isolated/testable modules. this is one of those and hopefully we can migrate more functionality into a state that we can test it without requiring an emulator.

src/js_types.hpp
cesarvr: yep :), good catch 👍

src/test/README.md
cesarvr: Just thought that this addition (Catch3 and the JSC boilerplate) might deserve a justification as to why I think is good to have it.

src/test/test.sh
cesarvr: Because it should worked me, edge cases are welcome to contribute with a PR :).

kneth (commented):
As our code base is heavily underdocumented, I suggest that comments like in https://github.com/realm/realm-js/blob/develop/src/js_set.hpp is added.

src/common/object/node/object.hpp
```diff
@@ -0,0 +1,157 @@
...
+    void set_collection(IOCollection *_collection) { collection = _collection; }
+
+    void set_observer(ObjectObserver *_observer) { observer = _observer; }
+
+    template <class VM, void callback(method::Arguments)>
+    void add_method(std::string &&name) {
+        auto object = get();
+
+        auto method = [=](const auto &info) {
+            callback({info.Env(), observer, collection, info.Length(),
```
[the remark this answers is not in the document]
cesarvr: > Do we need a HANDLESCOPE here?
No, because the JSObject class shouldn't be not concerned about the current scope that the user (the dev) want to use.

PR #3633, opened 2021-03-10
Throw meaningful error if SO load fails on Android

kraenhansen:
## What, How & Why?

With this change a user would see a more meaningful error if loading the librealmreact.so fails from being loaded in an Android app without libjsc.so:

> `Realm JS does not support the Hermes engine yet. Express your 💚  on https://github.com/realm/realm-js/issues/2455`

Screenshot 2021-03-10 at 10 04 32

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests


steffenagger (approved):
LGTM (and good idea)

kneth (approved):

muhammadwafa:
When this lib will support hermes ?

PR #3653, opened 2021-03-17
Mention CMake for Android NDK in README.md

kneth:


scripts/build-android.js
```diff
@@ -170,6 +171,14 @@ function getCmakePath(sdkPath) {
     return process.platform === 'win32' ? 'cmake.exe' : 'cmake';
 }
 
+function getCmakeVersion(sdkPath) {
+    let dirs = fs.readdirSync(`${sdkPath}/cmake`);
+    if (dirs.length === 0) {
+        throw new Error(`No CMake installation found in ${sdkPath}/cmake`);
+    }
+    return dirs.sort()[dirs.length - 1];
```
blagoev, line 179: This
```
const d = ["3.6.4111459", "3.20", "3.18.1", "3.10.2.4988404"]
console.log(d.sort())
```

will result in 
```
["3.10.2.4988404", "3.18.1", "3.20", "3.6.4111459"]
```

So it will not always take the latest cmake version available in the SDK.
Is this the desired behavior?

Also it will be great if the build script outputs which version it will use when it chooses it. As the other places in the script do for other environment variables etc. This will help with debugging possible issues later.

kneth: It was on purpose that I picked the latest version. But I agree I would be good to log the version. I'll add it.

blagoev: The change will NOT always picking up the latest version. That was my question?

kneth: At least it was my intention.

PR #3662, opened 2021-03-25
Add SSL configuration options to sync configuration.

kneth:
## What, How & Why?

When using a proxy server, it might be required to disable SSL validation.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* [x] jsdoc files updated

### DIFF
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -35,7 +35,7 @@
  */
 
 /**
- * This describes the different options used to create a {@link Realm} instance with Realm Cloud synchronization.
+ * This describes the different options used to create a {@link Realm} instance with MongoDB Realm synchronization.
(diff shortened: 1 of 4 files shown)


docs/sync.js
```diff
@@ -50,6 +51,14 @@
  *    file to complete and then open. If not set, the Realm will be downloaded before opened.
  */
 
+ /**
+ * This describes the different options used to create a {@link Realm} instance with Realm Platform synchronization.
```
bmunkholm, line 55: "Realm Platform" - that's not a term we use. Also in line 38, it's called "Realm Cloud", which is also not a term we use :-)
It's "MongoDB Realm"

kneth: Damn, just copied from v6.x. I'll fix it - thanks!

PR #3665, opened 2021-03-26
Make AppClass::transport_generator initialized at const-init time

RedBeard0531:
## What, How & Why?
<!-- Describe the changes and give some hints to guide your reviewers if possible. -->
<!-- E.g. reference to other repos: This closes realm/realm-sync#??? -->

This fixes the "bad_function_call" issue where it was invoked while null                                                                                                                                                                                             (probably actually uninitialized). I'm not sure why it wasn't                                                                                                                                                                                                        initialized yet, since the compiler should have ensured that it was.                                                                                                                                                                                                 There is likely a compiler bug there but I'll need to dig in more. To                                                                                                                                                                                                work around that, I made it a simple function pointer, which lambdas can                                                                                                                                                                                             implicitly convert to, and all of that should be possible at const-init                                                                                                                                                                                              time (basically, it should be burned into the binary image with nothing                                                                                                                                                                                              happening at runtime). Unfortunately due to a different compiler bug,                                                                                                                                                                                                (https://developercommunity.visualstudio.com/t/const-init-of-function-pointers-from-lambdas/1383098),                                                                                                                                                                this requires putting a `+` in front of the lambda to explicitly decay                                                                                                                                                                                               it to a function pointer, rather than relying on the implicit conversion.  


## ☑️ ToDos
<!-- Add your own todos here -->
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/js_network_transport.hpp
+++ b/src/js_network_transport.hpp
@@ -179,6 +179,8 @@ struct JavaScriptNetworkTransport : public app::GenericNetworkTransport {
     }
 
     using SendRequestHandler = void(ContextType m_ctx, const app::Request request, std::function<void(const app::Response)> completion_callback);
+
+    // This needs to be a plain function pointer to ensure that AppClass::transport_factory is correctly initialized on MSVC builds.
     using NetworkTransportFactory = std::unique_ptr<app::GenericNetworkTransport>(*)(ContextType, Dispatcher);
 
     JavaScriptNetworkTransport(ContextType ctx, Dispatcher eld) : m_ctx(ctx), m_dispatcher(std::move(eld)) {};
--- a/src/js_network_transport.hpp
+++ b/src/js_network_transport.hpp
@@ -179,7 +179,7 @@ struct JavaScriptNetworkTransport : public app::GenericNetworkTransport {
     }
 
     using SendRequestHandler = void(ContextType m_ctx, const app::Request request, std::function<void(const app::Response)> completion_callback);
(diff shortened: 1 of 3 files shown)


kneth (commented):
I think it is a bit over my C++ skills. Feel free to merge when CI passes.

src/js_network_transport.hpp
```diff
@@ -179,7 +179,7 @@ struct JavaScriptNetworkTransport : public app::GenericNetworkTransport {
     }
 
     using SendRequestHandler = void(ContextType m_ctx, const app::Request request, std::function<void(const app::Response)> completion_callback);
-    using NetworkTransportFactory = std::function<std::unique_ptr<app::GenericNetworkTransport>(ContextType, Dispatcher)>;
+    using NetworkTransportFactory = std::unique_ptr<app::GenericNetworkTransport>(*)(ContextType, Dispatcher);
```
fronck, line 182: We should probably leave a comment here as to why the C-style function pointer is necessary.  Just so someone doesn't "clean up" the code and re-introduce the bug.

fronck (changes requested):
Please leave a couple of comments for posterity about why the workaround is here, so we don't inadvertently reverse it.

Otherwise, LGTM.

src/js_network_transport.hpp
RedBeard0531: :+1:

PR #3544, opened 2021-01-29
Cesarvr/mixed for linked realm object

cesarvr:
This PR adds [Typed_Link](https://realm.io/docs/javascript/latest/#relationships) support to Realm-JS. 
Allowing the user to use the Mixed field type to store this relationships and primitive types.

PR #3669, opened 2021-03-29
Set Datatype

### LINKED ISSUE #3460: Set: Test sync works against Cloud

fronck:
## What, How & Why?

This PR implements support for Realm's new Set datatype on JS.

This PR closes #3460, #3386, #3384, #3378.

Partially addressed:
 * Object creation / manipulation API (awaiting [PR 3544](https://github.com/realm/realm-js/pull/3544))
 
## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated
* [-] Chrome debug API is updated if API is available on React Native


src/js_schema.hpp
```diff
@@ -169,6 +174,53 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
...
+            prop.object_type = "";
+        }
+        else if (prop.object_type == "decimal128") {
+            prop.type |= PropertyType::Decimal | PropertyType::Set;
+            prop.object_type = "";
+        }
+        else if (prop.object_type == "objectId") {
+            prop.type |= PropertyType::ObjectId | PropertyType::Set;
+            prop.object_type = "";
+        }
```
kneth, line 213: UUID has been merged to `develop` so it should be possible to add it.

fronck: The schema parsing is fixed.  Will add unit tests

src/js_object_accessor.hpp
```diff
@@ -233,8 +234,8 @@ class NativeAccessor {
         return false;
     }
 
-    bool is_same_set(realm::object_store::Set const& set, ValueType const& value) const {
-        throw std::runtime_error("'Set' type support is not implemented yet");
+    bool is_same_set(realm::object_store::Set const &set, ValueType const &value) const {
+        return false;
```
takameyer, line 238: Is it not possible to return true here?

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+
+        // add another person
+        realm.write(() => {
+            people[0].Persons.add({FirstName: 'bob', Age: 99});
+        });
+
+        people = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, people.length, "There should be one 'People' entry")
+        TestCase.assertEqual(3, people[0].Persons.length, "Persons Set length should be 3");
+//        console.log("Joe's name:  " + people.Persons[0].FirstName);
```
takameyer, line 167:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+
+        realm.write(() => {
+            // insert two people
+            realm.create(people_schema.name, {Persons: [
+                {FirstName: "Joe",  Age: 4},
+                {FirstName: "Sue",  Age: 53},
+            ]});
+        });
+
+
```
takameyer, line 154:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+            people[0].Names.delete("Sue");
+        });
+        people = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, people.length, "There should be one 'People' entry")
+        TestCase.assertEqual(3, people[0].Names.length, "Persons Set length should be 3");
+        
+        realm.close();
+    },
+
+
```
takameyer, line 126:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+function closeAfterUpload(realm) {
+    return realm.syncSession.uploadAllLocalChanges().then(() => realm.close());
+}
+
+const Realm = require('realm');
+const TestCase = require('./asserts');
+const schemas = require('./schemas');
+const Utils = require('./test-utils');
+const { Decimal128, ObjectId } = require("bson");
+//const { Test } = require('tslint');
```
takameyer, line 40:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+'use strict';
+const AppConfig = require("./support/testConfig");
+
+
+
```
takameyer, line 23:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+        people = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, people.length, "There should be one 'People' entry")
+        TestCase.assertEqual(3, people[0].Persons.length, "Persons Set length should be 3");
+//        console.log("Joe's name:  " + people.Persons[0].FirstName);
+        let onfe = people[0].Persons.get(0);
+
+        let filteredsues = people[0].Persons.filtered('FirstName = "Sue"');
+
+        console.log("Sue's age:  " + filteredsues[0].Age);
+
```
takameyer, line 173: There doesn't seem to be any assertions on `filteredsues`.  Is this test perhaps not complete?

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+            sync: {
+                user,
+                partitionValue: "_id",
+                _sessionStopPolicy: "immediately", // Make it safe to delete files after realm.close()
+            },
+            schema: [schema]
+        };
+        const realm = await Realm.open(config);
+
+
```
takameyer, line 297:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+            primaryKey: "_id",
+            properties: {
+                _id: "int",
+                intSet: "int<>",
+            }
+        }
+        
+        const appConfig = AppConfig.integrationAppConfig;
+        const app = new Realm.App(appConfig);
+//        Realm.App.Sync.setLogLevel(app, 'all');
```
takameyer, line 323:
```suggestion
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,397 @@
...
+        });
+        await realm.syncSession.uploadAllLocalChanges();
+        objects = realm.objects(schema.name);
+        objcount = objects.length;
+        // there should still only be one object
+        TestCase.assertEqual(1, objcount, "Number of objects should be 1");
+        // .. but the object's Set should have two elements
+        TestCase.assertEqual(0, objects[0].intSet.length, "Length of Set should be 0");
+
+
```
takameyer, line 394:
```suggestion
```

src/js_types.hpp
```diff
@@ -507,7 +510,7 @@ inline bool Value<T>::is_valid_for_property_type(ContextType context, const Valu
             && (type != PropertyType::Object || list->get_object_schema().name == object_type);
     };
 
-    if (!realm::is_array(type)) {
+    if (!realm::is_array(type) && !realm::is_set(type)) {
```
takameyer, line 513: It seems that there is no explicit test here for if `set` `is_valid_for_property_type`.  But perhaps thats legitimate, I do not have much knowledge on how this function is being used.

takameyer (approved):

src/js_object_accessor.hpp
fronck: Well, that depends on the semantics we are going for.  It should always be `false` for different sets -- even if they contain the same elements -- because they create different objects in Core.
We could return `true` if a Set object is compared to itself, I guess.

tests/js/set-tests.js
fronck: Yeah, still working on more coverage on unit tests.  I'll make sure to incorporate the feedback, though :+1:

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+function closeAfterUpload(realm) {
+    return realm.syncSession.uploadAllLocalChanges().then(() => realm.close());
+}
+
+const Realm = require('realm');
+const TestCase = require('./asserts');
+const schemas = require('./schemas');
+const Utils = require('./test-utils');
+const { Decimal128, ObjectId } = require("bson");
```
kraenhansen, line 36: I think you might need to use `Realm.BSON` instead of `require("bson")` here.

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+    pathSeparator = '\\';
+}
+
+const fs = isNodeProcess ? nodeRequire('fs-extra') : require('react-native-fs');
+
+module.exports = {
+    async testSetSchema() {
+        // test that short (JS) and canonical schema types yield
+        // the same results
+        const shorthand_schema = {
```
kraenhansen, line 53: Let's keep the JS convention here and use camelCasing for local variable names.
I'll refrain from repeating this comment throughout the file 🙂 
```suggestion
        const shorthandSchema = {
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        TestCase.assertEqual(sh_schema.properties, can_schema.properties, 
+            "Canonical and shorthand schemas should have idendical properties");
+    },
+
+    async testSetStringAddDelete() {
+        // test add/delete operations on Sets of type string
+        const people_schema = {
+            name: "Person",
+            properties: {
+                Names: "string<>"
```
kraenhansen, line 85: No need to deviate from conventions here either 🙂 
```suggestion
                names: "string<>"
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        const shorthand_realm = new Realm({ schema: [shorthand_schema] });
+        const sh_schema = shorthand_realm.schema;
+        shorthand_realm.close();
+
+        const canonical_realm = new Realm({ schema: [canonical_schema] });
+        const can_schema = canonical_realm.schema;
+        canonical_realm.close();
+
+        TestCase.assertEqual(sh_schema.properties, can_schema.properties, 
+            "Canonical and shorthand schemas should have idendical properties");
```
kraenhansen, line 77: Personally I wouldn't bother with these messages, the assertion library should clearly state the issue and jumping to the offending line would make it obvious what is going on. On the other hand, when test cases are copy-pasted (as they often are) I've seen very frequently that these messages are not updated to match the code and they become more of a liability than a help to the developer.

Again - this is a personal preference, feel free to disagree 🙂

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        realm.close();
+    },
+
+
+    async testSetObjectAddDelete() {
+        const person_schema = {
+            name: "Person",
+            properties: {
+                FirstName: 'string',
+                Age: 'int'
```
kraenhansen, line 129: Same as above.
```suggestion
                firstName: 'string',
                age: 'int'
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        realm.write(() => {
+            // insert two people
+            realm.create(people_schema.name, {
+                Names: ["John", "Sue"]
+            });
+        });
+
+        let people = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, people.length, "There should be one 'People' entry")
+        TestCase.assertEqual(2, people[0].Names.length, "Persons Set length should be 2");
```
kraenhansen, line 101: I suggest using the size property here, since that's closer to the specification we're implementing towards.
```suggestion
        TestCase.assertEqual(2, people[0].Names.size, "Persons Set size should be 2");
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+//         });
+//         people = realm.objects(people_schema.name);
+
+//         TestCase.assertEqual(1, people.length, "There should be one 'People' entry")
+//         TestCase.assertEqual(2, people[0].Persons.length, "Persons Set length should be 2");
+
+        // TODO: add another 'People'
+    },
+
+    async testSetUtilityFunctions() {
```
kraenhansen, line 189: I would probably split these into multiple tests to make it easier to debug a potential regression in the future.

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        let teams = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, teams.length, "There should be only one JS team");
+        let js_team = teams[0];
+        TestCase.assertEqual(3, js_team.Names.length, "There should be 3 people in the JS team");
+        TestCase.assertEqual(true, js_team.Names.has("Kenneth"), "Kenneth should be in the JS team");
+        TestCase.assertEqual(false, js_team.Names.has("Jason"), "Jason shouldn't be in the JS team");
+
+        // add one JS team member, delete another one
+        realm.write(() => {
+            js_team.Names.add("Kræn").delete("Kenneth");
```
kraenhansen, line 220: I don't think he would like that 😆 

In all seriousness, Alice, Bob and Charlie are often better used here.

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        });
+        teams = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, teams.length, "There should be only one JS team");
+        js_team = teams[0];
+        TestCase.assertEqual(3, js_team.Names.length, "There should be 3 people in the JS team");
+        TestCase.assertEqual(false, js_team.Names.has("Kenneth"), "Kenneth shouldn't be in the JS team");
```
kraenhansen, line 226: 🙀

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        // TODO: add another 'People'
+    },
+
+    async testSetUtilityFunctions() {
+        const people_schema = {
+            name: "Person",
+            properties: {
+                Names: "string<>"
+            }
+        };
```
kraenhansen, line 195: Changing this might make your examples easier to read?

```suggestion
        const teamSchema = {
            name: "Team",
            properties: {
                names: "string<>"
            }
        };
```

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+
+        realm.write(() => {
+            // insert two people
+            realm.create(people_schema.name, {
+                Names: ["John", "Sue"]
+            });
+        });
+
+        let people = realm.objects(people_schema.name);
+        TestCase.assertEqual(1, people.length, "There should be one 'People' entry")
```
kraenhansen, line 100: I believe it's a general convention for assertion frameworks ([including ours](https://github.com/realm/realm-js/blob/master/tests/js/asserts.js#L60)) to take the `actual` value as the first argument and the `expected` value as a second argument.

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        TestCase.assertEqual(2, teams.length, "There should be two teams");
+        js_team = teams[0];
+        java_team = teams[1];
+        TestCase.assertEqual(0, js_team.Names.length, "There should be no people in the JS team");
+        TestCase.assertEqual(2, java_team.Names.length, "There should be 2 people in the Java team");
+
+        realm.close();
+    },
+
+    // TODO:  Move Sync tests to separate file
```
kraenhansen, line 263: Why is that?

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        const appConfig = AppConfig.integrationAppConfig;
+        const app = new Realm.App(appConfig);
+        const credentials = Realm.Credentials.anonymous();
+
+        const user = await app.logIn(credentials);
+        const config = {
+            sync: {
+                user,
+                partitionValue: "_id",
+                _sessionStopPolicy: "immediately", // Make it safe to delete files after realm.close()
```
kraenhansen, line 287: Shouldn't we register a logger and ensure no errors are emitted? Since a sync error wouldn't result in an error being thrown when we call our user-facing API.

tests/js/set-tests.js
```diff
@@ -0,0 +1,392 @@
...
+        const app = new Realm.App(appConfig);
+//        Realm.App.Sync.setLogLevel(app, 'all');
+        const credentials = Realm.Credentials.anonymous();
+
+        const user = await app.logIn(credentials);
+        const config = {
+            sync: {
+                user,
+                partitionValue: "_id",
+                _sessionStopPolicy: "immediately", // Make it safe to delete files after realm.close()
```
kraenhansen, line 327: Same as above - we should probably register a logger to ensure no sync related errors are logged.

tests/mongodb/config.json
```diff
@@ -1,6 +1,6 @@
 {
-    "app_id": "auth-integration-tests-denli",
-    "config_version": 20180301,
+    "app_id": "auth-integration-tests-bhnql",
+    "config_version": 20200603,
```
kraenhansen, line 3: We should really migrate this to the [`realm-app-importer`](https://github.com/realm/realm-js/tree/master/packages/realm-app-importer) to avoid this being constantly overridden.

kraenhansen (changes requested):
I think we should get closer the to JS Set API, let's have a call about my suggestions 🙂

src/js_types.hpp
fronck: `is_valid_for_property_type` is invoked when elements are inserted into the Set.  It asserts that the new element conforms to the type of Set it is inserted into.

src/js_object_accessor.hpp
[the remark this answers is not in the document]
fronck: It's not stubbed.  It's implemented the same as for List.
The premise is that two sets can never be the same, as they create different object in the core database.  We could make an exception if one compares a Set to itself.

src/js_schema.hpp
```diff
@@ -80,6 +79,11 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
         type = type.substr(0, type.size() - 2);
     }
 
+    if (type.ends_with("<>")) {
```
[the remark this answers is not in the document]
fronck: I'll fix the API ticket.  `{}` has been changed to be used for Dict, so Set now uses `<>`.

src/js_schema.hpp
```diff
@@ -169,6 +174,57 @@ static inline void parse_property_type(StringData object_name, Property& prop, S
...
+        else {
+            if (is_nullable(prop.type)) {
+                throw std::logic_error(util::format("List property '%1.%2' cannot be optional", object_name, prop.name));
+            }
+            if (is_array(prop.type)) {
+                throw std::logic_error(util::format("List property '%1.%2' must have a non-list value type", object_name, prop.name));
+            }
+            prop.type |= PropertyType::Object | PropertyType::Set;
+        }
+    }
```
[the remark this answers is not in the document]
fronck: Yeah, it's pretty hideous.  I as actually thinking that it should be delegated into the Set class itself rather than in `js_schema.hpp`

tests/js/set-tests.js
fronck: Indeed.  The unit tests are incomplete ATM.  There should be more coverage, and the commented-out bits are just there for posterity when the unit test PR comes in.

tests/js/set-tests.js
fronck: Awww :)

tests/js/set-tests.js
fronck: I'm actually not sure.  It just seems to be the current convention that Sync tests don't reside in the same place as the other tests.
They seem to be guarded both in inclusion ([here](https://github.com/realm/realm-js/blob/5316468d40a6c5ab26825e1ac0fc289665029ffc/tests/js/index.js#L63)) and in the tests themselves by `if (!global.enableSyncTests) return;`.  I'm not sure if this convention has a deeper purpose.

src/js_set.hpp
```diff
@@ -0,0 +1,403 @@
...
+#include "js_types.hpp"
+#include "js_util.hpp"
+
+#include <realm/object-store/shared_realm.hpp>
+#include <realm/object-store/set.hpp>
+
+namespace realm {
+namespace js {
+
+template<typename JSEngine>
```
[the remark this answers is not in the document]
fronck: It is a copy-paste.  However, I think it's good practice to use more descriptive typenames than `T` if the template accepts a narrow list of types.  Also, `JSEngine` is distinct from the `T` used in the Set class.

tests/js/set-tests.js
fronck: I like having the messages there.  For example, if you test that the size of `teams` is on several occasions, Jamine's output won't tell you which of the tests has failed.  Adding different messages will help locate the failing test.

src/js_object_accessor.hpp
kneth: We could compare the (table key, object key) pair, right?

src/js_set.hpp
```diff
@@ -0,0 +1,403 @@
...
+    static void filtered(ContextType, ObjectType, Arguments &, ReturnValue &);
+
+    // observable
+    static void add_listener(ContextType, ObjectType, Arguments &, ReturnValue &);
+    static void remove_listener(ContextType, ObjectType, Arguments &, ReturnValue &);
+    static void remove_all_listeners(ContextType, ObjectType, Arguments &, ReturnValue &);
+
+     std::string const name = "Set";
+
+     MethodMap<T> const methods = {
```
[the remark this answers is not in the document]
kneth: For `List` and `Results` we implement these methods in JavaScript (by copying methods from the prototype of `Array` - see `lib/collection-methods.js`). If our `Set` is `Array`ish, we can do the same here.

src/js_set.hpp
```diff
@@ -0,0 +1,466 @@
...
+    // if (size == 0) {
+    //     set->verify_in_transaction();
+    //     return_value.set_undefined();
+    // }
+    // else {
+    //     get_index(ctx, this_object, size - 1, return_value);
+    //     list->remove(size - 1);
+    // }
+
+    // TODO: assert that args[0] is a number
```
kneth, line 292: `Value::is_number()` might be useful.

src/js_set.hpp
```diff
@@ -0,0 +1,466 @@
...
+    auto const value_type = set->get_type();
+
+    switch_on_type(value_type, [&](auto const type_indicator) -> void {
+        try {
+            using element_type = std::remove_pointer_t<decltype(type_indicator)>;
+            int requested_index = Value::validated_to_number(ctx, args[0]);
+            realm::Mixed element_value = set->template get<std::remove_pointer_t<decltype(type_indicator)>>(requested_index);
+            return_value.set(element_value);
+        } catch (...) {
+            // TODO:  propagate exception to JS
```
kneth, line 304: If you throw a C++ exception, it will be caught and rewritten to a JS exception. You might not need the `try`/`catch`.

src/js_set.hpp
```diff
@@ -0,0 +1,466 @@
...
+    if (type == realm::PropertyType::Object) {
+        object_type = set.get_object_schema().name;
+    }
+    if (!Value::is_valid_for_property_type(ctx, value, type, object_type)) {
+        throw TypeErrorException("Property", object_type ? object_type : local_string_for_property_type(type), Value::to_string(ctx, value));
+    }
+}
+
+template<typename T>
+void SetClass<T>::add_listener(ContextType ctx, ObjectType this_object, Arguments &args, ReturnValue &return_value) {
```
kneth, line 446: Maybe a comment saying that `args` is validated elsewhere.

tests/js/set-tests.js
kneth: I find them useful when asserting in a loop - good to know which iteration is failing.

tests/js/set-tests.js
kneth: The convention comes from the time where we didn't run sync tests on RN (which is still true).

src/js_set.hpp
fronck: Done 👍

src/js_set.hpp
fronck: You're right.  Fixed the comment.

src/js_schema.hpp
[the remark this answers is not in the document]
fronck: FWIW, I've moved it to `realm::js::set::derive_property_type()`.  That removes a lot of clutter in `js_schema.hpp`.

tests/js/set-tests.js
fronck: Do we want to keep both these guards as standard practice?

src/js_set.hpp
```diff
@@ -0,0 +1,403 @@
...
+//         {"sum", wrap<compute_aggregate_on_collection<ListClass<T>, AggregateFunc::Sum>>},
+//         {"avg", wrap<compute_aggregate_on_collection<ListClass<T>, AggregateFunc::Avg>>},
+        {"addListener", wrap<add_listener>},
+        {"removeListener", wrap<remove_listener>},
+        {"removeAllListeners", wrap<remove_all_listeners>},
+     };
+
+     PropertyMap<T> const properties = {
+         {"size", {wrap<get_length>, nullptr}},
+         {"length", {wrap<get_length>, nullptr}},       // length property is required for, e.g., JavaScript serialization
```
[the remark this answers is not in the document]
kraenhansen: I believe we should implement a `toJSON` instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior

src/js_object_accessor.hpp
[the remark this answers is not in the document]
kraenhansen: > We could make an exception if one compares a Set to itself.

Isn't that the intended semantics of this function in the first place?

tests/js/set-tests.js
kraenhansen: Scrap that. We should upload, delete the client-side Realm file and download the changes from the server to ensure we can roundtrip correctly.

src/js_set.hpp
fronck: Done

tests/js/set-tests.js
kraenhansen: Scrap that, we should probably upload changes, delete the client-side realm and re-download to verify the round-trip.

src/js_object_accessor.hpp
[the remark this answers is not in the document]
fronck: Seems reasonable.  The `==` overload in Object Store does, in fact, compare columns and keys from the database.

fronck: Yes.  I've changed `is_same_set` to call the `==` overload operator from Object Store's Set, which does exactly that.

tests/js/set-tests.js
fronck: Yeah.  It actually makes the error messages from failed tests confusing to do it this way.  Expected and actual values are swapped.  I've fixed the tests.

kneth (approved):
We will need to update `Realm~ObjectSchemaProperty` and the example for `Realm~ObjectSchema` in `docs/realm.js`.

docs/set.js
```diff
@@ -0,0 +1,74 @@
...
+ * (see {@linkplain Realm~ObjectSchemaProperty ObjectSchemaProperty}).
+ *
+
+ * Sets mostly behave like normal JavaScript Sets, with a few exceptions:
+ * They can only store values of a single type (indicated by the `type` 
+ * and `optional` properties of the Set).
+ * They can only be modified inside a {@linkplain Realm#write write} transaction.
+ * Unlike JavaScript's Set, Realm~Set does NOT make any guarantees about the 
+ * traversal order of `values()`, `entries()`, `keys()`, or `forEach` iterations.
+ *
```
takameyer, line 30: Perhaps add a note, that if the developer wishes order to be established, they must implement it themselves.

tests/js/set-tests.js
fronck: @kneth @kraenhansen  I moved the Sync tests to a separate file.  Please resolve if that's a suitable solution.

docs/set.js
fronck: Done.

tests/js/set-tests.js
```diff
@@ -0,0 +1,483 @@
...
+const Realm = require('realm');
+const TestCase = require('./asserts');
+
+let pathSeparator = '/';
+const isNodeProcess = typeof process === 'object' && process + '' === '[object process]';
+const isElectronProcess = typeof process === 'object' && process.versions && process.versions.electron;
+
+if (isNodeProcess && process.platform === 'win32') {
+    pathSeparator = '\\';
+}
```
kraenhansen, line 37:
```suggestion
```

Same as the other file - this is not used.

tests/js/set-tests.js
```diff
@@ -0,0 +1,483 @@
...
+const require_method = require;
+function nodeRequire(module) {
+    return require_method(module);
+}
+
+
+const Realm = require('realm');
+const TestCase = require('./asserts');
+
+let pathSeparator = '/';
```
kraenhansen, line 31:
```suggestion
```

Same as the other file - this is not used.

types/index.d.ts
```diff
@@ -330,6 +330,45 @@ declare namespace Realm {
         readonly prototype: List<any>;
     };
 
+
+    /**
+     * Set
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Set.html }
+     */
+     interface Set<T> extends Collection<T> {
```
kraenhansen, line 338: Wouldn't it make more sense for this to be a `class` instead of an interface? (that way we can avoid lines 368 - 370)
```suggestion
     class Set<T> extends Collection<T> {
```

kraenhansen (approved):
I have a few suggestions, nothing major. Great job!

packages/realm-app-importer/src/RealmAppImporter.ts
```diff
@@ -131,7 +131,7 @@ export class RealmAppImporter {
     }
 
     private loadStichJson(appTemplatePath: string) {
-        const stitchJsonPath = path.resolve(appTemplatePath, "stitch.json");
+        const stitchJsonPath = path.resolve(appTemplatePath, "config.json");
```
kraenhansen, line 134: I just saw this - why is this needed?

kraenhansen:
Do we support construction and assignment of a set property by passing in a JS Set?

```typescript
const values = new Set();
values.add(1).add(2).add(3);

const realm = new Realm({ schema: [{ name: "Thing", properties: { values: "int<>" } }] });

realm.write(() => {
  realm.create("Thing", { values });
});
```

If no, it would be great if we did. I imagine users what to buffer values in a JS Set before adding them to a database.
If yes, it would be great if we had tests for that.

lib/extensions.js
```diff
@@ -90,6 +92,30 @@ module.exports = function(realmConstructor, environment) {
...
+    //     value: function (_, cache = new Map()) {
+    //         let elementArray = Array.from(this.values());
+    //         return elementArray.map((item, index) =>
+    //             item instanceof realmConstructor.Object ? item.toJSON(index.toString(), cache) : item);
+    //     },
+
+    //     writable: true,
+    //     configurable: true,
+    //     enumerable: false
+    // });
```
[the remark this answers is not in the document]
fronck: Already fixed after feedback from @takameyer 👍

lib/set-methods.js
```diff
@@ -0,0 +1,34 @@
...
+
+'use strict';
+
+
+var forEachMethod = function(callback) {
+    const elements = Array.from(this.values());
+    elements.forEach(element => callback(element));
+};
+
+var toJSONMethod = function() {
```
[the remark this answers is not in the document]
fronck: Also, re-written in a later commit.

lib/set-methods.js
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+'use strict';
+
+module.exports = function(realmConstructor) {
+    var forEachMethod = function(callback) {
```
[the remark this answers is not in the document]
fronck: Please take a look at the new version of `set-methods.js` and LMK if it looks better.

packages/realm-app-importer/src/RealmAppImporter.ts
fronck: The newer versions of the "Stitch" Docker image has changed the naming convention.

tests/js/set-sync-tests.js
```diff
@@ -0,0 +1,243 @@
...
+                partitionValue: "_id",
+                _sessionStopPolicy: "immediately", // Make it safe to delete files after realm.close()
+            },
+            schema: [schema]
+        };
+        const realm = await Realm.open(config);
+
+        // clear out any lingering documents
+        realm.write(() => {
+            realm.deleteAll();
```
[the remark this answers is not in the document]
kneth: Objects server-side are not deleted so if you run the test twice, you will have all objects twice.

tests/js/set-sync-tests.js
```diff
@@ -0,0 +1,243 @@
...
+        });
+
+        const integerSet = [1, 2, 3, 4, 5, 6, 7];
```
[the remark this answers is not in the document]
fronck: `integerSet` has been renamed to `integerArray`

tests/js/set-sync-tests.js
```diff
@@ -0,0 +1,243 @@
...
+        // make sure everything is in the database
+        let integers = realm.objects(schema.name)[0];
+        TestCase.assertEqual(integers.intSet.size, 7, "There should be 7 integers");
+
+        const realmPath = realm.path;
+
+        realm.close();
+
+        // make sure we don't have a local copy of the realm
+        fs.unlinkSync(realmPath);
```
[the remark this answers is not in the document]
fronck: Done.

tests/js/set-sync-tests.js
```diff
@@ -0,0 +1,243 @@
...
+        TestCase.assertEqual(syncedIntegers.size, 7, "There still should be 7 integers");
+
+        const intsValues = Array.from(syncedIntegers.values());
+        const sameInts = intsValues.map((value, index) => {
+            return value == integerSet[index];
+        });
+        const isSameInts = sameInts.reduce((prev, curr) => {
+            return prev && curr;
+        });
+        TestCase.assertTrue(isSameInts, "Downloaded integers should be the same as uploaded integers");
```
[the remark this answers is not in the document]
fronck: Good point.  I've changed the search algorithm.

packages/realm-app-importer/src/RealmAppImporter.ts
fronck: The changes to RealmAppImporter have been backed out.

kneth (approved):

kneth (approved):

PR #3678, opened 2021-04-07
Enable tests of aggregation functions

### LINKED ISSUE #3674: Error using objects.max function: Not implemented
## Goals

After updating from version `10.2.0` to `10.3.0`, one of our functions stopped working. Created a test function and tried to get max value on Objects using `max` resulting in an error.

## Expected Results

Get the max value from list of objects.

## Actual Results

Got the following error:
Screenshot_20210405-094908

## Steps to Reproduce

1. Create Realm
2. Get objects
3. Attempt to get max value from objects

## Code Sample

```
Model = {
  name: "Test",
  properties: {
    count: { type: "int" },
  },
};

// Receives list of objects
const objects = realm.objects("TEST");
// attempt to call max on list of objects
return objects.max("count");
// receives Not implemented
```

- Realm JS SDK Version: 10.3.0
- Node or React Native: React Native 0.64
- Client OS & Version: Android 8.0.0 and iOS 14.4.2

kneth:
## What, How & Why?

This closes #3674

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* On React Native, calling an aggregate function would fail with error `Not implemented`. ([#3674](https://github.com/realm/realm-js/issues/3674), since v10.2.0)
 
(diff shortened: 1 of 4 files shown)


steffenagger (approved):
LGTM (but let's re-include the aggregation-tests in `js/tests/js/list-tests.js` as well).

PR #3702, opened 2021-04-29
More testing of Realm.Set

kneth:
* Enable `[]` on `Realm.Set`
* Test `Realm.Set.has()`
* Test set of mixed

### DIFF
--- a/src/js_set.hpp
+++ b/src/js_set.hpp
@@ -37,7 +37,7 @@ class NativeAccessor;
 namespace set {
     /**
  * @brief Derive and apply property flags for \ref Set.
- * 
+ *
  * @param object_name Name of the Set object (for error reporting purposes)
  * @param prop (mutable) Property object that will be changed to be correct for \ref Set
  * @exception std::logic_error Thrown if the the prop argument contains an invalid property configuration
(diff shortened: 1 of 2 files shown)


tests/js/set-tests.js
```diff
@@ -154,37 +188,36 @@ module.exports = {
 
     //
     // test filtering on object properties
-    async testSetObjectFilter() {
+    testSetObjectFilter() {
```
fronck, line 191: Why remove the `async` from the test functions?

steffenagger: No need, `await` is not utilized inside the function.

src/js_set.hpp
```diff
@@ -162,6 +162,7 @@ struct SetClass : ClassDefinition<T, realm::js::Set<T>, CollectionClass<T>> {
 
      MethodMap<T> const methods = {
         {"add", wrap<add>},
+        {"get", wrap<get>},
```
fronck, line 165: Is this required?  We had a long discussion about whether to expose a `get` function during the main implementation work.

kneth: It is used in the test, I enabled?

tests/js/set-tests.js
kneth: I don't see any async operations, so I believe it is not required.

src/js_set.hpp
fronck: Yes, but the test was written and commented out prior to the decision to remove the `get` accessor was taken.
However, since `Set` is also a `Collection`, there are other ways to do index-based accession without exposing `get` on `Set`.

kneth: I will change it to `[]`.

PR #3732, opened 2021-05-11
Integration tests of `mixed` values

### LINKED ISSUE #3463: Mixed: Test sync works against Cloud

kraenhansen:
## What, How & Why?

This closes #3463 by adding tests that write, upload, redownload and read values of all supported mixed types.

It also adds three Mocha hooks to the integration test suite:
- `importAppBefore` calls the app importer to import an app
- `authenticateUserBefore` ensures an authenticated users
- `openRealmBefore` opens a local or synced Realm before the test runs

This also adds a `Mixed` type to our TypeScript declaration.

## ☑️ ToDos
* [ ] 📝 Changelog entry: I don't think this is needed, as this is just adding tests.
* [x] 🚦 Tests


cesarvr (approved):
LGTM

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -0,0 +1,13 @@
+import { importApp, TemplateReplacements } from "../utils/import-app";
+
+export function importAppBefore(name: string, replacements: TemplateReplacements = {}, logLevel: Realm.App.Sync.LogLevel = "warn") {
+  before(async function(this: Partial<AppContext> & Mocha.Context) {
+    this.timeout(10000);
```
takameyer, line 5: Why the timeout?  Perhaps add a comment and maybe we can think of a way to conditionally await some sort of readiness.

integration-tests/tests/src/tests/credentials/anonymous.ts
```diff
@@ -1,18 +1,14 @@
...
-  before(async function() {
-    app = await importApp("simple");
-  });
+  importAppBefore("simple");
 
-  it.skip("authenticates", async function() {
+  it("authenticates", async function(this: AppContext) {
     const credentials = Credentials.anonymous();
-    const user = await app.logIn(credentials);
+    const user = await this.app.logIn(credentials);
```
takameyer, line 11: I think that setting `this.app` implicitly in the `importAppBefore` hook is perhaps too magical.  Without looking into the implementation of said hook, I would not know that `this.app` is being set.

integration-tests/tests/src/tests/import-app-util.ts
```diff
@@ -10,7 +10,7 @@ describe.skipIf(environment.missingServer, "importApp utility", function() {
     const app = await importApp("simple");
     expect(app).instanceOf(App);
     expect(app.id.startsWith("simple")).equals(true);
-  });
+  }).timeout(10000);
```
takameyer, line 13: Anyway we can get rid of the timeout?  This overtime will really start slowing the CI down.

integration-tests/tests/src/tests/realm-constructor.ts
```diff
@@ -24,6 +24,10 @@ import { IPerson, PersonSchema, DogSchema } from "../schemas/person-and-dogs";
 const RealmAsAny = Realm as any;
 
 describe("Realm#constructor", () => {
+    beforeEach(() => {
+        Realm.clearTestState();
+    });
+
```
takameyer, line 30: I like how this is being explicitly done now.  Good change! :)

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -0,0 +1,134 @@
...
+        this.realm.create<MixedClass>("MixedClass", {
+          _id: this._id,
+          value: this.value,
+          // Adding a few other unrelated elements to the list
+          list: [this.value, 123, false, "something-else"],
+        });
+      });
+    });
+  
+    itUploadsDeletesAndDownloads();
```
takameyer, line 66: I assume that the db state is persisted when this test is called?  Is there cleanup required here?

takameyer (approved):
Very readable and easy to follow.  Lots of testing magic, but it's all in the context of the test and one can quickly understand what is happening.

integration-tests/tests/src/hooks/import-app-before.ts
kraenhansen: Oh, this is not calling `setTimeout` internally. This is just to prevent the hook from failing if it takes more than 2s (and less than 10s in this case) to import the app.

integration-tests/tests/src/tests/credentials/anonymous.ts
kraenhansen: Yeah, I agree this is not how most Mocha tests are written and having the `this.app` injected is not obvious where it came from. However, it (the global `Mocha.Context`) is a feature of Mocha to transfer values between hooks, suites and tests. As far as I can tell, this is the purpose of the feature.

My hope is that these hooks will be used in every sync related test and if we start relying more on them, we should also ensure proper documentation of their use for every new developer on the project to read. And as such, it won't be "magical" to the average developer contributing to the tests, because they would be used everywhere in the sync tests. 🤞

integration-tests/tests/src/tests/import-app-util.ts
kraenhansen: Again, this is not waiting 10s - it just ensures that the test won't fail if it takes more than 2s.

integration-tests/tests/src/tests/realm-constructor.ts
kraenhansen: Yeah - it turned out we didn't want to do this after every test, if multiple tests are using the same open Realm.

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: > I assume that the db state is persisted when this test is called?
Depends on what you mean by "db state is persisted".
The test awaits the database being uploaded before deleting it.

> Is there cleanup required here?
Nope, that's done by the `after` hook in `openRealmBefore`.

integration-tests/tests/src/hooks/import-app-before.ts
takameyer: A very good.  So this will return if it takes too long.  I'm happy with that :)

kraenhansen: Yeah, it will return (or basically abort / fail the suite) it the execution of the hook takes too much time (10s in this case).

Issue #3733, opened 2021-05-12
Types for `User#callFunction` doesn't match implementation across platforms


kraenhansen:
## Goals

A strongly typed and unified API across Realm JS and Realm Web.

## Expected Results

I would expect calling a MongoDB Realm function would have the same signature across our supported platforms.

## Actual Results

The [TypeScript declarations](https://github.com/realm/realm-js/blob/master/types/app.d.ts#L495) and [Realm Web implementation](https://github.com/realm/realm-js/blob/master/packages/realm-web/src/User.ts#L322-L324) takes arguments using ["rest parameter"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) syntax. Whereas the Realm JS implemeation and [JS docs](https://github.com/realm/realm-js/blob/master/docs/sync.js#L698-L704) takes an array of arguments as its 2nd positional argument.

We might consider deprecating and consider the `callFunction` method private altogether, since we intent user to call functions in the slightly more convenient syntax:

```
user.functions.sum(1, 2);
```

## Version of Realm and Tooling

- Realm JS SDK Version: Realm JS 10.4.0 and Realm Web 1.2.1
- Node, React Native and Web.
- Client OS & Version: N/A
- Which debugger for React Native: None

PR #3734, opened 2021-05-12
Realn.User.callFunction() defaults to an empty array as argument

### LINKED ISSUE #3718: user.callFunction returns 'Assertion failed' if called with no args instead of simple error
## Note
I am reporting this issue as I see the following message in the terminal 

`IMPORTANT: Please report this at https://github.com/realm/realm-core/issues/new`

## Goals
Call a simple function which requires no args as authenticated user.

## Expected Results
{{user.callFunction("some_function_name") }}

should either throw a simple error explaining that `callFunction` expects  both name and args or default args to empty array if none are passed.

## Actual Results
```
./vendor/realm-core/src/realm/object-store/util/bson/bson.hpp:201: [realm-core-10.6.0] Assertion failed: m_type == Bson::Type::Array
0   realm.node                          0x0000000105d8752c _ZN5realm4utilL18terminate_internalERNSt3__118basic_stringstreamIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 28
1   realm.node                          0x0000000105d8782b _ZN5realm4util9terminateEPKcS2_lOSt16initializer_listINS0_9PrintableEE + 395
2   realm.node                          0x00000001058cb986 _ZN5realm2js9UserClassINS_4node5TypesEE13call_functionEN4Napi3EnvENS5_6ObjectERNS0_9ArgumentsIS3_EERNS0_11ReturnValueIS3_EE + 1174
3   realm.node                          0x00000001058c6315 _ZN5realm2js4wrapIXadL_ZNS0_9UserClassINS_4node5TypesEE13call_functionEN4Napi3EnvENS6_6ObjectERNS0_9ArgumentsIS4_EERNS0_11ReturnValueIS4_EEEEEENS6_5ValueERKNS6_12CallbackInfoE + 293
4   realm.node                          0x00000001058ce57b _ZZN4Napi12InstanceWrapIN5realm4node13WrappedObjectINS1_2js9UserClassINS2_5TypesEEEEEE29InstanceMethodCallbackWrapperEP10napi_env__P20napi_callback_info__ENKUlvE_clEv + 139
5   realm.node                          0x00000001058ce49a _ZN4Napi12InstanceWrapIN5realm4node13WrappedObjectINS1_2js9UserClassINS2_5TypesEEEEEE29InstanceMethodCallbackWrapperEP10napi_env__P20napi_callback_info__ + 42
6   node                                0x000000010005eefa _ZN6v8impl12_GLOBAL__N_123FunctionCallbackWrapper6InvokeERKN2v820FunctionCallbackInfoINS2_5ValueEEE + 122
7   node                                0x000000010024a488 _ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE + 616
8   node                                0x0000000100249a49 _ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE + 521
9   node                                0x00000001002491b2 _ZN2v88internalL26Builtin_Impl_HandleApiCallENS0_16BuiltinArgumentsEPNS0_7IsolateE + 258
10  node                                0x00000001009ce9d9 Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit + 57
11  node                                0x0000000100954284 Builtins_InterpreterEntryTrampoline + 676!!! IMPORTANT: Please report this at https://github.com/realm/realm-core/issues/new/choosesh
```

## Steps to Reproduce
Define a function named 'test_function' like so in the realm UI and in the settings allow an api-key authenticated user to call it via user id authentication. 

```
exports = function() {
  return "HELLO";
};
```

From node log in a user and try call the function without passing args.

```
const app = new App({ id: appId });

const credentials = Credentials.serverApiKey(apiKey);

const user = await app.logIn(credentials);

user.callFunction('test_function');
```

* Realm JS SDK Version: ^10.4.0
* Node: 12.18.2
* Client OS & Version: macOS Mojave 10.14.6

kneth:
## What, How & Why?

This closes #3718

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,26 @@
-10.4.1-rc.3 Release notes (2021-5-10)
+x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * None.
 
 ### Fixed
 * Fixed the naming of `url` (now `baseUrl`) property on an app config to match the TypeScript declaration and other SDKs. ([#3612](https://github.com/realm/realm-js/issues/3612))
+* `Realm.User.callFunction()` could crash if no arguments were applied. ([#3718](https://github.com/realm/realm-js/issues/3718), since v10.0.0)
+### Compatibility
+* MongoDB Realm Cloud.
+* APIs are backwards compatible with all previous releases of Realm JavaScript in the 10.x.y series.
+* File format: generates Realms with format v20 (reads and upgrades file format v5 or later for non-synced Realm, upgrades file format v10 for synced Realms).
 
+### Internal
+* Using Realm Core v10.6.0.
+
+10.4.1-rc.3 Release notes (2021-5-10)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* None.
 ### Compatibility
 * MongoDB Realm Cloud.
 * APIs are backwards compatible with all previous releases of Realm JavaScript in the 10.x.y series.
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -675,7 +675,7 @@ class User {
 
     /**
      * Links a user to another credentials. This is useful when linking
-     * different account togteher.
+     * different account together.
      * @param {Realm.Credentials} credentials
      * @returns {Promise<void>} - resolves when the user has been linked with the other credentials.
      */
@@ -698,7 +698,7 @@ class User {
     /**
      * Calls the named server function as this user.
      * @param {string} name - name of the function to call
-     * @param {any[]} args - list of arguments to pass, if no arguments are passes, `[]` is used
+     * @param {any[]} args = [] - list of arguments to pass
      * @return {Promise<any>} - resolves when the function terminates.
      */
     callFunction(name, args) { }
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -698,7 +698,7 @@ class User {
     /**
      * Calls the named server function as this user.
      * @param {string} name - name of the function to call
-     * @param {any[]} args - list of arguments to pass
+     * @param {any[]} args - list of arguments to pass, if no arguments are passes, `[]` is used
      * @return {Promise<any>} - resolves when the function terminates.
      */
     callFunction(name, args) { }
--- a/lib/user.js
+++ b/lib/user.js
@@ -33,6 +33,9 @@ const instanceMethods = {
     },
 
     async callFunction(name, args, service = undefined) {
+        if (!args) {
+            args = [];
+        }
         const cleanedArgs = cleanArguments(args);
         const stringifiedArgs = EJSON.stringify(cleanedArgs, { relaxed: false });
         const result = await promisify(cb => this._callFunction(name, stringifiedArgs, service, cb));
--- a/lib/user.js
+++ b/lib/user.js
@@ -32,10 +32,7 @@ const instanceMethods = {
         return promisify(cb => this._logOut(cb));
     },
 
-    async callFunction(name, args, service = undefined) {
-        if (!args) {
-            args = [];
-        }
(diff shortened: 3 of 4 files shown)


docs/sync.js
```diff
@@ -698,7 +698,7 @@ class User {
     /**
      * Calls the named server function as this user.
      * @param {string} name - name of the function to call
-     * @param {any[]} args - list of arguments to pass
+     * @param {any[]} args - list of arguments to pass, if no arguments are passes, `[]` is used
      * @return {Promise<any>} - resolves when the function terminates.
      */
     callFunction(name, args) { }
```
RedBeard0531, line 704: Can you do `args = []` here? I prefer putting things like that in code, rather than prose, although the esdoc html generator has some weird limitations. If not, you should probably at least change the type on line 701 to indicate optionality.

Also, we need to fix this since it is extremely wrong... https://github.com/realm/realm-js/blob/364448f3267e4f733018a248fe7d5a370f625a15/types/app.d.ts#L495

lib/user.js
```diff
@@ -33,6 +33,9 @@ const instanceMethods = {
     },
 
     async callFunction(name, args, service = undefined) {
+        if (!args) {
```
RedBeard0531, line 36: Why do this rather than `args = []` in the function signature?

CHANGELOG.md
```diff
@@ -1,11 +1,26 @@
-10.4.1-rc.3 Release notes (2021-5-10)
+x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * None.
 
 ### Fixed
 * Fixed the naming of `url` (now `baseUrl`) property on an app config to match the TypeScript declaration and other SDKs. ([#3612](https://github.com/realm/realm-js/issues/3612))
```
RedBeard0531, line 7: Was it correct to move this from `10.4.1-rc.3` to `x.x.x`? Is this just fixing a bug that it was in the wrong place before, or are you introducing a bug by moving it?

RedBeard0531 (changes requested):

docs/sync.js
kraenhansen: Just to make this obvious. The types you're referring to are shared with Realm Web.
I've created an issue to track fixing this: https://github.com/realm/realm-js/issues/3733

CHANGELOG.md
kneth: I forgot to add the template when I merged a PR yesterday so I am fixing it now.

kraenhansen (approved):
Looking good 🚢

RedBeard0531 (approved):
I'd rather we update the .d.ts files to reflect the correct types in this commit, but if there are reasons why we can't, I'm fine with deferring.

steffenagger:
> I'd rather we update the .d.ts files to reflect the correct types in this commit, but if there are reasons why we can't, I'm fine with deferring.

Our current TS declaration is:
```ts
callFunction(name: string, ...args: any[]): Promise<any>;
```
Thats not right if `callFunction` previously expected an array as second param? (@kraenhansen does realm-web handle this differently?)

Now it should just be:
```ts
callFunction(name: string, args?: any[]): Promise<any>;
```
right?

kneth:
@RedBeard0531 @steffenagger I believe that https://github.com/realm/realm-js/issues/3733 is trying to capture your concerns.

PR #3741, opened 2021-05-13
Fix the Issue Adding Link Objects to Dictionaries 

cesarvr:
This fix the issue the problem happens when we insert a new Link Realm object will cause the Dictionary->Mixed component to mistakenly report that Link Types (Realm Objects) are not supported.


src/common/object/jsc/support.hpp
```diff
@@ -0,0 +1,8 @@
+//
```
kneth, line 1: Is this file needed?

src/test/jsc_object.js
```diff
@@ -6,6 +6,17 @@ let null_dict = null_dictionary || {}
 // eslint-disable-next-line no-undef
 assert_true(dict.X !== undefined && dict.doSomething !== undefined);  // Testing successful object creation.
 
+
+/*
+    It seems that Object.values(dict) is not supported on JSC for non-static fields, it would be interesting to
```
kneth, line 11: You can create an issue for it and put it on the backlog so we don't forget it.

tests/js/dictionary-tests.js
```diff
@@ -461,9 +460,43 @@ module.exports = {
...
+        realm.write(()=> {
+                    let child = realm.create(Child.name, { num: 555 } )
+                    let child2 = realm.create(Child.name, { num: 666 } )
+                    realm.create(DictSchema.name, {
+                                 dict: { children1: child, children2: child2 },
+                                 dict2: { children1: child2, children2: child }
+                                 })
+                    })
+        
+        let l = realm.objects(DictSchema.name)[0].dict
```
kneth, line 491: The test seems to be incomplete. What happens when you uncomments the lines below?

PR #3746, opened 2021-05-19
dictionary toJSON fix

steffenagger:
As our `toJSON` implementation keeps track (a "cache") of `Realm.Object`s traversed, additional handling is needed for dictionaries. Even though dictionaries behaves more like JS objects, dictionaries can contain object references which still needs a helping hand, in case of circular structures.


cesarvr (approved):
LGTM

takameyer (approved):

PR #3748, opened 2021-05-19
Fixed  `Realm not defined` error under Jest

fronck:
## What, How & Why?

Fixed error that would cause `Realm.Set`'s iterators to throw a `Realm not defined` error.
This closes RJS-1196.

## ☑️ ToDos

* [x] 📝 Changelog entry
~* [ ] 🚦 Tests~
~* [ ] 📝 Public documentation PR created or is not necessary~

~* [ ] jsdoc files updated~

### DIFF
--- a/lib/collection-methods.js
+++ b/lib/collection-methods.js
@@ -55,7 +55,7 @@ Object.defineProperty(iteratorPrototype, Symbol.iterator, {
 
 ['entries', 'keys', 'values'].forEach(function(methodName) {
     var method = function() {
-        const isSet = this instanceof Realm.Set;
+        const isSet = (this.constructor && this.constructor.name && this.constructor.name === 'Set');
         var self = (this.type === "object" || isSet) ? this.snapshot() : this;
         var index = 0;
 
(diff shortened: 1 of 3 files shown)


lib/collection-methods.js
```diff
@@ -55,7 +55,7 @@ Object.defineProperty(iteratorPrototype, Symbol.iterator, {
 
 ['entries', 'keys', 'values'].forEach(function(methodName) {
     var method = function() {
-        const isSet = this instanceof Realm.Set;
+        const isSet = (this.constructor && this.constructor.name && this.constructor.name === 'Set');
```
kneth, line 58: I would suggest that we add a comment why `instanceof Realm.Set` doesn't work.

kneth (approved):

lib/collection-methods.js
kraenhansen: Wouldn't it be better to repeat the pattern you introduced for `set-methods.js`? https://github.com/realm/realm-js/blob/ca7051f8dc566101a8a89444575d16d5167ed83f/lib/set-methods.js#L21-L30

Basically refactoring `collections-methods.js` to export a function that takes the `Realm` constructor as an argument.

elliottkember:
I also had this error with v10.5 while running React Native under a debugger. I was able to fix it by adding `global.Realm = Realm` to a root file, but this would solve it in a nicer way.

takameyer (approved):

liamjones:
Was this fix abandoned/unneeded? I see the branch was deleted without merging.

takameyer:
@liamjones This branch was removed do to some merge conflicts, but the problem should have been fixed.  Can you make a new issue with some steps to reproduce `Realm not defined`?  We would appreciate it 😃

liamjones:
Sure, it's happening in one of our app monorepo projects so I'll see if I can reduce it down to a minimal repro.

Acetyld:
Can this be fixed asap?

PR #3749, opened 2021-05-19
New datatypes TS updates

steffenagger:
This PR exposes a (simple) `Realm.Dictionary` interface declaration, and updates the `RealmInsertionModel<T>` type, to include `Realm.Set`.

**Warning:** this PR references https://realm.io/docs/javascript/latest/api/Realm.Dictionary.html, which I believe hasn't been setup yet.

* [ ] jsdoc files updated **_(this will be done in a separate PR)_**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -231,7 +231,7 @@ declare namespace Realm {
      * Dictionary
      * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Dictionary.html }
      */
-    interface Dictionary<T = unknown> { [key: string]: T };
+    interface Dictionary<T = unknown> { [key: string]: T }
 
     /**
      * Collection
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -227,6 +227,12 @@ declare namespace Realm {
      */
     type SortDescriptor = [string] | [string, boolean];
 
+    /**
+     * Dictionary
+     * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Dictionary.html }
```
steffenagger, line 232: I don't believe this exists yet?

kneth: True. We will need to update the API doc too.

kneth (approved):

steffenagger:
Converted back to a draft. Dictionary needs more work, and is a bit tricky.

takameyer (approved):

takameyer (approved):

kraenhansen:
How would a user read a value from the dictionary with this declaration?

takameyer:
@kraenhansen They would manually define their dictionary structure as an interface and make an intersection with Realm.Dictionary.
`Realm.Dictionary & MyDictionaryInterface`

PR #3755, opened 2021-05-27
Quickfix: Run Realm.App._clearAppCache() on load.

### LINKED ISSUE #3668: App login function freezes/hangs when React-Native environment is reloaded (press R twice) on Android
## Goals

Fix freezing login function on App class in React-Native Android application after reloading.

## Expected Results

Login function should not freeze.

## Actual Results

Login function freezes/hangs forever after logging in into application when the application is reloaded.

## Steps to Reproduce

We have a React-Native application running as Android application (in emulator, but similar thing happens on real device). When I reload the application in React-Native environment (pressing R key twice) and I try to log in then the application will freeze forever on login function in App class.

What I already tried:
- `realmApp.allUsers[key].loggedIn` - false (only one user visible in allUsers)
- `realmApp.allUsers[key].logOut()` - didn't help
- `realmApp.currentUser` = returns null
- deleting realm files with adb shell commands - didn't help
- killing the application from memory - after starting the application it works again (just by closing where the application is running in the background and starting it from there doesn't help)
- reinstalling the application (`react-native run-android`) - works again

## Code Sample

```
const getRealmApp = () => {
  const appConfig = {
    id: Config.REALM_DIR,
    timeout: 30000,
  };
  return new Realm.App(appConfig);
};

export const realmApp = getRealmApp();



export const signIn = async jwt => {
  const creds = Realm.Credentials.jwt(jwt);
  await realmApp?.logIn(creds);  // <- the application will hang on this line
};
```

- Realm npm package: 10.2.0
- Realm JS SDK Version: ?
- react-native: 0.63.3
- Android: 10
- node: v14.9.0
- Which debugger for React Native: 0.10.7 (not connected)

steffenagger:
This PR executes `Realm.App._clearAppCache()` on load, to re-matchup C++ & JS environments for RN hot-reloading.

This closes #3668

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * A warning to polyfill `crypto.getRandomValues` was triggered prematurely ([#3714](https://github.com/realm/realm-js/issues/3714), since v10.4.0)
+* App login function to not hang after React-Native environment was reloaded ([#3668](https://github.com/realm/realm-js/issues/3668))
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/lib/index.js
+++ b/lib/index.js
@@ -65,5 +65,7 @@ require("./extensions")(realmConstructor, environment);
 
 const versions = utils.getVersions();
 realmConstructor.App._setVersions(versions);
+// Temporary solution to clear the app cache when a new JS context is created (this is require for hot reloading).
+realmConstructor.App._clearAppCache();
 
 module.exports = realmConstructor;


kraenhansen:
I don't know if this is a huge problem, but it seems we're already clearing this cache natively on iOS from [react-native/ios/RealmReact/RealmReact.mm#L267](https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmReact.mm#L267) (which calls [src/jsc/jsc_init.cpp#L57-L62](https://github.com/realm/realm-js/blob/master/src/jsc/jsc_init.cpp#L57-L62)).

To avoid calling it twice, we might either want to remove it from the native code entirely or add it to the Android native code instead of calling it from JS?

kneth:
It might be safer to do it when you load Realm, but I agree with @kraenhansen that we shouldn't do it twice.

kneth (approved):

kraenhansen:
Okay. I realize now that we're clearing other caches when running on Android here: https://github.com/realm/realm-js/blob/master/src/android/jsc_override.cpp#L137 - we should probably just add the `realm::app::App::clear_cached_apps();` after that line.

PR #3772, opened 2021-06-01
Validate configuration

### LINKED ISSUE #3771: Disallow in-memory for sync Realms
If `inMemory` is set to true for a synced Realm, the fatal error `terminating with uncaught exception of type realm::LogicError: Durability setting (as passed to the SharedGroup constructor) was not consistent across the session` can be observed.

In general, it is not possible to sync a Realm to memory only, and Realm Core assumes that the app developer is aware of this.

It is a bug that Realm JavaScript isn't validating the Realm configuration and terminates early so the app developer becomes aware of it.

kneth:
## What, How & Why?

This closes #3771 

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


fronck (approved):

nirinchev (changes requested):
Is `disableFormatUpgrade` an option compatible with sync? I would imagine that this would prevent Sync from writing to the file. Additionally, we should express these incompatibilities in the typescript definitions so that people who are using an editor that respects ts definitions will get early feedback if they try to construct an invalid config.

nirinchev (approved):

types/index.d.ts
```diff
@@ -107,25 +107,38 @@ declare namespace Realm {
...
+    interface ConfigurationWithSync extends BaseConfiguration {
+        sync: SyncConfiguration,
+        migration?: never,
+        inMemory?: never,
+        deleteRealmIfMigrationNeeded?: never,
+        disableFormatUpgrade?: never
+    }
+
+    interface ConfigurationWithoutSync extends BaseConfiguration {
+        sync?: never,
```
takameyer, line 129:
```suggestion
        sync: SyncConfiguration;
        migration?: never;
        inMemory?: never;
        deleteRealmIfMigrationNeeded?: never;
        disableFormatUpgrade?: never;
    }

    interface ConfigurationWithoutSync extends BaseConfiguration {
        sync?: never;
```
The interface properties should end in `;`

takameyer (changes requested):
Just a formatting issue that should be addressed.

I am not completely satisfied with how the typings resolve for configuration, but I have no alternative solution at the moment.  If the developer explicitly uses `ConfigurationWithSync`, then the TS errors look nice, but if they use the `Configuration` union, the TS errors are quite vague.  If I come up with a better solution later, I will make a separate PR.

Otherwise, LGTM

takameyer (approved):

Issue #2784, opened 2020-04-01
FR: electron-rebuild and ssl libs on Windows


xmedeko:
When using `realm-js` with Electron, I have to run `electron-rebuild`. The rebuild recompiles realms from sources, so it needs SSL libs installed on `C:\src\vcpkg\installed\x64-windows-static\lib`. Is it possible to configure the path somehow or to make it relative? E.g. some env variable like `REALM_SSL_LIBS` etc.? See also [my forum post](https://community.mongodb.com/t/electron-rebuild-and-ssl-libs-on-windows/1776).

Note: I know Electron is not officially supported platform, assume this as a FR to improve building `realm-js` from sources 😃 

Note 2: the `README.md` has some obsolete info about Node versions and the build process.

## Version of Realm and Tooling

- Realm JS SDK Version: 5.0.2
- Node: Node 12, Electron 8
- Client OS & Version: Windows 10

kneth:
@xmedeko Sorry for the late reply.

No, there is currently no way to customize the path. I would imagine it is possible to use an environment variable in the build file (https://github.com/realm/realm-js/blob/master/realm.gypi#L240). You are welcome to come up with a pull request to solve it :smile:.

PR #3777, opened 2021-06-03
Make vcpkg path configurable via an environment variable

kraenhansen:
## What, How & Why?

The CMakeLists.txt has a hardcoded path the vcpkg toolchain.
This PR makes the path configurable via an environment variable as suggested by [the CMake integration guide](https://github.com/microsoft/vcpkg/blob/master/docs/users/integration.md#cmake-toolchain-file-recommended-for-open-source-cmake-projects).

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kneth (approved):

kraenhansen:
This fixes https://github.com/realm/realm-js/issues/2784 too.

CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ if(DEFINED CMAKE_JS_VERSION)
         elseif(CMAKE_GENERATOR_PLATFORM MATCHES "^[Xx]64$")
             set(VCPKG_TARGET_TRIPLET "x64-windows-static" CACHE STRING "")
         endif()
-        set(CMAKE_TOOLCHAIN_FILE "C:\\src\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake")
+        if(DEFINED ENV{VCPKG_ROOT})
+            set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake" CACHE STRING "")
```
fealebenpae, line 11:
```suggestion
            set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake")
```
We don't need to save the variable in the cache.

fealebenpae (approved):

CMakeLists.txt
kraenhansen: I'm simply following the guide 🙂 Any reason this would hurt us?

fealebenpae: It doesn't hurt, but it doesn't do anything much either. Normally you would use `set(CACHE)` if you don't want to overwrite a variable if it's already defined in a super scope. We don't have a super scope for this variable, though, and indeed setting it here should trump any other definitions. Technically, I needn't have set `VCPKG_TARGET_TRIPLET` like that either, and that's likely a bad copy/paste job on my part. Mind you, these variable are only considered by the `project()` command so beyond that their value is not considered.
If you do want to save it in the cache you should use `FILEPATH` instead of `STRING` as the entry type.

PR #3790, opened 2021-06-07
Changes to existing files to prep for hermes support

RedBeard0531:
Main thing is that some types are not default constructible and have no null
state. Additionally, some types are no longer copyable.

## ☑️ ToDos
(Removing all TODOs since this is just an internal code change with no user impact on its own)

<del>

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

</del>

### DIFF
--- a/src/js_types.hpp
+++ b/src/js_types.hpp
@@ -279,8 +279,8 @@ struct Object {
     static void set_property(ContextType ctx, ObjectType&& obj, const String<T>& field, const ValueType& val, PropertyAttributes attributes = None) {
         set_property(ctx, obj, field, val, attributes);
     }
-    static void set_property(ContextType ctx, ObjectType&& obj, uint32_t field, const ValueType& val, PropertyAttributes attributes = None) {
-        set_property(ctx, obj, field, val, attributes);
+    static void set_property(ContextType ctx, ObjectType&& obj, uint32_t field, const ValueType& val) {
+        set_property(ctx, obj, field, val);
     }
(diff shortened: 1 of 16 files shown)


src/js_types.hpp
```diff
@@ -270,8 +276,14 @@ struct Object {
     static ValueType get_property(ContextType c, const ObjectType &o, const std::string &s) { return get_property(c, o, StringData(s)); }
     static ValueType get_property(ContextType, const ObjectType &, const String<T> &);
     static ValueType get_property(ContextType, const ObjectType &, uint32_t);
-    static void set_property(ContextType, const ObjectType &, const String<T> &, const ValueType &, PropertyAttributes attributes = None);
-    static void set_property(ContextType, const ObjectType &, uint32_t, const ValueType &);
+    static void set_property(ContextType ctx, ObjectType&& obj, const String<T>& field, const ValueType& val, PropertyAttributes attributes = None) {
```
RedBeard0531, line 279: Explanation, since it isn't obvious what changed here or why: JSI doesn't support setting properties on const Objects, so I had to remove const here. But you can't pass a temporary to a non-const ref argument, so just doing that prevents things like `set_property(ctx, get_global(), ...)`, which we do in a few places. So I added `Object&&` overloads, that just directly call the `Object&` overload.

blagoev (approved):
👍

kneth (approved):

PR #3794, opened 2021-06-08
Android RN should use same cache clearing logic as iOS

### LINKED ISSUE #3668: App login function freezes/hangs when React-Native environment is reloaded (press R twice) on Android

RedBeard0531:
## What, How & Why?

This makes it also call `realm::app::App::clear_cached_apps()`, which
fixes #3668. This also makes it more maintainable by having a single place to
put all cache clearing logic for all RN platforms.

This PR replaces and closes #3755.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

<del>
* [ ] jsdoc files updated
</del>

### DIFF
--- a/src/android/jsc_override.cpp
+++ b/src/android/jsc_override.cpp
@@ -134,7 +134,7 @@ static JSGlobalContextRef create_context(JSContextGroupRef group, JSClassRef glo
     swap_function();
 
     // Clear cache from previous instances.
-    realm::_impl::RealmCoordinator::clear_all_caches();
+    RJSInvalidateCaches();
 
     RJSInitializeInContext(ctx);
     realmContextInjected = true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * A warning to polyfill `crypto.getRandomValues` was triggered prematurely ([#3714](https://github.com/realm/realm-js/issues/3714), since v10.4.0)
+* Don't hang when using the network after hot-reloading an RN app. ([#3668](https://github.com/realm/realm-js/issues/3668))
 
 ### Compatibility
 * MongoDB Realm Cloud.


src/android/jsc_override.cpp
```diff
@@ -134,7 +134,7 @@ static JSGlobalContextRef create_context(JSContextGroupRef group, JSClassRef glo
     swap_function();
 
     // Clear cache from previous instances.
-    realm::_impl::RealmCoordinator::clear_all_caches();
+    RJSInvalidateCaches();
```
RedBeard0531, line 137: For reference, here is this function: https://github.com/realm/realm-js/blob/5b0eabf940d843b76835b6e889817f4052c0b5c3/src/jsc/jsc_init.cpp#L57-L62

kneth (approved):

kraenhansen (approved):

PR #3799, opened 2021-06-15
Extend the .gitignore for tests

takameyer:
## What, How & Why?

The current tests are generating artifacts that do not belong in the repository.
This adds these to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,11 @@ tests/react-test-app/ios/out.txt
 crash.log
 tests/react-test-app/ios/Pods/
 tests/react-test-app/ios/ReactTests.xcworkspace/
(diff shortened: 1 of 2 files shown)


kraenhansen (approved):
Historically we've been an heavy user of the root level .gitignore file. I'm wondering if this should be placed besides the other rules related to tests? https://github.com/realm/realm-js/blob/develop/.gitignore#L54-L74
Alternatively some of those rule might have to be moved into the tests/.gitignore file?

takameyer:
@kraenhansen I actually would have placed them at the root level, had their not been a .gitignore in the test directory.  I have no problem with merging them to the root level.

kneth (approved):

PR #3801, opened 2021-06-16
Improve Dictionary Typescript Types

### LINKED ISSUE #3798: Improve Typescript Typings for Dictionary
Dictionary is tricky to type, since it is possible to run into naming collisions between its methods (put, delete, etc) and the dictionary element keys.  For now, we have landed on a solution where the developer will simply make an intersection between their defined dictionary structure and the given dictionary methods.

For example:

```java
interface MyDictionary{
  a:string;
  b?:number;
}class Journal extends Realm.Object {
  tags?: Realm.Dictionary & MyDictionary;
...
```

It would be desirable to handle this through generics and also make sure that the naming collisions are impossible to override.

takameyer:
## What, How & Why?
    
Basically making dictionaries more dynamic.  There is also a bit of refactoring done for the RealmInsertionModel to make it more readable

This closes #3798

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -231,34 +231,28 @@ declare namespace Realm {
      * Dictionary
      * @see { @link https://realm.io/docs/javascript/latest/api/Realm.Dictionary.html }
      */
+    
+    type Dictionary<ValueType = Mixed> = DictionaryBase<ValueType> & {
+        [key: string]: ValueType;
+    }
 
-     //interface Dictionary<T = unknown> { [key: string]: T }
-
-
-     interface Dictionary{
-        /**
-         * @returns An object for JSON serialization.
-         */
-        toJSON(): Array<any>;
-
+    interface DictionaryBase<ValueType = Mixed> {
         /**
          * @returns Adds given element to the dictionary
          */
-        put(element:{[key:string]:any}): void;
+        put(element:{[key:string]: ValueType}): void;
 
         /**
          * @returns Removes given element from the dictionary
          */
-        remove(element:{[key:string]:any}): void;
+        remove(element:{[key:string]: ValueType}): void;
 
         /**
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -572,20 +570,36 @@ type ExtractPropertyNamesOfType<T, PropType> = {
...
-type RealmCollectionRemappedModelPart<T> = {
-    [K in keyof T]?: T[K] extends Realm.Collection<infer GT> ? Array<GT | RealmInsertionModel<GT>> : never
+type RealmListsRemappedModelPart<T> = {
+    [K in ExtractPropertyNamesOfType<T, Realm.List<any>>]?: T[K] extends Realm.List<infer GT> ? Array<GT | RealmInsertionModel<GT>> : never
 }
 
 /**
- * Joins T stripped of all keys which value extends Realm.Collection and all inherited from Realm.Object,
- * with only the keys which value extends Realm.List or Realm.Set, remapped as Arrays.
+ * Exchanges properties defined as Realm.List<Model> with an optional Array<Model | RealmInsertionModel<Model>>.
```
kraenhansen, line 580: This doc string probably needs updating.

types/index.d.ts
```diff
@@ -236,20 +236,16 @@ declare namespace Realm {
 
 
      interface Dictionary{
```
nirinchev, line 238: Shouldn't the interface declaration be generic since in Core you can constrain the dictionary values to be of specific type?

kneth (approved):

types/index.d.ts
takameyer: Yes, if they can be constrained, then we should definitely reflect that.

takameyer: @nirinchev Dictionary now allows the value to be constrained

PR #3805, opened 2021-06-17
Allow object spread syntax on realm objects

### LINKED ISSUE #2844: Spread operation not works
## Goals

Spread operation works on older versions , After upgrading from 3.4.1 to 5.0.3 , Spread operation not works

## Expected Results

Spread operation works on older versions , I passed ream object as props to react component by spreading it

```
<MyComponent {...realmObject}>
```

## Actual Results

Speading creates empty object

## Steps to Reproduce

```
onst CarSchema = {
  name: 'Car',
  properties: {
    make:  'string',
    model: 'string',
    miles: {type: 'int', default: 0},
  }
};

Realm.open({schema: [CarSchema]})
  .then(realm => {
    // Create Realm objects and write to local storage
    realm.write(() => {
      const myCar = realm.create('Car', {
        make: 'Honda',
        model: 'Civic',
        miles: 1000,
      });
      myCar.miles += 20; // Update a property value
    });

    // Query Realm for all cars with a high mileage
    const car = realm.objects('Car')[0];

   const spread = {...car};
  console.log(spread)


    // Remember to close the realm when finished.
    realm.close();
  })
  .catch(error => {
    console.log(error);
  });
```

- Realm JS SDK Version:  5.0.3
- Node or React Native: 10.15.3 - 0.62.2 
- Client OS & Version: ? Windows 10 10.0.18362

RedBeard0531:
## What, How & Why?

This adds a `Proxy` wrapper to each object, but only traps `ownKeys` and `getOwnPropertyDescriptor`, not `get` or `set` operations. This means that it has ~20-40% overhead in tight loops rather than a full order of magnitude from prior solutions.

If we want to, there is an easy way to allow the user to opt-in or out getting the Proxy wrapper, without too much additional code or complexity. But of course it is simpler both for us and users if there isn't an additional knob to consider.

This is still 🚧 WIP 🚧 . In particular I only changed the behavior on Node. I want to get feedback on whether we like this direction before doing the JSC implementation and adding some tests.

Fixes #2844

## ☑️ ToDos
* [ ] address all `XXX` comments
* [ ] add JSC support

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/node/node_class.hpp
+++ b/src/node/node_class.hpp
@@ -179,7 +179,8 @@ Napi::Reference<Napi::External<typename ClassType::Internal>> WrappedObject<Clas
 
 
 struct SchemaObjectType {
-    Napi::FunctionReference constructor;
+    Napi::FunctionReference factory;     // Use this to construct the object via Call() not New()
+    Napi::FunctionReference constructor; // Use this to check if the constructor is the same.
 };
 
 template <typename ClassType>
@@ -222,6 +223,11 @@ private:
 
     static Napi::Function init_class(Napi::Env env);
 
+    static const Napi::FunctionReference& get_or_create_realm_object_factory(Napi::Env env,
+                                                                             Napi::Function& constructor,
+                                                                             const realm::ObjectSchema& schema,
+                                                                             const SharedRealm& realm);
+
     static Napi::ClassPropertyDescriptor<WrappedObject<ClassType>>
     setup_method(Napi::Env env, const std::string& name, node::Types::FunctionCallback callback);
     static Napi::ClassPropertyDescriptor<WrappedObject<ClassType>>
@@ -235,7 +241,7 @@ private:
     static void property_setter(const Napi::CallbackInfo& info);
     static std::vector<Napi::PropertyDescriptor>
     create_napi_property_descriptors(Napi::Env env, const Napi::Object& constructorPrototype,
-                                     const realm::ObjectSchema& schema, bool redefine = true);
+                                     const realm::ObjectSchema& schema, Napi::Array& ownKeys, bool redefine = true);
 };
 
 template <>
@@ -312,7 +318,7 @@ inline auto& ObjectWrap<ClassType>::get_nativeMethods()
 template <typename ClassType>
 inline auto& ObjectWrap<ClassType>::get_schemaObjectTypes()
 {
-    static std::unordered_map<std::string, std::unordered_map<std::string, SchemaObjectType*>*> s_schemaObjectTypes;
+    static std::unordered_map<std::string, std::unordered_map<std::string, SchemaObjectType>> s_schemaObjectTypes;
     return s_schemaObjectTypes;
 }
 
@@ -1146,8 +1152,6 @@ Napi::Object ObjectWrap<ClassType>::create_instance(Napi::Env env, Internal* int
     return scope.Escape(factory).As<Napi::Object>();
 }
 
-inline static void schema_object_type_constructor(const Napi::CallbackInfo& info) {}
-
 template <typename ClassType>
 void ObjectWrap<ClassType>::internal_finalizer(Napi::Env, typename ClassType::Internal* internal)
 {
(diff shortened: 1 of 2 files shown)


src/node/node_class.hpp
```diff
@@ -284,7 +287,7 @@ inline auto& ObjectWrap<ClassType>::get_nativeMethods() {
  
 template<typename ClassType>
 inline auto& ObjectWrap<ClassType>::get_schemaObjectTypes() {
-	static std::unordered_map<std::string, std::unordered_map<std::string, SchemaObjectType*>*> s_schemaObjectTypes;
+	static std::unordered_map<std::string, std::unordered_map<std::string, SchemaObjectType>> s_schemaObjectTypes;
```
RedBeard0531, line 290: While I was in here, I refactored the code a bit to reflect some of the simplifications I did in the hermes port. I tried to reduce a lot of the duplicated logic (eg there were 4 different places where we were constructing the object)

src/node/node_class.hpp
```diff
@@ -1093,118 +1082,80 @@ inline std::vector<Napi::PropertyDescriptor> ObjectWrap<ClassType>::create_napi_
 			node::String* name = get_cached_property_name(propName);
 			auto descriptor = Napi::PropertyDescriptor::Accessor<property_getter, property_setter>(name->ToString(env), napi_enumerable, (void*)name);
 			properties.push_back(descriptor);
+			// XXX should computed properties be reported as enumerable own keys?
```
RedBeard0531, line 1085: This is a question for the team. I don't know which properties make sense to show up in a spread / have `instance.hasOwnProperty()` return true for.

If the model is that computed properties are like getters on a JS class, then they probably shouldn't. But if they are "like stored properties but computed on demand", then they probably should.

On the other hand, maybe this is something that calls for practicality over purity, and if it is more convenient for users if it shows up in a spread, maybe it is worth just making that happen, regardless of what the model would suggest. We should also be aware that it may be expensive to eagerly compute properties that are not needed, but I suppose that applies as much or more to fetching the fields from the DB.

src/node/node_class.hpp
```diff
@@ -1235,40 +1186,83 @@ Napi::Object ObjectWrap<ClassType>::create_instance_by_schema(Napi::Env env, Nap
...
+
+					}
+				};
+				return function factory() {
+					const obj = new Type();
+					if (!(obj instanceof Type))
+						throw new Error("Realm object constructor must not return another value");
+
+					// Note: we could allow user to opt-out of Proxy by returning obj here, if the
+					// perf overhead is too high for them. Perhaps using some bool on the constructor?
```
RedBeard0531, line 1223: Also a note/question for the team

src/node/node_class.hpp
```diff
@@ -1093,118 +1082,80 @@ inline std::vector<Napi::PropertyDescriptor> ObjectWrap<ClassType>::create_napi_
...
 	//if we are creating a RealmObject from schema with no user defined constructor
 	if (constructor.IsEmpty()) {
-		//1.Check by name if the constructor is already created for this RealmObject 
-		if (!schemaObjects->count(schemaName)) {
-
-			//2.Create the constructor
-
-			//get or create the RealmObjectClass<T> constructor
+		//1.Check by name if the factory is already created for this RealmObject 
+		// XXX should we also check if `constructor` is empty?
```
RedBeard0531, line 1112: This seems like a bug in the old code. I kept the behavior for now, but I'm not sure I should have. In the case where the user passes a ctor in, we check that it is the *same* ctor, and rebuild the factory if not. Should we do the same if they switch from using a ctor to not using one?

blagoev (commented):
This seems like an acceptable approach. The key find is we can skip providing `get` and `set` traps which is a big relief implementation wise. I agree and would vote for not providing a knob and either go with this or not. The spread operator is essentially cloning the object and we need to decide and settle on if we are fine with this as a general principle of how Realm should be used.

src/node/node_class.hpp
```diff
@@ -1195,123 +1187,87 @@ ObjectWrap<ClassType>::create_napi_property_descriptors(Napi::Env env, const Nap
...
+        schemaObjectConstructor = env.Global()
+                                      .Get("Function")
+                                      .As<Napi::Function>()
+                                      .New({
+                                          Napi::String::New(env, "name"), // argument name
+                                          Napi::String::New(env, "return {[name]: function () {}}[name];") // body
+                                      })
+                                      .As<Napi::Function>()
+                                      .Call({Napi::String::New(env, schema.name)})
+                                      .As<Napi::Function>();
```
kraenhansen, line 1239: Couldn't we use something like https://github.com/nodejs/node-addon-api/blob/main/doc/function.md#new for that?

```suggestion
        schemaObjectConstructor = Napi::Function::New(env, [] {}, schema.name);
```

tests/js/object-tests.js
```diff
@@ -777,4 +777,50 @@ module.exports = {
       obj.getPropertyType("foo");
     }, new Error("No such property: foo"));
   },
+
+  testObjectSpread: function () {
+    let realm = new Realm({ schema: [schemas.IntPrimary] });
```
takameyer, line 782: We need to test this with a class that extends `Realm.Object`, I believe this has a bug.

takameyer (commented):
I think I finally understand this code.  Added some comments and suggestions.

kneth:
It is now supported by v12

PR #3838, opened 2021-07-08
Housekeeping

takameyer:
Just updating eslint dependencies that were causing package-lock failures in npm 7.
Also added a Podfile.lock for the test app.


kneth (approved):
Do we need a comment in `README.md` to say the npm v7 is recommended?

takameyer:
The new lock file is backwards compatible with npm 6.

PR #3841, opened 2021-07-09
Catalyst Unit Tests

### LINKED ISSUE #3839: Add Catalyst Tests
Catalyst support is out. A Catalyst test phase in Jenkins must be added.

* Extend ReactTestApp to support catalyst
* Extend test.sh to test Catalyst
* Add step to Jenkins

takameyer:
This adds a testing phase for catalyst in the ReactTestApp

This closes #3839

## ☑️ ToDos
* [x] 📝 Changelog entry


kneth (approved):

tests/ReactTestApp/ios/Podfile.lock
```diff
@@ -272,7 +272,7 @@ PODS:
     - React-perflogger (= 0.64.2)
   - ReactNativeExceptionHandler (2.10.10):
     - React-Core
-  - RealmJS (10.5.0):
+  - RealmJS (10.6.0):
```
kraenhansen, line 275: It's fine this is updated here, but I'm in doubt if we even want / need to keep this checked in? These versions and hashes are derived from the NPM packages, which already have the `package-lock.json` to enforce a deterministic install.

kraenhansen (approved):
Looking good.
Once these changes are applied to the xCode project of the integration tests, I would love for them to get documented such that we can more easily upgrade React Native version in the future.

tests/ReactTestApp/ios/Podfile.lock
takameyer: Sure, makes sense.  I've always committed it before, but it really doesn't "need" to be there.

PR #3843, opened 2021-07-09
Fix for Set prototype resolution

### LINKED ISSUE #3788: "Realm is not defined" when iterating through Realm Sets using Jest
I'm getting the error: "ReferenceError: Realm is not defined" when attempting to iterate through a Realm set using Jest. 

Realm version: 10.5.0-beta.2 (although I've tested on 10.5.0-beta.1 as well)

Sample unit test:

```javascript
test("should work with the Set data type", async () => {
    const characterSchema = {
      name: "Character",
      primaryKey: "_id",
      properties: {
        _id: "objectId",
        name: "string",
        levelsCompleted: "int<>",
        inventory: "string<>",
      },
    };
      schema: [characterSchema],
    });

    let link, hunter;
    realm.write(() => {
      link = realm.create("Character", {
        _id: new BSON.ObjectId(),
        name: "Link",
        inventory: ["elixir", "compass", "glowing shield"],
        levelsCompleted: [4, 9],
      });
      hunter = realm.create("Character", {
        _id: new BSON.ObjectId(),
        name: "Hunter",
        inventory: ["estus flask", "gloves", "rune"],
        levelsCompleted: [1, 2, 5, 24],
      });
    });
    
    hunter.inventory.forEach((inventoryItem) => {
      console.log("The hunter has an inventory item:", inventoryItem);
    });

    realm.write(() => {
      realm.delete(link);
      realm.delete(hunter);
    });
    // close the realm
    realm.close();
  });
```

Some notes:

* Switching from a forEach loop to an alternative iterator method will still result in the error. 
* If you comment out the `forEach` method in the jest test you will not get the Realm undefined error. 
* If this code (minus the jest specific stuff) is run outside of a jest suite, for instance in a js file, it runs fine without error.

fronck:
## What, How & Why?

Fix type resolution for Set so that it works under Jest, node, and RN.

This closes #3788

## ☑️ ToDos

* [x] 📝 Changelog entry
~* [ ] 🚦 Tests~
~* [ ] 📝 Public documentation PR created or is not necessary~

~* [ ] jsdoc files updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* Fixed TypeScript definition of `Realm.Dictionary.remove()`. (since v10.6.0)
-* Fixed Realm.Object#toJSON as it threw "Right-hand side of 'instanceof' is not an object". (since v10.6.0)
+* Fix TypeScript definition of `Realm.Dictionary.remove()`. (since v10.6.0)
+* Fixed `Realm not defined` error experienced when using `Realm.Set` iterators under Jest
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/lib/collection-methods.js
+++ b/lib/collection-methods.js
@@ -55,11 +55,11 @@ Object.defineProperty(iteratorPrototype, Symbol.iterator, {
 
 ['entries', 'keys', 'values'].forEach(function(methodName) {
     var method = function() {
-    // checks for Set datatype.  Required for compatibility with node.js, Jest, and RN
-    // instanceof Set is not performing correctly in Jest tests, so we need to use an alternate method 
-    const isSet = ((this.constructor && this.constructor.name === 'Set') || this instanceof Realm.Set);
-    var self = (this.type === "object" || isSet) ? this.snapshot() : this;
-    var index = 0;
+        const isSet = ((this.constructor && this.constructor.name && this.constructor.name === 'Set') || this instanceof Realm.Set);
+//        const isSet = (this.constructor && this.constructor.name && this.constructor.name === 'Set');
+//        const isSet = this instanceof Realm.Set;
+        var self = (this.type === "object" || isSet) ? this.snapshot() : this;
+        var index = 0;
 
     return Object.create(iteratorPrototype, {
       next: {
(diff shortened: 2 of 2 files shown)


lib/collection-methods.js
```diff
@@ -55,7 +55,8 @@ Object.defineProperty(iteratorPrototype, Symbol.iterator, {
 
 ['entries', 'keys', 'values'].forEach(function(methodName) {
     var method = function() {
-        const isSet = this instanceof Realm.Set;
+        // checks for Set datatype.  Required for compatibility with node.js, Jest, and RN
```
takameyer, line 58:
```suggestion
        // checks for Set datatype.  Required for compatibility with node.js, Jest, and RN
        // instanceof Set is not performing correctly in Jest tests, so we need to use an alternate method 
```
is it only Jest?

takameyer (approved):
My only suggestion would be to add a failing test to this PR. Otherwise LGTM

kneth (approved):
We might apply the same trick to Dictionary.

CHANGELOG.md
```diff
@@ -4,8 +4,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* Fixed TypeScript definition of `Realm.Dictionary.remove()`. (since v10.6.0)
-* Fixed Realm.Object#toJSON as it threw "Right-hand side of 'instanceof' is not an object". (since v10.6.0)
+* Fix TypeScript definition of `Realm.Dictionary.remove()`. (since v10.6.0)
```
fronck, line 7:
```suggestion
* Fixed TypeScript definition of `Realm.Dictionary.remove()`. (since v10.6.0)
* Fixed Realm.Object#toJSON as it threw "Right-hand side of 'instanceof' is not an object". (since v10.6.0)
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Fixed
 * Fixed TypeScript definition of `Realm.Dictionary.remove()`. (since v10.6.0)
 * Fixed Realm.Object#toJSON as it threw "Right-hand side of 'instanceof' is not an object". (since v10.6.0)
+* Fixed `Realm not defined` error experienced when using `Realm.Set` iterators under Jest
```
fronck, line 9:
```suggestion
* Fixed `Realm not defined` error experienced when using `Realm.Set` iterators under Jest  (closes #3788)
```

PR #3847, opened 2021-07-12
Improve error message when using an array as value for a dictionary

### LINKED ISSUE #3730: Improve error message when inserting a list into a mixed property
## Goals

Clear and actionable error messages.

## Expected Results

An error message such as:

> A mixed property doesn't support lists of values

## Actual Results

The following error is thrown:

> Error: Only Realm instances are supported.

1. Open a Realm with a schema declaring a mixed property.
2. Try inserting a list of values

```typescript
const realm = new Realm({
  schema: [{
    name: "MixedClass",
    properties: { value: "mixed" },
  }],
});
realm.write(() => {
  realm.create("MixedClass", { value: [123, false, "hello"] });
});
```

- Realm JS SDK Version: 10.5.0-beta.1
- Node or React Native: Tested on Node.js
- Client OS & Version: N/A

kneth:
## What, How & Why?

Since an array is an object, the error message when using an array as value for a dictionary can be hard to decipher.

This closes #3730 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


fronck (approved):
:+1:

src/js_links.hpp
```diff
@@ -65,28 +65,32 @@ template <typename T>
...
     std::shared_ptr<Realm> realm;
 
   public:
     MixedLink(std::shared_ptr<Realm> _realm): realm{_realm} {}
 
-    Mixed wrap(Context context, Value const& value) {
+    Mixed wrap(Context context, ValueType const& value) {
+
+        if (Value::is_array(context, value)) {
+            throw std::runtime_error("An array cannot be converted to Mixed.");
```
kraenhansen, line 80: Technically, nothing is being "converted" here.
From the endusers point of view, they're trying to store an array as the value of a `Mixed` property.
Which is why I suggested the following message:
> A mixed property doesn't support lists of values

Perhaps "support" is not as precise as we want to be. Another alternative could be:

> Arrays are not a mixed value

or

> Cannot use an array as a Mixed value

kneth: Or maybe `A mixed property cannot contain an array of values`?

takameyer (approved):
👍

fronck (approved):
:+1:

takameyer (approved):
Looks good!

tests/js/mixed-tests.js
```diff
@@ -196,4 +198,30 @@ module.exports = {
...
+
+  testMixedValuesAsArray() {
+    const MixedSchema = {
+      name: "MixedClass",
+      properties: { value: "mixed" },
+    };
+
+    const realm = new Realm({ schema: [MixedSchema] });
+
+    let objectsBefore = realm.objects(MixedSchema.name);
```
kraenhansen, line 210:
```suggestion
    const objectsBefore = realm.objects(MixedSchema.name);
```

kraenhansen (approved):
There's a couple of places in the test where you're using `let` which might as well be `const`. Not a big deal though.

PR #3850, opened 2021-07-13
ESLinting

kraenhansen:
## What, How & Why?

This updates the eslint configurations across our Lerna sub-packages.

Another PR will introduce linting for the C++ part of our codebase.

## ☑️ ToDos
* [ ] Revisit the changes from on [kh/eslint-and-prettier](https://github.com/realm/realm-js/tree/kh/eslint-and-prettier)
* [x] Attempt to hoist eslint config and plugin packages
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


.github/workflows/pr-linting.yml
```diff
@@ -0,0 +1,26 @@
...
+    
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+          registry-url: https://registry.npmjs.org/
+      # Install the root package (--ignore-scripts to avoid downloading or building the native module)
+      - run: npm ci --ignore-scripts
+        name: Install root package dependencies
```
takameyer, line 22: Can we do `name` before `run`?  That reads better for me.

package.json
```diff
@@ -45,9 +45,11 @@
     "CMakeLists.txt"
   ],
   "scripts": {
+    "lerna:bootstrap": "lerna bootstrap",
```
takameyer, line 48: It's more convenient to type `npx lerna bootstrap` than `npm run lerna:bootstrap`.  How about `npm run bootstrap`?

kneth (approved):

integration-tests/tests/package.json
```diff
@@ -11,7 +11,7 @@
     "build:watch": "tsc --watch",
     "app-importer": "realm-app-importer serve ../realm-apps",
     "mocha": "ts-mocha --project ./tsconfig.json --require src/utils/inject-node-globals.ts --watch-extensions ts,js src/index.ts",
-    "lint": "tslint -p .",
+    "lint": "eslint --ext .js,.ts .",
```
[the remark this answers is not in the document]
fronck: We've been talking about using the Clang linter for C++ files, like Core.  Isn't that why it's constrained?

fronck (approved):

package.json
kraenhansen: You can still type `npx lerna bootstrap`? I think `bootstrap` as a package script name is a bit too generic.

kraenhansen:
I found the need to do a different approach. Sorry that I have to ask for a re-review.

.prettierrc.js
```diff
@@ -1,7 +1,4 @@
 module.exports = {
-    jsxBracketSameLine: true,
     trailingComma: 'all',
-    tabWidth: 4,
```
kraenhansen, line 4: This change might be controversial. Prettier's default is 2 spaces.

.prettierrc.js
```diff
@@ -1,7 +1,4 @@
...
-    arrowParens: 'avoid',
-    // printWidth: 120,
+    printWidth: 120,
```
kraenhansen, line 3: I remember Lubo specifically requesting this and I agree that the default 80 chars feels a bit restricted on todays monitors.

fronck (approved):

fronck (changes requested):
It seems like there's some installation needed on the CI.  The eslint command fails

.prettierrc.js
kneth: It is roughly the same as what is recommended for Realm Core.

.prettierrc.js
kneth: I think it is pretty common in JS circles.

contrib/linting.md
```diff
@@ -0,0 +1,44 @@
...
+- Writing code in a shared style makes way easier to read across files.
+- As with any static analysis of code, linting has the potential to find bugs before our users do.
+
+## Setup your editor to format on save
+
+To help yourself keep your code passing linters at all times, it's highly recommended to setup your editor to help you.
+
+### Visual Studio Code
+
+- Install and enable the ["eslint" extension for VS Code](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).
```
kneth, line 14: We could add it to `extensions.json` (https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions)

kneth (approved):

contrib/linting.md
kraenhansen: I believe I already did that 🙂

.github/workflows/integration-tests.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
           registry-url: https://registry.npmjs.org/
+      - name: Install npm v7
+        run: npm install -g npm@7
+        # TODO: Remove the following line once we've found a workaround for https://github.com/dotnet/msbuild/issues/5726
+        if: ${{ matrix.os != 'windows-latest' }}
```
takameyer, line 41: I think it reads better to have the `if` before the `run`.

.prettierrc.js
takameyer: This is great.  Excited for 2 spaces and a constrained width.  120 seems appropriate even for working directly on the laptop monitor

package.json
takameyer: I agree.  Then I will just keep using `npx lerna bootstrap`

package.json
```diff
@@ -141,7 +141,8 @@
     "typescript": "^3.8.2"
```
takameyer, line 141: I actually want to address the `eslint-plugin-*` devDependencies above.  There seem to be many, and they also could be updated.  Would it be possible to update and also constrain the plugins to the package that uses them?

takameyer (commented):
Good work here.  Happy to see it was possible to share the eslint dependencies and rules across packages.

package.json
kraenhansen: A lot of them are shared across packages.
I would argue that hoisting them to the root is the simplest, makes them easier to upgrade and have no downsides (except the root package.json being a bit more cluttered). My eventual vision would be to make the `realm` package a sub-package like all the others and in that way simplify what we're publishing to our users.

Historically I've had issues with plugins and the `eslint` package having to be co-located in the same `node_modules` directory. I haven't verified that's the case now with NPM v7, but I suspect this might still be a limitation making it difficult to move them into the sub-packages.

takameyer (approved):

fronck (approved):
:+1:

PR #3853, opened 2021-07-13
Fix TS definition of Realm.Dictionary.remove()

kneth:
## What, How & Why?

A fix of `Realm.Dictionary.remove()`'s TypeScript definition.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -261,9 +261,10 @@ declare namespace Realm {
         set(element:{[key:string]: ValueType}): void;
 
         /**
-         * @returns Removes given element from the dictionary
+         * Removes given element from the dictionary
+         * @returns The dictionary
          */
-        remove(element:{[key:string]: ValueType}): void;
+        remove(key:string|string[]): DictionaryBase<ValueType>;
 
         /**
          * @returns void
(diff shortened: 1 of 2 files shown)


kraenhansen (approved):

types/index.d.ts
```diff
@@ -261,9 +261,10 @@ declare namespace Realm {
         set(element:{[key:string]: ValueType}): void;
```
takameyer, line 262: Is `set` also returning a Dictionary?

kneth: Good catch. Yes, it does.

takameyer: Cool.  The comment also needs to be updated

takameyer (approved):

types/index.d.ts
```diff
@@ -258,12 +258,13 @@ declare namespace Realm {
         /**
          * @returns Adds given element to the dictionary
```
takameyer, line 259:
```suggestion
         * Adds given element to the dictionary
         * @returns The dictionary
```

PR #3854, opened 2021-07-13
Fix integration workflow script

takameyer:
The build ios script invocation is updated to the new syntax.


takameyer:
@kraenhansen Noticed that your lint branch was failing.  My text search tool skipped hidden directories, like `.github`, so I didn't update the build-ios script here.  Small change that will fix the GHA for linting.

kraenhansen (approved):
Great - thanks!

PR #3861, opened 2021-07-17
Fix Integration Test and Linter Pipelines

takameyer:
There seems to be a problem with the android-emulator-runner action.  Using the workaround posted in this issue:
https://github.com/ReactiveCircus/android-emulator-runner/issues/168

The linting pipeline is running ubuntu-latest, which does not have cocoa pods.  One of the lerna dependencies (realm-react-native-tests) was invoking `pod install`.  Changed this to use the npm package `pod-install` which will not crash on non-mac systems.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -236,6 +236,7 @@ jobs:
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: 29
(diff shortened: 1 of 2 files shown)


takameyer:
I will merge this right away to unblock other pull requests.

takameyer:
The linting pipeline will pass after the next linting PR

PR #3867, opened 2021-07-21
Andrew/small things

takameyer:
## What, How & Why?
In the process of my last task, I had written some todos down that I thought I would take care of.

* remove obsolete `packaging` directory
* refactor start-server into its own script with package.json entry


scripts/start-sync-server.sh
```diff
@@ -0,0 +1,36 @@
...
+if [[ -z "$(command -v docker)" ]]; then
+  echo "starting stitch requires docker"
+  exit 1
+fi
+EXISTING_STITCH=$(docker ps | grep "mongodb-realm-test-server.*9090")
+if [[ -n "$EXISTING_STITCH" ]]; then
+  echo "found existing stitch running, not attempting to start another"
+else
+  echo "no existing stitch instance running in docker, attempting to start one"
+  . "${SRCROOT}/dependencies.list"
```
kraenhansen, line 14: I would like if we use the more explicit `source` command instead of the alias ".".
```suggestion
  source "${SRCROOT}/dependencies.list"
```

kraenhansen (changes requested):
Great idea. I have a small optional suggestion.

kraenhansen (approved):

PR #3886, opened 2021-08-04
Fix some errors in Node.js caused by react-native

ibilux:
This React-Native lines caused some problems for Svelte Node.js App.
I had to comment `NativeModules` and `debugHosts, debugPort` avoid errors in `Node.js`.

### DIFF
--- a/lib/browser/index.js
+++ b/lib/browser/index.js
@@ -16,7 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import { NativeModules } from "react-native";
+//import { NativeModules } from "react-native";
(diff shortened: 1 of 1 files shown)


kneth:
Files under `lib/browser` should only be included in React Native apps. The lines you are removing are essential for debugging React Native apps.

Could you explain how they caused you problems?

ibilux:
> Files under `lib/browser` should only be included in React Native apps. The lines you are removing are essential for debugging React Native apps.
> 
> Could you explain how they caused you problems?

I use realm-js in a sveltekit app. It rendered the app as a web app in front-end.
When I used realm-js as a sveltekit backed-end it didn't cause any problem.

PR #3895, opened 2021-08-10
Upgrade to Realm Core v11.2.0.

### LINKED ISSUE #3561: currentUser.profile is not persistent after app restart
`##` Goals

When logging in a user I use metadata which is added to their currentUser.profile property which contain their partitionKey values needed to open a realm.
On app.login() the user object contains the profile property which contains all the needed metadata.
After the app is restarted, the user object persists and is still logged in, but the currentUser.profile property is empty.

## Expected Results

app.currentUser should persist when the app is restarted, including the profile property.

## Actual Results

app.currentUser.profile is empty after app restarts.

## Code Sample
```
import Realm from 'realm';
import {getRealmApp} from './getRealmApp';

const app = getRealmApp();
const AuthContext = React.createContext(null);

const AuthProvider = ({children}) => {
  const [user, setUser] = useState(app.currentUser);

const signIn = async ({username, password}) => {
       getCustomJwt({username, password})
      .then((jwt) => {
        const credentials = Realm.Credentials.jwt(jwt);
        // Authenticate the user
        app
          .logIn(credentials)
          .then((loggedInUser) => {
            setUser(loggedInUser);
          })
          .catch((err) => console.error('Failed to log in', err));
      })
      .catch((err) => console.log(err));
  };

  return (
    <AuthContext.Provider
      value={{
        signIn,
        user,
      }}>
      {children}
    </AuthContext.Provider>
  );
};
const useAuth = () => {
  const auth = useContext(AuthContext);
  if (auth == null) {
    throw new Error('useAuth() called outside of a AuthProvider?');
  }
  return auth;
};

export {AuthProvider, useAuth};
```

- Realm JS SDK Version: 10.1.4
- Node or React Native: React Native
- Client OS & Version: Android 8.0.0
- Which debugger for React Native: Chrome DevTools

kneth:
## What, How & Why?

This closes #3561.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


fronck (approved):
:+1:

PR #3917, opened 2021-08-20
Set the GHA job timeout for RN tests to 60 minutes.

nirinchev:
## What, How & Why?
This adds an explicit timeout for RN test jobs set to 60 minutes to avoid situations where they run forever.


kneth (approved):

takameyer:
Fair change to implement.  Still need to look into why it hangs, but this is a good compromise until we can look into it.

PR #3918, opened 2021-08-23
Disable analytics on Github Actions

kneth:
## What, How & Why?

Disable analytics on Github Actions in order to prevent screwed builder statistics.


kraenhansen (changes requested):
As far as I know, piping the environment variable into `$GITHUB_ENV` is primarily used when the value is generated at runtime. I would rather like to see this change added declaratively as an `env` object on the entire job (just under the `runs-on` key), just like the `DAY_OF_WEEK` on the first example on [the documentation](https://docs.github.com/en/actions/reference/environment-variables).

kneth:
@kraenhansen I have moved setting the environment variable to the top so it will be done only once.

kraenhansen (approved):

takameyer (commented):
Looks good

PR #3929, opened 2021-08-30
Disable analytics on CI

nirinchev:
## What, How & Why?

Most CI services will set the environment variable `CI` to some value, so it's a good idea to check for it and make sure we don't count CI builds as actual developers.

* [Github Actions](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)
* [Travis CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)
* [Appveyor](https://www.appveyor.com/docs/environment-variables/)
* [Jenkins](https://github.com/jenkinsci/jenkins/pull/5370)

Additionally, it modifies the Gradle analytics plugin to not check for the value of `REALM_DISABLE_ANALYTICS` since that is what we do on node/electron and React Native for iOS.

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/react-native/android/analytics.gradle
+++ b/react-native/android/analytics.gradle
@@ -57,8 +57,9 @@ class SendAnalyticsTask extends DefaultTask {
     def sendAnalytics() {
        try {
             def env = System.getenv()
-            def disableAnalytics= env['REALM_DISABLE_ANALYTICS']
-            if (disableAnalytics == null || disableAnalytics != "true") {
+            def disableAnalytics = env['REALM_DISABLE_ANALYTICS'] != null
+            def isCI = env['CI'] != null
+            if (!disableAnalytics && !isCI) {
                 send()
             }
        } catch(all) {}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * <Either mention core version or upgrade>
 * <Using Realm Core vX.Y.Z>
 * <Upgraded Realm Core from vX.Y.Z to vA.B.C>
+* Disable analytics if the `CI` environment variable is set to some value.
 
 10.7.0 Release notes (2021-8-30)
 =============================================================
--- a/lib/submit-analytics.js
+++ b/lib/submit-analytics.js
@@ -95,8 +95,8 @@ function isAnalyticsDisabled() {
     // ignore error
   }
 
-  // Or if the user has specifically opted-out
-  return "REALM_DISABLE_ANALYTICS" in process.env;
+  // If the user has specifically opted-out or if we're running in a CI environment
+  return "REALM_DISABLE_ANALYTICS" in process.env || "CI" in process.env;
 }
 
 function sha256(data) {
--- a/react-native/ios/RealmReact/RealmAnalytics.mm
+++ b/react-native/ios/RealmReact/RealmAnalytics.mm
@@ -227,7 +227,7 @@ static NSDictionary *RLMAnalyticsPayload() {
 }
 
 void RLMSendAnalytics() {
-    if (getenv("REALM_DISABLE_ANALYTICS") || !RLMIsDebuggerAttached()) {
+    if (getenv("REALM_DISABLE_ANALYTICS") || getenv("CI") || !RLMIsDebuggerAttached()) {
         return;
     }
 


react-native/android/analytics.gradle
```diff
@@ -57,8 +57,9 @@ class SendAnalyticsTask extends DefaultTask {
     def sendAnalytics() {
        try {
             def env = System.getenv()
-            def disableAnalytics= env['REALM_DISABLE_ANALYTICS']
-            if (disableAnalytics == null || disableAnalytics != "true") {
+            def disableAnalytics = env['REALM_DISABLE_ANALYTICS'] != null
+            def isCI = env['CI'] != null
+            if (!disableAnalytics && !isCI) {
```
kraenhansen, line 62: This would have the unfortunate side-effect of disabling analytics when `REALM_DISABLE_ANALYTICS=false` which was not the semantics that the previous implementation had, I don't know how we document this, but perhaps this is not what you ment?

nirinchev: It matches the semantics for iOS and Node/Electron so it made sense to align on all platforms:

https://github.com/realm/realm-js/blob/48144ae85fe02d27663101a1fef9c927e6cca29a/lib/submit-analytics.js#L99

https://github.com/realm/realm-js/blob/48144ae85fe02d27663101a1fef9c927e6cca29a/react-native/ios/RealmReact/RealmAnalytics.mm#L230

I feel like just setting `REALM_DISABLE_ANALYTICS` to any value should stop analytics submission. For example, realm-js's own CI workflow sets `REALM_DISABLE_ANALYTICS` to `1` rather than `true`, and I assume many users might feel these are interchangeable. Rather than special truthy and falsy values, it's more straightforward to just check for existence. If a user wants to enable analytics, they should just not set the environment variable.

react-native/android/analytics.gradle
```diff
@@ -57,8 +57,9 @@ class SendAnalyticsTask extends DefaultTask {
     def sendAnalytics() {
        try {
             def env = System.getenv()
-            def disableAnalytics= env['REALM_DISABLE_ANALYTICS']
-            if (disableAnalytics == null || disableAnalytics != "true") {
+            def disableAnalytics = env['REALM_DISABLE_ANALYTICS'] != null
```
kneth, line 60:
```suggestion
            def disableAnalytics = env['REALM_DISABLE_ANALYTICS'] != null || env['REALM_DISABLE_ANALYTICS'] == "true"
```

react-native/android/analytics.gradle
```diff
@@ -57,8 +57,9 @@ class SendAnalyticsTask extends DefaultTask {
...
+            def isCI = env['CI'] != null
```
kneth, line 61:
```suggestion
            def isCI = env['CI'] != null || env['CI'] == "true"
```

react-native/android/analytics.gradle
kneth: For the integration tests, we have https://github.com/realm/realm-js/blob/master/.github/workflows/integration-tests.yml#L4 (`REALM_DISABLE_ANALYTICS` is not `null`).

Moreover, we should set `REALM_DISABLE_ANATYTICS` for all test runner in `Jenkinsfile` (as we do for MacOS: https://github.com/realm/realm-js/blob/master/Jenkinsfile#L486).

react-native/android/analytics.gradle
nirinchev: `!= null` is already stronger than `== true`. If `REALM_DISABLE_ANALYTICS` is set to `true`, then it won't be null.

react-native/android/analytics.gradle
nirinchev: Similarly with `REALM_DISABLE_ANALYTICS`, I think it makes sense to check for existence rather than a particular value. I don't suppose many developers have a `CI` environment variable set on their machines.

kraenhansen (approved):
It's fine by me to simplify the semantics of `REALM_DISABLE_ANALYTICS`. Have you checked if we have to documentation that needs updating too?

nirinchev:
The current docs mention setting an env variable, but are not prescriptive about the value, which would be in line with the changes in this PR:

https://github.com/realm/realm-js/blob/48144ae85fe02d27663101a1fef9c927e6cca29a/lib/submit-analytics.js#L32-L34

https://github.com/realm/realm-js/blob/48144ae85fe02d27663101a1fef9c927e6cca29a/react-native/ios/RealmReact/RealmAnalytics.h#L40-L42

`analytics.gradle` doesn't contain a similar text, but I can add it there for consistency. I don't believe we have docs outside the source code for this.

kneth:
@nirinchev Once https://github.com/realm/realm-js/pull/3928 is merged, you can rebase and the failing CI jobs should be fixed

kneth:
Merging as I don't believe the jest test runner is a real issue

PR #3935, opened 2021-08-31
send extra HTTP headers with mongoClient and function calls

tgerk:
I propose modifying the realm-web package with an extension to Fetcher class to utilize headers provided in any of three
ways:  extension of the App object, extension of the User object, or an option on creating the mongoClient service.

Though it may seem a Typescript heresy, it is perfectly normal to extend Javascript objects, for example:  Object.defineProperties(realmApp.currentUser, {
 "organization": identifier,
 "headers": { get: () => ({
  'X-MyCorp-tenant-organization': identifier
 }) }
}

When that HTTP header is sent with every request, the user's tenant organization should then be accessible from request context in functions, rules, and filters.  The presence or absence of the header can contribute to secure & reasonable behavior.

Looking forward to your feedback!

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kneth:
@tgerk 

packages/realm-web/src/Fetcher.ts
```diff
@@ -64,14 +64,19 @@ function asyncIteratorFromResponseBody(body: unknown): AsyncIterable<Uint8Array>
   }
 }
 
+interface UserWithOptionalHeaders extends User {
+  /** extra headers defined by the user */
+  headers?: { [key: string]: string };
```
takameyer, line 69: I think `headers` should be added directly to the User class.

kraenhansen:
I believe I understand your use-case, you want to associate additional data with your user as they're requesting a MongoDB function, right? Have you tried other ways of associating this data with the user, such as https://docs.mongodb.com/realm/users/enable-custom-user-data or simply adding this as a parameter of the function you're calling? If yes, can you help me understand why this is not sufficient for your use-case?

packages/realm-web/src/User.ts
```diff
@@ -331,7 +331,10 @@ export class User<
   }
 
   /** @inheritdoc */
-  public mongoClient(serviceName: string): Realm.Services.MongoDB {
+  public mongoClient(serviceName: string, options?: { extraHeaders?: { string: string }}): Realm.Services.MongoDB {
```
takameyer, line 334: There should be a `headers` type that can be used by the three introduced changes. 
`type headers = { [key: string]: string }`

packages/realm-web/src/User.ts
```diff
@@ -34,12 +34,12 @@ const DEFAULT_DEVICE_ID = "000000000000000000000000";
...
 
 interface HydratableUserParameters {
-  app: App<any>;
+  app: App<any, any>;
   id: string;
 }
 
 interface UserParameters {
-  app: App<any>;
+  app: App<any, any>;
```
takameyer, line 42: Why were these lines changed?  Was there a typing issue here?

takameyer (commented):
I have a few comments that should be taken into consideration.  We would also appreciate the tests being extended.

packages/realm-web/src/User.ts
tgerk: Yes.  I followed the build process in twice, once as a commit following tag realm-web-v1.3.0 & this was not a problem.  Problem occurred when I cherry-picked my commit onto master branch.  My build process both times was git ```submodule update --recursive; npx lerna bootstrap``` before running build in dir packages/realm-web.  (I did not get past the cmake requirement to do a full build at top-level of the monorepo.)

packages/realm-web/src/User.ts
tgerk: nice, will do.

packages/realm-web/src/Fetcher.ts
tgerk: I see pros and cons, and tensions between OO Typescript and Javascript that I alluded in my commit comment.  Referring to the 'O' in SOLID, my claim is: this feature is not intrinsic to existing App or User classes, but Fetcher only.  In support, ``LocationUrlContext``` is an invented trait/interface created in Fetcher module to overlay the App class; I created this special interface in the same spirit.  Changes in User module had only to do with passing constructor options to Fetcher.   On the con side, I think I agree that the feature should not be hidden & undocumented.

packages/realm-web/src/User.ts
tgerk: 09:31 ~/Code/realm-js >node -v; npm -v
v16.5.0
7.19.1

tgerk:
> I believe I understand your use-case: You want to associate additional data with your user as they're requesting a MongoDB function, right?
> 
> Have you tried other ways of associating this data with the user, such as [custom user data](https://docs.mongodb.com/realm/users/enable-custom-user-data) or simply adding this as a argument of the function you're calling? If yes, can you help me understand why this is not sufficient for your use-case?

Consider that the additional data is associated with the request.  My app is served on a wildcard domain, and via some provisioning service, the domain maps to a tenant id used as an attribute of docs in Atlas collections.  The user (specifically Realm user) may be anonymous & have no custom data.  This is the web-client perspective.

From the Realm application (server-side) perspective, if the HTTP header is absent I can write a data source filter to reliably, succinctly eliminate tenant-owned docs from search.  This is one fail-stupid barrier to augment data security for my tenants.

And additional to the security perspective, developers (myself included) are error-prone & likely somewhere to forget to include a find parameter.  Just as they are likely to work around a complex model class that would have to rewrite every request.  I'm looking for the one-and-done developer experience, I guess.

Note on user custom-data:  my authenticated users could be authorized for none, one, or many organizations so the filter logic could become quite complex.  By mapping a client-contextual organization into the Realm request I think I'm improving user-experience by matching end-user expectations.

tgerk:
> I have a few comments that should be taken into consideration. We would also appreciate the tests being extended.

I added four tests, for each of the three individual sources of headers App/User/mongoClient and the interaction of all three.

packages/realm-web/src/Fetcher.ts
```diff
@@ -64,14 +64,21 @@ function asyncIteratorFromResponseBody(body: unknown): AsyncIterable<Uint8Array>
   }
 }
 
+export type headers = { [key: string]: string };
```
takameyer, line 67: Types and interfaces should be capitalised
```suggestion
export type Headers = { [key: string]: string };
```

packages/realm-web/src/Fetcher.ts
takameyer: One use case that I'm pondering is what if you want to store some header data when currentUser is set?  For example, you want to store a timestamp of when the user logged or establish a non-user specific session ID, in which you want included as header data in your requests.  I think being able to configure the headers for the user object before constructing the fetcher would be desirable in this case.

packages/realm-web/src/Fetcher.ts
tgerk: On second look, found the same Headers type already declared & imported from realm-network-transport.

I chose to re-export it from Fetcher, for use in User without adding to its immediate dependencies.

packages/realm-web/src/Fetcher.ts
```diff
@@ -225,7 +226,7 @@ export class Fetcher implements LocationUrlContext {
     this._extraHeaders = extraHeaders;
   }
 
-  clone(config: Partial<FetcherConfig>) {
+  clone(config: Partial<FetcherConfig>): Fetcher {
```
tgerk, line 229: linter complained this was missing

packages/realm-web/src/Fetcher.ts
tgerk: Couldn't we consider your use case as property of the App instead of currentUser?  The way Fetcher objects are constructed and cloned through the App and User constructors gives it (the Fetcher) a reference to both.  If a property "headers" on User or "extraHeaders" on App exists at call time, Fetcher will forward those with the HTTP request.  The properties could be dynamic, a getter function.  There is no requirement of "before constructing the fetcher"

There is one case where the extra headers are static, this is in call to User.mongoClient (which is a sort of constructor).  The extra headers option becomes a property of the clone of User's fetcher & this is most accessible to the casual user.

I don't have a strong opinion on typing, but suggest adding an optional property (to be defined elsewhere) on Users means the same should be done on App.  Since Fetcher's only knowledge of its App dependency is through LocationUrlContext interface, the "extraHeaders" optional property should remain in the interface, but the UserContextWithOptionalHeaders is redundant.

packages/realm-web/src/Fetcher.ts
```diff
@@ -194,27 +203,31 @@ export class Fetcher implements LocationUrlContext {
   private readonly transport: NetworkTransport;
   private readonly userContext: UserContext;
   private readonly locationUrlContext: LocationUrlContext;
+  private readonly _extraHeaders?: Headers;
```
takameyer, line 206: I would remove the `_` for the `extraHeaders` variable.  One could argue that `_` be used a prefix for private variables, but in that case it should be done everywhere and be a separate PR.

Also, I think the prefix `extra` could also be eliminated.

tgerk: no quarrel with any of that

last push has been rebased onto today's master

takameyer: seems the `_` is still there?

tgerk: For "reasons" switched _my_ default branch to "main", also tripping out on the four-space two-space indent revolution between the branch I'm using here locally (based on realm-web-v1.3.0) and the one I'm pushing to you.  Anyway, I think this problem is addressed correctly now.

packages/realm-web/src/App.ts
```diff
@@ -175,7 +185,7 @@ export class App<
       this.storage.clear();
       // A failed hydration shouldn't throw and break the app experience
       // Since this is "just" persisted state that unfortunately got corrupted or partially lost
-      console.warn("Realm app hydration failed:", err.message);
+      console.warn("Realm app hydration failed:", (err as Error).message);
```
tgerk, line 188: more eslint

tgerk:
CI failed on missing password for a Docker container:  not sure how I can facilitate that.

tgerk:
Sending @kraenhansen a friendly ping.

kraenhansen:
Sorry for keeping this hanging @tgerk. We're really busy getting Hermes support out. I suspect I'll be able to have a proper look at this some time during October.

tgerk:
thanks for the update

tgerk:
I've performed a squash and rebase onto current master (to resolve merge conflicts).

Includes simplified headers implementation in Fetcher compared to earlier reviews.  Now merges headers on each clone.

packages/realm-web/src/User.ts
```diff
@@ -96,6 +96,16 @@ export class User<
...
 
+  /**
+   * an extended property that may be set, removed or defined by others on an object
+   *  instance statically or as a dynamic getter function
+   * HTTP headers to be sent with mongoClient and functions requests
+   *
+   * @type {Headers}
+   * @memberof User
+   */
+  public headers?: Headers;
```
kraenhansen, line 107: This would be the first public and writable property in the API.
My initial though is that we would like this to be a write-only API, through a setter method (`setHeaders` or something ...) instead of a public writable property on `User`.

packages/realm-web/src/App.ts
```diff
@@ -111,6 +111,16 @@ export class App<
...
 
+  /**
+   * an extended property that may be set/removed on this object statically,
+   *  or defined as a dynamic getter function
+   * will be sent with mongoClient and functions requests
+   *
+   * @type {Headers}
+   * @memberof User
+   */
+  public headers?: Headers;
```
kraenhansen, line 122: I would like this to be a `private` member of `App` only available though the configuration passed to the constructor.

tgerk: To be accessible from Fetcher, the property needs to be public in the LocationUrlContext and UserContext interfaces. I propose "public readonly" prop to be set by constructor option in Fetcher and App classes (also `clone` option for Fetcher).  User class will have a public getter, no setter, and utility methods for adding or removing key/value pairs.

tgerk:
_Gotcha_:`Headers` is a type defined by realm-network-transport package & different than Headers from Web Fetch API.  This turns out to be helpful, since context that is dynamic in nature can be accommodated by using generic objects with dynamic **getter** properties.  My `User::setContext` method takes advantage of this.  Its `value` param accepts a function returning a string, as well as just a plain string.

For example, part of my User context is a capabilities JWT that is updated periodically by a 3rd-party SSO service.  I can call ```myRealmUser.setContext('X-MyApp-Capabilites', () => ssoProvider.acquireOrRefreshCapToken(...) )```

PR #3937, opened 2021-09-01
Return value from realm.write

### LINKED ISSUE #2237: Return value from realm.write
Allow returning value from `realm.write` 

E.g.

```js
const user = realm.write(() => {
  return realm.create('User', {})
})
```

If you accept this feature request, I can provide PR.

zepedr0:
## What, How & Why?

Allow returning value from realm.write

E.g.

```js
const user = realm.write(() => {
  return realm.create('User', {})
})
```

This closes #2237 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -779,10 +779,10 @@ declare class Realm {
     removeAllListeners(name?: string): void;
 
     /**
-     * @param  {()=>void} callback
-     * @returns void
+     * @param  {()=>ReturnValueType} callback
+     * @returns {ReturnValueType}
      */
-    write(callback: () => void): void;
+    write<ReturnValueType>(callback: () => ReturnValueType): ReturnValueType;
 
     /**
      * @returns void
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -88,6 +88,8 @@ jobs:
       - run: npm ci --ignore-scripts
       # Bootstrap lerna sub-packages (builds the packages and Realm JS native module)
       - run: npx lerna bootstrap --scope realm-electron-tests --include-dependencies
+        env:
+          npm_config_realm_local_prebuilds: ${{ github.workspace }}/prebuilds
       # Run tests (using xvfb-run to enable opening windows in a virtual framebuffer)
       - name: Run tests (Electron / Main)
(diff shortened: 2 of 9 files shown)


types/index.d.ts
```diff
@@ -779,10 +779,10 @@ declare class Realm {
     removeAllListeners(name?: string): void;
 
     /**
-     * @param  {()=>void} callback
-     * @returns void
+     * @param  {()=>ReturnValueType} callback
+     * @returns {ReturnValueType}
      */
-    write(callback: () => void): void;
+    write<ReturnValueType>(callback: () => ReturnValueType): ReturnValueType;
```
kraenhansen, line 785: Does this `ReturnValueType` need a default value or is it able to infer this entirely from the callback? I would have thought you needed to use the `infer` keyword or `ReturnType` helper type or at least `write<ReturnValueType = unknown>(...)`.

takameyer: I worked on this with José.  The type of the return value is inferred from the callback return value (we tested this).  That being said, the default value should probably be set to unknown (the value is just typed as `void` if nothing is returned).

takameyer (approved):
👍🏼

takameyer:
As in the other PR, the changelog should be updated.

types/index.d.ts
kraenhansen: I just tested it in the TS playground - it works as I expect it to. Good job!
https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwHcYsMQAeAJQD4AKMKCCAIyjAGsAueGgSngF4q8Cjy4UA3AChJGAJ4AHBAFUAziBgD4Ab0nw98VFAC2ILiozFUAcykBfaWDzn4yNRv6FipGrwFCd+vCOqM6u6lyq6poBgfqGJlwARABSOAAWqIkANLr6tlKBcBjIMPhhMHY8UpLBKjgQIAB0EDhWNOWN8SBVkkA

kneth (approved):

fronck (approved):

kraenhansen (approved):
Much better :)

takameyer (approved):

PR #3938, opened 2021-09-01
Get Realm.Object property type

### LINKED ISSUE #3646: Mixed: Get underlying type of a mixed property
Since all numeric types (int, long, double) are represented as a `number` when read from our SDK, it makes sense to implement a method to get the underlying type of the property value, if your application (such as MongoDB Realm Studio) care about this.

The API could be an instance method on `Realm.Object`:

```js
realmObject.getType(propertyName);
```

`getType()` implies it works for any property (and not just Mixed properties).

zepedr0:
## What, How & Why?

Since all numeric types (int, long, double) are represented as a `number` when read from our SDK, it makes sense to implement a method to get the underlying type of the property value, if your application (such as MongoDB Realm Studio) cares about this. Throws an exception if a property with the given name does not exist.

E.g.

```js
realmObject.getPropertyType(propertyName);
```

This closes #3646 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/src/js_realm_object.hpp
+++ b/src/js_realm_object.hpp
@@ -79,6 +79,7 @@ struct RealmObjectClass : ClassDefinition<T, realm::js::RealmObject<T>> {
     static void add_listener(ContextType, ObjectType, Arguments &, ReturnValue &);
     static void remove_listener(ContextType, ObjectType, Arguments &, ReturnValue &);
     static void remove_all_listeners(ContextType, ObjectType, Arguments &, ReturnValue &);
+    static void get_property_type(ContextType, ObjectType, Arguments &, ReturnValue &);
 
     static void get_realm(ContextType, ObjectType, ReturnValue &);
 
@@ -101,6 +102,7 @@ struct RealmObjectClass : ClassDefinition<T, realm::js::RealmObject<T>> {
         {"addListener", wrap<add_listener>},
         {"removeListener", wrap<remove_listener>},
         {"removeAllListeners", wrap<remove_all_listeners>},
+        {"getPropertyType", wrap<get_property_type>},
     };
 
     PropertyMap<T> const properties = {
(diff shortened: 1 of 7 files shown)


kraenhansen (approved):
Great job! Feel free to merge as soon as the tests are green.

tests/js/object-tests.js
```diff
@@ -724,4 +724,36 @@ module.exports = {
       }, 2000);
     });
   },
+
+  testGetPropertyType: function () {
+    const realm = new Realm({
+      schema: [schemas.AllTypes, schemas.TestObject, schemas.LinkToAllTypes],
```
kraenhansen, line 730: I would love to see a test using a property of type `mixed` as well. To verify that it's the underlying type and not simply the type from the schema which gets returned.

bmunkholm:
This also needs an updated changelog.md

src/js_realm_object.hpp
```diff
@@ -415,6 +417,24 @@ void RealmObjectClass<T>::remove_all_listeners(ContextType ctx, ObjectType this_
...
+    if (!realm_object) {
+        throw std::runtime_error("Invalid 'this' object");
+    }
+
+    const Property* prop = realm_object->get_object_schema().property_for_public_name(property_name);
+    if (!prop) {
+        throw std::invalid_argument(util::format("No such property: %1", property_name));
+    }
+
+    return_value.set(prop->type_string());
```
kneth, line 436: To get the JavaScript names, you need to call `TypeErrorException::type_string(prop)` (https://github.com/realm/realm-js/blob/master/src/js_realm.cpp#L78).

If `prop->type_string() == "mixed"`, you need to unwrap the property. It is something like:

```
auto obj = realm_object->obj();
DataType typ = obj.get().get_type();
// convert typ to string
```

takameyer (approved):
👍🏼

kraenhansen (approved):

PR #3953, opened 2021-09-09
Fix for ios integration tests

takameyer:
The cache key syntax was incorrect in a way that all platforms were sharing the same cache.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kneth (approved):
LGTM

PR #3986, opened 2021-09-28
Create github workflow to build and publish jsdocs (#3762)

### LINKED ISSUE #3762: Automate upload API doc when releasing
As part of releasing Realm JS, we should upload the API doc as a CI step. It can be done in the JSDoc step: https://github.com/realm/realm-js/blob/v10/Jenkinsfile#L85-L94

iamer:
The workflow is triggered for PRs that change documentation files as well as for
major and minor release tags (vX.Y.0).
Additionally for tags the documentation is published to an S3 bucket under
a direcotry with the version as its name.

For uploading to work, three secrets need to be configured: DOCS_S3_ACCESS_KEY,
DOCS_S3_SECRET_KEY and DOCS_S3_BUCKET_NAME

## What, How & Why?

This closes #3762

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/jsdoc.yml
+++ b/.github/workflows/jsdoc.yml
@@ -1,0 +1,45 @@
+name: JSDoc
+on:
+  pull_request:
+    paths:
+      # documentation
+      - "docs/**"
+      - "*.md"
+      # changes to this workflow
+      # seems there is no way to get this easily yet https://github.com/actions/runner/issues/853
+      - ".github/workflows/jsdoc.yml"
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+env:
+  REALM_DISABLE_ANALYTICS: 1
+jobs:
+  jsdoc:
+    name: JSDoc
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: "recursive"
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+          registry-url: https://registry.npmjs.org/
+      - name: Install npm v7
+        run: npm install -g npm@7
+      # Install the root package (--ignore-scripts to avoid downloading or building the native module)
+      - name: Install root package dependencies
+        run: npm ci --ignore-scripts
+      - name: Run JSDoc documentation build
+        run: npm run jsdoc
+      - name: Configure AWS Credentials
+        if: startsWith(github.ref, 'refs/tags/v')
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.DOCS_S3_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/jsdoc.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          node-version: 14
+          registry-url: https://registry.npmjs.org/
+      - name: Install npm v7
+        run: npm install -g npm@7
+      # Install the root package (--ignore-scripts to avoid downloading or building the native module)
+      - name: Install root package dependencies
+        run: npm ci --ignore-scripts
+      - name: Install subpackages dependencies
+        run: npm run lerna:bootstrap
```
kraenhansen, line 35: Are you sure this is needed?

```suggestion
```

.github/workflows/jsdoc.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Configure AWS Credentials
+        if: startsWith(github.ref, 'refs/tags/v')
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.DOCS_S3_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.DOCS_S3_SECRET_KEY }}
+          aws-region: us-west-1
+      - name: Upload docs
+        if: startsWith(github.ref, 'refs/tags/v')
+        run:  aws s3 sync --acl public-read docs/output/realm s3://${{ secrets.DOCS_S3_BUCKET_NAME }}
```
kraenhansen, line 47:
```suggestion
        run: aws s3 sync --acl public-read docs/output/realm s3://${{ secrets.DOCS_S3_BUCKET_NAME }}
```

kraenhansen: I'm wondering if `DOCS_S3_BUCKET_NAME` need to be considered a secret?

kraenhansen: Also, doesn't destination path need to contain the current package version?

iamer: Bucket name was easier to configure as a variable. I would need to know the bucket name to hardcode it.
The documentation is already built in a directory named as current package version, so the step uploads it as is.
For example docs/output/realm/10.2.0 is uploaded to 10.2.0

.github/workflows/jsdoc.yml
iamer: I had copied this from another workflow, I will remove it.

kneth (approved):

kraenhansen:
> Wouldn't that upload docs on PRs even though things haven't been released?

I was under the impression that tagging a commit with a version communicates that this has (or will very soon) be published.
To me CI should perform the tagging of the commit as a part of the actions performed when publishing a release and this workflow could then pick up that signal and publish the docs accordingly.

PR #4020, opened 2021-10-12
Make logo dark-mode friendly

takameyer:
## What, How & Why?

Added SVG in the logo that detects dark-mode preference from the users system settings and changes the text color based on said preference.
Unfortunately, if the user has their system setting to "dark" and their github preferences to "light", it will display the wrong fill color.  I assume this should work for 99% of the vistors reading the Readme.


takameyer:
> It's not a showstopper, but this doesn't appear to work on Safari 😭
> 
> Safari
> 
> Screen Shot 2021-10-12 at 23 02 20
> 
> Chrome
> 
> Screen Shot 2021-10-12 at 23 02 24

Oh no! It seems safari does not support this feature in SVGs embedded inside an .  So close!

kneth:
I have approved the change. It is not a full-proof solution but it is an improvement to what we have now.

PR #4027, opened 2021-10-14
Deferred compilation on Android

kraenhansen:
## What, How & Why?

The React Native project doesn't promise ABI / API stability of any shared object libraries shipped via the React Native package. In addition to this the end-user determines the NDK (hence the shared C++ STL) used when building the their app.

In principle this means that we can only publish a prebuilt shared object library file under the following assumptions:
- Our prebuilt binary is installed alongside a specific subset of React Native versions (any version ABI compatible with the version we used when compiling our prebuilt binary).
- The end-users app is compiled with the same NDK version that we used when building our prebuilt binary. This ensures ABI stability with the shared object C++ STL shipped with the app.

In practice however:
- The breakage of the ABI might happen on the next release or ten releases from now, nobody really knows. We might be putting in a lot of work to build resilience towards a scenario that in practice rarely happens. Historically we've been "lucky" that we've mostly been integrating with ABIs of the Android platform and haven't had deep integrations with ABIs exposed by React Native. With the upcoming changes to support Hermes via JSI, we are not allowed that luxury.
- We can always upgrade the React Native version, recompile and publish a new version of our library as a version of React Native ships with an ABI breaking change. If this happens a lot, it would be a bummer for our users (and our documentation and support crew) that the have to constantly know (possibly by NPM peerDepenency specifications or an NPM post-install script) what versions of Realm JS are compatible with which versions of React Native.

The goal of this PR is to increase resilience towards breaking ABI & API changes of the React Native JavaScript Interface (JSI) on Android. I'm putting this into draft for the time being, as it does complicated our build process and developer experience significantly and puts extra requirements on the end-users setup:
- An expected version CMake via the SDK manager.
- ~Building the app with an alpha release of the Android Gradle plugin. Although I still need to verify that this is the case when building Realm Core from prebuilt binaries.~ This doesn't seem to be an issue anymore (see "Bump Gradle version to 7.2, Bump Kotlin version to 1.5.31 ([9ae3367431](https://github.com/facebook/react-native/commit/9ae3367431428748f5486c782199beb4f9c6b477) by [@svbutko](https://github.com/svbutko))" in the [React Native Changelog](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#android-specific-12))

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen:
Closing this as https://github.com/realm/realm-js/pull/6650 is realizing the same goal 🎉

Issue #3723, opened 2021-05-07
Reset password doesn't work in realm-web


jeanbaptistebeau:
## Goals

Reset the password of a user.

## Expected Results

The `resetPassword` function completes without errors and the password is reset.

## Actual Results

Getting the error `invalid token data`.

## Steps to Reproduce

- Replace the appID and the user email in the code below
- Run the script
- Copy paste the token and token ID you received by email in the script below
- Comment the `callResetPasswordFunction` and uncomment the `resetPassword`
- Run the script again

## Code Sample

```
const Realm = require("realm");

const appID = "appID";
const app = new Realm.App({ id: appID, timeout: 10000 });

const tokenId = "someTokenID";
const token = "someToken";

app.emailPasswordAuth.callResetPasswordFunction("someUserEmail", "password", []).then(
	(value) => {
		console.log("SUCCESS");
	},
	(error) => {
		console.log(error);
	}
);

// app.emailPasswordAuth.resetPassword("password", token, tokenId).then(
// 	(value) => {
// 		console.log("SUCCESS");
// 	},
// 	(error) => {
// 		console.log(token);
// 		console.log(tokenId);
// 		console.log(error);
// 	}
// );

```
## Version of Realm and Tooling

- Realm JS SDK Version: 10.0.0-rc.2
- Node or React Native: Node 14.9.0
- Client OS & Version: MacOS Big Sur 11.3 Beta

Initially tracked here: https://developer.mongodb.com/community/forums/t/consistently-getting-invalid-token-data-when-trying-to-reset-user-password/100020/4

jeanbaptistebeau:
EDIT: Updated to Realm JS SDK Version 10.4.0 and it works. 

It doesn't work when in a webscript script imported using: 
`<script src="https://unpkg.com/realm-web@1.2.0/dist/bundle.iife.js"></script>`

jeanbaptistebeau:
I get `invalid token data` error

jeanbaptistebeau:
Any news about that? Is it a real bug or am I doing something wrong?

kushpetal:
Same issue here. Getting the same error.

ryan-byrne:
For folks still having this issue, the documentation had the **wrong order of arguments**
```js
app.emailPasswordAuth.resetPassword('newPassword', token, tokenId)
```
Should be:
```js
app.emailPasswordAuth.resetPassword(token, tokenId, 'newPassword')
```
Which fixed the issue for me

[...]

PR #4041, opened 2021-10-19
Add consistent object-based API for `EmailPasswordAuth` methods, deprecate positional argument API

### LINKED ISSUE #3943: Easy and consistent API for emailPassword methods
As reported in #3723, the API of `Realm.Auth.EmailPasswordAuth.resetPassword()` is not consistent between Realm Web and Realm JavaScript. Moreover, the TypeScript definition says on thing, and the API documentation says something else. The root cause is positional arguments (all strings) which is error prone for both our users and for ourself.

We should deprecate all current methods of `Realm.Auth.EmailPasswordAuth` (to be removed in version 11) and implement a new variant of the API:

```javascript
callResetPasswordFunction({ email: string, password: string }, args: [BSON])
confirmUser({ token: string, id: string })
registerUser({ email: string, password: string })
resendConfirmationEmail({ email: string }) // for consistency
resetPassword({ password: string, token: string, id: string })
sendResetPasswordEmail({ email: string }) // for consistency
```

Methods should log a warning if called with deprecated signature.

tomduncalf:
## What, How & Why?

This PR introduces a consistent, object/dictionary-based API for the `EmailPasswordAuth` methods, instead of using positional arguments, which can be error-prone and unclear. This closes #3943 and #3723.
 
The existing API is supported until the next major version, but will output deprecation warnings to the console when called in development mode of the form: `Deprecation warning from Realm: resetPassword(password, token, token_id) is deprecated and will be removed in a future major release. Consider switching to resetPassword({ password, token, token_id }).`

As the same logic is required to handle the deprecation in multiple places across `realm` and `realm-web`, and we may wish to deprecate more positional argument based APIs in future, a shared function `handleDeprecatedPositionalArgs` has been introduced, in a new Lerna package `@realm.io/common`. This function allows handling both the new and old forms of arguments in a consistent way, until the old way is deprecated at which point we can remove the call (see https://github.com/realm/realm-js/pull/4041#discussion_r733478780). This package can serve as a place to store other common code which does not warrant its own dedicated package between the two platforms in future.

### `registerUser`

Old:
```js
registerUser('john@example.com', 'some-secure-password');
```

New:
```js
registerUser({ email: 'john@example.com', password: 'some-secure-password' });
```

### `confirmUser`

Old:
```js
confirmUser('someToken', 'someTokenId');
```

New:
```js
confirmUser({ token: 'someToken', tokenId: 'someTokenId' })
```

### `resendConfirmationEmail`

Old:
```js
resendConfirmationEmail('john@example.com');
```

New:
```js
resendConfirmationEmail({ email: 'john@example.com' });
```

### `retryCustomConfirmation`

Old:
```js
retryCustomConfirmation('john@example.com');
```

New:
```js
retryCustomConfirmation({ email: 'john@example.com' });
```

### `sendResetPasswordEmail`

Old:
```js
sendPasswordEmail('john@example.com');
```

New:
```js
sendPasswordEmail({ email: 'john@example.com' });
```

### `resetPassword`

Old:
```js
resetPassword('some-secure-password', 'someResetToken', 'someResetTokenId');
```

New:
```js
resetPassword({ password: 'some-secure-password', token: 'someResetToken', tokenId 'someResetTokenId' });
```

### `callResetPasswordFunction`

Old:
```js
callResetPasswordFunction(
  'john@example.com', 
  'some-secure-password', 
  'extraArgToFunction1', 
  'extraArgToFunction2'
);
```

New:
```js
callResetPasswordFunction(
  { 
    email: 'john@example.com', 
    password: 'some-secure-password' 
  }, 
  'extraArgToFunction1', 
  'extraArgToFunction2'
);
```

## Questions:
- Do we want to only log warnings in development mode, or should we do it in all modes?
- Should we mention something in the Compatibility section?

## Notes:
- CI is failing until the `@realm.io/common` package is published, which I won't do until it has been reviewed. This is because currently our CI runs `npm install` before running Lerna, which pulls in code from npm only. I had a hacky(?) workaround in https://github.com/realm/realm-js/pull/4041/commits/c9d4bcef56090507a2ced962dab18539065b0a4b but after discussion with Kræn, we decided the real solution is to move `realm` to a Lerna package also, so for now we can accept having to first publish the package before CI passes.

## ☑️ ToDos
Todo:
* [x] Add workflow for running package tests
* [x] 📝 Changelog entry
* [x] 🚦 Tests
* [x] Test in sample apps etc
  * https://github.com/mongodb/docs-realm/pull/1438

* [x] jsdoc files updated


packages/realm-common/src/environment.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Set the implementation used for `isDevelopmentMode`. This allows each entry point
+ * (node vs DOM) to specify an appropriate implementation, while maintaing a common
+ * interface.
+ *
+ * @param impl A function returning a boolean indicating whether the user's app is running
+ * in development mode or not.
+ */
+export const setIsDevelopmentModeImpl = (impl: typeof isDevelopmentModeImpl): void => {
```
kraenhansen, line 39: Is there a reason you choose a `callback` pattern over simply setting a boolean on import?

.github/workflows/package-unit-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+      - "src/**"
+      - "types/**"
+      - "react-native/**"
+      - "vendor/**"
+      # Integration tests
+      - "integration-tests/**"
+      # No need to run when updating documentation
+      - "!**.md"
+      # Run this on updates to the workflow
+      - ".github/workflows/integration-tests.yml"
```
kraenhansen, line 22: These paths looks like they we're copied 1-to-1.

```suggestion
      # Source code
      - "packages/realm-network-transport/**"
      - "packages/realm-common/**"
      # No need to run when updating documentation
      - "!**.md"
      # Run this on updates to the workflow
      - ".github/workflows/package-unit-tests.yml"
```

packages/realm-common/src/environment.ts
tomduncalf: Hm, good point, the development mode status is not going to change in a given session. I'll simplify it

integration-tests/environments/react-native/package.json
```diff
@@ -53,6 +53,8 @@
     "realm-app-importer": "*"
   },
   "localDependencies": {
+    "@realm.io/common": "../../../packages/realm-common",
+    "realm-network-transport": "../../../packages/realm-network-transport",
```
kraenhansen, line 57: This shouldn't be needed (or they should be removed again) after https://github.com/realm/realm-js/pull/4054 merge (depending on what happens first 😅 )

tomduncalf: 👍  I am just reviewing that one now so if we merge that first I can update this after!

packages/realm-common/src/environment.ts
```diff
@@ -16,29 +16,26 @@
...
+export const isDevelopmentMode = (): boolean => isDevelopmentModeImpl();
 
-  if (typeof window !== "undefined") {
-    // We are in a browser, check if we are in react dev mode
-    return typeof global.__DEV__ !== "undefined" && global.__DEV__;
-  } else {
-    // We are not in a browser, check if NODE_ENV is anything other than production
-    return process.env?.NODE_ENV !== "production";
-  }
+/**
```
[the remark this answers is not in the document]
kraenhansen: Is there a reason you choose a callback pattern over simply setting a boolean on import?

PR #4050, opened 2021-10-29
Colorized compiler diagnostics if using Ninja

kneth:
## What, How & Why?

Adding colorized compiler diagnostics.

Screenshot 2021-10-29 at 12 00 43

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(CheckCXXCompilerFlag)
+
 # compiling for Node.js, need to set up toolchains before project() call
 if(DEFINED CMAKE_JS_VERSION)
     if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -29,11 +31,20 @@ endif()
 cmake_minimum_required(VERSION 3.15)
 project(RealmJS)
 
-# Use ccache if available
-find_program(CCACHE_PROGRAM ccache)
-if(CCACHE_PROGRAM)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
-endif()
```
kraenhansen, line 36: Why would this not break the use of ccache?

kneth: It looks like a mistake.

CMakeLists.txt
```diff
@@ -49,6 +66,13 @@ if(DEFINED CMAKE_JS_VERSION)
     include(NodeJSTargets)
 endif()
 
+if (NOT MSVC)
+    # Ninja buffers output so we need to tell the compiler to use colors even though stdout isn't a tty.
+    if("${CMAKE_GENERATOR}" STREQUAL "Ninja")
```
kraenhansen, line 71: Is there any reason this needs to be wrapped in a string template?
```suggestion
    if(CMAKE_GENERATOR STREQUAL "Ninja")
```

fronck (approved):
:+1:

PR #4054, opened 2021-10-29
Avoid `local-install` and rely on `metro.config.js` instead

kraenhansen:
## What, How & Why?

Inspired by https://github.com/realm/realm-js/pull/3883 this PR uses a more comprehensive approach to resolve linked dependencies (generated by Lerna) and their peer dependencies.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests


integration-tests/environments/react-native/metro.config.js
```diff
@@ -20,36 +20,81 @@ const path = require("path");
...
-  ]),
-);
-
 function readJson(...pathSegemnts) {
   const filePath = path.join(...pathSegemnts);
   return JSON.parse(fs.readFileSync(filePath, "utf8"));
 }
 
-const watchFolders = Object.values(localDependencies);
+function getLinkedDependencies(packagePath, recursive = true, exclude = new Set()) {
```
tomduncalf, line 28: Could we add a high level description of what this process is doing/why it exists in a comment, as it's not necessarily obvious? Just to make it clear for anyone new to the codebase, without having to read the details. Something like

```
// Metro bundler does not support symlinks. In order to link to our local version 
// of Realm JS in a way which Metro can follow, we need to instead add the module 
// to `resolver.blockList`, and then in `resolver.extraNodeModules`, override the 
// locations for any symlinked modules by specifying a mapping from their name 
// to their actual location on disk.

// In order to achieve this, we recursively check if each file/directory in `node_modules` 
// is a symlink or not. If it is, and if it is specified in our main `package.json` (i.e. 
// it is a direct dependency of our app), we add it to the `blockList` and 
// add a mapping for it. We also add it to the list of watched folders, so
// that any changes to it are picked up by hot reloading.
```

kneth (approved):

tomduncalf (approved):
Looks like a nicer solution!

integration-tests/environments/react-native/metro.config.js
kraenhansen: > Do you think it would make sense to move this logic to somewhere shared.

Definitely. If other apps would be copy-pasting this, it would probably make more sense to copy it somewhere shared first. But I would cross that bridge once another of our apps adopts this.

tomduncalf: > But I would suggest crossing that bridge once another of our apps adopts this.

Makes sense! Maybe @takameyer's React test apps could switch to making use of this for consistency?

integration-tests/environments/react-native/metro.config.js
```diff
@@ -16,40 +16,96 @@
...
+  const links = files
+    .filter((f) => f.isSymbolicLink())
+    .map(({ name }) => {
+      const linkPath = fs.realpathSync(path.join(nodeModulesPath, name));
+      const linkPackageJson = readJson(linkPath, "package.json");
+      const peerDependencies = Object.keys(linkPackageJson.peerDependencies || {});
+      return { name, path: linkPath, peerDependencies };
+    });
+  // We're only interested in actual dependencies
+  const dependencyLinks = links.filter(({ name }) => name in packageJson.dependencies && !exclude.has(name));
```
takameyer, line 53: In `realm-react`, I have some conflicts with `devDependencies` as well, since the tests make use of the `babel-runner` and `react-native`. Might be worthwhile to check `name in packageJson.devDependencies` as well.

integration-tests/environments/react-native/metro.config.js
```diff
@@ -16,40 +16,96 @@
...
+function getLinkedDependencies(packagePath, recursive = true, exclude = new Set()) {
+  const packageJson = readJson(packagePath, "package.json");
+  const nodeModulesPath = path.join(packagePath, "node_modules");
+  const files = fs.readdirSync(nodeModulesPath, { encoding: "utf8", withFileTypes: true });
+  const links = files
+    .filter((f) => f.isSymbolicLink())
+    .map(({ name }) => {
+      const linkPath = fs.realpathSync(path.join(nodeModulesPath, name));
+      const linkPackageJson = readJson(linkPath, "package.json");
+      const peerDependencies = Object.keys(linkPackageJson.peerDependencies || {});
```
takameyer, line 49: Might be worthwhile to grab the `optionalDependencies` as well

integration-tests/environments/react-native/metro.config.js
```diff
@@ -16,40 +16,96 @@
...
-  ]),
-);
-
 function readJson(...pathSegemnts) {
   const filePath = path.join(...pathSegemnts);
   return JSON.parse(fs.readFileSync(filePath, "utf8"));
 }
 
-const watchFolders = Object.values(localDependencies);
+function getLinkedDependencies(packagePath, recursive = true, exclude = new Set()) {
```
takameyer, line 40: Do we need the `recursion` flag?  As this is not a utility function being used for multiple things, I think having support for `recursion` just complicates the code.

integration-tests/environments/react-native/metro.config.js
takameyer: @tomduncalf Possibly, yes.  I actually am not using symbolic links.  I will have to test this out.

integration-tests/environments/react-native/metro.config.js
```diff
@@ -16,40 +16,96 @@
...
-// Block any node_module loaded from the local dependency, outside of the app's node_modules
-const blockList = Object.values(localDependencies).map(
-  (localPath) => new RegExp(`^${escape(path.join(localPath, "node_modules"))}\\/.*$`),
-);
+const blockedPaths = [];
+for (const pkg of linkedDependencies) {
+  // Block the links
+  blockedPaths.push(path.join(__dirname, "node_modules", pkg.name));
+  // Block any peer dependencies of the links
+  blockedPaths.push(...pkg.peerDependencies.map((peer) => path.join(pkg.path, "node_modules", peer)));
```
takameyer, line 77: `optionalDependencies` here as well

takameyer (approved):
I approve.  Just a couple considerations, which will make this script probably work in `realm react` as well.

takameyer:
There are references to `install-local` in the `README.md` of `integration-tests/environments/react-native` that should be updated.

integration-tests/environments/react-native/metro.config.js
kraenhansen: The default is actually to recurse (as that's needed for our purpose). For example Realm JS has a symbolically linked dependency on `realm-network-transport`, which we also want to resolve. I do agree that since we're always recursing, having that as an argument is overkill (although it was useful when debugging).

integration-tests/environments/react-native/metro.config.js
kraenhansen: I would like to understand your use-case for using `optionalDependencies`. In most cases what I actually want to use is `peerDependencyMeta` to mark individual `peerDependencies` optional: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

kraenhansen:
I suggest merging this when CI is green and postpone the tasks needed to reuse this in other packages for another PR (which does exactly that).

integration-tests/environments/react-native/metro.config.js
takameyer: Ah, I overlooked the default value.  I suppose if it has some use, we can keep it in.

integration-tests/environments/react-native/metro.config.js
takameyer: I will attempt to add this to my project and we can go over it when I get it working.  I'm just looking ahead at probable hangups.  I have a devDependency on `react-native` which is an `optionalDependency` for anything using that module.  In that case, I want to resolve any conflicts there.

takameyer: However, it is possible this is just a workaround for my problem.  Maybe there is a smarter way to resolve this.  Lets not hold up this PR with that though

kraenhansen:
Merging since we're only waiting for the Node.js tests on Windows.

PR #4061, opened 2021-11-02
Removed unneeded files from our archive

### LINKED ISSUE #4060: Remove source files from NPM package
Since we're shipping prebuilt binaries for all supported platforms, there's no reason to include the `src`, `vendor` directories nor any build scripts / cmake configurations.

This was mentioned as an a-aside in a comment on another PR https://github.com/realm/realm-js/pull/3626#issuecomment-825608686

I wouldn't consider this a breaking change, since all public documentation I can find instructs users to `git clone` our repository to build from source.

kraenhansen:
## What, How & Why?

This closes #4060 cutting our NPM unpacked archive by 30 MB from ~320 MB to ~290 MB.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
I've drafted this again, since it looks like we're relying on building from source in our tests: https://github.com/realm/realm-js/blob/kh/less-files-in-archive/scripts/test.sh#L453

PR #4070, opened 2021-11-05
Enable CMake compile_commands.json and add appropriate VS Code config and setup instructions

tomduncalf:
## What, How & Why?

This change should enable better (not perfect, but better!) Intellisense for C++ code in VS Code. It tells CMake to generate a `build/compile_commands.json` file, which the [clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) extension can parse (it needs to be symlinked in the root directory for it to pick it up).

image

image

image

image

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


contrib/building.md
```diff
@@ -168,6 +169,14 @@ mklink /j assets "../../../../../data"
 
 Note: If you have cloned the repo previously make sure you remove your `node_modules` directory since it may contain stale dependencies which may cause the build to fail.
 
+### Visual Studio Code setup for C++
+
+If you are using Visual Studio Code as your editor, you can get greatly improved C++ Intellisense by installing the [ccls](https://marketplace.visualstudio.com/items?itemName=ccls-project.ccls) extension.
+
+This should pick up the configuration in `.vscode/c_cpp_properties.json`, which references the `build/compile_commands.json` file generated by CMake, enabling full Intellisense.
+
+Other editors should also be able to be configured to use th `compile_commands.json` file.
```
kraenhansen, line 178:
```suggestion
Other editors should also be able to be configured to use the `compile_commands.json` file.
```

contrib/building.md
```diff
@@ -168,6 +169,14 @@ mklink /j assets "../../../../../data"
 
 Note: If you have cloned the repo previously make sure you remove your `node_modules` directory since it may contain stale dependencies which may cause the build to fail.
 
+### Visual Studio Code setup for C++
```
kneth, line 172: Do you need the configuration mentioned in https://code.visualstudio.com/docs/cpp/config-clang-mac? Or does `clangd` handle it for you?

kneth (approved):

contrib/building.md
tomduncalf: I think the `clangd` extension should handle this, it's independent of the MS C++ extension as I understand it... I have the MS Intellisense disabled and it seems to work but if you could verify on your setup that would be great!

contrib/building.md
```diff
@@ -168,6 +169,14 @@ mklink /j assets "../../../../../data"
...
+
+If you are using Visual Studio Code as your editor, you can get greatly improved C++ Intellisense by installing the [clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) extension. This should prompt you to disable the built in C++ extensions Intellisense, but if not you should do this in Settings, searching for `cpp intellisense`.
```
blagoev, line 174: why not add this as a suggestion extension in VS Code for this repo

tomduncalf: I had no idea that was possible! Nice, will update 👍

tomduncalf:
Going ahead with merge as it is only the glitchy iOS test failing

PR #4078, opened 2021-11-10
Retrying launching the iOS integration tests

kraenhansen:
## What, How & Why?

This is an attempt at stabilizing the React Native integration tests running on iOS by retrying the build / lunch command if tests doesn't complete the tests within 5 minutes from the previous build command invocation exits.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


tomduncalf (approved):
Seems reasonable until we determine the root cause

PR #4082, opened 2021-11-16
Automated Client Reset with Discard Local behavior

### LINKED ISSUE #3699: Automatic Client Reset w. “seamless loss”
#### Provide a less disruptive client reset experience with a new “seamless loss” mode that will reset the clients to the state of the server, but will lose local client changes.
* [x] API design
* [x]Add parsing of sync configuration (C++)
* [x] Add functor classes for the two new call backs (C++)
* [x] API doc
* [x] TS defs
* [x] Rebase on master
* [x] Update to latest core commit
* Tests
  * [x] Simulate client reset (discard local)
  * [x] Simulate failure on client reset
  * [x] Update usage of _simulateError()

PR: [https://github.com/realm/realm-js/pull/4082](https://github.com/realm/realm-js/pull/4082)

kneth:
## What, How & Why?

This closes #3699 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,23 @@
+
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Added templates for Expo.
+* A new mode `discardLocal` for client reset is introduce. The old behavior is supported (but deprecated) through the `manual` mode. The new mode will discard any local changes, and a fresh copy of the Realm will be downloaded. An example of the configuration:
+
+```js
+const config = {
+  // ...
+  sync: {
+    // ...
+    clientReset: {
+      mode: "discardLocal",
+      clientResyncBefore: (localRealm, remoteRealm) => { /* ... */ },
+      clientResyncAfter: (localRealm) => { /* ... */ },
+    },
+  },
+};
+```
 
 ### Fixed
 * If the option `user` in a sync configuration was not a `Realm.User` object could lead to a crash. ([#1348](https://github.com/realm/realm-js/issues/1348), since v10.0.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,11 @@ const config = {
 * Removed undocumented, outdated and unused `ListView` component exported via `realm/react-native`. See [@realm.io/react](https://www.npmjs.com/package/@realm.io/react) for a modern integration with React.
 * Sending the correct version when submitting analytics requests on Android, as a side-effect of [#4114](https://github.com/realm/realm-js/pull/4114/files#diff-58e627175b916d5dcb05c3c8bd5b10fb18cd72ce6e40e41d8d1c51e984fe32e8L73-R73).
 * Removed TypeScript dependency from individual sub-packages so all packages share the root version, and updated the root TypeScript version to `4.5.2`.
+<<<<<<< HEAD
 * Consuming TypeScript directly from the integration test environments, instead of transpiling first.
 * Adding a new private `@realm/metro-config` package to share this across any React Native app in our repo that reference other packages via symbolic links.
+=======
+>>>>>>> dc9b6ef61 (Upgrade to Realm Core v11.7.0)
 * Upgraded Realm Core from v11.6.1 to v11.7.0.
 
 10.10.1 Release notes (2021-11-18)
@@ -584,8 +587,12 @@ NOTE: This release has a number of fixes compared to v10.3.0-rc.1. For a complet
 * File format: generates Realms with format v20 (reads and upgrades file format v5 or later for non-synced Realm, upgrades file format v10 for synced Realms).
 
 ### Internal
+<<<<<<< HEAD
 * Upgraded Realm Core from v10.1.3 to v10.1.4.
 * Upgraded Realm Sync from v10.1.4 to v10.1.5.
+=======
+* None.
+>>>>>>> dc9b6ef61 (Upgrade to Realm Core v11.7.0)
 
 10.1.0 Release notes (2020-12-8)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Added templates for Expo.
-* A new mode `discardLocal` for client reset is introduce. The old behavior is supported (but deprecated) through the `manual` mode. The new mode will discard any local changes, and a fresh copy of the Realm will be downloaded. An example of the configuration:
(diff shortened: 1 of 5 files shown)


docs/sync.js
```diff
@@ -34,6 +34,15 @@
  * @property {string} version - The version of the app.
  */
 
+/**
+ * This describes the options to configure client reset.
+ * @typedef {Object} Realm.App.Sync~ClientResetConfiguration
+ * @property {string} mode - Either "manual" (deprecated, see also `Realm.App.Sync.initiateClientReset()`) or "discardLocal" (download a fresh copy from the server).
+ * @property {callback(localRealm)|null} [clientResetBefore] - called before sync initiate client reset.
```
takameyer, line 41:
```suggestion
 * @property {callback(localRealm)|null} [clientResetBefore] - called before sync initiates a client reset.
```

src/js_sync.hpp
```diff
@@ -28,8 +28,10 @@
 #include "logger.hpp"
 
 #include "platform.hpp"
+#include "realm/object-store/shared_realm.hpp"
```
takameyer, line 31: Shouldn't this use angle brackets instead of quotes?

src/js_sync.hpp
```diff
@@ -912,6 +985,53 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
...
+            if (!Value::is_undefined(ctx, client_reset_mode_value)) {
+                std::string client_reset_mode = Value::validated_to_string(ctx, client_reset_mode_value, "mode");
+                if (client_reset_mode == client_reset_manual) {
+                    config.sync_config->client_resync_mode = realm::ClientResyncMode::Manual;
+                }
+                else if (client_reset_mode == client_reset_discard_local) {
+                    config.sync_config->client_resync_mode = realm::ClientResyncMode::DiscardLocal;
+                }
+                else {
+                    throw std::invalid_argument("Unknown argument for clientReset.mode");
```
takameyer, line 1011: I think we could print the invalid argument and also suggest the valid options 'manual' and 'discardLocal'

src/js_sync.hpp
```diff
@@ -168,6 +173,68 @@ class SessionClass : public ClassDefinition<T, WeakSession> {
...
+    {
+        HANDLESCOPE(m_ctx)
+
+        typename T::Value arguments[] = {
+            create_object<T, RealmClass<T>>(m_ctx, new SharedRealm(local_realm)),
+            create_object<T, RealmClass<T>>(m_ctx, new SharedRealm(remote_realm)),
+        };
+        Function<T>::callback(m_ctx, m_func, 2, arguments);
+        local_realm->close();
+        remote_realm->close();
```
takameyer, line 200: Once close is called on a sharedRealm, all instances are closed.  I wonder if we need to document that after a reset occurs, the realm needs to be reopened?

takameyer (commented):
Looks good, just a few comments.

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
...
+
+```js
+const config = {
+  // ...
+  sync: {
+    // ...
+    clientReset: {
+      mode: "discardLocal",
+      clientResyncBefore: (localRealm, remoteRealm) => { /* ... */ },
+      clientResyncAfter: (localRealm) => { /* ... */ },
```
ironage, line 15: These signatures are out of date.

src/js_sync.hpp
ironage: I'm not sure about the Realm instance lifecycle in js, but the intention of this feature is that the original Realm instance can continue to be used after the reset, without having to reopen it. Can we avoid closing the instance?

docs/sync.js
```diff
@@ -34,6 +34,15 @@
...
+ * @property {callback(localRealm, remoteRealm)|null} [clientResetAfter] - called after client reset has been executed.
```
ironage, line 42: Up to you, but it might be more clear to use the terms `before` and `after`. Technically, these instances are both pointing to the same underlying Realm, just different versions of it.

tests/js/session-tests.js
```diff
@@ -545,7 +547,93 @@ module.exports = {
...
+          mode: "discardLocal",
+          clientResetBefore: (localRealm) => {
+            beforeCalled = true;
+            TestCase.assertEqual(localRealm.objects("Dog").length, 1, "local");
+          },
+          clientResetAfter: (localRealm, remoteRealm) => {
+            afterCalled = true;
+            TestCase.assertEqual(localRealm.objects("Dog").length, 1, "local");
+            // TestCase.assertEqual(remoteRealm.objects("Dog").length, 1, "remote");
+            // TestCase.assertNotEqual(remoteRealm.path, localRealm.path);
```
ironage, line 618: Since the most recent change, these paths will actually be the same. The `before` state is a an older frozen version of the same Realm and the `after` state is the live most up to date version after the reset has been committed.

docs/sync.js
kneth: Good point

src/js_sync.hpp
kneth: It will follow the code standard -> yes!

tests/js/session-tests.js
kneth: Got it; it doesn't make sense to compare the two paths.

src/js_sync.hpp
kneth: I need to close these two instances, since they are created on a different thread (the sync client's thread). If I don't close them, I will get `realm::IncorrectThreadException: Realm accessed from incorrect thread`. Notice that the test demonstrate that the Realm doesn't need to be reopened.

src/js_sync.hpp
```diff
@@ -915,6 +988,55 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
         }
         config.sync_config->stop_policy = session_stop_policy;
 
+        // Client reset
+        //
+        // i)  manual: the error handler is called with the proper error code and a client reset is initiated
+        // ii) discardLocal: the sync client handles it but notifications are send before and after
```
fronck, line 994:
```suggestion
        // ii) discardLocal: the sync client handles it but notifications are sent before and after
```

tests/js/session-tests.js
```diff
@@ -542,9 +543,10 @@ module.exports = {
         const config = getSyncConfiguration(user, partition);
         config.sync.error = (sender, error) => {
           try {
+            console.log(JSON.stringify(error));
```
fronck, line 546:
```suggestion

```

PR #4084, opened 2021-11-16
Realm Web: Use `globalThis` instead of `window` global

kraenhansen:
## What, How & Why?

This updates Realm Web to adopt best practices in regards to usage of the `window` property, namely use [the `globalThis` instead](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) to avoid runtime errors when the package if the browser bundle get loaded on a runtime not providing the browser (read `window` global) API.

Incidentally this also fix a minor inconvenience when using Realm Web from a Cloudflare Worker.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


takameyer (approved):
Looks like a reasonable change. 👍🏼

PR #4087, opened 2021-11-17
Add core `.clang-format` and format all source

tomduncalf:
## What, How & Why?

Add consistent formatting to our C++ files using the same `.clang-format` as core

Applied to existing files using ```for i in `find src -name "*.?pp" | grep -v catch`;  do echo $i; clang-format --style=file $i -i; done``` from the root directory

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,136 @@
+---
+BasedOnStyle: LLVM
+TabWidth: 4
+UseTab: Never
+---
+Language:  JavaScript
+DisableFormat: true
+---
+Language:        Cpp
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortLambdasOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: Yes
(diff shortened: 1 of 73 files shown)


.clang-format
```diff
@@ -0,0 +1,136 @@
...
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: true
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+  - SECTION
+  - TEST_CASE
+  - DYNAMIC_SECTION
+  - BENCHMARK
```
kraenhansen, line 76: We could probably get rid of these.

.clang-format
```diff
@@ -0,0 +1,136 @@
...
+SpaceBeforeParens: ControlStatementsExceptForEachMacros
+SpaceBeforeRangeBasedForLoopColon: true
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
```
kraenhansen, line 129: This doesn't look right.

tomduncalf: Hm, no it doesn't – I'll ask core about this, as it's copied from there?

.clang-format
```diff
@@ -0,0 +1,136 @@
+---
+BasedOnStyle: LLVM
+TabWidth: 4
```
kraenhansen, line 3: It might make sense to just default to `2` since that's what our JS code does?

.clang-format
```diff
@@ -0,0 +1,136 @@
...
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+StatementMacros:
+  - Q_UNUSED
+  - QT_REQUIRE_VERSION
+TabWidth:        4
+UseTab:          Never
```
kraenhansen, line 134: It might actually make sense to just remove all of this and rely on a preset instead for now.
I know this will take us in a different direction than Realm Core, but again it also seems to me like some of the preferences they've selected seems kind of arbitrary.

```suggestion
```

tomduncalf: Let's ask the team and see – I guess it doesn't matter too much whether we align with core or not, I assumed it would make sense to but perhaps not!

kraenhansen: We agreed to take it all for now. This is something we can always revisit if someone feels strongly either way.

kraenhansen (approved):

kneth (approved):

PR #4088, opened 2021-11-17
Quote all args for integration test run, so that quoted multi-word grep strings are preserved

tomduncalf:
## What, How & Why?

Previously, quoted multi-word arguments were being split up into their words, meaning that running integration tests with a multi-word `--grep` did not produce the expected results.

Before:
```
➜  tests git:(master) npm start -- --grep "fails when passed an array with an object without 'properties'"

> realm-integration-tests@0.1.0 start
> node start.js --watch "--grep" "fails when passed an array with an object without 'properties'"
...
[mocha] > ts-mocha --project ./tsconfig.json --require src/utils/inject-node-globals.ts --watch-extensions ts,js src/index.ts "--watch" "--grep" "fails" "when" "passed" "an" "array" "with" "an" "object" "without" "properties"
```

After:
```
➜  tests git:(td/preserve-quoted-args-for-integration-tests) npm start -- --grep "fails when passed an array with an object without 'properties'"

> realm-integration-tests@0.1.0 start
> node start.js --watch "--grep" "fails when passed an array with an object without 'properties'"
...
[mocha] > ts-mocha --project ./tsconfig.json --require src/utils/inject-node-globals.ts --watch-extensions ts,js src/index.ts "--watch" "--grep" "fails when passed an array with an object without 'properties'"
```

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/start.js
+++ b/integration-tests/tests/start.js
@@ -25,9 +25,12 @@ const concurrently = require("concurrently");
 
 const extraArgs = process.argv.slice(2);
 
-concurrently([{ command: `npm:mocha -- ${extraArgs.join(" ")}` }, { command: "npm:app-importer" }], {
-  killOthers: ["failure", "success"],
-}).catch((tasks) => {
+concurrently(
+  [{ command: `npm:mocha -- ${extraArgs.map((arg) => `"${arg}"`).join(" ")}` }, { command: "npm:app-importer" }],
+  {
+    killOthers: ["failure", "success"],
+  },
+).catch((tasks) => {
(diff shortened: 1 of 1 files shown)


kraenhansen (approved):
This is better, but still a bit annoying that the concurrently API doesn't take an array of arguments instead.

tomduncalf:
Yeah... I can't see a better way to do it, any ideas welcome!

kraenhansen:
Yeah - I think any improvements over this would require a PR to `concurrently`.

PR #4090, opened 2021-11-19
Run prettier on type definitions

tomduncalf:
## What, How & Why?

I noticed these did not match our prettier formatting. Unfortunately the diff looks terrifying due to the way GH presents it!

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


tomduncalf:
We are not using prettier 🙃  and eslint makes a mess of this file

PR #4096, opened 2021-11-22
Add linting check for C++ code

tomduncalf:
## What, How & Why?

This adds checking C++ code against the `clang-format` rules to our GHA lint CI

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/pr-linting.yml
+++ b/.github/workflows/pr-linting.yml
@@ -23,3 +23,5 @@ jobs:
         run: npm run lerna:bootstrap
       - name: Run linting of subpackages
         run: npm run lerna:lint -- --no-bail --concurrency 1 --stream --no-prefix
+      - name: Run linting of C++ code
+        run: for i in `find src -name "*.?pp" | grep -v catch`; do clang-format --dry-run --Werror $i; done
--- a/.github/workflows/pr-linting.yml
+++ b/.github/workflows/pr-linting.yml
@@ -24,4 +24,4 @@ jobs:
       - name: Run linting of subpackages
         run: npm run lerna:lint -- --no-bail --concurrency 1 --stream --no-prefix
       - name: Run linting of C++ code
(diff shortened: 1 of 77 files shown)


takameyer (approved):
Looks good to me

.github/workflows/pr-linting.yml
```diff
@@ -23,3 +23,5 @@ jobs:
         run: npm run lerna:bootstrap
       - name: Run linting of subpackages
         run: npm run lerna:lint -- --no-bail --concurrency 1 --stream --no-prefix
+      - name: Run linting of C++ code
+        run: for i in `find src -name "*.?pp" | grep -v catch`; do clang-format --dry-run --Werror $i; done
```
kraenhansen, line 27: I was wondering if we wanted to use the clang-format NPM package and add a script to the package running this command?

Also, isn't there a more compact way of running this on a directory of files?

kraenhansen: I guess using xargs will eliminate the for loop https://stackoverflow.com/a/36046965
Guess that's a matter of taste 🙂

tomduncalf: Ah yes I forgot about the clang-format package – I'll take a look and try that. I have no strong feelings about how we run it, I tend to use for loops because I always forget the xargs syntax haha, but it is maybe a bit neater so I'll change

kraenhansen: Probably a matter of taste ... I tend to forget the for-loop bash syntax 🙈

tomduncalf: As usual with shell, both are quite awkward syntaxes 😁

tomduncalf: Update to use the npm package which I agree is cleaner

kraenhansen (approved):

PR #4097, opened 2021-11-22
Immutable Realm

takameyer:
## What, How & Why?
Live objects usefulness in Realm-JS has outlived its usefulness.  This PR will attempt to resolve this by having an 'immutable' mode which ensures that data returned from Realm does not change unexpectedly.

This closes # 2578

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


takameyer:
Went a different direction with this.

PR #4107, opened 2021-11-25
Remove typescript dependency from child packages and upgrade root version to latest

tomduncalf:
## What, How & Why?

We currently have mismatched Typescript versions across our packages, which can result in incorrect Intellisense in VS Code. This PR removes the explicit TS dependency from all child packages (except the standalone ts-example one) in favour of inheriting the parent version.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/contrib/building.md
+++ b/contrib/building.md
@@ -13,7 +13,9 @@
             * [ccache](#ccache)
    * [Cloning the repository](#cloning-the-repository)
       * [Cloning the repository on Windows](#cloning-the-repository-on-windows)
-      * [Visual Studio Code setup for C++](#visual-studio-code-setup-for-c)
+      * [Visual Studio Code setup](#visual-studio-code-setup)
+         * [TypeScript](#typescript)
+         * [C++](#c)
    * [Building Realm JS](#building-realm-js-1)
       * [Building for iOS](#building-for-ios)
       * [Building for Android](#building-for-android)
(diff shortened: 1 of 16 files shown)


tomduncalf:
(This did pass CI before I pushed again, so hopefully caught anything which was broken by the upgrade)

kraenhansen (approved):

contrib/building.md
```diff
@@ -35,7 +37,7 @@
       * [Debugging failing Github Actions CI tests](#debugging-failing-github-actions-ci-tests)
    * [Testing against real apps](#testing-against-real-apps)
 
-<!-- Added by: tom.duncalf, at: Tue 23 Nov 2021 10:03:05 GMT -->
+<!-- Added by: tom.duncalf, at: Thu 25 Nov 2021 12:07:16 GMT -->
```
takameyer, line 40: Is this comment block necessary?  We have git blame for this 😆

takameyer (approved):
Excellent work here

contrib/building.md
tomduncalf: It's added by the `gh-md-toc` tool... but we can suppress it actually, I can do a separate PR for that updating the command too

tomduncalf: In fact I had to do a rebase anyway so I added it to this PR :) https://github.com/realm/realm-js/pull/4107/commits/2380a7fdb4685d6d664e2a2d1fe16ca48a414fab

takameyer: Ah, that explains it.  Cool tool

tomduncalf:
TODO unformat changelog 🙄

PR #4113, opened 2021-11-26
Removing the React Example app

kraenhansen:
## What, How & Why?

This removes the outdated React Example app, since:
- It uses an outdated version of React Native (v0.59.9), which makes it fail builds on the `hermes` branch.
- It shows usage of a component (our `ListView`) that we’re not actively maintaining.
- It should probably be rewritten to use hooks via the `@realm.io/react` package anyway.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Add `.clang-format` file for formatting C++ code, formatted all existing code and added a lint check for C++.
 * Updated Docker URL to new canonical URL of `ghcr.io`
 * Excluding the `react-native/android/build` directory from the NPM package.
+* Removed the `examples/ReactExample` app.
```
kneth, line 20:
```suggestion
* Removed the `examples/ReactExample` app. See [FindOurDevices](https://github.com/realm/FindOurDevices) for a modern example app.
```

kneth (approved):

PR #4120, opened 2021-11-30
Hermes: Fixing a couple of breaking tests

kraenhansen:
## What, How & Why?

This fixes a couple of breaking tests.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/src/js_object_accessor.hpp
+++ b/src/js_object_accessor.hpp
@@ -554,23 +554,21 @@ struct Unbox<JSEngine, Obj> {
 
         auto current_realm = native_accessor->m_realm;
         auto js_object = Value::validated_to_object(native_accessor->m_ctx, value);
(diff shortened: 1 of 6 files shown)


src/js_object_accessor.hpp
```diff
@@ -554,23 +554,21 @@ struct Unbox<JSEngine, Obj> {
 
         auto current_realm = native_accessor->m_realm;
         auto js_object = Value::validated_to_object(native_accessor->m_ctx, value);
-        auto realm_object = get_internal<JSEngine, RealmObjectClass<JSEngine>>(native_accessor->m_ctx, js_object);
```
kraenhansen, line 557: The primary purpose of this change is to move the call to `get_internal` into the block where we actually know we have an object pointing to a `RealmObjectClass` instance.

fronck (approved):
LGTM :+1:

PR #4148, opened 2021-12-14
Pass arguments correctly to callResetPasswordFunction

### LINKED ISSUE #4143: Cannot run reset password function using Node SDK

All the time

### Description

I use NodeJS Realm SDK.

I expect my password reset function to be called when I do the following:
```
await app.emailPasswordAuth.callResetPasswordFunction('user@example.com', 'myNewPassword')
```
or at least when I use new API as this is described [here](https://github.com/realm/realm-js/issues/3943):
```
await app.emailPasswordAuth.callResetPasswordFunction({ email: 'user@example.com', password: 'myNewPassword' })
```
But both approaches result in the same error.
 

### Stacktrace & log output

```shell
Error: Error: resetDetails must be of type 'object', got (user@example.com)
at /myProject/node_modules/realm/lib/email-password-auth-methods.js:76:35
at /myProject/node_modules/realm/lib/utils.js:52:7
at new Promise (<anonymous>)
	at promisify (/myProject/node_modules/realm/lib/utils.js:51:12)
	at EmailPasswordAuth.callResetPasswordFunction (/myProject/node_modules/realm/lib/email-password-auth-methods.js:76:12)
	at _callee$ (/myProject/utils/resetPassword.js:11:31)
	at tryCatch (/myProject/node_modules/regenerator-runtime/runtime.js:63:40)
	at Generator.invoke [as _invoke] (/myProject/node_modules/regenerator-runtime/runtime.js:294:22)
	at Generator.next (/myProject/node_modules/regenerator-runtime/runtime.js:119:21)
	at asyncGeneratorStep (/myProject/utils/resetPassword.js:11:103)
```

Yes, always

### Reproduction Steps

1. Enable email/password auth for Realm Application
2. Configure it to run custom password reset function
3. Register a user
4. Try to run password reset function for registered user:
```
const app = new Realm.App({id: id})
await app.emailPasswordAuth.callResetPasswordFunction({ email: 'user@example.com', password: 'newUserPassword' })
```

### Version

10.10.1

MongoDB Realm (i.e. Sync, auth, functions)

### Are you using encryption?

No, not using encryption

Debian GNU/Linux 9.13 (stretch)

### Build environment

_No response_

kneth:
## What, How & Why?

This closes #4143 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests
* [ ] Add `resetFunction` to the test set up (see https://docs.mongodb.com/realm/authentication/email-password/#run-a-password-reset-function)

* ~[ ] jsdoc files updated~

### DIFF
--- a/tests/js/user-tests.js
+++ b/tests/js/user-tests.js
@@ -272,6 +272,24 @@ module.exports = {
     TestCase.assertEqual(err.message, "already confirmed", "User should already be confirmed.");
   },
 
+  async testResetPassword() {
+    let app = new Realm.App(appConfig);
+
+    const validEmail = randomVerifiableEmail();
+    const validPassword = "password123456";
+    await app.emailPasswordAuth.registerUser({ email: validEmail, password: validPassword });
+
+    const newPassword = "password654321";
+    await app.emailPasswordAuth.callResetPasswordFunction({ email: validEmail, password: newPassword });
+
+    // see if we can log in
+    let creds = Realm.Credentials.emailPassword(validEmail, newPassword);
+    let user = await app.logIn(creds);
+    TestCase.assertTrue(user instanceof Realm.User);
+
+    await user.logOut();
+  },
+
   async testFunctions() {
     let app = new Realm.App(appConfig);
     let credentials = Realm.Credentials.anonymous();
--- a/tests/js/user-tests.js
+++ b/tests/js/user-tests.js
@@ -279,7 +279,7 @@ module.exports = {
     const validPassword = "password123456";
     await app.emailPasswordAuth.registerUser({ email: validEmail, password: validPassword });
 
-    const newPassword = "password654321";
+    const newPassword = "realm_tests_do_reset654321";
     await app.emailPasswordAuth.callResetPasswordFunction({ email: validEmail, password: newPassword });
 
     // see if we can log in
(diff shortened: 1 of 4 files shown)


kraenhansen (approved):
If it passes, this looks reasonable :)

tests/js/user-tests.js
```diff
@@ -272,6 +272,24 @@ module.exports = {
...
 
+  async testResetPassword() {
+    let app = new Realm.App(appConfig);
+
+    const validEmail = randomVerifiableEmail();
+    const validPassword = "password123456";
+    await app.emailPasswordAuth.registerUser({ email: validEmail, password: validPassword });
+
+    const newPassword = "realm_tests_do_reset654321";
+    await app.emailPasswordAuth.callResetPasswordFunction({ email: validEmail, password: newPassword });
```
tomduncalf, line 283: Would it be worthing adding another test which calls this with the optional extra `...args`? e.g. 

```
await app.emailPasswordAuth.callResetPasswordFunction({ email: validEmail, password: newPassword }, "test", 123);
```

I'm not sure if/how we could verify that the args are actually being passed to the function, but it would check that we don't get an error calling it at least

tomduncalf (approved):
Thanks for fixing this and adding a test!

tests/js/user-tests.js
takameyer: I guess the function could send the args back in the resolved promise?

tomduncalf: Yeah that's a good idea – I actually have no idea how the Realm app for these tests comes into being (e.g. where the template which contains the password reset function is)!

fronck (approved):
LGTM is all is green 👍

PR #4155, opened 2021-12-16
Fixed `MongoDBCollection#watch` on React Native

### LINKED ISSUE #3494: collection.watch() fails on React Native SDK
## Goals
I'm trying to monitor collection changes on a remote MongoDB Atlas database.

## Expected Results
I followed the example on the [official](https://docs.mongodb.com/realm/react-native/mongodb#std-label-react-native-mongodb-watch) Realm docs to use the collection.watch() functionality for the RN SDK. The example works with Node, and it returns any newly inserted objects in the collection, which is what I'm going for. The exact same code, however, does not work in my RN project.

## Actual Results
I get the following error message

```
EROR [TypeError: Invalid attempt to iterate non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.]
```

## Steps to Reproduce
The error can be reproduced running the following code with the specified SDK and RN versions at the bottom. I call this function inside the `componentDidMount()` function, which looks like this:

 ```js
 componentDidMount() {
    this.watchMessageCollection().catch((err) => console.log('ERROR', err));
  }
 ```
And watchMessageCollection():

```js
  async watchMessageCollection() {
    const {user} = this.props.session;
    const db = user.mongoClient('mongodb-atlas').db('main');
    const messages_collection = db.collection('chat_messages');
    const generator = messages_collection.watch();

    for await (const change of generator) {
      const {documentKey, fullDocument} = change;
      console.log(`new document: ${documentKey}`, fullDocument);
    }
  }

```
I tried running this function on different project files with the same result, so there's something about the generator itself that's off. The user auth works on my RN project, as do the database and collection connections.

- Realm JS SDK Version: 10.1.2
- Node: v15.1.0
- React Native: 0.63.4
- OS: Ubuntu 20.04.1 LTS

kraenhansen:
## What, How & Why?

This closes #3494 by adding documentation on installing polyfills and a babel plugin, plus calling the React Native fetch polyfill with the right option.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually locally)

* [x] jsdoc files updated

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,7 @@
     "./header.eslintrc"
   ],
   "env": {
-    "es2017": true
+    "es2020": true
   },
   "parserOptions": {
     "ecmaVersion": 2018
--- a/packages/realm-network-transport/rollup.config.js
+++ b/packages/realm-network-transport/rollup.config.js
@@ -64,6 +64,21 @@ export default [
     ],
   },
   {
+    input: "src/react-native/index.ts",
+    output: [
+      {
+        file: pkg["react-native"],
+        format: "es",
+      },
+    ],
+    plugins: [
+      typescript({
+        tsconfig: "src/react-native/tsconfig.json",
+      }),
+      nodeResolve(),
+    ],
+  },
+  {
     input: "types/generated/index.d.ts",
     output: {
       file: pkg.types,
--- a/packages/realm-network-transport/src/DefaultNetworkTransport.ts
+++ b/packages/realm-network-transport/src/DefaultNetworkTransport.ts
@@ -29,6 +29,7 @@ import type {
 export class DefaultNetworkTransport implements NetworkTransport {
   public static fetch: Fetch;
   public static AbortController: AbortController;
+  public static extraFetchOptions: Record<string, unknown> | undefined;
 
   public static DEFAULT_HEADERS = {
     "Content-Type": "application/json",
@@ -69,6 +70,7 @@ export class DefaultNetworkTransport implements NetworkTransport {
     try {
       // We'll await the response to catch throw our own error
       return await DefaultNetworkTransport.fetch(url, {
+        ...DefaultNetworkTransport.extraFetchOptions,
         signal, // Used to signal timeouts
         ...rest,
       });
--- a/packages/realm-network-transport/src/react-native/index.ts
+++ b/packages/realm-network-transport/src/react-native/index.ts
@@ -1,0 +1,25 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2021 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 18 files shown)


.eslintrc.json
```diff
@@ -6,7 +6,7 @@
     "./header.eslintrc"
   ],
   "env": {
-    "es2017": true
+    "es2020": true
```
kraenhansen, line 9: This is needed to be able to use the `globalThis` keyword.

packages/realm-network-transport/rollup.config.js
```diff
@@ -63,6 +63,21 @@ export default [
       nodeResolve(),
     ],
   },
+  {
+    input: "src/react-native/index.ts",
```
kraenhansen, line 67: We're adding a new "react-native" bundle to separate the code that runs on "react-native" from that which run on web.

packages/realm-network-transport/src/react-native/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { DefaultNetworkTransport } from "../DefaultNetworkTransport";
+import { AbortController, Fetch } from "../types";
+
+DefaultNetworkTransport.fetch = globalThis.fetch.bind(globalThis) as Fetch;
+DefaultNetworkTransport.AbortController = globalThis.AbortController as AbortController;
+// Setting this non-standard option to enable text streaming
+// See https://github.com/react-native-community/fetch#enable-text-streaming
+DefaultNetworkTransport.extraFetchOptions = {
+  reactNative: { textStreaming: true },
```
kraenhansen, line 29: See https://github.com/react-native-community/fetch#enable-text-streaming for details on why we're adding this.

packages/realm-network-transport/src/DefaultNetworkTransport.ts
```diff
@@ -69,6 +70,7 @@ export class DefaultNetworkTransport implements NetworkTransport {
     try {
       // We'll await the response to catch throw our own error
       return await DefaultNetworkTransport.fetch(url, {
+        ...DefaultNetworkTransport.extraFetchOptions,
```
kraenhansen, line 73: I tried setting this only on the streaming requests, but that yield the following in my test app:

```
 LOG  Running "ReactNativeTestApp" with {"rootTag":61,"initialProps":{}}
 WARN  Invalid responseType: blob
 ERROR  {"code": -1, "message": "undefined is not an object (evaluating 'options.blobId')"}
 WARN  Invalid responseType: blob
 LOG  Unable to symbolicate stack trace: undefined is not an object (evaluating 'options.blobId')
 WARN  Invalid responseType: blob
 WARN  Invalid responseType: blob
 WARN  Invalid responseType: blob
```

packages/realm-network-transport/src/react-native/index.ts
kneth: Please add it as a comment

kneth (approved):

takameyer (approved):
Awesome.  Do we need to prominently document this somewhere, besides in the comments?

packages/realm-network-transport/src/react-native/index.ts
kraenhansen: It's already there - two lines above the line I commented on here :)

kraenhansen:
> Awesome. Do we need to prominently document this somewhere, besides in the comments?

Besides the changes I'm proposing to the TypeScript docs and JSDoc, I would say the instructions on adding the babel plugin and globals polyfill needs to be inserted on https://docs.mongodb.com/realm/sdk/react-native/examples/query-mongodb/#real-time-change-notifications as well.

PR #4162, opened 2021-12-20
Quick and dirty

kneth:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -77,6 +77,28 @@ module.exports = function (realmConstructor) {
     return `${name}#${id}`;
   };
 
+  // Overriding Object.is()
+  Object.is = function (x, y) {
+    if (x instanceof realmConstructor.Object && y instanceof realmConstructor.Object) {
+      return x._objectId() === y._objectId();
+    }
+
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is
+    // SameValue algorithm
+    if (x === y) {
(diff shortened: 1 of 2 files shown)


lib/extensions.js
```diff
@@ -77,6 +77,28 @@ module.exports = function (realmConstructor) {
...
+      // if x and y are both 0 of the same sign.
+      // This checks for cases 1 and 2 above.
+      return x !== 0 || 1 / x === 1 / y;
+    } else {
+      // return true if both x AND y evaluate to NaN.
+      // The only possibility for a variable to not be strictly equal to itself
+      // is when that variable evaluates to NaN (example: Number.NaN, 0/0, NaN).
+      // This checks for case 3.
+      return x !== x && y !== y;
+    }
```
kraenhansen, line 99: I would definitely just store the `const originalObjectIs = Object.is.bind(Object);` and defer to that instead of writing our own fallback.

lib/extensions.js
```diff
@@ -77,6 +77,28 @@ module.exports = function (realmConstructor) {
     return `${name}#${id}`;
   };
 
+  // Overriding Object.is()
+  Object.is = function (x, y) {
```
kraenhansen, line 81: This should be moved to a separate file that is manually required in by the user, to ensure it's an opt-in feature.

PR #4168, opened 2021-12-21
Rename @realm.io/react to @realm/react and make Realm importable

takameyer:
This commit renames the realm-react module and makes it possible to
import Realm from `@realm/react`.

## ☑️ ToDos
* [x] 📝 Changelog entry


packages/realm-react/package.json
```diff
@@ -12,7 +12,8 @@
     "lint": "eslint --ext .tsx ."
   },
   "dependencies": {
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "realm": "*"
```
kraenhansen, line 16: Shouldn't this be set as a peer dependency now?
To signal that users need to install it alongside.
I think it would be okay if this was more specific on its version range.

kraenhansen: Do you need an update to the README as well?

takameyer: It is a peer dependency and I have added installing `realm` to the readme

kraenhansen: Oh. I'm sorry I didn't expand to see that.

kraenhansen (approved):

kraenhansen:
Looks like the GitHub workflows also need the package name updated.

kneth (approved):

PR #4180, opened 2021-12-29
 Fix realm-web integration tests

takameyer:
The app-importer was updated, but the tests for realm-web were not run.  Some updates to the harness were necessary.


kraenhansen (commented):
Looks reasonable! Thanks for getting this fixed.

PR #4193, opened 2022-01-06
Enable strictNullChecks for integration tests

tomduncalf:
## What, How & Why?

This change is required for flexible sync. Ideally we'd enable `strict: true` but this throws a lot more errors so just doing `strictNullChecks` for now.

I've tried to do runtime assertions where I thought it might make sense (e.g. there's a chance the thing might not exist, if something is badly broken, so throw) and otherwise use non-null assertions, but happy to change if people think we should take a different approach.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -54,19 +54,16 @@ export function openRealmHook(config: LocalConfiguration | SyncedConfiguration =
   };
 }
 
-export function closeRealm(this: RealmContext): void {
+export function closeRealm(this: Partial<RealmContext> & Mocha.Context): void {
   if (this.realm) {
     this.realm.close();
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore Ignore error as we ensure the realm is reopened
     delete this.realm;
   } else {
     throw new Error("Expected a 'realm' in the context");
   }
+
   if (this.config) {
     Realm.deleteFile(this.config);
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore Ignore error as we ensure the config is recreated
     delete this.config;
   } else {
     throw new Error("Expected a 'config' in the context");
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -46,6 +46,9 @@ export function openRealmHook(config: LocalConfiguration | SyncedConfiguration =
       } as Realm.Configuration;
       this.realm = new Realm(this.config);
       // Upload the schema, ensuring a valid connection
+      if (!this.realm.syncSession) {
+        throw new Error("No syncSession found on realm");
+      }
       await this.realm.syncSession.uploadAllLocalChanges();
     }
   };
@@ -54,12 +57,16 @@ export function openRealmHook(config: LocalConfiguration | SyncedConfiguration =
 export function closeRealm(this: RealmContext): void {
   if (this.realm) {
     this.realm.close();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Ignore error as we ensure the realm is reopened
     delete this.realm;
   } else {
     throw new Error("Expected a 'realm' in the context");
   }
   if (this.config) {
     Realm.deleteFile(this.config);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Ignore error as we ensure the config is recreated
     delete this.config;
   } else {
     throw new Error("Expected a 'config' in the context");
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -33,7 +33,7 @@ describe("Realm objects", () => {
   describe("Interface & object literal", () => {
     it("can be created", () => {
       const realm = new Realm({ schema: [PersonSchema] });
-      let john: IPerson;
+      let john!: IPerson;
 
       realm.write(() => {
         john = realm.create<IPerson>(PersonSchema.name, {
@@ -51,7 +51,7 @@ describe("Realm objects", () => {
 
     it("can have it's properties read", () => {
       const realm = new Realm({ schema: [PersonSchema] });
-      let john: IPerson;
+      let john!: IPerson;
 
       realm.write(() => {
         john = realm.create<IPerson>(PersonSchema.name, {
@@ -77,6 +77,7 @@ describe("Realm objects", () => {
       });
 
       const john = realm.objectForPrimaryKey<IPersonWithId>(PersonSchemaWithId.name, _id);
+      if (!john) throw new Error("Object not found");
 
       expect(john).instanceOf(Realm.Object);
       expect(john._id.equals(_id)).equals(true);
@@ -86,7 +87,7 @@ describe("Realm objects", () => {
 
     it("can be updated", () => {
       const realm = new Realm({ schema: [PersonSchemaWithId] });
-      let john: IPersonWithId;
+      let john!: IPersonWithId;
       const _id = new Realm.BSON.ObjectId();
 
       realm.write(() => {
@@ -160,7 +161,7 @@ describe("Realm objects", () => {
   describe("Class Model", () => {
     it("can be created", () => {
       const realm = new Realm({ schema: [Person] });
-      let john: Person;
+      let john!: Person;
 
       realm.write(() => {
         john = realm.create(Person, {
@@ -202,6 +203,7 @@ describe("Realm objects", () => {
       });
 
       const john = realm.objectForPrimaryKey(PersonWithId, _id);
+      if (!john) throw new Error("Object not found");
 
       expect(john).instanceOf(PersonWithId);
       expect(john._id.equals(_id)).equals(true);
@@ -211,7 +213,7 @@ describe("Realm objects", () => {
 
     it("can be updated", () => {
       const realm = new Realm({ schema: [PersonWithId] });
-      let john: PersonWithId;
+      let john!: PersonWithId;
       const _id = new Realm.BSON.ObjectId();
 
       realm.write(() => {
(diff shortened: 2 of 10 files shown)


integration-tests/tests/src/hooks/open-realm-before.ts
```diff
@@ -46,6 +46,9 @@ export function openRealmHook(config: LocalConfiguration | SyncedConfiguration =
       } as Realm.Configuration;
       this.realm = new Realm(this.config);
       // Upload the schema, ensuring a valid connection
+      if (!this.realm.syncSession) {
+        throw new Error("No syncSession found on realm");
+      }
       await this.realm.syncSession.uploadAllLocalChanges();
```
kraenhansen, line 52:
```suggestion
      if (this.realm.syncSession) {
        await this.realm.syncSession.uploadAllLocalChanges();
      }
```

tomduncalf: My thinking was that if the `syncSession` does not exist here, something has gone badly wrong, as we've created a realm with a sync config, so we should throw

integration-tests/tests/src/hooks/open-realm-before.ts
```diff
@@ -54,12 +57,16 @@ export function openRealmHook(config: LocalConfiguration | SyncedConfiguration =
 export function closeRealm(this: RealmContext): void {
   if (this.realm) {
     this.realm.close();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Ignore error as we ensure the realm is reopened
```
kraenhansen, line 61: Perhaps `this` should be a `Partial<RealmContext>` instead?

integration-tests/tests/src/hooks/open-realm-before.ts
kraenhansen: Ahh sorry, I hadn't expanded the code snippet. I agree, we're in a branch where sync is expected.

integration-tests/tests/src/hooks/open-realm-before.ts
tomduncalf: Yeah that would work!

takameyer: What was the ts error here?

tomduncalf: `The operand of a 'delete' operator must be optional` which is a new one to me!

integration-tests/tests/src/tests/objects.ts
```diff
@@ -33,7 +33,7 @@ describe("Realm objects", () => {
   describe("Interface & object literal", () => {
     it("can be created", () => {
       const realm = new Realm({ schema: [PersonSchema] });
-      let john: IPerson;
+      let john!: IPerson;
 
       realm.write(() => {
         john = realm.create<IPerson>(PersonSchema.name, {
```
kraenhansen, line 39: Actually, with https://github.com/realm/realm-js/pull/3937 merged, we might as well do:

```suggestion
      const john = realm.write(() => {
        return realm.create<IPerson>(PersonSchema.name, {
```

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -484,7 +484,8 @@ describe("JSON serialization", () => {
       afterEach(() => {
         if (realm) {
           realm.write(() => {
-            realm.deleteAll();
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            realm!.deleteAll();
```
kraenhansen, line 488: I'm surprised this is needed ...

integration-tests/tests/src/hooks/open-realm-before.ts
takameyer: Ahhh...ok that makes sense.  Since `this` is typed with non-optionals.  So `Partial<RealmContext>` is appropriate

integration-tests/tests/src/utils/benchmark.ts
```diff
@@ -35,7 +35,8 @@ export function itPerforms(title: string, fn: () => void, options?: Partial<Benc
       maximumFractionDigits: 0,
     });
     const sd = Number(result.sd.toFixed(2));
-    this.test.title += ` (${ops} ops/sec, ±${sd}%)`;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    this.test!.title += ` (${ops} ops/sec, ±${sd}%)`;
```
kraenhansen, line 39: Does this fly? Using the `!` operator is really just hiding a potential issue.

```suggestion
    if (this.test) {
      this.test.title += ` (${ops} ops/sec, ±${sd}%)`;
    }
```

integration-tests/tests/src/tests/objects.ts
takameyer: Why are we not using classes for integration tests?  (Question is out of scope for this PR, just curious)

integration-tests/tests/src/tests/serialization.ts
tomduncalf: Yeah I have been confused by this before. I believe the logic is that TS can't infer how the callback is called, so it could be being called async and `realm` has changed in the interim to being `null`

image

integration-tests/tests/src/utils/benchmark.ts
tomduncalf: I was making the assumption that `test` would always be defined. Though if mocha type defs mark it as optional, I guess there maybe is some edge case! I think it might be appropriate to throw if `test` is undefined as it would mean something weird has happened?

integration-tests/tests/src/tests/serialization.ts
kraenhansen: But `realm` is captured when the lambda is created? It can't change value from that point on 🤔

integration-tests/tests/src/tests/objects.ts
tomduncalf: Oh I had no idea #3937 was a thing, nice!

integration-tests/tests/src/tests/serialization.ts
kraenhansen: It seems like a TSC bug to me.

tomduncalf: Hmm yeah... maybe the TS logic is a bit faulty, I think this is described in https://github.com/microsoft/TypeScript/issues/10982#issuecomment-247993896

kraenhansen: It works as expected here: https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3nKEC2AXPCgK4QTwA+ABQN7wBmSSRARAEayvwC+A3ACgBASwbwAFDnwBKeLQHxFiVJkbN4AXmwhceAHRMkgpcvRZoedsCiaJsjQD45Ck0uTokEEHohIA5pI6+AbM0sZK-AI8AkA

integration-tests/tests/src/tests/objects.ts
```diff
@@ -77,6 +77,7 @@ describe("Realm objects", () => {
       });
 
       const john = realm.objectForPrimaryKey<IPersonWithId>(PersonSchemaWithId.name, _id);
+      if (!john) throw new Error("Object not found");
```
takameyer, line 80: Can we `expect(john).to.not.be(undefined)`, or does that defeat the purpose of this PR?

integration-tests/tests/src/tests/serialization.ts
takameyer: I hope its not

tomduncalf: That's weird...

kraenhansen: Ahh. Perhaps https://github.com/realm/realm-js/pull/4193/files#diff-13ec3f38b0cf2f5f7842047684a331ad82a406f144a3e9ac84cad52718bd824eR491 is the offender :)

tomduncalf: I think it's actually because `realm` is a `let` https://github.com/realm/realm-js/pull/4193/files#diff-13ec3f38b0cf2f5f7842047684a331ad82a406f144a3e9ac84cad52718bd824eR475

See https://www.typescriptlang.org/play?ssl=2&ssc=1&pln=2&pc=4#code/PTAEBMFMGMBsEMBOlTQPYDsDOAXUz5YBbALlAwFdZZQAfAAoG9QAzNNMgIgCMlPQAvgG4AULEh4CxMpWp0mrdl16J+A0AF5QzNh1A8+g0SICWLUAAopRAJTaRoR6ky5FaTfkiEiAOl2inZ2w8BCJucHgPCzsNAD57QMD0bDRxH1g0AHMrL2I-dhsAp2ERAREgA

takameyer (commented):
I think the test should make assertions instead of throwing exceptions.  Maybe this warrants a  quick discussion.

integration-tests/tests/src/tests/serialization.ts
kraenhansen: Oh my ... I learned something today.

integration-tests/tests/src/tests/objects.ts
tomduncalf: We could, but this wouldn't satisfy Typescript so we'd have to do non-null assertions or whatever. I thought that as its an edge case (like should never happen), throwing is the least lines of code required to both fail the test and keep TS happy, but the test error may be a little less clear if you did hit the edge case

takameyer (approved):

integration-tests/tests/src/tests/objects.ts
takameyer: Ok...that makes sense.  We need to keep TS happy 🍼

kraenhansen (approved):
I would love for me or you (probably me) to do a follow-up PR making this `strict` instead of simply `strictNullChecks`.

tomduncalf:
If you have the time, that would be awesome, but if not I am happy to look at it once my current bit of work is out of the way (probably end of the week). I'll create an issue to track

tomduncalf:
https://github.com/realm/realm-js/issues/4200

PR #4203, opened 2022-01-10
Clean up JavaScriptCore & legacy Chrome debugger RPC layer

kraenhansen:
## What, How & Why?

Now that our support for JavaScriptCore on the `hermes` branch is entirely through the JSI (JavaScript Interface), we can remove our direct implementation to JavaScriptCore and the legacy Chrome Debugger implementation (which was implemented tightly coupled with the JSC API).

## The experience of running in legacy Chrome debugging mode

Users starting an app that requires Realm from Chrome debugging mode will now experience the following:

> This version of Realm JS doesn't support the legacy Chrome Debugger

error on iOSerror on Android

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually)

* [ ] jsdoc files updated
* [x] Chrome debug API is updated if API is available on React Native (for the last time)


lib/react-native/index.js
```diff
@@ -25,25 +25,35 @@ if (typeof Reflect !== "undefined" && Reflect.construct) {
...
+// @see https://github.com/realm/realm-js/issues/2126
+const Realm = globalThis.Realm;
+
 // Otherwise, we must be in a "normal" react native situation.
 // In that case, the Realm type should have been injected by the native code.
 // If it hasn't, the user likely forgot to install the RealmJS CocoaPod
-if (typeof global.Realm === "undefined") {
+if (typeof Realm === "undefined") {
   throw new Error(
     'Missing Realm constructor. Did you run "pod install"? Please see https://realm.io/docs/react-native/latest/#missing-realm-constructor for troubleshooting',
```
fronck, line 49: This URL ends up in a 403.
This error message isn't mentioned in our current documentation, but can be found in the [legacy documentation](https://docs.mongodb.com/realm-legacy/docs/javascript/3.6.0/index.html) and in [this blog post](https://www.mongodb.com/developer/how-to/build-offline-first-react-native-mobile-app-with-expo-and-realm/).

kraenhansen: I agree, it does make sense to update the URL in this PR too.

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+### Breaking change
+* Removed all code related to the legacy Chrome Debugger.
```
```suggestion
* Removed all code related to the legacy Chrome Debugger. Please use [Flipper](https://fbflipper.com/) as debugger.
```

lib/react-native/index.js
```diff
@@ -25,25 +25,35 @@ if (typeof Reflect !== "undefined" && Reflect.construct) {
   Reflect.construct.sham = 1;
 }
 
+const usingLegacyChromeDebugger = typeof DedicatedWorkerGlobalScope !== "undefined";
+
+if (usingLegacyChromeDebugger) {
+  throw new Error("This version of Realm JS doesn't support the legacy Chrome Debugger");
```
kneth, line 31:
```suggestion
  throw new Error("This version of Realm JS doesn't support the legacy Chrome Debugger. Please use Flipper instead.");
```

kneth (approved):
Feels so good!

lib/index.js
```diff
@@ -17,46 +17,19 @@
...
-      if (global.__DEV__) {
-        return require("./browser").default; // (exported as ES6 module)
-      } else {
-        throw new Error("Can´t use debugger if __DEV__ isn´t true.");
-      }
-    default:
-      throw new Error("Unexpected execution environment (" + environment + ")");
-  }
+const environment = utils.getEnvironment();
+if (environment !== "node.js" || environment !== "electron") {
```
fronck, line 23: Can "reactnative" or "jscore" no longer be our environment here?

fronck (approved):
Nice work getting rid of this old cruft :+1:

lib/index.js
kraenhansen: Right. With [the `react-native` main field in the `package.json`](https://github.com/realm/realm-js/blob/hermes/package.json#L34), this code will no longer be hit on React Native:

lib/react-native/index.js
fronck: I think we should remove the link altogether.  The docs site is no longer something we control, so the URL may change at any time..

kraenhansen:
Merging with these failing tests:
- Integration Tests via Node.js v14 on Window (debug & release build): The `npm ci` isn't happening with NPM v7 fails the build. It does pass on v16, so I think it's okay to disregard this for this PR and fix it later.
- Test runners on Linux failed, but it already passed on a previous commit and it succeeded on MacOS with this commit, I don't see a need to rerun the build to verify.

crabby786:
I am switching to sqlite , just becouse realm does not support chrome debugger (but flipper, dont know why)

PR #4215, opened 2022-01-13
Update Jenkinsfile to use NPM CLI v8 on Windows

### LINKED ISSUE #4202: Upgrade nodist, enabling switching to NPM v7 on CI
We're unable to `npm ci` on Windows since we've switched to a new version of the `package-lock.json`. See https://github.com/nullivex/nodist/issues/244 as the original issue blocking this upgrade.

kraenhansen:
## What, How & Why?

This fixes #4202

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,8 @@ import groovy.json.JsonOutput
 repoName = 'realm-js'
 
 platforms = ['win32-ia32', 'win32-x64', 'darwin-x64', 'darwin-arm64', 'linux-x64', 'linux-arm']
-nodeTestVersion = '12.22.5'
+nodeTestVersion = '16.13.1'
+npmVersion = '8.1.2'
 
 //Changing electron versions for testing requires upgrading the spectron dependency in tests/electron/package.json to a specific version.
 //For more see https://www.npmjs.com/package/spectron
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -236,8 +237,11 @@ def buildWindows(nodeVersion, arch) {
   return {
     myNode('windows && nodejs') {
       unstash 'source'
-
+      bat "nodist add ${nodeVersion}"
+      bat "nodist npm add ${npmVersion}"
```
fealebenpae, line 241:
```suggestion
      bat """
        nodist add ${nodeVersion}
        nodist npm add ${npmVersion}
      """
```
You can use a single shell block to run both commands.

Jenkinsfile
```diff
@@ -505,9 +509,11 @@ def testWindows(nodeVersion) {
     node('windows && nodist') {
       unstash 'source'
       bat "nodist add ${nodeVersion}"
+      bat "nodist npm add ${npmVersion}"
```
fealebenpae, line 512: You can run both commands in a single block, see previous suggestion.

fealebenpae (approved):

Jenkinsfile
kraenhansen: Thanks for the suggestion. I like the simplicity of separating these. In my opinion it makes it simpler to determine what command failed.

PR #4219, opened 2022-01-13
Remove outdated release instructions

tomduncalf:
## What, How & Why?

These instructions seem to be outdated/incomplete so should point to the wiki

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kneth (commented):
I think the markdown file is more accurate. We have almost abandoned the old wiki.

tomduncalf:
In which case I will update the Markdown and remove the wiki page :)

PR #4228, opened 2022-01-17
Prevent dangling objects on JSCRuntime destruction

### LINKED ISSUE #4115: Fix "JSCRuntime destroyed with a dangling API object"
When running with hermes disabled, our "old" test app crash at this assertion:

[JSCRuntime destroyed with a dangling API object](https://github.com/facebook/react-native/blob/v0.66.3/ReactCommon/jsi/JSCRuntime.cpp#L393-L394)

This hints that we're leaking JavaScript objects. Which would also be valuable to get resolved when running with other engines.

kraenhansen:
## What, How & Why?

This closes #4115 by introducing a separate static data structure (the `NotificationBucket` class) that owns `NotificationToken`s returned by object store APIs. This data structure can be cleaned as the JS runtime gets torn down to ensure no objects are left protected as the JS runtime gets destructed.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually verified the assertion in JSCRuntime.cpp no longer happens)

### DIFF
--- a/src/js_notifications.hpp
+++ b/src/js_notifications.hpp
@@ -45,13 +45,13 @@ class NotificationHandle;
 template <typename T>
 class NotificationBucket {
     using ProtectedFunction = Protected<typename T::Function>;
-
-    static inline std::map<IdType, std::vector<std::pair<ProtectedFunction, NotificationToken>>> s_tokens;
+    using TokensMap = std::map<IdType, std::vector<std::pair<ProtectedFunction, NotificationToken>>>;
 
 public:
     static void emplace(NotificationHandle<T>& handle, ProtectedFunction&& callback, NotificationToken&& token)
     {
         if (handle) {
+            auto& s_tokens = get_tokens();
             s_tokens[handle].emplace_back(std::move(callback), std::move(token));
         }
         else {
@@ -61,12 +61,14 @@ public:
 
     static void clear()
     {
+        auto& s_tokens = get_tokens();
         s_tokens.clear();
     }
 
     static void erase(NotificationHandle<T>& handle)
     {
         if (handle) {
+            auto& s_tokens = get_tokens();
             s_tokens.erase(handle);
         }
     }
@@ -74,6 +76,7 @@ public:
     static void erase(NotificationHandle<T>& handle, ProtectedFunction&& callback)
     {
         if (handle) {
+            auto& s_tokens = get_tokens();
             auto& tokens = s_tokens[handle];
             auto compare = [&callback](auto&& token) {
                 return typename ProtectedFunction::Comparator()(token.first, callback);
@@ -83,6 +86,20 @@ public:
         else {
             throw std::runtime_error("Cannot erase notifications using an unset handle");
         }
+    }
+
+    /**
+     * @brief Get the static tokens object.
+     *
+     * @note We cannot use a simple static object on the namespace, because we are referencing this across multiple
+     * translation units and this could lead to code using this before it's initialized.
+     *
+     * @return auto& A reference to the static map of tokens.
+     */
+    inline static TokensMap& get_tokens()
+    {
+        static TokensMap s_tokens;
+        return s_tokens;
     }
 };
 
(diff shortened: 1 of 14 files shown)


tests/ReactTestApp/ios/ReactTestApp.xcodeproj/project.pbxproj
```diff
@@ -315,40 +313,6 @@
...
-			);
-			name = "[CP] Embed Pods Frameworks";
-			outputFileListPaths = (
-				"${PODS_ROOT}/Target Support Files/Pods-ReactTestApp-ReactTestAppTests/Pods-ReactTestApp-ReactTestAppTests-frameworks-${CONFIGURATION}-output-files.xcfilelist",
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "\"${PODS_ROOT}/Target Support Files/Pods-ReactTestApp-ReactTestAppTests/Pods-ReactTestApp-ReactTestAppTests-frameworks.sh\"\n";
-			showEnvVarsInLog = 0;
-		};
```
kraenhansen, line 351: To be honest, I don't know exactly how this change ended up here, but on the other hand, it does look like a leftover from a previous upgrade.

src/jsi/jsi_init.cpp
```diff
@@ -46,18 +47,11 @@ extern "C" void realm_jsi_invalidate_caches()
...
 }
-extern "C" void realm_hermes_invalidate_caches()
-{
-    // Close all cached Realms
-    realm::_impl::RealmCoordinator::clear_all_caches();
-    // Clear the Object Store App cache, to prevent instances from using a context that was released
-    realm::app::App::clear_cached_apps();
-    // Ensure all registered invalidators get notified that the runtime is going away.
-    realm::js::Context<realmjsi::Types>::invalidate();
-}
```
kraenhansen, line 60: This was left in after a bad merge.

src/js_notifications.hpp
```diff
@@ -0,0 +1,163 @@
...
+#include <map>
+#include <vector>
+
+#include <realm/object-store/collection_notifications.hpp>
+
+#include "js_types.hpp"
+
+namespace realm::js::notifications {
+
+using IdType = unsigned long long;
```
kneth, line 30: Why not `size_t`?

kneth (approved):
It is good to see that we are getting the notification under control.

src/js_notifications.hpp
kraenhansen: @fronck can answer to this: I believe this was to prevent a very small range if we started publishing for more embedded devices.

fronck: `size_t` is the natural pointer size on a system.  Is it not guaranteed to be any larger than 64k.  So, just in the interest of not getting nasty surprises if we were to, for example, get into the IoT space, I wanted to force it to hold more than that.

src/js_notifications.hpp
```diff
@@ -0,0 +1,163 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2021 Realm Inc.
```
fronck, line 3:
```suggestion
// Copyright 2022 Realm Inc.
```

fronck: ;)

tests/ReactTestApp/App.ios.js
```diff
@@ -25,6 +25,9 @@ export default class ReactTests extends React.Component {
   render() {
     return (
       <View style={styles.container}>
+        <Text>
+          Engine: {global.HermesInternal ? 'Hermes' : 'JavaScriptCore'}
```
fronck, line 29: Nice little tweak - I like it :+1:

talhajibrantariq:
Hello, I have realm v 10.24.0 and react native 0.64. 

I'm still getting this error whenever i refresh the application.

Are these changes merged in 10.24? If yes, then is there any workaround?

PR #4277, opened 2022-01-27
Split JSDoc workflow and fix uploading

tomduncalf:
## What, How & Why?

The current workflow uploads the docs to the wrong location for archival, and does not update the copy of the docs at /latest. This change splits the workflow in two - one which builds the JSDocs for every pull request but does not publish them, and one which only runs on vX.Y.Z tag builds, and uploads the docs both to /latest and for archival.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/jsdoc-publish.yml
+++ b/.github/workflows/jsdoc-publish.yml
@@ -1,0 +1,48 @@
+# This workflow builds the JSDocs and publishes them to https://docs.mongodb.com/realm-sdks/js/latest/.
+# It also uploads them to a versioned directory for archival purposes.
+#
+# We only run this workflow for "vX.Y.Z" tags, not "vX.Y.Z-beta", as the docs are pushed to the
+# site as "latest" as part of the build and we don't want to publish beta docs to live
+
+name: JSDoc Publish
+on:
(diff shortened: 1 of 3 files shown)


.github/workflows/jsdoc-publish.yml
```diff
@@ -0,0 +1,48 @@
...
+# It also uploads them to a versioned directory for archival purposes.
+#
+# We only run this workflow for "vX.Y.Z" tags, not "vX.Y.Z-beta", as the docs are pushed to the
+# site as "latest" as part of the build and we don't want to publish beta docs to live
+
+name: JSDoc Publish
+on:
+  push:
+    tags:
+      - v[0-9]+\.[0-9]+\[0-9]+
```
cbush, line 11: Missing `.`?
```suggestion
      - v[0-9]+\.[0-9]+\.[0-9]+
```

tomduncalf: Ooh good spot!

cbush (approved):

PR #4279, opened 2022-01-28
Enable mixed tests for flexible sync

tomduncalf:
## What, How & Why?

This also adds deleting the Realm app in between test runs, as it can potentially cause slow down on a long running server if it has lots of subscriptions (this was added trying to fix a bug which it was not actually the root cause of, but I think it's worth keeping in).

This depends on running the BaaS server with PR REALMC-9471 for now, as that enables import/export of flexible sync apps.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm-app-importer/src/AppImportServer.ts
```diff
@@ -142,4 +144,20 @@ export class AppImportServer {
...
+
+    // Match a URL like /app/my-app-client-app-id
+    const appIdMatches = req.url.match(/^\/app\/([0-9a-z-]+)$/);
+    if (!appIdMatches || !appIdMatches[1]) {
+      res.statusCode = 400;
+      throw new Error("Expected a URL in the format: /app/my-app-client-app-id");
+    }
+
+    return this.importer.deleteApp(appIdMatches[1]);
+  }
```
kraenhansen, line 162: Great addition!

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -31,4 +31,14 @@ export function importAppBefore(
...
+
+  // Delete our app after we have finished, otherwise the server can slow down
+  // (in the case of flexible sync, with lots of apps with subscriptions created)
+  after(async function (this: Partial<AppContext> & Mocha.Context) {
+    if (!this.app) {
+      throw new Error("No app on context when trying to delete app");
+    }
+
+    await deleteApp(this.app.id);
+  });
```
kraenhansen, line 43: I like the change, but I'm also wondering if this could make it harder to inspect the server-side logs when debugging an issue. At some point we might want to have some sort of "manual mode" where the tests will skip importing and deleting an app and instead simply use an app-id provided when starting the test.

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -107,46 +125,48 @@ function describeRoundtrip(typeName: string, value: Value, testValue: ValueTeste
...
 
-  describeRoundtrip("string", "test-string");
+  describeRoundtrip(flexibleSync, "string", "test-string");
 
   // Unsupported:
   // describeSimpleRoundtrip("undefined", undefined);
 
   const buffer = new Uint8Array([4, 8, 12, 16]).buffer;
-  describeRoundtrip("data", buffer, (value: ArrayBuffer) => {
+  describeRoundtrip(flexibleSync, "data", buffer, (value: ArrayBuffer) => {
```
kraenhansen, line 148: I feel this is approaching the limits of positional arguments.

integration-tests/tests/src/tests/sync/upload-delete-download.ts
```diff
@@ -18,7 +18,7 @@
 
 import { closeAndReopenRealm } from "../../utils/close-realm";
 
-export function itUploadsDeletesAndDownloads(): void {
+export function itUploadsDeletesAndDownloads(flexibleSync = false): void {
```
kraenhansen, line 21: I believe it would be more clean if this was read from the `realm.syncSession.config` instead of getting passed as an argument.

packages/realm-app-importer/src/AppImporter.ts
```diff
@@ -136,6 +150,25 @@ export class AppImporter {
     return { appId };
   }
 
+  public async deleteApp(clientAppId: string): Promise<void> {
+    await this.logIn();
```
kraenhansen, line 154: Hmm. I just noticed we're doing this in `importApp` as well. It was written originally as a once pr process thing, but with the `realm-app-importer serve`, we shouldn't log-in again and again. I've [created an issue](https://github.com/realm/realm-js/issues/4280) to tackle this.

kraenhansen (approved):
Great job! I have just minor comments.

integration-tests/tests/src/hooks/import-app-before.ts
tomduncalf: Maybe I can add an environment context option to skip deleting the app for now, in case you want to inspect the state of the app after a test run?

integration-tests/tests/src/tests/sync/mixed.ts
tomduncalf: Aha you might be right, I'll look at changing it

integration-tests/tests/src/tests/sync/upload-delete-download.ts
tomduncalf: Makes sense!

packages/realm-app-importer/src/AppImporter.ts
tomduncalf: 👍

PR #4284, opened 2022-01-31
Fix logo in darkmode

takameyer:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen:
So the logos will be side-by-side on https://www.npmjs.com/package/realm? (which is currently broken in dark-mode btw).

takameyer:
@kraenhansen That's true!  So basically they implemented a non-standard method of doing this that breaks everywhere but here :D

kraenhansen:
There's always the possibility of packing for NPM with an altered README.md, but that's a bit involved for the problem at hand.

kneth (approved):
LGTM

fronck (approved):

nirinchev:
> There's always the possibility of packing for NPM with an altered README.md, but that feels a bit involved for the problem at hand.

As with everything in npm, apparently [there's a package for that](https://www.npmjs.com/package/strip-gh-theme-links) 😄 And it even has a github action you can plug before `npm publish`.

takameyer:
@nirinchev I added a bash script to do the same.  We can use it later when we have automated releases

Issue #4160, opened 2021-12-17
iOS crash when the app is being terminated/swiped out


cristianoccazinsp:
### Description

Regular use. The database is opened only once on app startup and never closed, and there are a few permanently rendered components with listeners to about two different collections (`addListener`).

When you swipe out the app, there will be ocasional crashes not visible to the user, unless the OS reports a "the app has crashed, do you want to report it?" prompt.

### Stacktrace & log output

```shell
CrashReporter Key:  0efe0ef54dbc94f6507345177af9b6ae7583f073
Hardware Model:     iPhone13,3
Process:            zinspector3
Identifier:         com.zinspector.zinspector3
Version:            3.2.11
Role:               Background
OS Version:         iOS 15.1.1
Exception Type:     EXC_BAD_ACCESS 
Exception Subtype:  KERN_INVALID_ADDRESS


EXC_BAD_ACCESS: Attempted to dereference garbage pointer 0x8.

[...]

PR #4298, opened 2022-02-02
Fix instances where SubscriptionSet was still referred to as Subscriptions

tomduncalf:
## What, How & Why?

This fixes https://github.com/realm/realm-js/issues/4160

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


types/index.d.ts
```diff
@@ -893,7 +893,7 @@ declare namespace Realm {
             removeAll: () => number;
         }
 
-        const MutableSubscriptions: {
+        const MutableSubscriptionSet: {
             new(): never; // This type isn't supposed to be constructed manually by end users.
             readonly prototype: SubscriptionSet;
```
kraenhansen, line 898: Should this be `MutableSubscriptionSet`?
```suggestion
            readonly prototype: MutableSubscriptionSet;
```

tomduncalf: Indeed it should

kraenhansen (approved):

PR #4299, opened 2022-02-02
Enabling integration tests to import apps into atlas

kraenhansen:
## What, How & Why?

This fixes the reading of `REALM_BASE_URL` from the app importer and adds reading of a `mongodbClusterName` from environment (mocha remote context) to enable importing of apps into the realm-dev enviroment.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests (locally and manually)


integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -23,7 +23,7 @@ export function importAppBefore(
   logLevel: Realm.App.Sync.LogLevel = (environment.syncLogLevel as Realm.App.Sync.LogLevel) || "warn",
 ): void {
   before(async function (this: Partial<AppContext> & Mocha.Context) {
-    this.timeout(10000);
+    this.timeout(120000); // Let's give it 2 minutes
```
tomduncalf, line 26: Wow, the import can take 2 minutes?

integration-tests/tests/src/utils/import-app.ts
```diff
@@ -24,24 +24,23 @@ export type ErrorResponse = { message: string; appId: never };
...
+function getUrls() {
   // Try reading the app importer URL out of the environment, it might not be accessiable via localhost
   const { appImporterUrl, mongodbRealmBaseUrl } = environment;
-  const url = typeof appImporterUrl === "string" ? appImporterUrl : "http://localhost:8091";
-
-  return { appImporterUrl: url, mongodbRealmBaseUrl: mongodbRealmBaseUrl as string };
-};
+  return {
+    appImporterUrl: typeof appImporterUrl === "string" ? appImporterUrl : "http://localhost:8091",
+    baseUrl: typeof mongodbRealmBaseUrl === "string" ? mongodbRealmBaseUrl : "http://localhost:9090",
```
tomduncalf, line 32:
```suggestion
    mongodbRealmBaseUrl: typeof mongodbRealmBaseUrl === "string" ? mongodbRealmBaseUrl : "http://localhost:9090",
```

for consistency maybe?

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -211,13 +211,29 @@ function performTests(flexibleSync: boolean) {
...
+          "services/mongodb/config.json": {
+            type: "mongodb-atlas",
+            config: {
+              clusterName: mongodbClusterName,
+              readPreference: "primary",
+              wireProtocolEnabled: false,
+            },
+          },
+        }
+      : undefined;
```
tomduncalf, line 228: Would it make sense to move this to somewhere reusable? I expect we would want to run the flexible sync tests against realm-dev, for example, and it looks like it would need the same replacement logic.

Could it be inside `importApp` for example?

tomduncalf (approved):

integration-tests/tests/src/utils/import-app.ts
kraenhansen: I'll change the `mongodbRealmBaseUrl` to `realmBaseUrl` to keep it consistent with the `REALM_BASE_URL` environment variable passed to the `realm-app-importer`, but I think in the context of this file, `baseUrl` can't be understood and it makes it simpler to add into the config of the `App` constructor below.

integration-tests/tests/src/hooks/import-app-before.ts
kraenhansen: Yeah - I was surprised. I notice this is fast when there's no database connection, I think it might be adding a change to the MongoDB server deployment to allow access from BaaS. Applying an atlas configuration takes ~ 1 minute from my experience.

tomduncalf: Ah OK, might be worth changing the comment to give a bit more context? We could change the timeout to `2 * 60 * 1000` to make the two seconds part clear

integration-tests/tests/src/utils/import-app.ts
tomduncalf: OK 👍

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: I moved this to a `getDefaultReplacements` in the `import-app.ts`.

tomduncalf: Great!

kraenhansen:
The failing tests seem unrelated to the change.

PR #4304, opened 2022-02-03
Update toJSON to gracefully handle invalid state

kraenhansen:
## What, How & Why?

This fixes an issue where an error would be thrown when trying to serialize an invalid object.

```
Error: Accessing object of type MixedClass which has been invalidated or deleted
```

Instead of erroring, i suggest simply returning an empty object / array and warn the user.

Note: We hit this error in our integration tests as the `Realm.Object` or `Realm.Collection` is getting stringified. This typically happens when a `Realm.Object` or `Realm.Collection` is stored on the Mocha context, either manually or when asserting something about an object and that assertion fails. When Mocha Remote tries to stringify the object or collection, it's been invalidated already since it sends off messages after the test has passed / failed and the after hooks have run (which close the realm, hence invalidates the object).

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


kraenhansen:
Closing this in favour of https://github.com/realm/realm-js/pull/4307.

PR #4330, opened 2022-02-10
Explicitly flush the React Native UI task queue whenever we call from C++ to JS

tomduncalf:
## What, How & Why?

React Native has its own JS "microtask" queue, implemented in https://github.com/facebook/react-native/blob/main/Libraries/Core/Timers/JSTimers.js. Any asynchronous work, e.g. `setTimeout` or `setImmediate`, is added to this queue, and the queue is flushed whenever a message is sent from React Native's native core to JS: https://github.com/facebook/react-native/blob/main/Libraries/BatchedBridge/MessageQueue.js#L108.

However, our native to JS messages are not being passed via this abstraction - instead, we hook directly into the JS engine. This means that React Native is not aware that when Realm has done some async work, we might need to update the UI (and therefore flush the task queue), and this can result in Realm-related UI updates not showing until some action is taken which causes React Native to send a message from the core to JS (e.g. touching the screen), which flushes this task queue, resolving pending promises and updating the UI.

Two solutions were considered for whenever Realm has performed some work which might require a UI update:
1. Make Realm JS "play by the rules" that React Native expects native code to use, and e.g. send a dummy message via React Native whenever we have done work.
2. Expose the internal `callReactNativeMicrotasks` function from React Native in a way that we can directly call that from C++ whenever we have done work.

We decided to go with option 2, as it is simpler to implement and we felt we can implement it in a reasonably clean way using the Realm constructor. This does mean we are depending on React Native internal implementation details, but as we already are potentially quite tightly coupled to RN versions, we didn't feel this was too much of an issue.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/jsc/jsc_function.hpp
+++ b/src/jsc/jsc_function.hpp
@@ -29,6 +29,17 @@ inline JSValueRef jsc::Function::call(JSContextRef ctx, const JSObjectRef& funct
 {
     JSValueRef exception = nullptr;
     JSValueRef result = JSObjectCallAsFunction(ctx, function, this_object, argc, arguments, &exception);
+
+    // Flush the React Native UI task queue whenever we call into JS from C++ – see `_flushUiTaskQueue` in
+    // `lib/extensions.js` for detailed explanation of why this is necessary.
+    JSObjectRef global_object = JSContextGetGlobalObject(ctx);
+    JSObjectRef realm_constructor = jsc::Object::validated_get_constructor(ctx, global_object, "Realm");
+    JSObjectRef flush_ui_task_queue =
+        jsc::Object::validated_get_function(ctx, realm_constructor, "_flushUiTaskQueue");
+    JSValueRef flush_ui_task_queue_exception = nullptr;
+    // Call it directly rather than via `call_method` to avoid an infinite loop
+    JSObjectCallAsFunction(ctx, flush_ui_task_queue, nullptr, 0, {}, &flush_ui_task_queue_exception);
+
     if (exception) {
         throw jsc::Exception(ctx, exception);
     }
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -71,6 +71,24 @@ module.exports = function (realmConstructor) {
     enumerable: false,
   });
 
+  // React Native has its own JS task queue system of "microtasks". This queue is usually flushed
+  // whenever a native -> JS call is made via the React Native bridge (see `__callReactNativeMicrotasks`
+  // in `react-native/Libraries/BatchedBridge/MessageQueue.js`).
+  //
+  // As Realm does not pass messages via the RN bridge, but instead hooks directly in to the JS engine,
+  // we need to explicitly tell React Native to flush the task queue whenever anything happens in Realm
+  // which could cause a UI update – in practice, this means any time we call into JS from C++. If we don't
+  // do this, UI updates will not be flushed until some other event occurs which does cause a native -> JS
+  // call, for example touching the screen.
+  //
+  // This exposes the React Native internal `callReactNativeMicrotasks` function on the Realm constructor,
+  // so that we can call it from C++ whenever we make a function call into JS (see jsc_function.hpp)/
+  try {
+    realmConstructor._flushUiTaskQueue = require("react-native/Libraries/Core/Timers/JSTimers.js").callReactNativeMicrotasks;
+  } catch (e) {
+    realmConstructor._flushUiTaskQueue = function () {};
+  }
+
   const getInternalCacheId = (realmObj) => {
     const { name, primaryKey } = realmObj.objectSchema();
     const id = primaryKey ? realmObj[primaryKey] : realmObj._objectId();


src/jsc/jsc_function.hpp
```diff
@@ -29,6 +29,17 @@ inline JSValueRef jsc::Function::call(JSContextRef ctx, const JSObjectRef& funct
 {
     JSValueRef exception = nullptr;
     JSValueRef result = JSObjectCallAsFunction(ctx, function, this_object, argc, arguments, &exception);
+
+    // Flush the React Native UI task queue whenever we call into JS from C++ – see `_flushUiTaskQueue` in
+    // `lib/extensions.js` for detailed explanation of why this is necessary.
+    JSObjectRef global_object = JSContextGetGlobalObject(ctx);
+    JSObjectRef realm_constructor = jsc::Object::validated_get_constructor(ctx, global_object, "Realm");
+    JSObjectRef flush_ui_task_queue =
```
kraenhansen, line 37: I think a check for this being `undefined` before calling it would be good.

src/jsc/jsc_function.hpp
```diff
@@ -29,6 +29,17 @@ inline JSValueRef jsc::Function::call(JSContextRef ctx, const JSObjectRef& funct
...
+        jsc::Object::validated_get_function(ctx, realm_constructor, "_flushUiTaskQueue");
+    JSValueRef flush_ui_task_queue_exception = nullptr;
```
kraenhansen, line 39: I wonder if we would want to throw this `flush_ui_task_queue_exception` if it does fail?

src/jsc/jsc_function.hpp
tomduncalf: Yeah, agreed, it's not a fatal error

src/jsc/jsc_function.hpp
```diff
@@ -29,6 +29,17 @@ inline JSValueRef jsc::Function::call(JSContextRef ctx, const JSObjectRef& funct
 {
     JSValueRef exception = nullptr;
     JSValueRef result = JSObjectCallAsFunction(ctx, function, this_object, argc, arguments, &exception);
+
+    // Flush the React Native UI task queue whenever we call into JS from C++ – see `_flushUiTaskQueue` in
```
kraenhansen, line 33: I wonder if it makes sense to move this to an `inline`d function that you call from [`jsc::Function::construct`](https://github.com/realm/realm-js/pull/4330/files#R57) too? It might potentially also schedule async work.

kraenhansen (approved):
🤞  this looks promising.

src/jsc/jsc_function.hpp
blagoev: I would suggest use pure JSC calls here. `validated_get_function` will throw if this function is not found. 
Then of course `extensions.js` should not set this to a an empty function.  but leave it undefined if this is not found in `JSTimers.js`

blagoev: We are already caching some things here. https://github.com/realm/realm-js/blob/master/src/jsc/jsc_init.cpp#L29
Basically you need to create a protected handle to this function on init time as we do  here https://github.com/realm/realm-js/blob/master/src/jsc/jsc_class.hpp#L60
Then you can call it any time like we do for `Obect.defineProperty` here https://github.com/realm/realm-js/blob/master/src/jsc/jsc_class.hpp#L513

tomduncalf: That's really helpful, thanks, will take a look later/tomorrow!

blagoev: and you can use `JSValueIsUndefined` to check if this one didn't get a valid function on init

src/jsc/jsc_function.hpp
```diff
@@ -29,6 +29,17 @@ inline JSValueRef jsc::Function::call(JSContextRef ctx, const JSObjectRef& funct
...
+    // `lib/extensions.js` for detailed explanation of why this is necessary.
+    JSObjectRef global_object = JSContextGetGlobalObject(ctx);
+    JSObjectRef realm_constructor = jsc::Object::validated_get_constructor(ctx, global_object, "Realm");
+    JSObjectRef flush_ui_task_queue =
+        jsc::Object::validated_get_function(ctx, realm_constructor, "_flushUiTaskQueue");
+    JSValueRef flush_ui_task_queue_exception = nullptr;
+    // Call it directly rather than via `call_method` to avoid an infinite loop
+    JSObjectCallAsFunction(ctx, flush_ui_task_queue, nullptr, 0, {}, &flush_ui_task_queue_exception);
+
     if (exception) {
```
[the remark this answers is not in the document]
kneth: In line 31 you do a call, and in line 41 you do a second call. And finally on line 43 you check if the first call did throw an exception. Furthermore, you don't check if the call in line 41 is throwning (`flush_ui_task_queue_exception`)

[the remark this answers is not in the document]
tomduncalf: Oh sorry, I didn't realise you were talking about the exception! I guess maybe there's no need to update the UI if there was an exception in the initial call, I'll think if there's any case where that would be useful. 

@kraenhansen also mentioned maybe we should check `flush_ui_task_queue_exception`... I'm undecided as flushing the task queue is, to some extent, a "nice to have" – we have not been doing this so far, and the majority of the time it works OK! – so I don't know if we want to be throwing errors if we fail to flush it for some reason, as it is "recoverable" (in that when the user next taps the screen, everything will be correct). But I'm open to opinions here.

tomduncalf:
After further testing, this fix doesn't work 100% of the time – we still see occasional freezes when running the tests, although far fewer. Needs further research – one option is sending a dummy React Native C++ -> JS event with every call into JS, which does seem to work reliably but is perhaps not an optimal solution...

tomduncalf:
Superseded by https://github.com/realm/realm-js/pull/4389

PR #4331, opened 2022-02-10
Implement non-synced Realm to synced Realm conversion

### LINKED ISSUE #4136: Implement Convert Non-Sync to Sync Realm
Requires Realm Core v10.9.0.

See also https://github.com/realm/realm-core/pull/5136

fronck:
## What, How & Why?

Refactor of the `writeCopyTo` functionality.  The method can now convert any (local/synced/encrypted) Realm format to any other format.
Some conversions are only supported when an output Realm configuration is passed to `writeCopyTo`, which is now the preferred way to use the method.

This closes #4136

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated


src/js_realm.hpp
```diff
@@ -1382,41 +1382,153 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+        realm::Realm::Config config;
+        ObjectDefaultsMap defaults;
+        ConstructorMap constructors;
+        bool schema_updated = get_realm_config(ctx, args.count, args.value, config, defaults, constructors);
+
+        realm->export_to(config);
+    }
 
-    realm->write_copy(path, encryption_key.get());
+    return;
```
kneth, line 1531: Is this needed?

src/js_realm.hpp
```diff
@@ -1382,41 +1382,153 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+            if (Value::is_valid(syncConfigValue) && Value::is_object(ctx, syncConfigValue)) {
+                copy_mode |= COPYMODE_TOSYNC;
+            }
+            else {
+                copy_mode |= COPYMODE_TOLOCAL;
+            }
+        } break;
+
+        case COPYMODE_WITHPATH: {
+            if (args.count > 2) {
```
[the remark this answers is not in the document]
fronck: True.  I left it in intentionally, though.  Given that the C++ code is resolving function overloading from JS, I wanted to have each individual cope path validate its inputs.

src/js_realm.hpp
fronck: No.  I'll ditch it.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Added `Realm.App#deleteUser(user)` to delete a sync user from a MongoDB Realm app. ([#4006](https://github.com/realm/realm-js/issues/4006))
+* Extended `Realm.writeCopyTo()` functionality to allow conversion of non-synced Realms to synced Realms ([#4136](https://github.com/realm/realm-js/issues/4136)).
```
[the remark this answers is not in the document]
fronck: 👍

docs/realm.js
```diff
@@ -307,16 +307,24 @@ class Realm {
   compact() {}
 
   /**
-   * Writes a compacted copy of the Realm to the given path.
+   * configuration.
```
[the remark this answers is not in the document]
fronck: It's queued locally -- I'm bundling up changes and pushing once.

types/index.d.ts
```diff
@@ -1153,6 +1153,19 @@ declare class Realm {
      */
     writeCopyTo(path: string, encryptionKey?: ArrayBuffer | ArrayBufferView): void;
 
+    /**
```
tomduncalf, line 1156: Just a question – I notice that the wording between this and the JSDoc is slightly different (because the JSDoc has to cover both cases in one comment). 

If we were to move to generating docs from the Typescript definitions exclusively, are you happy that the docs for both forms of `writeCopyTo` in the Typescript definitions cover everything that is in the JSDoc? It appears so, but just confirming.

src/js_realm.hpp
```diff
@@ -1382,41 +1382,151 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+ * @param this_object Realm object passed from JS
+ * @param args Either `<string>, [string]`, or `object` -- see function brief.
+ * @param return_value none
+ */
 template <typename T>
 void RealmClass<T>::writeCopyTo(ContextType ctx, ObjectType this_object, Arguments& args, ReturnValue& return_value)
 {
     args.validate_maximum(2);
 
+    enum CopyMode {
```
tomduncalf, line 1404: It's probably more of a reflection on the kind of software I've written (I've not come across that much "working with bits" code, so I'm not that comfortable reading it), and I am quite possibly missing something quite obvious, but I'd be interested to know the advantages of doing it this way versus something more "explicit" e.g.

```
enum CopyType {
  WithPath,
  WithConfig
};

enum CopyDestination {
  ToSync,
  ToLocal
};

CopyType copy_type = CopyType.WithPath;
CopyDestination copy_destination = CopyDestination.ToSync; 
bool copy_encrypted = false;

// ...

if (Value::is_object(ctx, firstParamValue)) {
  copy_type = CopyType.WithConfig;
}

// ... 

switch (copy_type) {
  case CopyType.WithConfig:
    // ...
    if (Value::is_valid(syncConfigValue) && Value::is_object(ctx, syncConfigValue)) {
      copy_destination = CopyDestination.ToSync;
    }
    else {
     copy_destination = CopyDestination.ToLocal;
    }
    break;
  case CopyType.WithPath:
    // ....
}

// etc ...

if (copy_destination == CopyDestination.ToLocal) {
  realm->write_copy(output_path, encryption_key.get());
}
else if (copy_destination == CopyDestination.ToSync) {
  // ... 
}
````

File: tests/js/session-tests.js [hunk 7]
```diff
@@ -1152,6 +1152,225 @@ module.exports = {
...
+    const realm3Path = realm1Path + "copy3.realm";
+    realm1.writeCopyTo(realm3Path);
+
+    /*
+      Test 3:  open a copy of our realm with a new user and a new
+        partition key.  We expect it to fail because of the mismatch
+        in partition keys
+    */
+    const user3 = await app.logIn(credentials3);
+    const otherPartition = Utils.genPartition();
```
tomduncalf, line 1279:
```suggestion

    /*
      Test 3:  open a copy of our realm with a new user and a new
        partition key.  We expect it to fail because of the mismatch
        in partition keys
    */
    const realm3Path = realm1Path + "copy3.realm";
    realm1.writeCopyTo(realm3Path);

    const user3 = await app.logIn(credentials3);
    const otherPartition = Utils.genPartition();
```

Maybe slightly clearer if this action comes after the description?

tomduncalf (approved):

docs/realm.js
```diff
@@ -307,16 +307,23 @@ class Realm {
...
-   * Writes a compacted copy of the Realm to the given path.
+   * Writes a compacted copy of the Realm α) to the given path or β) with the given output
+   *
+   * For invocation with α):
+   *   * Input Realms may be local or synced, encrypted or non-encrypted
+   *   * Output Realms will be local only, encrypted or non-encrypted
+   *
+   * For invocation with β):
+   *   * Input Realms may be local or synced, encrypted or non-encrypted
+   *   * Output Realms will be local or synced, encrypted or non-encrypted, depending on the configuration passed to the function
```
takameyer, line 318: I think we should have some examples of what one could achieve with this function.  

- encrypt unencrypted realm
- attempt to make a smaller realm (compaction)
- convert a local realm to a sync-realm

And as stated below, focus on the `configuration` as the input argument and perhaps make a note that `(path, encryptionKey)` is deprecated and support may be removed in the next major release.

takameyer: We should also document that the configuration must contain a `path` and will perform encryption if given an `encryptionKey`

src/js_realm.hpp
```diff
@@ -1382,41 +1382,151 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+                throw std::runtime_error("`path` property must exist in output configuration");
+            }
+            else {
+                throw std::runtime_error("`path` property must be a string");
+            }
+
+            // check whether encryption key property exists, and whether it's a string
+            ValueType encKeyValue = Object::get_property(ctx, output_config, "encryptionKey");
+            // `encryptionKey` is optional..
+            if (Value::is_valid(encKeyValue)) {
```
takameyer, line 1459: does `is_valid` return true if it's not set?

src/js_realm.hpp
```diff
@@ -1382,41 +1382,151 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+                if (Value::is_string(ctx, encKeyValue)) {
+                    copy_mode |= COPYMODE_ENCRYPTED;
+                    encryption_key = Value::to_binary(ctx, encKeyValue);
+                }
+            }
+            else {
+                throw std::runtime_error("encryptionKey property must be a Binary");
```
takameyer, line 1466: Does `is_valid` prove that its not a binary?

src/js_realm.hpp
```diff
@@ -1382,41 +1382,151 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+            else if (!Value::is_valid(pathValue)) {
+                throw std::runtime_error(
+                    "`path` parameter must be first parameter to `writeCopyTo(<path>, [encryption key])`");
+            }
+            else {
+                throw std::runtime_error("`path` parameter must be a string");
+            }
+
+            if (args.count == 2) {
+                // a second parameter is given -- it must be an encryption key for the destination Realm
```
takameyer, line 1500: Why does this act differently to the configuration copy mode?  Couldn't this also transfer a string to binary?

tests/js/session-tests.js
```diff
@@ -1152,6 +1152,225 @@ module.exports = {
...
+
+    realmSynced.close();
+    realmLocalCopy.close();
+    realmLocal.close();
+  },
+
+  /*
+    Test the functionality of the writeCopyTo() operation on synced realms.
+  */
+  testSyncWriteCopyTo2: async function () {
```
takameyer, line 1367: My suggestion would be to change this one to `testSyncWriteCopyToWithPath` and the other to `testSyncWriteCopyToWithConfig`

takameyer (commented):
I have a few questions that I would like to be addressed before this is merged.

src/js_realm.hpp
fronck: `is_valid` returns the equivalent of
`return _value == nullptr;`
I.e., it tells you whether the `Value`'s internal data pointer points to something that isn't `null`

src/js_realm.hpp
fronck: No.  I'm missing an `else` guard here, thanks.

src/js_realm.hpp
fronck: I'm not sure what you mean.  The configuration copy mode only takes a single parameter to `writeCopyTo`

tests/js/session-tests.js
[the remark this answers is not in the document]
fronck: It's not different.  It's a relic from when I has two implementations of `writeCopyTo`.  I'll get rid of it, thanks.

src/js_realm.hpp
takameyer: The configuration mode takes the `encryptedKey` value from the configuration, tests if it is a string and then transforms it into a binary. (lines 1460 to 1463).
If the second argument was a string instead of a ArrayBuffer, then couldn't we also convert the argument into a binary instead of throwing an exception?

types/index.d.ts
fronck: I believe that it should be covered.  It's just much easier to do in two separate blocks 😄

tests/js/session-tests.js
fronck: 👍

src/js_realm.hpp
```diff
@@ -1382,41 +1382,151 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
 void RealmClass<T>::writeCopyTo(ContextType ctx, ObjectType this_object, Arguments& args, ReturnValue& return_value)
 {
     args.validate_maximum(2);
 
+    enum CopyMode {
+        COPYMODE_WITHPATH = 1,
+        COPYMODE_WITHCONFIG = 1 << 1,
+        COPYMODE_TOSYNC = 1 << 2,
+        COPYMODE_TOLOCAL = 1 << 3,
+        COPYMODE_ENCRYPTED = 1 << 4,
```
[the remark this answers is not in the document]
fronck: N/A in the refactored code.

tests/js/session-tests.js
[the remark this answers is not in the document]
fronck: N/A in the refactored code.

src/js_realm.hpp
fronck: N/A in the refactored implementation

src/js_realm.hpp
```diff
@@ -1382,41 +1390,200 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
     return_value.set(realm->compact());
 }
 
+/**
+ * @brief Helper function for `writeCopyTo()` -- parses and validated parameters in a config structure passed from JS
```
kneth, line 1394:
```suggestion
 * @brief Helper function for `writeCopyTo()` -- parses and validates parameters in a config structure passed from JS
```

src/js_realm.hpp
```diff
@@ -1382,41 +1390,200 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
 
-    if (args.count == 0) {
-        throw std::runtime_error("At least path has to be provided for 'writeCopyTo'");
+    // validate 1)
+    if (args.count > 1) {
+        throw std::runtime_error("`writeCopyTo(<config>)` accepts only one parameter");
     }
 
-    SharedRealm realm = *get_internal<T, RealmClass<T>>(ctx, this_object);
+    // validate 2)
```
kneth, line 1417: Why not use `validated_to_object(args[0], "`config` parameter must be an object")`?

src/js_realm.hpp
```diff
@@ -1382,41 +1390,200 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+     * 2) args[0] must be an object
+     * 3) args[0].path must be present and be a string
+     * 4) args[0].encryptionKey may be present, and must be a binary if it is present
+     * 5) args[0].sync may be present, and must be an object if it is present
+     */
 
-    if (args.count == 0) {
-        throw std::runtime_error("At least path has to be provided for 'writeCopyTo'");
+    // validate 1)
+    if (args.count > 1) {
```
kneth, line 1413: What if there is zero arguments? Will `args[0]` fail? Or can we assume that `writeCopyTo()` is validating?

You could use `validate_argument_count(args.count, 1, "`writeCopyTo(<config>)` accepts only one parameter")` to avoid the `if` and `throw`.

tests/js/session-tests.js
```diff
@@ -1297,6 +1297,274 @@ module.exports = {
...
+      schema: [schemas.DogForSync],
+      path: "dogsSynced.realm",
+    };
+
+    // shim config for the synced realms
+    const configSyncEnc = {
+      sync: {
+        // TODO: export_to doesn't fail is there's no user..?
+        // TODO:  Realm.open() fails after export_to() of synced -> local
+        user: user,
```
kneth, line 1380:
```suggestion
        user,
```

kneth (approved):
I like the comments in the C++ code, in particular on the rules on how to validate the input arguments.

My comments are only minor, and I don't need to review again before merging.

src/js_realm.hpp
fronck: Done :+1:

src/js_realm.hpp
fronck: Done :+1:

src/js_realm.hpp
fronck: Fixed :+1:

takameyer (approved):
Just a few suggestions, but on the whole I approve.
Is there documentation somewhere on how to convert from sync to non-sync realms?  
I was thinking we should provide examples on how to do that, but that is probably better placed in docs pages.

docs/realm.js
takameyer: I think these examples are better left to the docs.  I assume a user would land on this method trying to implement conversions with guidance from documentation or customer support.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* `Realm.writeCopyTo()` can now be invoked with a `Realm.Configuration` as parameter, which will create a copy of the Realm the conforms to the given configuration -- this allows for a much wider range of conversions between synced/local Realms and encrypted/non-encrypted Realms.
```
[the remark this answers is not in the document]
fronck: :+1:

docs/realm.js
```diff
@@ -307,18 +307,25 @@ class Realm {
...
    *
```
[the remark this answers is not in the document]
fronck: :+1:

docs/realm.js
```diff
@@ -307,18 +307,25 @@ class Realm {
...
-   * When invoked on a synced realm, a copy of the realm is created that any user can open and
-   * resume synchronization with the server.
    *
    * Note that if this method is called from within a write transaction, the current data is written,
    * not the data from the point when the previous write transaction was committed.
-   * @param {string} path path to save the Realm to.
-   * @param {ArrayBuffer|ArrayBufferView} [encryptionKey] - Optional 64-byte encryption key to encrypt the new file with.
+   * @param {string|Realm~Configuration} pathOrConfig a α) path to save the Realm to, OR β) {@link Realm~Configuration | Configuration} that describes the output realm.
+   * @param {ArrayBuffer|ArrayBufferView} [encryptionKey] - Optional 64-byte encryption key to encrypt the new file with.  Must not be present when
+   * β) a {@link Realm~Configuration | Configuration} is given as first parameter.
```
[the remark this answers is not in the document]
fronck: :+1:

src/js_realm.hpp
```diff
@@ -1382,41 +1390,194 @@ void RealmClass<T>::compact(ContextType ctx, ObjectType this_object, Arguments&
...
+template <typename T>
+realm::Realm::Config RealmClass<T>::write_copy_to_helper_deprecated(ContextType ctx, ObjectType this_object,
+                                                                    Arguments& args)
+{
+    /* Validation rules:
+     * 1) there must be one or two parameters
+     * 2) first parameter must be a string
+     * 3) second parameter, if present, must be a binary
+     */
+
```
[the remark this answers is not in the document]
fronck: :+1:

PR #4338, opened 2022-02-11
Fix types for Object Listener

### LINKED ISSUE #4337: Improve `Realm.Object#addListener` type signature
### Problem

Our current types of `Realm.Object#addListener` are currently not specific enough, they take a callback of a function that takes any `Realm.Object` not the concrete type of object that the listener got added on.

### Solution

Here's a small snippet from the TypeScript playground to show a possible solution, using the `this` type.

```typescript
type ListenerCallback<T> = (obj: T) => void;

abstract class RealmObject {
  addListener(callback: ListenerCallback<this>): void {
    // 
  }
}

class Car extends RealmObject {
  name: string = "";
  age: number = 123;
}

const car = new Car();

car.addListener(obj => {
  const keys = Object.keys(obj) as (keyof typeof obj)[];
  keys.every(key => {
    const value = obj[key];
  });
});
```

takameyer:
When a class extends Realm.Object, adding a listener to an instance of the
object would not correctly type the object returned in the callback.
With this change, the callback returns an object of the same type as the
instance it was called from.

This closes #4337

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


.gitignore-revs
```diff
@@ -1,3 +1,4 @@
 c76131748cbb048e78a987eb234d8a9609e26430
 f38d8cf7cac992767a451d2f3cf3310e3546bf6f
 fa80f12bf3fbfe1337ae69e12c59051fac347001
+99e64a8e990d2579bcfe382f755edbc61f70af4e
```
kraenhansen, line 4: Was this on purpose?

takameyer: Yes...the types document still had our merge commit from version 10.5.0

takameyer: Somehow that seems to exist across multiple SHAs.

kraenhansen (approved):

types/index.d.ts
```diff
@@ -251,16 +251,16 @@ declare namespace Realm {
         /**
          * @returns void
          */
-        addListener(callback: ObjectChangeCallback): void;
+        addListener(callback: ObjectChangeCallback<this>): void;
```
tomduncalf, line 254: Nice trick ✨

takameyer: Game changer.

PR #4351, opened 2022-02-16
Add an example of connecting a Realm React app to sync

tomduncalf:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/example/App.tsx
+++ b/packages/realm-react/example/App.tsx
@@ -25,19 +25,23 @@ import { Realm } from "@realm/react";
 import TaskContext from "./app/models/Task";
 import { AppMain } from "./AppMain";
 
+const APP_ID = "<your app ID>";
+
 // 2. Add your Realm app details here, using whatever login mechanism you wish
-const app = new Realm.App({ id: "application-0-bzsvu" });
+const app = new Realm.App({ id: APP_ID });
 const credentials = Realm.Credentials.anonymous();
 
 const App = () => {
-  // 3. Store the logged in user in state
-  const [user, setUser] = useState<Realm.User | undefined>();
+  // 3. Store the logged in user in state, this will be null when you first start the app
+  const [user, setUser] = useState<Realm.User | null>(app.currentUser);
 
   const { RealmProvider } = TaskContext;
 
-  // 4. On application startup, login the user
+  // 4. On application startup, login the user if not already logged in
   useEffect(() => {
     (async () => {
+      if (user) return;
+
       const loggedInUser = await app.logIn(credentials);
       // 5. Set the logged in user in state
       setUser(loggedInUser);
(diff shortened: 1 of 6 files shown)


packages/realm-react/example/App.tsx
```diff
@@ -15,114 +15,44 @@
...
 
-function App() {
-  const realm = useRealm();
-  const result = useQuery(Task);
+import TaskContext from "./app/models/Task";
+import { AppMain } from "./AppMain";
 
-  const tasks = useMemo(() => result.sorted("createdAt"), [result]);
+// 2. Add your Realm app details here, using whatever login mechanism you wish
+const app = new Realm.App({ id: "application-0-bzsvu" });
```
takameyer, line 29: Is that a valid App ID?

packages/realm-react/example/App.tsx
```diff
@@ -15,114 +15,44 @@
...
-          <TaskList tasks={tasks} onToggleTaskStatus={handleToggleTaskStatus} onDeleteTask={handleDeleteTask} />
-        )}
-      </View>
-    </SafeAreaView>
-  );
-}
+  // 4. On application startup, login the user
+  useEffect(() => {
+    (async () => {
+      const loggedInUser = await app.logIn(credentials);
```
takameyer, line 41: I think we could also check for current user here, so that we don't create a bunch of anon users when reloading this code.

packages/realm-react/example/App.tsx
```diff
@@ -15,114 +15,44 @@
...
+const credentials = Realm.Credentials.anonymous();
 
-  const handleAddTask = useCallback(
-    (description: string): void => {
-      if (!description) {
-        return;
-      }
+const App = () => {
+  // 3. Store the logged in user in state
+  const [user, setUser] = useState<Realm.User | undefined>();
```
takameyer, line 34: This could be initialised with `app.currentUser`

takameyer: Just thinking about possible hooks.  We might be able to extract a lot of this into a `useAnonUser` hook:
```
const user = useAnonUser(app)
```
--or--
```
const {userAnonUser} = createAppHooks(app)
const App = () => {
const user = useAnonUser();
...
}
```

takameyer: It's also possible we will want an AppProvider...but perhaps just a generator function that returns hooks with an `Realm.App` as an argument would suffice.

takameyer:
I put a bunch of my thoughts down.  I think this is already a good example on how someone could get up and going with Sync in their app.  We should probably try out all the different auth possibilities and think about what could make sense to help users with their authentication flow.

PR #4353, opened 2022-02-17
How to disable analytics

kneth:
When we build and test locally, we should not submit analytics to the DW.

### DIFF
--- a/contrib/building.md
+++ b/contrib/building.md
@@ -51,6 +51,12 @@ The following dependencies are required. All except Xcode can be installed by fo
 - [Android NDK 21.0](https://developer.android.com/ndk/downloads/index.html)
 - [Android CMake](https://developer.android.com/ndk/guides/cmake)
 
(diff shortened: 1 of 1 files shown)


tomduncalf (approved):

contrib/building.md
```diff
@@ -51,6 +51,12 @@ The following dependencies are required. All except Xcode can be installed by fo
 - [Android NDK 21.0](https://developer.android.com/ndk/downloads/index.html)
 - [Android CMake](https://developer.android.com/ndk/guides/cmake)
 
+Moreover, in order to avoid introducting false positives in our analytics dataset, it is highly recommended to disable analytics by add the following to your shell configuration:
```
takameyer, line 54:
```suggestion
Moreover, in order to avoid introducing false positives in our analytics dataset, it is highly recommended to disable analytics by adding the following to your shell configuration:
```

takameyer (changes requested):

takameyer (approved):
Just fixing some spelling and grammar.  Then it can be merged :)

PR #4163, opened 2021-12-20
New analytics

kneth:
## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 📝 `Compatibility` label is updated or copied from previous entry~
* [x] 🚦 Tests
* ~[ ] 📱 Check the React Native/other sample apps work if necessary~
* ~[ ] 📝 Public documentation PR created or is not necessary~
* ~[ ] 💥 `Breaking` label has been applied or is not necessary~

*If this PR adds or changes public API's:*
* ~[ ] typescript definitions file is updated~
* ~[ ] jsdoc files updated~
* ~[ ] Chrome debug API is updated if API is available on React Native~

PR #4355, opened 2022-02-17
Fix bug in submit analytics script when running "npm install" in the realm-js repository

kneth:
## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/scripts/submit-analytics.js
+++ b/scripts/submit-analytics.js
@@ -118,8 +118,25 @@ function isAnalyticsDisabled() {
 }
 
 function getRealmVersion() {
-  const packageJson = fse.readJsonSync([getProjectRoot(), "node_modules", "realm", "package.json"].join(path.sep));
-  return packageJson["version"];
+  let version = "unknown";
+  let packageJsonPath = "";
+  try {
+    packageJsonPath = [getProjectRoot(), "package.json"].join(path.sep);
+    const packageJson = fse.readJsonSync(packageJsonPath);
+    if (packageJson["name"] === "realm") {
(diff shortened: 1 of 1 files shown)


scripts/submit-analytics.js
```diff
@@ -118,8 +118,25 @@ function isAnalyticsDisabled() {
...
 function getRealmVersion() {
-  const packageJson = fse.readJsonSync([getProjectRoot(), "node_modules", "realm", "package.json"].join(path.sep));
-  return packageJson["version"];
+  let version = "unknown";
+  let packageJsonPath = "";
+  try {
+    packageJsonPath = [getProjectRoot(), "package.json"].join(path.sep);
+    const packageJson = fse.readJsonSync(packageJsonPath);
+    if (packageJson["name"] === "realm") {
+      version = packageJson["version"];
```
kraenhansen, line 127: This will simply fall through and report "unknown" if no error is thrown?

I suggest [relying on the `npm_config_local_prefix`](https://github.com/realm/realm-js/pull/4163#discussion_r792568357). If that directory's package.json has "realm" as its name, simply skip submitting analytics all together.

kneth: It is using `npm_config_local_prefix` through `getProjectRoot()`. But in some cases (in `integration-tests` if I remember correctly), it is not set.

 > If that directory's package.json has "realm" as its name, simply skip submitting analytics all together.

It makes sense to add this to `isAnalyticsDisabled()`.

PR #4374, opened 2022-02-25
Add script to generate JS template apps from TS

tomduncalf:
## What, How & Why?

Currently we have to manually sync our JS template apps with our TS template apps, which has led to them slipping out of sync. 

This PR adds a script which strips the types out of the TS apps and writes the output into the JS template apps, using the [detype](https://www.npmjs.com/package/detype) npm package (tsc does not preserve comments and whitespace).

It also updates the JS apps with the output of the script.

A future enhancement might be to have the `react-native-template-realm-ts` as the "master" app, then we copy the `app` directory from there into the Expo template as part of this script, and generate the JS apps. This would mean we only have to maintain one set of app code, but each template could have its own supporting code.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/templates/scripts/convert-ts-to-js.sh
+++ b/templates/scripts/convert-ts-to-js.sh
@@ -11,6 +11,8 @@ convert() {
   pushd $1
   # The Babel config causes detype to output non-ES2015 JS, so temporarily hide it
   mv babel.config.js babel.config.js.ignore
+  # Delete any generated JS files in the target directory in case we have renamed or deleted something
+  find $2/app -name "*.js" -exec rm {} +
   # Strip types from every .ts* file (except those in node_modules) and write the output
   # to the corresponding file in react-native-template-realm-js, with the extension changed
   # to ".js" (regardless of whether the original was ".tsx" or ".ts")
--- a/templates/scripts/convert-ts-to-js.sh
+++ b/templates/scripts/convert-ts-to-js.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# This script converts the TypeScript tempalte in `react-native-template-realm-ts` to plain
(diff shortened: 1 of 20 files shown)


templates/scripts/convert-ts-to-js.sh
```diff
@@ -0,0 +1,24 @@
...
+set -o pipefail
+
+convert() {
+  pushd $1
+  # The Babel config causes detype to output non-ES2015 JS, so temporarily hide it
+  mv babel.config.js babel.config.js.ignore
+  # Strip types from every .ts* file (except those in node_modules) and write the output
+  # to the corresponding file in react-native-template-realm-js, with the extension changed
+  # to ".js" (regardless of whether the original was ".tsx" or ".ts")
+  for i in `find . -path ./node_modules -prune -o -name "*.ts*" -print`; do npx detype $i $2/${i/\.ts*/\.js} ; done
```
tomduncalf, line 17: @takameyer I'm wondering if maybe this should only convert files in `app/`, and delete all the `*.ts*` files in there first in case we delete/rename a file... WDYT?

takameyer: That makes sense.  We should probably move the `App.tsx` into `app/` as well?

takameyer (approved):
This is amazing!  We will have to ensure to get this into our CI at some point

templates/scripts/convert-ts-to-js.sh
tomduncalf: See https://github.com/realm/realm-js/pull/4374/commits/a4917cbaf1f28718287ea7f4b53d831eb5b0f76e

Issue #2869, opened 2020-05-08
Refactor package structure


RealmBot:
* Make realm-web, realm-network-transport and integration tests as Lerna packages
* Move to full Lerna packages or NPM workspaces once v10.0.0 is stabilized and its in master
* Discuss later to split realm-js into node.js and RN packages
* Restructure CI for specific platforms once realm-js has been moved

https://docs.google.com/document/d/1YIjSSdxykoSv8wlpgyNlQkb47tESeKYDr5yx9sneLpk/edit

kraenhansen:
Considering the (lack of) maintenance of Lerna and the maturity of NPM workspaces, I would like for us to consider migrating to a NPM workspaces based solution.

PR #4378, opened 2022-02-28
Migrating from Lerna to NPM workspaces

kraenhansen:
## What, How & Why?

This closes (part of) #2869

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (tested running the tests manually locally)


package.json
```diff
@@ -141,7 +152,7 @@
     "shelljs": "^0.8.5",
     "shx": "^0.2.2",
     "ts-node": "^10.4.0",
-    "typescript": "^4.5.2"
+    "typescript": "4.4.4"
```
tomduncalf, line 155: Was this downgrade intentional?

packages/realm-react/example/android/app/build.gradle
```diff
@@ -120,6 +120,11 @@ def jscFlavor = 'org.webkit:android-jsc:+'
  */
 def enableHermes = project.ext.react.get("enableHermes", false);
 
+/**
+ * Architectures to build native code for in debug.
+ */
+def nativeArchitectures = project.getProperties().get("reactNativeDebugArchitectures")
```
tomduncalf, line 126: Are these changes a result of us standardising on a single version of dependencies or something else? Also are the various lint fixes a result of us unifying our lint configs?

tomduncalf:
I think this looks great, nice to remove complexity, good work!

So with npm workspaces, there's no longer the concept of "bootstrapping" an individual subpackage, you always just run `npm i --workspaces` from the root?

I tried running `packages/realm-react/example` by:

1. `npm i --workspaces`
2. `cd packages/realm-react/example`
3. `npx pod-install`

And I get:

```
[!] No podspec found for `React-logger` in `../../node_modules/react-native/ReactCommon/logger`
Couldn't install Pods. Updating the Pods project and trying again...
Command `pod install` failed.
└─ Cause: No podspec found for `React-logger` in `../../node_modules/react-native/ReactCommon/logger`
```

Not sure what is wrong here, I can't see an obvious hard coded path that is causing it...

contrib/building.md
```diff
@@ -266,13 +266,9 @@ The generated docs can be found in `docs/output/realm/<version>/index.html`.
 We've decided to slowly migrate this repository to a mono-repository containing multiple packages (stored in the `./packages` directory). To install and link these, run (from the `realm-js` repo root directory):
 
 ```sh
-npx lerna bootstrap
+npm install --workspaces
 ```
 
-Note: you must successfuly build Realm JS for [iOS](#building-for-ios) and [Android](#building-for-android) before running `lerna`, or the command may fail.
-
-Please familiarise yourself with [Lerna](https://github.com/lerna/lerna) to learn how to add dependencies to these packages.
```
takameyer, line 274: Any such doc for `workspaces` that we should reference?

integration-tests/environments/react-native/ios/Podfile
```diff
@@ -1,5 +1,5 @@
-require_relative '../node_modules/react-native/scripts/react_native_pods'
-require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+require_relative '../../../../node_modules/react-native/scripts/react_native_pods'
```
takameyer, line 1: Does it not install RN in the project folder?

integration-tests/package.json
```diff
@@ -4,13 +4,7 @@
   "description": "A wrapper module that can install and run Realm JS tests in various environments",
   "private": true,
   "scripts": {
-    "start": "npm start --prefix tests --",
-    "test": "npm run build --prefix tests && concurrently -m 1 npm:test:*",
-    "test:node": "npm test --prefix environments/node",
```
takameyer, line 9: The readme should be updated to reflect these changes

package.json
kraenhansen: Yeah - unfortulately the eslint plugins aren't compatible with the newer version. I suggest a followup PR to get this and the eslint plugins upgraded.

packages/metro-config/index.js
```diff
@@ -17,132 +17,24 @@
...
-          for (const i in candidatePaths) {
-            console.warn(`- ${candidatePaths[i]}`, i === 0 ? " (returning this)" : "");
-          }
-        }
-        return candidatePaths[0];
-      } else {
-        console.error("Failed to resolve", name);
-        return undefined;
-      }
+function build({ projectRoot, watchFolders }) {
```
takameyer, line 22: Did `workspaces` fix this complexity?  Or is this just a new discovery for making the symbolic links problem go away in a simpler way?

packages/realm-react/example/package.json
```diff
@@ -3,33 +3,32 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "android": "react-native run-android",
-    "ios": "react-native run-ios",
+    "android": "concurrently -n metro,rn npm:start 'react-native run-android'",
```
takameyer, line 6: What is this doing?

package.json
tomduncalf: OK, I was trying to remember if there was a specific reason I upgraded to 4.5.2. I don't know if there was – so I guess if everything typechecks then this is fine for now

integration-tests/environments/react-native/ios/Podfile
kraenhansen: With npm workspaces common dependencies are hoisted to the highest level, in this case the root's `node_modules`.

integration-tests/package.json
kraenhansen: Ahh. Thanks I got away from that.
... I wonder why I deleted those 🤔

packages/metro-config/index.js
kraenhansen: It was possible to really simplify this, which is great! 😄 👍

packages/realm-react/example/android/app/build.gradle
kraenhansen: These are simply from upgrading the React Native version of the realm-react/example app. I wanted to align on a single version of React Native across the sub-packages. Strictly that might not have been needed, I could revert if this is an issue.

packages/metro-config/index.js
takameyer: That IS great! 👍🏼

kraenhansen:
> So with npm workspaces, there's no longer the concept of "bootstrapping" an individual subpackage, you always just run `npm i --workspaces` from the root?

I had to revert this, it turns out the `--workspaces` is already default on these (install, update, etc.) commands.

As [the docs](https://docs.npmjs.com/cli/v8/commands/npm-install#workspaces) say:

> Commands that operate on the node_modules tree (install, update, etc.) will link workspaces into the node_modules folder.

contrib/building.md
kraenhansen: I added a link to the NPM docs: https://docs.npmjs.com/cli/v8/using-npm/workspaces

integration-tests/package.json
kraenhansen: I added them back in.. I started this branch trying to do something else ...

packages/realm-react/example/package.json
kraenhansen: The React Native CLI is not happy to be run when it's not a located in the `node_modules` of the app.
More specifically the metro server which pops up has an issue with this, which is why I suggest migrating to something that doesn't "pop up" the Metro terminal.

integration-tests/environments/react-native/ios/RealmReactNativeTests.xcodeproj/project.pbxproj
```diff
@@ -393,7 +393,7 @@
...
+			shellScript = "export RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../../../../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../../../../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
```
kneth, line 396: Should we move the shell script of the Xcode project? It will make it easier to review changes in the future.

lerna.json
```diff
@@ -1,12 +0,0 @@
-{
```
kneth, line 1: Nice :smile:

package.json
```diff
@@ -77,6 +77,20 @@
     "lint:fix:cpp": "find src -name \"*.?pp\" | xargs clang-format -i",
     "postinstall": "node scripts/submit-analytics.js"
   },
+  "workspaces": [
```
kneth, line 80: Is it a requirement that we manually maintain the list of workspaces?

packages/realm-react/example/ios/RealmReactExample.xcodeproj/project.pbxproj
```diff
@@ -393,7 +393,7 @@
...
+			shellScript = "export RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../../../../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../../../../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
```
kneth, line 396: Same question as before: should we move the script out of Xcode?

packages/metro-config/index.js
```diff
@@ -17,132 +17,24 @@
...
+  const extraNodeModules = new Proxy(
```
kneth, line 23: What a simplification 🎉

kneth (commented):
I appreciate that we are moving away from unmaintained Lerna. Using the features of the standard package manager is a good idea.

package.json
kraenhansen: Yeah. We can use wildcards here, but scripts (ex `npm run build --workspaces`) are ran on the sub-packages in the order they appear in this list, so for things like the `realm-network-transport` where some packages are depending on it being built before they can build themselves, we need to be explicit on this order.

This is a drawback to how Lerna worked, where packages would be installed based on their dependency ordering (any package depending on "package A" would be installed (hence built) after "package A").

packages/metro-config/index.js
kraenhansen: Yeah - this is awesome.

packages/realm-react/example/ios/RealmReactExample.xcodeproj/project.pbxproj
kraenhansen: I've deleted it entirely.

kraenhansen:
This is awaiting merge and release of https://github.com/vercel/turborepo/pull/1145.

example/ios/RealmTsTemplate.xcodeproj/project.pbxproj
```diff
@@ -376,25 +375,6 @@
 			shellScript = "\"${PODS_ROOT}/Target Support Files/Pods-RealmTsTemplate-RealmTsTemplateTests/Pods-RealmTsTemplate-RealmTsTemplateTests-resources.sh\"\n";
 			showEnvVarsInLog = 0;
 		};
-		FD10A7F022414F080027D42C /* Start Packager */ = {
```
kraenhansen, line 379: This is being removed, because it's not really needed and it expects `react-native` to be located in the `./node_modules` relative to the app.

integration-tests/environments/react-native/ios/RealmReactNativeTests.xcodeproj/project.pbxproj
```diff
@@ -377,25 +376,6 @@
 			shellScript = "\"${PODS_ROOT}/Target Support Files/Pods-RealmReactNativeTests-RealmReactNativeTestsTests/Pods-RealmReactNativeTests-RealmReactNativeTestsTests-resources.sh\"\n";
 			showEnvVarsInLog = 0;
 		};
-		FD10A7F022414F080027D42C /* Start Packager */ = {
```
kraenhansen, line 380: See comment above for `example/ios/RealmTsTemplate.xcodeproj/project.pbxproj`.

PR #4411, opened 2022-03-14
Lint: Importing Realm to fix linting realm-react/testApp

kraenhansen:
## What, How & Why?

This fixes three lint errors in the `realm-react/testApp`:

```
realm-js/packages/realm-react/testApp/App.tsx
  56:19  error  'Realm' is not defined  no-undef
  80:19  error  'Realm' is not defined  no-undef

realm-js/packages/realm-react/testApp/app/components/TaskItem.tsx
  25:16  error  'Realm' is not defined  no-undef

realm-js/packages/realm-react/testApp/app/components/TaskList.tsx
  25:10  error  'Realm' is not defined  no-undef
  25:31  error  'Realm' is not defined  no-undef
  26:37  error  'Realm' is not defined  no-undef
  27:31  error  'Realm' is not defined  no-undef

✖ 7 problems (7 errors, 0 warnings)
```

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 🚦 Tests
* [x] 📱 Check the React Native/other sample apps work if necessary
* [x] 📝 Public documentation PR created or is not necessary
* [x] 💥 `Breaking` label has been applied or is not necessary

PR #4416, opened 2022-03-15
Fix crash when accessing out of bounds index of a collection

takameyer:
Collection will now return `undefined` for out of bound indexes.
Added a test to cover this.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/cachedCollection.ts
+++ b/packages/realm-react/src/cachedCollection.ts
@@ -24,8 +24,8 @@ function getCacheKey(id: string) {
 }
 
 /**
-  * Arguments object for {@link cachedCollection}.
-  */
+ * Arguments object for {@link cachedCollection}.
+ */
 type CachedCollectionArgs<T> = {
   /**
    * The {@link Realm.Collection} to proxy
@@ -40,7 +40,7 @@ type CachedCollectionArgs<T> = {
    * (derived) version of the collection to reuse the same cache, preventing excess new object
    * references being created.
    */
-  collectionCache?: Map<string, T>;
+  objectCache?: Map<string, T>;
   /**
    * Optional flag specifying that this is a derived (`sorted` or `filtered`) version of
    * an existing collection, so we should not create or remove listeners or clear the cache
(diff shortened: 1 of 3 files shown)


packages/realm-react/src/cachedCollection.ts
```diff
@@ -100,6 +95,12 @@ export function createCachedCollection<T extends Realm.Object>({
       // If the key is numeric, check if we have a cached object for this key
       const index = Number(key);
       const object = target[index];
+
+      // If there is no object at this index, return undefined
+      if (!object) {
```
kraenhansen, line 100: Could this be falsy for another reason? I was thinking if this might (eventually) be used for lists of primitive values, where the empty string, number 0 or `false`.

```suggestion
      if (typeof object === "undefined") {
```

takameyer: This cache isn't used for lists of primitives, but it never hurts to be explicit.

takameyer: I am a bit worried about it being "null" for some reason.  In that case `typeof object` would return `"object"`

takameyer: I've updated it to check for both

kraenhansen: The only reason I can think of where it would be `null` would be for elements in a nullable list of objects.
There is a difference between an element being `null` and an out of bounds access (which should be `undefined`), in which case you should ensure not to return `undefined` when the element was `null`. I've updated my suggestion accordingly.

takameyer: And I have updated the code with your suggestion and added comments

kraenhansen:
I see you accidentally stages changes to the `vendor/realm-core` (just like I did on #4411)

takameyer:
Should be good now

packages/realm-react/src/cachedCollection.ts
```diff
@@ -100,6 +95,17 @@ export function createCachedCollection<T extends Realm.Object>({
       // If the key is numeric, check if we have a cached object for this key
       const index = Number(key);
       const object = target[index];
+
+      // If the collection is modeled in a way that objects can be null
+      // then we should return null instead of undefined to stay symantically
```
kraenhansen, line 100:
```suggestion
      // then we should return null instead of undefined to stay semantically
```

kraenhansen (approved):

kneth (approved):

PR #4417, opened 2022-03-15
Class models: Enforce `extends Realm.Object`

### LINKED ISSUE #4412: Realm#constructor should throw if schema class doesn’t extend Realm.Object.

kraenhansen:
## What, How & Why?

This closes #4412 by checking the prototype of constructors passed into the `Realm` constructor.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/class-models.ts
+++ b/integration-tests/tests/src/tests/class-models.ts
@@ -1,0 +1,75 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2022 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { expect } from "chai";
+
+import Realm from "realm";
+
+describe("Class models", () => {
+  describe("as schema element", () => {
+    beforeEach(() => {
+      Realm.clearTestState();
+    });
+
+    it("fails without a schema static", () => {
+      class Person extends Realm.Object {}
+      expect(() => {
+        new Realm({ schema: [Person as any] });
+      }).throws("must have a 'schema' property");
+    });
+
+    it("fails without a schema.properties static", () => {
+      class Person extends Realm.Object {
+        static schema = { name: "Person" };
+      }
+      expect(() => {
+        new Realm({ schema: [Person as any] });
+      }).throws("properties must be of type 'object'");
+    });
+
+    it("fails if it doesn't extend Realm.Object", () => {
+      class Person {
+        name!: string;
+        static schema: Realm.ObjectSchema = {
+          name: "Person",
+          properties: { name: "string" },
+        };
+      }
+      expect(() => {
+        new Realm({ schema: [Person as any] });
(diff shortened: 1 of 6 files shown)


integration-tests/tests/src/tests/class-models.ts
```diff
@@ -0,0 +1,75 @@
...
+      }).throws("Class 'Person' must extend Realm.Object");
+
+      // Mutate the name of the object schema to produce a more detailed error
+      Person.schema.name = "Foo";
+      expect(() => {
+        new Realm({ schema: [Person as any] });
+      }).throws("Class 'Person' (declaring 'Foo' schema) must extend Realm.Object");
+    });
+
+    it("is allowed", () => {
```
takameyer, line 64: It's probably covered elsewhere in the tests, but the only case I could think of adding would be plain object schemas, just to ensure the changes made left that flow alone.

integration-tests/tests/src/tests/class-models.ts
```diff
@@ -0,0 +1,75 @@
...
+  describe("as schema element", () => {
+    beforeEach(() => {
+      Realm.clearTestState();
+    });
+
+    it("fails without a schema static", () => {
+      class Person extends Realm.Object {}
+      expect(() => {
+        new Realm({ schema: [Person as any] });
+      }).throws("must have a 'schema' property");
```
kneth, line 33: Is it possible to get the class name in the error message?

integration-tests/tests/src/tests/class-models.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    it("fails without a schema.properties static", () => {
+      class Person extends Realm.Object {
+        static schema = { name: "Person" };
+      }
+      expect(() => {
+        new Realm({ schema: [Person as any] });
+      }).throws("properties must be of type 'object'");
```
kneth, line 42: Is it possible to get the class name in the error message?

kneth (approved):
Only minor comments - not blocking in any way

integration-tests/tests/src/tests/class-models.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    it("is allowed", () => {
+      class Person extends Realm.Object {
+        name!: string;
+        static schema: Realm.ObjectSchema = {
+          name: "Person",
+          properties: { name: "string" },
+        };
+      }
+      new Realm({ schema: [Person] });
```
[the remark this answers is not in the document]
kraenhansen: This doesn't really add anything though? If `new Realm()` throws the tests fails, I see little need to expect / assert that?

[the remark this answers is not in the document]
takameyer: If there's no assertion, do we need a test?

integration-tests/tests/src/tests/class-models.ts
kraenhansen: Yeah. The tests for plain objects remain in the legacy test suite, I'm not moving these over in this PR, but they do pass.

integration-tests/tests/src/tests/class-models.ts
kraenhansen: Sure, but that code is not related to the changes I'm making in this PR.
I thought about leaving out the test, but put it in for completeness.

integration-tests/tests/src/tests/class-models.ts
kraenhansen: Sure, but that code is not really related to the changes I'm making in this PR.
I thought about leaving out the test, but put it in for completeness.

integration-tests/tests/src/tests/class-models.ts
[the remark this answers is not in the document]
takameyer: Ok...I did some reading and found this to be valid (challenging my assumptions :) ).  It's called as smoke test if it has no assertion.  If it's broken (smoking) then it failed the test 😆

takameyer (approved):

fronck (approved):

PR #4418, opened 2022-03-15
Add Realm.Configuration.dataInitialization

kneth:
## What, How & Why?

A simple mechanism to populate a Realm when opened the first time.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added `Realm.Configuration.initialDataTransaction` to populate a new Realm with initial data.
+* Added `Realm.Configuration.onFirstOpen` to populate a new Realm with initial data.
 
 ### Fixed
 * Fixed issue that could cause mangling of binary data on a roundtrip to/from the database ([#4278](https://github.com/realm/realm-js/issues/4278), since v10.1.4).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added `Realm.Configuration.dataInitialization` to populate a Realm initially.
+* Added `Realm.Configuration.dataInitialization` to populate a new Realm with initial data.
 
 ### Fixed
 * Fixed issue that could cause mangling of binary data on a roundtrip to/from the database ([#4278](https://github.com/realm/realm-js/issues/4278), since v10.1.4).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added `Realm.Configuration.dataInitialization` to populate a Realm initially.
 
 ### Fixed
 * Fixed issue that could cause mangling of binary data on a roundtrip to/from the database ([#4278](https://github.com/realm/realm-js/issues/4278), since v10.1.4).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added `Realm.Configuration.dataInitialization` to populate a new Realm with initial data.
+* Added `Realm.Configuration.initialDataTransaction` to populate a new Realm with initial data.
 
 ### Fixed
 * Fixed issue that could cause mangling of binary data on a roundtrip to/from the database ([#4278](https://github.com/realm/realm-js/issues/4278), since v10.1.4).
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -378,6 +378,10 @@ class Realm {
  *     - `usedSize` - The total bytes used by data in the file.
  *   It returns `true` to indicate that an attempt to compact the file should be made. The compaction
  *   will be skipped if another process is accessing it.
+ * @property {callback(realm)} [dataInitialization] - The function called when opening a Realm for
+ *   the first time. The function can populate the Realm prior to open it. When calling the callback,
+ *   the Realm will be in a write transaction. The callback takes one argument:
+ *     - `realm` - The Realm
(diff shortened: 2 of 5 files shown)


src/js_realm.hpp
```diff
@@ -689,6 +689,33 @@ bool RealmClass<T>::get_realm_config(ContextType ctx, size_t argc, const ValueTy
...
+
+                    auto realm_ptr = new SharedRealm(realm);
+                    ValueType arguments[] = {
+                        create_object<T, RealmClass<T>>(ctx, realm_ptr),
+                    };
+                    try {
+                        Function<T>::call(ctx, data_initialization_function, 1, arguments);
+                    }
+                    catch (...) {
+                        realm_ptr->reset();
```
kraenhansen, line 711: I don't like that an exception from user code could be hidden here.

src/js_realm.hpp
```diff
@@ -689,6 +689,33 @@ bool RealmClass<T>::get_realm_config(ContextType ctx, size_t argc, const ValueTy
                 config.should_compact_on_launch_function = std::move(should_compact_on_launch_functor);
             }
 
+            static const String data_initalization_string = "dataInitialization";
```
kraenhansen, line 692: Is this name prior art? (ex. from the Java SDK?)

src/js_realm.hpp
```diff
@@ -689,6 +689,33 @@ bool RealmClass<T>::get_realm_config(ContextType ctx, size_t argc, const ValueTy
...
+                config.initialization_function = [=](SharedRealm realm) {
+
+                    auto realm_ptr = new SharedRealm(realm);
+                    ValueType arguments[] = {
+                        create_object<T, RealmClass<T>>(ctx, realm_ptr),
+                    };
+                    try {
+                        Function<T>::call(ctx, data_initialization_function, 1, arguments);
+                    }
+                    catch (...) {
```
kraenhansen, line 710: Could this be more DRY with a finally block?

src/js_realm.hpp
```diff
@@ -689,6 +689,33 @@ bool RealmClass<T>::get_realm_config(ContextType ctx, size_t argc, const ValueTy
...
+            if (!Value::is_undefined(ctx, data_initialization_value)) {
+                if (config.schema_mode == SchemaMode::Immutable) {
+                    throw std::invalid_argument("Cannot set 'dataInitalization' when 'readOnly' is set.");
+                }
+
+                FunctionType data_initialization_function =
+                    Value::validated_to_function(ctx, data_initialization_value);
+                config.initialization_function = [=](SharedRealm realm) {
+
+                    auto realm_ptr = new SharedRealm(realm);
```
kraenhansen, line 703: Why allocate a new shared Realm on the heap when one is passed as argument to the lambda?

docs/realm.js
```diff
@@ -378,6 +378,10 @@ class Realm {
  *     - `usedSize` - The total bytes used by data in the file.
  *   It returns `true` to indicate that an attempt to compact the file should be made. The compaction
  *   will be skipped if another process is accessing it.
+ * @property {callback(realm)} [dataInitialization] - The function called when opening a Realm for
+ *   the first time. The function can populate the Realm prior to open it. When calling the callback,
```
takameyer, line 382:
```suggestion
 *   the first time. The function can populate the Realm prior to opening it. When calling the callback,
```

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added `Realm.Configuration.dataInitialization` to populate a Realm initially.
```
takameyer, line 4:
```suggestion
* Added `Realm.Configuration.dataInitialization` to populate a new Realm with initial data.
```

takameyer (approved):
Great quick win here :)

src/js_realm.hpp
kneth: I am calling `reset()` to release the ownership.

src/js_realm.hpp
kneth: Java calls it [`initialDataTransaction`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java#L102) which actually describes it well.

src/js_realm.hpp
kraenhansen: But won't realm_ptr be a pointer to a heap allocated `std::shared_ptr<Realm>`? I.e. do you actually free the memory allocated for the smartpointer itself by calling `reset` on it?

Anyways - why can't you use the `realm` passed as argument to the lambda directly?

fronck: I agree, it seems like `realm_ptr` would stick around and not be de-allocated.
And, because `realm` is passed as copy-by-value, you should have a fresh `shared_ptr` ready to go there.

kneth: @fronck @kraenhansen I have slightly modified the code, and I believe I have addressed your concerns.

fronck (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added `Realm.Configuration.dataInitialization` to populate a new Realm with initial data.
```
kraenhansen, line 4: This needs to be updated with the final name of the property.

src/js_realm.hpp
kraenhansen: I can't say I love either of those. I see you use `initializer` in your test. I actually like that better 🤔

src/js_realm.hpp
kraenhansen: Okay, I tried passing `&realm` but get a double free. Since `create_object` calls `set_internal` it does assume ownership of the pointer and therefore needs a brand new internal object. I can't say I love an API that implicitly assumes ownership like that.

kraenhansen (approved):
I just have a final remark, nothing blocking.

src/js_realm.hpp
takameyer: If we are discussing naming, I have a few other suggestions to throw in the ring:
 `onRealmInit`, `onFirstOpen`, `initData`

PR #4427, opened 2022-03-17
Class models: `Realm.Object#constructor`

### LINKED ISSUE #4415: Adapt type declarations to class-based models extending Realm.Object

kraenhansen:
## What, How & Why?

This closes #4415, #4414, #4413 by implementing a constructor for `Realm.Object` which calls `Realm#create` to create objects in the database. To enabled this, class-based model objects read from the database will not have their constructors called, but instead we construct plain objects and set their prototype after creation.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


src/js_realm_object.hpp
```diff
@@ -162,6 +164,29 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
...
+    Arguments create_arguments{ctx, create_args.size(), create_args.data()};
+    ReturnValue result{ctx};
+    RealmClass<T>::create(ctx, realm, create_arguments, result);
+    ObjectType realm_object_object = Value::validated_to_object(ctx, result);
+
+    // Copy the internal from the constructed object onto this_object
+    auto realm_object = get_internal<T, RealmObjectClass<T>>(ctx, realm_object_object);
+    // The finalizer on the ObjectWrap (applied inside of set_internal) will delete the `new_realm_object`
+    auto new_realm_object = new realm::js::RealmObject<T>(*realm_object);
+    set_internal<T, RealmObjectClass<T>>(ctx, this_object, new_realm_object);
```
kraenhansen, line 187: Since calling `set_internal` assumes taking ownership of the object behind the `internal` argument, we need to do this dance to allocate a separate `RealmObject` to avoid a double free.

src/node/node_class.hpp
```diff
@@ -1439,6 +1411,7 @@ Napi::Value ObjectWrap<ClassType>::constructor_callback(const Napi::CallbackInfo
         return scope.Escape(env.Null()); // return a value to comply with Napi::FunctionCallback
     }
     else {
+        // XXX: Remove this now that we declare a constructor for the RealmObjectClass
```
kraenhansen, line 1414: I need to remember to try removing this 🙂

tests/js/realm-tests.js
```diff
@@ -1140,20 +1140,18 @@ module.exports = {
...
     });
 
-    TestCase.assertThrowsContaining(() => {
-      realm.write(() => {
-        realm.create("InvalidObject", { intCol: 1 });
-      });
-    }, "Realm object constructor must not return another value");
+    realm.write(() => {
+      realm.create("InvalidObject", { intCol: 1 });
+    });
```
kraenhansen, line 1154: This is no longer an issue since the `InvalidObject` function is no longer called on creation.

tests/spec/helpers/reporters.js
```diff
@@ -31,5 +31,8 @@ jasmine.getEnv().addReporter(
     spec: {
       displayPending: true,
     },
+    summary: {
+      displayStacktrace: true,
```
kraenhansen, line 35: I've added this to improve debuggability of failing tests in the legacy test suite.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 
 ### Breaking change
-* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object`.
+* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulled an object of that type is read from the database.
```
```suggestion
* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulling an object of that type is read from the database.
```

kraenhansen:
> Do the changes to `ObjectWrap<ClassType>::create_instance_by_schema()` have any impact on the spread operator?

None intentional ☺️ The changes on the open PR will most likely have a couple of conflicts with this PR depending on which is merged first.

src/node/node_class.hpp
```diff
@@ -1307,52 +1314,18 @@ Napi::Object ObjectWrap<ClassType>::create_instance_by_schema(Napi::Env env, Nap
...
         // define the properties on the prototype of the schema object constructor
         if (properties.size() > 0) {
             constructorPrototype.DefineProperties(properties);
         }
 
-        instance = schemaObjectConstructor.New({});
-        if (!instance.InstanceOf(schemaObjectConstructor)) {
-            throw Napi::Error::New(env, "Realm object constructor must not return another value");
-        }
+        instance = ObjectCreate.Call({constructorPrototype}).As<Napi::Object>();
```
kraenhansen, line 1327: This change is important. We're creating plain object and setting the prototype instead of calling the constructor provided by the user.

packages/realm-react/src/useQuery.tsx
```diff
@@ -47,7 +47,9 @@ export function createUseQuery(useRealm: () => Realm) {
    * @param type - The object type, depicted by a string or a class extending Realm.Object
    * @returns a collection of realm objects or an empty array
    */
-  return function useQuery<T>(type: string | ({ new (): T } & Realm.ObjectClass)): Realm.Results<T & Realm.Object> {
+  return function useQuery<T>(
+    type: string | ({ new (...args: any): T } & Realm.ObjectClass),
```
takameyer, line 51: I was hoping with class models that we could simplify these types to just be `Realm.ObjectClass<T>`.  Is that possible?

src/js_realm_object.hpp
```diff
@@ -162,6 +164,29 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
...
+    auto constructor = Object::validated_get_object(ctx, this_object, "constructor");
+    auto realm = Value::validated_to_object(ctx, args[0], "realm");
+    auto values = Value::is_undefined(ctx, args[1]) ? Object::create_empty(ctx) : args[1];
+
+    // Create an object
+    std::vector<ValueType> create_args{constructor, values};
+    Arguments create_arguments{ctx, create_args.size(), create_args.data()};
+    ReturnValue result{ctx};
+    RealmClass<T>::create(ctx, realm, create_arguments, result);
+    ObjectType realm_object_object = Value::validated_to_object(ctx, result);
```
takameyer, line 181: I love this variable name :D So this is a c++ Realm object transformed to a JS object?

src/js_types.hpp
```diff
@@ -461,7 +461,7 @@ struct Object {
     static typename ClassType::Internal* get_internal(ContextType ctx, const ObjectType&);
 
     template <typename ClassType>
-    static void set_internal(ContextType ctx, const ObjectType&, typename ClassType::Internal*);
+    static void set_internal(ContextType ctx, ObjectType&, typename ClassType::Internal*);
```
takameyer, line 464: What did removing `const` achieve?

tests/js/list-tests.js
```diff
@@ -1840,58 +1840,42 @@ module.exports = {
 
   testClassObjectCreation: function () {
     class TodoItem extends Realm.Object {
-      constructor(description) {
-        super();
-        this.id = new ObjectId();
```
takameyer, line 1845: Why remove `id`?  What is the equivalent way to do this now?

takameyer (commented):
I would need a bit of guidance in the c++ code.  Before I approve, I would like to know how we are handling computed values in the class.  Maybe that's coming in a future phase?

src/js_realm_object.hpp
kraenhansen: Yep. This is a JS object wrapping a C++ Realm object.
To be honest, this name is be cause I have another `realm_object` below 🙂 

I could rename this to `js_realm_object`?

src/js_types.hpp
kraenhansen: This makes the code able to mutate the "target" object when setting the internal, something we didn't have to do from "outside" of the engine specific code before. I need when setting the internal on JSC, but as this type is shared across all engines I need to update it here as well.

tests/js/list-tests.js
kraenhansen: Since `id` was a primary key I got errors (duplicated primary key) when creating a new object from a managed object here: https://github.com/realm/realm-js/blob/kh/class-models-object-constructor/tests/js/list-tests.js#L1878

I can add it back in but remove the primary key?

src/js_realm_object.hpp
takameyer: That would be way less funny 😆  But more clear

tests/js/list-tests.js
takameyer: Ok thats fine, then there's really no need for it.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 
 ### Breaking change
-* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object`.
+* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulling an object of that type is read from the database.
```
tomduncalf, line 5:
```suggestion
* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulling an object of that type from the database.
```

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 
 ### Breaking change
-* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object`.
+* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulling an object of that type from the database.
```
tomduncalf, line 5: Do you think there'd be any value in adding a small example of how to migrate a class to the new style?

src/js_realm_object.hpp
tomduncalf: I think `js_realm_object` is clearer also

src/js_realm_object.hpp
```diff
@@ -162,6 +164,29 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
     }
 }
 
+template <typename T>
```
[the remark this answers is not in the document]
fronck: Agreed.  Doxygen documentation for new functions is also something we've agreed to do.

src/jsc/jsc_class.hpp
```diff
@@ -624,16 +622,25 @@ typename ClassType::Internal* ObjectWrap<ClassType>::get_internal(JSContextRef c
 }
 
 template <typename ClassType>
-void ObjectWrap<ClassType>::set_internal_property(JSContextRef ctx, JSObjectRef& instance,
-                                                  typename ClassType::Internal* internal)
+void ObjectWrap<ClassType>::set_internal(JSContextRef ctx, JSObjectRef& instance,
```
fronck, line 625: What's the reason for renaming this function?
Also, given that it's been significantly reworked, a Doxygen header would be appreciated.

fronck (commented):
I'd like to see some documentation added to the functions that have been added/significantly changed.

integration-tests/tests/src/tests/class-models.ts
```diff
@@ -72,4 +73,45 @@ describe("Class models", () => {
       new Realm({ schema: [Person] });
     });
   });
+
+  describe("#constructor", () => {
```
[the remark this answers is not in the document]
kraenhansen: This made me realise that the property initialisers fires "too late": I.e. after the object gets created into the database.
We agreed in a meeting that we need to document this behaviour and I've started on a scope to rectify this, making it less likely for users to hit this.

packages/realm-react/src/useQuery.tsx
kraenhansen: I tried but failed to update it to anything simpler. It has to match this signature: https://github.com/realm/realm-js/blob/master/types/index.d.ts#L1093

src/js_realm_object.hpp
kraenhansen: I renamed to `js_realm_object`.

kraenhansen: Since this is thrown away and only used to read its "internal", perhaps I should call it `temporary_realm_object` or `throwaway_realm_object` instead?

takameyer: `tmp_realm_object`

src/js_realm_object.hpp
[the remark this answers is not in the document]
kraenhansen: I've added some documentation.

src/jsc/jsc_class.hpp
kraenhansen: I've added some documentation.

CHANGELOG.md
kraenhansen: I've added an example of usage to the changelog.

src/jsc/jsc_class.hpp
kraenhansen: The reason I've renamed it is because now I need it exposed externally (it was only used internally to the class before). The matching getter was named `get_internal` and I though the "property" part wasn't adding much clarity anyway.

CHANGELOG.md
```diff
@@ -2,10 +2,31 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 
 ### Breaking change
-* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object`.
-
-### Enhancements
-* None.
+* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulling an object of that type from the database. Existing classes already extending `Realm.Object` now needs to call the `super` constructor passing two arguments:
```
```suggestion
* Model classes passed as schema to the `Realm` constructor must now extend `Realm.Object` and will no longer have their constructors called when pulling an object of that type from the database. Existing classes already extending `Realm.Object` now need to call the `super` constructor passing two arguments:
```

CHANGELOG.md
```diff
@@ -2,10 +2,31 @@ x.x.x Release notes (yyyy-MM-dd)
...
+  - `realm`: The realm to create the object in.
```
kneth, line 6:
```suggestion
  - `realm`: The Realm to create the object in.
```

src/js_realm_object.hpp
```diff
@@ -162,6 +164,42 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
     }
 }
 
+/**
+ * @brief Implements the constructor for a Realm.Object, calling the `Realm#create` instance method to create an
+ * object in the database.
+ *
+ * @note This differes from `RealmObjectClass<T>::create_instance` as it is executed when end-users construct a `new
```
kneth, line 171:
```suggestion
 * @note This differs from `RealmObjectClass<T>::create_instance` as it is executed when end-users construct a `new
```

src/js_realm_object.hpp
```diff
@@ -162,6 +164,42 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
...
+ * Realm.Object()` (or another user-defined class extending `Realm.Object`), whereas `create_instance` is called when
+ * reading objects from the database.
+ *
+ * @tparam T Engine specific types.
```
kneth, line 175: `@param` and not `@tparam`?

src/js_realm_object.hpp
```diff
@@ -162,6 +164,42 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
...
+ * @param ctx JS context
```
kneth, line 176:
```suggestion
 * @param ctx JS context.
```

src/jsc/jsc_class.hpp
```diff
@@ -623,17 +621,37 @@ typename ClassType::Internal* ObjectWrap<ClassType>::get_internal(JSContextRef c
     return realmObjectInstance->m_object.get();
 }
 
+/**
+ * @brief Stores data on the `instance` JS object, making it possible to retrieve the internal object insteance at a
+ * later point, using `ObjectWrap<ClassType>::get_internal`.
+ *
+ * @note This hands over ownership of the object pointed to by the `internal` pointer.
+ *
+ * @tparam ClassType The class implementing the JS interface (from C++).
```
kneth, line 630: `@param` instead of `@tparam`?

src/jsc/jsc_class.hpp
```diff
@@ -623,17 +621,37 @@ typename ClassType::Internal* ObjectWrap<ClassType>::get_internal(JSContextRef c
     return realmObjectInstance->m_object.get();
 }
 
+/**
+ * @brief Stores data on the `instance` JS object, making it possible to retrieve the internal object insteance at a
```
kneth, line 625:
```suggestion
 * @brief Stores data on the `instance` JS object, making it possible to retrieve the internal object instance at a
```

kneth (approved):
Mostly fixing typos in comments 😄

src/js_realm_object.hpp
kraenhansen: This was generated by my editor: It's used to describe [a template parameter](https://www.doxygen.nl/manual/commands.html#cmdtparam).

src/js_realm_object.hpp
```diff
@@ -162,6 +164,42 @@ typename T::Object RealmObjectClass<T>::create_instance(ContextType ctx, realm::
...
+ * @tparam T Engine specific types.
+ * @param ctx JS context.
+ * @param this_object JS object being returned to the user once constructed.
+ * @param args Arguments passed by the user when calling the constructor.
+ */
+template <typename T>
+void RealmObjectClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments& args)
+{
+    // Parse aguments
+    args.validate_between(1, 2);
```
kraenhansen, line 184: Expect two required arguments here.

fronck (approved):
:+1:

src/js_realm_object.hpp
kraenhansen: I've pushed code to do this.

PR #4432, opened 2022-03-20
Additional test of Realm.objectForPrimaryKey()

### LINKED ISSUE #4431: ObjectForPrimaryKey passed into CurrentUser crashes
[BUG] If you are using objectForPrimaryKey and you pass in the currentUser instead of currentUser.id as the primary key, it will crash

kneth:
## What, How & Why?

This closes #4431

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


takameyer (approved):
I also tried to reproduce this, and got the same error, which looks fine.
There must have been something else that crashed the application.

Issue #4375, opened 2022-02-25
useObject inside FlatList throws Exception in HostFunction: Cannot create asynchronous query while in a write transaction


mfbx9da4:
### How frequently does the bug occur?

All the time

### Description

Calling `useObject` inside a flat list item component can throw the Exception "Cannot create asynchronous query while in a write transaction" if a `realm.write()` occurs around the same time. 

My guess cause is that the FlatList is doing some multithreading work to do the layouting of the item. As you will see in my example, the offending writes do not call addListener inside the transaction.

### Stacktrace & log output

```shell
Error: Exception in HostFunction: Cannot create asynchronous query while in a write transaction

This error is located at:
    in Message (at App.tsx:121)
    in RCTView (at View.js:32)
    in View (at VirtualizedList.js:2073)
    in VirtualizedListCellContextProvider (at VirtualizedList.js:2088)
    in CellRenderer (at VirtualizedList.js:814)
    in RCTScrollContentView (at ScrollView.js:1674)
    in RCTScrollView (at ScrollView.js:1792)
    in ScrollView (at ScrollView.js:1818)
    in ScrollView (at VirtualizedList.js:1268)
    in VirtualizedListContextProvider (at VirtualizedList.js:1100)
    in VirtualizedList (at FlatList.js:645)
    in FlatList (at App.tsx:118)
    in RCTSafeAreaView (at SafeAreaView.js:51)
    in SafeAreaView (at App.tsx:116)
    in FooList (at App.tsx:155)
    in App (at renderApplication.js:50)
    in RCTView (at View.js:32)
    in View (at AppContainer.js:92)
    in RCTView (at View.js:32)
    in View (at AppContainer.js:119)
    in AppContainer (at renderApplication.js:43)
    in Onin(RootComponent) (at renderApplication.js:60)

```

### Can you reproduce the bug?

Yes, always

### Reproduction Steps

```tsx
import React, { useEffect, useState } from 'react'
import { FlatList, SafeAreaView, Text } from 'react-native'
import RNFS from 'react-native-fs'
import Realm from 'realm'

export function App() {
  const r = useWaitForRealm()
  const [initialized, setInitialized] = useState(false)

  useEffect(() => {
    if (!r) return
    const asyncEffect = () => {
      // cleanup the db
      const fooResults = realm.objects<Foo>('foo')
      realm.write(() => {
        for (const x of fooResults) {
          realm.delete(x)
        }
      })
      setInitialized(true)
    }
[...]

PR #4446, opened 2022-03-25
Fix for adding event listeners while in a write transaction

takameyer:
## What, How & Why?
If a component is setup to render a collection that make use of both useQuery and useObject, it is possible to setup data updates in a way to set the event listeners while in a write transaction.  This can occur if the writes are happing on the event loop (async) and multiple transactions occur one after the other with no break.  As soon as the second write transaction is called, it will flush all event listener events.  In the issue, the flushing of the event causes a component with useObject to render, which will apply an event listener on the object.  This is not allowed within a write transaction.  
The fix makes use of `setImmediate`, which will push the event listener apply function onto the event loop.  Since, in this case, the write transaction is already on the event loop, the listener will be applied immediately after the write transaction has finished, but still allow the component to finish rendering.
This was also added to collections, as an object could possibly have a list property, and would end up having the same issue.

This closes # [4375](https://github.com/realm/realm-js/issues/4375)

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -1,3 +1,16 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* Adding event listeners while in a write transaction ([#4375](https://github.com/realm/realm-js/issues/4375))
+### Compatibility
+* None.
+
+### Internal
+* None.
+
 0.2.1 Release notes (2022-03-24)
 =============================================================
 ### Enhancements
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -4,7 +4,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* Adding event listeners while in a write transaction ([#4375](https://github.com/realm/realm-js/issues/4375))
+* Fixed potential "Cannot create asynchronous query while in a write transaction" error with `useObject` due to adding event listeners while in a write transaction ([#4375](https://github.com/realm/realm-js/issues/4375), since v0.1.0)
+
 ### Compatibility
 * None.
 
--- a/packages/realm-react/src/__tests__/useQueryHook.test.tsx
+++ b/packages/realm-react/src/__tests__/useQueryHook.test.tsx
@@ -58,6 +58,8 @@ const useRealm = () => {
 
 const useQuery = createUseQuery(useRealm);
 
+const awaitEventLoop = () => new Promise((resolve) => setTimeout(resolve, 0));
+
 const testDataSet = [
   { _id: 1, name: "Vincent", color: "black and white", gender: "male", age: 4 },
   { _id: 2, name: "River", color: "brown", gender: "female", age: 12 },
@@ -78,11 +80,14 @@ describe("useQueryHook", () => {
     });
     realm.close();
   });
+
   afterEach(() => {
     Realm.clearTestState();
   });
-  it("can retrieve collections using useQuery", () => {
+
+  it("can retrieve collections using useQuery", async () => {
     const { result } = renderHook(() => useQuery<IDog>("dog"));
+    await awaitEventLoop();
     const collection = result.current;
 
     const [dog1, dog2, dog3] = testDataSet;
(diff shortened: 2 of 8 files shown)


packages/realm-react/src/cachedObject.ts
```diff
@@ -104,7 +104,13 @@ export function createCachedObject<T extends Realm.Object>({
     }
   };
 
-  object.addListener(listenerCallback);
+  // Add this on the next tick, in case there is a write transaction occuring immediately after creation of this object
+  // see https://github.com/realm/realm-js/issues/4375
+  setImmediate(() => {
+    if (object.isValid()) {
```
kraenhansen, line 110: Are you sure this was actually needed?

takameyer: It will throw if it's not valid

kneth (approved):
Do we need to document the usage of `setImmediate()`? The origin of my question is the big yellow "note" in https://reactnative.dev/docs/timers.

packages/realm-react/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* Adding event listeners while in a write transaction ([#4375](https://github.com/realm/realm-js/issues/4375))
```
tomduncalf, line 7: I think we should aim for changelogs which are more easy to understand for end users (e.g. what was the consequence of the bug, in addition to the underlying cause)

```suggestion
* Fixed potential "Cannot create asynchronous query while in a write transaction" error with `useObject` due to adding event listeners while in a write transaction ([#4375](https://github.com/realm/realm-js/issues/4375), since v0.1.0)

```

packages/realm-react/src/__tests__/useQueryHook.test.tsx
```diff
@@ -58,6 +58,8 @@ const useRealm = () => {
 
 const useQuery = createUseQuery(useRealm);
 
+const awaitEventLoop = () => new Promise((resolve) => setTimeout(resolve, 0));
```
tomduncalf, line 61: Could we add a comment summarising why we need to call this?

packages/realm-react/src/cachedCollection.ts
```diff
@@ -158,7 +158,10 @@ export function createCachedCollection<T extends Realm.Object>({
   };
 
   if (!isDerived) {
-    cachedCollectionResult.addListener(listenerCallback);
+    // Add this on the next tick, in case there is a write transaction occuring immediately after creation of this collection
+    setImmediate(() => {
```
tomduncalf, line 162: I wonder if [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask) might be appropriate. My understanding is it fires slightly earlier than `setImmediate`/`setTimeout`, not sure if it would be "too early" or not

packages/realm-react/src/__tests__/useQueryRender.test.tsx
```diff
@@ -354,4 +371,34 @@ describe.each`
 
     expect(queryByText("756c")).toBeNull();
   });
+  // This replicates the issue https://github.com/realm/realm-js/issues/4375
+  it("will handle multiple async transactions", async () => {
+    const { queryByTestId } = await setupTest({ queryType, useUseObject: true });
+    const asyncEffect = async () => {
```
tomduncalf, line 377: If this is a well known naming convention then no worries, but I thought this could be named more plainly

```suggestion
    const performTest = async () => {
```

packages/realm-react/src/__tests__/useQueryRender.test.tsx
```diff
@@ -354,4 +371,34 @@ describe.each`
...
+        const id = i;
+        testRealm.write(() => {
+          testRealm.create(Item, { id, name: `${id}` }, Realm.UpdateMode.Modified);
+        });
+        testRealm.write(() => {
+          const item = testRealm.objectForPrimaryKey(Item, id);
+          if (item) {
+            item.name = `${id + 100}`;
+          }
+        });
```
tomduncalf, line 393: So do we only see the issue if we create an item _and_ update it in the same "event loop"? Just curious because the original user's repro was a bit different (created an object, `await sleep(0)`, created another object), if this is the easiest way to repro then all good

takameyer: The `sleep(0)` had the same result, so we could potentially add that back in just to make sure that also is fixed.  It's a bit of a strange concept, since this is already on the event loop.  Curious about how those are tasks get prioritised.  A `sleep(1)` will hide the issue completely.

packages/realm-react/src/cachedCollection.ts
takameyer: Seems to function in both cases. I'll have to review both methods to decide which is better suited for this.

packages/realm-react/src/cachedObject.ts
takameyer: I've added a comment

packages/realm-react/src/__tests__/useQueryRender.test.tsx
takameyer: I've added the `sleep(0)` so that this code mirrors the issue explained by the user.

packages/realm-react/src/cachedCollection.ts
tomduncalf: Ahh OK, let's leave it as is then!

PR #4452, opened 2022-03-29
Fix Timeout Error for Logged out User

### LINKED ISSUE #4453: When TimeOut behaviour is set to "openLocalRealm", error is thrown if user logs out before the timeout

All the time

### Description

When TimeOut behaviour is set to "openLocalRealm", error is thrown if user logs out before the timeout.
If the timeout time is reached before logout, then there is no error.

### Stacktrace & log output

```shell
Error: User is no longer valid.
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:104:6 in reportException
at node_modules/react-native/Libraries/Core/ExceptionsManager.js:172:19 in handleException
at node_modules/react-native/Libraries/Core/setUpErrorHandling.js:24:6 in handleError
at node_modules/expo-error-recovery/build/ErrorRecovery.fx.js:12:21 in ErrorUtils.setGlobalHandler$argument_0
```
```


Yes, always

### Reproduction Steps

Open a realm containing the configuration:


```javascript
sync: {
  //...
    existingRealmFileBehavior: {
    type: "downloadBeforeOpen",
    timeOutBehavior: "openLocalRealm",
    timeOut: 30 * 1000 // arbitrary timeout value
  },
}
```

Perform a login and immediately logout.  After 30 seconds the error will appear.

### Version

10

MongoDB Realm (i.e. Sync, auth, functions)

### Are you using encryption?

No, not using encryption

Node, macos

### Build environment

### Cocoapods version

_No response_

takameyer:
## What, How & Why?

When opening Realm again with the same user and the timeOutBehaviour "openLocalRealm", logging the user out immediately would crash as the timeout event would try and open a local realm with the logged out user session.

This closes #4453

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -257,9 +257,7 @@ module.exports = function (realmConstructor) {
                   return;
                 }
                 // Clear the fallback timeOut if it has been started
-                if (timeoutId) {
-                  clearTimeout(timeoutId);
-                }
+                clearTimeout(timeoutId);
                 if (error) {
                   reject(error);
                 } else {
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -197,6 +197,10 @@ module.exports = function (realmConstructor) {
 
         // First configure any timeOut and corresponding behavior.
         let openPromises = [];
+
+        // Id of the timer triggering the timeout
+        let timeoutId = null;
+
         if (config.sync[behavior] !== undefined && config.sync[behavior].timeOut !== undefined) {
           let timeOut = config.sync[behavior].timeOut;
           if (typeof timeOut !== "number") {
@@ -223,7 +227,7 @@ module.exports = function (realmConstructor) {
 
           openPromises.push(
             new Promise((resolve, reject) => {
-              setTimeout(() => {
+              timeoutId = setTimeout(() => {
                 if (asyncOpenTask) {
                   asyncOpenTask.cancel();
                   asyncOpenTask = null;
@@ -251,6 +255,10 @@ module.exports = function (realmConstructor) {
                 // actually invoke this, so recheck here.
                 if (cancelled) {
                   return;
+                }
+                // Clear the fallback timeOut if it has been started
+                if (timeoutId) {
+                  clearTimeout(timeoutId);
                 }
                 if (error) {
                   reject(error);
--- a/tests/js/open-behavior-tests.js
+++ b/tests/js/open-behavior-tests.js
@@ -357,8 +357,8 @@ module.exports = {
   },
 
   testNewFile_downloadBeforeOpen_openLocalOnTimeOut_existing: async function () {
-    //This is a regression test.  If one were to logout before the openLocalRealm timeout
-    //then the timeout would try to open a local realm with the timed out user
+    // This is a regression test.  If one were to logout before the openLocalRealm timeout
+    // then the timeout would try to open a local realm with the timed out user
 
     // 1. Add data to server Realm from User
     // 2. Open Realm again with User
@@ -383,8 +383,6 @@ module.exports = {
       };
 
       const realm = await Realm.open(config);
-
-      console.log(realm.path);
 
       realm.write(() => {
         realm.create(schemas.DogForSync.name, { _id: new ObjectId(), name: "Lola" });
--- a/tests/js/open-behavior-tests.js
+++ b/tests/js/open-behavior-tests.js
@@ -357,7 +357,9 @@ module.exports = {
   },
 
   testNewFile_downloadBeforeOpen_openLocalOnTimeOut_existing: async function () {
-    // This is a regression test.  If one were to logout before the openLocalRealm timeout
+    // This is a regression test for the following issue:
+    // https://github.com/realm/realm-js/issues/4453
+    // If one were to logout before the openLocalRealm timeout
     // then the timeout would try to open a local realm with the timed out user
 
     // 1. Add data to server Realm from User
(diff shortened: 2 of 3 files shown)


tests/js/open-behavior-tests.js
```diff
@@ -356,6 +356,70 @@ module.exports = {
     }
   },
 
+  testNewFile_downloadBeforeOpen_openLocalOnTimeOut_existing: async function () {
+    //This is a regression test.  If one were to logout before the openLocalRealm timeout
```
kneth, line 360: A like to the issue would be nice

kneth (approved):

lib/extensions.js
```diff
@@ -228,6 +232,11 @@ module.exports = function (realmConstructor) {
                   asyncOpenTask.cancel();
                   asyncOpenTask = null;
                 }
+                // Resolve immediately if the other realm open promise has already finished
+                if (promiseResolved) {
+                  return resolve();
```
kraenhansen, line 237: If anything it should reject, I don't like that it's resolving to `undefined`, if a bug gets introduced in the future this might make `Realm.open` return promises that doesn't resolve to a `Realm`: If anything it should `reject`.

tests/js/open-behavior-tests.js
takameyer: I'll have to create one then.  Or should I link the help ticket directly?

lib/extensions.js
takameyer: I'm happy with rejecting

kraenhansen:
I suggest an alternative fix where you `clearTimeout` of the timer returned here: https://github.com/realm/realm-js/blob/andrew/timeOutBehavior-fix/lib/extensions.js#L230

lib/extensions.js
takameyer: I've updated the code as per your suggestion using `clearTimeout`

lib/extensions.js
```diff
@@ -197,6 +197,10 @@ module.exports = function (realmConstructor) {
 
         // First configure any timeOut and corresponding behavior.
         let openPromises = [];
+
+        // This will be set to true by the first promise to resolve.  The next promise will use this to stop execution
+        let fallbackTimeOutId = null;
```
kraenhansen, line 202:
```suggestion
        // Id of the timer triggering the timeout
        let timeoutId = null;
```

There are no other timeout ids.

takameyer: Good catch.  Fix is in

lib/extensions.js
```diff
@@ -252,6 +256,10 @@ module.exports = function (realmConstructor) {
                 if (cancelled) {
                   return;
                 }
+                // Clear the fallback timeOut if it has been started
+                if (timeoutId) {
+                  clearTimeout(timeoutId);
+                }
```
kraenhansen, line 262: In an ideal scenario I would love this to go into a `finally(() => { /
 ...*/ })` on the [`openPromise`](https://github.com/realm/realm-js/blob/andrew/timeOutBehavior-fix/lib/extensions.js#L274) - but let's save that for another PR 😄

kraenhansen (approved):

lib/extensions.js
kraenhansen, line 262: Calling this with `null` is actually okay: https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout#notes

```suggestion
                clearTimeout(timeoutId);
```

tests/js/open-behavior-tests.js
takameyer: created https://github.com/realm/realm-js/issues/4453

tests/js/open-behavior-tests.js
```diff
@@ -356,6 +356,70 @@ module.exports = {
...
+    //then the timeout would try to open a local realm with the timed out user
```
tomduncalf, line 361:
```suggestion
    // This is a regression test.  If one were to logout before the openLocalRealm timeout
    // then the timeout would try to open a local realm with the timed out user
```

tests/js/open-behavior-tests.js
```diff
@@ -356,6 +356,70 @@ module.exports = {
...
+          user,
+          partitionValue,
+          _sessionStopPolicy: "immediately",
+        },
+      };
+
+      const realm = await Realm.open(config);
+
+      console.log(realm.path);
+
```
tomduncalf, line 388:
```suggestion
```

tomduncalf (approved):

tests/js/open-behavior-tests.js
takameyer: doooh!

lib/extensions.js
takameyer: Sounds good.  Someday we should just rewrite this entirely.  I'm also kind of unhappy about `timeOut` vs `timeout`.  Maybe a breaking change ticket is in order

tests/js/open-behavior-tests.js
```diff
@@ -356,6 +356,68 @@ module.exports = {
...
+          partitionValue,
+          _sessionStopPolicy: "immediately",
+          existingRealmFileBehavior: {
+            type: "downloadBeforeOpen",
+            timeOut: 1000,
+            timeOutBehavior: "openLocalRealm",
+          },
+        },
+      };
+      await Realm.open(config);
```
kraenhansen, line 412: As you're awaiting the promise returned by realm.open before you log out, why would this trigger the original bug?
Did this throw before you did the fix?

PR #4470, opened 2022-04-04
@realm/react AppProvider and useApp

takameyer:
## What, How & Why?
Create a provider component that opens a Realm.App with a given
configuration.
Create a hook to access the app component.

This closes # ???

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/example/app/AppSync.tsx
+++ b/example/app/AppSync.tsx
@@ -1,10 +1,10 @@
 import React, {useEffect, useMemo} from 'react';
+import {useApp} from '@realm/react';
+import {StyleSheet, Text} from 'react-native';
 
 import {Task} from './models/Task';
 import {TaskRealmContext} from './models';
 import {TaskManager} from './components/TaskManager';
-import {useApp} from '@realm/react';
-import {StyleSheet, Text} from 'react-native';
 
 const {useRealm, useQuery} = TaskRealmContext;
 
--- a/example/app/AppSync.tsx
+++ b/example/app/AppSync.tsx
@@ -3,6 +3,8 @@ import React, {useEffect, useMemo} from 'react';
 import {Task} from './models/Task';
 import {TaskRealmContext} from './models';
 import {TaskManager} from './components/TaskManager';
+import {useApp} from '@realm/react';
+import {StyleSheet, Text} from 'react-native';
 
 const {useRealm, useQuery} = TaskRealmContext;
 
@@ -12,6 +14,7 @@ type AppProps = {
 
 export const AppSync: React.FC<AppProps> = ({userId}) => {
   const realm = useRealm();
+  const app = useApp();
   const result = useQuery(Task);
 
   const tasks = useMemo(() => result.sorted('createdAt'), [result]);
@@ -22,5 +25,17 @@ export const AppSync: React.FC<AppProps> = ({userId}) => {
     });
   }, [realm, result]);
 
-  return <TaskManager tasks={tasks} userId={userId} />;
+  return (
+    <>
+      <Text style={styles.idText}>Syncing with app id: {app.id}</Text>
+      <TaskManager tasks={tasks} userId={userId} />
+    </>
+  );
 };
+
+const styles = StyleSheet.create({
+  idText: {
+    color: '#999',
+    paddingHorizontal: 20,
+  },
+});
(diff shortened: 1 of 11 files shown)


example/sync.config.js
```diff
@@ -1,6 +1,6 @@
 export const SYNC_CONFIG = {
   // Set `enabled` to `true` to enable sync.
-  enabled: false,
+  enabled: true,
   // Add your Realm App ID here if sync is enabled.
-  appId: '<Your App ID>',
+  appId: 'realmtemplate-waxdf',
```
tomduncalf, line 5: 🙃

takameyer: Everytime 😅

takameyer: git update-index --assume-unchanged [<file> ...]

tomduncalf: Perhaps it is worth modifying this to use an environment variable (and then e.g. have a `.env` file with your app id in which can be `gitignore`d)?

takameyer: Done.  It's probably best practice to inject your app id through an env var anyway.

packages/realm-react/src/__tests__/AppProvider.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+describe("AppProvider", () => {
+  it("returns the configured app with useApp", async () => {
+    const wrapper = ({ children }: { children: React.ReactNode }) => (
+      <AppProvider id="someId" app={{ name: "someName", version: "42" }} baseUrl="someurl">
+        {children}
+      </AppProvider>
+    );
+    const { result } = renderHook(() => useApp(), { wrapper });
+    const app = result.current;
+    expect(app.id).toBe("someId");
```
tomduncalf, line 33: Should we also check the app `name`, `version` and `baseUrl` (assuming there's a way to access them)?

packages/realm-react/src/__tests__/AppProvider.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    await act(async () => {
+      fireEvent.press(changeSchemaButton);
+      await new Promise((resolve) => setTimeout(resolve, 0));
+    });
+
+    // Changing the realm provider configuration will cause a comlete new remount
+    // of the child component.  Therefore it must be retreived again
+    const newSchemaNameContainer = getByTestId("appId");
+
+    expect(newSchemaNameContainer).toHaveTextContent("newId");
```
tomduncalf, line 68: Could we not just check `app.id` rather than rendering it to a component?

example/app/AppSync.tsx
```diff
@@ -3,6 +3,8 @@ import React, {useEffect, useMemo} from 'react';
 import {Task} from './models/Task';
 import {TaskRealmContext} from './models';
 import {TaskManager} from './components/TaskManager';
+import {useApp} from '@realm/react';
+import {StyleSheet, Text} from 'react-native';
```
tomduncalf, line 7: I think these should be grouped with the `react` import on line 1

example/sync.config.js
```diff
@@ -1,6 +1,6 @@
 export const SYNC_CONFIG = {
   // Set `enabled` to `true` to enable sync.
-  enabled: false,
+  enabled: Boolean(process.env.TEMPLATE_SYNC_ENABLED) || false,
```
tomduncalf, line 3: Maybe we can be explicit about what value we expect? I don't mind too much though!
```suggestion
  enabled: process.env.TEMPLATE_SYNC_ENABLED === '1' || false,
```

tomduncalf (changes requested):
Looks good but just some small things

packages/realm-react/src/__tests__/AppProvider.test.tsx
takameyer: Oddly enough, they are not accessible

tomduncalf: Ah cool, not to worry

packages/realm-react/src/__tests__/AppProvider.test.tsx
takameyer: Not so trivial

kneth (approved):

PR #4498, opened 2022-04-19
Update package.json to include a `unpkg` field

kraenhansen:
## What, How & Why?

This adds a `unpkg` field to the package.json to simplify the URL users need to paste when consuming the IIFE bundle using unpkg.com.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-web/scripts/postversion.ts
+++ b/packages/realm-web/scripts/postversion.ts
@@ -30,7 +30,7 @@ const changelogPath = path.resolve(__dirname, "../CHANGELOG.md");
 
 // Update the readme file to use the new version in the script-tag.
 const readmeContent = fs.readFileSync(readmePath, "utf8");
-const readmeContentReplaced = readmeContent.replace(/realm-web@[^/]+/, `realm-web@${version}`);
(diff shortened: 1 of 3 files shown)


packages/realm-web/scripts/postversion.ts
```diff
@@ -30,7 +30,7 @@ const changelogPath = path.resolve(__dirname, "../CHANGELOG.md");
 
 // Update the readme file to use the new version in the script-tag.
 const readmeContent = fs.readFileSync(readmePath, "utf8");
-const readmeContentReplaced = readmeContent.replace(/realm-web@[^/]+/, `realm-web@${version}`);
+const readmeContentReplaced = readmeContent.replace(/realm-web@[^"/]+/, `realm-web@${version}`);
```
tomduncalf, line 33: Is this change deliberate?

tomduncalf (approved):

kneth (approved):
LGTM

packages/realm-web/scripts/postversion.ts
kraenhansen: Yeah. Now that the version is the last part of the URL it needed a " in the pattern.

PR #4509, opened 2022-04-20
Close sync sessions on hot reload to prevent race condition

### LINKED ISSUE #4506: Hot reload of a flexible sync app can cause a crash
* [BUG] (Andrew) Just for js-sdk.  Hot reload of a sync app can cause a crash: terminating with uncaught exception of type realm::MultipleSyncAgents: Multiple sync agents attempted to join the same session.  This happened when sync was not enabled in the backend and Realm.open was not returning anything (reproduction)

Investigate issue before continuing to determine level of effort

tomduncalf:
## What, How & Why?

This closes #4506.
 
This helps prevent a race condition when hot reloading a React Native app which uses sync which led to a `Multiple sync agents attempted to join the same session` exception, by immediately closing any sync sessions when the hot reload is requested (more details in the corresponding core PR, https://github.com/realm/realm-core/pull/5411).

There is still an issue where if you disable sync for your app and hot reload after the initial load, you hit the exception. This needs more investigation, no reference to the Realm is being kept around in the `RealmCoordinator` in that case – perhaps a core issue, but hopefully an edge case.

## How to test

To reproduce the problem prior to this change:
1. Open a sync app
2. Disable sync
3. Hot reload (press "R" in the React Native packager window)
4. You should hit the exception

After this change, the hot reload should be successful.

Note that this is just an easy way to trigger this condition, it can (and has) been hit in normal usage if the sync client takes longer to finish its current upload than the RN app takes to restart.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * Fixed issue where React Native apps would sometimes show stale Realm data until the user interacted with the app UI ([#4389](https://github.com/realm/realm-js/issues/4389), since v10.0.0)
+* Fixed race condition leading to potential crash when hot reloading an app using Realm Sync ([4509](https://github.com/realm/realm-js/pull/4509), since v10.12.0)
 * Adding an object to a Set, deleting the parent object, and then deleting the previously mentioned object causes crash ([#5387](https://github.com/realm/realm-core/issues/5387), since v10.5.0)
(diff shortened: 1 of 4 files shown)


kneth (approved):
Do we need to port it to https://github.com/realm/realm-js/tree/hermes?

tomduncalf:
Yeah we do I think @kneth, not sure how much of this has changed in Hermes though!

Issue #4518, opened 2022-04-22
Realm react caching doesn't work for non-primitive values


mfbx9da4:
### How frequently does the bug occur?

All the time

### Description

All lines like this `console.log('foo?.date changed unfortunately')` should not be firing on every button click.

```
import React, { useEffect, useState } from 'react'
import { Button, SafeAreaView, Text } from 'react-native'
import RNFS from 'react-native-fs'
import SplashScreen from 'react-native-splash-screen'
import Realm from 'realm'
SplashScreen.hide()

export function App() {
  const r = useWaitForRealm()
  const [initialized, setInitialized] = useState(false)

  useEffect(() => {
    if (!r) return
    const asyncEffect = () => {
      // Cleanup the db
      const fooResults = realm.objects<Foo>('foo')
      realm.write(() => {
        for (const x of fooResults) {
          realm.delete(x)
        }
        realm.create<Foo>(
          'foo',
          {
            id: 'foo',
            date: new Date(),
            set: ['0'] as any,
            dictionary: { foo: 'bar' },
          },
          Realm.UpdateMode.Modified
        )
      })
      setInitialized(true)
    }
    void asyncEffect()
  }, [r])

  if (!initialized) return null

  return <FooList />
}

let i = 0
const sleep = (milliseconds: number) => new Promise(r => setTimeout(r, milliseconds))
function FooList() {
  const fooResults = useQuery<Foo>(() => realm.objects<Foo>('foo'))
  console.log('fooResults', fooResults)
  const [count, setCount] = useState(0)

  const foo = useObject<Foo>('foo', 'foo')

  useEffect(() => {
    console.log('foo?.date changed unfortunately')
  }, [foo?.date])

  useEffect(() => {
    console.log('foo?.set changed unfortunately')
  }, [foo?.set])

  useEffect(() => {
    console.log('foo?.dictionary changed unfortunately')
  }, [foo?.dictionary])

  return (
    <SafeAreaView style={{ margin: 20 }}>
      <Text>{foo?.id}</Text>
      <Button onPress={() => setCount(count + 1)} title={`Count: ${count}`} />
    </SafeAreaView>
  )
}

// #region === Setup the Realm instance (start) ===
// You can skip reading this bit, I've left it here so it can be easily reproduced.
const FooSchema: Realm.ObjectSchema = {
  name: 'foo',
  primaryKey: 'id',
  properties: {
    id: 'string',
    date: 'date?',
[...]

PR #4519, opened 2022-04-22
Fix #4513

### LINKED ISSUE #4513: Incorrect wrapping of proxy objects on every render

All the time

### Description

https://github.com/realm/realm-js/blob/master/packages/realm-react/src/useObject.tsx#L78

```
    // Wrap object in a proxy to update the reference on rerender ( should only rerender when something has changed )
    return new Proxy(object, {});
```

The intention of this code is that the reference will only change if the object has changed to play nicely with reacts immutability conventions. However, the hook is not invoked only when the hook fires. It is invoked anytime the parent component renders. The component could be rendering because a totally unrelated change to some other component state which will cause unnecessary Proxy wrapping, unnecessarily new references and break React's immutability conventions.

Easier to explain with code

```
const initialObject = {};
function useObject() {
  return new Proxy(initialObject, {});
}

export default function App() {
  const [count, setCount] = useState(0);
  const object = useObject();
  useEffect(() => {
    // This effect will be invoked everytime the component
    // renders (eg click the button), even though the object
    // never actually changed
    console.log("object changed", object);
  }, [object]);

  return (
    <div className="App">
      <h1>{count}</h1>
      <button onClick={() => setCount((x) => x + 1)}>Click me</button>
    </div>
  );
}
```

https://codesandbox.io/s/modern-field-7ety98?file=/src/App.js:69-642

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

_No response_

### Version

main

Local Database only

### Are you using encryption?

No, not using encryption

ios

### Build environment

### Cocoapods version

_No response_

mfbx9da4:
## What, How & Why?
Fixes https://github.com/realm/realm-js/issues/4513 and https://github.com/realm/realm-js/issues/4913

### DIFF
--- a/packages/realm-react/src/useObject.tsx
+++ b/packages/realm-react/src/useObject.tsx
@@ -52,7 +52,7 @@ export function createUseObject(useRealm: () => Realm) {
     // the cachedObject can force the component using this hook to re-render when a change occurs.
     const [, forceRerender] = useReducer((x) => x + 1, 0);
 
-    const objectRef = useRef<Realm.Object & T | null>(null);
+    const objectRef = useRef<(Realm.Object & T) | null>(null);
 
     // Wrap the cachedObject in useMemo, so we only replace it with a new instance if `primaryKey` or `type` change
     const { object, tearDown } = useMemo(
--- a/packages/realm-react/src/useObject.tsx
+++ b/packages/realm-react/src/useObject.tsx
@@ -16,7 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 import Realm from "realm";
-import { useEffect, useReducer, useMemo } from "react";
+import { useEffect, useReducer, useMemo, useRef } from "react";
 import { createCachedObject } from "./cachedObject";
 
 // In order to make @realm/react work with older version of realms
@@ -52,15 +52,26 @@ export function createUseObject(useRealm: () => Realm) {
     // the cachedObject can force the component using this hook to re-render when a change occurs.
     const [, forceRerender] = useReducer((x) => x + 1, 0);
 
+    const objectRef = useRef<Realm.Object & T | null>(null);
+
     // Wrap the cachedObject in useMemo, so we only replace it with a new instance if `primaryKey` or `type` change
     const { object, tearDown } = useMemo(
       // TODO: There will be an upcoming breaking change that makes objectForPrimaryKey return null
       // When this is implemented, remove `?? null`
       () =>
-        createCachedObject({
-          object: realm.objectForPrimaryKey(type, primaryKey) ?? null,
-          updateCallback: forceRerender,
-        }),
+        {
+          const updateCallback = () => {
+            // Wrap object in a proxy to update the reference on rerender ( should only rerender when something has changed )
+            objectRef.current = cachedObject.object ? new Proxy(cachedObject.object, {}) : null;
+            forceRerender();
+          }
+          const cachedObject = createCachedObject({
+            object: realm.objectForPrimaryKey(type, primaryKey) ?? null,
+            updateCallback: forceRerender,
+          })
+          objectRef.current = cachedObject.object;
+          return cachedObject
+        },
       [type, realm, primaryKey],
     );
 
@@ -74,7 +85,6 @@ export function createUseObject(useRealm: () => Realm) {
       return null;
     }
 
-    // Wrap object in a proxy to update the reference on rerender ( should only rerender when something has changed )
-    return new Proxy(object, {});
+    return objectRef.current
   };
 }
--- a/packages/realm-react/src/useObject.tsx
+++ b/packages/realm-react/src/useObject.tsx
@@ -85,6 +85,7 @@ export function createUseObject(useRealm: () => Realm) {
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


mfbx9da4:
I didn't spend too long on this so might need some tweaking or cleaning up? @takameyer 

This won't fix https://github.com/realm/realm-js/issues/4518

This also won't fix the fact that:
 
> Using the same object via useObject in two different components will return different references 

(as mentioned in 4513). 

I imagine the code will have changed significantly to solve these latter two issues

takameyer:
@mfbx9da4 Thanks for the PR!

Slightly different take than my idea:
```tsx
return function useObject<T>(type: string | { new (): T }, primaryKey: PrimaryKey): (T & Realm.Object) | null {
    const realm = useRealm();

    // Create a forceRerender function for the cachedObject to use as its updateCallback, so that
    // the cachedObject can force the component using this hook to re-render when a change occurs.
    const [version, forceRerender] = useReducer((x) => x + 1, 0);

    // Wrap the cachedObject in useMemo, so we only replace it with a new instance if `primaryKey` or `type` change
    const { object, tearDown } = useMemo(
      // TODO: There will be an upcoming breaking change that makes objectForPrimaryKey return null
      // When this is implemented, remove `?? null`
      () =>
        createCachedObject({
          object: realm.objectForPrimaryKey(type, primaryKey) ?? null,
          updateCallback: forceRerender,
        }),
      [type, realm, primaryKey],
    );

    // Invoke the tearDown of the cachedObject when useObject is unmounted
    useEffect(() => {
      return tearDown;
    }, [tearDown]);

    return useMemo(() => {
      // If the object has been deleted or doesn't exist for the given primary key, just return null
      if (object === null || object?.isValid() === false) {
        return null;
      } else {
        // Wrap object in a proxy to update the reference on rerender ( should only rerender when something has changed )
        return new Proxy(object, {});
      }
      // We need version as a dependency in order to trigger a new object reference
    }, [version, object]);
  };
```
We will need to write a test case for this and the other issues in order to avoid regressions.

mfbx9da4:
Yeah cool! I like your way of doing it. It would be cleaner IMO if the cached object API was tweaked though

```
function useObject<T>(type: string | { new (): T }, primaryKey: PrimaryKey): (T & Realm.Object) | null {
    const realm = useRealm();

    const cachedObject = useMemo<T & Realm.Object | null>(() => createCachedObject(realm, type, primaryKey), [realm, type, primaryKey]);
    const [version, setVersion] = useState(0)
    

    useEffect(() => {
        return cachedObject?.subscribe(() => setVersion(x => x + 1));
    }, [cachedObject]);

    return useMemo(() => new Proxy(cachedObject.object, {}), [cachedObject, version]);
  };
```

takameyer (changes requested):
Looks like we need to rebase and write a regression test.  If you got some time to do that, we would be very grateful 🙂 
Otherwise it may be some time before I get to it.

takameyer:
Adding @tomduncalf and @kraenhansen to the review as well

packages/realm-react/src/useObject.tsx
```diff
@@ -52,15 +52,26 @@ export function createUseObject(useRealm: () => Realm) {
...
     const { object, tearDown } = useMemo(
       // TODO: There will be an upcoming breaking change that makes objectForPrimaryKey return null
       // When this is implemented, remove `?? null`
       () =>
-        createCachedObject({
-          object: realm.objectForPrimaryKey(type, primaryKey) ?? null,
-          updateCallback: forceRerender,
-        }),
+        {
+          const updateCallback = () => {
```
sdandois, line 63: updateCallback is unused.

packages/realm-react/src/useObject.tsx
```diff
@@ -74,7 +85,6 @@ export function createUseObject(useRealm: () => Realm) {
       return null;
     }
 
-    // Wrap object in a proxy to update the reference on rerender ( should only rerender when something has changed )
-    return new Proxy(object, {});
+    return objectRef.current
```
takameyer, line 88:
```suggestion
    // Return the current object reference, which should only cause a rerender when something has changed 
    return objectRef.current
```

packages/realm-react/src/useQuery.tsx
```diff
@@ -65,8 +77,6 @@ export function createUseQuery(useRealm: () => Realm) {
       return tearDown;
     }, [tearDown]);
 
-    // This makes sure the collection has a different reference on a rerender
-    // Also we are ensuring the type returned is Realm.Results, as this is known in this context
-    return new Proxy(collection as Realm.Results<T & Realm.Object>, {});
+    return collectionRef.current;
```
takameyer, line 80:
```suggestion
    // Return the current collection reference, which should only cause a rerender when something in the visible collection has changed 
    return collectionRef.current;
```

takameyer (changes requested):
The code looks great, but I would like to see a regression test in both `useQueryRender` and `useObjectRender` that verifies that this fix works.  Functionally it should trigger a re-render on the parent component (using `useState`) and verify that the children were not re-rendered.

takameyer:
I've squashed the current changes into a single commit and am continuing this PR on my own here:
https://github.com/realm/realm-js/pull/4958

PR #4538, opened 2022-04-29
[@realm/tools] Set up the package. Initial import of realm-schema.

kneth:
## What, How & Why?

Growing out of our recent hack-week, I want to have a place for small and useful (to our users and us) tools.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests


packages/realm-tools/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "realm-schema": "./dist/realm-schema.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc -p tsconfig.build.json",
+    "lint": "eslint --ext .js,.ts .",
+    "prepare": "rm -rf dist && npm run build",
+    "start": "ts-node src/cli.ts",
```
takameyer, line 15: `cli.ts` does not exist
```suggestion
    "start": "ts-node src/realm-schema.ts",
```

packages/realm-tools/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Building
+
+Source code for all tools are found in `src`. The output of the build process is found in `dist`.
+
+
+```sh
+npm install
+npm run build
+```
```
takameyer, line 31: Maybe add a run script here, like `node dist/realm-schema.js`

takameyer (commented):
Is the idea to have multiple tools in the future that live under the same `cli` command?  Or should we have seperate packages for each individual tools.  Something like `@realm/md-schema`

packages/realm-tools/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@realm/tools",
+  "version": "0.0.1",
+  "description": "Tools for work with Realm",
```
kraenhansen, line 4:
```suggestion
  "description": "Tools useful for working with Realms",
```

packages/realm-tools/src/realm-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import Realm from "realm";
+import { parse } from "ts-command-line-args";
+import * as fs from "fs";
+
+interface Relationship {
+  from: string;
+  to: string;
+}
```
kraenhansen, line 27: A type seems more appropriate here.

```suggestion
type Relationship = {
  from: string;
  to: string;
}
```

packages/realm-tools/src/realm-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface IExtractSchemaArgs {
+  inputFileName?: string;
+  outputFileName?: string;
+  format?: string;
+  help?: boolean;
+}
```
kraenhansen, line 34: Same as above.

```suggestion
type ExtractSchemaArgs = {
  inputFileName?: string;
  outputFileName?: string;
  format?: string;
  help?: boolean;
}
```

packages/realm-tools/src/realm-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+    ],
+  },
+);
+
+let fd = 0;
+if (args.outputFileName && args.outputFileName !== "-") {
+  fd = fs.openSync(args.outputFileName, "w");
+}
+
+const writer = (line: string) => {
```
kraenhansen, line 114: This might as well be a function.

```suggestion
function writer(line: string) {
```

kraenhansen (approved):
I second Andrew's question - I don't be curious if we're going for a single CLI with multiple commands or a package exposing multiple CLIs.

packages/realm-tools/package.json
kneth: > Is the idea to have multiple tools in the future that live under the same cli command?

No, and I will remove `start`.

packages/realm-tools/README.md
```diff
@@ -0,0 +1,62 @@
...
+Example:
+
+```sh
+npx realm-schema -i samples/default.realm -o MyClasses.md -f mermaid
+```
+
+The output will look like
+
+```mermaid
+classDiagram
```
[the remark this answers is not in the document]
kneth: Does the Realm file contains any tables?

takameyer:
This needs `typescript` as a devDependency, otherwise the build script will not work.  I tried installing this from another directory and it complained that `tsc` wasn't installed.

packages/realm-tools/README.md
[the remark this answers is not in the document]
takameyer: I tried the demo Realm from realm-studio and it works.

takameyer:
>Is the idea to have multiple tools in the future that live under the same cli command? Or should we have seperate packages for each individual tools. Something like @realm/md-schema

I'll answer my own question here.  One can establish an array for `bin` in the package.json.  So we could have a collection of packages under one project.  I'm happy with that.

packages/realm-tools/src/realm-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+  outputFileName?: string;
+  format?: string;
+  help?: boolean;
+};
+
+type FormatterFunction = (r: Realm) => void;
+type Formatters = {
+  [key: string]: FormatterFunction;
+};
+
```
takameyer, line 40: [NIT] In a later PR, we should also include a helpful `-h` output for unexpected flags/inputs/etc.

packages/realm-tools/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "Kenneth Geisshirt <kenneth.geisshirt@mongodb.com> (https://github.com/kneth)"
+  ],
+  "license": "See the actual license in the file LICENSE",
+  "dependencies": {
+    "realm": ">=10",
+    "ts-command-line-args": "^2.2.1"
+  },
+  "devDependencies": {
+    "@types/command-line-args": "^5.2.0",
+    "@types/node": "^17.0.29"
```
takameyer, line 43: [Blocker] `typescript` needs to be added to `devDependencies`, otherwise anyone consuming this will be forced to install `typescript`.

packages/realm-tools/src/realm-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+const mermaid: FormatterFunction = (realm: Realm) => {
+  writer("classDiagram");
```
takameyer, line 42: [NIT] The output markdown file does not contain the `mermaid` wrapper syntax.  I think it should.

takameyer (changes requested):
I attempted to install this in a bare project, and it failed due to `tsc` not being installed.  To fix this, `typescript` should be added to `devDependencies`

packages/realm-tools/README.md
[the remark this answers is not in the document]
kneth: It doesn't work so nicely with dictionaries and sets.

kneth:
> Is the idea to have multiple tools in the future that live under the same cli command?

My idea is that `@realm/tools` will with time become a collection of unrelated tools. Some tools will likely be more useful for us that for our users while other tools like be useful for all.

packages/realm-tools/README.md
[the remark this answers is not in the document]
kneth: It supports set and dictionary now

packages/realm-tools/src/realm-schema.ts
kneth: Done!

kneth:
I have resolved all comments (including the blocking one). I will merge when CI permit.

PR #4541, opened 2022-04-29
Add a workaround for React Native `Proxy` objects pre-JSI not working properly with JSC.

### LINKED ISSUE #4507: the result from `useQuery` is not usable in adding subscriptions.

tomduncalf:
## What, How & Why?
As described in https://github.com/realm/realm-js/issues/4507#issuecomment-1112237740, React Native `Proxy` objects which do not use JSI do not seem to be compatible with JSC – it doesn't "unwrap" them properly so can't see the wrapped type.

This means that you cannot pass a result from `useQuery` in `@realm/react` to the subscription set mutation functions, as it is wrapped in a Proxy.

This workaround stores the unproxied Results as a non-enumerable field on the proxied object, so that we can check for its presence and pass that to C++ instead of the proxied version when calling subscription set mutation functions.

Note that this problem seems to go away with JSI in `v11` (regardless of whether Hermes is being used) so another option is that we just accept the issue for now, and wait until `v11` becomes `master`

This closes #4507

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


lib/mutable-subscription-set.js
```diff
@@ -0,0 +1,50 @@
...
+// in a proxy) to be passed to the subscription mutation methods, we need to store
+// the unproxied results on the proxy object as a non-enumerable field with a symbol key
+// (see `useQuery.ts` in @realm/react), then in here, we check if that field exists, and
+// if so we pass the original unproxied results to C++.
+//
+// Once our v11 branch is merged, we can revert this change as JSI React Native `Proxy`s
+// work fine, by reverting PR #4541.
+const instanceMethods = (realmConstructor) => ({
+  add(query, options) {
+    return this._add(query[symbols.REALM_REACT_PROXIED_OBJECT] || query, options);
```
kneth, line 35: I have some reservations about `REALM_REACT_PROXIED_OBJECT` - it is leaking one package to another and in the opposite direction of the dependency. Could we find a neutral name?

kneth (approved):
My only comment is about naming of a constant and it should not block merging this PR.

tomduncalf:
Closing in favour of waiting for `v11`

lib/mutable-subscription-set.js
takameyer: Perhaps `ORIGINAL_REALM_COLLECTION`

lib/mutable-subscription-set.js
```diff
@@ -0,0 +1,50 @@
...
+  },
+
+  remove(query) {
+    return this._remove(query[symbols.REALM_REACT_PROXIED_OBJECT] || query);
+  }
```
takameyer, line 40: Seems we are missing some linter fixes in this file

takameyer (changes requested):
Seems like a very good way to getting around this issue.  I would, however, like to see a test for this so that we can ensure it still works after v11 when we remove this code.

packages/realm-common/src/symbols.ts
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+export const REALM_REACT_PROXIED_OBJECT = Symbol("REALM_REACT_PROXIED_OBJECT");
```
kraenhansen, line 19: I think a comment would be good on the purpose and usefulness of this.

packages/realm-react/package.json
```diff
@@ -13,7 +13,8 @@
   },
   "dependencies": {
     "lodash": "^4.17.21",
-    "realm": ">=10"
+    "realm": ">=10",
+    "@realm.io/common": "*"
```
kraenhansen, line 17: I suspect existing package locks won't upgrade this if the version isn't bumped.
As we have [`forceLocal` enabled](https://github.com/realm/realm-js/blob/master/lerna.json#L12), there shouldn't be an issue with this version not matching the package.json of `react-common`.

```suggestion
    "@realm.io/common": "^0.1.2"
```

lib/mutable-subscription-set.js
kraenhansen, line 35: I'd suggest a shorter name. It's already in a namespace from the package name.

```suggestion
    return this._add(query[symbols.PROXY_TARGET] || query, options);
```

kraenhansen (approved):
I have just minor suggestions.

lib/mutable-subscription-set.js
tomduncalf: I went with `PROXY_TARGET` as per Kræn's suggestion

packages/realm-common/src/symbols.ts
tomduncalf: Added

packages/realm-react/package.json
tomduncalf: Got it, thanks, fixed

lib/mutable-subscription-set.js
tomduncalf: Good suggestion, changed to this 👍

tomduncalf:
> Seems like a very good way to getting around this issue. I would, however, like to see a test for this so that we can ensure it still works after v11 when we remove this code.

@takameyer I like the idea of adding a test but I am struggling to think of a good way to do it, because in order to get access to `realm.subscriptions`, you need to have a valid sync config, which means you need a Realm app to test against. This would mean we'd need to do one of:
1. Make the Realm React tests support working with the app importer and move some of the Mocha hooks etc. used to interact with it from `integration-tests` into `realm/common`. This would be the right approach but feels like a lot of work for one test for a temporary(ish) test.
2. Make the integration tests somehow test useQuery when running in React Native. Because it's a hook, this might mean setting up a whole React app to test in. Again, a lot of work and would also be breaking the separation of concerns in a weird way possibly (not sure if the main integration tests should be testing Realm React, although you could argue it's appropriate)
3. Add some way of exposing subscriptions without a valid sync session for testing purposes, which feels quite hacky

If you have other ideas let me know though as I agree it would be good, but I'm not sure if it's worth a fairly major bit of work to achieve. I might be missing something easier. As is, we'll have to remember to test it again with v11 which I agree is not ideal!

kraenhansen (approved):

PR #4544, opened 2022-05-02
Singleton useObject

### LINKED ISSUE #4521: useObject will return referentially different objects for the same object

All the time

### Description

Given the immutability conventions of react you would expect that 

```jsx
const object = useObject('foo', 0)
const object2 = useObject('foo', 0)
assert(object === object2)
```

However this is not the case.

Referential equality is commonplace for the equivalent of useObject in all the most common state libraries such as redux, mobx and zustand. It's very common to use hooks instead of prop drilling and therefore natural to use `useObject` multiple times for the same object. This issue could produce potential gotchas for users such as unnecessary re-renders. 

One solution would be to maintain an object cache, each object would only have a single realm listener. The cache entry would track the number of active hooks there are for the object via list of plain JS listener callbacks. When the component unmounts the listener would be removed from the list of listeners. When the list of listeners drops to 0 the cache entry would be purged.

This model would likely have some other benefits 
* Reduce the overhead of a new `useObject` as a new realm object instance and listener wouldn't have to be made if it was already in the cache
* Less potential gotchas with creating listeners e.g. #4375 

A similar approach could be applied to `useQuery`

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

_No response_

### Version

main

Local Database only

### Are you using encryption?

No, not using encryption

ios

### Build environment

### Cocoapods version

_No response_

mfbx9da4:
## What, How & Why?
To play nicely with React's immutability conventions this PR introduces a singleton object which is returned by useObject. 

This closes #4521

⚠️ For the time being this is untested illustrative code. Opened as a draft PR to open discussion with @takameyer. 

## ☑️ ToDos

* [x] Return singleton object
* [x] One realm listener per singleton object
* [x] Handle deleted or invalid objects gracefully
* [x] Handle objects not yet inserted gracefully
* [x] Support type being a constructor
* [ ] Make all existing tests pass
* [ ] Test listening to object before it is inserted
* [ ] Test listening to object after it is deleted and then inserted again
* [ ] Test there are no memory leaks after 
* [ ] Test referential equality
* [ ] Do the same for collections
...
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/__tests__/useObjectRender.test.tsx
+++ b/packages/realm-react/src/__tests__/useObjectRender.test.tsx
@@ -170,6 +170,9 @@ const TestComponent: React.FC<{ testID?: string }> = ({ testID }) => {
     return <View testID={testID} />;
   }
 
+  const favoriteName = list?.favoriteItem?.name;
+  console.log("list?.favoriteItem?.name", favoriteName);
+
   return (
     <View testID={testID}>
       <FlatList testID="list" data={list?.items ?? []} keyExtractor={keyExtractor} renderItem={renderItem} />;
@@ -198,8 +201,8 @@ const TestComponent: React.FC<{ testID?: string }> = ({ testID }) => {
         </TouchableHighlight>
       </View>
       {list?.favoriteItem && (
-        <View testID={`favoriteItem-${list?.favoriteItem.name}`}>
-          <Text>{list?.favoriteItem.name}</Text>
+        <View>
+          <Text testID="favoriteItemText">{favoriteName}</Text>
         </View>
       )}
     </View>
@@ -286,7 +289,7 @@ describe("useObject: rendering objects with a Realm.List property", () => {
         forceSynchronousNotifications(testRealm);
       });
 
-      expect(getByTestId(`favoriteItem-${object.items[0].name}`)).toBeTruthy();
+      expect(getByTestId("favoriteItemText")).toHaveTextContent(object.items[0].name);
 
       await act(async () => {
         testRealm.write(() => {
@@ -295,7 +298,7 @@ describe("useObject: rendering objects with a Realm.List property", () => {
         forceSynchronousNotifications(testRealm);
       });
 
-      expect(getByTestId(`favoriteItem-${object.items[1].name}`)).toBeTruthy();
+      expect(getByTestId("favoriteItemText")).toHaveTextContent(object.items[1].name);
 
       await act(async () => {
         testRealm.write(() => {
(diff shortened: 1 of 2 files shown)


takameyer:
@mfbx9da4 Thanks for submitting this.  I'll look it over with the team.  Next week I may have some time to add some tests to verify this.

takameyer:
Ill have more time to review this next week.  Really appreciate the effort here!

takameyer:
@mfbx9da4  We have been discussing internally how we want to generally handle external PRs.  I'll first quote an upcoming update to our contribution guide:

>[...] indicate how you would like us to support you. We will happily guide you and work with you to move the PR to a point where it can be merged. It might require considerable work at your end to meet our expectations (code quality, API docs, TS defs, tests, etc.). In the case you want to move on and not work with us, please let us know. If the PR meets our expectations, we will merge it - and if it doesn't, we will either take over or close it, depending on the requirement on our time and our current priorities.

So, let us know your intentions with the PR 🙂.  We have this on our backlog, but it will probably take a while before this gets prioritised on our end.  If you would like to work on it, it may take some time with the back and forth and a bit of patience, but will in the end be more likely to be accepted into our code base.

takameyer:
@mfbx9da4 Sorry for the delay.  I've added @tomduncalf and @kraenhansen to the PR as well.  Let's work together to get this in.

packages/realm-react/src/__tests__/useObjectHook.test.tsx
```diff
@@ -84,11 +84,23 @@ describe("useObject hook", () => {
     expect(object).toMatchObject(dog2);
   });
 
-  it("object is null", () => {
+  it("object is undefined", () => {
```
takameyer, line 87: symantically this should be `null` if the object doesn't exist

packages/realm-react/src/__tests__/useObjectHook.test.tsx
```diff
@@ -84,11 +84,23 @@ describe("useObject hook", () => {
...
+  it("identical useObject calls return the same reference", () => {
+    const [, dog2] = testDataSet;
+    const { result } = renderHook(() => useObject<IDog>("dog", dog2._id));
+    const { result: result2 } = renderHook(() => useObject<IDog>("dog", dog2._id));
+
+    const object = result.current;
+    const object2 = result2.current;
+
+    expect(object).toBeDefined();
+    expect(object).toEqual(object2);
```
takameyer, line 104: This passes no matter if the reference is different.  Change to `toBe`

TommyTheTribe:
Does anyone know why this branch has not been merged ? Thanks

kraenhansen:
@TommyTheTribe looks stale to me. I see @takameyer made a few comments which was not addressed and now it also needs a rebase.

takameyer:
@TommyTheTribe I added tests for the changes made here and they failed. We haven't had the resources available to tackle this ourselves just yet.

Issue #4535, opened 2022-04-28
Process does not exit when using `new Realm.App` on Node


tomduncalf:
**To reproduce:**

1. Create a new node project with the following code in `index.js`:
```
const Realm = require("realm");
const app = new Realm.App({ id: "myapp-abcde" });
console.log(`app id: ${app.id`})
// process.exit() <- need to call this to actually finish running
```
1. Observe that `node index.js` never exits

This issue seems to have started happening in `realm@10.2.0`. Possibly related: https://github.com/realm/realm-js/issues/3525.

May be the real fix for https://github.com/realm/realm-js/issues/4530, in which case we can revert that change.

kneth:
We have investigated it but haven't reached a conclusion. Attaching a debugger to the node process provides no Realm related stack traces:

```
(lldb) t 2
* thread #2
    frame #0: 0x00007ff80ae7b34e libsystem_kernel.dylib`kevent + 10
libsystem_kernel.dylib`kevent:
->  0x7ff80ae7b34e <+10>: jae    0x7ff80ae7b358            ; <+20>
    0x7ff80ae7b350 <+12>: movq   %rax, %rdi
    0x7ff80ae7b353 <+15>: jmp    0x7ff80ae771c5            ; cerror_nocancel
    0x7ff80ae7b358 <+20>: retq
(lldb) bt
* thread #2
  * frame #0: 0x00007ff80ae7b34e libsystem_kernel.dylib`kevent + 10
    frame #1: 0x000000010e852964 node`uv__io_poll + 948
    frame #2: 0x000000010e83f561 node`uv_run + 433
    frame #3: 0x000000010dedd839 node`node::WorkerThreadsTaskRunner::DelayedTaskScheduler::Run() + 361
    frame #4: 0x00007ff80aeb34e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #5: 0x00007ff80aeaef6b libsystem_pthread.dylib`thread_start + 15
(lldb) t 3
* thread #3
    frame #0: 0x00007ff80ae793ea libsystem_kernel.dylib`__psynch_cvwait + 10
libsystem_kernel.dylib`__psynch_cvwait:
->  0x7ff80ae793ea <+10>: jae    0x7ff80ae793f4            ; <+20>
    0x7ff80ae793ec <+12>: movq   %rax, %rdi
    0x7ff80ae793ef <+15>: jmp    0x7ff80ae771c5            ; cerror_nocancel
[...]

PR #4556, opened 2022-05-05
Add test for clean exit from node

tomduncalf:
## What, How & Why?

This reproduces https://github.com/realm/realm-js/issues/4535

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kneth (approved):

integration-tests/tests/src/node/clean-exit.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { execSync } from "child_process";
+
+describe("Clean exit for Node.js scripts", () => {
+  // Repro for https://github.com/realm/realm-js/issues/4535 - currently still failing
+  it.skip("exits cleanly when creating a new Realm.App", () => {
+    execSync(
+      `node -e 'const Realm = require("realm"); const app = new Realm.App({ id: "myapp-abcde" }); Realm.clearTestState();'`,
+      {
+        timeout: 5000,
```
kraenhansen, line 27: I think this extends the timeout of the test? I.e. if it takes 3 seconds for the process to exit, it will fail although it didn't timeout. Could we read this timeout from the test's context instead?
```suggestion
        timeout: this.timeout,
```
```suggestion
        timeout: 5000,
```

kraenhansen, line 27: I think this extends the timeout of the test? I.e. if it takes 3 seconds for the process to exit, it will fail although it didn't timeout. Could we read this timeout from the test's context instead?
```suggestion
        timeout: this.timeout,
```
```suggestion
        timeout: 5000,
```

tomduncalf: Ah yeah, good spot. How about this change: https://github.com/realm/realm-js/pull/4556/commits/924c7dbc6e1039999bde00b1df889ffd10692971 so that it never exceeds either the test timeout or 5000ms? 

I think 5000ms should always be enough for this test in any case, if you were running the suite with say 60 second timeout for some reason, there's no point in this one waiting for 60 seconds also.

kraenhansen: Good idea 👍 Looks great.

kraenhansen (approved):

PR #4557, opened 2022-05-05
UserProvider and useUser

### LINKED ISSUE #4456: Create useApp and AppProvider
Create a Provider that on render creates a context based on the arguments passed to the provider. Create a hook that returns opened app.

takameyer:
## What, How & Why?
Add UserProvider and useUser hook to allow access to the user within any place in the application.

Note: There are no tests, as this is unreasonable to test without actually spinning up a BAAS server.  This is an unrealistic expectation for this work and should be done in a greater effort to refactor tests.  I have, however, throughly tested this with the example application.  If we do start getting bugs, we should find a way to create regression tests.

We will also need a follow up PR to update the Realm dependency as soon as the new release is out, as this is dependent on the new App and User listener events.

This closes #4456

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/RealmProvider.tsx
+++ b/packages/realm-react/src/RealmProvider.tsx
@@ -19,8 +19,15 @@
 import React, { useEffect, useState, useRef } from "react";
 import Realm from "realm";
 import { isEqual } from "lodash";
+import { useUser } from "./UserProvider";
 
-type ProviderProps = Realm.Configuration & {
+type PartialRealmConfiguration = {
+  [P in keyof Omit<Realm.Configuration, "sync">]?: Realm.Configuration[P];
+} & {
+  sync?: Partial<Realm.SyncConfiguration>;
+};
+
+type ProviderProps = PartialRealmConfiguration & {
   fallback?: React.ComponentType<unknown> | React.ReactElement | null | undefined;
 };
 
@@ -62,6 +69,10 @@ export function createRealmProvider(
    */
   return ({ children, fallback: Fallback, ...restProps }) => {
     const [realm, setRealm] = useState<Realm | null>(null);
+
+    // Automatically set the user in the configuration if its been set.
+    const user = useUser();
+
     // We increment `configVersion` when a config override passed as a prop
     // changes, which triggers a `useEffect` to re-open the Realm with the
     // new config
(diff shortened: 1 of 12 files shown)


packages/realm-react/src/RealmProvider.tsx
```diff
@@ -78,11 +89,21 @@ export function createRealmProvider(
...
       const combinedConfig = mergeRealmConfiguration(realmConfig, restProps);
-      if (!areConfigurationsIdentical(configuration.current, combinedConfig)) {
-        configuration.current = combinedConfig;
-        setConfigVersion((x) => x + 1);
+
+      // If there is a user in the current context and not one set by the props, then use the one from context
+      const combinedConfigWithUser =
+        user && !combinedConfig.sync?.user
+          ? mergeRealmConfiguration(combinedConfig, { sync: { user } })
+          : combinedConfig;
```
kraenhansen, line 97: Would this simplification be possible?

```suggestion
      const combinedConfigWithUser = mergeRealmConfiguration({ sync: { user } }, combinedConfig);
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -19,8 +19,15 @@
 import React, { useEffect, useState, useRef } from "react";
 import Realm from "realm";
 import { isEqual } from "lodash";
+import { useUser } from "./UserProvider";
 
-type ProviderProps = Realm.Configuration & {
+type PartialRealmConfiguration = {
+  [P in keyof Omit<Realm.Configuration, "sync">]?: Realm.Configuration[P];
+} & {
```
kraenhansen, line 26: Is this different from this simpler expression?

```suggestion
type PartialRealmConfiguration = Omit<Realm.Configuration, "sync"> & {
```

takameyer: Yes, since the `?` is removed.  However this might be better:
```
type PartialRealmConfiguration = Omit<Partial<Realm.Configuration>, "sync"> & {
  sync?: Partial<Realm.SyncConfiguration>;
};
```

packages/realm-react/src/RealmProvider.tsx
takameyer: With this suggestion, it is possible for a non-sync application to have {sync: {user: undefined}} included in the configuration.  But maybe there is a way to simplify this.

takameyer:
```
 const combinedConfigWithUser =
        combinedConfig?.sync && user ? mergeRealmConfiguration({ sync: { user } }, combinedConfig) : combinedConfig;
```
This is a little simpler.

kraenhansen: > it is possible for a non-sync application to have {sync: {user: undefined}}

Yeah, I figured this could be an edge-case too and wondered if `mergeRealmConfiguration` implemented a deletion of the `sync` property if all members has `undefined` values or something.

kraenhansen (approved):

packages/realm-react/src/RealmProvider.tsx
takameyer: That's perhaps another possibility

PR #4562, opened 2022-05-06
Fix for Flexible Sync Configuration

### LINKED ISSUE #4552: Flexible sync type definitions require `strictNullChecks` to be enabled
The type definitions for flexible sync require `strictNullChecks` to be enabled in the user's `tsconfig.json` otherwise they will get errors: https://github.com/realm/realm-js/blob/master/types/index.d.ts#L144-L145

I can't remember the details of why this is so, but Kræn and I did quite a bit of investigation and concluded that the type discrimination using the `flexible: true` and `flexible?: false | undefined` did not work properly without strict null checks. I think there was a somewhat logical explanation for this, I can try to cast my mind back if useful!

We thought this would be a non-issue in practice for the majority of users, but unfortunately it turns out that Expo [defaults to strict mode disabled](https://docs.expo.dev/guides/typescript/#project-configuration) so there is a chance that Expo users will hit this issue. I'm not sure if there is a good workaround, but it warrants further investigation.

takameyer:
## What, How & Why?
This makes the flexible boolean incompatible with paritionValue,
without setting tsconfig to false.  It does however not allow,
flexible to be false, but that's a better tradeoff in the end.

This closes #4552

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Adding an object to a Set, deleting the parent object, and then deleting the previously mentioned object causes crash ([#5387](https://github.com/realm/realm-core/issues/5387), since v10.5.0)
 * Synchronized Realm files which were first created using SDK version released in the second half of August 2020 would be redownloaded instead of using the existing file, possibly resulting in the loss of any unsynchronized data in those files (since v10.10.1).
 * Fixed buildtime error where our package would be ignored due to use of the deprecated `podspecPath` and `sharedLibraries` keys in the `react-native.config.js`. ([#4553](https://github.com/realm/realm-js/issues/4553))
-* Fixed flexible sync configuration types to be compatible with `strict: false`. ([#4552](https://github.com/realm/realm-js/issues/4552))
+* Fixed flexible sync configuration types to be compatible with Typescript when `strict` mode is disabled. ([#4552](https://github.com/realm/realm-js/issues/4552))
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Adding an object to a Set, deleting the parent object, and then deleting the previously mentioned object causes crash ([#5387](https://github.com/realm/realm-core/issues/5387), since v10.5.0)
 * Synchronized Realm files which were first created using SDK version released in the second half of August 2020 would be redownloaded instead of using the existing file, possibly resulting in the loss of any unsynchronized data in those files (since v10.10.1).
 * Fixed buildtime error where our package would be ignored due to use of the deprecated `podspecPath` and `sharedLibraries` keys in the `react-native.config.js`. ([#4553](https://github.com/realm/realm-js/issues/4553))
+* Fixed flexible sync configuration types to be compatible with `strict: false`. ([#4552](https://github.com/realm/realm-js/issues/4552))
 
 ### Compatibility
 * MongoDB Realm Cloud.
--- a/integration-tests/tests/src/tests/sync/flexible.ts
+++ b/integration-tests/tests/src/tests/sync/flexible.ts
@@ -152,18 +152,19 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
         }).to.throw("'partitionValue' cannot be specified when flexible sync is enabled");
       });
 
-      it("accepts { flexible: false } and a partition value", function () {
-        expect(() => {
-          new Realm({
-            sync: {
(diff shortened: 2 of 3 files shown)


takameyer:
Just for a little more background on these changes, I found that instead of trying to make a type that supports `{flexible: false, partitionValue: "something"}`, we could just make `{flexible: false}` completely incompatible and only support `{flexible: true, partitionValue: never}` and `{flexible: never, partitionValue: "something"}`.
This has the advantage of fixing the type errors we were having with `strict: false`.
I think this is a good tradeoff, as I do not really see the use case for `flexible: false`.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -152,18 +152,19 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
-            sync: {
-              _sessionStopPolicy: SessionStopPolicy.Immediately,
-              flexible: false,
-              user: this.user,
-              partitionValue: "test",
-            },
-          });
-        }).to.not.throw();
-      });
+      // NOTE: This is not a compatible configuration anymore and will cause a typescript error
```
kneth, line 155: `NIT`: Can it be removed?

kneth (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Adding an object to a Set, deleting the parent object, and then deleting the previously mentioned object causes crash ([#5387](https://github.com/realm/realm-core/issues/5387), since v10.5.0)
 * Synchronized Realm files which were first created using SDK version released in the second half of August 2020 would be redownloaded instead of using the existing file, possibly resulting in the loss of any unsynchronized data in those files (since v10.10.1).
 * Fixed buildtime error where our package would be ignored due to use of the deprecated `podspecPath` and `sharedLibraries` keys in the `react-native.config.js`. ([#4553](https://github.com/realm/realm-js/issues/4553))
+* Fixed flexible sync configuration types to be compatible with `strict: false`. ([#4552](https://github.com/realm/realm-js/issues/4552))
```
tomduncalf, line 14: I think it's good to clarify that we are talking about a Typescript and not Realm feature here 
```suggestion
* Fixed flexible sync configuration types to be compatible with Typescript when `strict` mode is disabled. ([#4552](https://github.com/realm/realm-js/issues/4552))
```

types/index.d.ts
```diff
@@ -130,9 +130,8 @@ declare namespace Realm {
         clientResetAfter?: ClientResetAfterCallback;
     }
 
-    interface BaseSyncConfiguration<ClientResetModeT = ClientResetMode>{
+    interface BaseSyncConfiguration{
```
tomduncalf, line 133: `comment` I guess the idea with the generic was to avoid duplicating the `clientReset` key but for whatever reason I didn't end up going through with it 🤷‍♂️ happy to remove for now

types/index.d.ts
```diff
@@ -141,16 +140,16 @@ declare namespace Realm {
         error?: ErrorCallback;
     }
 
-    // Note: This type does not work correctly without strictNullChecks enabled –
-    // a config of { flexible: true } will incorrectly have a type error
+    // Note: Since there really is no use case for {flexible: false}, we are
+    // hard setting this to only allowing "true".
```
tomduncalf, line 144: It's not a big deal but maybe we should explain the reason behind this

```suggestion
    // We only allow `flexible` to be `true` or `undefined` - `{ flexible: false
    // }` is not allowed. This is because TypeScript cannot discriminate that
    // type correctly with `strictNullChecks` disabled, and there's no real use
    // case for `{ flexible: false }`.
```

tomduncalf (approved):
👍 Thanks for finding this workaround! I've only made minor suggestions which you can choose to ignore if preferred

types/index.d.ts
takameyer: I just noticed it wasn't being used so I thought I'd remove it.  Not really related to this PR, but also nothing ground breaking

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: I'll do what Tom suggests here.

kraenhansen: Using [`@ts-expect-error`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) is even better for this use-case as it will fail if no error was raised.

takameyer: Good idea.  That is now implemented.

kraenhansen (approved):

PR #4568, opened 2022-05-10
Add `initialSubscriptions` sync config option to bootstrap initial set of flexible sync subscriptions

tomduncalf:
## What, How & Why?

This closes https://github.com/realm/realm-js/issues/4515, by introducing an `initialSubscriptions` option on the Realm's `sync` config, which allows you to bootstrap an initial set of subscriptions (using the same callback format as `subscriptions.update`). This will return a promise, resolved when the new subscriptions have been synchronised – i.e. if the user calls `await Realm.open`, it will not resolve until all the initial subscription data has been synchronised.

This is as part of the additive changes to flexible sync, in response to feedback that this a common use case. It's worth pointing out that this functionality should probably reside in core, but it was determined that there was not time to implement that, so for now each SDK is implementing it separately.

By default, the callback is only called when the Realm does not already exist, to allow the user to bootstrap an initial subscription set. The `rerunOnStartup` flag allows the user to rerun the callback every time the Realm is opened – the intended use case for this is that a user could implement e.g. a date range query in their subscription, which is updated every time the app is opened, in order to work around the lack of date range operators in RQL.

For example, to implement a subscription for "all people born in the last 30 days", which updates the date range every time you start the app, you could do something like:
```js
const thirtyDaysAgo = new Date(new Date().setDate(new Date().getDate() - 30));
const config = {
  // ...
  sync: {
    flexible: true,
    user,
    initialSubscriptions: {
      update: (realm) => {
        realm.subscriptions.update((subs) => {
          subs.add(
            realm.objects("Person").filtered("dateOfBirth > $0", thirtyDaysAgo), 
            // This is a named subscription, so will replace any existing subscription with the same name
            { name: "People30Days" }
          );
        });
      },
      rerunOnStartup: true,
    },
  },
};
```

One gotcha to be aware of with this is that if you inline your config with the call to `Realm.open` and then try to use the `realm` variable that you are declaring inside the `update` callback, it won't work as it's not yet defined – you need to use the Realm instance provided as a callback argument. I'm not sure if there's anything we can do about this (except try to signpost it if we think it's a big issue?) and `onFirstOpen` presumably has the same issue.

```js
  // ...
  sync: {
    initialSubscriptions: {
      update: () => {
        realm.... // <-- realm is undefined
```  

### Alternatives considered 

I am open to feedback on the API for this, as I do feel it's perhaps a bit awkward that we are introducing a new concept that is actually quite similar to `onFirstOpen`. `initialSubscriptions.update` is presented as a callback for updating subscriptions, but in actual fact the user could do anything in this callback, so it could be a more generic thing. I felt like this was the "least bad" solution in the end – alternatives I considered, and the reasons for rejecting them are:

- Reusing the existing `onFirstOpen` config option and making it `await`able (right now it does not return a promise), and adding a counterpart `onOpen` which is called every time the Realm opens to cover the `rerunOnStartup` case
  - `onFirstOpen` does not work for calling `subscriptions.update`, because the call to `onFirstOpen` from [core](https://github.com/realm/realm-core/blob/master/src/realm/object-store/shared_realm.cpp#L470) is already in a write transaction. Therefore calling `subscriptions.update` from here fails, as `update` opens its own write transaction.
  - As `onFirstOpen` is implemented using [`initialization_function` in core](https://github.com/realm/realm-core/blob/master/src/realm/object-store/shared_realm.cpp#L470),  this would require changes to core to add the equivalent for `onOpen` (which requires touch a lot of files to pass it through the various configs and layers) and to make it `await`able. We would also need to change the semantics of `onFirstOpen` and/or `update` with regard to write transactions. 
  - This seemed like too complex a change, especially as the intention is for this initial subscription functionality to eventually be moved to core.
- Have the implementation in the SDK like it is in this PR, but make it a more generic solution (e.g. add a top level callback on the config called `onOpen`, with a `rerunOnOpenOnStartup` flag, which lets the user do initial work they want of any kind – essentially an SDK-side `await`able replacement for `onFirstOpen`)
  - This would create an awkward duplication with `onFirstOpen` – would it be a replacement for it? Or are there some semantic differences?
  - Is this even something we want to add? It would need some proper design and consideration before adding a new generic concept.
  - For the initial subscriptions use case (which is currently the only one, and the one we want to streamline), this would require users to remember to `return realm.subscriptions.waitForSynchronization();` from their `onOpen` callback, otherwise we wouldn't wait for synchronisation.
    - We could add a `waitForSynchronisation` flag to do this implicitly, but that feels messy.
- Implementing the solution purely in JS in `lib/extensions.js`
  - My initial implementation did this, which was simpler in terms of the code, but did not work for calls to `new Realm` – right now all our config options are supported for both `new Realm` and `Realm.open` so I thought we should uphold this.
 
One other idea I did have was:

- Make the `initialSubscriptions.update` callback internally wrapped in `subscriptions.update` and change its signature to `update: (subs: MutableSubscriptionSet, realm: Realm)`
  - This would perhaps make it a bit more ergonomic for users as they don't need to call `realm.subscriptions.update`, and it would guide them more clearly to just using this callback for updating subscriptions
  - I felt like this might be introducing unnecessary complexity and was somewhat redundant as you already have the `realm`, but looking at it with fresh eyes maybe this is actually a nicer solution? Thoughts welcome!

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add ability to listen to changes to `Realm.App` and `Realm.User`. ([#4455](https://github.com/realm/realm-js/issues/4455))
+* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function which bootstraps a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs) (#4561[https://github.com/realm/realm-js/pull/4561]).
 
 ### Fixed
 * Fixed issue where React Native apps would sometimes show stale Realm data until the user interacted with the app UI. ([#4389](https://github.com/realm/realm-js/issues/4389), since v10.0.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,30 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Enhancements
 * Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function to bootstrap a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs, using the `rerunOnOpen` flag). (#4561[https://github.com/realm/realm-js/pull/4561])
 
+    Example usage:
+    ```ts
+    const thirtyDaysAgo = new Date(new Date().setDate(new Date().getDate() - 30));
+    const config = {
+      // ...
+      sync: {
+        flexible: true,
+        user,
+        initialSubscriptions: {
+          update: (realm) => {
+            realm.subscriptions.update((subs) => {
+              subs.add(
+                realm.objects("Person").filtered("dateOfBirth > $0", thirtyDaysAgo),
+                // This is a named subscription, so will replace any existing subscription with the same name
+                { name: "People30Days" }
+              );
+            });
+          },
+          rerunOnStartup: true,
+        },
+      },
+    };
+    ```
+
 ### Fixed
 * Flexible sync would not correctly resume syncing if a bootstrap was interrupted. ([realm/realm-core#5466](https://github.com/realm/realm-core/pull/5466), since v10.12.0)
 * The sync client may upload corrupted internal data leading to a fatal error from the sync server. ([realm/realm-core#5460](https://github.com/realm/realm-core/pull/5460), since v10.16.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function which bootstraps a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs). (#4561[https://github.com/realm/realm-js/pull/4561])
+* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function to bootstrap a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs, using the `rerunOnOpen` flag). (#4561[https://github.com/realm/realm-js/pull/4561])
 
 ### Fixed
 * Flexible sync would not correctly resume syncing if a bootstrap was interrupted. ([realm/realm-core#5466](https://github.com/realm/realm-core/pull/5466), since v10.12.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function which bootstraps a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs). (#4561[https://github.com/realm/realm-js/pull/4561])
 
 ### Fixed
 * Flexible sync would not correctly resume syncing if a bootstrap was interrupted. ([realm/realm-core#5466](https://github.com/realm/realm-core/pull/5466), since v10.12.0)
@@ -22,7 +22,6 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add ability to listen to changes to `Realm.App` and `Realm.User`. ([#4455](https://github.com/realm/realm-js/issues/4455))
-* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function which bootstraps a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs) (#4561[https://github.com/realm/realm-js/pull/4561]).
 
 ### Fixed
 * Fixed issue where React Native apps would sometimes show stale Realm data until the user interacted with the app UI. ([#4389](https://github.com/realm/realm-js/issues/4389), since v10.0.0)
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -94,7 +94,7 @@
  * instance to allow you to setup the initial set of subscriptions. See
  * {@link Realm.App.Sync.SubscriptionSet#update} for more information.
  * @property {Boolean} rerunOnOpen - optional flag. If `true`, the
- * {@link updateCallback} will be rerun every time the Realm is opened (e.g.
+ * {@link update} callback will be rerun every time the Realm is opened (e.g.
  * every time a user opens your app), otherwise (by default) it will only be run
  * if the Realm does not yet exist.
  */
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -93,7 +93,7 @@
  * {@link Realm.App.Sync~MutableSubscriptionSet} instance and the {@link Realm}
  * instance to allow you to setup the initial set of subscriptions. See
  * {@link Realm.App.Sync.SubscriptionSet#update} for more information.
- * @property {Boolean} returnOnStartup - optional flag. If `true`, the
+ * @property {Boolean} rerunOnOpen - optional flag. If `true`, the
  * {@link updateCallback} will be rerun every time the Realm is opened (e.g.
  * every time a user opens your app), otherwise (by default) it will only be run
  * if the Realm does not yet exist.
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -76,10 +76,8 @@
  *     user,
  *     flexible: true,
  *     initialSubscriptions: {
- *       update: realm => {
- *         realm.subscriptions.update(subs => {
- *           subs.add(realm.objects('Task'));
- *         })
+ *       update: (subs, realm) => {
+ *         subs.add(realm.objects('Task'));
  *       }
  *     }
  *   },
@@ -91,9 +89,9 @@
  * // subscriptions fully synchronised.
  * ```
  * @typedef {Object} Realm.App.Sync~InitialSubscriptionsConfiguration
- * @property {callback(realm)} update - callback called with the {@link Realm} instance
- * to allow you to setup the initial set of subscriptions by calling
- * `realm.subscriptions.update`. See
+ * @property {callback(realm)} update - callback called with a
+ * {@link Realm.App.Sync~MutableSubscriptionSet} instance and the {@link Realm}
+ * instance to allow you to setup the initial set of subscriptions. See
  * {@link Realm.App.Sync.SubscriptionSet#update} for more information.
  * @property {Boolean} returnOnStartup - optional flag. If `true`, the
  * {@link updateCallback} will be rerun every time the Realm is opened (e.g.
--- a/docs/sync.js
+++ b/docs/sync.js
@@ -48,7 +48,9 @@
  * @typedef {Object} Realm.App.Sync~SyncConfiguration
  * @property {Realm.User} user - A {@link Realm.User} object obtained by calling `Realm.App.logIn`.
  * @property {Realm.App.Sync~SSLConfiguration} [ssl] - SSL configuration.
- * @property {string|number|BSON.ObjectId|null} partitionValue - The value of the partition key.
+ * @property {boolean} flexible - Whether to use flexible sync (if `true`) or partition based sync (default)
+ * @property {string|number|BSON.ObjectId|null} partitionValue - The value of the partition key. Only valid if using partition based sync.
+ * @property {Realm.App.Sync~InitialSubscriptionsConfiguration} initialSubscriptions - Optional object to configure the setup of an initial set of flexible sync subscriptions to be used when opening the Realm. Only valid if using flexible sync. See {@link Realm.App.Sync~InitialSubscriptionsConfiguration}.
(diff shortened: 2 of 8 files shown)


src/js_realm.hpp
```diff
@@ -785,17 +787,100 @@ bool RealmClass<T>::get_realm_config(ContextType ctx, size_t argc, const ValueTy
...
+    ObjectType sync_object = Value::validated_to_object(ctx, sync_value);
+
+    ValueType initial_subscriptions_value = Object::get_property(ctx, sync_object, "initialSubscriptions");
+    if (Value::is_undefined(ctx, initial_subscriptions_value)) {
+        return;
+    }
+    ObjectType initial_subscriptions_object = Value::validated_to_object(ctx, initial_subscriptions_value);
+
+    ValueType update_value = Object::get_property(ctx, initial_subscriptions_object, "update");
+    FunctionType update_function = Value::validated_to_function(ctx, update_value, "update");
```
kneth, line 853: `NIT`: Checking if `update_value` is undefined might be a good idea

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add ability to listen to changes to `Realm.App` and `Realm.User`. ([#4455](https://github.com/realm/realm-js/issues/4455))
+* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function which bootstraps a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs) (#4561[https://github.com/realm/realm-js/pull/4561]).
```
```suggestion
* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function which bootstraps a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs). (#4561[https://github.com/realm/realm-js/pull/4561])
```

kneth (commented):
First pass: the code looks good.

I will respond to the many questions later once I have digested them :smile:

kraenhansen:
### My thoughts after reading the PR description

> Reusing the existing `onFirstOpen` config option and making it `await`able

When reading your initial description, this was an alternative that I had in mind as well. I think I understand the tradeoffs, although perhaps the first drawback of  `subscriptions.update` failing because it opens its own write transaction could probably be fixed, by checking if it's already in a transaction. Or is that too complicated? I get that since this is relying on a core primitive and the idea here is to implement a solution without touching core, this is probably not the right alternative.

I'd love to see an eventual unified API based on this.

> Make the `initialSubscriptions.update` callback internally wrapped in `subscriptions.update`

I like this idea as aligns more with the use-case we have in mind for this API and reduces boilerplate.

CHANGELOG.md
kraenhansen, line 5:
```suggestion
* Added an `initialSubscriptions` option to the `sync` config, which allows users to specify a subscription update function bootstrapping a set of flexible sync subscriptions when the Realm is first opened (or every time the app runs) (#4561[https://github.com/realm/realm-js/pull/4561]).
```

docs/sync.js
```diff
@@ -61,6 +63,44 @@
...
+ *
+ * // At this point, the Realm will be open with the data for the initial set
+ * // subscriptions fully synchronised.
+ * ```
+ * @typedef {Object} Realm.App.Sync~InitialSubscriptionsConfiguration
+ * @property {callback(realm)} update - callback called with the {@link Realm} instance
+ * to allow you to setup the initial set of subscriptions by calling
+ * `realm.subscriptions.update`. See
+ * {@link Realm.App.Sync.SubscriptionSet#update} for more information.
+ * @property {Boolean} returnOnStartup - optional flag. If `true`, the
```
kraenhansen, line 98: Is this name aligned with the other SDKs? Technically it is re-running when the Realm is opened, right? Not when the app is starting up (although it might if that ends up opening the Realm).

In any case it has a typo.

```suggestion
 * @property {Boolean} rerunOnStartup - optional flag. If `true`, the
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -193,6 +193,168 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+            );
+          });
+
+          it("throws an error if `rerunOnStartup` is not a boolean", async function () {
+            this.config = getConfig(this.user, {
+              update: () => {
+                // no-op
+              },
+              rerunOnStartup: "yes please",
+            } as any);
```
kraenhansen, line 242:
```suggestion
            @ts-expect-error Intentionally testing the wrong type
            this.config = getConfig(this.user, {
              update: () => {
                // no-op
              },
              rerunOnStartup: "yes please",
            });
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -193,6 +193,168 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+          };
+        }
+
+        describe("error", function () {
+          afterEach(function () {
+            Realm.deleteFile(this.config);
+          });
+
+          it("throws an error if no `update` function is provided", async function () {
+            this.config = getConfig(this.user, {} as any);
```
kraenhansen, line 219: This suggestion applies to a couple of other places in this file.

```suggestion
            @ts-expect-error Intentionally testing the wrong type
            this.config = getConfig(this.user, {});
```

types/index.d.ts
```diff
@@ -147,6 +147,48 @@ declare namespace Realm {
...
+         */
+        initialSubscriptions?: {
+            /**
+             * Callback called with the {@link Realm} instance to allow you to setup the
+             * initial set of subscriptions by calling `realm.subscriptions.update`.
+             * See {@link Realm.App.Sync.SubscriptionSet.update} for more information.
+             */
+            update: (realm: Realm) => void;
+            /**
+             * If `true`, the {@link updateCallback} will be rerun every time the Realm is
```
kraenhansen, line 186:
```suggestion
             * If `true`, the {@link update} callback will run every time the Realm is
```

docs/sync.js
```diff
@@ -61,6 +63,44 @@
...
+ * {@link updateCallback} will be rerun every time the Realm is opened (e.g.
```
kraenhansen, line 99:
```suggestion
 * {@link update} callback will be rerun every time the Realm is opened (e.g.
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -193,6 +193,168 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+          ) {
+            const { realm, config } = await openRealm(user, extraConfig);
+
+            expect(realm.subscriptions).to.have.length(1);
+            expect(realm.subscriptions.state).to.equal(Realm.App.Sync.SubscriptionsState.Complete);
+
+            if (closeRealmAfter) {
+              closeRealm(realm, config);
+              return undefined;
+            }
```
kraenhansen, line 284: Might make sense to wrap this in a try-finally block to ensure it happens even if the assertions above fails.

lib/extensions.js
```diff
@@ -268,8 +270,30 @@ module.exports = function (realmConstructor) {
...
+          // in `js_realm.hpp`), we need to return a promise which waits for the
+          // new subscriptions to be fully synchronised, then returns the Realm.
+          if (initialSubscriptions.rerunOnStartup || !realmExists) {
+            return realm.subscriptions.waitForSynchronization().then(() => {
+              return realm;
+            });
+          } else {
+            return realm;
+          }
+        });
```
kraenhansen, line 295: If you made the chained promise callback an async function, you could simplify the logic at the bottom of the function body. Just a thought.

```suggestion
        let openPromise = Promise.race(openPromises).then(async (realm) => {
          const { initialSubscriptions } = config.sync;

          // If `initialSubscriptions` was not specified, return the Realm immediately
          if (!initialSubscriptions) {
            return realm;
          }

          // If an update has been run by C++ (which performs all the validation
          // and runs the actual update function, see `handle_initial_subscriptions`
          // in `js_realm.hpp`), we need to return a promise which waits for the
          // new subscriptions to be fully synchronised, then returns the Realm.
          if (initialSubscriptions.rerunOnStartup || !realmExists) {
            await realm.subscriptions.waitForSynchronization();
          }
          return realm;
        });
```

kraenhansen (approved):
I have only minor comments and suggestions. Great (pragmatic) solution.

takameyer:
>One other idea I did have was:
Make the initialSubscriptions.update callback internally wrapped in subscriptions.update and change its signature to update: (subs: MutableSubscriptionSet, realm: Realm)

I actually was going to suggest this when I read the initial implementation.  I think this is the right way to go, as it also guides the developer towards the intention of the callback.  As it's already in the `update` transaction, they would be less tempted to start doing `realm.write` (is that even allowed within an `update`?).

takameyer (approved):
Nice work!  Impressive test coverage :) Just have a few suggestions in documentation.  Would also be up for discussing the shape of the interface as well.

kneth:
> Implementing the solution purely in JS 

Moving to an async API only (disallow `new Realm()`) will feel more JSish, and I suggest that we consider it. Of course the benefit of a pure JS implementation of this particular feature is en extra bonus.

> [...] make it a more generic solution

In a way it feel awkward to have `onOpen` - in particular if we move to have `Realm.open()` only. And `beforeResolve` is even worse. It might be better to have a number of less generic (more specialized) attributes.

src/js_realm.hpp
tomduncalf: I believe this is already handled by `validated_to_function`, I've added a test for this case which passes (not yet pushed)

tomduncalf:
I modified this so the `update` callback behaves like `subs.update`, except it passes a second `realm` argument to use (as the Realm does not exist at the time of configuring it). It was a slightly fun C++ change, so if someone could review that would be great: https://github.com/realm/realm-js/pull/4568/commits/3ac2b7fd3a78876030da691b4ee28fe98c8268d0

docs/sync.js
tomduncalf: Renamed to `rerunOnOpen`, though will see if other SDKs have a better name

integration-tests/tests/src/tests/sync/flexible.ts
tomduncalf: 👍

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -193,6 +193,168 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+              flexible: true,
+              user,
+              initialSubscriptions,
+            },
+          };
+        }
+
+        describe("error", function () {
+          afterEach(function () {
+            Realm.deleteFile(this.config);
```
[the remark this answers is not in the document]
tomduncalf: All these tests result in an error, so a Realm should never be opened?

types/index.d.ts
tomduncalf: 👍

docs/sync.js
tomduncalf: 👍

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -193,6 +193,168 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+
+            return realm;
+          }
+
+          it("returns a promise", async function () {
+            const result = openRealm(this.user, {});
+            expect(result).to.be.instanceOf(Promise);
+
+            const { realm, config } = await result;
+            closeRealm(realm, config);
```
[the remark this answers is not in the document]
tomduncalf: I suppose not, I'll wrap in try/catch/finally!

CHANGELOG.md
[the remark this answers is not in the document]
tomduncalf: Done, nice idea to include the example

tomduncalf:
Think I've addressed all changes, if someone (@kraenhansen or @kneth maybe) could give final approval when they get a chance. Thanks all!

kneth (approved):

src/js_subscriptions.hpp
```diff
@@ -585,16 +606,16 @@ MutableSubscriptionSetClass<T>::create_instance(ContextType ctx, realm::sync::Mu
...
+ * instance to point to the updated SubscriptionSet.
  *
  * @param ctx JS context
  * @param this_object \ref ObjectType wrapping the SubscriptionSet
- * @param args \ref Arguments structure:
- *   Argument 1: A callback which receives a mutable version of the SubscriptionSet as its
- *     argument, and which updates the SubscriptionSet as required
- *   Argument 2: A callback to be called when the state of the SubscriptionSet is "Complete"
- *      or "Error" after the update has been applied (see `wait_for_synchronization_impl`).
+ * @param args \ref Arguments structure: Argument 1: A callback which receives a
```
kneth, line 614: :shipit: The formatting could be better

tomduncalf: Agreed, I'll update on another PR

PR #4569, opened 2022-05-11
Remove unnecessary steps from GHA workflow

takameyer:
## What, How & Why?
Release publish workflow was taking too long.  The release binaries for Realm-JS were being built, which are not required for releasing @realm/react.  This PR removes these steps from the workflow

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/publish-realm-react.yml
+++ b/.github/workflows/publish-realm-react.yml
@@ -28,7 +28,7 @@ jobs:
         run: npm ci --ignore-scripts
 
       - name: Bootstrap @realm/react
-        run: npm install realm --prefix packages/realm-react
+        run: npm install realm --prefix packages/realm-react --no-save
         env:
           npm_config_realm_local_prebuilds: ${{ github.workspace }}/prebuilds
 
--- a/.github/workflows/publish-realm-react.yml
+++ b/.github/workflows/publish-realm-react.yml
@@ -24,27 +24,8 @@ jobs:
       - name: Install npm v7
         run: npm install -g npm@7
 
-      # Caches
-      - uses: actions/cache@v2
-        name: cached node_modules
-        id: cache-node-modules
-        with:
-          path: node_modules
-          key: node-modules-${{ github.job }}-${{ hashFiles('package-lock.json') }}
-
-      - name: ccache
-        uses: hendrikmuhs/ccache-action@v1
-        with:
-          key: realm-react
-
-      - name: Prepend ccache executables to the PATH
-        run: echo PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH" >> $GITHUB_ENV
-
       - name: Install realm node dependencies
         run: npm ci --ignore-scripts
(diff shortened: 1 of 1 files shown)


kraenhansen:
This does however mean your tests wouldn't be able to pass if they were based on an unreleased version of the `realm` package. Which might actually be desirable, but could lead to difficulties reproducing test failures locally.

takameyer:
>This does however mean your tests wouldn't be able to pass if they were based on an unreleased version of the realm package. Which might actually be desirable, but could lead to difficulties reproducing test failures locally.

@kraenhansen That would actually be desirable.  I would not want to make a release of `@realm/react` that was dependent on a non-released version of `realm`.  Seems like a feature, not a bug.
That being said, PR tests should be based on the `realm` in master.

.github/workflows/publish-realm-react.yml
```diff
@@ -24,28 +24,9 @@ jobs:
...
-        run: echo PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH" >> $GITHUB_ENV
-
       - name: Install realm node dependencies
         run: npm ci --ignore-scripts
 
-      - name: Prebuild release binary
-        run: npx prebuild
-
       - name: Bootstrap @realm/react
         run: npx lerna bootstrap --scope @realm/react --include-dependencies
```
kraenhansen, line 31: I havn't personally tested this, but I'd imagine something like this would be less error prone.

```suggestion
        run: npm install realm --prefix packages/realm-react
```

takameyer: I tested it and it did something strange :D I'll go the traditional route and run `npm install` in the `packages/realm-react` directory

takameyer: Tested, this works just fine :)

takameyer:
>If that's the goal, it shouldn't use lerna bootstrap to install (since that will link realm, including the lib parts but download a pre-built binary which might be incompatible). Instead it should simply npm install possibly even npm install realm instead.

This is a good solution, since we don't have any `*` dependencies in `@realm/react`.

.github/workflows/publish-realm-react.yml
```diff
@@ -24,30 +24,11 @@ jobs:
...
-
       - name: Install realm node dependencies
         run: npm ci --ignore-scripts
 
-      - name: Prebuild release binary
-        run: npx prebuild
-
       - name: Bootstrap @realm/react
-        run: npx lerna bootstrap --scope @realm/react --include-dependencies
+        run: npm install realm --prefix packages/realm-react
```
kraenhansen, line 31: You probably want to throw `--no-save` on as well to prevent changes to the `package.json`.

```suggestion
        run: npm install realm --prefix packages/realm-react --no-save
```

kraenhansen (approved):

PR #4575, opened 2022-05-16
Expose realm ref when using RealmProvider in @realm/react

### LINKED ISSUE #4571: Expose realm ref when using RealmProvider in @realm/react
### Problem

Hello,

Thank you for this amazing service. When opening a `Realm`, one could handle client resets with the following code:

```ts
const handleRealmSyncError: Realm.ErrorCallback = useCallback(
  (_session, error) => {
    if (realmRef.current) {
      if (error.name === 'ClientReset') {
        realmRef.current.close();
        Realm.App.Sync.initiateClientReset(realmApp, realmPath);
        realmRef.current = null;
      }
    }
  },
  [realmPath],
);
```

and opening a Realm with the following sync config:

```ts
sync: {
  user: currentRealmUser,
  error: handleRealmSyncError,
},
```

But with `@realm/react`, we cannot use the `useRealm` hook in the same component where `RealmProvider` is declared, and thus there is no way to close the `Realm` in the error handler.

This enhancement should be a high priority as the production app would simply break in the event of a client reset and users would have to reinstall a newer version of the app with the updated schema.

### Solution

As mentioned by @takameyer, we should provide a ref prop to the RealmProvider and set the value when it’s rendered, so that it can be accessed in the error handler in the same component.

Dealbreaker

takameyer:
## What, How & Why?
It was found that we were missing a critical feature to the `RealmProvider`, which was a way to access the configured realm from outside of the provider.  This is being done through a forwarded ref, which is passed as a property into the provider.
Upon implementation, it was also apparent that this would be a useful addition to `AppProvider` as well.  

This closes #4571

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/README.md
+++ b/packages/realm-react/README.md
@@ -250,6 +250,20 @@ const AppWrapper = () => {
   )
 }
 ```
+
+In some cases, it may be necessary to access the configured Realm from outside of the `RealmProvider`, for instance, implementing a client reset fallback.  This can be done by creating a `ref` with `useRef` and setting the `realmRef` property of `RealmProvider`.
+
+```tsx
+const AppWrapper = () => {
+  const realmRef = useRef<Realm|null>(null)
+
+  return (
+    <RealmProvider realmRef={realmRef}>
+      <App/>
+    <RealmProvider>
+  )
+}
+```
 ### Dynamically Updating a Realm Configuration
 
 It is possible to update the realm configuration by setting props on the `RealmProvider`.  The `RealmProvider` takes props for all possible realm configuration properties.
@@ -290,6 +304,20 @@ const SomeComponent = () => {
 	const app = useApp();
 
 	//...
+}
+```
+
+It is also possible to receive a reference to the app outside of the `AppProvider`, through the `appRef` property.  This must be set to a React reference returned from `useRef`.
+
+```tsx
+const AppWrapper = () => {
+  const appRef = useRef<Realm.App|null>(null)
+
+  return (
+    <AppProvider appRef={appRef}>
+      <App/>
+    <AppProvider>
+  )
 }
 ```
 
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -1,7 +1,8 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Add realmRef property to RealmProvider to access the configured realm outside of the provider component ([#4571](https://github.com/realm/realm-js/issues/4571))
+* Add realmRef property to `RealmProvider` to access the configured realm outside of the provider component ([#4571](https://github.com/realm/realm-js/issues/4571))
+  * Additionally appRef on `AppProvider` was added to provide access to `Realm.App` from outside the provider component
 
 ### Fixed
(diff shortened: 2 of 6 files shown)


packages/realm-react/README.md
```diff
@@ -250,6 +250,20 @@ const AppWrapper = () => {
   )
 }
 ```
+
+In some cases, it may be necessary to access the configured Realm from outside of the `RealmProvider`, for instance, implementing a client reset fallback.  This can be done by creating a `ref` with `useRef` and setting the `realmRef` property of `RealmProvider`.
+
+```tsx
+const AppWrapper = () => {
+  const realmRef = useRef<Realm|null>(null)
+
```
tomduncalf, line 259: Would it be worth showing a simple config here that uses the `realmRef`, just so it's really clear how to use this?

```suggestion
// Handle client reset errors by closing the Realm and initiating a client reset
const handleRealmSyncError: Realm.ErrorCallback = useCallback(
  (_session, error) => {
    if (realmRef.current) {
      if (error.name === 'ClientReset') {
        realmRef.current.close();
        Realm.App.Sync.initiateClientReset(realmApp, realmPath);
      }
    }
  },
  [realmPath],
);
```

tomduncalf: Maybe there is something even simpler we could show though...

tomduncalf (approved):
Nice solution 👍

kraenhansen:
An alternative solution to the original issue would be to pass the realm instance into the callback provided through `config.sync.error`. Right now it takes `session` and `error`, how about exposing the `Realm` instance on the `session` object instead?

tomduncalf:
I wondered about the same @kraenhansen but thought maybe this solution was more general, e.g. if you wanted to access the Realm in your top level component for some other reason. Maybe you're right that we should favour simplicity unless we can find another use case though

takameyer:
>An alternative solution to the original issue would be to pass the realm instance into the callback provided through config.sync.error. Right now it takes session and error, how about exposing the Realm instance on the session object instead? That would be way simpler in my opinion.

@tomduncalf @kraenhansen 
In the linked issue, the user would need access to both the Realm and the Realm.App.  Would you also want to provide that app within the session object as well?  I'm not so familiar with the session object, but I'll take a look at the code. 

I thought this solution would be more flexible, since it would cover other use cases.  Adding realm to `session` would solve only the current issue.  On the other hand, maybe this is the only use case.  Thoughts?

packages/realm-react/README.md
takameyer: I originally did throw that in there, which was what inspired me to add this strategy to the `AppProvider` as well.  I would like to review how a client reset is done and see if I can simplify this.

takameyer: I think I'll leave this as is for now.  I don't want to show deprecated use cases in the documentation.

thespacemanatee:
@takameyer Is it possible to release a beta-6 for hermes builds and bump @realm/react to include this commit? Would be nice to have `ClientReset` functionality as we are gearing up for our first release soon, thank you!

takameyer:
@thespacemanatee I'm releasing `@realm/react` v0.3.1 now.

takameyer:
@thespacemanatee thanks for the heads up.  We will get that taken care of.

PR #4579, opened 2022-05-17
Add calls to flush_ui_queue for Hermes

tomduncalf:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/jsi/jsi_init.hpp
+++ b/src/jsi/jsi_init.hpp
@@ -18,6 +18,14 @@
 
 #pragma once
 
+#include <functional>
+
+namespace realm {
+namespace js {
+std::function<void()> flush_ui_queue;
+} // namespace js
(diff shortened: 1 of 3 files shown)


src/jsi/jsi_init.hpp
```diff
@@ -18,6 +18,14 @@
 
 #pragma once
 
+#include <functional>
+
+namespace realm {
+namespace js {
+std::function<void()> flush_ui_queue;
```
tomduncalf, line 25: This needed to be moved to the header file so that it is visible to `jsi_function.hpp`, I guess this is what I missed and created some convoluted workaround for in the original!

PR #4580, opened 2022-05-18
Type Fix for Returned Schemas

takameyer:
## What, How & Why?
The schema returned from an open realm has a different structure than the input schema
for configuration.  This structure has now been typed along with updated documentation.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [x] jsdoc files updated


types/index.d.ts
```diff
@@ -67,13 +87,21 @@ declare namespace Realm {
...
         primaryKey?: string;
         embedded?: boolean;
+    }
+
+    interface ObjectSchema extends BaseObjectSchema {
         properties: PropertiesTypes;
     }
 
+    interface InternalObjectSchema extends BaseObjectSchema {
+        properties: InternalPropertiesTypes;
```
kraenhansen, line 102:
```suggestion
    interface CanonicalObjectSchema extends BaseObjectSchema {
        properties: CanonicalPropertiesTypes;
```

takameyer:
I'll put this back in draft.   Appears the tests are failing.

integration-tests/tests/src/tests/dynamic-schema-updates.ts
```diff
@@ -72,7 +72,7 @@ describe("realm._updateSchema", () => {
     // Copy the schema
     const updatedSchema = [...realm.schema];
     // Locate the Dog schema
-    const dogSchema = updatedSchema.find((s) => s.name === "Dog");
+    const dogSchema = updatedSchema.find((s) => s.name === "Dog") as Realm.ObjectSchema;
```
takameyer, line 75: Curious if anyone has any thoughts on this.  Since the schema is being dynamically updated, I have cast it to the "input configuration" version of the schema so that it can be modified in this manner.

docs/realm.js
```diff
@@ -499,6 +499,33 @@ class Realm {
  *   name. Queries can be done using both the public and the underlying property name.
  */
 
+/**
+ * The retreived schema from an open Realm has a different structure as the input schema.
```
kneth, line 503:
```suggestion
 * The retreived schema from an open Realm might have a different structure as the input schema.
```

kneth: :shipit: We rarely use the canonical schema notation in examples

integration-tests/tests/src/tests/dynamic-schema-updates.ts
kneth: Yeah, it is a good discussion to take. I mean, the canonical schema notation is mostly to tools and not humans.

kneth (approved):
I suggest that it is merged but it leaves us with a decision to make. Maybe the upcoming TS model classes will have it simpler to make the decision.

docs/realm.js
takameyer: This change would denote that it also might return ObjectSchema?  I want to show that this isn't the case.  `realm.schema[n]` always returns a CanonicalObjectSchema.

integration-tests/tests/src/tests/dynamic-schema-updates.ts
takameyer: I think in this case casting it is the best solution.  This is only for Realm Studio

takameyer: > But shouldn't the CanonicalObjectSchema be assignable to an ObjectSchema?

I do not think that's possible without breaking the property type when reading the schema of an opened realm.  But I am open to be proven wrong.  
The one important difference between the `CanonicalObjectSchema` and the `ObjectSchema` is, that the `properties` are always objects in the `CanonicalObjectSchema`.  If we make `CanonicalObjectSchema` assignable to the `ObjectSchema`, then the `properties` must also possibly be a `string`.  This would make, for example, `realm.schema[0].properties.foo.mapTo` display a type error.  Removing this type error is the point of this PR.
Do you know a better way to type this in a way that accomplishes the goal of the PR?

kraenhansen:
```suggestion
    const dogSchema = updatedSchema.find((s) => s.name === "Dog");
```

kraenhansen (approved):

PR #4588, opened 2022-05-24
Realm js release workflows

### LINKED ISSUE #3763: Automate releasing
Releasing Realm JavaScript is a manual process, and we should be able to more most of it to CI. Pushing a new git tag should initialize the entire release process:

- [ ] Set version (CHANGELOG.md, package.json, xcode project, dependencies.list)
- [ ] Build and upload binaries for all platforms
- [ ] Publish npm package
- [ ] Upload API documentation
- [ ] Notify in #realm-releases
- [x]  ~Close Jira release~

iamer:
## What, How & Why?
Add release preparation github actions workflow

This closes #3763

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -48,3 +48,4 @@ jobs:
           draft: true
           body: An automated PR for next release.
           commit-message: "[${{ steps.update-changelog.outputs.new-version }}] Bump version"
+          token: ${{ secrets.REALM_CI_PAT }}
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,50 @@
+name: Prepare Release
+
+on: workflow_dispatch
+
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      # action is at https://github.com/realm/ci-actions/tree/main/update-changelog
+      - name: Update Changelog
+        id: update-changelog
+        uses: realm/ci-actions/update-changelog@v2
+        with:
+          changelog: ${{ github.workspace }}/CHANGELOG.md
+
+      - name: Update package.json
+        uses: jacobtomlinson/gha-find-replace@0dfd0777cc234ef6947ec1f20873c632114c4167 #! 0.1.4
+        with:
+          find: '"version": "[^"]*"'
+          replace: '"version": "${{ steps.update-changelog.outputs.new-version }}"'
+          include: ${{ github.workspace }}/package.json
+
+      - name: Update package-lock.json
+        run: npm install --ignore-scripts
+
+      - name: Update dependencies.list
+        uses: jacobtomlinson/gha-find-replace@0dfd0777cc234ef6947ec1f20873c632114c4167 #! 0.1.4
+        with:
+          find: 'VERSION=.*'
+          replace: 'VERSION=${{ steps.update-changelog.outputs.new-version }}'
+          include: ${{ github.workspace }}/dependencies.list
+
+      - name: Update xcode project
+        uses: jacobtomlinson/gha-find-replace@0dfd0777cc234ef6947ec1f20873c632114c4167 #! 0.1.4
+        with:
+          find: 'CURRENT_PROJECT_VERSION = [^;]*'
+          replace: 'CURRENT_PROJECT_VERSION = ${{ steps.update-changelog.outputs.new-version }}'
+          include: react-native/ios/RealmReact.xcodeproj/project.pbxproj
+
+      - name: Create Release PR
+        uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad #! 3.10.1
+        with:
+          branch: release/${{ steps.update-changelog.outputs.new-version }}
+          title: Prepare for ${{ steps.update-changelog.outputs.new-version }}
+          draft: true
+          body: An automated PR for next release.
+          commit-message: "[${{ steps.update-changelog.outputs.new-version }}] Bump version"
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -45,7 +45,6 @@ jobs:
         with:
           branch: release/${{ steps.update-changelog.outputs.new-version }}
           title: Prepare for ${{ steps.update-changelog.outputs.new-version }}
-          draft: true
           body: An automated PR for next release.
           commit-message: "[${{ steps.update-changelog.outputs.new-version }}] Bump version"
           token: ${{ secrets.REALM_CI_PAT }}
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -17,11 +17,9 @@ jobs:
           changelog: ${{ github.workspace }}/CHANGELOG.md
 
       - name: Update package.json
-        uses: jacobtomlinson/gha-find-replace@0dfd0777cc234ef6947ec1f20873c632114c4167 #! 0.1.4
-        with:
-          find: '"version": "[^"]*"'
-          replace: '"version": "${{ steps.update-changelog.outputs.new-version }}"'
-          include: ${{ github.workspace }}/package.json
+        run: |
+          jq '.version="${{ steps.update-changelog.outputs.new-version }}"' package.json > package.tmp
+          mv package.tmp package.json
 
       - name: Update package-lock.json
         run: npm install --ignore-scripts
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,114 @@
+name: Publish Release
+on: workflow_dispatch
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: Setup node version
+      uses: actions/setup-node@v2
+      with:
+        node-version: '14'
+        cache: 'npm'
+        cache-dependency-path: '**/package-lock.json'
+
+    - name: Setup NPM version
+      run: npm install -g npm@7
+
+    - name: Get NPM cache directory
+      id: npm-cache-dir
+      run: echo "::set-output name=dir::$(npm config get cache)"
+
+    - name: Restore NPM cache
+      id: npm-cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+
+    - name: Install node modules
+      run: npm ci --ignore-scripts
+
+    - name: Download prebuild PR artifacts
+      uses: dawidd6/action-download-artifact@d0f291cf39bd21965ea9c4c6e210fc355c3844ed
+      with:
+        workflow: pr-realm-js.yml
+        commit: ${{ github.sha }}
+        path: ${{ github.workspace }}
+        workflow_conclusion: completed
+
+    - name: Read version
+      id: get-version
+      run: |
+        pkgVersion=$(npm pkg get version)
+        echo "::set-output name=version::$pkgVersion"
+      shell: bash
+
+    - name: Npm Publish realm-js-${{ steps.get-version.outputs.version }}
+      run: npm publish ${{ steps.get-version.outputs.version }}
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+    - name: Find Release PR
+      uses: juliangruber/find-pull-request-action@f9f7484f8237cf8485e5ab826e542ba5dd9e9c6e
+      id: find-pull-request
+      with:
+        branch: ${{ github.ref }}
+
+    - name: Merge Pull Request
+      uses: juliangruber/merge-pull-request-action@8a13f2645ad8b6ada32f829b2fae9c0955a5265d
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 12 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Update Changelog
+        id: update-changelog
+        uses: realm/ci-actions/update-changelog@v2
```
takameyer, line 14: Where is this action located?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,116 @@
...
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+
+    - name: Install node modules
+      run: npm ci --ignore-scripts
+
+    - name: Download prebuild PR artifacts
```
takameyer, line 40: Will this also obtain the `.xcframework` and Android build artifacts?

takameyer (commented):
Seems that this is headed in the right direction, but we will need to test.  I have add some questions and suggestions.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,53 @@
...
+          changelog: ${{ github.workspace }}/CHANGELOG.md
+
+      - name: Update package.json
+        uses: jacobtomlinson/gha-find-replace@0dfd0777cc234ef6947ec1f20873c632114c4167 #! 0.1.4
+        with:
+          find: '"version": "[^"]*"'
+          replace: '"version": "${{ steps.update-changelog.outputs.new-version }}"'
+          include: ${{ github.workspace }}/package.json
+
+      - name: Update package-lock.json
```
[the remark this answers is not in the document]
iamer: OK will do that instead of modifying the lock file

.github/workflows/publish-release.yml
iamer: once all builds in the PR workflow are enabled then it should contain them.

takameyer: Ok, just wanted to make sure.  The example you linked only contains the `node` binaries

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,116 @@
...
+
+    - name: Create vNext PR
+      id: vnext-pr
+      uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad
+      with:
+        branch: prepare-vnext
+        title: Prepare for vNext
+        body: Update Changelog for vNext
+        delete-branch: true
+        base: main
```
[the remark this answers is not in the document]
iamer: Somehow I missed the base: main, thanks for pointing it out.
It makes more sense to let the action use the current commit in case the release is being created from a branch (when support is added for that)?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,114 @@
...
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        number: ${{ steps.find-pull-request.outputs.number }}
+        method: squash
+
+    - name: Extract Changelog
+      run: scripts/extract-changelog.sh ${{ steps.get-version.outputs.version }} > ExtractedChangelog.md
+
+    - name: Create Github Release
+      uses: ncipollo/release-action@10c84d509b28aae3903113151bfd832314964f2e
+      with:
```
kneth, line 78: If the npm dist-tag is not `latest`, we would like to set `prerelease` to true.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,50 @@
...
+          find: 'CURRENT_PROJECT_VERSION = [^;]*'
+          replace: 'CURRENT_PROJECT_VERSION = ${{ steps.update-changelog.outputs.new-version }}'
+          include: react-native/ios/RealmReact.xcodeproj/project.pbxproj
+
+      - name: Create Release PR
+        uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad #! 3.10.1
+        with:
+          branch: release/${{ steps.update-changelog.outputs.new-version }}
+          title: Prepare for ${{ steps.update-changelog.outputs.new-version }}
+          draft: true
```
nirinchev, line 48: I'd suggest passing `draft: false` here.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: jacobtomlinson/gha-find-replace@0dfd0777cc234ef6947ec1f20873c632114c4167 #! 0.1.4
+        with:
+          find: 'CURRENT_PROJECT_VERSION = [^;]*'
+          replace: 'CURRENT_PROJECT_VERSION = ${{ steps.update-changelog.outputs.new-version }}'
+          include: react-native/ios/RealmReact.xcodeproj/project.pbxproj
+
+      - name: Create Release PR
+        uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad #! 3.10.1
+        with:
+          branch: release/${{ steps.update-changelog.outputs.new-version }}
```
nirinchev, line 46: Can we also add `token: ${{ secrets.REALM_CI_PAT }}` to the arguments? That way the newly created PR will trigger a PR build.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    - name: Read version
+      id: get-version
+      run: |
+        pkgVersion=$(npm pkg get version)
+        echo "::set-output name=version::$pkgVersion"
+      shell: bash
+
+    - name: Npm Publish realm-js-${{ steps.get-version.outputs.version }}
+      run: npm publish ${{ steps.get-version.outputs.version }}
```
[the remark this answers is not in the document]
nirinchev: I suggest that we provide the tag manually as a workflow input. That can default to `latest` and we can override it if necessary.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,114 @@
...
+        bodyFile: ExtractedChangelog.md
+        name: ${{ steps.get-version.outputs.version }}
+        tag: v${{ steps.get-version.outputs.version }}
+        token: ${{ secrets.GITHUB_TOKEN }}
+        draft: false
+
+    - name: Upload Github Release artifacts
+      run: npm prebuild --upload ${{ secrets.GITHUB_TOKEN }}
+
+    - name: 'Post to #realm-releases'
```
nirinchev, line 88: That should probably be the last step in the workflow as it's not unlikely that it fails. The `release-to-slack` action will attempt to convert markdown to the slack post format, but that's not always flawless, especially when more complex markdown is used.

nirinchev (commented):
The publish release workflow doesn't appear to be uploading the API docs to S3 - can we add that as well?

.github/workflows/publish-release.yml
amtep: Hi I'm assisting iamer on this PR.

Instead of moving the job around, would it be better to add `continue-on-error: true` here? That way the workflow retains its logical order, and also the whole release won't be marked as "failed" just because it wasn't posted on slack.

amtep:
@cla-bot check

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,143 @@
...
+    - name: Npm Publish realm-js-${{ steps.get-version.outputs.version }}
+      run: npm publish --tag '${{ inputs.tag}}' realm-js@${{ steps.get-version.outputs.version }}
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+    - name: Find Release PR
+      uses: juliangruber/find-pull-request-action@f9f7484f8237cf8485e5ab826e542ba5dd9e9c6e
+      id: find-pull-request
+      with:
+        branch: ${{ github.ref }}
```
takameyer, line 78: Curious how this step works.  How does it find the correct PR?  What is `github.ref` and how is it set?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,143 @@
...
+      uses: ncipollo/release-action@10c84d509b28aae3903113151bfd832314964f2e
+      with:
+        bodyFile: ExtractedChangelog.md
+        name: ${{ steps.get-version.outputs.version }}
+        tag: v${{ steps.get-version.outputs.version }}
+        token: ${{ secrets.GITHUB_TOKEN }}
+        draft: false
+        prerelease: ${{ steps.get-version.outputs.prerelease }}
+
+    - name: Prepare Android and iOS artifacts for uploading
```
takameyer, line 100: Not sure if this step is necessary.  The artifacts are published in the npm release

takameyer: And in any case, you would not want to pull the contents of the .xcframework out of that folder.  That should all stay together

takameyer (approved):
I'd like to get this merged.  I do have some questions, but I don't think this holds back this PR

.github/workflows/publish-release.yml
amtep: > Not sure if this step is necessary. The artifacts are published in the npm release

People keep telling me that, but I don't see them :) Which files exactly do you want in the Release page on github? Without this step it's these four:

realm-v10.20.0-napi-v5-darwin-x64.tar.gz
realm-v10.20.0-napi-v5-linux-x64.tar.gz
realm-v10.20.0-napi-v5-win32-ia32.tar.gz
realm-v10.20.0-napi-v5-win32-x64.tar.gz

.github/workflows/publish-release.yml
amtep: The idea is that you run the publish-release step on the branch that was made for the release PR. So github.ref is just the ref for that branch.

.github/workflows/publish-release.yml
amtep: Since the kneth-pr-workflow has been publishing releases, you can find an example here https://github.com/realm/realm-js/releases/tag/v10.20.0

.github/workflows/publish-release.yml
amtep: When you choose "publish release" from the list of manually triggered actions, you get a dropdown choice of which branch to run it on. There you can select the release branch.

github.ref is set by github itself at the start of every workflow, and it's the git ref (= commit or tag or branch or whatever) that triggered the workflow

takameyer: Oh great, that sounds like a good workflow.  Thanks for the clarification

.github/workflows/publish-release.yml
takameyer: > People keep telling me that, but I don't see them :) 

If you make an empty node project and type `npm install realm`, the release binaries will be located in:
* `node_modules/realm/react-native/ios/realm-js-ios.xcframework/`
* `node_modules/realm/react-native/android/src/main/jniLibs/`
You can also verify this by doing `npm pack`, which will output a `.tgz.`.  This should also contain those files.

> Which files exactly do you want in the Release page on github?

I would probably just zip up the entire `react-native` directory.  But i'm not really sure what value any of that provides.  Our users are typically not downloading the artifacts from github.  They are using `npm install`.

.github/workflows/publish-release.yml
takameyer: I'm happy with `continue-on-error: true`

.github/workflows/publish-release.yml
amtep: Okay :) Would it be better to have no release artifacts on the github Release page? That would mean we can cut a couple of steps from the workflow.

The source archive is always there by default.

takameyer: Yes, you can cut those steps out.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,48 @@
+name: Prepare Release
+
+on: workflow_dispatch
```
takameyer, line 3: Can we also add a prefix to the release?  We sometimes do `-beta` or `-rc` releases, such as:
`11.0.0-rc.1`

amtep: Hmm there is currently support for a "tag" in the Publish Release step. It gets passed along in the npm publish step. It defaults to "latest", and if it's not "latest", then the release is marked as a prerelease on github.

Would it make sense to use that tag in the version as well? Both in the github release and in the git tag for the release. But not in the version passed to npm publish, since it already has "tag" there.

Then I would also make it so that it doesn't merge the PR if the tag is not "latest', because you probably want to use the same branch for a real release later.

amtep: Okay :)
I can add a "prerelease" variable next to "tag", default empty. And then handle the version logic based on that.

I would guess the 11.0.0 prereleases are made from a different branch than master? Since presumably they contain breaking changes.  (We also still need to handle the case of releasing from a base branch other than master)

So I still think it's best if the prerelease string is set in the "publish release" step, so that you can do one "prepare release" that creates a release branch, then optionally multiple prerelease "publish release" steps from there (which don't merge the branch into master), then a final "publish release" when the release is ready.

But if you want it the other way around, where each prerelease gets its own release branch named with the prerelease tag and a new one is made for every prerelease, let me know.

Btw, if you want to merge this PR soon then I can optionally work on the prerelease feature as a new PR. Since it's kind of orthogonal to the current PR.

takameyer: Yes, let's do this at a later time.  One can always change the PR prepared by the workflow anyway.

templates/scripts/publish_templates.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+templates_location="$(dirname $(readlink -e $0))/.."
```
takameyer, line 4: Might be a good idea to keep this an explicit array of directories.  There are more than just templates within the `templates` directory (`scripts`, `docs`)

PR #4594, opened 2022-05-25
Working with Github Actions

kneth:
## What, How & Why?

Documenting Github Actions and workflows


contrib/working-with-github-actions.md
```diff
@@ -0,0 +1,12 @@
...
+We are using [Github Actions](https://github.com/features/actions) to automate building, testing and releasing.
+
+You find the workflows in the "Actions" menu.
+
+![GithubActions](assets/github-actions-menu.png)
+
+
+## Delete apps and clusters
+
+In most cases our test suites will delete Atlas clusters and apps but not always. If you need to delete them manually, you can use the workflow "Wipe all clusters and apps".
```
nirinchev, line 12: Might be worth noting that this workflow doesn't attempt to distinguish between "active" and "leftover" apps so will happily delete clusters and apps that are currently being used by PR builds.

nirinchev (approved):

tomduncalf (approved):
Good idea 👍

contrib/working-with-github-actions.md
kneth: Good point. I will add a comment about that.

PR #4600, opened 2022-05-26
Switch to building with Xcode 13

tgoyne:
In preparation for dropping Xcode 12 support in core. I think I found all of the places where Xcode versions are mentioned.


kneth (approved):
Thank you!

PR #4612, opened 2022-06-02
Updated cmake config to read arch from `NODE_ARCH`

kraenhansen:
## What, How & Why?

This fixes cmake configuration to respect the `NODE_ARCH`, making it compile for arm64 on an darwin-arm64 by default.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually - locally)


package.json
```diff
@@ -58,7 +58,8 @@
     "build": "npm run build-changes",
     "build:arm": "npm run build-changes:arm",
     "build-changes": "cmake-js build --debug",
-    "build-changes:arm": "NODE_ARCH_ARM=1 cmake-js build --debug",
+    "build-changes:x86": "cmake-js build --debug --arch x86",
```
kneth, line 61: Shouldn't it be `x86_64`?

kraenhansen: Good catch!

kneth (approved):

PR #4613, opened 2022-06-02
Cleaning up: Remove old, unused scripts

### LINKED ISSUE #4402: Security vulnerabilities in dependencies in request and https-proxy-agent

All the time

### Description

npm dependencies request and https-proxy-agent have security vulnerabilities. Are there any plans to update the dependency to a version that does not have vulnerability or to change the dependency to an alternative one?

We are using realm as a local mobile database only on react native. Could you also advise if the security vulnerabilities are relevant in the context of a local mobile database?

### request
Vulnerability: sonatype-2017-0655
CVSS Score: 5.9
Component: request : 2.88.2
Fix Available: No recommended versions are available for the current component

The request package is vulnerable to Weak Authentication Algorithm. The OAuth.prototype.buildBodyHash function in oauth.js uses SHA-1 for authentication which is no longer considered cryptographically secure. With enough resources an attacker might be able to crack the authentication mechanism and cause security attacks.

### https-proxy-agent
Vulnerability: sonatype-2019-0419
CVSS Score: 5.9
Component: https-proxy-agent : 2.2.4
Fix Available: https-proxy-agent : 3.0.0

The https-proxy-agent package is vulnerable to Man-in-the-Middle. The ondata and onsocket functions in index.js does not upgrade TLS connections like it normally does when connecting to a proxy that doesn't return a 200 response to the initial CONNECT request. Consequently, a MitM listening to this could potentially view sensitive data that would have otherwise been hidden by TLS.

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

Run vulnerability scan (Nexus IQ) to produce the following vulnerabilities:
sonatype-2017-0655
sonatype-2019-0419

### Version

10.13.0

Local Database only

### Are you using encryption?

Yes, using encryption

React Native

### Build environment

### Cocoapods version

_No response_

kneth:
## What, How & Why?

This closes #4402 

We have a number of unused scripts, and dependencies related to the scripts. 

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


takameyer (approved):
I have never used these scripts.  I think its safe to remove them :)

PR #4628, opened 2022-06-08
Upgrade to NDK r23c

### LINKED ISSUE #3905: Upgrade to Android NDK r23
* [x] Upgrade NDK to r23 (`Dockerfile.android`)
* [x] Document version (`README.md` and `contrib/building.md`)
* [x] Enable LTO

~We will have to wait for [https://github.com/android/ndk/issues/1560](https://github.com/android/ndk/issues/1560) be resolved.~

See also https://github.com/actions/virtual-environments/issues/5595

kneth:
## What, How & Why?

This closes #3905 

It simplifies building by using the same CMake and Ninja installation to build for all supported platforms (see `contrib/building.md` for details).

**Note**: Using latest version of Ninja with Github Actions leads to a fatal error. An older version works just fine. I observed the same error as in https://github.com/ninja-build/ninja/issues/2048.

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


contrib/building.md
```diff
@@ -78,8 +78,8 @@ npm install
 #### iOS
 
 ```sh
-# Install cocoapods and cmake
-brew install cocoapods cmake
+# Install cocoapods
```
takameyer, line 81:
```suggestion
# Install cmake
```

armhf.Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get install -y \
         ninja-build \
         nodejs \
         libnode-dev:armhf \
-        npm \
+        ninja-build \
```
takameyer, line 16:
```suggestion
```

🚢 This is repeated above

contrib/building.md
takameyer: 🚢 comment is not reflecting the script

takameyer (approved):
👏🏼 That doesn't look trivial!  Just a couple things.  Other than that, LGTM

tomduncalf (approved):
Looks good as far as I can tell

scripts/build-android.js
```diff
@@ -87,7 +84,7 @@ for (const arch of architectures) {
     "-GNinja",
     `-DANDROID_NDK=${ndkPath}`,
     `-DANDROID_ABI=${arch}`,
-    `-DCMAKE_MAKE_PROGRAM=${sdkPath}/cmake/${cmakeVersion}/bin/ninja`,
```
kraenhansen, line 90: What's the reason we can rely on the built-in `ninja` and we don't have to use the one provided by the Android SDK?

armhf.Dockerfile
```diff
@@ -22,15 +22,12 @@ ENV NVM_DIR /tmp/.nvm
...
-RUN cd /opt \
-    && curl -O -J https://cmake.org/files/v3.15/cmake-3.15.2-Linux-x86_64.tar.gz \
-    && tar zxf cmake-3.15.2-Linux-x86_64.tar.gz
-
-ENV PATH "/opt/cmake-3.15.2-Linux-x86_64/bin:$PATH"
\ No newline at end of file
+RUN mkdir -p /home/jenkins/cmake && \
+    cd /home/jenkins/cmake && \
+    curl -O -J https://cmake.org/files/v3.23/cmake-3.23.2-linux-x86_64.tar.gz &&\
+    tar zxf cmake-3.23.2-linux-x86_64.tar.gz
```
kraenhansen, line 32: Is there a lower bound to the version of `cmake` we should be using?
If so, we should maby assert that in the build-android.js script?

kraenhansen:
My only concern with this, is that as we're shipping a pre-built binary, we need to ensure we're compiling with the same C++ standard library as the developer building their app. It's my understanding that the ABI of the standard library is provided by the NDK, is that correct? If so, how do we ensure these are matching out endusers? (this relates to the comment in the description of #4027)

scripts/build-android.js
kneth: Yes, this way you only have to install ninja once (currently you have to install it twice for local development: node.js + Android)

armhf.Dockerfile
kneth: It is handled by the root `CMakeList.txt` (`cmake_minimum_required`) but I can see there is a mismatch.

Issue #4350, opened 2022-02-16
iOS - Can't find variable: globalThis


cristianoccazinsp:
### Description

Using `10.20.0-alpha.2` (but without actually using Hermes on iOS), reported the above crash for a user right on app startup.

### Stacktrace & log output

```shell
ReferenceError
Can't find variable: globalThis

    node_modules/realm-network-transport/dist/bundle.react-native.es.js:91:43 
    node_modules/metro-runtime/src/polyfills/require.js:340:6 loadModuleImplementation
    node_modules/realm/lib/extensions.js:60:45 exports
    node_modules/realm/lib/react-native.js:45:24 
    node_modules/metro-runtime/src/polyfills/require.js:340:6 loadModuleImplementation
    src/db/index.js:8 
    node_modules/metro-runtime/src/polyfills/require.js:340:6 loadModuleImplementation
    src/App.js:31 
    node_modules/metro-runtime/src/polyfills/require.js:340:6 loadModuleImplementation
    node_modules/metro-runtime/src/polyfills/require.js:200:45 guardedLoadModule
    index.js:86:29 
    node_modules/regenerator-runtime/runtime.js:63:44 tryCatch
    node_modules/regenerator-runtime/runtime.js:294:30 
    node_modules/regenerator-runtime/runtime.js:63:44 tryCatch
    node_modules/regenerator-runtime/runtime.js:155:28 invoke
    node_modules/regenerator-runtime/runtime.js:165:19 
    node_modules/promise/setimmediate/core.js:37:14 tryCallOne
    node_modules/promise/setimmediate/core.js:123:25 
    node_modules/react-native/Libraries/Core/Timers/JSTimers.js:248:17 
    node_modules/react-native/Libraries/Core/Timers/JSTimers.js:112:6 _callTimer
    node_modules/react-native/Libraries/Core/Timers/JSTimers.js:166:15 _callReactNativeMicrotasksPass
    node_modules/react-native/Libraries/Core/Timers/JSTimers.js:418:11 callReactNativeMicrotasks
    node_modules/react-native/Libraries/BatchedBridge/MessageQueue.js:393:11 value
    node_modules/react-native/Libraries/BatchedBridge/MessageQueue.js:135:11 
    node_modules/react-native/Libraries/BatchedBridge/MessageQueue.js:370:8 value
    node_modules/react-native/Libraries/BatchedBridge/MessageQueue.js:134:17 value
    [native code] value
    [native code] value
```
```


[...]

PR #4647, opened 2022-06-14
Fix other uses of globalThis breaking iOS 11/12 support

tomduncalf:
## What, How & Why?

https://github.com/realm/realm-js/issues/4350

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/lib/mongo-client.js
+++ b/lib/mongo-client.js
@@ -21,6 +21,8 @@ const { DefaultNetworkTransport } = require("realm-network-transport");
 
 const { cleanArguments, getEnvironment } = require("./utils");
 
+const globalThisOrWindow = typeof globalThis === "object" ? globalThis : window;
+
 /**
  * Iterates the global to ensure all globals needed for watching are available.
  */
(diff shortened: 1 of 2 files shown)


lib/mongo-client.js
```diff
@@ -21,6 +21,8 @@ const { DefaultNetworkTransport } = require("realm-network-transport");
 
 const { cleanArguments, getEnvironment } = require("./utils");
 
+const globalThisOrWindow = typeof globalThis === "object" ? globalThis : window;
```
takameyer, line 24: This might be an issue.  Does `mongo-client` run in React Native?

tomduncalf: Hm yeah I kind of did this fix on autopilot. I wonder what the correct version of this for RN is.

tomduncalf:
Based on Andrew's comment I thought it was worth using a proper polyfill and shared code, closing in favour of https://github.com/realm/realm-js/pull/4653

PR #4648, opened 2022-06-15
Andrew/app importer refactor

takameyer:
## What, How & Why?
Replacing the usage of `realm-cli` with REST calls.  This was due intermittent failures with the `realm-cli`, which were hard to determine the cause.  With this change, it is easier to determine what failed and at which stage of the import.

There are also cleanups and improvements to the integration tests:
* removal of unused dependencies
* unique database names per deployed application

I attempted to keep the integration test fixes to this [PR](https://github.com/realm/realm-js/pull/4736), but some parts were too dependent on the refactoring efforts here.

Both PRs should be merged to determine what is left to do.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


.github/workflows/integration-tests.yml
```diff
@@ -6,6 +6,7 @@ env:
 
 on:
   pull_request:
+    types: [ labeled ]
```
kraenhansen, line 9: Why was this added?

PR #4653, opened 2022-06-16
Implement polyfill for globalThis

### LINKED ISSUE #4350: iOS - Can't find variable: globalThis

tomduncalf:
## What, How & Why?

This closes #4350. I decided to replace the existing inline implementations with a more robust polyfill taken from [core-js](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/global.js), and to put it in `realm.io/common` so we don't need to always repeat the same snippet

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


CHANGELOG.md
```diff
@@ -181,6 +181,7 @@ const Person = {
 * Logging out too quickly can cause an error if the timeout behavior is set to `openLocalRealm` ([#4453](https://github.com/realm/realm-js/issues/4453), since v10.0.0)
 * Released `realm-network-transport` to adopt the changes published to fix `globalThis` undefined issue for older devices. ([#4350](https://github.com/realm/realm-js/issues/4350), since v10.0.0)
 * Fixed flexible sync crash when updating subscriptions after token expiry. ([#4421](https://github.com/realm/realm-js/issues/4421), since v10.12.0)
+* Fixed remaining uses of `globalThis` undefined issue, causing Realm to not load on iOS 11/12. ([#4350](https://github.com/realm/realm-js/issues/4350))
```
kneth, line 184: :shipit:  Do we know if it has always been an issue - or was the bug introduced in a specific version?

kneth (approved):

kraenhansen (commented):
I'm wondering if we could simplify the solution, but simply documenting the use of the [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env#shippedproposals) Babel preset?

tomduncalf:
> I'm wondering if we could simplify the solution, but simply documenting the use of the [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env#shippedproposals) Babel preset?

As in we say to users that they need to use `@babel/preset-env` in their project if they are on an old iOS version?

kraenhansen:
> As in we say to users that they need to use @babel/preset-env in their project if they are on an old iOS version?

Yeah. Or perhaps https://www.npmjs.com/package/babel-plugin-transform-globalthis?

kraenhansen:
Regardless, if we choose to discourages the explicit usage of `globalThis`, I think this PR should ensure `eslint` errors on that.

tomduncalf:
I was following the pattern of what we've already done for `globalThis` in a couple of places, which is to fix our library so that it works with all supported OS versions. I _think_ I prefer that approach (we do the nasty stuff so users don't have to do any workarounds) but I am also happy if the majority would prefer to ask users to install a Babel plugin for this reasonably edge case.

@kneth @takameyer any thoughts?

Good suggestion re: the lint rule

kraenhansen:
> we do the nasty stuff so users don't have to do any workarounds

I agree - and ideally we would apply this in a build step before publishing our package for consumption.
I'd expect this to be easier to implement in the future as I plan on adding rollup to the package as part of the binding generator project.

tomduncalf:
>we would apply this in a build step before publishing our package for consumption.
Got it, this would be the ideal (we use the Babel plugin on our code before publish). In the interim until we have the infra in place to do that, are you OK with this solution?

takameyer:
> any thoughts?
I think we should do the work for our users here.  If it's possible that we implement the babel plugin in our own codebase, I would prefer this.

takameyer (approved):
LGTM

kneth:
> any thoughts?

I agree with @takameyer

tomduncalf:
@tomduncalf Plan: merge this PR, rebase v11 against master, rebase v11 PR to remove cherrypick commit and merge

tomduncalf:
If everyone is happy with this, I'll publish the new common version and rerun CI

kraenhansen:
> I added the eslint rule in all public facing packages: b92a7f8

I think it should be enough to add it to the root package and have the other packages inherit?

tomduncalf:
> I think it should be enough to add it to the root package and have the other packages inherit?

Ah I didn't realist that they inherited! Even better, I'll update

Issue #4657, opened 2022-06-17
Problem After Recently Updating Realm JS


s1tony:
I have been successfully developing a React Native application using Realm and specifically Flexible Sync.  Things were working until recently, after updating to Realm JavaScript v10.19.1.

I am receiving the following message back when attempting to call subscriptions.update:

TypeError: undefined is not an object (evaluating '_$$_REQUIRE(_dependencyMap[0], "@realm.io/common").symbols.PROXY_TARGET')

I am developing on a Windows 10 box for the Android platform.

Any ideas?

hachther:
The issue here is the version of @realm.io/common. 
The version installed by realm is "0.1.1" but the needed version is "0.1.2". 

See below I fixed my issue:
```
yarn add @realm.io/common
rm -rf node_modules/realm/node_modules/@realm.io
```

tomduncalf:
Hi @s1tony @hachther, thanks for the report. 

It looks like we missed updating the dependency in the package.json, sorry for the trouble. @hachther's workaround should help temporarily.

I will look to get a fixed version published as soon as possible and will update this issue when I have done so. I will also look into why our automated tests didn't catch this issue!

tomduncalf:
Based on my testing, a clean install of `realm` should pull in the right version of `@realm.io/common` (0.1.2), as it is published on npm and our `package.json` specifies `^0.1.1`... but it seems that if you already have common 0.1.1 installed, npm won't upgrade it, so if you are upgrading Realm you can hit this.

Therefore @s1tony a workaround for now should be just to delete your `node_modules` and re-install, which should pull in `@realm.io/common@0.1.2` – no need to manually install it.

Regardless we should fix this in our package.json, which I'm doing now.

s1tony:
Thank you so much.

PR #4660, opened 2022-06-20
Update realm.io/common to 0.1.2

tomduncalf:
## What, How & Why?

This fixes https://github.com/realm/realm-js/issues/4657.

I'm going to look separately if I can create a failing test for this cas.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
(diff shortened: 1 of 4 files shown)


takameyer (approved):
This seems to be a common problem when we update our internal packages.  We should set something up to catch this for us (or automate an update for dependent packages)

kraenhansen:
I'd expect the package-lock.json to update from this change?

tomduncalf:
Yep @takameyer, I'll create a ticket for that as it is very easy to miss (and quite hard to catch as our `package.json` specifies `^0.1.1` so in many cases will pull in the right version for a clean install regardless, but seemingly not for an upgrade install).

@kraenhansen Good point, I'll do an npm/lerna to update them

tomduncalf:
Added package lock in https://github.com/realm/realm-js/pull/4660/commits/fb436c6f9c37764334ae32a8464d362b4c5db7bc

kraenhansen (approved):

Issue #4667, opened 2022-06-21
Binary mirroring not working since node-pre-gyp to prebuild-install migration


Fiouz:
### How frequently does the bug occur?
All the time

### Description
# Context
* Closed network environment (intranet), where build machines have no IP routing to the internet (and especially https://static.realm.io)
* Build machines have IP routing to an HTTP mirroring machine (e.g. Sonatype Nexus / JFrog Artifactory), so they can still access content from public repositories, by using internal URLs (as long as mirroring rules have been provisioned) such as https://corporate.mirror.lan/repository/static.realm.io
* Build machines NOT provisioned for native binary compilation
* Build machines configured to redirect realm binary download HTTP request to the internal mirror `npm config set realm_binary_host_mirror https://corporate.mirror.lan/repository/static.realm.io`, according to https://github.com/mapbox/node-pre-gyp/blob/v1.0.9/README.md#download-binary-files-from-a-mirror (for older versions of realm) / https://github.com/prebuild/prebuild-install/blob/v7.1.1/README.md#custom-binaries (for newer versions of realm)

# Command
```
npm --loglevel=verbose install realm@10.20.0-beta.5
```

# Expected behavior
`npm install` should success, and since our build machines are not able to compile native components, success would imply that the binary was successfully downloaded from the configured HTTP mirror.

Since the mirror setting between `node-pre-gyp` and `prebuild-install` has the same spelling, we expect a working mirroring for older versions to also work for newer versions of realm.

See also: #2501

# Actual behavior
When `prebuild-install` constructs the mirror URL, the URL has a different shape compared to the "original" URL, making it unable to download anything from the mirror.

Generated URL for mirroring:

```
<mirror_prefix>/v10.20.0-beta.5/realm-v10.20.0-beta.5-napi-v4-win32-x64.tar.gz
```

Original URL (without mirroring):

```
<original_prefix>/realm-js-prebuilds/10.20.0-beta.5/realm-v10.20.0-beta.5-napi-v4-win32-x64.tar.gz
```

(URL obtained by manually executing the `prebuild-install --verbose` command from the `node_modules/realm` directory)

This only affects latest versions, previous versions that relied on `node-pre-gyp` (e.g. `6.1.8`) are properly mirrored, i.e. the following works fine with binary mirroring enabled:

```
npm --loglevel=verbose install realm@6.1.8
```

### Stacktrace & log output
```shell
npm verb stack Error: command failed
[...]

PR #4672, opened 2022-06-24
Update package.json to discard tag_prefix to make npm binary mirroring work

tomduncalf:
## What, How & Why?

A user reported in https://github.com/realm/realm-js/issues/4667 that they cannot install Realm using npm "binary mirroring", which allows it to work in an environment with no internet connection and instead, an internal artefact server – I guess quite a common set up in finance, etc. This happened since v10, it used to work in v6.

They did some [research](https://github.com/realm/realm-js/issues/4667#issuecomment-1163201076) and suggested that this change would fix the issue. They tested this branch and it did indeed fix the issue, but I have to confess I don't really understand the consequences of the change so would welcome any thoughts!

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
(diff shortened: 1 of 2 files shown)


kneth (approved):
LGTM. I wonder if the experience will be better if we migrate to Github releases instead of using S3?

PR #4674, opened 2022-06-24
Fix toJSON behaviour for Dictionary on JSC pre-v11 (fixes #4658)

### LINKED ISSUE #4658: Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication)

All the time

### Description

@kneth 

Hello guys!

As the title above, I'm facing the error after installing `realm-js`:   </br> `Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication)`

This error was caused because has more than one react native version in my project. 
After check `yarn.lock`, I see that `@realm/react@^0.3.1` has installed `react-native@>=0.59` => This is causes error

Why `@realm/react@^0.3.1` does install it?

And now, I can run my project until I remove the `@realm/react` package. (This means `react-native@>=0.59` has been removed)

Pls, help me, How to resolve this case?

Screen Shot 2022-06-19 at 11 47 21 PM
Screen Shot 2022-06-19 at 11 49 13 PM

### Stacktrace & log output

```shell
ERROR  TypeError: null is not an object (evaluating 'endpoint.startsWith')
 ERROR  Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication). A frequent cause of the error is that the application entry file path is incorrect.
      This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native.
 LOG  redux-persist/stateReconciler: rehydrated keys 'app, user, _persist'
 ERROR  Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication). A frequent cause of the error is that the application entry file path is incorrect.
      This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native.
```

Yes, always

### Reproduction Steps

_No response_

### Version

```
"realm": "^10.19.1"
"@realm/react": "^0.3.1",
```

Local Database only

### Are you using encryption?

No, not using encryption

```
MacOS: 12.4 (21F79) 
iPhone OS: 15.4.1
```

### Build environment

```
System:
    OS: macOS 12.4
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 273.85 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.1 - ~/.nvm/versions/node/v16.14.1/bin/node
    Yarn: 1.18.0 - /usr/local/bin/yarn
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.1/bin/npm
    Watchman: 2022.03.21.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /Users/mac/.rvm/gems/ruby-2.7.4/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.2, iOS 15.2, macOS 12.1, tvOS 15.2, watchOS 8.3
    Android SDK:
      API Levels: 23, 28, 29, 30, 31, 32
      Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 32.0.0
      System Images: android-27 | Google Play Intel x86 Atom, android-29 | Intel x86 Atom_64, android-30 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7935034
    Xcode: 13.2.1/13C100 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.11 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.67.2 => 0.67.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

### Cocoapods version

CocoaPods: 1.11.3

tomduncalf:
## What, How & Why?

This fixes the behaviour of `toJSON` on objects with Dictionary fields on JSC, reported in https://github.com/realm/realm-js/issues/4658. 

Before this fix, Dictionary fields would incorrectly not be converted to JSON, resulting in an exception if a field of the Dictionary of the JSONified object was modified, as it would try to write to the Realm outside of a write transaction.

The root cause of this is similar to https://github.com/realm/realm-js/pull/4541, in that Proxy objects do not seem to be properly "transparent" on JSC – in this case, `instanceof Realm.Dictionary` would return `false` for a proxied Dictionary, whereas it should return `true` (as JS `Proxy` objects are supposed to be totally transparent). This should be fixed with JSI/Hermes.

I searched the codebase for anywhere else that might be affected by this bug and couldn't see anywhere.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -206,6 +206,30 @@ describe("Dictionary", () => {
     });
   });
 
+  describe("toJSON", function () {
+    openRealmBefore({
+      schema: [
+        {
+          name: "Item",
+          properties: { dict: "{}" },
+        },
+      ],
+    });
+
+    it("calling toJSON on an object with a Dictionary field removes the Proxy from the Dictionary", function (this: RealmContext) {
+      const object = this.realm.write(() => {
+        return this.realm.create<Item>("Item", { dict: { something: "test" } });
+      });
+      const jsonObject = object.toJSON();
+
+      // Previously this would throw on JSC, because the Dictionary was still a Proxy,
+      // so modifying it tried to write to the Realm outside of a write transaction
+      expect(() => {
+        jsonObject.dict.something = "test2";
+      }).to.not.throw();
+    });
+  });
+
   type ValueGenerator = (realm: Realm) => DictValues;
 
   type TypedDictionarySuite = {
--- a/lib/dictionary.js
+++ b/lib/dictionary.js
@@ -16,8 +16,15 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+const { symbols } = require("@realm.io/common");
+
 const dictionaryHandler = {
   get(target, key) {
+    // Allows us to detect if this is a proxied Dictionary on JSC pre-v11. See realm-common/symbols.ts for details.
+    if (key === symbols.IS_PROXIED_DICTIONARY) {
+      return true;
+    }
+
     if (key === "toJSON") {
       return function () {
         const keys = target._keys();
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -16,6 +16,8 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+const { symbols } = require("@realm.io/common");
+
 let getOwnPropertyDescriptors =
   Object.getOwnPropertyDescriptors ||
   function (obj) {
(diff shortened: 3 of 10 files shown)


takameyer:
@tomduncalf Looks like we need to publish `@realm/common` in order for the tests to run

takameyer:
@tomduncalf oh yeah...that is a bit annoying.  And the change by itself in a PR doesn't really bring much, as it's very dependent on the other code being written.  We will just have to wait for that wonderful world where `realm` is also a `package`.

takameyer (approved):
LGTM

kneth (approved):

PR #4681, opened 2022-06-27
Not compatible with RN 0.69

kneth:


tomduncalf:
I'm not sure if this applies to `master`, or just `v11` – I'll be looking into it a bit more so maybe let's hold off merging for now? LGTM otherwise though

tomduncalf:
I was able to run 0.69 against both v11 and Hermes so we don’t need this, my bad!

takameyer:
@tomduncalf What about `10.19.3`?  That's not working, right?

tomduncalf:
Sorry typo, I meant `v11` and `master`. Master worked with just upgrading my project's dependency. v11 I had to recompile with the new RN version installed so it does require a new release

PR #4686, opened 2022-06-29
Fix android test app failing to build

tomduncalf:
## What, How & Why?

I have no idea why, but the app recently started failing to build for me with the error

```
Execution failed for task ':app:checkDebugAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
   > The minCompileSdk (31) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-30).
     Dependency: androidx.appcompat:appcompat:1.4.1.
     AAR metadata file: /Users/tom.duncalf/.gradle/caches/transforms-3/d3d5b4e7c41854b612980f161edd44cd/transformed/appcompat-1.4.1/META-INF/com/android/build/gradle/aar-metadata.properties.
```

This fixes it, and matches React Native's SDK compatibility 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
+++ b/integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
@@ -15,7 +15,8 @@
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustResize"
+        android:exported="true">
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Internal
 * Fixed analytics tests to reflect the fact that framework versions are read from node_modules, not package.json. ([#4687](https://github.com/realm/realm-js/pull/4687))
 * Adding response type checking to the realm-app-importer. ([#4688](https://github.com/realm/realm-js/pull/4688))
-* <Either mention core version or upgrade>
-* <Using Realm Core vX.Y.Z>
-* <Upgraded Realm Core from vX.Y.Z to vA.B.C>
+* Updated integration test app to target Android SDK 31 ([#4686](https://github.com/realm/realm-js/pull/4686))
 
 10.19.3 Release notes (2022-6-27)
 =============================================================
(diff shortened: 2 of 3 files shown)


integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,8 @@
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustResize"
+        android:exported="true">
```
kneth, line 19: Isn't this the fix? See https://developer.android.com/about/versions/12/behavior-changes-12#exported

kneth (approved):

integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
tomduncalf: Ah yes sorry it should be false, good spot

tomduncalf: Except... the app will not start with it set to false 🙈

takameyer:
Seems it isn't building in the tests.

takameyer:
Maybe we should also update RN 🙈

tomduncalf:
Yeah I have a branch with updated RN actually, I can PR that

integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
tomduncalf: @kneth I believe we need this to be `true`, because
>If the app component includes the [LAUNCHER](https://developer.android.com/reference/android/content/Intent#CATEGORY_LAUNCHER) category, set android:exported to true
and we do include that category: https://github.com/realm/realm-js/blob/e2919ec573e410e5fa630b79669856230939dfd2/integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml#L22

The app fails to start with `false` in any case

kneth: My point is that this is more important than the `compileSdkVersion`.

tomduncalf: Oh I see, I had to upgrade the `compileSdkVersion` first to get to this error. Not sure why no one else was affected!

PR #4697, opened 2022-07-01
Td/debug android cpp in android studio

tomduncalf:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
```diff
@@ -9,7 +9,8 @@
       android:icon="@mipmap/ic_launcher"
       android:roundIcon="@mipmap/ic_launcher_round"
       android:allowBackup="false"
-      android:theme="@style/AppTheme">
+      android:theme="@style/AppTheme"
+      android:usesCleartextTraffic="true">
```
fronck, line 13: [Non-blocker]  Is this change intentional?  Do we need to use non-HTTS traffic?  ([reference](https://developer.android.com/guide/topics/manifest/application-element) (massive webpage without anchors :( ))

fronck (approved):
LGTM 👍

integration-tests/environments/react-native/android/app/src/main/AndroidManifest.xml
tomduncalf: I’ll double check that it’s needed though

integration-tests/environments/react-native/android/app/build.gradle
```diff
@@ -161,6 +161,11 @@ android {
                     abiFilters nativeArchitectures.split(',')
                 }
             }
+
+            // Do not strip debug symbols from the Realm library
```
takameyer, line 165: In general, can we add notes to anything custom we have added to the `react-native` build files?  This will be most helpful in upgrading `react-native` so that we don't accidentally delete something unintentionally.

takameyer (approved):
Amazing work!  Bravo! 🤖

integration-tests/environments/react-native/android/app/build.gradle
tomduncalf: Yeah I think this is a good idea, perhaps we could prefix the comments with `REALM-JS` or something, so that we can clearly spot our customised bits?

takameyer: Sounds good.  Don't forgot the other code blocks as well 🙂

tomduncalf: I don't think we need it there, as the `react-native/build.gradle` is something we've created ourselves to build Realm on Android, rather than a customised RN file?

tomduncalf:
I realised I missed a couple of pre-requisite steps in the docs btw in case anyone is trying this based on the old instructions

integration-tests/environments/react-native/android/app/build.gradle
```diff
@@ -186,6 +191,13 @@ android {
 
         }
     }
+
+    // Add the Realm source files to the Android Studio project so that we can add breakpoints
```
takameyer, line 195:
```suggestion
    // REALM_JS Add the Realm source files to the Android Studio project so that we can add breakpoints
```

tomduncalf: Ah sorry got ya

kneth (commented):
Great to document this as it has never been easy to debug C++ code on device

PR #4700, opened 2022-07-04
Closing connections after calling `return` on `AsyncIntegrator` returned from `watch`

kraenhansen:
## What, How & Why?

This closes an issue, where calling `return` on the `AsyncIterator` returned by `watch` wouldn't free up the network connection.

The existing code was assuming that `return` would be called on the underlying iterator of a generator function, which is not explicit in the specification. The fix is to get a reference to the response body iterator and explicitly call `return` on this, while deferring all other implementation to a renamed `watchImpl` method.

The syntax involved in unwrapping and rewrapping the iterator is a bit involved, since `watch` must return synchronously while the response body iterator is obtained through a promise.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests (manually, locally)

### DIFF
--- a/packages/realm-web/CHANGELOG.md
+++ b/packages/realm-web/CHANGELOG.md
@@ -8,7 +8,7 @@
 * None
 
 ### Fixed
-* None
+* Fixed an issue where closing a watch stream, by calling `return` on the `AsyncIterator` wouldn't propagate to closing the underlying network connection. ([#4700](https://github.com/realm/realm-js/pull/4700))
 
 ### Internal
 * None
--- a/packages/realm-web/CHANGELOG.md
+++ b/packages/realm-web/CHANGELOG.md
@@ -8,7 +8,7 @@
 * None
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


packages/realm-web/src/index.ts
```diff
@@ -20,6 +20,8 @@ export * as BSON from "bson";
 
 import { App } from "./App";
 
+console.log("Hello from realm-web");
+
```
fronck, line 24:
```suggestion
```

fronck (approved):
LGTM 👍

kneth (approved):
:lgtm:

packages/realm-web/CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 * None
 
 ### Fixed
-* None
+* Fixed an issue where closing a watch stream, by calling `return` on the `AsyncIterator` wouldn't propagate to closing the underlying network connection. ([#4700](https://github.com/realm/realm-js/pull/4700))
```
tomduncalf, line 11:
```suggestion
* Fixed an issue where closing a watch stream, by calling `return` on the `AsyncIterator`, wouldn't propagate to closing the underlying network connection. ([#4700](https://github.com/realm/realm-js/pull/4700))
```

tomduncalf (approved):

PR #4703, opened 2022-07-04
Fix types and add a test for flexible sync client reset 

tomduncalf:
## What, How & Why?

I wanted to confirm this was working after a question from the docs team. It turns out our TS definition to only allow "manual" mode was a bit incorrect so I fixed it, and added a simple test which replicates the PBS one in the old tests

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


tomduncalf:
Merging this as it is a recreated #4702 (created to test out if it was my code making Catalyst CI fail somehow, but it seems to just be pure chance)

PR #4710, opened 2022-07-08
Fix missing use of protected_ctx causing crash when initialSubscriptions throws an error on Android

tomduncalf:
## What, How & Why?

This was working on iOS but crashing on Android, presumably the lifetime of the `ctx` is slightly different between the two platforms

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


takameyer (approved):
Ship it!

kneth (approved):

PR #4717, opened 2022-07-11
Update `building.md` to most recent steps

kraenhansen:
## What, How & Why?

When installing the build chain for Android on my M1 mac, I found that the [AdoptOpenJDK](https://github.com/AdoptOpenJDK/homebrew-openjdk) cask had been deprecated in favour of `temurin`. I also found that I needed to manually install the `cli-tools` to get the `sdkmanager` CLI.

### DIFF
--- a/contrib/building.md
+++ b/contrib/building.md
@@ -84,13 +84,12 @@ brew install cocoapods cmake
 
 #### Android
 
-First, install OpenJDK 8 (later versions are not compatible with the project and with Android's `sdkmanager` tool):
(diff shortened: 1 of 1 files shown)


tomduncalf (approved):
While it might not be relevant for too long, do you know if this still works on Intel mac also?

kraenhansen:
> Do you know if this still works on Intel mac also?

I'd assume so. I don't think `temurin` is M1 dependent.

PR #4730, opened 2022-07-15
Add default logger for tests

tomduncalf:
## What, How & Why?

This adds a default logger for the integration tests, as we currently get no output on Android as it forwards `stdout` to `/dev/null`. The log is colourised to help with reading it.

image

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/hooks/import-app-before.ts
+++ b/integration-tests/tests/src/hooks/import-app-before.ts
@@ -17,6 +17,8 @@
 ////////////////////////////////////////////////////////////////////////////
 import { deleteApp, importApp, TemplateReplacements } from "../utils/import-app";
 
+const REALM_LOG_LEVELS = ["all", "trace", "debug", "detail", "info", "warn", "error", "fatal", "off"];
+
 export function importAppBefore(
   name: string,
   replacements?: TemplateReplacements,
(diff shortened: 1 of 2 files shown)


takameyer (approved):
:shipit: The colors are great.  Good idea!

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -30,6 +32,14 @@ export function importAppBefore(
     } else {
       this.app = await importApp(name, replacements);
       Realm.App.Sync.setLogLevel(this.app, logLevel);
+      // Set a default logger as Android does not forward stdout
+      Realm.App.Sync.setLogger(this.app, (level, message) => {
+        const magentaDate = `\x1b[35m${new Date().toISOString().replace("T", " ").replace("Z", "")}`;
```
kraenhansen, line 37: The full date timestamps feel a bit excessive to me. Perhaps an HH:MM:SS.ms would be enough?

Nit: I'd personally split this in two lines, I don't like debugging expressions inside of template strings (same goes for the line below).

kraenhansen (approved):

integration-tests/tests/src/hooks/import-app-before.ts
tomduncalf: Yea fair point maybe the data isn't that useful!

PR #4742, opened 2022-07-28
Fix for integration tests with skip_runner true on iOS

tomduncalf:
Currently we try to adb reverse ports which fails, we only need do this for android

## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


takameyer (approved):
Makes sense.  Good catch!  :shipit:

kraenhansen:
Woups. Thanks 😊

PR #4764, opened 2022-08-10
Fix ccache on M1

takameyer:
Homebrew installs ccache in a different directory on the M1.
This checks if ccache exists in the path rather then a specific location.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/scripts/ccache-clang++.sh
+++ b/scripts/ccache-clang++.sh
@@ -1,11 +1,10 @@
 #!/bin/sh
-if type -p /usr/local/bin/ccache >/dev/null 2>&1; then
+if command -v ccache &> /dev/null; then
   export CCACHE_MAXSIZE=10G
   export CCACHE_CPP2=true
   export CCACHE_HARDLINK=true
   export CCACHE_SLOPPINESS=file_macro,time_macros,include_file_mtime,include_file_ctime,file_stat_matches
-  exec /usr/local/bin/ccache /usr/bin/clang++ "$@"
+  exec ccache clang++ "$@"
 else
   exec clang++ "$@"
 fi
-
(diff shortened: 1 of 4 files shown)


scripts/ccache-clang++.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/sh
-if type -p /usr/local/bin/ccache >/dev/null 2>&1; then
+if command -v ccache &> /dev/null; then
   export CCACHE_MAXSIZE=10G
   export CCACHE_CPP2=true
   export CCACHE_HARDLINK=true
   export CCACHE_SLOPPINESS=file_macro,time_macros,include_file_mtime,include_file_ctime,file_stat_matches
-  exec /usr/local/bin/ccache /usr/bin/clang++ "$@"
+  exec ccache /usr/bin/clang++ "$@"
```
kraenhansen, line 7:
```suggestion
  exec ccache clang++ "$@"
```

I wonder if clang++ could also just be read from the path?

scripts/ccache-clang++.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/sh
-if type -p /usr/local/bin/ccache >/dev/null 2>&1; then
+if command -v ccache &> /dev/null; then
```
kraenhansen, line 2: Is this Linux compatible?

takameyer: It's POSIX, so it should be

scripts/ccache-clang++.sh
takameyer: As long as it's in the path...which it should be.

kraenhansen (approved):

PR #4769, opened 2022-08-11
Remove values array from `Realm.Object#create`

kraenhansen:
## What, How & Why?

This feature was never documented and it depends on the order of properties in the schema, which feels like an implementation detail to me.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,7 @@
 * File format: generates Realms with format v23 (reads and upgrades file format v5 or later for non-synced Realm, upgrades file format v10 or later for synced Realms).
 
 ### Internal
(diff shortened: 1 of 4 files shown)


kneth (approved):
Please update `CHANGELOG.md`

kneth (approved):
My approval is based on the assumption that you will update the tests before merging

kraenhansen:
Closing this, since I added this behaviour behind a flag to the v12 of the SDK: https://github.com/realm/realm-js/blob/main/packages/realm/src/flags.ts#L20-L23 and we can remove this entirely in v13.

PR #4779, opened 2022-08-15
Fix end-to-end tests

takameyer:
It would seem the assumption that waitForSynchronization would
allow the latest data to propogate is no longer valid.
We now require a call to downloadAllServerChanges before we make assertions.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/tests/sync/flexible.ts
+++ b/integration-tests/tests/src/tests/sync/flexible.ts
@@ -1391,7 +1391,7 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
         return realm.create<IPerson>(FlexiblePersonSchema.name, { _id: new BSON.ObjectId(), name: "Tom", age: 36 });
       });
 
-      await realm.subscriptions.waitForSynchronization();
+      await realm?.syncSession?.uploadAllLocalChanges();
 
       return { person, id: person._id };
     }
@@ -1425,6 +1425,9 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
       expect(newRealm.objectForPrimaryKey(FlexiblePersonSchema.name, id)).to.be.undefined;
 
       await newRealm.subscriptions.update((mutableSubs) => subsUpdateFn(mutableSubs, newRealm));
+
+      // This was added to ensure all server changes are downloaded before any tests assertions
+      await newRealm?.syncSession?.downloadAllServerChanges();
 
       return { id, newRealm };
     }
@@ -1480,6 +1483,8 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
         mutableSubs.add(newRealm.objects(FlexiblePersonSchema.name).filtered("age > 30"));
       });
 
+      await newRealm?.syncSession?.uploadAllLocalChanges();
+
       newRealm.addListener("change", () => {
         expect(newRealm.objectForPrimaryKey(FlexiblePersonSchema.name, id)).to.not.be.undefined;
       });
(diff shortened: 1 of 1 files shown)


kneth (approved):
LGTM

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1391,7 +1391,7 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
         return realm.create<IPerson>(FlexiblePersonSchema.name, { _id: new BSON.ObjectId(), name: "Tom", age: 36 });
       });
 
-      await realm.subscriptions.waitForSynchronization();
+      await realm?.syncSession?.uploadAllLocalChanges();
```
tomduncalf, line 1394: This makes sense. I guess we should rename the function `addPersonAndWaitForUpload`, and change the comment to `Add a Person object and wait for the change to be uploaded`

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1426,6 +1426,9 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
 
       await newRealm.subscriptions.update((mutableSubs) => subsUpdateFn(mutableSubs, newRealm));
 
+      // This was added to ensure all server changes are downloaded before any tests assertions
+      await newRealm?.syncSession?.downloadAllServerChanges();
```
tomduncalf, line 1430: I'm surprised that this one is necessary as I thought `waitForSynchronization` (which `await .subscriptions.update` calls) should ensure all the matching objects are downloaded – or is that not the case?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1480,6 +1483,8 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
         mutableSubs.add(newRealm.objects(FlexiblePersonSchema.name).filtered("age > 30"));
       });
 
+      await newRealm?.syncSession?.uploadAllLocalChanges();
```
tomduncalf, line 1486: This one does not make sense to me, as we have not created any local objects prior to calling it (except in `addPersonAndResyncWithSubscription`, which you have fixed) – we've only added a subscription, which should be handled by `await subs.update`.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1501,6 +1506,8 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
```
tomduncalf, line 1509: Again I don't think this one should be necessary as we didn't create any local objects prior to calling it

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: That's a good change 👍🏼

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: This is actually not needed 👍🏼

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: It fails without this :(

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: Removed

PR #4785, opened 2022-08-17
Tomasz/object key

tmarzec:
## What, How & Why?
This adds the functionality of retrieving Object Keys from realm objects and retrieving objects by their Object Keys.
It is needed by [the realm/realm-flipper-plugin](https://github.com/realm/realm-flipper-plugin).

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [x] 🚦 Tests


kraenhansen (approved):
Looks legit 👍

fealebenpae (approved):
💯

src/js_realm_object.hpp
```diff
@@ -295,6 +297,22 @@ std::vector<String<T>> RealmObjectClass<T>::get_property_names(ContextType ctx,
     return names;
 }
 
+template <typename T>
```
takameyer, line 300: Would be nice to have a comment here explaining the arguments.
For example: https://github.com/realm/realm-js/blob/master/src/js_realm.hpp#L827
We have decided as a team that any new functions should come with comment block headers.

takameyer (commented):
Post merge suggestion :)

PR #4819, opened 2022-08-23
Added exit on missing emulator

kraenhansen:
## What, How & Why?

Adds a check on the error message from retrieval of the app process id, to test if the emulator has gone away and exit the runner if it did.

I verified that this indeed exists when terminating the emulator while running locally.


takameyer (approved):
Works.  LGTM

PR #4826, opened 2022-08-23
Fixed an offset error in Node buffer handling

### LINKED ISSUE #3781: NAPI DataView and TypedArray conversions ignore byteOffset
https://github.com/realm/realm-js/blob/f2fd5e1956b9a6448bf410adeb92557dc5be64a6/src/node/node_buffer.hpp#L48-L60

That code should be adding the byte offset of the view to the start address of the ArrayBuffer. Instead, it is just returning the start address with no offsetting.

fronck:
## What, How & Why?

Fix an error on our type conversion of Node's binary buffers where the target type's `byteOffset` wouldn't be respected.
Note:  this code is never called at present. 

This closes #3781.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kneth (approved):
LGTM

PR #4830, opened 2022-08-25
make bindgen generate into sdk package

RedBeard0531:
This is my attempt to simplify getting the binding generated into the sdk package. There should be no workflow changes, but we can add them if we want to. I verified that the SDK still compiles with `tsc` and that I can still run my mini tests, but I haven't tried running the SDK tests yet.


packages/bindgen/src/templates/typescript.ts
```diff
@@ -143,18 +143,19 @@ export function generateTypeScript({ spec: rawSpec, file }: TemplateContext): vo
...
 
-  const js = file("native.js", "eslint");
+  const js = file("native.mjs", "eslint");
   js("// This file is generated: Update the spec instead of editing this file directly");
   js(`
-    import bindings from 'bindings';
-    const nativeModule = bindings("realm.node")
+    import { createRequire } from 'node:module';
+    const require = createRequire(import.meta.url);
+    const nativeModule = require("./realm.node");
```
RedBeard0531, line 151: I couldn't tell if it would be better to hide the fact that this is a `require()` function from the bundlers or not. Either way, easy enough to change.

kraenhansen (approved):
Looking good.

PR #4840, opened 2022-08-30
Match types for User#callFunction implementation across platforms

### LINKED ISSUE #3733: Types for `User#callFunction` doesn't match implementation across platforms

takameyer:
## What, How & Why?
The unifies the call signature of `User#callFunction` across all documentation
and platforms.

This closes #3733 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/lib/user.js
+++ b/lib/user.js
@@ -31,7 +31,7 @@ const instanceMethods = {
   },
 
   async callFunction(name, ...args) {
-    this._callFunctionOnService(name, undefined, ...args);
+    return this._callFunctionOnService(name, undefined, ...args);
   },
 
   async _callFunctionOnService(name, serviceName, ...args) {
@@ -110,7 +110,7 @@ const instanceMethods = {
         get(target, name, receiver) {
           if (typeof name === "string" && name != "inspect") {
             return (...args) => {
-              return user._callFunctionOnService(name, serviceName, args);
+              return user._callFunctionOnService(name, serviceName, ...args);
             };
           } else {
             return Reflect.get(target, name, receiver);
--- a/lib/user.js
+++ b/lib/user.js
@@ -30,10 +30,14 @@ const instanceMethods = {
     return promisify((cb) => this._logOut(cb));
   },
 
-  async callFunction(name, args = [], service = undefined) {
+  async callFunction(name, ...args) {
+    this._callFunctionOnService(name, undefined, ...args);
+  },
+
+  async _callFunctionOnService(name, serviceName, ...args) {
     const cleanedArgs = cleanArguments(args);
     const stringifiedArgs = EJSON.stringify(cleanedArgs, { relaxed: false });
-    const result = await promisify((cb) => this._callFunction(name, stringifiedArgs, service, cb));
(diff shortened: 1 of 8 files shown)


kneth (approved):
LGTM. But if it is only documentation changes, it is a breaking change?

types/index.d.ts
```diff
@@ -36,7 +36,7 @@ declare namespace Realm {
      * PropertyType
      * @see { @link https://realm.io/docs/javascript/latest/api/Realm.html#~PropertyType }
      */
-    type PropertyType = string | 'bool' | 'int' | 'float' | 'double' | 'decimal128' | 'objectId' | 'string' | 'data' | 'date' | 'list' | 'linkingObjects';
+    type PropertyType = 'bool' | 'int' | 'float' | 'double' | 'decimal128' | 'objectId' | 'string' | 'data' | 'date' | 'list' | 'linkingObjects';
```
takameyer, line 39: Fix this

takameyer:
@kneth Went over this with @kraenhansen.  This is a breaking change,  as it changes the `callFunction` implementation in `realm-js`:
```
    user.callFunction("sum", 1, 2, 3); // Valid
    user.callFunction("sum", [1, 2, 3]); // Invalid
```

lib/user.js
```diff
@@ -98,15 +102,15 @@ const instanceMethods = {
...
+  _functionsOnService(serviceName) {
     const user = this;
     return new Proxy(
       {},
       {
         get(target, name, receiver) {
           if (typeof name === "string" && name != "inspect") {
             return (...args) => {
-              return user.callFunction(name, args, service);
+              return user._callFunctionOnService(name, serviceName, args);
```
kraenhansen, line 113: I think these arguments should be spread into the call.

```suggestion
              return user._callFunctionOnService(name, serviceName, ...args);
```

takameyer: Good catch.  We also needed to return the call to  `_callFunctionOnService`.  Jenkins tests should pass now

PR #4863, opened 2022-09-05
Statically cache the String value of g_internal_field for a small performance gain

tomduncalf:
## What, How & Why?
This optimisation yields a small (~5% with Hermes disabled, ~2% with Hermes enabled) performance improvement with our benchmark suite, as it avoids the overhead of repeatedly creating the JSI String object.

image

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/jsi/jsi_class.hpp
+++ b/src/jsi/jsi_class.hpp
@@ -240,6 +240,10 @@ public:
     // Also, may need to suppress destruction.
     inline static std::optional<JsiFunc> s_ctor;
 
+    // Cache the JSI String instance representing the field name of the internal
+    // C++ object for the lifetime of the current env, as this is a hot path
+    inline static std::optional<fbjsi::String> s_js_internal_field_name;
+
     /**
      * @brief callback for invalid access to index setters
      * Throws an error when a users attemps to write to an index on a type that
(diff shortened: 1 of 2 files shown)


src/jsi/jsi_class.hpp
```diff
@@ -490,7 +490,9 @@ class ObjectWrap {
 
     static Internal* get_internal(JsiEnv env, const JsiObj& object)
     {
-        auto internal = object->getProperty(env, g_internal_field);
+        static const auto js_internal_field = fbjsi::String::createFromAscii(env, g_internal_field);
```
RedBeard0531, line 493: This will break on reload since it will continue to use the string from the old env.

tomduncalf: Ah right yeah, good point. I guess there's no way round that? It's not a huge gain anyway so I'm happy to close this if that's best

RedBeard0531: Make it a class-static like s_ctor with the same logic to set it on module load and reset on em teardown. This isnt a perfect solution because it assumes at most one env at a time, but it will at least work as well as the rest of the codebase.

tomduncalf: Updated in https://github.com/realm/realm-js/pull/4863/commits/e32fc5b82711ace47849baefcb0c836b48404653

tomduncalf:
Waiting on https://github.com/realm/realm-js/pull/4869 for lint to pass

kraenhansen (approved):

src/jsi/jsi_class.hpp
```diff
@@ -490,7 +495,11 @@ class ObjectWrap {
 
     static Internal* get_internal(JsiEnv env, const JsiObj& object)
     {
-        auto internal = object->getProperty(env, g_internal_field);
+        if (!s_js_internal_field_name) {
```
RedBeard0531, line 498:
```suggestion
        if (REALM_UNLIKELY(!s_js_internal_field_name)) {
```

I doubt is makes much difference, but might help some. And the whole point of this is perf, so might as well squeeze out as much as we can.

RedBeard0531 (approved):

tomduncalf:
Ignoring known failure on Android

PR #4869, opened 2022-09-05
Update clang-format npm package version to match system clangd on newer Macs, and update code style to match

tomduncalf:
## What, How & Why?

The default `clangd` installed on newer Macs formats empty classes differently. As the system version is used by the VS Code plugin, but the version used for the lint job is fixed (it is packaged in the `clang-format` npm package), this results in lint errors if you rely on your editor to format the code on newer Macs.

Our [config](https://github.com/realm/realm-js/blob/master/.clang-format#L47-L49) would seem to be setup such that these empty classes should in fact be split but perhaps some new option has been added that we haven't set. Given that it's not a huge deal, and also that the config structure is pretty complex so figuring it out could take a while, I thought the pragmatic solution was to update the `clang-format` package to match the newer `clangd`.

⚠️  This may require you to update your `clangd` version (e.g. via Homebrew and add it to your path) if you are on an older Mac.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kneth (approved):
I like the way you have formatted these empty classes.

takameyer (approved):
LGTM :shipit:

PR #4871, opened 2022-09-05
WIP using JSI's NativeState class for faster access to native C++ object when Hermes enabled

tomduncalf:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/jsi/jsi_class.hpp
+++ b/src/jsi/jsi_class.hpp
@@ -170,6 +170,20 @@ public:
     T obj;
 };
 
+// Class extending JSI's NativeState, used to store the native C++ object associated
+// with a JS object when Hermes is being used. See set_internal for more info.
+template <typename T>
+class WrappedState : public fbjsi::NativeState {
+public:
+    template <typename... Args, typename = std::enable_if_t<std::is_constructible_v<T, Args...>>>
+    WrappedState(Args&&... args)
+        : obj(std::forward<Args>(args)...)
+    {
+    }
+
+    T obj;
+};
+
 template <typename T>
 inline T& unwrap(Wrapper<T>& wrapper)
 {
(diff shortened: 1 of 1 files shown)


tomduncalf:
@kraenhansen @RedBeard0531 I wanted to open this WIP to start a discussion around supporting the new [NativeState API](https://github.com/facebook/hermes/blob/main/API/jsi/jsi/jsi.h#L147) in Hermes master, which provides a way to store a C++ object on a JSI object and allows us to have much faster access to the native object when Hermes is enabled. It is not currently in an RN release, so the build will fail (you need to build RN from source to make it work).

It seems to work (just running the performance tests) but I am sure I'll have missed something as this stuff is complex! 

The main gotcha here is that the NativeState API is not implemented for the [JSCRuntime](https://github.com/facebook/react-native/blob/main/ReactCommon/jsi/JSCRuntime.cpp#L916) and instead throws a `logic_error`, so I am catching that and falling back to the old property method, as I couldn't find a nicer way to detect if Hermes in enabled.

Another gotcha is this will tie us to a minimum of whatever RN version includes this API. Not sure if that is an issue or if there's a good way to work around that if so.

tomduncalf:
OK adding the try/catch makes this like 20x slower with JSC (compared to just having the "catch" code there), lol, I guess that won't fly... maybe we could store a static flag `native_state_supported`, initially set to `true`, then if we hit a `std::logic_error` when trying to use it we set the flag false (and in future skip trying). Any better ideas @RedBeard0531?

tomduncalf:
I pushed a rough version of what I described above but I'll leave the comment intact in case the history gets confusing! https://github.com/realm/realm-js/pull/4871/commits/2a8ccebbab638c7ad36e18b595aea53f7bd7fb94

PR #4876, opened 2022-09-07
Stable roundtrip tests

kraenhansen:
## What, How & Why?

This awaits the existence of an object instead of using `objectForPrimaryKey` in the roundtrip tests.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/sync/mixed.ts
+++ b/integration-tests/tests/src/tests/sync/mixed.ts
@@ -123,8 +123,16 @@ function describeRoundtrip({
     it("reads", async function (this: RealmContext) {
       await setupTest(this.realm);
 
-      const obj = this.realm.objectForPrimaryKey<MixedClass>("MixedClass", this._id);
-      if (!obj) throw new Error("Object not found");
+      const obj = await new Promise<MixedClass>((resolve) => {
+        this.realm
+          .objects<MixedClass>("MixedClass")
+          .filtered("_id = $0", this._id)
+          .addListener(([obj]) => {
+            if (obj) {
+              resolve(obj);
+            }
+          });
+      });
 
       expect(typeof obj).equals("object");
       // Test the single value
--- a/integration-tests/tests/src/utils/sleep.ts
+++ b/integration-tests/tests/src/utils/sleep.ts
@@ -1,0 +1,27 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2022 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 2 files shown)


integration-tests/tests/src/utils/sleep.ts
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+declare function setTimeout(cb: () => void, ms: number): void;
+
+export function sleep(ms = 1000): Promise<void> {
```
kraenhansen, line 21: This was useful when debugging. I've written it a few times before, so this time I decided to just leave it in for the future.

takameyer (approved):
LGTM :shipit:

integration-tests/tests/src/utils/sleep.ts
kneth: Yes, we all write this when debugging and remove it before committing. It's good to have - thanks.

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -123,8 +123,16 @@ function describeRoundtrip({
     it("reads", async function (this: RealmContext) {
       await setupTest(this.realm);
 
-      const obj = this.realm.objectForPrimaryKey<MixedClass>("MixedClass", this._id);
-      if (!obj) throw new Error("Object not found");
+      const obj = await new Promise<MixedClass>((resolve) => {
```
kneth, line 126: It can wait until we need it, but a generic method `waitForObject<T>(realm, id): Promise<T>` could be useful.

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -123,8 +123,16 @@ function describeRoundtrip({
...
+        this.realm
+          .objects<MixedClass>("MixedClass")
+          .filtered("_id = $0", this._id)
+          .addListener(([obj]) => {
+            if (obj) {
+              resolve(obj);
```
kneth, line 132: Are we fine with not removing the listener?

kneth (approved):

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: Sure. When the Realm gets closed after the test (the `openRealmBefore` hook does that) the listeners wont fire again. And even if they did, they could only resolve a promise that already got resolved, which is a no-op.

tomduncalf (approved):

PR #4881, opened 2022-09-07
Refactored manual transaction tests

kraenhansen:
## What, How & Why?

This refactors manual transaction tests.

The first test didn't really test anything other than roll back when `cancelTransaction` gets called, so I renamed it. The other actually verifies something that I think is unexpected (creating an invalid object throws but leaves an object in a strange state), which I noted with a TODO. We should create an issue to track fixing that.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests


kneth (approved):

kraenhansen:
The failure on Android seems unrelated to the change and instead related to a faulty AVD cache.

PR #4887, opened 2022-09-08
Update prepare-release GHA workflow to override version

kraenhansen:
## What, How & Why?

This use the newly merged feature of the `update-changelog` action to allow us to provide an override of the version when preparing a new version.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,6 +1,12 @@
 name: Prepare Release
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Use this to provide a version, instead of deriving it from the changelog.
+        required: false
+        type: string
 
 jobs:
   prepare-release:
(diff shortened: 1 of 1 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -12,8 +18,9 @@ jobs:
       # action is at https://github.com/realm/ci-actions/tree/main/update-changelog
       - name: Update Changelog
         id: update-changelog
-        uses: realm/ci-actions/update-changelog@v2
+        uses: realm/ci-actions/update-changelog@master
```
tomduncalf, line 21: Is this a deliberate change, i.e. we would rather use the master version of actions rather than released ones?

tomduncalf (approved):

.github/workflows/prepare-release.yml
kraenhansen: I asked @nirinchev - it seems that's what other repositories are doing (and I didn't feel like pushing a new tag to the ci-actions repo 🙃 )

takameyer (approved):

PR #4888, opened 2022-09-08
[CI] Use a more specific cache key for vcpkg

fealebenpae:
If Core picks up a newer vcpkg submodule or changes its vcpkg manifest we need to invalidate any existing vcpkg caches.


kneth (approved):
Thank you!

PR #4890, opened 2022-09-08
[CI] Use a unshallow fetch on the vcpkg submodule

fealebenpae:
We need this because we build an older OpenSSL version and this requires the full vcpkg submodule history.


kneth (approved):
Tested on https://github.com/realm/realm-js/pull/4882

PR #4895, opened 2022-09-09
Td/v11 maybe tweaks

tomduncalf:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen (approved):
Looks reasonable to me.

src/jsi/jsi_class.hpp
```diff
@@ -125,10 +125,21 @@ inline void ObjectSetPrototypeOf(JsiEnv env, const fbjsi::Value& target, const f
...
-    auto obj = js::globalType(env, "Object");
-    return env(obj.getPropertyAsFunction(env, "create").callWithThis(env, obj, proto)).asObject();
+    if (REALM_UNLIKELY(!s_object)) {
+        s_object = js::globalType(env, "Object");
+    }
+    if (REALM_UNLIKELY(!s_object_create)) {
+        s_object_create = (*s_object).getPropertyAsFunction(env, "create");
+    }
+
+    return env((*s_object_create).callWithThis(env, *s_object, proto)).asObject();
```
RedBeard0531, line 142:
```suggestion
    return env(s_object_create->callWithThis(env, *s_object, proto)).asObject();
```

FYI, (almost) any time you do `(*x).y` you can instead do `x->y`. The only exceptions are when they are overloaded to do different things (bad idea) or `*x` is provided but `x->` isn't (sometimes a good idea). Neither issue applies to `std::optional` though.

Out of curiousity is there any perf difference between this and `env(s_object_create->call(env, proto)).asObject()` (that is, stop attaching `Object` as `this`)? I wasn't sure if it was safe to call the static methods on `Object` without a bound this and I wasn't able to test this while I was writing it, so I went for the conservative option of always doing the equivalent of `Object.foo()` rather than `const foo = Object.foo; foo()`. But it seems somewhat unlikely that it would matter, and I could imagine it being faster to not pass the `this` argument. IIRC there were several places where I did something like this. If it makes any difference it may be worth searching for `callWithThis` calls and seeing how many can be replaced with just `call`. (In a separate PR)

src/jsi/jsi_class.hpp
```diff
@@ -463,13 +483,20 @@ class ObjectWrap {
...
         set_internal(env, obj, ptr);
 
-        auto wrapper = (*s_ctor)->getProperty(env, "_proxyWrapper");
-        if (!wrapper.isUndefined()) {
-            obj = env(wrapper.asObject(env).asFunction(env).call(env, std::move(obj.get()))).asObject();
+        if (REALM_UNLIKELY(!s_wrapper)) {
+            s_wrapper = (*s_ctor)->getProperty(env, "_proxyWrapper");
+        }
+
+        if (!(*s_wrapper).isUndefined()) {
```
RedBeard0531, line 498: Given that we are creating `_proxyWrapper` in C++, we might as well drop it in `s_wrapper` (which could be an `optional<Function>`) when we create it (if it is needed). We could then just do `if (s_wrapper)` which will be slightly faster since it won't need to call into the engine to ask if it is undefined.

RedBeard0531 (approved):

src/jsi/jsi_class.hpp
tomduncalf: > FYI, (almost) any time you do (*x).y you can instead do x->y
Ah, of course, obvious in hindsight haha.

> Out of curiousity is there any perf difference between this and env(s_object_create->call(env, proto)).asObject()
Will take a look at this, thanks!

tomduncalf: >Out of curiousity is there any perf difference between this and env(s_object_create->call(env, proto)).asObject()

This does seem to yield a small (<5%) performance increase, good suggestion!

tomduncalf: Warrants more in depth testing probably as I just did a few runs and eyeballed the results, but yeah makes sense!

PR #4901, opened 2022-09-13
Align How Enums are Defined

### LINKED ISSUE #1955: Align how enums are defined
Currently Enums in the API use different definitions:

`SubscriptionState`:

```
// JS
        // Keep these value in sync with subscription_state.hpp
        realmConstructor.Sync.SubscriptionState = {
            Error: -1,      // An error occurred while creating or processing the partial sync subscription.
            Creating: 2,    // The subscription is being created.
            Pending: 0,     // The subscription was created, but has not yet been processed by the sync server.
            Complete: 1,    // The subscription has been processed by the sync server and data is being synced to the device.
            Invalidated: 3, // The subscription has been removed.
        };

// TypeScript
    enum SubscriptionState {
         Error,
         Creating,
         Pending,
         Complete,
         Invalidated,
    }

```

`SessionState`:

```
// JS
    /**
     * Gets the current state of the session.
     * Can be either:
     *  - "active": The session is connected to the Realm Object Server and is actively transferring data.
     *  - "inactive": The session is not currently communicating with the Realm Object Server.
     *  - "invalid": A non-recoverable error has occurred, and this session is semantically invalid. A new session should be created.
     * @type {string}
     */
    get state() {}

// TypeScript:
        readonly state: 'invalid' | 'active' | 'inactive';
```

`ConnectionState`:

    // JS
        /**
         * Gets the current state of the connection to the server. Multiple sessions might share the same underlying
         * connection. In that case, any connection change is sent to all sessions.
         *
         * Can be either:
         *  - Realm.Sync.ConnectionState.Disconnected: No connection to the server is available.
         *  - Realm.Sync.ConnectionState.Connecting: An attempt to connect to the server is in progress.
         *  - Realm.Sync.ConnectionState.Connected: The connection to the server is active and data can be synchronized.
         *
         * Data will only be synchronized with the Realm ObjectServer if this method returns `Connected` and `state()`
         * returns `Active` or `Dying`.
         *
         * @type {string}
         */
        connectionState() {}
    
    // TypeScript
        enum ConnectionState {
            Disconnected = "disconnected",
            Connecting = "connecting",
            Connected = "connected",
        }This is confusing and we should align it. Enums doesn't exist in JS and there is apparently no strong consensus on using either strings or integers to represent them. So after talking to @kraenhansen we agreeded that the most optimal approach would  be to use Strings as it massively increase debuggability compared to integer representations.

E.g. Enums should be defined like they are for `ConnectionState`. This means we need to align `SubscriptionState` and create a new `SessionState` enum for Typescript. We possibly need to do this elsewhere as well, e.g. Progress listeners.

This is breaking change.

takameyer:
## What, How & Why?
This addresses a breaking change we want to make where typed "string" unions are replaced with respective enums.

This closes #1955

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -613,16 +613,30 @@ declare namespace Realm {
         Connected = "connected",
     }
 
+    enum SessionState {
+        Invalid = "invalid",
+        Active = "active",
+        Inactive = "inactive",
+    }
+
+    enum ProgressDirection {
+        Download = "download",
+        Upload = "upload",
+    }
+
+    enum ProgressMode {
+        ReportIndefinitely = "reportIndefinitely",
+        ForCurrentlyOutstandingWork = "forCurrentlyOutstandingWork",
+    }
+
     type ProgressNotificationCallback = (transferred: number, transferable: number) => void;
-    type ProgressDirection = 'download' | 'upload';
-    type ProgressMode = 'reportIndefinitely' | 'forCurrentlyOutstandingWork';
 
     type ConnectionNotificationCallback = (newState: ConnectionState, oldState: ConnectionState) => void;
 
     namespace App.Sync {
         class Session {
             readonly config: SyncConfiguration;
-            readonly state: 'invalid' | 'active' | 'inactive';
+            readonly state: SessionState;
             readonly url: string;
             readonly user: User;
             readonly connectionState: ConnectionState;
@@ -763,7 +777,7 @@ declare namespace Realm {
              * of the `Subscriptions`. You should not use a superseded SubscriptionSet,
              * and instead obtain a new instance from {@link Realm.subscriptions}.
              */
-            Superseded = "Superseded",
+            Superseded = "superseded",
         }
 
         /**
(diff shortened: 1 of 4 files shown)


kneth (approved):

kneth:
@tomduncalf @cmelchior Please be aware that the changes to `Realm.Auth.EmailPasswordAuth` are cover by a private Jira ticket.

types/index.d.ts
```diff
@@ -613,16 +613,30 @@ declare namespace Realm {
         Connected = "connected",
     }
 
+    enum SessionState {
+        Invalid = "invalid",
```
takameyer, line 617: Implement this in JS

kraenhansen (approved):
Looks about right to me 🙂

kraenhansen:
It looks like we forgot the `Realm#addListener` which takes a `string` for event name, which should actually take an enum 🤔

PR #4908, opened 2022-09-14
use ccache in test job to speed up test app builds

iamer:


kraenhansen:
Try adding a step with printing the ccache stats after the test has run, it might reveal if the compilation is indeed being cached.
I found that running the iOS app locally produced a lot of uncacheable cache misses and tried to apply the official guidelines: https://reactnative.dev/docs/build-speed

iamer:
> Try adding a step with printing the ccache stats after the test has run, it might reveal if the compilation is indeed being cached.

The ccache post step added automatically by the action does that already.

kneth (approved):
LGTM (disclaimer: I don't know much about ccache)

.github/workflows/pr-realm-js.yml
```diff
@@ -120,9 +120,7 @@ jobs:
 
       - name: Configure ccache
         if: ${{ runner.os != 'Windows' }}
-        run: |
-          ccache --set-config="compiler_check=content"
-          ccache --show-config
+        run: ccache --set-config="compiler_check=content"
```
kraenhansen, line 123: I would like a comment with the origin of this config / why it's needed at all.

.github/workflows/pr-realm-js.yml
```diff
@@ -239,6 +237,18 @@ jobs:
         if: ${{ runner.os == 'Windows' }}
         uses: ilammy/msvc-dev-cmd@v1
 
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1
+        #TODO: figure out using ccache or equivalent on windows
+        if: ${{ matrix.variant.environment == 'react-native' }}
+        with:
+          key: ${{ runner.os }}-${{ matrix.variant.os }}-test
```
kraenhansen, line 245: Is it on purpose that we don't reuse the cache from the job above? If so, I'd like a comment on why we can't / won't make sense to reuse the key from job above.

scripts/ccache-clang++.sh
```diff
@@ -2,8 +2,12 @@
 if command -v ccache &> /dev/null; then
   export CCACHE_MAXSIZE=10G
   export CCACHE_CPP2=true
+  export CCACHE_DIRECT=true
+  export CCACHE_DEPEND=true
   export CCACHE_HARDLINK=true
-  export CCACHE_SLOPPINESS=file_macro,time_macros,include_file_mtime,include_file_ctime,file_stat_matches
+  export CCACHE_FILECLONE=true
+  export CCACHE_INODECACHE=true
+  export CCACHE_SLOPPINESS=file_macro,time_macros,include_file_mtime,include_file_ctime,file_stat_matches,modules,ivfsoverlay,pch_defines,system_headers
```
kraenhansen, line 10: It would be great with a comment linking to whatever resource you've used to determine which settings to use here.

scripts/ccache-clang.sh
```diff
@@ -2,8 +2,12 @@
 if command -v ccache &> /dev/null; then
   export CCACHE_MAXSIZE=10G
   export CCACHE_CPP2=true
+  export CCACHE_DIRECT=true
+  export CCACHE_DEPEND=true
   export CCACHE_HARDLINK=true
-  export CCACHE_SLOPPINESS=file_macro,time_macros,include_file_mtime,include_file_ctime,file_stat_matches
+  export CCACHE_FILECLONE=true
+  export CCACHE_INODECACHE=true
+  export CCACHE_SLOPPINESS=file_macro,time_macros,include_file_mtime,include_file_ctime,file_stat_matches,modules,ivfsoverlay,pch_defines,system_headers
```
kraenhansen, line 10: Same as with ccache-clang++.sh

kraenhansen (approved):
Great job. I could swear we did try something very similar before 🤔

.github/workflows/pr-realm-js.yml
iamer: No reason other than simplicity. The build job ccache cache key uses matrix.variant.arch which doesn't exist in this matrix. It could make sense to introduce a matching variable in this matrix, but I don't know what to set it to for other reach-native targets. If you have an idea what those could be then I will add them and share the cache across the jobs.

kraenhansen: Can you simply set it to `undefined`, `null` or equivalent?

PR #4909, opened 2022-09-14
Append ClusterName to AppName for CI Tests

takameyer:
## What, How & Why?
This change appends the `clusterName` to any deployed apps when deploying against the BaaS in Atlas.  This was done to ensure the apps are cleaned up properly when the tests are finished.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/utils/import-app.ts
+++ b/integration-tests/tests/src/utils/import-app.ts
@@ -56,8 +56,10 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
   // When running on CI we connect through mongodb-atlas instead of local-mongodb
   const { mongodbClusterName } = environment;
   if (typeof mongodbClusterName === "string") {
+    const replacements = { "config.json": { name: `${name}-${mongodbClusterName}` } };
     if (name === "with-db") {
       return {
+        ...replacements,
         "services/mongodb/config.json": {
           type: "mongodb-atlas",
           config: {
@@ -72,6 +74,8 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
       };
     } else if (name === "with-db-flx") {
       return {
+        ...replacements,
+        "config.json": { name: `${name}-${mongodbClusterName}` },
         "services/mongodb/config.json": {
           type: "mongodb-atlas",
           config: {
@@ -85,6 +89,7 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
         },
       };
     }
+    return replacements;
   }
   return {};
 }
--- a/integration-tests/tests/src/utils/import-app.ts
+++ b/integration-tests/tests/src/utils/import-app.ts
@@ -56,10 +56,12 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
   // When running on CI we connect through mongodb-atlas instead of local-mongodb
   const { mongodbClusterName } = environment;
   if (typeof mongodbClusterName === "string") {
-    const replacements = { "config.json": { name: `${name}-${mongodbClusterName}` } };
+    const appName = `${name}-${mongodbClusterName}`;
+    const appNameReplacement = { "config.json": { name: appName } };
+
     if (name === "with-db") {
       return {
-        ...replacements,
+        ...appNameReplacement,
         "services/mongodb/config.json": {
           type: "mongodb-atlas",
           config: {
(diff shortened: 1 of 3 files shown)


integration-tests/tests/src/utils/import-app.ts
```diff
@@ -56,8 +56,10 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
   // When running on CI we connect through mongodb-atlas instead of local-mongodb
   const { mongodbClusterName } = environment;
   if (typeof mongodbClusterName === "string") {
+    const replacements = { "config.json": { name: `${name}-${mongodbClusterName}` } };
```
kraenhansen, line 59: Could we move this to a separate `appName` const and reuse that below?
I remember Nikola mentioning that the app name needs to be below 31 symbols - how do we guarantee that?

kraenhansen: I remember Nikola mentioning that the app name needs to be below 31 symbols - how do we guarantee that?

integration-tests/tests/src/utils/import-app.ts
```diff
@@ -72,6 +74,8 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
       };
     } else if (name === "with-db-flx") {
       return {
+        ...replacements,
```
kraenhansen, line 77: Perhaps we need a bit deeper merge of these objects?

integration-tests/tests/src/utils/import-app.ts
takameyer: We could throw an exception 🤷🏼 That's in our hands to create valid app names.  The clusterName.length should stay consistent in CI.

integration-tests/tests/src/utils/import-app.ts
takameyer: At least in the case of the base `config.json`, the object is flat

integration-tests/tests/src/utils/import-app.ts
```diff
@@ -72,6 +74,8 @@ export function getDefaultReplacements(name: string): TemplateReplacements {
...
+        "config.json": { name: `${name}-${mongodbClusterName}` },
```
takameyer, line 78: I will remove this line

packages/realm-app-importer/src/AppImporter.ts
```diff
@@ -233,9 +233,10 @@ export class AppImporter {
     }
   }
 
-  private loadAppConfigJson(appTemplatePath: string): AppConfig {
+  private loadAppConfigJson(appTemplatePath: string, replacements: TemplateReplacements = {}): AppConfig {
     const configJsonPath = path.resolve(appTemplatePath, "config.json");
-    return this.loadJson(configJsonPath);
+    const configJson = this.loadJson(configJsonPath);
+    return replacements["config.json"] ? { ...configJson, ...replacements["config.json"] } : configJson;
```
kraenhansen, line 239: This seems simpler.

```suggestion
    return { ...configJson, ...replacements["config.json"] };
```

packages/realm-app-importer/src/AppImporter.ts
```diff
@@ -233,9 +233,10 @@ export class AppImporter {
     }
   }
 
-  private loadAppConfigJson(appTemplatePath: string): AppConfig {
+  private loadAppConfigJson(appTemplatePath: string, replacements: TemplateReplacements = {}): AppConfig {
```
kraenhansen, line 236: Perhaps this should take the `appName` as a (optional) parameter instead?

integration-tests/tests/src/utils/import-app.ts
kraenhansen: Never mind - i assumed this was provided as an argument, didn't read all the code above.

packages/realm-app-importer/src/AppImporter.ts
takameyer: Yeah, we can try that.  The underlying code now looks like this:
```typescript
  private loadAppConfigJson(appTemplatePath: string, appNameOverride?: string): AppConfig {
    const configJsonPath = path.resolve(appTemplatePath, "config.json");
    const configJson = this.loadJson(configJsonPath);
    return appNameOverride ? { ...configJson, ...{ name: appNameOverride } } : configJson;
  }
```

takameyer: and now:
```typscript
    const appNameOverride = { ...replacements["config.json"] }?.name as string;
 ```

takameyer: Kind of like the other way better, as we are "loading" `AppConfigJson`, which could possibly include other template replacements other than `name`

packages/realm-app-importer/src/AppImporter.ts
takameyer: Good point, i'll change it

kneth (approved):
LGTM

PR #4910, opened 2022-09-14
Pinning NDK and use `--active-arch-only` on Android

kraenhansen:
## What, How & Why?

Fixed pinning of the NDK used to build our prebuilt binaries for Android. Our previous version, Realm JS v10.21.0 was compiled with NDK 25.1.8937393, which could result in unpredictable crashes, since this needs to match the one being used by React Native.

This closes potential crashes of `10.21.0` and the crashing integration tests on `v11`.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


takameyer (approved):
LGTM.  I remember trying the `--active-arch-only` flag, but this did not seem to make much of a difference in build times.

kneth (approved):

PR #4928, opened 2022-09-16
objectForPrimaryKey should return null instead of undefined

### LINKED ISSUE #3966: objectForPrimaryKey should return null instead of undefined
### Problem

When an object is not found with `objectForPrimaryKey`, `undefined` is returned.  This is syntactically not correct.  It should return a `null` value, since the object does not exist.  `undefined` should be reserved for unset properties.

### Solution

Change the implementation of `objectForPrimaryKey` to return `null`.
Update the types as follows:
```
objectForPrimaryKey<T>(type: string, key: Realm.PrimaryKey): (T & Realm.Object) | null;
objectForPrimaryKey<T extends Realm.Object>(type: {new(...arg: any[]): T; }, key: Realm.PrimaryKey): T | null;
```

takameyer:
## What, How & Why?
`Realm#objectForPrimaryKey` now returns `null` instead of `undefined` when no object is found

This closes #3966

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ### Notes
 Based on Realm JS v10.21.1: See changelog below for details on enhancements and fixes introduced between this and the previous pre release (which was based on Realm JS v10.19.5).
 
-### Breaking change
+### Breaking changes
+* When no object is found calling `Realm#objectForPrimaryKey`, `null` is returned instead of `undefined`
 * Removed deprecated positional arguments to Email/Password authentication functions
     * The following functions now only accept object arguments:
     ```javascript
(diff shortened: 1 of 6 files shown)


kneth (approved):

kraenhansen (approved):

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@
 ### Notes
 Based on Realm JS v10.21.1: See changelog below for details on enhancements and fixes introduced between this and the previous pre release (which was based on Realm JS v10.19.5).
 
-### Breaking change
+### Breaking changes
+* When no object is found calling `Realm#objectForPrimaryKey`, `null` is returned instead of `undefined`
+* Removal of deprecated functions, which should be replaced with `Realm.Credentials#apiKey`:
```
kraenhansen, line 8: I think you accidentally included some changelog from another PR here.

PR #4938, opened 2022-09-21
Babel plugin changes

tomduncalf:
## What, How & Why?

This PR contains all additional changes I've made to Kraen's Babel Plugin branch, except the template app changes which are in a [separate PR](https://github.com/realm/realm-js/pull/4964/files), because they can't be merged yet.

I've also created a [separate PR](https://github.com/realm/realm-js/pull/4967) containing the entire Babel plugin pointing to `v11` if we'd prefer to review it all in one.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/babel-plugin/README.md
```diff
@@ -0,0 +1,239 @@
...
+   ```js
+   // Existing babel.config.js content is commented out
+   // module.exports = {
+     // presets: ['module:metro-react-native-babel-preset'],
+
+     // --------------------------
+     // Add the following plugins:
+     plugins: [
+       '@realm/babel-plugin',
+       ['@babel/plugin-proposal-decorators', { legacy: true }],
```
kraenhansen, line 97: I was thinking that we should also publish a `@realm/babel-config` which would check / add these plugins, to simplify consumption. Perhaps not right away, but eventually.

packages/babel-plugin/src/plugin/index.ts
```diff
@@ -99,6 +112,50 @@ function getRealmTypeForTSTypeReference(path: NodePath<types.TSTypeReference>):
...
   } else if (isRealmTypeAlias(path, "Mixed") || isRealmTypeAlias(path, "Mixed", null)) {
     return { type: "mixed" };
+  } else if (isRealmTypeAlias(path, "LinkingObjects")) {
+    const classPropertyNode = path.parentPath?.parentPath?.node;
+    if (
+      // Keep TS happy
+      !types.isClassProperty(classPropertyNode) ||
+      classPropertyNode.optional
+    ) {
+      throw new Error(getLinkingObjectsError("Properties of type LinkingObjects cannot be optional"));
```
kraenhansen, line 122: 🥇 for preconditions + errors ... Perhaps we're missing some elsewhere?

tomduncalf: Yeah probably – I actually meant to ask your opinion on whether this is overkill or not? i.e. should we just rely on the type definitions? 

I think throwing an explicit error is probably better though, as sometimes people will ignore TS errors

packages/babel-plugin/README.md
tomduncalf: That is a nice idea!

packages/babel-plugin/src/plugin/index.ts
kraenhansen: Would it be possible to send this to babel somehow, in a way that wouldn't make it completely unusable if a property is misconfigured, but simply error through to the bundler. I don't know what's available for babel plugins to report errors or warnings.

tomduncalf: Yeah that is a good idea – I'll do some research.

tomduncalf: Actually I think this should be a `throw`, as if you replace it with e.g. a `console.warn`, you then end up with `Property 'xxx' of type 'linking objects' cannot be nullable.` – so you are specifying an invalid schema.

However if you use `throw path.buildCodeFrameError(errorMsg)`, you get a better error message which shows the source of the error, so I will switch to that.

image

tomduncalf: I couldn't really think of any other good preconditions. The type argument for List/Dictionary/Set is optional and if they supply more than 1 type argument for some reason it will still work as intended... doesn't really seem worth checking. Can you think of anywhere else?

packages/babel-plugin/README.md
```diff
@@ -0,0 +1,235 @@
+# Realm Babel Plugin
+
+## Introduction
+
+The Realm Babel plugin enables you to define your Realm models using standard Typescript syntax - no need to define a separate schema.
```
kneth, line 5:
```suggestion
The Realm Babel plugin enables you to define your Realm models using standard TypeScript syntax - no need to define a separate schema.
```

packages/babel-plugin/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Restrictions
+
+### All class properties will be added to the Realm schema
+
+There is currently no way to specify a property on your class which should not be persisted to the Realm.
+
+### Classes extending Realm.Object cannot be constructed with `new` outside of a write transaction
+
+This plugin does not change the behaviour of `Realm.Object`, which cannot be constructed using `new` outside of a write transaction - there is no concept of a Realm.Object which is not stored in a Realm. Constructing a Realm object with `new` inside a write transaction will create a new object in the Realm - see [class-based models in the CHANGELOG](/CHANGELOG.md#enhancements-4).
```
kneth, line 235:
```suggestion
This plugin does not change the behaviour of `Realm.Object`, which cannot be constructed using `new` outside of a write transaction - there is no concept of a `Realm.Object` which is not stored in a Realm. Constructing a Realm object with `new` inside a write transaction will create a new object in the Realm - see [class-based models in the CHANGELOG](/CHANGELOG.md#enhancements-4).
```

packages/babel-plugin/README.md
```diff
@@ -0,0 +1,235 @@
...
+| `index`   | none                          | Specifies that the decorated property should be indexed by Realm.                                    |
+| `mapTo`   | `(realmPropertyName: string)` | Specifies that the decorated property should be stored as `realmPropertyName` in the Realm database. |
+
+The example below shows both decorators in use:
+
+```ts
+import Realm from "realm";
+import { mapTo, index } from "@realm/babel-plugin";
+
+export class Task extends Realm.Object {
```
kneth, line 216: Do we need to change `Realm.Object` to `Realm.Object<Task>`?

kneth (approved):

packages/babel-plugin/README.md
tomduncalf:
```suggestion
export class Task extends Realm.Object<Task> {
```

tomduncalf:
@kraenhansen I made the changes we talked about yesterday:

- Split out into new PRs and remove from this:
    - RequiredFields change
    - Default property function
- Decorators: 
    - Remove rollup and export decorators from realm
    - Remove decorators from the babel output + make them throw if used at runtime
- Readme:
    - Copy paste error on ObjectId in readme
    - Add the other types to Realm.Types in extensions.js
    - Remove unknown from README. 
    - Note that provided by Realm are exported at the top level as well as in types

packages/babel-plugin/README.md
kneth: I have created https://github.com/realm/realm-js/issues/5009 to track it

packages/babel-plugin/src/plugin/index.ts
```diff
@@ -163,38 +225,141 @@ function getRealmTypeForClassProperty(path: NodePath<types.ClassProperty>): Real
...
+
+function findDecoratorIdentifier(
+  decoratorsPath: NodePath<types.Decorator>[],
+  name: string,
+): NodePath<types.Decorator> | undefined {
+  return decoratorsPath.find(
+    (d) =>
+      d.node &&
+      isRealmDecoratorWithName(d.node.expression, name) &&
+      // ((types.isIdentifier(d.node.expression) && d.node.expression.name === name) ||
```
kneth, line 247: Can it be removed?

packages/babel-plugin/src/plugin/index.ts
```diff
@@ -163,38 +225,141 @@ function getRealmTypeForClassProperty(path: NodePath<types.ClassProperty>): Real
...
+function findDecoratorCall(
+  decoratorsPath: NodePath<types.Decorator>[],
+  name: string,
+): { decoratorNode: NodePath<types.Decorator>; callExpression: types.CallExpression } | undefined {
+  const node = decoratorsPath.find(
+    (d) =>
+      d.node &&
+      types.isCallExpression(d.node.expression) &&
+      isRealmDecoratorWithName(d.node.expression.callee, name) &&
+      // types.isIdentifier(d.node.expression.callee) &&
```
kneth, line 266: Can it be removed?

kneth (approved):

PR #4958, opened 2022-09-29
Fix #4513

### LINKED ISSUE #4513: Incorrect wrapping of proxy objects on every render

takameyer:
## What, How & Why?
Update useObject and useQuery to wrap their return value with proxy only when it needs to be rerendered

This closes #4513

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -4,15 +4,10 @@
 * None
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None
+* Fix non-realm re-renders events from creating a new instance of collections and objects returned from `useObject` and `useQuery` ([#4513](https://github.com/realm/realm-js/issues/4513), since v0.2.0, special thanks to @mfbx9da4 for the PR)
 
 ### Compatibility
-* React Native >= v0.64.0
-* Atlas App Services.
-* Realm Studio v12.0.0.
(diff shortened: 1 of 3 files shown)


tomduncalf (approved):
Makes sense to me 👍

PR #4966, opened 2022-09-30
Add toJSON implementation and testing

gagik:
## What, How & Why?

This is a breaking change in sense that it reintroduces `toJSON` functionality with the binding generator but removes `JSONSerializerReplacer`. Instead, users are advised to use external libraries such as [flatted](https://www.npmjs.com/package/flatted) and [@ungap/structured-clone](https://github.com/ungap/structured-clone/#readme) if they need to deal with serialising objects with circular references into a JSON string. 

The expected functionality of `toJSON` is therefore to generate valid, plain JavaScript objects with correct circular references so those can be used by the packages mentioned above. Current implementation supports Objects and Collections. It is nearly completely ported from pre-bindgen implementation,  see [this for objects](https://github.com/realm/realm-js/blob/138e77133b78a1d28fdfe4b81e6b13f7cbfe7064/lib/extensions.js#L82) and [this for collections](https://github.com/realm/realm-js/blob/138e77133b78a1d28fdfe4b81e6b13f7cbfe7064/lib/extensions.js#L65). Some changes were made to adapt both to the current state of the API and possibly improve performance by using a 2D map.

This PR is not complete and still needs feedback regarding some approaches and testing, see my review for details.

This closes https://github.com/realm/realm-js/issues/4949

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests
* [] 💥 `Breaking` label has been applied or is not necessary

* [ ] jsdoc files updated


integration-tests/tests/src/tests/serialization.ts
```diff
@@ -39,26 +39,12 @@ import circularCollectionResult from "../structures/circular-collection-result.j
...
-  });
-});
-
 type TestSetup = {
   name: string;
   schema: (Realm.ObjectSchema | Realm.ObjectClass)[];
   testData: (realm: Realm) => unknown;
 };
 
-interface ICacheIdTestSetup {
```
gagik, line 55: I am deleting these cached tests since we're not using a serializer... if we decide to add `flatted` tests here a format like this is probably useful.

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -89,49 +75,17 @@ const testSetups: TestSetup[] = [
         // Playlists
         const p1 = realm.create<IPlaylistNoId>(PlaylistSchemaNoId.name, {
           title: "Playlist 1",
-          songs: [
-            s1,
-            s2,
-            s3,
-            {
-              artist: "Unique artist 1",
```
gagik, line 97: I had to delete non-pre-created objects since those do not seem to be supported right now.. it does make these tests less different from one another though so I am not sure multiple playlist objects are needed (also it makes no difference whether primary keys are supplied or not in functionality now, so maybe I should remove that too).

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -312,57 +213,18 @@ const testSetups: TestSetup[] = [
...
-              _id: 7,
-              artist: "Unique artist 4",
-              title: "Unique title 4",
-            },
-            s3,
-          ],
+          songs: [s3],
           related: [p1],
         });
         const p3 = realm.create(PlaylistWithId, {
```
gagik, line 224: again, not sure if all these objects for testing are valuable as they are not meaningfully different now

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -458,34 +265,23 @@ describe("JSON serialization", () => {
...
-
           // Check that linked list is not a Realm entity.
           expect(serializable.related).not.instanceOf(Realm.Collection);
           // But is a plain Array
           expect(Array.isArray(serializable.related)).equals(true);
-
           // Check that the serializable object is the same as the first related object.
           // (this check only makes sense because of our structure)
-          expect(serializable).equals(serializable.related[0]);
+          const relatedObjects = serializable.related as Array<typeof serializable>;
```
gagik, line 276: not sure if using `typeof serializable` is okay here / in general, not too familiar with practices just yet.

packages/realm/src/Collection.ts
```diff
@@ -39,6 +41,20 @@ export abstract class Collection<T = unknown, ChangeCallbackType = unknown> impl
     throw new Error("Not yet implemented");
   }
 
+  toJSON(_?: string, cache = new Map()): Array<DefaultObject> {
+    // return this.map((item, index) =>
```
gagik, line 45: original implementation used map, since we do not have it defined I just went with regular loop

packages/realm/src/Collection.ts
```diff
@@ -39,6 +41,20 @@ export abstract class Collection<T = unknown, ChangeCallbackType = unknown> impl
     throw new Error("Not yet implemented");
   }
 
+  toJSON(_?: string, cache = new Map()): Array<DefaultObject> {
```
gagik, line 44: in here and in object `toJSON`, I used `DefaultObject` as expected return value. This to me feels like a way to make sure it is a "plain" object for sake of future JSON libraries but maybe there is a better way to go about it.

packages/realm/src/Object.ts
```diff
@@ -43,6 +43,8 @@ const PROXY_HANDLER: ProxyHandler<RealmObject<any>> = {
   },
 };
 
+type JSONCacheMap = Map<string, Map<string, DefaultObject>>;
```
gagik, line 46: I am defining a type here for readability, not sure if this is meant to be done somewhere / removed.
I am also not sure if maps are best here because they do have to be initialised

packages/realm/src/Object.ts
```diff
@@ -120,13 +122,52 @@ class RealmObject<T = DefaultObject> {
...
+    if (existing) {
+      return existing;
+    }
+
+    // Create new result, and store in cache.
+    const result: DefaultObject = {};
+    this._setOwnCache(cache, result);
+
+    // Move all enumerable keys to result, triggering any specific toJSON implementation in the process.
+    // TODO: Consider Object.keys(this)
```
gagik, line 140: is there any value to `.key(..).concat(...prototype)` now? again I skipped as when I started only looping was possible so I am unsure whether this is not equivalent with the concat.

packages/realm/src/Object.ts
```diff
@@ -120,13 +122,52 @@ class RealmObject<T = DefaultObject> {
...
+        continue; // continue
+      }
+
+      if (value === null || value === undefined) {
+        result[key] = value;
+      } else if (value instanceof RealmObject || value instanceof Realm.Collection) {
+        // recursively trigger `toJSON` for Realm instances with the same cache.
+        result[key] = value.toJSON(key, cache);
+      } else if (value instanceof Realm.Dictionary /* TODO: value[Symbol.IS_PROXIED_DICTIONARY] */) {
+        // TODO: Consider dictionary compatibility comment.
```
gagik, line 157: I am quite confused by this `Symbol.IS_PROXIED_DICTIONARY` comment / information. there also were no dictionary tests so I think I would need to add them once they are implemented.

gagik (commented):
Some things I am not sure about

integration-tests/tests/src/tests/serialization.ts
kraenhansen: > non-pre-created objects

What do you mean by that?

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -164,31 +118,17 @@ const testSetups: TestSetup[] = [
...
           title: "Playlist 1",
-          songs: [
-            s1,
-            s2,
-            s3,
-            { artist: "Unique artist 1", title: "Unique title 1" },
-            { artist: "Unique artist 2", title: "Unique title 2" },
-          ],
+          songs: [s1, s2, s3],
+          related: [],
```
kraenhansen, line 122: Is this needed?

integration-tests/tests/src/tests/serialization.ts
kraenhansen: I'd be okay with you removing those, as they're testing a detail of the implementation that we're no longer doing.

packages/realm/src/Collection.ts
kraenhansen: I'd probably want to more this to `OrderedCollection` and then you'd be able to use the `map` method.

packages/realm/src/Object.ts
```diff
@@ -120,13 +122,52 @@ class RealmObject<T = DefaultObject> {
...
+      if (typeof value === "function") {
+        continue; // continue
+      }
+
+      if (value === null || value === undefined) {
+        result[key] = value;
+      } else if (value instanceof RealmObject || value instanceof Realm.Collection) {
+        // recursively trigger `toJSON` for Realm instances with the same cache.
+        result[key] = value.toJSON(key, cache);
+      } else if (value instanceof Realm.Dictionary /* TODO: value[Symbol.IS_PROXIED_DICTIONARY] */) {
```
kraenhansen, line 156: I think we should simply defer to the `Realm.Dictionary#toJSON` here.

kraenhansen: I'd appreciate a test covering dictionaries too.

packages/realm/src/Object.ts
```diff
@@ -156,6 +197,24 @@ class RealmObject<T = DefaultObject> {
...
+  /**
+   * @internal
+   * Set cache belonging to this object to given value
+   */
+  private _setOwnCache(cache: JSONCacheMap, value: DefaultObject) {
+    if (!cache.has(this[INTERNAL].table.name)) {
+      cache.set(this[INTERNAL].table.name, new Map());
+    }
+    cache.get(this[INTERNAL].table.name)?.set(this._objectKey(), value);
+  }
```
kraenhansen, line 216: Could be refactored to free functions or a class.

packages/realm/src/Object.ts
kraenhansen: Moving it to a separate file makes sense in my opinion.

packages/realm/src/Object.ts
```diff
@@ -156,6 +197,24 @@ class RealmObject<T = DefaultObject> {
...
+  /**
+   * @internal
+   * Set cache belonging to this object to given value
+   */
+  private _setOwnCache(cache: JSONCacheMap, value: DefaultObject) {
+    if (!cache.has(this[INTERNAL].table.name)) {
+      cache.set(this[INTERNAL].table.name, new Map());
+    }
+    cache.get(this[INTERNAL].table.name)?.set(this._objectKey(), value);
```
kraenhansen, line 215:
```suggestion
    const { name } = this[INTERNAL].table;
    let tableMap = cache.get(name);
    if (!tableMap) {
      tableMap = new Map();
      cache.set(name, tableMap);
    }
    tableMap.set(this._objectKey(), value);
```

packages/realm/src/Dictionary.ts
```diff
@@ -87,6 +90,13 @@ const PROXY_HANDLER: ProxyHandler<Dictionary> = {
   },
 };
 
+function toJSON(this: Dictionary, _?: string, cache = new JSONCacheMap()): DefaultObject {
```
gagik, line 93: TEMPORARY: For the sake of testing

packages/realm/src/Object.ts
```diff
@@ -120,13 +123,37 @@ class RealmObject<T = DefaultObject> {
...
+    // Check if current objectId has already processed, to keep object references the same.
+    const existing = cache.find(this);
+    if (existing) {
+      return existing;
     }
-    return { ...this };
+    const result: DefaultObject = {};
+    cache.add(this, result);
+    // Move all enumerable keys to result, triggering any specific toJSON implementation in the process.
+    Object.entries(this).forEach(([key, value]) => {
```
gagik, line 143: Reason why this is a `forEach` instead of a .`entries(...).map(...)` is because we need to add the object into cache before we start looping through it.

integration-tests/tests/package.json
```diff
@@ -26,6 +26,7 @@
     "@types/chai-as-promised": "^7.1.4",
     "@types/mocha": "^5.2.6",
     "@types/node": "^16.11.12",
+    "@ungap/structured-clone": "^1.0.1",
```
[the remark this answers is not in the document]
kraenhansen: > I do think we cover a good enough ground without them already when we compare objects.

In that case - I'd suggest leaving it out to keep things simple.

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -17,545 +17,187 @@
...
+      this.birthdaysSerialized = {
+        dict: {
+          Bob: "August",
+          Tom: "January",
+        },
+      };
+      // Dictionary object test
+      this.birthdays = this.realm.create<IBirthdays>("Birthdays", this.birthdaysSerialized);
+
+      // See #4980.
```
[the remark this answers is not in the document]
kraenhansen: Feel free to merge #4980 and rebase this PR.

packages/realm/src/Dictionary.ts
```diff
@@ -217,4 +220,18 @@ export class Dictionary<T = unknown> extends Collection<string, T, [string, T],
...
+  // /**
+  //  * Returns a plain object representation with possible circular references
+  //  * from the dictionary for JSON serialization.
+  //  * @returns A plain object
+  //  */
+  // @ts-expect-error We're exposing methods in the users value namespace
+  toJSON(_?: string, cache = new JSONCacheMap()): DefaultObject {
+    return Object.fromEntries(
+      Object.entries(this)
+        .filter(([k]) => k != "toJSON")
```
kraenhansen, line 233: I'm surprised this would ever happen, perhaps you're missing a non-enumerable property definition of the `toJSON` in the dictionary?

packages/realm/src/Object.ts
```diff
@@ -93,15 +96,11 @@ class RealmObject<T = DefaultObject> {
...
           result[property.name] = defaultValue;
-        } else if (
-          !(property.type & binding.PropertyType.Collection) &&
-          !(property.type & binding.PropertyType.Nullable) &&
-          created
-        ) {
-          throw new Error(`Missing value for property '${property.name}'`);
-        }
+        } else if (created) {
+        throw new Error(`Missing value for property '${property.name}'`);
```
kraenhansen, line 100: Do you remember why this was needed?

kraenhansen (approved):
I have a few comments. This is very close to mergeable.

packages/realm/src/Object.ts
gagik: this was a mess-up on my end from rebasing. I'll revert this.

packages/realm/src/Dictionary.ts
gagik: I will need to check this again, it may actually not be happening since I originally wrote this when I was adding toJSON to dictionaries in a different way

gagik: yeah this is not necessary. I removed it.

kraenhansen (approved):

PR #4977, opened 2022-10-03
Test case: Add simple reference list test

gagik:
Simple test to show error with plain objects in list fields.


packages/realm/src/tests/list-objects.test.ts
```diff
@@ -0,0 +1,66 @@
...
+          {
+            name: "Bob",
+            age: 10,
+          },
+          this.tom,
+        ],
+      });
+    });
+    expect(this.alice.friends[0].name).equals("Bob");
+    expect(this.alice.friends[1].tom).equals("Tom");
```
kraenhansen, line 64:
```suggestion
    expect(this.alice.friends[1].name).equals("Tom");
```

PR #4981, opened 2022-10-05
Add Realm.Configuration.cache

kneth:
## What, How & Why?

Expose Realm Core's `cache` configuration option.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [x] jsdoc files updated

### DIFF
--- a/docs/realm.js
+++ b/docs/realm.js
@@ -426,6 +426,7 @@ class Realm {
  *    what fits in memory, but it is not persistent and will be removed when the last instance
  *    is closed. This option is incompatible with `sync`.
  * @property {boolean} [readOnly=false] - Specifies if this Realm should be opened as read-only.
+ * @property {boolean} [cache=true] - Cache Realm instances. 
  * @property {boolean} [disableFormatUpgrade=false] - Specifies if this Realm's file format should
  *    be automatically upgraded if it was created with an older version of the Realm library.
  *    If set to `true` and a file format upgrade is required, an error will be thrown instead.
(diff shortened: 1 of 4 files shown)


docs/realm.js
```diff
@@ -426,6 +426,7 @@ class Realm {
  *    what fits in memory, but it is not persistent and will be removed when the last instance
  *    is closed. This option is incompatible with `sync`.
  * @property {boolean} [readOnly=false] - Specifies if this Realm should be opened as read-only.
+ * @property {boolean} [cache=true] - Cache Realm instances.
```
bmunkholm, line 429: Couldn't this use an explanation as to when to cache and when not to?

bmunkholm: Were we caching previously?

tgoyne: No, this is something we specifically do not expose publicly.

takameyer (approved):
LGTM

docs/realm.js
takameyer: There's a lint error here that should be fixed.

nirinchev: I think you should fix/reference count close rather than expose cache.

takameyer: It's a workaround until we have the time and resources to fix said issue.

bmunkholm: Is the plan then to make a breaking change later to remove the option once this real issue is fixed?

As a minimum, there has to be a descriptive parameter description.

PR #4992, opened 2022-10-07
Check peerDependencies in analytics post-install script

kneth:
## What, How & Why?

It looks our builder analytics is overestimating node.js usage. 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 ### Internal
 * Upgraded Realm Core from v12.6.0 to v12.9.0. ([#4932](https://github.com/realm/realm-js/issues/4932) and [#4983](https://github.com/realm/realm-js/issues/4983)
 * Added ARM/Linux build guide.
+* Check peer dependencies in analytics post-install script.
 
 ## 10.21.1 (2022-09-15)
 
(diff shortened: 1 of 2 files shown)


tomduncalf (approved):
Seems like a reasonable change code-wise, I don't know about how it might or might not solve any overestimating

kraenhansen:
What is the problem being solved here?

kneth:
@kraenhansen Mostly defensiveness as we have seen an uptake of node.js which we cannot explain easily - only if Electron or React Native are not in `dependencies` or `devDependencies`.

kraenhansen:
Some package need to declare react-native or electron as dependency, perhaps resolving it using `require.resolve("react-native/package.json")` would be more robust?

> which we cannot explain easily

I would assume some assumption of the script is wrong all-together. I'd suggest investigating the root cause of the uptake as an alternative to merging this PR. I fear this change would be causing more confusion than solving any issue.

For one, we should probably test the script with all known package managers being used out there (including but possibly not limited to yarn and pnpm).

kneth:
I am closing this PR as we believe it requires more investigation: https://github.com/realm/realm-js/issues/4998

PR #4997, opened 2022-10-10
Improve toJSON functionality and remove JSONSerializerReplacer

gagik:
## What, How & Why?
This is a breaking change that removes `JSONSerializerReplacer` and improves the current `toJSON` functionality. It also improves current tests, organizing them better and adding cases for dictionaries. 
With removal `JSONSerializerReplacer`, users are advised to use external libraries such as [flatted](https://www.npmjs.com/package/flatted) and [@ungap/structured-clone](https://github.com/ungap/structured-clone/#readme) if they need to deal with serialising objects with circular references into a JSON string.
The expected functionality of `toJSON` is therefore to generate valid, plain JavaScript objects with correct circular references so those can be used by the packages mentioned above. `toJSON` is implemented by Realm Object, Dictionary, and Collection.

Related to https://github.com/realm/realm-js/pull/4966.

## ☑️ ToDos

* [X] 📝 Changelog entry
* [X] 🚦 Tests
* [X] 🔀 Executed flexible sync tests locally if modifying flexible sync

* [X] jsdoc files updated
* [X] Chrome debug API is updated if API is available on React Native

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,7 @@ Based on Realm JS v10.21.1: See changelog below for details on enhancements and 
 
 ### Enhancements
 * Small improvement to performance by caching JSI property String object [#4863](https://github.com/realm/realm-js/pull/4863)
-* Small improvement to performance for `toJSON` which should make it useful for cases where a plain representations of Realm entities are needed, e.g. when inspecting them for debugging purposes through `console.log(realmObj.toJSON())`.  
+* Small improvement to performance for `toJSON` which should make it useful for cases where a plain representations of Realm entities are needed, e.g. when inspecting them for debugging purposes through `console.log(realmObj.toJSON())`. ([#4997](https://github.com/realm/realm-js/pull/4997)) 
 
 ### Fixed
 * None
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,7 @@ Based on Realm JS v10.21.1: See changelog below for details on enhancements and 
   ```
   * A typo was fixed in the `SubscriptionsState` enum, in which `SubscriptionsState.Superseded` now returns `superseded` in place of `Superseded`
 * `"discardLocal"` is now the default client reset mode. ([#4382](https://github.com/realm/realm-js/issues/4382))
-
+* Removed `Realm.JsonSerializationReplacer`. Use circular JSON serialization libraries such as [@ungap/structured-clone](https://www.npmjs.com/package/@ungap/structured-clone) and [flatted](https://www.npmjs.com/package/flatted) for stringifying Realm entities that have circular structures. The Realm entities' `toJSON` method returns plain objects and arrays (with circular references if applicable) which makes them compatible with any serialization library that supports stringifying plain JavaScript types. ([#4997](https://github.com/realm/realm-js/pull/4997))
 ### Enhancements
 * Small improvement to performance by caching JSI property String object [#4863](https://github.com/realm/realm-js/pull/4863)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ Based on Realm JS v10.21.1: See changelog below for details on enhancements and 
 
 ### Enhancements
 * Small improvement to performance by caching JSI property String object [#4863](https://github.com/realm/realm-js/pull/4863)
+* Small improvement to performance for `toJSON` which should make it useful for cases where a plain representations of Realm entities are needed, e.g. when inspecting them for debugging purposes through `console.log(realmObj.toJSON())`.  
 
 ### Fixed
 * None
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ Based on Realm JS v10.21.1: See changelog below for details on enhancements and 
   * A typo was fixed in the `SubscriptionsState` enum, in which `SubscriptionsState.Superseded` now returns `superseded` in place of `Superseded`
 * `"discardLocal"` is now the default client reset mode. ([#4382](https://github.com/realm/realm-js/issues/4382))
 * Removed `Realm.JsonSerializationReplacer`. Use circular JSON serialization libraries such as [@ungap/structured-clone](https://www.npmjs.com/package/@ungap/structured-clone) and [flatted](https://www.npmjs.com/package/flatted) for stringifying Realm entities that have circular structures. The Realm entities' `toJSON` method returns plain objects and arrays (with circular references if applicable) which makes them compatible with any serialization library that supports stringifying plain JavaScript types. ([#4997](https://github.com/realm/realm-js/pull/4997))
+
 ### Enhancements
 * Small improvement to performance by caching JSI property String object [#4863](https://github.com/realm/realm-js/pull/4863)
 
--- a/integration-tests/tests/src/tests/serialization.ts
+++ b/integration-tests/tests/src/tests/serialization.ts
@@ -19,543 +19,180 @@
 import { expect } from "chai";
 import Realm from "realm";
 
-import {
-  IPlaylist as IPlaylistNoId,
-  ISong as ISongNoId,
-  PlaylistSchema as PlaylistSchemaNoId,
-  SongSchema as SongSchemaNoId,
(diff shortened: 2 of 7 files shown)


src/js_realm_object.hpp
```diff
@@ -352,6 +354,22 @@ void RealmObjectClass<T>::get_object_key(ContextType ctx, ObjectType object, Arg
...
+{
+    args.validate_maximum(0);
+
+    auto realm_object = get_internal<T, RealmObjectClass<T>>(ctx, object);
+    if (!realm_object) {
+        throw std::runtime_error("Invalid 'this' object");
+    }
+
+    const Obj& obj = realm_object->obj();
+    auto table_key = obj.get_table()->get_key();
```
gagik, line 369: This seems like something to address in core: there is a private `get_table_key()` method which is just a shortcut to this. I think it'd be good to make that method public since I imagine this would be nice to use in contexts like this (and you can access it anyways so no reason not to).

src/js_realm_object.hpp
```diff
@@ -110,6 +111,7 @@ struct RealmObjectClass : ClassDefinition<T, realm::js::RealmObject<T>> {
         {"linkingObjectsCount", wrap<linking_objects_count>},
         {"_isSameObject", wrap<is_same_object>},
         {"_objectKey", wrap<get_object_key>},
+        {"_tableKey", wrap<get_table_key>},
```
gagik, line 114: I am exposing `_tableKey` so objects can be stored in cache according to table and object keys; this is what we decided would be good when working on this in bindgen. Let me know if a different approach could be better.

CHANGELOG.md
```diff
@@ -82,7 +82,7 @@ Based on Realm JS v10.21.1: See changelog below for details on enhancements and
   ```
   * A typo was fixed in the `SubscriptionsState` enum, in which `SubscriptionsState.Superseded` now returns `superseded` in place of `Superseded`
 * `"discardLocal"` is now the default client reset mode. ([#4382](https://github.com/realm/realm-js/issues/4382))
-
+* Removed `Realm.JsonSerializationReplacer`. Use circular JSON serialization libraries such as [@ungap/structured-clone](https://www.npmjs.com/package/@ungap/structured-clone) and [flatted](https://www.npmjs.com/package/flatted) for stringifying Realm entities that have circular structures. The Realm entities' `toJSON` method returns plain objects and arrays (with circular references if applicable) which makes them compatible with any serialization library that supports stringifying plain JavaScript types. ([#4997](https://github.com/realm/realm-js/pull/4997))
```
gagik, line 85: Let me know if this could be worded better / any details missing.

gagik (commented):
Just some things worth noting.

lib/extensions.js
```diff
@@ -73,29 +92,31 @@ module.exports = function (realmConstructor) {
...
-  const getInternalCacheId = (realmObj) => {
-    const { name, primaryKey } = realmObj.objectSchema();
-    const id = primaryKey ? realmObj[primaryKey] : realmObj._objectKey();
-    return `${name}#${id}`;
-  };
+  Object.defineProperty(realmConstructor.Dictionary.prototype, "toJSON", {
+    value: function toJSON(_, cache = new Map()) {
+      const result = {};
+      for (const k of this._keys()) {
+        const v = this.getter(k);
```
gagik, line 99: this does not seem to be the cleanest approach to iterate through a dictionary keys/values... but I had a lot of trouble trying other ways.

CHANGELOG.md
```diff
@@ -82,9 +82,11 @@ Based on Realm JS v10.21.1: See changelog below for details on enhancements and
...
 * `"discardLocal"` is now the default client reset mode. ([#4382](https://github.com/realm/realm-js/issues/4382))
+* Removed `Realm.JsonSerializationReplacer`. Use circular JSON serialization libraries such as [@ungap/structured-clone](https://www.npmjs.com/package/@ungap/structured-clone) and [flatted](https://www.npmjs.com/package/flatted) for stringifying Realm entities that have circular structures. The Realm entities' `toJSON` method returns plain objects and arrays (with circular references if applicable) which makes them compatible with any serialization library that supports stringifying plain JavaScript types. ([#4997](https://github.com/realm/realm-js/pull/4997))
 
 ### Enhancements
 * Small improvement to performance by caching JSI property String object [#4863](https://github.com/realm/realm-js/pull/4863)
+* Small improvement to performance for `toJSON` which should make it useful for cases where a plain representations of Realm entities are needed, e.g. when inspecting them for debugging purposes through `console.log(realmObj.toJSON())`.
```
gagik, line 89: not sure if there is more to say / smth different to say.

types/index.d.ts
```diff
@@ -296,9 +296,9 @@ declare namespace Realm {
         entries(): [string, any][];
 
         /**
-         * @returns An object for JSON serialization.
+         * @returns A plain object for JSON serialization.
```
gagik, line 299: I thought it would be good to specify it is a plain object

CHANGELOG.md
kneth: I think it is detailed enough for our users.

src/js_realm_object.hpp
kneth: It makes a lot of sense as the pair (table key, object key) should be unique within a Realm file.

src/js_realm_object.hpp
kneth: That's an interesting observation, and I suggest that you bring it up with the team working on the bindgen project (core's public API is part of the project).

kneth (approved):

integration-tests/tests/src/tests/serialization.ts
```diff
@@ -19,543 +19,180 @@
...
+  // Expected serialized plain object result
+  serialized: Record<string, any>;
+}
 
-      return circularCollectionResultWithIds;
-    },
-  },
-];
+const commonTests: Record<string, TestSetup> = {
+  //@ts-expect-error subject and serialized are set before tests run.
```
kraenhansen, line 84: Can you share a bit more about this? It feels like this could be avoided.

lib/extensions.js
kraenhansen: For what it's worth, it seem equally bad as what you're removing from dictionary.js.

types/index.d.ts
```diff
@@ -296,9 +296,9 @@ declare namespace Realm {
...
          */
-        toJSON(): any;
+        toJSON(): Record<string, any>;
```
kraenhansen, line 301: Could this (easily) be the less permissive `unknown`? If yes, I'd opt for that.

```suggestion
        toJSON(): Record<string, unknown>;
```

types/index.d.ts
```diff
@@ -390,6 +383,11 @@ declare namespace Realm {
         addListener(callback: DictionaryChangeCallback): void;
         removeListener(callback: DictionaryChangeCallback): void;
         removeAllListeners(): void;
+
+        /**
+         * @returns A plain object for JSON serialization.
+         */
+        toJSON(): Record<string, any>;
```
kraenhansen, line 390:
```suggestion
        toJSON(): Record<string, unknown>;
```

types/index.d.ts
```diff
@@ -401,9 +399,9 @@ declare namespace Realm {
         readonly optional: boolean;
 
         /**
-         * @returns An object for JSON serialization.
+         * @returns An array of plain objects for JSON serialization.
          */
-        toJSON(): Array<any>;
+        toJSON(): Array<Record<string, any>>;
```
kraenhansen, line 404:
```suggestion
        toJSON(): Array<Record<string, unknown>>;
```

integration-tests/tests/src/tests/serialization.ts
gagik: so, I can either make the subject and serialized fields as optional in which case I won't have this TS error here but then I will have to try to avoid all errors where I use subject but it could be undefined (even though it will never be undefined because it will be set before each test). This was my way of avoiding this but I can see how this is hacky, I can look into alternative

gagik (commented):
I just realized this PR seems to be updating `vendor/real-core`? That's odd.

kraenhansen (approved):

PR #4999, opened 2022-10-11
Update types for min, max, avg & sum

### LINKED ISSUE #4994: Aggregation functions return `undefined` when they should return `null`

Seen once

### Description

Our types declare that the `min`, `max` and `sum` can return `null`:

```typescript
min(property?: string): number | Date | null;
max(property?: string): number | Date | null;
sum(property?: string): number | null;
avg(property?: string): number;
```

But the implementation shows that `min`, `max` and `avg` actually return `undefined` when there are no objects in the collection:

```typescript
const realm = new Realm({
  schema: [{
    name: "Item",
    properties: { value: "int" }
  }],
});
realm.objects("Item").min("value"); // returns `undefined`
realm.objects("Item").max("value"); // returns `undefined`
realm.objects("Item").avg("value"); // returns `undefined`
realm.objects("Item").sum("value"); // returns `0`
```

We are actually testing, to ensure this behaviour ☝️
https://github.com/realm/realm-js/blob/master/tests/js/list-tests.js#L1389-L1392

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

_No response_

### Version

10.21.1

Local Database only

### Are you using encryption?

No, not using encryption

Tested on Node.js

### Build environment

_No response_

### Cocoapods version

_No response_

kraenhansen:
## What, How & Why?

This closes #4994.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


kneth (approved):

poonamdhomane:
HOw to use this operator without grouping ?

PR #5020, opened 2022-10-18
Add Realm listeners

### LINKED ISSUE #5012: Implement listeners on Realm instances

gagik:
## What, How & Why?
Adds Realm listeners and new integrationt tests for it to bindgen.

This closes #5012.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm/src/Realm.ts
```diff
@@ -50,7 +50,9 @@ import {
   normalizeObjectSchema,
   toArrayBuffer,
 } from "./internal";
+import { BindingListeners } from "./Listeners";
```
kraenhansen, line 53:
```suggestion
  BindingListeners
} from "./internal";
```

packages/realm/src/Listeners.ts
```diff
@@ -54,3 +56,43 @@ export class Listeners<CallbackType> {
     this.listeners.clear();
   }
 }
+
+/** @internal */
+type BindingListenerCallback = (r: Realm, name: string, schema?: Realm.ObjectSchema[]) => void;
+/** @internal */
+export class BindingListeners {
```
kraenhansen, line 63:
```suggestion
export class RealmListeners {
```

I'd suggest this class gets a file of its own 👍

packages/realm/src/Listeners.ts
```diff
@@ -54,3 +56,43 @@ export class Listeners<CallbackType> {
...
+  /**
+   * Keeps tracked of registered listener callbacks for Realm class notifications.
+   */
+  constructor(private name: string) {}
```
kraenhansen, line 67: I'd suggest a name closer to the semantics.

```suggestion
  constructor(private eventName: string) {}
```

Since this is passed to the callback as an argument and it has only three applicable values, I'd suggest using a string-based enum for this.

packages/realm/src/Listeners.ts
```diff
@@ -54,3 +56,43 @@ export class Listeners<CallbackType> {
...
+  private listeners = new Set<BindingListenerCallback>();
+
+  // Combined callback which runs all listener callbacks in one call.
+  callback(realm: Realm): void {
```
kraenhansen, line 71: Would it make sense for this `realm` to be a private member of the class instead? It will always get the same Realm passed in as callback, we might as well make that explicit.

packages/realm/src/Realm.ts
```diff
@@ -511,19 +530,44 @@ export class Realm {
     });
   }
 
-  addListener(): unknown {
+  addListener(name: string, callback: RealmListenerCallback): void {
```
kraenhansen, line 533:
```typescript
enum RealmEventName {
  Change = "change",
  Schema = "schema",
  BeforeNotify = "beforenotify",
}
```

```suggestion
  addListener(name: RealmEventName, callback: RealmListenerCallback): void {
```

integration-tests/tests/src/tests/dynamic-schema-updates.ts
```diff
@@ -18,40 +18,39 @@
...
+    expect(this.realm.schema).to.be.an("array");
     // There is a function defined on the Realm
-    expect(realm._updateSchema).to.be.a("function");
+    expect(this.realm._updateSchema).to.be.a("function");
     // Expect no enumerable field on the schema property
-    expect(Object.keys(realm.schema)).to.not.contain("update");
+    expect(Object.keys(this.realm.schema)).to.not.contain("update");
     // TODO: This function gets put on the schema to
-    // expect(realm.schema.update).to.be.a("function");
+    // expect(this.realm.schema.update).to.be.a("function");
```
gagik, line 38: is this something that is planned to be added?

packages/realm/src/Realm.ts
```diff
@@ -49,6 +49,9 @@ import {
   ClassHelpers,
   normalizeObjectSchema,
   toArrayBuffer,
+  RealmListeners,
+  RealmEventName,
```
gagik, line 53: not sure if you have thought about this being called `RealmEventName`. I imagine `RealmListenerEventName` seems redundant/weird since listener listens to events

packages/realm/src/RealmListeners.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type RealmEventName = "change" | "schema" | "beforenotify";
+enum RealmEvent {
+  Change = "change",
+  Schema = "schema",
+  BeforeNotify = "beforenotify",
+}
+export type RealmListenerCallback = (r: Realm, name: RealmEventName, schema?: Realm.ObjectSchema[]) => void;
+
+// Temporary functions to work between event names and corresponding enums.
```
gagik, line 29: this is what I mean by it feels a little unnecessary to do all this just to store the name as an enum but if we do not want to change user-facing API this is useful for future change.

packages/realm/src/RealmListeners.ts
```diff
@@ -0,0 +1,92 @@
...
+      // No need to add a listener twice
+      return;
+    }
+    // Store the listener.
+    this.listeners.add(callback);
+  }
+
+  remove(callback: RealmListenerCallback): void {
+    // TODO: consider whether bool of delete is useful or not.
+    this.listeners.delete(callback);
```
gagik, line 86: The listener return a bool if it exists or not, this seems like more useful than the legacy `void` function in some scenarios (same could apply to add...) but I am leaving it this way for consistency with old API.

integration-tests/tests/src/tests/listeners.ts
```diff
@@ -0,0 +1,173 @@
...
+      });
+    });
+    it("should work with multiple listeners of different types", function (this: RealmContext, done) {
+      let changeListenerRuns = 0;
+      let schemaListenerRuns = 0;
+      this.realm.addListener("change", () => {
+        changeListenerRuns += 1;
+      });
+      this.realm.addListener("beforenotify", () => {
+        changeListenerRuns += 1;
```
kraenhansen, line 110: This looks incorrect.

packages/realm/src/RealmListeners.ts
kraenhansen: You could do something a bit more reflective like this: https://www.typescriptlang.org/play?#code/KYOwrgtgBASsCGAbCBRAbqALgOXhYUA3gFBRQDCAFvCAOYEC8UARAMbV3DMA0pUAyu2AR4UJswDOQkTz4AhYADMA9gCdg2ZZgCWigJ5iWAIyVrQW3XtkBfYsUVgQrHcpBRtE9Flz4AFCDxgAC4oCUxVbToAShD4CQlgVUwJKAD8dxS4JFQMEBxAoj5dKF8AQgB5IwArYGcAOjQkMGAJXyzkLzyfYCi6yNZEMAATFv9AqKjCsjJMSlVlAHdU4CWUVXnVXwADAFUQYAAPAAdazGAhqGBczFSCgHIAEkI04Gs7raiAbj5bW3tHZzaVxQFTKXxXLAhMIRaJTDKdfJ+CF5L58IzwTbIzCov4OJwuNzozHXELtHLecZw1iuCTKRDAOqIZS0XzMIlQVhIekXBbaWY8S7XHF2UGs9g0ejMVGi5gLeZ0KXfIA

packages/realm/src/RealmListeners.ts
```diff
@@ -0,0 +1,92 @@
...
+  // Combined callback which runs all listener callbacks in one call.
+  callback(): void {
+    let schema: Realm.ObjectSchema[] | undefined;
+    if (this.eventType === RealmEvent.Schema) {
+      schema = this.realm.schema;
+    }
+    const name = nameFromEvent(this.eventType);
+    for (const callback of this.listeners) {
+      callback(this.realm, name, schema);
+    }
```
kraenhansen, line 72: Since your enum values are strings, you should be able to pass them directly.

```suggestion
    for (const callback of this.listeners) {
      callback(this.realm, this.eventType, schema);
    }
```

kraenhansen (approved):
Minor suggestions.

integration-tests/tests/src/tests/listeners.ts
gagik: that's true... interestingly enough this actually does not affect tests directly. `beforenotify` is ran AFTER the checks so this is still running correctly in terms of checking that both `schema` and `change` are run... `beforenotify` runs after that, so this addition is meaningless. which is a little odd to me, I am still quite confused by what the point of this event is...

packages/realm/src/RealmListeners.ts
gagik: I think this assertion already ends up being indirectly being done by the Real as it ends up looking through all possible values before running this function for the specific one. If you mean for constructing these listener objects, they get initialised in the Realm with constant string, so Typescript ends up asserting types anyways.

PR #5025, opened 2022-10-20
Add better documentation for bindgen.

gagik:
## What, How & Why?
Adds better documentation for methods and fields of Realm instance classes. I came up with some principles that decisions were made when I was not sure how to organize, prioritising consistency with legacy documentation as well as clear formatting from language server when viewed from VSCode.

- General prioritisation order of comments: `@params > @throws > @returns > @example > @since`. 
- All tags start with uppercase i.e. `@returns The sum`. The only exception is if it is referring to boolean / other references that happen to lowercase, i.e. @returns `true` if...
- No tags are followed by a dash (this was in legacy docs and is actually ignored by JSDoc but for sake of consistency this is removed).
- No tags other than `@throws` should specify types since this is TypeScript.
- `@throws` is followed by Error type (i.e. `{Error}`) then either If or When.
- `@returns` is followed by The / A if applicable. If it returns a boolean it follows the format: @returns `true` if X, `false` if not.
- `@since` is kept since it would be easier to remove it then add it back.
- I tried to replace links with italicised bold text i.e.` {@link Realm.Bold bold} => ***bold***`. This is because the links in i.e. VSCode were not useful and just led to the source code when clicked which did not seem appropriate in this context. In my opinion best developer experience would be for the links to link to the documentation website for the specific function. I did not see italicised bold text used before, so adding this specific styling to some words will hopefully make it easier to go through them and consider re-adding links in the future. I do think there are other instances where a link to doc would be useful so this is not exhaustive.
- All methods that override built-in JS parent class methods (i.e. `forEach`, `map`) have not been commented as they already have built-in documentation that is inherited.

Let me know if any of these "rules" seem unreasonable and I could change it, they are mostly following the old docs.

<!This closes # ??? -- link to an existing issue -->

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/Collection.ts
+++ b/packages/realm/src/Collection.ts
@@ -18,6 +18,22 @@
 
 import { CallbackRegistrator, IllegalConstructorError, Listeners, binding } from "./internal";
 
+/**
+ * Abstract base class containing methods shared by Realm **List**, **Dictionary**, and **Results**.
+ *
+ * A Realm Collection is a homogenous sequence of values of any of the types
+ * that can be stored as properties of Realm objects. A collection can be
+ * accessed in any of the ways that a normal Javascript Array can, including
+ * subscripting, enumerating with `for-of` and so on.
+ *
+ * A Collection always reflect the current state of the Realm. The one exception to this is
+ * when using `for...in` or `for...of` enumeration, which will always enumerate over the
+ * objects which matched the query when the enumeration is begun, even if some of them are
+ * deleted or modified to be excluded by the filter during the enumeration.
+ *
+ * @memberof Realm
+ * @since 0.11.0
+ */
 export abstract class Collection<
   KeyType = unknown,
   ValueType = unknown,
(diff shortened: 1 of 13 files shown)


packages/realm/src/Realm.ts
```diff
@@ -132,9 +156,12 @@ export class Realm {
 
   /**
    * Get the current schema version of the Realm at the given path.
-   * @param  {string} path
-   * @param  {any} encryptionKey?
-   * @returns number
+   * @param path - The path to the file where the
```
kraenhansen, line 159: Not at a computer now, but I'm curious if the dash after param name is ignored? If not, it should probably go away.

packages/realm/src/Realm.ts
```diff
@@ -478,6 +632,15 @@ export class Realm {
     }
   }
 
+  /**
+   * Searches for a Realm object by its primary key.
+   * @param type - The type of Realm object to search for.
+   * @param primaryKey - The primary key value of the object to search for.
+   * @throws If type passed into this method is invalid or if the object type did
```
kraenhansen, line 639: I'm wondering if it make sense to pick an order of these annotations? So `@return` is always before `@throw` and perhaps we could add an eslint plugin for that? Just lower the possible variations among these comments.

packages/realm/src/Realm.ts
```diff
@@ -478,6 +632,15 @@ export class Realm {
...
+   *   not have a `primaryKey` specified in its {@link ObjectSchema}.
```
kraenhansen, line 640: Does this link resolve correctly?

gagik: links seem to work weirdly... they don't really link to meaningful places (they do link to code) so I saw thinking of removing them.

packages/realm/src/Realm.ts
gagik: The dash does get ignored but It could be good to make this consistent anyways.

packages/realm/src/App.ts
```diff
@@ -18,20 +18,56 @@
 
 import { binding, fs } from "./internal";
 
+/**
+ * This describes the options used to create a Realm.App instance.
+ * @property id - The id of the Atlas App Services application.
+ * @property baseUrl - The base URL of the Atlas App Services server.
+ * @property timeout - General timeout (in millisecs) for requests. TODO
```
gagik, line 25: Not sure if these fields are going to be added in the future.

packages/realm/src/Collection.ts
```diff
@@ -44,20 +59,83 @@ export abstract class Collection<
...
+
+  /**
+   * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/entries Array.prototype.keys}
+   * @returns Iterator with all key/value pairs in the collection
+   * @since 0.11.0
+   */
   abstract entries(): Iterable<EntryType>;
+
+  /**
+   * This is the same method as the {@link Collection.values} method.
```
gagik, line 84: perhaps I should remove all the links... they would be nice if they linked to the documentation website... but the way they are now they will link to the source... which isn't very useful.

packages/realm/src/Dictionary.ts
```diff
@@ -98,7 +96,12 @@ const PROXY_HANDLER: ProxyHandler<Dictionary> = {
 };
 
 /**
- * TODO: Make this extends Collection<T> (once that doesn't have a nummeric index accessor)
```
gagik, line 101: I'm assuming these TODOs are done?

packages/realm/src/List.ts
```diff
@@ -163,8 +199,49 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
     return internal.size;
   }
 
+  /** TODO
+   * Changes the contents of the list by removing value and/or inserting new value.
```
gagik, line 203: I am not sure how to handle multiple implementations of functions. One would hope they would inherit from one another... but they don't...

packages/realm/src/Realm.ts
gagik: The way old docs were made, return was always after the throw, so I kept it this way, generally I kept this order:
param > throws > return > since

packages/realm/src/Realm.ts
```diff
@@ -531,6 +700,15 @@ export class Realm {
     });
   }
 
+  /**
+   * Add a listener `callback` for the specified event `name`.
+   * @param eventName - The name of event that should cause the callback to be called.
+   *   DISCUSS: why does it say _Currently, only the "change" and "schema" events are supported_.
```
gagik, line 706: current docs say only change and schema are supported. Docs website also doesn't mention beforenotify still. Is this intentional or just outdated?

gagik: so I learnt now that beforeNotify was originally exposed in JS for primarily internal use, which is probably why the docs are like this. Perhaps it's worth not to mention beforeNotify at all not to confuse people about this niche event

gagik:
There are still some changes to be made but seems like good time to discuss things.

packages/realm/src/Dictionary.ts
```diff
@@ -240,7 +249,10 @@ export class Dictionary<T = unknown> extends Collection<string, T, [string, T],
   }
 
   /**
-   * @returns A plain object for JSON serialization.
+   * The plain object representation of the Dictionary for JSON serialization.
+   * Use circular JSON serialization libraries such as {@link https://www.npmjs.com/package/@ungap/structured-clone @ungap/structured-clone}
```
gagik, line 253: I am a little biased but I thought it'd be nice to mention this in the function comment itself. I imagine stringifying circular objects is a common enough usecase that would be a much better user experience to find out through simple IDE tips rather than going deep into docs (which don't exist yet anyways) or changelogs.

packages/realm/src/App.ts
kraenhansen: Most likely.

packages/realm/src/Dictionary.ts
kraenhansen: Yep.

kraenhansen:
@gagik if you find a moment, it would be awesome to get this rebased and merged 🤞

packages/realm/src/OrderedCollection.ts
```diff
@@ -364,14 +373,37 @@ export abstract class OrderedCollection<T = unknown, EntryType extends [unknown,
...
+  /**
+   * Returns the minimum value of the values in the collection or of the
+   * given property among all the objects in the collection, or `undefined`
+   * if the collection is empty.
+   *
+   * Only supported for int, float, double and date properties. `null` values
+   * are ignored entirely by this method and will not be returned.
+   *
+   * @param property For a collection of objects, the property to take the minimum of.
+   * @throws {Error} If no property with the name exists or if property is not numeric/date.
```
kraenhansen, line 403: Not a blocker, but this could be more specific.

```suggestion
   * @throws {TypeAssertionError} If no property with the name exists or if property is not numeric/date.
```

Issue #5006, opened 2022-10-13
Recent builds are not working with RHEL7


kneth:
### How frequently does the bug occur?

All the time

### Description

Version 10.21.0 doesn't work on older Linux systems.

### Stacktrace & log output

```shell
node:internal/modules/cjs/loader:1183
  return process.dlopen(module, path.toNamespacedPath(filename));
                 ^

Error: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /root/test/node_modules/realm/build/Release/realm.node)
    at Object.Module._extensions..node (node:internal/modules/cjs/loader:1183:18)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:94:18)
    at bindings (/root/test/node_modules/bindings/bindings.js:112:48)
    at getRealmConstructor (/root/test/node_modules/realm/lib/index.js:28:37)
    at Object.<anonymous> (/root/test/node_modules/realm/lib/index.js:53:26)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10) {
  code: 'ERR_DLOPEN_FAILED'
}
```

### Can you reproduce the bug?

Yes, always

### Reproduction Steps

`const Realm = require("realm");`

### Version

10.21.0

### What SDK flavour are you using?

Local Database only

### Are you using encryption?

No, not using encryption

### Platform OS and version(s)

RHEL7

### Build environment

Which debugger for React Native: ..

crazielips:
Same problem

kneth:
We have merged a PR which will enable builds for older versions of glibc in order to support Centos/RHEL 7. Binaries will be released from next release.

nickbeaulieu:
Will the fix for this issue be included in version `v10.24.0` or only in version `v11.2.0`? It is only mentioned in the major version 11 [release notes](https://github.com/realm/realm-js/releases/tag/v11.2.0) and not the major version 10 [release notes](https://github.com/realm/realm-js/releases/tag/v10.24.0). 

We had a report of this crash from a few of our users on RHEL.

nickbeaulieu:
Sorry for the delay.

[...]

PR #5036, opened 2022-10-26
Do Linux builds in CentOS 7 container

iamer:
Use the CentOS 7 based container to build for Linux x64. This is to avoid linking to newer Glibc versions causing errors on older Linux distros as seen in https://github.com/realm/realm-js/issues/5006

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -1,4 +1,4 @@
-name: pr-realm-js
+name: Pull request build and test
 on:
   push:
     branches:
@@ -9,6 +9,8 @@ jobs:
   build:
     name: Build for ${{ matrix.variant.os }} ${{ matrix.variant.arch }}
     runs-on: ${{ matrix.variant.runner }}
+    # if container is not set for a variant, this is a noop
+    container: ${{ matrix.variant.container }}
     env:
       REALM_DISABLE_ANALYTICS: 1
       NDK_VERSION: 21.0.6113669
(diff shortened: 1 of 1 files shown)


armhf.Dockerfile
```diff
@@ -22,15 +23,12 @@ ENV NVM_DIR /tmp/.nvm
...
  && chmod a+rwX -R $NVM_DIR
 
 # Ensure a new enough version of CMake is available.
 RUN cd /opt \
     && curl -O -J https://cmake.org/files/v3.15/cmake-3.15.2-Linux-x86_64.tar.gz \
     && tar zxf cmake-3.15.2-Linux-x86_64.tar.gz
 
-ENV PATH "/opt/cmake-3.15.2-Linux-x86_64/bin:$PATH"
\ No newline at end of file
+ENV PATH "/opt/cmake-3.15.2-Linux-x86_64/bin:$PATH"
```
kneth, line 34: Can we use the same version as x86_64?

kneth (commented):
If we move the ARM64 docker image to Debian 11, we can install gcc 10 which will open C++20.

kraenhansen (approved):

armhf.Dockerfile
iamer: I can look into both

takameyer (approved):

takameyer:
It seems the CI is failing.

iamer:
Yeah I was experimenting with doing the container builds in the PR workflow. I will revert that now.

iamer:
I reverted the experiment. So this workflow still depends on #5038

armhf.Dockerfile
iamer: IMHO, moving away from CentOS7 should be the target, not towards it..

kneth (approved):

PR #5038, opened 2022-10-27
Workflow to build and publish Linux containers for realm-js builds

iamer:
A manually triggered github actions workflow is used to build and publish
2 different container images:
* CentOS 7 based container to build binaries that don't depend on recent Glibc.
* Debian 10 based container to cross compile ARMHF binaries

### DIFF
--- a/.github/workflows/publish-containers.yml
+++ b/.github/workflows/publish-containers.yml
@@ -1,0 +1,47 @@
+name: Create and publish container images
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '**Dockerfile'
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - dockerfile: ./Dockerfile
+            image: ghcr.io/${{ github.repository }}/linux-x64
+          - dockerfile: ./armhf.Dockerfile
+            image: ghcr.io/${{ github.repository }}/linux-arm
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Log in to the Container registry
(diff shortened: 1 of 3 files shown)


kneth (commented):
Is it possible to automate it? If either `Dockerfile` or `amhf.Dockerfile` is changes, a new container is build and published?

kraenhansen:
What's the advantage of publishing? Is it unreasonable for the consuming workflow to run `docker build`? I'm asking because the latter would allow us to create a PR with adjustments to the Dockerfiles and see the results of that, without having to publish new images.

kraenhansen:
Oh - I reviewed the other PR now and understand that this is used via the `container` property on the other workflow.
An alternative would be to move the building of the image into the pr-workflow and publish it using the commit sha as tag? Perhaps it's possible to avoid pushing to the registry if it all happens in the same job?

iamer:
> What's the advantage of publishing? Is it unreasonable for the consuming workflow to run `docker build`? I'm asking because the latter would allow us to create a PR with adjustments to the Dockerfiles and see the results of that, without having to publish new images.

At the very least it saves 4 minutes, unless we also publish in the consuming workflow and use some cache like mechanism.

iamer:
The workflow without publishing would also not be the best optimized approach.
https://github.com/docker/build-push-action/blob/master/docs/advanced/share-image-jobs.md

iamer:
I can experiment with doing the build / publish as part of the other workflow with caching

kraenhansen (approved):

kraenhansen:
I'm okay with the way it is now :+1: At least it's publishing the the GCR linked to this repo and it might be useful to be able to pull when building locally in the future.

iamer:
So I tried, and bumped into a familiar issue. publishing from pull request workflows requires special permissions or using https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

takameyer (approved):

kneth (approved):

PR #5057, opened 2022-11-02
Adding a couple of object listener tests

kraenhansen:
## What, How & Why?

I wanted to ensure I understood the behaviour of the existing SDK around object listeners (and their removal), so I added a few tests to the integration tests. I also updated the object tests to use the `openRealmBefore` hook.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -48,7 +48,7 @@ describe("Realm objects", () => {
         expect(firstPerson).deep.equals(john);
       });
 
-      it("can have it's properties read", function (this: RealmContext) {
+      it("can have its properties read", function (this: RealmContext) {
         const { realm } = this;
 
         const john = realm.write(() => {
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -290,7 +290,7 @@ describe("Realm objects", () => {
 
     it("fires and supports removal across accessor objects", async function (this: RealmContext) {
       const alice = getAlice(this.realm);
-      const calls = [createPromiseHandle(), createPromiseHandle(), createPromiseHandle(), createPromiseHandle()];
+      const calls = [createPromiseHandle(), createPromiseHandle()];
       // Add a listener
       let callCount = 0;
       function listener() {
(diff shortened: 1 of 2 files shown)


integration-tests/tests/src/tests/objects.ts
```diff
@@ -24,138 +24,147 @@ import {
...
+        // Expect John to be the one and only result
+        const persons = realm.objects(PersonSchema.name);
+        expect(persons.length).equals(1);
+        const [firstPerson] = persons;
+        expect(firstPerson).deep.equals(john);
+      });
 
-    it("can have it's properties read", () => {
-      const realm = new Realm({ schema: [PersonSchema] });
+      it("can have it's properties read", function (this: RealmContext) {
```
takameyer, line 51:
```suggestion
      it("can have its properties read", function (this: RealmContext) {
```

integration-tests/tests/src/tests/objects.ts
```diff
@@ -24,138 +24,147 @@ import {
...
+        // Expect only one instance of 'PersonSchemaWithId' in db after all updates
+        const persons = realm.objects(PersonSchemaWithId.name);
+        expect(persons.length).equals(1);
+      });
     });
   });
 
   describe("Class Model", () => {
+    beforeEach(() => {
+      Realm.clearTestState();
```
takameyer, line 165: I guess this could also be abstracted into a beforeHook?

integration-tests/tests/src/tests/objects.ts
```diff
@@ -257,4 +266,86 @@ describe("Realm objects", () => {
...
+      const result = realm.objectForPrimaryKey<IPerson>("Person", ALICE_NAME);
+      if (!result) {
+        throw new Error("Expected an object");
+      }
+      return result;
+    }
+
+    it("fires and supports removal across accessor objects", async function (this: RealmContext) {
+      const alice = getAlice(this.realm);
+      const calls = [createPromiseHandle(), createPromiseHandle(), createPromiseHandle(), createPromiseHandle()];
```
takameyer, line 293: We only need 4 promise handles

takameyer (approved):
Just a couple small things.  Merge at will.

integration-tests/tests/src/tests/objects.ts
kraenhansen: I though the same, but choose not to, since I want the test to fail and not the before-hook, in case the class based models feature breaks.

integration-tests/tests/src/tests/objects.ts
kraenhansen:
```suggestion
      const calls = [createPromiseHandle(), createPromiseHandle()];
```

kraenhansen:

PR #5065, opened 2022-11-03
Indirect weak map and cache

kraenhansen:
## What, How & Why?

This adds two data structures we could add to improve the developer experience.

### `ImplicitWeakMap`

A map from some type of object (the key) into another type of object (the value), where a function (the hasher supplied at construction) is called to derive a hash of the key, which is used when looking up the value.
This makes it possible for multiple different key objects to get the same value object.
The map is considered weak in the sense that values are wrapped in a `WeakRef` before being inserted in the underling map. A value is also registered with a finalization registry, ensuring that their entry in the underlying map is removed when they get garbage collected, in an effort to make the entire `IndirectWeakMap` avoid leaks.

### `ImplicitWeakCache`

A cache of objects (the value) which can either be constructed on demand or retrieved from cache.
The cache is considered weak as it uses the `IndirectWeakMap` to store its values, making them available for garbage collection.

### DIFF
--- a/packages/realm/src/IndirectWeakCache.ts
+++ b/packages/realm/src/IndirectWeakCache.ts
@@ -16,13 +16,28 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import { IndirectWeakMap } from "./internal";
+import { HashFunction, IndirectWeakMap } from "./internal";
 
-export class WeakCache<K extends { $addr: bigint }, V extends object, Args extends unknown[]> {
-  private map = new IndirectWeakMap<K, V, bigint>(({ $addr }) => $addr);
+/**
+ * A cache of objects (the value) which can either be constructed on demand or retrieved from cache.
+ * The cache is considered weak as it uses the `IndirectWeakMap` to store it's values, making them available
+ * for garbage collection.
+ * @internal
+ */
+export class IndirectWeakCache<K extends object, V extends object, Args extends unknown[], H = unknown> {
+  private map: IndirectWeakMap<K, V, H>;
 
-  constructor(private ctor: { new (...args: Args): V }) {}
-
+  constructor(private ctor: { new (...args: Args): V }, hasher: HashFunction<K, H>) {
+    this.map = new IndirectWeakMap<K, V, H>(hasher);
+  }
+  /**
+   * Get an existing value from the cache or construct and store one in case of a miss.
+   * @param key Object passed to the hasher provided at construction of the cache.
+   * @param args An optional array of constructor arguments can be passed as well, which will be used in case of a cache miss
+   * to construct and store a new value object.
+   * @returns An existing or new value.
+   * @throws If `args` are not supplied and no object existed in the cache.
+   */
   get(key: K, args?: Args) {
     const existing = this.map.get(key);
     if (existing) {
--- a/packages/realm/src/IndirectWeakCache.ts
+++ b/packages/realm/src/IndirectWeakCache.ts
@@ -20,7 +20,7 @@ import { HashFunction, IndirectWeakMap } from "./internal";
 
 /**
  * A cache of objects (the value) which can either be constructed on demand or retrieved from cache.
- * The cache is considered weak as it uses the `IndirectWeakMap` to store it's values, making them available
+ * The cache is considered weak as it uses the `IndirectWeakMap` to store its values, making them available
  * for garbage collection.
  * @internal
  */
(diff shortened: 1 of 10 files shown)


packages/realm/src/IndirectWeakCache.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  /**
+   * Get an existing value from the cache or construct and store one in case of a miss.
+   * @param key Object passed to the hasher provided at construction of the cache.
+   * @param args An optional array of constructor arguments can be passed as well, which will be used in case of a cache miss
+   * to construct and store a new value object.
+   * @returns An existing or new value.
+   * @throws If `args` are not supplied and no object existed in the cache.
+   */
+  get(key: K, args?: Args) {
```
blagoev, line 41: Sometimes they call these `getOrCreate`

kraenhansen: Yeah - I think you're right. It would be good to signal the potential side-effect 👍

packages/realm/src/IndirectWeakCache.ts
```diff
@@ -0,0 +1,56 @@
...
+import { HashFunction, IndirectWeakMap } from "./internal";
+
+/**
+ * A cache of objects (the value) which can either be constructed on demand or retrieved from cache.
+ * The cache is considered weak as it extends the `IndirectWeakMap` to store its values, making them available
+ * for garbage collection.
+ * @internal
+ */
+export class IndirectWeakCache<
+  K extends object,
```
takameyer, line 28: In my opinion, we should start naming our generics.  I always have to read the implementation to deduce what these letters actually mean.  Thoughts?

kraenhansen: I agree .. ol' habits stick around.
I pushed a commit with some less abbreviated names 🙂

RedBeard0531: Meh. I think K and V are pretty classic abbreviations for Key and Value. And I don't think the Type suffix adds any value, both because they are capitalized, and because types are the only thing that TS supports as generic arguments (even things that look "valuey" like `true` and `"hello"` are promoted to "literal types" and are still considered types).

I agree that `Args` and `Hash` are good names though. But we don't need to go all Java Enterprise Edition on all of our names.

kraenhansen: Time to flip a 🪙 ? I don't feel strong either way.

packages/realm/src/IndirectWeakMap.ts
```diff
@@ -0,0 +1,76 @@
...
+    const existingRef = this.values.get(hash);
+    if (existingRef) {
+      this.registry.unregister(existingRef);
+    }
+    // Register the new value with the finalization registry, to prune its WeakRef from
+    // the map of values.
+    this.registry.register(value, hash, ref);
+    this.values.set(hash, ref);
+    return this;
+  }
```
RedBeard0531, line 61: This isn't a valid way to use a hash. It is legal for two semantically non-equal objects (meaning "real" equal, not the broken js `==` address equality check) to have the same hash. You seem to be treating the hash as if it is a bijection, and that if the hashs match, the backing objects must be semantically equal.

Perhaps the problem is just naming? If the behavior you want is that the function returns a value that can be used to uniquely identify the value in a map, then the conventional name for that is a "Key Function"

packages/realm/src/IndirectWeakCache.ts
takameyer: I mean, i'm not gonna die on this hill, but i've always felt being more explicit is better.  As long as we don't start using T, K, V for anything else that Type, Key and Value, i'm fine with it.  Just be explicit where it matters.

RedBeard0531:
My 2¢: I don't think we should do this right now. I don't think any other SDK does this (based on slack discussion) and the current JS SDK doesn't either. Adding weak maps for this use case is biting off a lot of complexity budget and I don't think it pays for itself. I'd rather we focus on reaching feature parity, and cross-platform support prior to tackling things like this. We can always add this later, even after we release V12.

GH comments probably isn't the best place to discuss this. Happy to discuss in person or on zoom.

packages/realm/src/IndirectWeakMap.ts
kraenhansen: Yeah. It's probably the naming. By hash I simply mean a one-way function. This is why I named it `IndirectWeakMap` instead of `HashWeakMap`.

The hasher is the function that produce the value which is used as the basis of equality checks. Do you have a better name for that?

kraenhansen: Perhaps `id` is better than `hash` in this case?

PR #5066, opened 2022-11-03
Finish `realm.delete`

takameyer:
Add support for delete results and lists
Added tests for delete list of primitives

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/Realm.ts
+++ b/packages/realm/src/Realm.ts
@@ -511,8 +511,11 @@ export class Realm {
       );
       const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
       table.removeObject(obj.key);
+    } else if (subject instanceof List) {
+      subject.internal.deleteAll();
+    } else if (subject instanceof Results) {
+      subject.internal.clear();
     } else if (Array.isArray(subject) || Symbol.iterator in subject) {
-      // TODO: Optimize this to not get the helper on every iteration
       for (const object of subject) {
         assert.instanceOf(object, RealmObject);
         const { objectSchema } = this.classes.getHelpers(object);
@@ -520,7 +523,7 @@ export class Realm {
         table.removeObject(object[INTERNAL].key);
       }
     } else {
-      throw new Error("Not yet implemented");
+      throw new Error("Can only delete objects, lists and results.");
     }
   }
 
--- a/packages/bindgen/spec.yml
+++ b/packages/bindgen/spec.yml
@@ -158,7 +158,7 @@ enums:
     values:
       TopLevel: 0
       Embedded: 1
-      TopLevelAsymmetric: 0x2 
+      TopLevelAsymmetric: 0x2
 
   DataType:
     values:
@@ -283,13 +283,13 @@ records:
       link_origin_property_name:
         type: std::string
         default: ""
-      is_primary: 
+      is_primary:
         type: bool
         default: false
       is_indexed:
         type: bool
         default: false
-      column_key: 
+      column_key:
         type: ColKey
         default: {}
   VersionID:
@@ -407,7 +407,7 @@ records:
 
   SyncClientTimeouts:
     fields:
-      connect_timeout: 
+      connect_timeout:
         type: uint64_t
         default: 120000
       connection_linger_time:
@@ -475,10 +475,10 @@ records:
     cppName: app::Request
     fields:
       method: HttpMethod
-      url: std::string 
+      url: std::string
       timeout_ms: uint64_t
       headers: std::map<std::string, std::string>
-      body: std::string 
+      body: std::string
       uses_refresh_token: bool
 
   Response:
@@ -604,7 +604,7 @@ classes:
       # apply_additive_changes: '(group: Group&, changes: std::vector<SchemaChanges>, update_indexes: bool)'
       # table_for_object_type
       # table_for_Schema_type
-      
+
       schema_from_group: '(group: Group) -> std::vector<ObjectSchema>'
 
       # some more...
@@ -680,6 +680,7 @@ classes:
       min: '(column: ColKey) -> util::Optional<Mixed>'
       average: '(column: ColKey) -> util::Optional<Mixed>'
       sum: '(column: ColKey) -> util::Optional<Mixed>'
+      clear: ()
 
       add_notification_callback: '(cb: (changes: const CollectionChangeSet&), keyPaths: KeyPathArray) -> NotificationToken'
 
@@ -714,7 +715,7 @@ classes:
 
       can_deliver_notifications: bool
       scheduler: std::shared_ptr<Scheduler>
-      
+
       is_closed: bool
 
       audit_context: AuditInterface*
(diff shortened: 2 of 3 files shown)


packages/realm-app-importer/src/DeleteAll.ts
```diff
@@ -0,0 +1,77 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+import fetch from "node-fetch";
+
+const getApps = async (headers: any) => {
```
kraenhansen, line 20: Did you mean to add this? If yes, I suggest moving this to a different PR.

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,19 @@ export class Realm {
...
-    } else if (Array.isArray(subject) || Symbol.iterator in subject) {
+    } else if (Array.isArray(subject) || subject instanceof Results) {
       // TODO: Optimize this to not get the helper on every iteration
       for (const object of subject) {
         assert.instanceOf(object, RealmObject);
         const { objectSchema } = this.classes.getHelpers(object);
         const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
         table.removeObject(object[INTERNAL].key);
       }
+    } else if (subject instanceof List) {
```
kraenhansen, line 522: I would move this branch up before the `Array.isArray`.

packages/realm-app-importer/src/DeleteAll.ts
takameyer: Nope

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,19 @@ export class Realm {
       );
       const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
       table.removeObject(obj.key);
-    } else if (Array.isArray(subject) || Symbol.iterator in subject) {
+    } else if (Array.isArray(subject) || subject instanceof Results) {
```
kraenhansen: I would add another branch above conditional on `subject instanceof Results` calling the `subject.clear()` (which needs to be added to the packages/bindgen/spec.yml first)

```suggestion
    } else if (Array.isArray(subject) || Symbol.iterator in subject) {
```

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,24 @@ export class Realm {
       );
       const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
       table.removeObject(obj.key);
+    } else if (subject instanceof List) {
+      const internalList = subject.internal;
+      internalList.removeAll();
+    } else if (subject instanceof Results) {
+      const internalResult = subject.internal;
+      internalResult.clear();
```
kraenhansen, line 519: Little need for a local variable.
```suggestion
      subject.internal.clear();
```

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,24 @@ export class Realm {
       );
       const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
       table.removeObject(obj.key);
+    } else if (subject instanceof List) {
+      const internalList = subject.internal;
+      internalList.removeAll();
```
kraenhansen, line 516: Little need for a local variable and [the existing SDK called deleteAll](https://github.com/realm/realm-js/blob/master/src/js_realm.hpp#L1408):

```suggestion
      subject.internal.deleteAll();
```

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,24 @@ export class Realm {
...
+      const internalResult = subject.internal;
+      internalResult.clear();
     } else if (Array.isArray(subject) || Symbol.iterator in subject) {
-      // TODO: Optimize this to not get the helper on every iteration
-      for (const object of subject) {
-        assert.instanceOf(object, RealmObject);
-        const { objectSchema } = this.classes.getHelpers(object);
-        const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
-        table.removeObject(object[INTERNAL].key);
+      if (subject.length > 0) {
```
kraenhansen, line 521: Why is this check needed?

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,24 @@ export class Realm {
...
+        // Get the objectSchema from the first element, and assume that all elements are of the same object type
+        const { objectSchema } = this.classes.getHelpers(subject[0]);
```
kraenhansen, line 523: This needs to happen after the assert that `object` is instance of `RealmObject`.

I actually think it would be better to leave it as it was and simply remove the TODO. It will be a lot less hot code now that you've added native support for `List` and `Result` and getting the helper on every object makes the API more useful as you can call it with an array of different objects.

packages/realm/src/Realm.ts
takameyer: This is done

packages/realm/src/Realm.ts
takameyer: Done.  Also did the same for Results

packages/realm/src/Realm.ts
takameyer: I could also assert `object` at that level as well.

takameyer:
```typescript
        assert.instanceOf(subject[0], RealmObject);
        const { objectSchema } = this.classes.getHelpers(subject[0]);
        for (const object of subject) {
          const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
          table.removeObject(object[INTERNAL].key);
        }
```

takameyer: But if for some reason, one was to just wildly put a bunch of realm objects of a different type into an array and try and delete it, this might cause some issues.  Maybe just putting it back the way it was is fine

packages/realm/src/Realm.ts
```diff
@@ -511,16 +511,21 @@ export class Realm {
...
+    } else if (subject instanceof Results) {
+      subject.internal.clear();
     } else if (Array.isArray(subject) || Symbol.iterator in subject) {
-      // TODO: Optimize this to not get the helper on every iteration
-      for (const object of subject) {
-        assert.instanceOf(object, RealmObject);
-        const { objectSchema } = this.classes.getHelpers(object);
-        const table = binding.Helpers.getTable(this.internal, objectSchema.tableKey);
-        table.removeObject(object[INTERNAL].key);
+      if (subject.length > 0) {
```
kraenhansen, line 519: No need for this check now.

```suggestion
```

kraenhansen (approved):

PR #5075, opened 2022-11-04
Adding a type for the key returned from create

kraenhansen:
## What, How & Why?

This suggested change introduce a new type for the API key and assert the key is a string as it is returned from the service after creation.


JacobOscarGunnarsson (approved):
LGTM

PR #5079, opened 2022-11-07
Bindgen: Apple, Facebook, Google, JWT & function Credentials

### LINKED ISSUE #4808: Credentials #3: apple, facebook, google, function, jwt

JacobOscarGunnarsson:
Exposes credentials (apple, Facebook, google, jwt and custom functions)through bindgen.
This closes #4808

## ☑️ ToDos
* [ ] 📝 make the imported test-apps more modular


integration-tests/tests/src/tests/credentials/jwt-encode/index.d.ts
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+declare module "jwt-encode" {
```
kraenhansen, line 19: I can't remember why this was originally needed in realm-web, but it there does seem to be a `@types/jwt-encode` package available for this: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jwt-encode/index.d.ts

integration-tests/tests/src/tests/credentials/jwt.ts
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+import { expect } from "chai";
+import { Credentials, User } from "realm";
+
+import { importAppBefore } from "../../hooks";
+import jwtEncode from "jwt-encode";
```
kraenhansen, line 22: It is a little concerning that the GitHub repo for this package is a 404 :/
From https://jwt.io/libraries?language=JavaScript I think we could go for something like https://www.npmjs.com/package/jose instead.

integration-tests/realm-apps/with-auth-providers/functions/customAuthentication/config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "5e622d21b55fffe5d410ebe6",
```
kraenhansen, line 2: I believe this could be removed as it's a template.

```suggestion
```

kraenhansen (approved):
I'm okay with merging this as is, but if you're up switching away from "jwt-encode", I think that would be a better long-term solution.

packages/realm/src/app-services/Credentials.ts
```diff
@@ -76,4 +77,49 @@ export class Credentials {
...
+    return new Credentials(binding.AppCredentials.custom(token));
+  }
+
+  /**
+   * Creates credentials with an Atlas App Services function and user identifier.
+   * @param payload A string identifying the user. Usually an identity token or a username.
+   * @return {Credentials} An instance of `Credentials` that can be used in {@linkcode Realm.App.logIn}.
+   */
+  static function(payload: string): Credentials {
+    return new Credentials(binding.AppCredentials.function(payload));
```
RedBeard0531, line 123: I changed this to take `EJson` in the spec, meaning that it will JSON-encode any argument passed to it. If you want to pass it pre-encoded JSON, you should probably change it back to `std::string` in the spec.

PR #5080, opened 2022-11-07
Implement App/User Listener

### LINKED ISSUE #5062: Add listeners to App and User

takameyer:
## What, How & Why?
Used the `subscribe` and `unsubscribe` methods to finish the app and user listener implementation.

This closes #5062

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/app-services/App.ts
+++ b/packages/realm/src/app-services/App.ts
@@ -35,10 +35,7 @@ export type AppConfiguration = {
 
 export type AppChangeCallback = () => void;
 
-type AppListenerToken = {
-  internal: binding.App;
-  token: binding.AppSubscriptionToken;
-};
+type AppListenerToken = binding.AppSubscriptionToken;
 
 // TODO: Ensure this doesn't leak
 const appByUserId = new Map<string, App>();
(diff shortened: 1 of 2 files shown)


packages/realm/src/app-services/App.ts
```diff
@@ -65,12 +70,13 @@ export class App {
 
   public userAgent = `RealmJS/${App.SDK_VERSION} (${App.PLATFORM_CONTEXT}, ${App.PLATFORM_OS}, v${App.PLATFORM_VERSION})`;
 
-  private listeners = new Listeners<AppChangeCallback, unknown>({
-    register() {
-      throw new Error("Not yet implemented");
+  private listeners = new Listeners<AppChangeCallback, AppListenerToken, [binding.App]>({
+    register(callback: () => void, internal: binding.App): AppListenerToken {
+      const token = internal.subscribe(callback);
+      return { internal, token };
```
kraenhansen, line 76: Sorry for the confusion on this when we went through it on the call. It shouldn't be needed to include the App internal into the token, since you can simple read it off the SDK `App` instance using `this.internal`.

```suggestion
  private listeners = new Listeners<AppChangeCallback, binding.AppSubscriptionToken>({
    register(callback: () => void): binding.AppSubscriptionToken {
      return this.internal.subscribe(callback);
```

takameyer: This is fixed now.  Needed to convert this into an arrow function so that the `App` context was binded to `this`.

packages/realm/src/app-services/User.ts
```diff
@@ -68,6 +74,16 @@ export class User<
   /** @internal */
   public internal: binding.SyncUser;
 
+  private listeners = new Listeners<UserChangeCallback, UserListenerToken, [binding.SyncUser]>({
```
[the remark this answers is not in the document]
takameyer: Also fixed

Issue #783, opened 2017-01-06
No way to access the deleted objects on collection notifications


mrackwitz:
The docs state this isn't possible yet, but support for this would be coming soon. I haven't found any issue tracking this though.

```
  // Update UI in response to deleted objects
  changes.deletions.forEach((index) => {
    // Deleted objects cannot be accessed directly
    // Support for accessing deleted objects coming soon...
    ...
  });
```

alazier:
Do you need this in the context of the global notifier? If so it exposes the old Realm in which case you should be able to access deleted objects from the old realm.

mrackwitz:
Initially I thought to use plain collection notifications on a synced Realm, but I'm now relying on the global listener instead as a workaround. This is good for now and not blocking me in any way.

ianpward:
Found the answer `var oldRealm = changeEvent.oldRealm.objects("RealmObject"); let deletedTask = oldRealm[index]; `

We should expose this in our docs @bigfish24

kneth:
We need to take up https://github.com/realm/realm-js/pull/5081 again

PR #5081, opened 2022-11-07
Halfbaked implementation of access to deleted object keys

jedelbo:
## What, How & Why?
This is an attempt to show how the issue explained in [realm-core#5220](https://github.com/realm/realm-core/issues/5220) could be solved. My knowledge on how things are handled in `cachedCollectio.ts` is zero, so please bare with me.

To be used as a basis for further discussion.

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/cachedCollection.ts
+++ b/packages/realm-react/src/cachedCollection.ts
@@ -133,6 +133,7 @@ export function createCachedCollection<T extends Realm.Object>({
   const listenerCallback: Realm.CollectionChangeCallback<(T & Realm.Object) | (unknown & Realm.Object)> = (
     listenerCollection,
     changes,
+    deletedObjects
   ) => {
     if (changes.deletions.length > 0 || changes.insertions.length > 0 || changes.newModifications.length > 0) {
       // TODO: There is currently no way to rebuild the cache key from the changes array for deleted object.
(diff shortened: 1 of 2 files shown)


takameyer (commented):
If I understand this correctly, there will now be an additional `deletedObjects` array sent to the change listener, which contains all the objectKeys which have been removed from the current collection.  

I think this would solve the current problem, but it's a bit, as you put it, halfbaked.  I think it would be a cleaner interface if the arguments to the listener were:
(listenerCollection, changesByIndex, changesByObjectKey)

The changesByObjectKey would then adhere to the same structure as changesByIndex (with records: insertions, newModifications, deletions).  This would increase flexibility of the listeners.

Another possibility would be a breaking change to `changes` which had a different structure:
```
{ 
changes:{
  byIndex: {
    insertions: [...], // array of indexes
    newModifications: [...],
    deletions: [...]
  },
  byObjectKey: {
    insertions: [...], //array of objectKeys
    newModifications: [...],
    deletions: [...]
  },
//deprecated
insertions: [...],
newModifications: [...],
deletions: [...]
}
```

takameyer:
We are currently using the `ObjectKey`.  I can expand on this PR to shape the user facing interface into what makes sense.  Thanks for taking the first steps!

leon-wbr:
How would I go about merging this with the most recent version of realm-js? I'm not good with C++, but I desperately need access to the deleted object keys (or the deleted objects).

It seems like implementing this would solve my problem, so I would like to work on it but don't know where to start.

I'm a bit lost as to where the CallbackWrapper is now. It seems that there needs to be changes to `realm/object-store`? I don't quite understand how the collection data (`listenerCollection`) is passed to `cachedCollection` and where I would add the additional attributes. Is it in `CollectionChangeCallback`?

jedelbo:
@leon-wbr Actually we found out that this was not needed after all. Unfortunately I can't remember the details.

leon-wbr:
That is unfortunate, because it would help solve #783, which is my problem. I'd like to implement a change listener like onDelete for every time an object is deleted, similar to what is possible in Ruby on Rails' ORM. It seems this is one of the only limited factors.

PR #5084, opened 2022-11-07
Change public api for custom-function credentials to take an object as parameter

JacobOscarGunnarsson:


packages/realm/src/app-services/Credentials.ts
```diff
@@ -116,10 +116,10 @@ export class Credentials {
 
   /**
    * Creates credentials with an Atlas App Services function and user identifier.
-   * @param payload A string identifying the user. Usually an identity token or a username.
+   * @param payload An object identifying the user. Usually an identity token or a username.
    * @return {Credentials} An instance of `Credentials` that can be used in {@linkcode Realm.App.logIn}.
    */
-  static function(payload: string): Credentials {
-    return new Credentials(binding.AppCredentials.function(payload));
+  static function(payload: object): Credentials {
```
kraenhansen, line 122: I tend to use this type for plain objects as it's nicer to work with, but for an arg type that is simply passed, the value-add might not be that great.

```suggestion
  static function(payload: Record<string, unknown>): Credentials {
```

That being said, I wonder if this could actually take any EJSON serialisable (primitives like a string or number included)?

For what it's worth [the documentation does talk about a "payload object"](https://www.mongodb.com/docs/atlas/app-services/authentication/custom-function/) .. so I guess object is fine 🤷

kraenhansen:
I'm sorry I didn't catch this in the previous review 👍

PR #5085, opened 2022-11-07
Install test refactored

kraenhansen:
## What, How & Why?

This refactors the install tests to:
- Run on both Android and iOS (previously it was just on iOS)
- Run with the new RN architecture enabled and disabled
- Run with RN@latest and RN@next (in an attempt to catch bugs early)
- Run with both jsc and hermes enabled
- Run at 7AM CET every Monday (to help us prioritise any bug on the upcoming planning meeting)


.github/workflows/install-test-react-native.yml
```diff
@@ -6,63 +6,133 @@ on:
...
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Install test *${{ needs.install.result }}*\n${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+                  }
+                }
+              ]
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TEST_WEBHOOK }}
```
kraenhansen, line 137: I will change this to "SLACK_TEAM_WEBHOOK" once it's merged.

takameyer:
Could we send the results of this test as a report in Slack?  Otherwise, we will forget to check this regularly.

.github/workflows/install-test-react-native.yml
takameyer: What channel will this show up in?

.github/workflows/install-test-react-native.yml
```diff
@@ -6,63 +6,133 @@ on:
...
+
+      - name: Install dependencies of the CLI
+        run: npm ci
+
+      - name: Initialize app
+        # Using "--skip-bundle-install" to let the setup-ruby action install the bundle
+        # Using "--skip-pod-install" to ensure it happens after setup-ruby has executed
+        run: npm run init -- --skip-bundle-install --skip-pod-install --realm-version ${{ matrix.realm-version }} --react-native-version ${{ matrix.react-native-version }} --engine ${{ matrix.engine }}
+
+      - name: CocoaPod cache
```
takameyer, line 80: We could skip this for non-ios platforms

.github/workflows/install-test-react-native.yml
```diff
@@ -6,63 +6,133 @@ on:
...
+        with:
+          # For posting a rich message using Block Kit
+          payload: |
+            {
+              "blocks": [
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Install test *${{ needs.install.result }}*\n${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```
takameyer, line 131: Is this enough information?  Would be great to see a table of the matrix and what passed or failed.

install-tests/react-native/patches/ccache-Podfile.patch
```diff
@@ -0,0 +1,18 @@
+--- app/ios/Podfile
++++ app/ios/Podfile
+@@ -39,5 +39,15 @@
+       :mac_catalyst_enabled => false
+     )
+     __apply_Xcode_12_5_M1_post_install_workaround(installer)
++    # Using ccache when building
++    installer.pods_project.targets.each do |target|
++      target.build_configurations.each do |config|
++        # Using the un-qualified names means you can swap in different implementations, for example ccache
```
takameyer, line 10: Where are we performing this swap?  Shouldn't there be a step somewhere to replace `clang` and `clang++` with a symbolic link to `ccache`?

install-tests/react-native/cli.ts
```diff
@@ -0,0 +1,263 @@
...
+      }
+
+      if (!skipBundleInstall) {
+        console.log(`Installing gem bundle (needed to pod-install for iOS)`);
+        exec("bundle", ["install"], { cwd: appPath });
+      }
+
+      if (!skipPodInstall) {
+        console.log(`Installing CocoaPods`);
+        exec("pod", ["install"], { cwd: path.resolve(appPath, "ios"), env });
```
takameyer, line 163: Maybe not an issue, but `npx pod-install` will automatically call `repo update` if this fails.

takameyer (approved):
Just a few clarifying questions and suggestions.  Nothing holding this back, except that it seems `realm-core` is being updated?

.github/workflows/install-test-react-native.yml
kraenhansen: `SLACK_TEST_WEBHOOK` goes to a private `#realm-js-bot-tests` channel and `SLACK_TEAM_WEBHOOK` goes to `#realm-js-team`.

.github/workflows/install-test-react-native.yml
kraenhansen: I think you're right .. I did try, but I can try a bit harder :)

install-tests/react-native/patches/ccache-Podfile.patch
kraenhansen: According to https://reactnative.dev/docs/build-speed#xcode-specific-setup this is enough to have XCode pick up the ccache executable if it's in the `PATH` and [that happens here](https://github.com/realm/realm-js/blob/kh/install-test-refactored/.github/workflows/install-test-react-native.yml#L62).

install-tests/react-native/cli.ts
kraenhansen: Let's add this for good measures then.

```suggestion
        // Use --no-repo-update to avoid updating the repo if the install fails
        exec("pod", ["install", "--no-repo-update"], { cwd: path.resolve(appPath, "ios"), env });
```

install-tests/react-native/cli.ts
```diff
@@ -0,0 +1,264 @@
...
+const __dirname = new URL(".", import.meta.url).pathname;
+
+const DEFAULT_APP_PATH = path.resolve(__dirname, "app");
+const APP_JS_PATH = path.resolve(__dirname, "App.js");
+const PATCHES_PATH = path.resolve(__dirname, "patches");
+const CCACHE_PODFILE_PATCH_PATH = path.resolve(PATCHES_PATH, "ccache-Podfile.patch");
+const JSC_PODFILE_PATCH_PATH = path.resolve(PATCHES_PATH, "jsc-Podfile.patch");
+const JSC_BUILD_GRADLE_PATCH_PATH = path.resolve(PATCHES_PATH, "jsc-build.gradle.patch");
+const PORT = 3000;
+const TIMEOUT = 5 * 60 * 1000; // 5 min should be pleanty of time from app has launched until message gets received
```
kneth, line 36:
```suggestion
const TIMEOUT = 5 * 60 * 1000; // 5 min should be plenty of time from app has launched until message gets received
```

kneth (approved):

PR #5092, opened 2022-11-09
Support for remote MongoDB Client and Push service Client

### LINKED ISSUE #4809: Atlas App Services #2
1. Remote MongoDB Client
1. Push service Client

The existing implementations:

* The legacy SDK (in JavaScript): https://github.com/realm/realm-js/blob/master/lib/mongo-client.js
* The Realm Web (in TypeScript): https://github.com/realm/realm-js/blob/master/packages/realm-web/src/services/MongoDBService/MongoDBCollection.ts

JacobOscarGunnarsson:
This closes #4809 

## ☑️ ToDos
* [ ] 📝 Add docs for types declared in MongoClient.ts??

### DIFF
--- a/packages/bindgen/spec.yml
+++ b/packages/bindgen/spec.yml
@@ -1036,9 +1036,16 @@ classes:
       userAPIKeyProviderClient:
         - sig: () -> UserAPIKeyProviderClient
           cppName: provider_client<app::App::UserAPIKeyProviderClient>
+      push_notification_client: '(service_name: const std::string&) -> PushClient'
       subscribe: '(observer: (app: IgnoreArgument<const App&>)) -> AppSubscriptionToken'
       unsubscribe: '(token: AppSubscriptionToken)'
       call_function: '(user: const SharedSyncUser&, name: std::string, args: EJsonArray, service_name: util::Optional<std::string>, cb: AsyncCallback<(result: Nullable<const EJson*>, err: util::Optional<AppError>)>)'
+
+  PushClient:
+    cppName: app::PushClient
+    methods:
+      register_device: '(registration_token: const std::string&, sync_user: const std::shared_ptr<SyncUser>&, completion: AsyncCallback<(err: util::Optional<AppError>)>&&)'
+      deregister_device: '(sync_user: const std::shared_ptr<SyncUser>&, completion: AsyncCallback<(err: util::Optional<AppError>)>&&)'
 
   UsernamePasswordProviderClient:
     cppName: app::App::UsernamePasswordProviderClient
--- a/packages/realm/src/app-services/MongoClient.ts
+++ b/packages/realm/src/app-services/MongoClient.ts
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import { Document, EJSON } from "bson";
-import { DefaultFunctionsFactory, User, createFactory } from "../internal";
+import { DefaultFunctionsFactory, User, binding, createFactory } from "../internal";
 
 /**
  * Options passed when finding a signle document
@@ -161,18 +161,20 @@ type Update = Record<string, unknown>;
 type AggregatePipelineStage = Record<string, unknown>;
 
 export class MongoClient<T extends Document> {
+  /** @internal */
+  private user: binding.SyncUser;
   databaseName: string;
   name: string;
-  user: User;
   serviceName: string;
   functions: DefaultFunctionsFactory;
 
   /**
    * Construct a remote collection of documents
    */
-  constructor(user: User, serviceName: string, databaseName: string, collectionName: string) {
+  /** @internal */
+  constructor(user: binding.SyncUser, serviceName: string, databaseName: string, collectionName: string) {
     this.user = user;
-    this.functions = createFactory(user, serviceName);
+    this.functions = createFactory(User.get(user), serviceName);
     this.databaseName = databaseName;
     this.name = collectionName;
     this.serviceName = serviceName;
--- a/packages/realm/src/app-services/MongoClient.ts
+++ b/packages/realm/src/app-services/MongoClient.ts
@@ -16,10 +16,52 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import { EJSON } from "bson";
-import { User, createFactory, DefaultFunctionsFactory } from "../internal";
-export type Filter {}
-export class MongoClient {
+import { Document, EJSON } from "bson";
+import { DefaultFunctionsFactory, User, createFactory } from "../internal";
+
+type FindOneOptions = {
+  readonly projection?: Record<string, unknown>;
+  readonly sort?: Record<string, unknown>;
+};
+
+type FindOneAndModifyOptions = FindOneOptions & {
+  readonly upsert?: boolean;
+  readonly returnNewDocument?: boolean;
+};
+
+type FindOptions = FindOneOptions & {
+  readonly limit?: number;
+};
+
+type CountOptions = {
+  readonly limit?: number;
+};
+
+type AggregatePipelineStage = Record<string, unknown>;
+type Filter = Record<string, unknown>;
+type Update = Record<string, unknown>;
+type NewDocument<T extends Document> = Omit<T, "_id"> & Partial<Pick<T, "_id">>;
+type InsertOneResult<IdType> = {
+  readonly insertedId: IdType;
+};
+type InsertManyResult<IdType> = {
+  readonly insertedIds: IdType[];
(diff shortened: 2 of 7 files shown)


packages/realm/src/app-services/MongoClient.ts
```diff
@@ -0,0 +1,220 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { Document, EJSON } from "bson";
+import { DefaultFunctionsFactory, User, createFactory } from "../internal";
+
```
JacobOscarGunnarsson, line 21: Should docs be included here?

packages/bindgen/spec.yml
```diff
@@ -1036,10 +1036,17 @@ classes:
...
+      push_notification_client: '(service_name: const std::string&) -> PushClient'
       subscribe: '(observer: (app: IgnoreArgument<const App&>)) -> AppSubscriptionToken'
       unsubscribe: '(token: AppSubscriptionToken)'
       call_function: '(user: const SharedSyncUser&, name: std::string, args: EJsonArray, service_name: util::Optional<std::string>, cb: AsyncCallback<(result: Nullable<const EJson*>, err: util::Optional<AppError>)>)'
 
+  PushClient:
+    cppName: app::PushClient
+    methods:
+      register_device: '(registration_token: const std::string&, sync_user: const std::shared_ptr<SyncUser>&, completion: AsyncCallback<(err: util::Optional<AppError>)>&&)'
+      deregister_device: '(sync_user: const std::shared_ptr<SyncUser>&, completion: AsyncCallback<(err: util::Optional<AppError>)>&&)'
```
nirinchev, line 1048: Mentioned this on Slack, but for visibility here as well - the Push Notifications functionality has been deprecated on the server, so it probably doesn't make a whole lot of sense to support it with the bindgen version of the SDK.

RedBeard0531: We are aiming for v12 (SDK-on-bindgen) being as close to a drop-in replacement for v11 (SDK-in-C++) as possible. If we want to remove this for v12, we should deprecate it in v11 ASAP.

nirinchev: I think we should - not sure how many people realize this functionality is deprecated on the server, but at some point next year, if they're using it, it'll stop working.

takameyer (approved):
Looks good 👍🏼

packages/realm/src/app-services/MongoClient.ts
```diff
@@ -0,0 +1,400 @@
...
+export class MongoClient<T extends Document> {
+  databaseName: string;
+  name: string;
+  user: User;
+  serviceName: string;
+  functions: DefaultFunctionsFactory;
+
+  /**
+   * Construct a remote collection of documents
```
RedBeard0531, line 171: Most method comments end in a `.` but some don't. Try to be consistent.

packages/realm/src/app-services/PushClient.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Authentication provider where users identify using an API-key.
+ */
+export class PushClient {
+  /** @internal */
+  private user: binding.SyncUser;
+  /** @internal */
+  public internal: binding.PushClient;
+
+  /** @internal */
```
RedBeard0531, line 29: I'm not sure we should mark the constructor as internal. That leads to no constructor being included in the .d.ts, which is equivalent to having an implicit `constructor()`: [playground](https://www.typescriptlang.org/play?stripInternal=true#code/MYGwhgzhAEBiD29oG8BQBfVqB2BTA7nIgBQCUA3FqJDAEJgBOa0L0A9AFQfQACAltgAuuBtjAhoHNqlbRg8bBEEMArsEHwGxMAC5o2FQFsARiNIpMmHAWj0tFdm2iCAFnxj4FAckHQRDTWg+ADNoAE94FWgAKxUlaBBEAGtoMF9XXGgAOgATLMEIIA)

I'll file a bug with TS because there _should_ be a way to have a type that consumers can't construct.

packages/realm/src/app-services/User.ts
```diff
@@ -58,7 +61,25 @@ export interface UserIdentity {
   providerType: ProviderType;
 }
 
+/** @internal */
+function cleanArguments(args: unknown[] | unknown): unknown[] | unknown {
+  if (Array.isArray(args)) {
+    return args.map((x) => cleanArguments(x));
+  } else if (typeof args === "object") {
```
RedBeard0531, line 68: No need for `else` following a `return`

packages/realm/src/app-services/User.ts
```diff
@@ -58,7 +61,25 @@ export interface UserIdentity {
...
+    const result: { [key: string]: unknown } = {};
+    for (const [k, v] of Object.entries(args as object)) {
+      if (typeof v !== "undefined") {
+        result[k] = v;
```
RedBeard0531, line 72: Shouldn't this be `result[k] = cleanArguments(v)`? Or is there a reason we don't want to recursively strip undefineds?

packages/realm/src/app-services/User.ts
```diff
@@ -58,7 +61,25 @@ export interface UserIdentity {
   providerType: ProviderType;
 }
 
+/** @internal */
+function cleanArguments(args: unknown[] | unknown): unknown[] | unknown {
+  if (Array.isArray(args)) {
+    return args.map((x) => cleanArguments(x));
+  } else if (typeof args === "object") {
+    const result: { [key: string]: unknown } = {};
+    for (const [k, v] of Object.entries(args as object)) {
```
RedBeard0531, line 70: You shouldn't need the `as object` cast. My best guess is that this is because `typeof null == "object"`, so we are entering this block incorrectly for null args. You probably want to make the `if` something like `if (args !== null and typeof args === "object")`.

Or even better, use an `if (args  === null || typeof args != "object") return args` early return.

packages/realm/src/app-services/User.ts
```diff
@@ -177,7 +198,7 @@ export class User<
    * Use this to call functions defined by the Atlas App Services application, as this user.
    */
   get functions(): FunctionsFactoryType {
-    throw new Error("Not yet implemented");
+    return createFactory(this as User, undefined);
```
RedBeard0531, line 201: Why do you need the `this as User` cast? I thought that would be implied on any method of User. Or is this to strip off the template params?

packages/realm/src/app-services/User.ts
```diff
@@ -256,7 +284,21 @@ export class User<
...
+      get serviceName() {
+        return serviceName;
+      },
+      db: (dbName: string) => {
+        return {
+          get name() {
+            return dbName;
+          },
+          collection: (collectionName: string) => {
+            return new MongoClient(this as User, serviceName, dbName, collectionName);
```
RedBeard0531, line 297: Ditto question about cast. If it is warning without the cast, I'd be a bit concerned that `this` is bound to something else here. Can you check what type TS thinks `this` is without the cast?

packages/realm/src/app-services/User.ts
```diff
@@ -242,8 +269,9 @@ export class User<
    *
    * @returns An service client with methods to register and deregister the device on the user.
```
RedBeard0531, line 270: Lets go ahead and add an `@deprecated` here with a link to https://www.mongodb.com/docs/atlas/app-services/reference/push-notifications/

RedBeard0531 (changes requested):

packages/realm/src/app-services/PushClient.ts
RedBeard0531: Bug filed: https://github.com/microsoft/TypeScript/issues/51486

packages/realm/src/app-services/User.ts
JacobOscarGunnarsson: This method was extracted from `utils.js` which looks like the following:
image
We actually don't call it there either, but your reasoning seems right and I think we should unless there's no reason to why we don't do it..

packages/realm/src/app-services/User.ts
JacobOscarGunnarsson: image
this is the error generated from removing the case

JacobOscarGunnarsson: A similar error is caused by the `return createFactory(this as User, undefined)`

PR #5093, opened 2022-11-10
Implement Keys for Spread Operator

### LINKED ISSUE #2844: Spread operation not works

takameyer:
## What, How & Why?
Removed the public 'keys' and 'entries' methods from Object. Replaced with symbol based method.

This closes #2844 and #3805

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm/src/Object.ts
```diff
@@ -252,13 +253,12 @@ export class RealmObject<T = DefaultObject> {
    */
   private readonly [INTERNAL_LISTENERS]!: ObjectListeners<T>;
 
-  // TODO: Find a way to bind this in
-  keys(): string[] {
-    throw new Error("This is expected to have a per-class implementation");
+  private keys(): string[] {
+    return Object.keys(this);
```
kraenhansen, line 257: This would result in an endless recursion if executed, but it's never called as [the `ClassHelper` provides a per-class implementation of the function](https://github.com/realm/realm-js/blob/bindgen/packages/realm/src/ClassMap.ts#L93-L101).

PR #5095, opened 2022-11-10
Implement getAllSyncSessions

takameyer:
## What, How & Why?
Add allSessions to SyncUser in bindgen. Transform result to JS SyncSession array.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/bindgen/spec.yml
```diff
@@ -993,6 +993,7 @@ classes:
       state: SyncUserState
       subscribers_count: count_t
     methods:
+      all_sessions: () -> std::vector<Nullable<SharedSyncSession>>
```
RedBeard0531, line 996: I don't think this needs to be `Nullable<>` since we should only get non-null pointers in the vector.

Also, this feels like it could be a property rather than a method. There isn't a hard-and-fast rule, but generally pure getters are turned into properties rather than methods.

packages/realm/src/app-services/Sync.ts
```diff
@@ -76,8 +76,9 @@ export namespace Sync {
     });
     app.internal.syncManager.setLoggerFactory(factory);
   }
-  export function getAllSyncSessions(user: User): SyncSession[] {
-    throw new Error("Not yet implemented");
+  export function getAllSyncSessions(user: User): SyncSession[] | null {
+    const allSessions = user.internal.allSessions();
+    return allSessions?.map((session) => new SyncSession(session)) ?? null;
```
RedBeard0531, line 81: Even with the `Nullable<>` in place, you didn't need the `?.`. That meant that the SyncSession pointers *inside* the vector/Array may be null, not that the vector itself may be null.

Did the generated TS types cause an error/warning without the `?.`? If so, I think that would be a bug in the generator.

RedBeard0531 (changes requested):

packages/bindgen/spec.yml
takameyer: Ok, i'll follow that guideline and make it into a property and remove `Nullable<>`.  Tests are passing and the SDK code is much simpler 👍🏼

packages/realm/src/app-services/Sync.ts
RedBeard0531: Crap, looks like there was a bug! That should have been `(null | SyncSession)[]` but was rendered as `null | SyncSession[]` which is parsed as `null | (SyncSession[])`. Will fix.

RedBeard0531: Fixed by https://github.com/realm/realm-js/commit/707f80a8b046eb35596672b65f73d45477c510a2

packages/bindgen/spec.yml
```diff
@@ -366,9 +366,9 @@ records:
   UserIdentity:
     cppName: SyncUserIdentity
     fields:
-      id: 
+      id:
         type: std::string
-      provider_type: 
+      provider_type:
         type: std::string
```
RedBeard0531, line 372: FYI: when you don't need a `default` you can just use the shorthand `name: Type` like `id: std::string`.

I know this was probably just an automatic whitespace cleanup, so don't feel like you need to fix this in your commit, but I wanted to mention it in case you end up declaring a record in the future.

CC @JacobOscarGunnarsson

RedBeard0531 (approved):
Much simpler without the unneccesary `Nullable<>`!

kraenhansen:
@takameyer what's the resolution on this? Did it land in another commit or does it need to be completely reworked to address the feedback from the review?

takameyer:
It landed in another commit #5302

PR #5098, opened 2022-11-10
Add remaining user-methods for bindgen

JacobOscarGunnarsson:
Standalone PR which complements #4808 and #4805 now that support for Bson is provided through the bindgen.

implements the following methods:

- `User.customData()`
- `User.refreshCustomData()`
- `User.profile()`
- `EmailPasswordAuthClient.callResetPassWordFunction()`

### DIFF
--- a/packages/realm/src/app-services/User.ts
+++ b/packages/realm/src/app-services/User.ts
@@ -184,7 +184,11 @@ export class User<
    * If this value has not been configured, it will be empty.
    */
   get customData(): CustomDataType {
-    throw new Error("Not yet implemented");
+    const result = this.internal.customData;
+    if (result === undefined) {
+      return {} as CustomDataType;
+    }
+    return result as CustomDataType;
   }
 
   /**
--- a/packages/realm/src/app-services/User.ts
+++ b/packages/realm/src/app-services/User.ts
@@ -195,7 +195,7 @@ export class User<
    * A profile containing additional information about the user.
    */
   get profile(): UserProfileDataType {
-    throw new Error("Not yet implemented");
+    return this.internal.userProfile as UserProfileDataType;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/realm/src/app-services/User.ts
```diff
@@ -70,6 +70,9 @@ export class User<
   /** @internal */
   public internal: binding.SyncUser;
 
+  // cached version of profile
+  private _profile: UserProfileDataType | undefined;
```
JacobOscarGunnarsson, line 74: we talked about not using underscore for private fields but I'm not sure how we actually want it so putting it out there like this for now

packages/realm/src/app-services/User.ts
```diff
@@ -163,14 +167,21 @@ export class User<
    * If this value has not been configured, it will be empty.
    */
   get customData(): CustomDataType {
-    throw new Error("Not yet implemented");
+    const result = this.internal.customData;
```
JacobOscarGunnarsson, line 170: maybe it would be worth having a cached type for this as well?

kraenhansen: It gets updated when the access token gets refreshed, so I think it's better to leave it un-memoized for now.
Also, it's not too expensive compared to how little it'll get called and end-users can always work around by caching it on the calling side if they experience performance issues.

packages/realm/src/app-services/User.ts
takameyer: I would name it `cachedProfile` and remove the `_`.

packages/realm/src/app-services/User.ts
```diff
@@ -163,14 +167,21 @@ export class User<
...
+    }
+    return result as CustomDataType;
   }
 
   /**
    * A profile containing additional information about the user.
    */
   get profile(): UserProfileDataType {
-    throw new Error("Not yet implemented");
+    if (!this._profile) {
```
takameyer, line 181: Does this cached `_profile` get invalidated somehow?

takameyer (approved):
Just a few comments, but nothing holding this back.

PR #5100, opened 2022-11-11
Fix SessionTests.testNonAcceptedPartitionValueTypes

### LINKED ISSUE #5064: Validate partition values in config validation
The SessionTests.testNonAcceptedPartitionValueTypes are failing, indicating that invalid partition values can be sent to the server.

JacobOscarGunnarsson:
Added validation for the `PartitionValue` provided when opening a `Realm`
This closes #5064

### DIFF
--- a/packages/realm/src/app-services/SyncConfiguration.ts
+++ b/packages/realm/src/app-services/SyncConfiguration.ts
@@ -103,6 +103,7 @@ export function toBindingSyncConfig(config: SyncConfiguration): binding.SyncConf
   }
   const { user, onError, _sessionStopPolicy } = config;
   assert.instanceOf(user, User, "user");
+  validatePartitionValue(config.partitionValue);
   const partitionValue = EJSON.stringify(config.partitionValue as EJSON.SerializableTypes);
   return {
     user: config.user.internal,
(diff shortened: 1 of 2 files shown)


packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -113,3 +114,26 @@ export function toBindingSyncConfig(config: SyncConfiguration): binding.SyncConf
       : binding.SyncSessionStopPolicy.AfterChangesUploaded,
   };
 }
+
+/** @internal */
+function validatePartitionValue(partitionValue: unknown) {
+  if (partitionValue === undefined) {
+    throw new Error(partitionValue + " is not an allowed PartitionValue");
+  }
+  if (typeof partitionValue == "number") {
```
takameyer, line 123:
```suggestion
  if (typeof partitionValue === "number") {
```
In general, I think `===` should be used instead of `==`, unless you have a reason to do so. `==` has so many edge cases, that it's possible to get some false positives if you don't know what it's doing.
In this case, it doesn't make much of a difference, as `typeof` will always return a string and `"number"` is also always going to be a string.  
For more on the matter: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -113,3 +114,26 @@ export function toBindingSyncConfig(config: SyncConfiguration): binding.SyncConf
       : binding.SyncSessionStopPolicy.AfterChangesUploaded,
   };
 }
+
+/** @internal */
+function validatePartitionValue(partitionValue: unknown) {
+  if (partitionValue === undefined) {
```
takameyer, line 120: We are missing a few other invalid cases here.  I would try and mirror what was written before in c++ (https://github.com/realm/realm-js/blob/master/src/js_sync.hpp#L77-L104)
For instance, currently the types `boolean`, `Date`, `Object`, `Array`, etc would all pass this validation.  In this case, I would verify that the value is typeof `string`, `number`, `objectId` or `null`.  If it isn't one of those, then throw.

takameyer (changes requested):

packages/realm/src/app-services/SyncConfiguration.ts
JacobOscarGunnarsson: hehe I think it was mostly my muscle memory that went for the `==`, I've also heard rumours about `==` being iffy in JS so I'll change it!

tests/js/session-tests.js
```diff
@@ -1046,6 +1046,9 @@ module.exports = {
 
   async testNonAcceptedPartitionValueTypes() {
     const testPartitionValues = [
+      true,
```
JacobOscarGunnarsson, line 1049: this addition may not be necessary but I see no harm in adding them as well

takameyer (approved):
Looks good!

PR #5102, opened 2022-11-11
consume the provided ThreadSafeReference in Realm.open async callback

RedBeard0531:
I confirmed that this solves the hanging issue that we were seeing in `testAddConnectionNotification` when combined with https://github.com/realm/realm-js/commit/abf636b52a833671456c336a31d3190e8aefdf1d. I'm making a PR since we had previously discussed that you didn't want to consume the TSR here. Since this matches the [existing SDK behavior](https://github.com/realm/realm-js/blob/8a63052553554e1634fbb7fa04f26f32709b472f/src/js_realm.hpp#L1228), and it ended up not being too gross, I think it is worth reconsidering. Let me know if you are happy with this direction.

There is also an issue where a forgotten `realm.close()` in one of our [legacy tests](https://github.com/realm/realm-js/blob/f2d7e1862f087c42a55a263348a88a1c3ef6da3e/tests/js/set-sync-tests.js#L63-L70) now crashes with the following error. Adding the `realm.close()` fixes this. I need to look into whether this is a real problem that could affect users or just an artifact of our testing with `clearTestState()` and `_sessionStopPolicy: "immediately"`.

```
terminate called after throwing an instance of 'realm::MultipleSyncAgents'
  what():  Multiple sync agents attempted to join the same session
```

### DIFF
--- a/packages/bindgen/spec.yml
+++ b/packages/bindgen/spec.yml
@@ -559,6 +559,7 @@ classes:
       base64_decode: '(input: StringData) -> BinaryData'
       make_logger_factory: '(log: (level: LoggerLevel, message: const std::string&) off_thread) -> LoggerFactory'
       simulate_sync_error: '(session: SyncSession&, code: int, message: StringData, type: StringData, is_fatal: bool)'
+      consume_thread_safe_reference_to_shared_realm: '(tsr: ThreadSafeReference) -> SharedRealm'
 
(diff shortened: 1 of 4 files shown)


kraenhansen:
@RedBeard0531 I suppose this could be closed? It looks to me as if this got merged in some other way.

PR #5112, opened 2022-11-14
remove crypto dependency

JacobOscarGunnarsson:
`jwt.ts` relied on the `crypto` module which isn't available on react-native. Solution was to simply remove it since using it would be overkill in this test-case


takameyer (approved):
Seems like a good compromise.

PR #5115, opened 2022-11-15
Jg/sync config

### LINKED ISSUE #5063: Make SyncSession#config return all properties
The existing SDK reads the config from core and constructs a new object from that. It contains the "onError" function and "customHttpHeaders" too, which is currently not exposed when reading the config from the binding. We could consider storing the sync config passed to the Realm constructor instead of deriving it from core, but that forces the SDK SyncSession constructor to take additional arguments that might not be available when it needs to be constructed.

JacobOscarGunnarsson:
This closes #5063 

Allows a user to get `customHttpHeaders` from a `syncSession.config`.
it was decided that the onError field should not be exposed.

### DIFF
--- a/tests/js/session-tests.js
+++ b/tests/js/session-tests.js
@@ -154,7 +154,6 @@ module.exports = {
     config.sync.customHttpHeaders = { language: "English" };
     const realm = new Realm(config);
     const session = realm.syncSession;
-    console.log(session.config);
     TestCase.assertEqual(
       "English",
       session.config.customHttpHeaders.language,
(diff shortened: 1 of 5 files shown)


tests/js/session-tests.js
```diff
@@ -367,8 +384,6 @@ module.exports = {
         };
         const realm = new Realm(config);
         const session = realm.syncSession;
-
-        TestCase.assertEqual(session.config.onError, config.sync.onError);
```
JacobOscarGunnarsson, line 371: This assert is removed in master as well, not sure if I should do it here also

takameyer: It should be removed

takameyer (approved):
LGTM

PR #5116, opened 2022-11-15
overload emailPasswordAuth

JacobOscarGunnarsson:
overloaded emailPassWordAuth to support positional arguments

### DIFF
--- a/packages/realm/src/app-services/Credentials.ts
+++ b/packages/realm/src/app-services/Credentials.ts
@@ -60,13 +60,19 @@ export class Credentials {
 
   /**
    * Creates credentials based on a login with an email address and a password.
-   * @param credentials An object with username and password for the user.
    * @return {Credentials} An instance of `Credentials` that can be used in {@linkcode App.logIn}.
    */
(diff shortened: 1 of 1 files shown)


packages/realm/src/app-services/Credentials.ts
```diff
@@ -60,13 +60,21 @@ export class Credentials {
...
-  static emailPassword(credentials: { email: string; password: string }): Credentials {
-    assert.string(credentials.email, "email");
-    assert.string(credentials.password, "password");
-    return new Credentials(binding.AppCredentials.usernamePassword(credentials.email, credentials.password));
+  static emailPassword(credentials: { email: string; password: string }): Credentials;
+  static emailPassword(email: string, password: string): Credentials;
+  static emailPassword(arg1: { email: string; password: string } | string, password?: string): Credentials {
+    let email;
+    if (typeof arg1 === "string") {
+      email = arg1;
```
RedBeard0531, line 70: Either do `const email = arg1` in this scope or just use `arg1` in place of  `email` in the call on line 72. Either way, delete line 68.

RedBeard0531 (approved):

PR #5119, opened 2022-11-17
Jg/client reset methods

JacobOscarGunnarsson:
early pull request for client reset support, still wip

### DIFF
--- a/integration-tests/tests/src/tests/sync/client-reset.ts
+++ b/integration-tests/tests/src/tests/sync/client-reset.ts
@@ -44,15 +44,15 @@ function getPartitionValue() {
   return new UUID().toHexString();
 }
 
-async function triggerClientReset(app: Realm.App, user: Realm.User): Promise<void> {
+async function triggerClientReset(app: App, user: User): Promise<void> {
   await user.functions.triggerClientReset(app.id, user.id);
 }
 
 async function waitServerSideClientResetDiscardUnsyncedChangesCallbacks(
   useFlexibleSync: boolean,
   schema: Realm.ObjectSchema[],
-  app: Realm.App,
-  user: Realm.User,
+  app: App,
+  user: User,
   actionBefore: (realm: Realm) => void,
   actionAfter: (beforeRealm: Realm, afterRealm: Realm) => void,
 ): Promise<void> {
@@ -100,8 +100,8 @@ async function waitServerSideClientResetDiscardUnsyncedChangesCallbacks(
 async function waitServerSideClientResetRecoveryCallbacks(
   useFlexibleSync: boolean,
   schema: Realm.ObjectSchema[],
-  app: Realm.App,
-  user: Realm.User,
+  app: App,
+  user: User,
   actionBefore: (realm: Realm) => void,
   actionAfter: (beforeRealm: Realm, afterRealm: Realm) => void,
 ): Promise<void> {
@@ -149,7 +149,7 @@ async function waitServerSideClientResetRecoveryCallbacks(
 async function waitSimulatedClientResetDiscardUnsyncedChangesCallbacks(
   useFlexibleSync: boolean,
   schema: Realm.ObjectSchema[],
-  user: Realm.User,
+  user: User,
   actionBefore: (realm: Realm) => void,
   actionAfter: (beforeRealm: Realm, afterRealm: Realm) => void,
 ): Promise<void> {
@@ -157,7 +157,7 @@ async function waitSimulatedClientResetDiscardUnsyncedChangesCallbacks(
     let afterCalled = false;
     let beforeCalled = false;
 
-    const modifiedConfig: Realm.Configuration = {
+    const modifiedConfig: Configuration = {
       schema,
       sync: {
         user,
@@ -201,7 +201,7 @@ async function waitSimulatedClientResetDiscardUnsyncedChangesCallbacks(
 async function waitSimulatedClientResetRecoverCallbacks(
   useFlexibleSync: boolean,
   schema: Realm.ObjectSchema[],
-  user: Realm.User,
+  user: User,
   actionBefore: (realm: Realm) => void,
   actionAfter: (beforeRealm: Realm, afterRealm: Realm) => void,
 ): Promise<void> {
@@ -209,7 +209,7 @@ async function waitSimulatedClientResetRecoverCallbacks(
     let afterCalled = false;
     let beforeCalled = false;
 
-    const modifiedConfig: Realm.Configuration = {
+    const modifiedConfig: Configuration = {
       schema,
       sync: {
         user,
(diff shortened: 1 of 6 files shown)


integration-tests/tests/src/tests/sync/client-reset.ts
```diff
@@ -44,26 +44,26 @@ function getPartitionValue() {
...
   const resetHandle = createPromiseHandle();
   let afterCalled = false;
   let beforeCalled = false;
-
-  const realm = new Realm({
+  const config: Configuration = {
     schema,
     sync: {
       user,
+      // @ts-expect-error this setting is not for users to consume
```
JacobOscarGunnarsson, line 66: is there any other recommended way to treat this ts-error?

packages/realm/src/Configuration.ts
```diff
@@ -82,10 +85,10 @@ type BaseConfiguration = {
 //   Manual = "manual",
 // }
```
JacobOscarGunnarsson, line 87: having this piece of code caused name collision, might need to double-check that it's not used elsewhere..

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -100,6 +106,84 @@ export function toBindingErrorHandler(onError: ErrorCallback) {
...
+      session.resetInternal();
+    };
+  }
+  // onManual must be true
+  return (sessionInternal: binding.SyncSession, bindingError: binding.SyncError) => {
+    const session = new SyncSession(sessionInternal);
+    const error = fromBindingSyncError(bindingError);
+    if (error instanceof ClientResetError) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      onManual!(session, error.config.path!);
```
JacobOscarGunnarsson, line 146: but confused here why TS can't prune onManual to be true after previous if statements

JacobOscarGunnarsson: Also it might be risky with `error.config.path!` is there a better way to fix this?

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -100,6 +106,84 @@ export function toBindingErrorHandler(onError: ErrorCallback) {
...
+  return (localRealmInternal: binding.Realm, tsr: binding.ThreadSafeReference, didRecover: boolean) => {
+    if (didRecover) {
+      onAfter(new Realm(localRealmInternal), new Realm(binding.Helpers.consumeThreadSafeReferenceToSharedRealm(tsr)));
+    } else {
+      const realm = new Realm(binding.Helpers.consumeThreadSafeReferenceToSharedRealm(tsr));
+      if (onFallback) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        onFallback(realm.syncSession!, realm.path);
+      } else {
+        // should we throw here?
```
JacobOscarGunnarsson, line 180: Not sure if we should throw here

integration-tests/tests/src/tests/sync/client-reset.ts
kneth: I don't think so. `_sessionStopPolicy` is only for our internal testing, and we don't wish our users to use it.

packages/realm/src/Configuration.ts
kneth: The bindgen project is "promoting" TypeScript to be first-class citizen, and it might have some implications of how the TypeScript API is described. Thank you for raising it - and we should double-check if we can remove this code section.

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -50,20 +55,62 @@ export type OpenRealmBehaviorConfiguration = {
...
+  | ClientResetRecoverUnsyncedChangesConfiguration
+  | ClientResetRecoverOrDiscardUnsyncedChangesConfiguration;
+
 export type BaseSyncConfiguration = {
   user: User;
   newRealmFileBehavior?: OpenRealmBehaviorConfiguration;
   existingRealmFileBehavior?: OpenRealmBehaviorConfiguration;
   onError?: ErrorCallback;
   /** @internal */
   _sessionStopPolicy?: SessionStopPolicy; // TODO: Why is this _ prefixed?
```
kneth, line 112: We might remove `_` now as it is marked `@internal`.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -100,6 +106,83 @@ export function toBindingErrorHandler(onError: ErrorCallback) {
   };
 }
 
+/** @internal */
+export function toBindingErrorHandlerWithOnManual(
+  onError: ErrorCallback | undefined,
+  onManual: ClientResetFallbackCallback | undefined,
+) {
+  if (!onError && !onManual) {
+    throw new Error("need either onError or onManual");
```
kneth, line 115: "or both"

packages/realm/src/app-services/SyncSession.ts
kneth: I can't imagine a situation where `path` is either undefined or null.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -100,6 +106,83 @@ export function toBindingErrorHandler(onError: ErrorCallback) {
...
+    return (sessionInternal: binding.SyncSession, bindingError: binding.SyncError) => {
+      const session = new SyncSession(sessionInternal);
+      const error = fromBindingSyncError(bindingError);
+      if (error instanceof ClientResetError) {
+        onError(session, error);
+      }
+      session.resetInternal();
+    };
+  }
+  // onManual must be true here
```
kneth, line 140: You can see this but I doubt the TypeScript compiler can. And I think it answers your question why `onManual` isn't pruned.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -113,6 +196,21 @@ export function toBindingStopPolicy(policy: SessionStopPolicy): binding.SyncSess
   }
 }
 
+/** @internal */
+export function toBindingClientResetMode(resetMode: ClientResetMode): binding.ClientResetMode {
+  if (resetMode === ClientResetMode.Manual) {
```
kneth, line 201: Could also be a `switch`

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -113,6 +196,21 @@ export function toBindingStopPolicy(policy: SessionStopPolicy): binding.SyncSess
...
+    return binding.ClientResetMode.Manual;
+  } else if (resetMode === ClientResetMode.DiscardUnsyncedChanges) {
+    return binding.ClientResetMode.DiscardLocal;
+  } else if (resetMode === ClientResetMode.RecoverUnsyncedChanges) {
+    return binding.ClientResetMode.Recover;
+  } else if (resetMode === ClientResetMode.RecoverOrDiscardUnsyncedChanges) {
+    return binding.ClientResetMode.RecoverOrDiscard;
+  } else {
```
kneth, line 209: With all the `return`s above, you really don't need `else`.

kneth (approved):
It looks good to me. I am happy to see that you are returning functions many places :smile:

Most of my comments are nitpick, and they should spoil the happiness to see client reset implemented in pure TypeScript.

PR #5133, opened 2022-11-23
fix armhf compilation

iamer:


kneth (approved):
Neat trick to use core's toolchain file

fealebenpae (approved):

PR #5134, opened 2022-11-23
Add aarch64 support to the debian container

iamer:


fealebenpae (approved):
Looks good, but then we shouldn't be calling it just `armhf.Dockerfile` anymore. How about `debian-multi-arm.Dockerfile`?

iamer:
I agree, was going to name it simply arm.Dockerfile to match the name of the image in the github registry. However, I found one reference to it in the Jenkinsfile and wasn't sure if there were more elsewhere? I can change the Jenkinsfile if we are sure there is no one else depending on it.

kneth (approved):

PR #5155, opened 2022-12-02
Avoid calling the dynamic default function twice.

### LINKED ISSUE #5140: @realm/babel-plugin - new Realm.BSON.ObjectId() is invoked twice for property initialization

All the time

### Description

Give the following class:
```typescript
export class Task extends Realm.Object<Task> {
  _id: Realm.BSON.ObjectId = new Realm.BSON.ObjectId();
  description!: string;
  isComplete: boolean = false;
  createdAt: Date = new Date();
  userId!: string;

  static primaryKey = '_id';
}
```

When two objects are created using:
```typescript
realm.write(() => {
  return new Task(realm, description, userId);
});
```
The resulting `_id` is two increments higher between the first and the second item.  Oddly enough, when using sync and Atlas, the `_id` is on increment smaller for the created items.  This causes various issues when trying to make modifications when permissions are set.

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

This bug exists in the current `example` application in `realm`.  

* Setup App Service and get the AppId
   * Enable username/password login
   * Enable Flexible Sync
       * Add `userId` to the queryable fields
       * Create a permissions object:
 ```json
{
  "rules": {},
  "defaultRoles": [
    {
      "name": "owner-read-write",
      "applyWhen": {},
      "read": {
        "userId": "%%user.id"
      },
      "write": {
        "userId": "%%user.id"
      }
    }
  ]
}
``` 
* Configure the `sync.config.js` in the example app with the AppId
* Run the example application
    * Create a new user
    * Create a few items
    * Check the "done" checkbox

The Items will disappear  

### Version

11 and above

Atlas App Services (auth, functions, etc.)

### Are you using encryption?

No, not using encryption

Platform Independent

### Build environment

None

### Cocoapods version

Irrelevant

kneth:
## What, How & Why?

This closes #5140 

## ☑️ ToDos

* [x] 📝 Changelog entry

### DIFF
--- a/packages/babel-plugin/CHANGELOG.md
+++ b/packages/babel-plugin/CHANGELOG.md
@@ -1,3 +1,17 @@
+## vNext
+
+### Breaking changes
+* None
+
+### Enhancements
+* None
+
+### Fixed
+* When a property has dynamical default value, the function was called twice. ([#5140](https://github.com/realm/realm-js/issues/5140), since v0.1.0)
+
+### Internal
+* None
+
 ## 0.1.0 (2022-10-18)
 
 ### Breaking changes
--- a/packages/babel-plugin/src/plugin/index.ts
+++ b/packages/babel-plugin/src/plugin/index.ts
@@ -325,6 +325,7 @@ function visitRealmClassProperty(path: NodePath<types.ClassProperty>) {
         properties.push(
           types.objectProperty(types.identifier("default"), types.arrowFunctionExpression([], valuePath.node)),
         );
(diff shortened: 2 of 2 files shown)


takameyer (approved):
I've tested with the example application, and the problem has gone away.  Would be great to be able to write tests for this somehow.

kneth:
> Would be great to be able to write tests for this somehow.

I can only agree. In the future we should have some integration tests.

PR #5180, opened 2022-12-15
Support Asymmetric Sync

elle-j:
## What, How & Why?

This feature is instead handled in PR #5301.

Updated our integration for asymmetric sync. If `asymmetric: true` in the object schema, then:

* Creating an object will return `undefined` rather than the object itself.
* Querying asymmetric object(s) will throw an error.

## ☑️ ToDos
* [x] 🚦 Tests


packages/realm/src/Realm.ts
```diff
@@ -754,9 +762,10 @@ export class Realm {
 
   /**
    * Returns all objects of the given {@link type} in the Realm.
-   * @param type The type of Realm objects to retrieve.
+   * @param type The type of Realm object to search for.
+   * @param objectKey The object key value of the object to search for.
    * @throws {@link Error} If type passed into this method is invalid or if the type is marked embedded or asymmetric.
-   * @returns Realm.Results that will live-update as objects are created and destroyed.
+   * @returns A {@link Realm.Object} or `undefined` if the object key is invalid.
```
kneth, line 768: Maybe "not found" instead of "invalid"

kneth (approved):

packages/realm/src/Realm.ts
elle-j: I put "invalid" here (rather than more specific) because it throws if the value cannot be convert to an `ObjKey`, or more specifically a `BigInt`.

`undefined` is returned if `err instanceof binding.InvalidObjKey`. And that error is thrown in the following case:

```ts
export function stringToObjKey(input: string): ObjKey {
  try {
    return BigInt(input) as unknown as ObjKey;
  } catch {
    throw new InvalidObjKey(input);
  }
}
```

But should I still change it? It's probably more user friendly with "not found".

PR #5187, opened 2022-12-20
Fix blank screen on user provider if fallback not passed

thecoorum:
## What, How & Why?
While implementing user authentication with `UserProvider` using the last one with no `fallback` provided gave me blank screen with no user. I digged into the code and noticed that `UserProvider` always expects `fallback` and doesn't return children if it's absent

### DIFF
--- a/packages/realm-react/src/UserProvider.tsx
+++ b/packages/realm-react/src/UserProvider.tsx
@@ -62,10 +62,13 @@ export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback, 
   }, [user, app]);
 
   if (!user) {
-    if (typeof Fallback === "function") {
-      return <Fallback />;
+    if (Fallback) {
+      if (typeof Fallback === "function") {
+        return <Fallback />;
+      }
+      return <>{Fallback}</>;
     }
-    return <>{Fallback}</>;
+    return children
   }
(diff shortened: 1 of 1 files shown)


thecoorum:
@cla-bot check

takameyer:
It returns the children if the `user` is logged in.  The `fallback` is there to provide a component to provide an authentication method.  What is your use case for not using the `fallback` for this?

thecoorum:
I found a workaround for my case, so I'm closing this PR

tw050x:
I found this thread after coming across the same issue. My use case is that I have no use for the fallback as login is handled elsewhere.

I had originally thought the same as @thecoorum due to the second bullet point in documentation found [here](https://www.mongodb.com/docs/realm/sdk/react-native/use-realm-react/#-userprovider-) stating that children are returned if no fallback is provided.

@takameyer Can this be reopened to get the code the align with the docs?

takameyer:
@tw050x The point of the `fallback` is to unsure that the `currentUser` is set on the `Realm.app` before rendering the child contents.  I am open to hearing about other use cases.  How would you like this changed to cover your use case?  Please feel free to open a feature request with some code examples showing the limitations of the current implementation and how it could be improved, along with current work arounds.

PR #5194, opened 2022-12-22
Refactor React Native integration tests for NPM workspaces

takameyer:
* This is inspired from the branch `kh/bindgen/react-native` @kraenhansen  and the following article: https://www.callstack.com/blog/setting-up-react-native-monorepo-with-yarn-workspaces
* Add the react-native environment to the workspaces, as well as any local dependency
* Reconfigure the react-native project to find all react-native dependencies

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen (approved):
Seems about right to me. It's annoying we need to deal with all these relative paths 😏

takameyer:
I also added rollup support and stubbed out the platform specific parts.

integration-tests/tests/package.json
```diff
@@ -19,7 +19,6 @@
     "realm": "*"
   },
   "devDependencies": {
-    "@thi.ng/bench": "^3.1.16",
```
kraenhansen, line 22: We should either find a different dependency and fix these performance tests, as I imagine we'll rely on them again very soon. Or alternatively fix the Metro config so it can deal with ES6 modules as expected.

takameyer: I'm on it right now.  I just wanted to unblock @RedBeard0531 and wasn't getting any further with this.

PR #5201, opened 2022-12-23
Kneth/rn71 ndk23

takameyer:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


scripts/build-android.js
```diff
@@ -81,7 +78,7 @@ for (const arch of architectures) {
     "-GNinja",
     `-DANDROID_NDK=${ndkPath}`,
     `-DANDROID_ABI=${arch}`,
-    `-DCMAKE_MAKE_PROGRAM=${sdkPath}/cmake/${cmakeVersion}/bin/ninja`,
+    `-DCMAKE_MAKE_PROGRAM=ninja`,
```
kraenhansen, line 81: Should we have a pre-condition check to verify that the correct version of `ninja` is going to be used?

integration-tests/tests/src/tests/sync/sync-as-local.ts
```diff
@@ -33,7 +33,6 @@ describe.skipIf(environment.missingServer, "Synced Realm as local", () => {
   });
 
   before(async function (this: RealmContext) {
-    this.timeout(5000);
```
kraenhansen, line 36: We should do something else here, to make it stable when running locally then?

takameyer: I have tried to find a way to set a default, but haven't been successful.  Perhaps we should just set it in the `package.json`

contrib/building.md
```diff
@@ -113,17 +113,16 @@ export ANDROID_EMULATOR_HOME=$HOME/.android
 export ANDROID_AVD_HOME=$HOME/.android/avd
 
 # Add the Android SDK tools to your PATH
-export PATH=$PATH:$ANDROID_SDK_ROOT/tools/bin
+export PATH=$PATH:$ANDROID_SDK_ROOT/to$ols/bin
```
kraenhansen, line 116: This looks like a mistake to me.

```suggestion
export PATH=$PATH:$ANDROID_SDK_ROOT/tools/bin
```

scripts/build-android.js
takameyer: Probably a good idea.  Merging for now, as we have complaints.  Lets look this over afterwards.

integration-tests/tests/src/tests/sync/sync-as-local.ts
takameyer: Let's look this over later

PR #5208, opened 2022-12-30
Refactor useObject

### LINKED ISSUE #4514: useObject doesn't react to the object being inserted

All the time

### Description

As a user, if I do the following `const foo = useObject('foo', fooId)` and foo is initially not in the database, I would expect the hook to re-render if the object with `fooId` is later inserted into realm. This is how most state libraries work in react. In practice, the way `useObject` works is that it will not re-render. The way to solve this is to do the following in the useObject hook:

Try to get the object by primary key. If it exists setup a listener. Otherwise listen to its collection. In the collection listener callback, if the object has been inserted, setup the listener.

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

_No response_

### Version

main

Local Database only

### Are you using encryption?

No, not using encryption

ios

### Build environment

_No response_

### Cocoapods version

_No response_

### LINKED ISSUE #5185: Component does not always re-render when array changes (useObject)

Seen always

### Description

My task app schema includes an array of subtasks. When I push a new subtask to the array, the component doesn't re-render _every time_. It only re-renders _every other_ time.

```
const TaskDisplay = ({_id})=>{

    const realm = useRealm()
    const task = useObject(Task, Realm.BSON.ObjectID(_id))

    const handleAddSubtask = (description) => {
        realm.write(() => {
          task.subtasks.push({description})
        })
    }

    console.log('RE-RENDER')

    return(
        <>
            <Button title={'Add Subtask'} onPress={()=>handleAddSubtask('This is just a subtask')}/>
        </>
    )
}
```
The interesting thing is that if I add in this Text below the Button, the component does re-render every time.

```{task ? task.subtasks.map((subtask,i)=><Text key={i}>{subtask.description}</Text>) : null}```

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

I created a sample app to demonstrate where this bug is happening for me..

1) Download and run the app
2) Click on 'Add task'
3) Click on 'Add subtask'
Expected result: 'RE-RENDER' is printed to the console
Actual result: 'RE-RENDER' is not printed to the console

https://github.com/mvrahas/task-app-test

### Version

0.4.1

Atlas Device Sync

### Are you using encryption?

Yes, using encryption

iOS 16.1.2

### Build environment

Expo 47.0.3

### Cocoapods version

_No response_

### PR DESCRIPTION
**Pull request by takameyer**
## What, How & Why?
* Prime any list properties with an cachedCollection so that updates fire correctly (Fixes #5185)
* Primary Keys as non-primative values would reset the cached objects, since their reference always changes
* Create a listener on the collection if the object doesn't exist, and rerender when it is created (Fixes #4514)

This closes #5185 and #4514

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/__tests__/useObjectRender.test.tsx
+++ b/packages/realm-react/src/__tests__/useObjectRender.test.tsx
@@ -361,7 +361,7 @@ describe("useObject: rendering objects with a Realm.List property", () => {
       const nameElement = getByTestId(`name${idString}`);
       const input = getByTestId(`input${idString}`);
 
-      expect(nameElement).toHaveTextContent(`${idString}`);
+      expect(nameElement).toHaveTextContent(idString);
 
       fireEvent.changeText(input, "apple");
 
@@ -386,7 +386,7 @@ describe("useObject: rendering objects with a Realm.List property", () => {
       const deleteButton = getByTestId(`deleteButton${idString}`);
       const nameElement = getByTestId(`name${idString}`);
 
-      expect(nameElement).toHaveTextContent(`${idString}`);
+      expect(nameElement).toHaveTextContent(idString);
       expect(itemRenderCounter).toHaveBeenCalledTimes(10);
 
       fireEvent.press(deleteButton);
--- a/packages/realm-react/src/__tests__/useObjectRender.test.tsx
+++ b/packages/realm-react/src/__tests__/useObjectRender.test.tsx
@@ -23,14 +23,14 @@ import Realm from "realm";
 import { createUseObject } from "../useObject";
 
 export class ListItem extends Realm.Object {
-  id!: number;
+  id!: Realm.BSON.ObjectId;
   name!: string;
   lists!: Realm.List<List>;
 
   static schema: Realm.ObjectSchema = {
     name: "ListItem",
     properties: {
-      id: "int",
+      id: "objectId",
       name: "string",
       lists: {
         type: "linkingObjects",
@@ -43,7 +43,7 @@ export class ListItem extends Realm.Object {
 }
 
 export class List extends Realm.Object {
-  id!: number;
+  id!: Realm.BSON.ObjectId;
   title!: string;
   items!: Realm.List<ListItem>;
   favoriteItem?: ListItem;
@@ -52,7 +52,7 @@ export class List extends Realm.Object {
   static schema: Realm.ObjectSchema = {
     name: "List",
     properties: {
-      id: "int",
+      id: "objectId",
       title: "string",
       items: "ListItem[]",
       favoriteItem: "ListItem",
@@ -81,7 +81,10 @@ const listRenderCounter = jest.fn();
 
 let testRealm: Realm = new Realm(configuration);
 
-const testCollection = [...new Array(100)].map((_, index) => ({ id: index, name: `${index}` }));
+const testCollection = [...new Array(100)].map(() => {
+  const id = new Realm.BSON.ObjectId();
+  return { id, name: id.toHexString() };
+});
 
 const useRealm = () => {
   testRealm = new Realm(configuration);
@@ -92,13 +95,15 @@ const useRealm = () => {
 
 const useObject = createUseObject(useRealm);
 
-const App = () => {
+const App = ({ renderItems = true }) => {
   return (
     <SetupComponent>
-      <TestComponent testID="testContainer" />
+      <TestComponent testID="testContainer" renderItems={renderItems} />
     </SetupComponent>
   );
 };
+
+const parentObjectId = new Realm.BSON.ObjectId();
 
 const SetupComponent = ({ children }: { children: JSX.Element }): JSX.Element | null => {
   const realm = useRealm();
@@ -113,7 +118,6 @@ const SetupComponent = ({ children }: { children: JSX.Element }): JSX.Element | 
   useEffect(() => {
     realm.write(() => {
       realm.deleteAll();
-      realm.create(List, { id: 1, title: "List", items: testCollection });
     });
     setSetupComplete(true);
   }, [realm]);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/realm-react/src/useObject.tsx [hunk 1]
```diff
@@ -82,3 +152,24 @@ export function createUseObject(useRealm: () => Realm) {
...
+  } else if (a instanceof Realm.BSON.ObjectId && b instanceof Realm.BSON.ObjectId) {
+    if (a.toHexString() === b.toHexString()) {
+      return true;
+    }
+  } else if (a instanceof Realm.BSON.UUID && b instanceof Realm.BSON.ObjectId) {
+    if (a.toHexString() === b.toHexString()) {
+      return true;
+    }
+  }
+  return false;
```
elle-j, line 174: An alternative. (And fixed a typo in the last if-block where it was `ObjectId` rather than `UUID`.)
```suggestion
  if (typeof a === "string" || typeof a === "number") {
    return a === b;
  }
  if (a instanceof Realm.BSON.ObjectId && b instanceof Realm.BSON.ObjectId) {
    return a.toHexString() === b.toHexString();
  }
  if (a instanceof Realm.BSON.UUID && b instanceof Realm.BSON.UUID) {
    return a.toHexString() === b.toHexString();
  }
  return false;
```

elle-j (approved):
Very interesting PR :) LGTM

gagik (approved):
Agree with LJ's comments but beyond those LGTM!

packages/realm-react/src/useObject.tsx
```diff
@@ -55,24 +55,94 @@ export function createUseObject(useRealm: () => Realm) {
...
+    useEffect(() => {
+      const collectionListener: Realm.CollectionChangeCallback<T & Realm.Object> = (collection, changes) => {
+        for (const index of changes.insertions) {
+          const object = collection[index];
+          const primaryKeyProperty = object.objectSchema().primaryKey;
+          if (primaryKeyProperty) {
+            //@ts-expect-error - if the primaryKeyProperty exists, then it is indexable. However, we don't allow it when we don't actually know the type of the object
+            const insertedPrimaryKey = object[primaryKeyProperty];
+            if (comparePrimaryKeys(insertedPrimaryKey, primaryKeyRef.current)) {
+              forceRerender();
```
appden, line 124: Minor: you can `break` here.

kraenhansen: Shouldn't you also `removeListener(collectionListener)` here?

packages/realm-react/src/useObject.tsx
```diff
@@ -55,24 +55,94 @@ export function createUseObject(useRealm: () => Realm) {
...
     useEffect(() => {
       return tearDown;
     }, [tearDown]);
 
+    // If the object doesn't exist, listen for insertions to the collection and force a rerender if the inserted object has the correct primary key
+    useEffect(() => {
+      const collectionListener: Realm.CollectionChangeCallback<T & Realm.Object> = (collection, changes) => {
+        for (const index of changes.insertions) {
+          const object = collection[index];
+          const primaryKeyProperty = object.objectSchema().primaryKey;
```
kraenhansen, line 119: This seem costly - is there any way this could be reused across items? They all share the same object schema.

packages/realm-react/src/useObject.tsx
```diff
@@ -55,24 +55,94 @@ export function createUseObject(useRealm: () => Realm) {
     // the cachedObject can force the component using this hook to re-render when a change occurs.
     const [, forceRerender] = useReducer((x) => x + 1, 0);
 
+    // Get the original object from the realm, so we can check if it exists
+    const originalObject = realm.objectForPrimaryKey<unknown>(type, primaryKey);
+
+    // Store the primaryKey as a ref, since when it is an objectId or UUID, it will be a new instance on every render
+    const primaryKeyRef = useRef(primaryKey);
```
kraenhansen, line 62: So what will happen if the calling code starts calling this hook with a different id? Do we have a test for that - if not, I'd suggest adding one.

packages/realm-react/src/__tests__/useObjectRender.test.tsx
```diff
@@ -157,12 +162,10 @@ const Item: React.FC<{ item: ListItem }> = React.memo(({ item }) => {
   );
 });
 
-const TestComponent: React.FC<{ testID?: string }> = ({ testID }) => {
-  const list = useObject(List, 1);
+const TestComponent: React.FC<{ testID?: string; renderItems?: boolean }> = ({ testID, renderItems }) => {
```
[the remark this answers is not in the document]
takameyer: I'll add a comment.  I needed to replicate a reported issue so I wrote a regression test.  The way to replicate is not so clearcut, so I needed make a mechanism to make it possible.

takameyer:
I've gotten all the PR changes.  Unless someone yells, I'll get this merged sometime tomorrow.

packages/realm-react/src/cachedObject.ts
```diff
@@ -49,18 +54,25 @@ type CachedObjectArgs = {
...
+export function createCachedObject({ object, realm, updateCallback }: CachedObjectArgs): CachedObject {
   const listCaches = new Map();
   const listTearDowns: Array<() => void> = [];
   // If the object doesn't exist, just return it with an noop tearDown
   if (object === null) {
     return { object, tearDown: () => undefined };
   }
 
+  // Create a cache for any Realm.List properties on the object
+  object.keys().forEach((key) => {
```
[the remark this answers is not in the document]
takameyer: done 👍🏼 thanks!

packages/realm-react/src/useObject.tsx
```diff
@@ -55,24 +55,96 @@ export function createUseObject(useRealm: () => Realm) {
...
+      if (!originalObjectRef.current) {
+        const collection = realm.objects(type);
+        collection.addListener(collectionListener);
+      }
+
+      return () => {
+        if (!originalObjectRef.current) {
+          // If the app is closing, the realm will be closed and the listener does not need to be removed
+          if (!realm.isClosed) {
+            const collection = realm.objects(type);
```
kraenhansen, line 141: I'm actually not sure what this will do. Are you sure adding a listener to one accessor object and removing it from another will actually stop notifications? I suspect you'd need to hold a reference to the original collection accessor object for this removal to work as expected.

takameyer: That's possible that this is the case and I will look into it.  

This specific case is not being tested, but should probably be proved in our unit tests somewhere.  

It is difficult to test such specifics within hooks, without creating extra an extra accessor layer on top of realm.  One could argue that we do that, but I would argue that we already have a test harness to test such specifics.

takameyer: Went over this in the morning.  At some point it would be good to find time to refactor code to be more testable.  The potential listener issues has now been fixed.

PR #5212, opened 2023-01-04
Assert ndk version

kraenhansen:
## What, How & Why?

I experienced a build failure and remembered that @takameyer mentioned that we're now building with an updated NDK. I want future contributors to fail faster.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/scripts/build-android.js
+++ b/scripts/build-android.js
@@ -21,6 +21,8 @@ const fs = require("fs-extra");
 const path = require("path");
 const exec = require("child_process").execFileSync;
 
+const EXPECTED_NDK_VERSION = "23.1.7779620";
+
 //simple validation of current directory.
 const rnDir = path.resolve(process.cwd(), "react-native");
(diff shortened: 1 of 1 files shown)


takameyer (approved):
Seems like a good addition, as I also got tripped up by this.  

One idea, that doesn't need to be addressed now.  In the environment setup for React Native the following env var is set:
```
export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
```
This folder has an `ndk` folder.  We could search if that exists there, and bypass the need to set `ANDROID_NDK` or `ANDROID_NDK_HOME`, by just looking for the desired NDK in that folder and showing an error if it's not installed.

kraenhansen:
I like that idea - I've created https://github.com/realm/realm-js/pull/5213 as a replacement for this PR.

PR #5213, opened 2023-01-04
Assert NDK version #2

kraenhansen:
## What, How & Why?

As per @takameyer's suggestion, we could simply expect that the SDK manager is used to download the NDK and infer and assert the path of the NDK relative to that. The only place the ANDROID_NDK is explicitly set is in the Dockerfile.android which isn't used by the new GHA.

This PR also updates the NDK version used on CI to explicitly match [that used by the RN template](https://github.com/facebook/react-native/blob/v0.71.0-rc.5/template/android/build.gradle#L11).

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -13,7 +13,7 @@ jobs:
     container: ${{ matrix.variant.container }}
     env:
       REALM_DISABLE_ANALYTICS: 1
-      NDK_VERSION: 23.2.8568313
+      NDK_VERSION: 23.1.7779620
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -146,8 +146,6 @@ jobs:
       - name: Build Android
         if: ${{ (matrix.variant.os == 'android') }}
         run: npm run prebuild:android -- --arch=${{matrix.variant.arch}}
-        env:
-          ANDROID_NDK: ${{ env.ANDROID_SDK_ROOT }}/ndk/${{ env.NDK_VERSION }}
```
kraenhansen, line 150: No need to provide this env variable anymore.

takameyer (approved):
LGTM :shipit:

kneth (approved):

PR #5215, opened 2023-01-04
Improve AppProvider and UserProvider rendering perf

appden:
## What, How & Why?

These providers would vend stale state into their context after props changing, which would require another unnecessary full re-render to get updated context values via `useEffect` hooks. Instead, now state is updated during render so as to immediately restart rendering before stale state is rendered down the context, which is more efficient.

For more information on this approach, please see the [Adjusting some state when a props changes](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) section of React beta docs on the "You Might Not Need an Effect" page.

### DIFF
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -16,7 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 import { isEqual } from "lodash";
-import React, { createContext, useContext, useEffect, useRef, useState } from "react";
+import React, { createContext, useContext, useLayoutEffect, useRef, useState } from "react";
 import Realm from "realm";
 
 /**
@@ -54,7 +54,7 @@ export const AppProvider: React.FC<AppProviderProps> = ({ children, appRef, ...a
     }
   }
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     if (appRef) {
       appRef.current = app;
     }
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -41,32 +41,24 @@ type AppProviderProps = Realm.AppConfiguration & {
 export const AppProvider: React.FC<AppProviderProps> = ({ children, appRef, ...appProps }) => {
   const configuration = useRef<Realm.AppConfiguration>(appProps);
 
-  const [app, setApp] = useState<Realm.App>(new Realm.App(configuration.current));
-
-  // We increment `configVersion` when a config override passed as a prop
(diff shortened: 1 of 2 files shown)


appden:
@cla-bot check

bmunkholm:
Hey Scott - lovely to see a PR from you again ;-) Thanks!

kraenhansen:
Thanks for your PR! @takameyer which is normally the one reviewing PRs for our React package is OoO until the beginning of next week. I hope we'll get this reviewed shortly thereafter.

packages/realm-react/src/AppProvider.tsx
```diff
@@ -41,32 +41,24 @@ type AppProviderProps = Realm.AppConfiguration & {
...
-        appRef.current = app;
-      }
+      setApp(new Realm.App(configuration.current));
     } catch (err) {
       console.error(err);
     }
-  }, [configVersion, setApp]);
+  }
+
+  useEffect(() => {
```
takameyer, line 57: I think this `useEffect` could also go away and the logic moved under line 51.  Thoughts?

takameyer (approved):
LGTM.  Learned something new about React today :) Thanks for the contribution.
Also good catch on the state initializers, as this would call the `Realm.App` constructor on every render.

Only have one suggestion/question.

packages/realm-react/src/AppProvider.tsx
appden: Thanks for taking a look at this @takameyer! Your asking this made me realize this *technically* should be `useLayoutEffect`, so I'll push that update. This is a nuanced one, especially with React's concurrent mode semantics coming to React Native. Refs are typically updated by React right before running layout effects, not during render since a render pass can be thrown away. While it's *fine* (ish) to update a ref you "own" during render (though that can lead to trouble too), refs you don't own should be updated as side effects because a render can be thrown away with concurrent mode. Also, another benefit here is that if `appRef` itself changes for some reason, then it'll get updated accordingly. 😀

takameyer: Now that you mention that, I will have to give the generated `RealmProvider` a look as well.  This could most likely also benefit from `useLayoutEffect` when binding the `realmRef`.

appden: Hi @takameyer, checking to see if this can be merged. I see a bunch of CI jobs are failing in the same way that appears to be unrelated...

```
An unexpected error occurred: Input required and not supplied: projectId
```

takameyer: @appden The fix is now released in [0.4.3](https://github.com/realm/realm-js/releases/tag/realm-react-v0.4.3)

PR #5216, opened 2023-01-05
Make integration tests conform to new parser

elle-j:
## What, How & Why?

This makes the integration tests conform to the new parser:
* Expected error messages have been updated.
* Tests that were using now-invalid schemas have been updated.
* The canonical object schema key `ctor` (previously `constructor`) and the canonical property schema key `name` which do not exist on the relaxed representations are no longer treated as invalid keys (even though we expect the relaxed representations). This is due to some integrations tests passing canonical objects.
* A bug with indexed properties was fixed.

Errors not fixed in this PR:
* Flexible-sync related errors.

## ☑️ ToDos
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/realm-constructor.ts
+++ b/integration-tests/tests/src/tests/realm-constructor.ts
@@ -159,7 +159,7 @@ describe("Realm#constructor", () => {
     it("fails when passed an object", () => {
       expect(() => {
         new RealmAsAny({ schema: {} });
-      }).throws("Expected 'schema (the realm schema)' to be an array, got an object");
+      }).throws("Expected 'realm schema' to be an array, got an object");
     });
 
     it("fails when passed an array with non-objects", () => {
@@ -171,19 +171,19 @@ describe("Realm#constructor", () => {
     it("fails when passed an array with empty object", () => {
       expect(() => {
         new RealmAsAny({ schema: [{}] });
-      }).throws("Expected 'name (the object schema name)' to be a string, got undefined");
+      }).throws("Expected 'name' on object schema to be a string, got undefined");
     });
 
     it("fails when passed an array with an object without 'properties'", () => {
       expect(() => {
         new RealmAsAny({ schema: [{ name: "SomeObject" }] });
-      }).throws("Expected 'SomeObject.properties' to be an object, got undefined");
+      }).throws("Expected 'properties' on 'SomeObject' to be an object, got undefined");
     });
 
     it("fails when passed an array with an object without 'name'", () => {
       expect(() => {
         new RealmAsAny({ schema: [{ properties: {} }] });
-      }).throws("Expected 'name (the object schema name)' to be a string, got undefined");
+      }).throws("Expected 'name' on object schema to be a string, got undefined");
     });
 
     function expectInvalidProperty(
@@ -192,7 +192,7 @@ describe("Realm#constructor", () => {
       addMessagePrefix = true,
     ) {
       if (addMessagePrefix) {
-        message = `Invalid type declaration for property 'InvalidObject.bad': ${message}`;
+        message = `Invalid type declaration for property 'bad' on 'InvalidObject': ${message}`;
       }
       expect(() => {
         new Realm({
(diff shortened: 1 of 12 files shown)


integration-tests/tests/src/tests/realm-constructor.ts
```diff
@@ -204,19 +211,35 @@ describe("Realm#constructor", () => {
       }).throws(message);
     }
 
-    it("fails when asking for a list of lists", () => {
+    // Old test: The new schema parser no longer allows combining shorthand and object notation.
+    it.skip("fails when asking for a list of lists", () => {
```
elle-j, line 215: I added `.skip` to some of the tests that were using schemas that are now invalid. A new test was added below. Not sure if we want to keep the `.skip` tests for reference or just delete them?

packages/realm/src/Configuration.ts
```diff
@@ -195,16 +197,32 @@ type BaseConfiguration = {
...
+  "embedded",
+  "asymmetric",
+  "properties",
+  // Not part of `ObjectSchema`
+  "ctor",
+]);
 
-const PROPERTY_SCHEMA_KEYS = new Set<keyof ObjectSchemaProperty>([
+// Need to use `CanonicalObjectSchemaProperty` rather than `ObjectSchemaProperty`
+// due to the same reasons as above.
```
elle-j, line 215: I'm using the name `ObjectSchemaProperty` here and leaving it up to the other renaming PR to modify this to `PropertySchema`.

packages/realm/src/Configuration.ts
```diff
@@ -231,7 +249,7 @@ export function validateConfiguration(config: unknown): asserts config is Config
  * Validate the data types of the fields of a user-provided realm schema.
  */
 export function validateRealmSchema(realmSchema: unknown): asserts realmSchema is Configuration["schema"][] {
-  assert.array(realmSchema, "the realm schema");
+  assert.array(realmSchema, "schema (the realm schema)");
```
kraenhansen, line 252: I'd say this would suffice - we want to keep these as short as possible to avoid information overload on the developers.

```suggestion
  assert.array(realmSchema, "realm schema");
```

In other parts of the code-base, I've actually tried to only use property names for this - to keep it even simpler, but perhaps that's too extreme.

packages/realm/src/Configuration.ts
```diff
@@ -258,11 +276,12 @@ export function validateObjectSchema(
...
-  } else {
-    // Schema is passed as an object
-    assert.object(objectSchema, "the object schema", false);
+  }
+  // Schema is passed as an object (ObjectSchema)
+  else {
+    assert.object(objectSchema, "object schema", false);
     const { name: objectName, properties, primaryKey, asymmetric, embedded } = objectSchema;
-    assert.string(objectName, "the object schema name");
+    assert.string(objectName, "name (the object schema name)");
```
kraenhansen, line 284:
```suggestion
    assert.string(objectName, "object schema name");
```

packages/realm/src/Configuration.ts
```diff
@@ -258,11 +276,12 @@ export function validateObjectSchema(
...
     assert.object(properties, `${objectName}.properties`, false);
```
kraenhansen, line 285: Just a suggestion - I think using the `.` here could cause some confusion.

```suggestion
    assert.object(properties, `properties on ${objectName}`, false);
```

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -234,6 +248,22 @@ describe("normalizePropertySchema", () => {
...
+
+    // -------------------------
+    // Indexed & Primary Keys
+    // -------------------------
+
+    itThrowsWhenNormalizing(
+      "string?",
+      "Optional properties cannot be used as a primary key.",
+      true, // Declare that it is a primary key.
+    );
```
kraenhansen, line 260: Incoming personal opinion:
For a third boolean positional argument, that you feel a need to provide a comment for, like here, it's often better for the entire function to take a single object argument or a third object argument, with a named parameter - making the code speak for itself:

```suggestion
    itThrowsWhenNormalizing(
      "string?",
      "Optional properties cannot be used as a primary key.",
      { isPrimary: true },
    );
```

kraenhansen (approved):
Minor suggestions.

integration-tests/tests/src/tests/realm-constructor.ts
kraenhansen: They can be deleted, just as we'll delete code instead of commenting it out. We can always use Git if we want to time-travel back and look at this.

packages/realm/src/Configuration.ts
elle-j: Thanks I'll update this!

packages/realm/src/Configuration.ts
elle-j: I'll go over the places where I use `.` and update it. E.g. see `Configuration.ts` where we validate the user input, I've used `"TheObjectName.thePropertyName"` but I see the confusion with how it looks like static types.

`#` was used before, can this be an option now? A good thing about being able to provide just one single "word" to the `assert` methods is due to the formatting of the error message from `TypeAssertionError.message()`.

Passing e.g. `"Person#age"` formats to `.. 'Person#age' ..`.
Passing e.g. `"age on Person"` formats to `.. 'age on Person' ..`.

We could have `TypeAssertionError.message()` accept a param to conditionally surround the value in single quotes, but then each of the methods on `assert` would also need this param.. Don't know how worth it it would be in order to instead be able to show e.g. `.. 'age' on 'Person' ..`

packages/realm/src/tests/schema-normalization.test.ts
elle-j: Amazing feedback, thank you!

kraenhansen (approved):
🚀  merge at will

PR #5247, opened 2023-01-12
move mixed tests to TS suite

JacobOscarGunnarsson:


integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,309 @@
...
+describe("Mixed", () => {
+  describe("primitive values", () => {
+    let realm: Realm;
+    beforeEach(() => {
+      Realm.clearTestState();
+      realm = new Realm({ schema: [SingleSchema] });
+    });
+    afterEach(() => {
+      realm.close();
+    });
```
kraenhansen, line 73: I would like this to use the [`openRealmBeforeEach` hook](https://github.com/realm/realm-js/blob/master/integration-tests/tests/src/hooks/open-realm-before.ts#L43) instead:

```suggestion
    openRealmBeforeEach({ schema: [SingleSchema] });
```

This calls `clearTestState` intenally by default.

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,309 @@
...
+    it("support primitives", function (this: RealmContext) {
```
kraenhansen, line 74: I would like this to use the [`openRealmBeforeEach` hook](https://github.com/realm/realm-js/blob/master/integration-tests/tests/src/hooks/open-realm-before.ts#L43) instead:

```suggestion
    openRealmBeforeEach({ schema: [SingleSchema] });
    it("support primitives", function (this: RealmContext) {
      const { realm } = this;
```

This calls `clearTestState` intenally by default.

kraenhansen:
I talked with Andrew about this in a PM.
I'd suggest we base these changes off `master` to make sure the migrated tests pass against the legacy SDK before running it against the new bindgen SDK. This way we'd be more sure that the migration of tests didn't miss something that the new SDK has introduced a bug for.

PR #5248, opened 2023-01-12
move array-buffer tests to ts-suite

JacobOscarGunnarsson:


integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,144 @@
...
+describe("ArrayBuffer", () => {
+  let realm: Realm;
+  beforeEach(() => {
+    Realm.clearTestState();
+    realm = new Realm({ schema: [SingleSchema] });
+  });
+
+  afterEach(() => {
+    realm.close();
+  });
```
kraenhansen, line 40: Same as your other PR - I'd like for us to use the `openRealmBeforeEach` hook here.

PR #5250, opened 2023-01-12
Fixing creation of named constuctors

kraenhansen:
## What, How & Why?

This fixes a "Hermes only" issue, where the name of a function implemented on a computed property of an object, got the empty string as `name`.

Note: I want to upstream an issue and patch for this, but found this workaround which honestly feels more clean than the original code did to me.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests


kraenhansen:
As this is a small change, I'll merge now - but please review in retrospect.

PR #5252, opened 2023-01-12
Upgrade to Realm Core 13.2.0

### LINKED ISSUE #5244: Update to Realm Core v13.2.0

kneth:
## What, How & Why?

This closes #5244 

## ☑️ ToDos

* [x] 📝 Changelog entry
* ~[ ] 🚦 Tests~

* ~[ ] jsdoc files updated~


lib/utils.js
```diff
@@ -134,40 +134,91 @@ module.exports = {
...
+          platform: Platform.OS,
           // Android reports a number ...
           platformVersion: `${Platform.Version}`,
+
+          deviceName,
+          deviceVersion,
+
+          cpuArch: "unknown", // Detecting CPU architecture on Android and iOS is done by seperate methods in platform.{mm, cpp)
+
+          frameworkName: "RN",
```
kraenhansen, line 184: Is this required by some spec? I think it would be most useful for a consumer of these metrics to not have it abbreviated.

```suggestion
          frameworkName: "React Native",
```

```suggestion
          frameworkName: "react-native",
```

src/android/platform.cpp
```diff
@@ -114,4 +114,15 @@ void print(const char* fmt, ...)
...
 }
+
+std::string get_cpu_arch()
+{
+#define QUOTE(name) #name
+#define STR(name) QUOTE(name)
+#define ABI_NAME STR(REALM_ANDROID_ABI)
+    return ABI_NAME;
+#undef STR
+#undef QUOTE
```
kraenhansen, line 125: This looks a bit suspicious to me 🙂 But I [read a bit more about it](https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html#Stringizing).

However, this could be simplified (including passing in the `REALM_ANDROID_ABI`) by relying on the GCC preprocessor directives:

```suggestion
#if defined(__arm__)
    return "armeabi-v7a";
#elif defined(__aarch64__)
    return "arm64-v8a";
#elif defined(__i386__)
    return "x86";
#elif defined(__x86_64__)
    return "x86_64";
#else
    return "unknown";
#endif
```

This has the benefit that we can control the values to match that of the other platforms.

lib/utils.js
```diff
@@ -134,40 +134,91 @@ module.exports = {
...
+          } else if (Platform.isTV()) {
+            deviceName = "AppleTV";
+          } else {
+            // RN doesn't support Apple Watch
+            deviceName = "iPhone";
+          }
+          deviceVersion = "unknown";
+        } else if (Platform.OS === "android") {
+          deviceName = `${Platform.Manufacturer}:${Platform.Brand}`;
+          deviceVersion = `${Platform.Model}`;
```
kraenhansen, line 168: The [react native docs](https://reactnative.dev/docs/platform#constants) on this, describes the `Model` as:
> The end-user-visible name for the Android device.

What is the intended semantics of this property exactly?

lib/utils.js
```diff
@@ -134,40 +134,91 @@ module.exports = {
...
+        if (Platform.OS === "ios") {
+          if (Platform.isPad()) {
+            deviceName = "iPad";
+          } else if (Platform.isTV()) {
+            deviceName = "AppleTV";
+          } else {
+            // RN doesn't support Apple Watch
+            deviceName = "iPhone";
+          }
+          deviceVersion = "unknown";
```
kraenhansen, line 165: You should be able to use `osVersion` here:

```suggestion
          deviceVersion = Platform.osVersion;
```

lib/utils.js
```diff
@@ -134,40 +134,91 @@ module.exports = {
...
+        } else if (Platform.OS === "android") {
+          deviceName = `${Platform.Manufacturer}:${Platform.Brand}`;
```
kraenhansen, line 167: I believe these are nested under a `constants` object:

```suggestion
          deviceName = `${Platform.constants.Manufacturer}:${Platform.constants.Brand}`;
```

kraenhansen (approved):
What specification are we implementing these versions towards? Their value should perhaps match some sort of schema to be useful?

lib/utils.js
kneth: I haven't found a way to tell me if it is an iPhone 10 or iPhone 11. `Platform.osVersion` will tell me the version of iOS or Android which is unrelated to the device.

JacobOscarGunnarsson (approved):
Aside from Kraens comments I think the changes for the sync connection parameter looks good! Also as we discussed, feel free to change the JS-specific values that are sent over to whatever you think is easiest consumable

lib/utils.js
kneth: The specification split "Samsung GT-I9100" so deviceName is "Samsung" and deviceVersion is "GT-I9100", and I have interpreted deviceVersion as the model.

src/android/platform.cpp
kneth: I am an old C hacker so I like the preprocessor 👴🏻 but your suggested is easier to read and doesn't rely on changes in `CMakeList.txt`

kneth:
> feel free to change the JS-specific values that are sent over to whatever you think is easiest consumable

I have updated the internal document to reflect the PR.

lib/utils.js
```diff
@@ -134,40 +134,92 @@ module.exports = {
...
+
     try {
       if (environment === "reactnative") {
         const { Platform } = require("react-native");
+        let deviceName = "unknown";
+        let deviceVersion = "unknown";
+        if (Platform.OS === "ios") {
+          if (Platform.isPad()) {
+            deviceName = "iPad";
+          } else if (Platform.isTV()) {
```
appden, line 159: Hey @kneth, these caused a regression since `Platform.isPad` and `Platform.isTV` are boolean properties (not methods).

```
Error getting versions: TypeError: false is not a function
```

Please see: https://reactnative.dev/docs/platform#ispad-ios

kneth: @appden Thank you for reporting. I have a fix in https://github.com/realm/realm-js/pull/5491.

PR #5254, opened 2023-01-12
migrate dictionary tests to ts

JacobOscarGunnarsson:
This one might need an extra eye since I've tried to move things from the legacy dictionary test to what I believe has been missing in the integration tests for dictionaries. Might be the case that I have missed something


integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -28,6 +28,66 @@ type Item<ValueType = Realm.Mixed> = {
...
+    const canonicalSchema: Realm.ObjectSchema = {
+      name: "CanonicalSchema",
+      properties: {
+        a: { type: "dictionary", objectType: "string", optional: false },
+        b: { type: "dictionary", objectType: "mixed", optional: true },
+        c: { type: "dictionary", objectType: "Child", optional: true },
+      },
+    };
+
+    beforeEach(() => {
```
JacobOscarGunnarsson, line 68: gonna change this as suggested in other PRs

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -333,4 +431,342 @@ describe("Dictionary", () => {
...
+          expect(cnt).equals(UPDATES + 1, "We expect ${UPDATES + 1} updates.");
+          fields.removeAllListeners();
+          realm.close();
+          resolve();
+        }, 1000);
+      });
+    });
+    // FIXME: Running the following test produce unexpected side-effects which breaks all subsequent tests.
+    // See https://github.com/realm/realm-js/issues/3834
+    it.skip("support insert notifications", () => {
```
JacobOscarGunnarsson, line 508: this wasn't used in the regular js file either but I thought it'd be worth moving over

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -333,4 +440,276 @@ describe("Dictionary", () => {
...
+        point.addListener(() => {
+          cnt++;
+        });
+      }
+
+      point.removeAllListeners();
+      this.realm.write(() => (point.x = 10));
+
+      return new Promise<void>((resolve, reject) => {
+        setTimeout(() => {
```
JacobOscarGunnarsson, line 558: I have squigglies here saying that it can't fund `setTimeout()`, but I'm given multiple import options and I can't figure out which one we want
image
image

integration-tests/tests/src/tests/dictionary.ts
takameyer: This works now, so you can uncomment and migrate this as well.

takameyer (approved):
Looks good.  Added to your comments

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -19,17 +19,89 @@
...
+  dict2: Record<string, IEmbeddedChild | null>;
+};
 type DictValues = { [key: string]: unknown };
 
 describe("Dictionary", () => {
+  describe("schema", () => {
+    it("throws on invalid schema", () => {
+      expect(() => {
+        new Realm({ schema: [DictWrongSchema] });
+      }).throws(
```
kraenhansen, line 98: You might need to `clearTestState` before this test or even better figure out what other test might be causing the error on CI (electron main process failed).

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -19,17 +19,89 @@
...
+        "Schema validation failed due to the following errors:\n- Property 'Dictionary.a' of type 'dictionary' has unknown object type 'wwwww'",
+      );
+    });
+  });
   describe("with unconstrained (mixed) values", () => {
-    openRealmBefore({
+    openRealmBeforeEach({
```
kraenhansen, line 104: I'm curious why this was needed. Every test might as well create their own `Item` object to use in their test and avoid cross talking that way.

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -205,10 +277,46 @@ describe("Dictionary", () => {
...
+      expect(item.dict.c).equals("hey");
+      item = this.realm.write(() => {
+        return this.realm.create<Item>("Item", {
+          dict: { a: 0, b: 0, c: "new value" },
+        });
+      });
+      expect(item.dict.a).equals(0);
+      expect(item.dict.b).equals(0);
+      expect(item.dict.c).equals("new value");
+    });
```
kraenhansen, line 297:
```suggestion
    });

```

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -205,10 +277,46 @@ describe("Dictionary", () => {
...
   });
 
   describe("toJSON", function () {
-    openRealmBefore({
+    openRealmBeforeEach({
```
kraenhansen, line 319: Same here. Couldn't the realm be used across tests here?

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -244,7 +352,7 @@ describe("Dictionary", () => {
 
   function describeTypedSuite({ extraSchema = [], type, goodValues, badValues, expectedError }: TypedDictionarySuite) {
     return describe(`with constrained '${type}' values`, () => {
-      openRealmBefore({
+      openRealmBeforeEach({
```
kraenhansen, line 355: And here 🙂

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -333,4 +441,262 @@ describe("Dictionary", () => {
...
+
+      await sleep(1000);
+      expect(this.realm.objects<IDictSchema>(DictSchema.name)[0].fields.field1).equals(5);
+      expect(cnt).equals(UPDATES + 1, "We expect ${UPDATES + 1} updates.");
+      fields.removeAllListeners();
+      this.realm.close();
+    });
+    // FIXME: Running the following test produce unexpected side-effects which breaks all subsequent tests.
+    // See https://github.com/realm/realm-js/issues/3834
+    it.skip("support insert notifications", async function (this: RealmContext) {
```
kraenhansen, line 494: Did you verify this is still the case?

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -333,4 +441,262 @@ describe("Dictionary", () => {
...
+
+      const half = data.filtered("fields['x'] >= 50");
+      const seventy = data.filtered("fields['x'] >= $0", 70);
+      expect(half.length).equals(N / 2, "We expect only 50 items, matching for field x.");
+      expect(seventy.length).equals(30, "We expect only 30 items, matching for field x.");
+    });
+  });
+  describe("notifications", () => {
+    openRealmBeforeEach({ schema: [DictSchema] });
+    it("support update notifications", async function (this: RealmContext) {
```
kraenhansen, line 463:
```suggestion
    // TODO: Refactor to not use a `sleep` and instead resolve a promise when the listener has fired the expected amount of times (and throw it that gets exceeded)
    it("support update notifications", async function (this: RealmContext) {
```

kraenhansen (approved):
I have nothing major here. Look forward to these getting merged and tested on `bindgen` so we can get back to refactoring these too.
Like Andrew mentioned, an empty newline between tests would improve readability.

integration-tests/tests/src/tests/dictionary.ts
JacobOscarGunnarsson: thought I did but after having a second glance at it I realised that I had written the tests in an invalid way, should work now! (If CI lets it through)

tests/js/dictionary-tests.js
```diff
@@ -1,810 +0,0 @@
...
-
-    let { x, y, z } = point;
-    TestCase.assertEqual(x, 1, "Should be an equals to: [1,3,2]");
-    TestCase.assertEqual(y, 2, "Should be an equals to: [1,3,2]");
-    TestCase.assertEqual(z, 3, "Should be an equals to: [1,3,2]");
-
-    TestCase.assertEqual(point[Symbol.for("x")], 1, "Should work with symbols");
-
-    realm.close();
-  },
```
kraenhansen, line 281: @JacobOscarGunnarsson it seem to me there's quite a lot of these tests that wasn't migrated?

PR #5255, opened 2023-01-13
Move array-buffer tests to integration tests

JacobOscarGunnarsson:


integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,230 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { Buffer } from "buffer";
+import { expect } from "chai";
+import Realm from "realm";
+import { openRealmBeforeEach } from "../hooks";
+
+const isNodeProcess = typeof process === "object" && process + "" === "[object process]";
```
JacobOscarGunnarsson, line 24: I have squigglys here telling me to do the following.
image
As I'm very unfamiliar with this area I thought it'd be worth asking here whether I should do it or not

kraenhansen: Nope. If you have Node.js specific tests, you can do one of three things:
- Determine if they're really Node.js specific or could be generalised a bit to make them run on every platform.
- If they really do test Node.js specific behaviour:
  - You can use the `describe.skipIf` or `it.skipIf` to skip the test if the `environment.node` isn't true. Here's [an example where we skip a test on Android](https://github.com/realm/realm-js/blob/master/integration-tests/tests/src/tests/realm-constructor.ts#L308).
  - If your test depends on importing a Node.js specific API (other than `path` or `fs` which are already injected as globals), you can move the test to the Node.js specific part of our test suite: https://github.com/realm/realm-js/tree/master/integration-tests/tests/src/node ... there you'll have all the Node.js runtime APIs available (such as the `process` global).

In this specific case, I'd probably use `it.skipIf(!environment.node, ...` for the "supports arrayBuffer-nodeBuffer" test and try removing the early return from the "supports inserting empty arrayBuffer" test to see if / why it doesn't work on other platforms as well.

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+      0x38,
+      0x48,
+      0x4e,
+      0xff,
+    ]);
+
+    this.realm.write(() => this.realm.create(SingleSchema.name, { a: view }));
+
+    const data = this.realm.objects<ISingleSchema>(SingleSchema.name)[0];
+    expect(data.a.byteLength).equals(view.byteLength, "Data size should be equals");
```
takameyer, line 76:
```suggestion
    expect(data.a.byteLength).equals(view.byteLength, "Data size should be equal");
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    const binary_view = new Uint8Array(data.a);
+
+    for (let i = 0; i < view.byteLength; i++) {
+      const p1 = view[i];
+      const p2 = binary_view[i];
+      expect(p1).equals(p2, "buffers should be equals");
```
takameyer, line 83:
```suggestion
      expect(p1).equals(p2, "buffers should be equal");
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+      0x38,
+      0x48,
+      0x4e,
+      0xff,
+    ]);
+
+    this.realm.write(() => this.realm.create(SingleSchema.name, { a: new DataView(view.buffer) }));
+
+    const data = this.realm.objects<ISingleSchema>(SingleSchema.name)[0];
+    expect(data.a.byteLength).equals(view.byteLength, "Data size should be equals");
```
takameyer, line 125:
```suggestion
    expect(data.a.byteLength).equals(view.byteLength, "Data size should be equal");
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    const binary_view = new Uint8Array(data.a);
+
+    for (let i = 0; i < view.byteLength; i++) {
+      const p1 = view[i];
+      const p2 = binary_view[i];
+
+      expect(p1).equals(p2, "buffers should be equals");
```
takameyer, line 133:
```suggestion
      expect(p1).equals(p2, "buffers should be equal");
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+      const p2 = ab_view[i];
+      expect(p1).equals(p2, "buffers should be equals");
+    }
+  });
+  it.skipIf(!environment.node, "supports arrayBuffer-nodeBuffer", function (this: RealmContext) {
+    const n_buffer = Buffer.from([0xca, 0xfe, 0xba, 0xbe]);
+    this.realm.write(() => this.realm.create(SingleSchema.name, { a: n_buffer }));
+
+    const data = this.realm.objects<ISingleSchema>(SingleSchema.name)[0];
+    expect(data.a.byteLength).equals(n_buffer.byteLength, "Data size should be equals");
```
takameyer, line 191:
```suggestion
    expect(data.a.byteLength).equals(n_buffer.byteLength, "Data size should be equal");
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+  properties: {
+    a: "data",
+  },
+};
+
+type ISingleSchema = {
+  a: ArrayBuffer;
+};
+
+describe("ArrayBuffer", () => {
```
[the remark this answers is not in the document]
takameyer: This would result in "ArrayBuffer type supports storing data from a built-in ArrayBuffer".  

Not a big deal though.

takameyer (approved):
Just a few suggestions.  Nice!

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+      0x70,
+      0x38,
+      0x48,
+      0x4e,
+      0xff,
+    ]);
+
+    this.realm.write(() => this.realm.create(SingleSchema.name, { a: new DataView(view.buffer) }));
+
+    const data = this.realm.objects<ISingleSchema>(SingleSchema.name)[0];
```
kraenhansen, line 124:
```suggestion
    const data = this.realm.write(() => this.realm.create(SingleSchema.name, { a: new DataView(view.buffer) }));
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+  },
+};
+
+type ISingleSchema = {
+  a: ArrayBuffer;
+};
+
+describe("ArrayBuffer type", () => {
+  openRealmBeforeEach({ schema: [SingleSchema] });
+  it("supports storing data from a bufferView", function (this: RealmContext) {
```
kraenhansen, line 37:
```suggestion
  it("supports storing data from a BufferView", function (this: RealmContext) {
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    const binary_view = new Uint8Array(data.a);
+
+    for (let i = 0; i < view.byteLength; i++) {
+      const p1 = view[i];
+      const p2 = binary_view[i];
+      expect(p1).equals(p2, "buffers should be equal");
+    }
+  });
+  it("supports storing data from a bufferDataView", function (this: RealmContext) {
```
kraenhansen, line 86: I suspect this refers to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView

```suggestion
  it("supports storing data from a DataView", function (this: RealmContext) {
```

integration-tests/tests/src/tests/array-buffer.ts
```diff
@@ -0,0 +1,219 @@
...
+    const binary_view = new Uint8Array(data.a);
+
+    for (let i = 0; i < view.byteLength; i++) {
+      const p1 = view[i];
+      const p2 = binary_view[i];
+
+      expect(p1).equals(p2, "buffers should be equal");
+    }
+  });
+  it("supports storing data from an arrayBuffer", function (this: RealmContext) {
```
kraenhansen, line 136: Since this is referring to an actual type: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer

```suggestion
  it("supports storing data from an ArrayBuffer", function (this: RealmContext) {
```

It could be updated in more places below too.

kraenhansen (approved):
Just minor comments.
Thanks for doing this!

PR #5256, opened 2023-01-13
move mixed tests to integration tests

JacobOscarGunnarsson:


integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,312 @@
...
+
+      mixedObjects = this.realm.objects<IMixedSchema>("MixedClass");
+      returnedData = [...new Uint8Array(mixedObjects[0].value as Iterable<number>)];
+      expect(returnedData).eql(uint8Values2);
+
+      this.realm.write(() => {
+        this.realm.deleteAll();
+      });
+
+      // Test with empty array
```
JacobOscarGunnarsson, line 271: this part still fails, will have a meeting about it later

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,312 @@
...
+      this.realm.write(() => ((data.a = null), "Should be the same null"));
+      expect(data.a).equals(null);
+
+      this.realm.write(() => ((data.a = undefined), "Should be the same null"));
+      expect(data.a).equals(null);
+    });
+    it("wrong type throws", function (this: RealmContext) {
+      expect(() => {
+        this.realm.write(() => this.realm.create(SingleSchema.name, { a: Object.create({}) }));
+      }).throws(Error, "Only Realm instances are supported.");
```
JacobOscarGunnarsson, line 166: in the previous branch the error message thrown was `Unable to convert an object with ctor 'Object' to a Mixed` which makes way more sense.. I don't know if I should either 1: leave it as is with an unclear error message, 2: change back to the regular string and leave the test as failing or 3: deal with why the wrong message is thrown.

takameyer (approved):
LGTM 👍🏼 Just some small things that were probably there in the legacy tests.

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,313 @@
...
+
+    it("support complex primitive types", function (this: RealmContext) {
+      const d128 = Decimal128.fromString("6.022e23");
+      const oid = new ObjectId();
+      const uuid = new UUID();
+      const date = new Date();
+
+      const data: ISingleSchema = this.realm.write(() =>
+        this.realm.create(SingleSchema.name, { a: oid, b: uuid, c: d128, d: date }),
+      );
```
kraenhansen, line 126: You might as well move the type to the `create` so you also get correct types on the second "values" argument too.

```suggestion
      const data = this.realm.write(() =>
        this.realm.create<ISingleSchema>(SingleSchema.name, { a: oid, b: uuid, c: d128, d: date }),
      );
```

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,313 @@
...
+    it("support nested types", function (this: RealmContext) {
+      this.realm.write(() => {
+        const r = this.realm.create(VertexSchema.name, { a: 1, b: 0, c: 0 });
+        const r2 = this.realm.create(VertexSchema.name, { a: 0, b: 1, c: 0 });
+        const r3 = this.realm.create(VertexSchema.name, { a: 0, b: 0, c: 1 });
+
+        this.realm.create(SingleSchema.name, { a: r, b: r2, c: r3, d: null });
+      });
+
+      let data: ISingleSchema = this.realm.objects<ISingleSchema>(SingleSchema.name)[0];
```
kraenhansen, line 181: Doing it this way, your test would be more robust to other tests that might write to the collection before it.

```suggestion
      let data = this.realm.write(() => {
        const r = this.realm.create(VertexSchema.name, { a: 1, b: 0, c: 0 });
        const r2 = this.realm.create(VertexSchema.name, { a: 0, b: 1, c: 0 });
        const r3 = this.realm.create(VertexSchema.name, { a: 0, b: 0, c: 1 });

        this.realm.create<ISingleSchema>(SingleSchema.name, { a: r, b: r2, c: r3, d: null });
      });
```

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -0,0 +1,313 @@
...
+      expect((data.c as IVertexSchema).b).equals(0, "Should be equal 0");
+      expect((data.c as IVertexSchema).c).equals(1, "Should be equal 1");
+
+      this.realm.write(() => {
+        const r = this.realm.create(MixNestedSchema.name, { a: 0, b: -1 });
+        const r1 = this.realm.create(MixNestedSchema.name, { a: 1, b: 0 });
+        this.realm.create(SingleSchema.name, { a: r, b: r1 });
+      });
+
+      data = this.realm.objects<ISingleSchema>(SingleSchema.name)[1];
```
kraenhansen, line 201:
```suggestion
      data = this.realm.write(() => {
        const r = this.realm.create(MixNestedSchema.name, { a: 0, b: -1 });
        const r1 = this.realm.create(MixNestedSchema.name, { a: 1, b: 0 });
        return this.realm.create<ISingleSchema>(SingleSchema.name, { a: r, b: r1 });
      });
```

kraenhansen, line 201: I'd also suggest naming this differently and use two `const` instead of a `let`.

kraenhansen (approved):
It's a bit hard for me to avoid making these suggestions. I'm okay with merging this as is and then getting back to this once we refactor.

PR #5258, opened 2023-01-13
Testing Bindgen SDK on React Native

kraenhansen:
## What, How & Why?

This continues the work of @RedBeard0531 and @takameyer to force the codebase into running the integration tests on React Native.

It picks the two commits off `andrew/bindgen/mathias-android` and adds temporary workarounds:
- ~Making the FS operations and all calling functions that we expose asynchronous.~
- ~Disables a failing JWT test.~
- ~Disables the use of WeakRef (which might be coming to the RN version of Hermes soon).~
- ~Disables a crashing test, related to queries on float properties.~

~This is work in progress and shouldn't be merged.~

### DIFF
--- a/src/android/CMakeLists.txt
+++ b/src/android/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.18.1)
 
+set(REACT_NATIVE_VERSION "0.71.0-rc.5")
+
 set(REACT_NATIVE_ROOT_DIR "${PACKAGE_ROOT_DIR}/node_modules/react-native")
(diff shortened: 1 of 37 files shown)


src/android/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.18.1)
 
+set(REACT_NATIVE_VERSION "0.71.0-rc.5")
```
kraenhansen, line 3: @RedBeard0531 should we capture an issue to get this derived? I also see it defined on line 7 of this file - any idea why it needed to go here too?

PR #5268, opened 2023-01-16
move linking-objects tests to ts

JacobOscarGunnarsson:


integration-tests/tests/src/tests/linking-objects.ts
```diff
@@ -0,0 +1,326 @@
...
+      });
+
+      this.realm.write(() => {
+        this.realm.delete(olivier);
+
+        expect(oliviersParents.length).equals(0);
+      });
+    });
+    it("filters work on linking object", function (this: RealmContext) {
+      let christine, olivier;
```
takameyer, line 152: could also write as:
```
const {christine, olivier} = realm.write(() => {
//...
return {christine, olivier}
}
```

takameyer (approved):
Look good!  Just a small suggestion

kraenhansen (approved):
Looks like you're missing a delete of the legacy test file (or perhaps you'll do that just before merging?)

PR #5280, opened 2023-01-17
Update size of Android binaries. Update compatibility matrix.

kneth:


takameyer:
Looks like CI is getting `Invalid Realm passed to bind_to_context()`.  We should probably investigate this somehow.

PR #5285, opened 2023-01-18
Migrate migration tests to integration tests

gagik:
A migration of migration tests.

### DIFF
--- a/integration-tests/tests/src/tests/migrations.ts
+++ b/integration-tests/tests/src/tests/migrations.ts
@@ -22,16 +22,16 @@ import { DogSchema, IDog, IPerson, PersonSchema } from "../schemas/person-and-do
 const TestSchema = {
   name: "Test",
   properties: {
-    prop0: "string",
-    prop1: "int",
+    name: "string",
+    age: "int",
   },
 };
 interface Test {
-  prop0: string;
-  prop1: number;
+  name: string;
+  age: number;
 }
 interface TestWithNewProperty extends Test {
-  prop2: string;
+  surname: string;
 }
 
 describe("Migrations", () => {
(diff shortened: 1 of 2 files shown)


integration-tests/tests/src/tests/migrations.ts
```diff
@@ -0,0 +1,521 @@
...
+        schemaVersion: 2,
+        onMigration: function (oldRealm, newRealm) {
+          // deleting a model which isn't target of linkingObjects works fine
+          newRealm.deleteModel("Dog");
+        },
+      });
+
+      expect(realm.objects("Person").length).equals(1);
+      expect(realm.objects("Dog").length).equals(0);
+      expect(realm.objects<IPerson>("Person")[0].dogs).equals(undefined);
```
gagik, line 457: This used to check equivalent of `expect(realm.objects<IPerson>("Person")[0].dogs.length).equals(0);` but seems like this behavior has changed now so I am now testing for this (which in my opinion makes more sense in a way). Is this intentional?

integration-tests/tests/src/tests/migrations.ts
```diff
@@ -0,0 +1,521 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+import Realm from "realm";
+import { expect } from "chai";
+import { DogSchema, IDog, IPerson, PersonSchema } from "../schemas/person-and-dogs";
+
+const TestSchema = {
+  name: "Test",
+  properties: {
+    prop0: "string",
```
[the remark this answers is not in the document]
takameyer: At least in tests, I wouldn't want to make a shared schema as this creates coupling between test suites.  I would like to avoid this, as when one needs to expand on the shared schema, it effects unrelated tests and could lead to unintended failures.

I do agree though, that the property names could be improved.

integration-tests/tests/src/tests/migrations.ts
```diff
@@ -0,0 +1,521 @@
...
+      expect(realm.objects("Test").length).equals(0); // objects should be gone
+
+      realm.close();
+
+      realm = new Realm({ schema: [TestSchema], schemaVersion: 2 });
+      expect(realm.objects("Test").length).equals(0);
+
+      realm.close();
+    });
+    it("should ignore deleteModel with non-existing models", () => {
```
takameyer, line 376: Our linter never picks this up, but I would prefer new lines between `it` clauses.

integration-tests/tests/src/tests/migrations.ts
takameyer: That doesn't seem correct, dogs is not supposed to be undefined and should always default to an empty list.  The old behavior was correct.

takameyer (approved):
LGTM

integration-tests/tests/src/tests/migrations.ts
gagik: I was able to find my error here and correct it, it had to do with the schemas I was passing.

integration-tests/tests/src/tests/migrations.ts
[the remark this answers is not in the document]
gagik: I am changing `prop0 => name, prop1 => age, prop2 => surname`

PR #5288, opened 2023-01-18
Fix a bug in parsing client reset configuration

kneth:
## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -61,6 +61,9 @@ module.exports = function (realmConstructor) {
   realmConstructor._UUID = realmConstructor.BSON.UUID;
   const { DefaultNetworkTransport } = require("@realm/network-transport");
   realmConstructor._networkTransport = new DefaultNetworkTransport();
+  realmConstructor._defaultOnDiscardCallback = (_, path) => console.log(`ClientReset onDiscard: path = ${path}`);
+  realmConstructor._defaultOnRecoveryCallback = (beforeRealm, afterRealm) =>
+    console.log(`ClientReset onRecovery: before path = ${beforeRealm.path}, after path = ${afterRealm.path}`);
 
   // Adds to cache when serializing an object for toJSON
   const addToCache = (cache, realmObj, value) => {
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -488,6 +488,7 @@ module.exports = function (realmConstructor) {
 
     realmConstructor.ClientResetMode = {
       Manual: "manual",
+      DiscardLocal: "discardLocal",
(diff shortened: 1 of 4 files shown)


takameyer (approved):
:shipit:

lib/extensions.js
```diff
@@ -61,6 +61,9 @@ module.exports = function (realmConstructor) {
   realmConstructor._UUID = realmConstructor.BSON.UUID;
   const { DefaultNetworkTransport } = require("@realm/network-transport");
   realmConstructor._networkTransport = new DefaultNetworkTransport();
+  realmConstructor._defaultOnDiscardCallback = (_, path) => console.log(`ClientReset onDiscard: path = ${path}`);
+  realmConstructor._defaultOnRecoveryCallback = (beforeRealm, afterRealm) =>
+    console.log(`ClientReset onRecovery: before path = ${beforeRealm.path}, after path = ${afterRealm.path}`);
```
kraenhansen, line 66: Or perhaps, just don't do anything at all?

src/js_sync.hpp
```diff
@@ -1303,21 +1303,23 @@ void SyncClass<T>::populate_sync_config(ContextType ctx, ObjectType realm_constr
...
+                            return Value::validated_to_function(ctx, callback_value);
+                        }
+                    };
 
                     FunctionType client_reset_discard_callback =
-                        Value::validated_to_function(ctx, client_reset_after_value);
+                        get_callback("onDiscard", "_defaultOnDiscardCallback");
                     FunctionType client_reset_recovery_callback =
-                        Value::validated_to_function(ctx, client_reset_recovery_value);
+                        get_callback("onRecovery", "_defaultOnRecoveryCallback");
```
kraenhansen, line 1322: I suppose that would make the two parameters optional then? In which case we should update the types: https://github.com/realm/realm-js/blob/v11/types/index.d.ts#L178C7-L179

lib/extensions.js
kneth: @takameyer You are thinking the same, right?

takameyer: Yeah, I was also thinking we could just do nothing there.  However, if we defaulted a `console.error` for sync errors, then it would also be visible that a client reset occurred.

takameyer (approved):
Other than possibly removing the `console.log` from the default callbacks, LGTM

elle-j (approved):
LGTM! :)

lib/extensions.js
kneth: It looks like @takameyer is in favor of the same.

PR #5290, opened 2023-01-18
Migrate set tests to integration tests 

gagik:


integration-tests/tests/src/tests/sets.ts
```diff
@@ -0,0 +1,493 @@
...
+      realm.write(() => {
+        // insert two people
+        realm.create(teamSchema.name, {
+          names: ["John", "Sue"],
+        });
+      });
+
+      let teams = realm.objects<Team>(teamSchema.name);
+      expect(teams.length).equals(1, "There should be one team");
+      expect(teams[0].names.size).equals(2, "Team Set size should be 2");
```
kraenhansen, line 84: This way we avoid implicitly testing `Results` too.

```suggestion
      const team = realm.write(() => {
        // insert two people
        return realm.create(teamSchema.name, {
          names: ["John", "Sue"],
        });
      });

      expect(team.names.size).equals(2, "Team Set size should be 2");
```

integration-tests/tests/src/tests/sets.ts
```diff
@@ -0,0 +1,493 @@
...
+        mixedValues.push(intObject);
+      });
+
+      realm.write(() => {
+        realm.create(MixedObjectSchema.name, { mixedField: mixedValues });
+      });
+
+      const objs = realm.objects<MixedObject>(MixedObjectSchema.name);
+      expect(1).equals(objs.length, "One MixedObject");
+      expect(mixedValues.length).equals(objs[0].mixedField.size, `${mixedValues.length} values in set`);
```
kraenhansen, line 141: As with my previous comment, I'd suggest holding on to the created object instead of going through the `Results`.

```suggestion
      const obj = realm.write(() => {
        return realm.create(MixedObjectSchema.name, { mixedField: mixedValues });
      });

      expect(mixedValues.length).equals(obj.mixedField.size, `${mixedValues.length} values in set`);
```

kraenhansen (approved):
Great job.
We should definitely clean this up once we've got it running against RN on `bindgen`.

integration-tests/tests/src/tests/sets.ts
```diff
@@ -0,0 +1,493 @@
...
+    name: "Team",
+    properties: {
+      names: "string<>",
+    },
+  };
+  interface Team {
+    names: Realm.Set<string>;
+  }
+
+  describe("Schema", () => {
```
takameyer, line 46: This could later be removed, as the tests for parsing schemas should cover this.

takameyer: We would need some sort of reminder.  We should be tracking what should later be refactored.

takameyer (approved):
LGTM 👍🏼

integration-tests/tests/src/tests/sets.ts
gagik: Would a TODO comment be sufficient or is it better to have something specific to this test refactoring, TOREFACTOR or something? @kraenhansen

kraenhansen: `// TODO` is fine 🙂
(alternatively TOREDO? 🙈 )

gagik: I guess I can just TODO then, it'll all be in same directory so should be pretty easy to figure out.

Issue #2009, opened 2018-09-06
Bulk insert into Realm


ogtimothymiller:
## Goals

Inserting collection of realm models into Realm

## Expected Results

insertOrUpdate like Realm Java 1.1.0+ (https://realm.io/blog/realm-java-1-1-0/)

## Actual Results

Can only insert new objects one at a time.

## Steps to Reproduce

See above

## Code Sample

```
realm.write(() => {
                        realm.create('ObjectName', {
                            foo: foo, baz: baz
                        })
                    })
```

## Version of Realm and Tooling

- Realm JS SDK Version: 2.15.3
- Node or React Native: Node
- Client OS & Version: macOS 10.13.6

nirinchev:
Do you have a reason to believe the performance of inserting objects one at a time is not sufficient for your use case?

ogtimothymiller:
I am inserting tens of thousands of objects, which takes a decent while to process.

It's strange java received this feature last year, and javascript does not have it after all this time.

ogtimothymiller:
I should clarify, this is a server-side node environment. I am inserting new data into the realm DB and client apps download a subset of this data.

bmunkholm:
@ogtimothymiller As long as you insert all your objects in the same transaction that should make it as fast as possible. It may be that other smaller things can be optimized, but the biggest gain is one transaction.
I've added a "Help Wanted" if anyone is interested in exploring optimization as it can be done entirely in this repo.

HughBerriez:
@nirinchev Am I missing something? I'm working on a react native project for a company that wants to persist 35,000+ objects locally (Android focused, at present). It seems the Realm Javascript api, Realm.create takes 47-50 seconds to write 1000 objects. Is there a sped up realm.insert* api ready and available to possibly speed things up? If not, I'll sadly have to make the decision to switch over to SQLite as our localDB solution. Hopefully I'm missing something - as the Java and Swift apis are fully fledged and seem to not have this problem.

bmunkholm:
@HughBerriez It shouldn't be that slow. Are you doing individual transactions or one big transaction?
If you want us to look closer at it, please create another issue, preferably with the code you use to insert.

HughBerriez:
[...]

PR #5299, opened 2023-01-20
Test Migration: Queries test

gagik:
The test are closer to being understandable than they used to be... but probably makes no sense to have such an outlier testing methodology for this specific test so could use refactoring _in the future?_


integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,11 +15,841 @@
...
+    realm.close();
+  });
+});
+
+import { IPerson, PersonSchema } from "../schemas/person-and-dogs";
+import { IContact, ContactSchema } from "../schemas/contact";
 
-import { IPerson, Person, PersonSchema } from "../schemas/person-and-dogs";
-import { IContact, Contact, ContactSchema } from "../schemas/contact";
 interface IPrimitive {
```
gagik, line 877: I left the old tests basically untouched and separate but I could also incorporate them in a larger describe.

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,11 +15,865 @@
...
+const SharedTestSuiteCases: Record<string, TestCase> = {
+  dateTests: {
+    schema: [{ name: "DateObject", properties: [{ name: "dateCol", type: "date" }] }],
+    objects: [
+      { type: "DateObject", value: [0] },
+      { type: "DateObject", value: [1] },
+      { type: "DateObject", value: [2] },
+    ],
+    tests: [
+      ["AssertLength", "DateObject", 2, "dateCol < $0", [2, "dateCol"]],
```
takameyer, line 68: We definitely need to refactor this.  Anyone starting to read this code from top to bottom will become immediately clueless as to what any of these tests means.  What may be a good alternative before merging is to add some comments about what these array values actually do.  This could help with future refactoring work.

integration-tests/tests/src/tests/queries.ts
takameyer: I think it's fine to leave them for now.  Maybe just add a comment about how these tests will be combined with the above in a later refactoring effort.

takameyer (approved):
Just want to take the opportunity to add some clarification so to the old test data, to make the future refactoring effort smoother.
Otherwise, LGTM and good work taking this one on.

kraenhansen (approved):
Minor suggestions (to refactoring that you might want to postpone).

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,11 +15,884 @@
...
+      ["AssertLength", "FloatObject", 2, "floatCol >= 0.0"],
+
+      ["AssertException", "FloatObject", "Unsupported comparison between type", "floatCol == false"],
+      ["AssertException", "FloatObject", "Cannot convert", "floatCol == 'not a float'"],
+      ["AssertException", "FloatObject", "Unsupported comparison between type", "floatCol == $0", "not a float"],
+      ["AssertException", "FloatObject", "Unsupported comparison operator", "floatCol BEGINSWITH 1"],
+      ["AssertException", "FloatObject", "Unsupported comparison operator", "floatCol CONTAINS 1"],
+      ["AssertException", "FloatObject", "Unsupported comparison operator", "floatCol ENDSWITH 1"],
+
+      // ["Disabled", "AssertException", "FloatObject", "floatCol = 3.5e+38"],
```
gagik, line 195: Not too sure what this is trying to test (there is no exception being thrown right now but I am not sure what expected behavior is these days considering this is from many years ago), should I remove this, is there meant to be support for "3e+12.." format? I can also add a regular test case for that. @takameyer @kraenhansen

takameyer: Interesting.  Perhaps it passes only in Node and not on React Native.  One would have to test.

kraenhansen: We had a conversation on Slack too:
> I'll lean towards having anything expressible as a JS `Number` should persist (as best as possible - with potential minor loss of precision) to the database file, without throwing.

gagik: I think then I will add a proper test case for this and see what happens, ty all

gagik:
Two notes here: @takameyer  @kraenhansen 
- I added some manual tests that seem to show that scientific notation queries seem to work fine across platforms on CI so I will keep that test as expected behavior instead of expecting an exception.
- Basically a good chunk of the test suite somewhat depends on support on passing schema configuration as an array which I was told is going away in bindgen, so makes it somewhat useless for our purposes, making it more important to refactor it now rather than later. I just thought of a couple of ways to make this a much nicer testing wise and essentially refactor it further without too much work, so I think I'll spend a bit of today doing that

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,129 +15,1026 @@
...
+      results = results.filtered(queryString, ...queryArgs);
+      expect(expectedResults.length).equals(results.length);
+      expect(expectedResults).to.deep.equal(results.map((el) => el[propertyToCompare]));
+    });
+  };
+
+  beforeEach(Realm.clearTestState);
+
+  describe("Basic types", () => {
+    openRealmBeforeEach({ schema: [NullableTypesObject] });
```
gagik, line 139: So I did a pretty large refactor again since to me it started seeming like trying to adapt the old structure wasn't actually saving us time and it would be good to just deal with this once and for all. Now it is still using the array test running structure but I hope it is much cleaner. I also combined some older tests together, so actually the file is more or less the same size and way less refactoring needed.

Some decisions that I made that may be worth disccusion:
1. I tried to avoid redeclaring schemas every time so `NullableTypesObject` is used quite a bit in basic types tests. It might still be cleaner to just open a new schema with single field every time, I just like having one `openRealmBeforeEach`.
2. I removed the separate optional types / null test and instead just mixed in null/undefined values into other type tests.

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,129 +15,1031 @@
...
+
+          [2, "uuidCol == null"],
+          [2, "null == uuidCol"],
+          [5, "uuidCol != null"],
+        ]);
+      });
+
+      it("throws with invalid queries", function (this: RealmContext) {
+        assertQueryException(this.realm, NullableTypesObject, [
+          ["Unsupported comparison between type", "dateCol == 'not a date'"],
```
takameyer, line 546: This is comparing dates instead of uuids?

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,129 +15,1031 @@
...
+          [5, "dataCol != $0", objects[0].dataCol],
+          [1, "dataCol BEGINSWITH $0", objects[0].dataCol],
+          [2, "dataCol BEGINSWITH $0", objects[1].dataCol],
+          [2, "dataCol ENDSWITH $0", objects[4].dataCol],
+          [3, "dataCol CONTAINS $0", objects[2].dataCol],
+
+          [1, "dataCol == null"],
+          [5, "dataCol != $0", null],
+        ]);
+      });
```
takameyer, line 443: Maybe make a todo for the types that are missing "invalid queries".  Currently they are missing for objectId and data.  Perhaps there are others.

integration-tests/tests/src/tests/queries.ts
takameyer: Wow, I can actually read the tests without jumping around the code a bunch.  Awesome job!

takameyer (approved):
Happy to see these tests finally in a readable state!  Noticed a couple things now that the code is readable (probably existed before the migration).  LGTM 🎉

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,129 +15,1037 @@
...
+      expect(expectedResults).to.deep.equal(
+        results.map((el) => el[propertyToCompare]),
+        `
+        Expected results not returned from query: ${queryString} ${JSON.stringify(queryArgs)}, 
+      `,
+      );
+    });
+  };
+
+  beforeEach(Realm.clearTestState);
```
kraenhansen, line 136: When calling `openRealmBeforeEach` an after-each hook calling `clearTestState` will also be registered, in which case I'd expect this is not needed.
```suggestion
```

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,129 +15,1037 @@
...
+      results = results.filtered(queryString, ...queryArgs);
+      expect(expectedResults.length).equals(results.length);
+      expect(expectedResults).to.deep.equal(
+        results.map((el) => el[propertyToCompare]),
+        `
+        Expected results not returned from query: ${queryString} ${JSON.stringify(queryArgs)}, 
+      `,
+      );
+    });
+  };
```
kraenhansen, line 134: Probably a matter of taste, but I'd suggest moving these out above of the `describe("Queries", () => {` and make then regular functions, instead of arrow functions (since they're not capturing a `this` of a parent scope anyway).

I also have a personal preference towards a for-of loop instead of calling `forEach` - we can talk more about this face-2-face, but basically, you're relying on the runtime implementation to call the callback synchronously - which it does on all runtimes I've seen, but then again ...

integration-tests/tests/src/tests/queries.ts
```diff
@@ -15,129 +15,1037 @@
...
+          t: new Date("2001-05-11T12:45:05"),
+        });
+        this.realm.create<IPrimitive>(PrimitiveSchema.name, {
+          s: "Here is a Unicorn 🦄 today",
+          b: false,
+          i: 44,
+          f: 1.41,
+          d: 4.67,
+          t: new Date("2004-02-26T10:15:02"),
+        });
```
kraenhansen, line 858: Not a part of your work, but this looks a bit verbose and would benefit from something like: https://github.com/realm/realm-js/issues/2009.

kraenhansen (approved):
Great job refactoring this! Just to clarify, this is no longer "passing schema configuration as an array" right? (I couldn't find anywhere it did).

integration-tests/tests/src/tests/queries.ts
gagik: That makes sense, will consider this thank you. With regards to arrow function vs regular functions, do you see regular functions as a default to use unless you'd benefit from not introducing another this into the scope? I think I have been looking at it from the opposite perspective (only use regular function if you need a local `this` scope).

gagik:
> passing schema configuration as an array

Yes! I completely got rid of that so should be good to go on bindgen 🤞

integration-tests/tests/src/tests/queries.ts
gagik: I am not sure how much less verbose can be made beyond maybe putting them into arrays and doing a forEach / arrayloop, which would hurt readability a bit

PR #5324, opened 2023-01-24
Jg/user master

JacobOscarGunnarsson:
some changes will be done here, right now I'm merging app-tests to master that uses the integration test-apps that exist in bindgen, I'll have to adopt those changes here as well


integration-tests/tests/package.json
```diff
@@ -39,7 +39,8 @@
   "dependencies": {
```
JacobOscarGunnarsson, line 39: the package-lock.json diff above seems quite big, could it be the case that I've done something weird here?

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+import {
+  randomNonVerifiableEmail,
+  randomPendingVerificationEmail,
+  randomVerifiableEmail,
+  uuid,
+} from "../../utils/generators";
+import { KJUR } from "jsrsasign";
+import { sleep } from "../../utils/sleep";
+
+function assertIsUser(user: Realm.User) {
```
kraenhansen, line 29: A very small thing: Technically it does an assertion, but you might want to lend the wording from `chai` here.

```suggestion
function expectIsUser(user: Realm.User) {
```

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+  const users = app.allUsers;
+  Object.keys(app.allUsers).forEach(async (id) => {
+    await app.removeUser(users[id]);
+  });
+}
+
+describe.skipIf(environment.missingServer, "User", () => {
+  describe("email password", () => {
+    importAppBefore("with-email-password");
+    it("login without username throws", async function (this: Mocha.Context & AppContext & RealmContext) {
```
kraenhansen, line 84: While this is correct, both [`AppContext` and `RealmContext` unions with `Mocha.Context` underneath](https://github.com/realm/realm-js/blob/bindgen/integration-tests/tests/src/typings.d.ts#L68-L73), so there's no need to include that here.

```suggestion
    it("login without username throws", async function (this: AppContext & RealmContext) {
```

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+    it("login with non existing user throws", async function (this: Mocha.Context & AppContext & RealmContext) {
+      const credentials = Realm.Credentials.emailPassword("foo", "pass");
+      return this.app
+        .logIn(credentials)
+        .then((user) => {
+          throw new Error("Login should have failed");
+        })
+        .catch((err) => {
+          expect(err.message).equals("invalid username/password");
+        });
```
kraenhansen, line 110: This (and a lot of the other tests) could fairly easily be refactored (now or later) to use "chai-as-promised":

```suggestion
      const credentials = Realm.Credentials.emailPassword("foo", "pass");
      await expect(this.app.logIn(credentials)).should.be.rejectedWith("invalid username/password");
```

kraenhansen (approved):
As a comment for later refactoring, it makes sense to merge these with the credentials tests we already have: https://github.com/realm/realm-js/tree/bindgen/integration-tests/tests/src/tests/credentials

integration-tests/tests/src/utils/generators.ts
```diff
@@ -15,10 +15,34 @@
...
+
 export function generatePartition() {
   return "xxxxxxxxxxxx4xxxyxxxxxxxxxxxxxxx".replace(/[xy]/g, function (c) {
     const r = (Math.random() * 16) | 0,
       v = c == "x" ? r : (r & 0x3) | 0x8;
     return v.toString(16);
   });
 }
+
+export function uuid() {
```
kraenhansen, line 27: Could this be merged with the `generatePartition` above?

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+    importAppBefore("with-db");
+    it("can perform operations on a collection via the client", async function (this: Mocha.Context &
+      AppContext &
+      RealmContext) {
+      const credentials = Realm.Credentials.anonymous();
+      const user = await this.app.logIn(credentials);
+
+      const mongo = user.mongoClient("mongodb");
+      //@ts-expect-error TYPEBUG: serviceName is a missing property on MongoDB interface
+      expect(mongo.serviceName).equals("mongodb");
```
kraenhansen, line 522: This should be fixed with the new SDK: https://github.com/realm/realm-js/blob/bindgen/packages/realm/src/app-services/MongoClient.ts#L166-L169
🤞

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+      assertIsSameUser(user, this.app.currentUser);
+      await user.logOut();
+    });
+  });
+
+  describe("properties and methods", () => {
+    describe("with anonymous", () => {
+      importAppBefore("with-db");
+      it("login and logout works", async function (this: Mocha.Context & AppContext & RealmContext) {
+        await removeExistingUsers(this.app);
```
takameyer, line 229: This seems like something that belongs in a `beforeEach` statement.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+        // Is now logged out.
+        expect(this.app.currentUser).to.be.null;
+      });
+
+      it("can fetch customData", async function (this: Mocha.Context & AppContext & RealmContext) {
+        await removeExistingUsers(this.app);
+        const credentials = Realm.Credentials.anonymous();
+        const user = await this.app.logIn(credentials);
+        const customData = user.customData;
+        // TODO: Enable custom user data in the app to test this e2e
```
takameyer, line 245: We should make a task for this.  Otherwise it will never happen

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+        expect(user2.isLoggedIn).to.be.false;
+        expect(user1.isLoggedIn).to.be.true;
+        expect(Object.keys(all).length).equals(2, "still holds references to both users");
+
+        await user1.logOut();
+        all = this.app.allUsers;
+        expect(user1.isLoggedIn).to.be.false;
+        expect(user2.isLoggedIn).to.be.false;
+        expect(Object.keys(all).length).equals(2, "still holds references to both users"); // FIXME: is this actually expected?
+      });
```
takameyer, line 341: new line between `it` statements 😉

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+        const user2 = await this.app.logIn(Realm.Credentials.emailPassword(validEmail, validPassword)).catch((err) => {
+          expect(err.message).equals("invalid username/password");
+          expect(err.code).equals(50);
+          didFail = true;
+        });
+        expect(user2).to.be.undefined;
+        expect(didFail).to.be.true;
+      });
+    });
+  });
```
takameyer, line 400: new line between describe statements

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+  describe("api-key auth", () => {
+    importAppBefore("with-api-key");
+    it("can create valid key", async function (this: Mocha.Context & AppContext & RealmContext) {
+      const validEmail = randomVerifiableEmail();
+      const validPassword = "test1234567890";
+      await this.app.emailPasswordAuth.registerUser({ email: validEmail, password: validPassword });
+      const user = await this.app.logIn(Realm.Credentials.emailPassword(validEmail, validPassword));
+      expect(user.apiKeys instanceof Realm.Auth.ApiKeyAuth).to.be.true;
+
+      // TODO: Enable when fixed: Disabling this test since the CI stitch integration returns cryptic error.
```
takameyer, line 446: We should see what happens when this is enabled

integration-tests/tests/src/utils/generators.ts
takameyer: Or replaced with `Realm.BSON.uuid()`?

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -0,0 +1,620 @@
...
+async function removeExistingUsers(app: Realm.App) {
+  const users = app.allUsers;
+  Object.keys(app.allUsers).forEach(async (id) => {
+    await app.removeUser(users[id]);
+  });
+}
+
+describe.skipIf(environment.missingServer, "User", () => {
+  describe("email password", () => {
+    importAppBefore("with-email-password");
```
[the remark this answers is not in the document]
JacobOscarGunnarsson: yeah I'm going back and forth with this, sometimes tests expect multiple authentication providers, in that case it becomes difficult to make the apps "modular", but on the other hand I also get concerned about making a gigantic app that would couple all tests together

PR #5325, opened 2023-01-24
Test Migration: Results Tests

gagik:
Results test migration

### DIFF
--- a/integration-tests/tests/src/tests/results.ts
+++ b/integration-tests/tests/src/tests/results.ts
@@ -1,0 +1,1072 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { expect } from "chai";
+import Realm, { BSON } from "realm";
+import { openRealmBeforeEach } from "../hooks";
+const { Decimal128, ObjectId, UUID } = Realm.BSON;
+
+const PersonObjectSchema = {
+  name: "PersonObject",
+  properties: {
+    name: "string",
+    age: "double",
+    married: { type: "bool", default: false },
+    children: { type: "list", objectType: "PersonObject" },
+    parents: { type: "linkingObjects", objectType: "PersonObject", property: "children" },
+  },
+};
+
+const DateObjectSchema = {
+  name: "Date",
+  properties: {
+    currentDate: "date",
+    nullDate: "date?",
+  },
+};
+
+class PersonObject extends Realm.Object {
+  name!: string;
+  age!: Realm.Types.Double;
+  married!: boolean;
+  children!: Realm.List<PersonObject>;
+  parents!: Realm.List<PersonObject>;
+}
+
+const TestObjectSchema = {
+  name: "TestObject",
+  properties: {
+    doubleCol: "double",
+  },
+};
+
+interface TestObject {
+  doubleCol: Realm.Types.Double;
+}
+
+const IntPrimarySchema = {
+  name: "IntPrimaryObject",
+  primaryKey: "primaryCol",
+  properties: {
+    primaryCol: "int",
+    valueCol: "string",
+  },
+};
+
+interface IntPrimaryObject {
+  primaryCol: Realm.Types.Int;
+  valueCol: string;
+}
+
+const DefaultValuesSchema = {
+  name: "DefaultValuesObject",
+  properties: {
+    boolCol: { type: "bool", default: true },
+    intCol: { type: "int", default: -1 },
+    floatCol: { type: "float", default: -1.1 },
+    doubleCol: { type: "double", default: -1.11 },
+    stringCol: { type: "string", default: "defaultString" },
+    dateCol: { type: "date", default: new Date(1.111) },
+    dataCol: { type: "data", default: new ArrayBuffer(1) },
+    objectCol: { type: "TestObject", default: { doubleCol: 1 } },
+    nullObjectCol: { type: "TestObject", default: null },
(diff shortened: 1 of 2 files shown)


integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1067 @@
...
+        this.realm.create<TestObject>("TestObject", { doubleCol: 1 });
+        this.realm.create<TestObject>("TestObject", { doubleCol: 2 });
+      });
+
+      const results = this.realm.objects<TestObject>("TestObject");
+      expect(results[0].doubleCol).equals(1.0);
+      expect(results[1].doubleCol).equals(2.0);
+      expect(results[2]).equals(undefined);
+      expect(results[-1]).equals(undefined);
+      // TODO: expect(Object.getPrototypeOf(results[0])).equals(TestObject.prototype);
```
gagik, line 299: I am not sure if this is worthwhile but previous test used to do some prototype checks, is this something that should be migrated over?

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1067 @@
...
+        boolCol: false,
+        stringCol: "hello",
+        intCol: 10,
+        floatCol: 10.0,
+        doubleCol: 10.0,
+        dateCol: new Date(10),
+        dataCol: new Uint8Array([0xd8, 0x21, 0xd6]),
+        decimal128Col: Decimal128.fromString("3e-10"),
+        objectIdCol: new ObjectId(`0000002a9a7969d24bea4cf1`),
+        uuidCol: new UUID(`f7adc509-49be-466f-b3d5-449696d9a601`),
```
gagik, line 939: I refactored the old test to make it more clear and also added some new Realm types to it, might be worth looking at this more.

gagik (commented):
I largely left the tests untouched in terms of their structure / behavior, mainly trying to group them and avoid unnecessary realm.open calls, instead opting for hooks.

Only 1 test (commented below) I changed significantly (essentially merging testResultsUpdateDataTypes and testResultUpdateDateColumn from legacy).

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1067 @@
...
+        objects.indexOf(object4);
+      }).throws("Realm object is from another Realm");
+    });
+
+    it("implements addListener", function (this: RealmContext) {
+      //@ts-expect-error Navigator could be defined in some environments.
+      if (typeof navigator !== "undefined" && /Chrome/.test(navigator.userAgent)) {
+        // eslint-disable-line no-undef
+        // FIXME: async callbacks do not work correctly in Chrome debugging mode
+        return Promise.resolve();
```
gagik, line 217: Not sure if something needs to be done about this

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1067 @@
...
+    dateCol: "date?",
+    dataCol: "data?",
+    decimal128Col: "decimal128?",
+    objectIdCol: "objectId?",
+    uuidCol: "uuid?",
+  },
+};
+
+describe("Results", () => {
+  beforeEach(Realm.clearTestState);
```
kraenhansen, line 140: This might not be needed, since it's default behaviour of `openRealmBeforeEach` to also clear test state after each test.

```suggestion
```

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1067 @@
...
+      }).throws("Realm object is from another Realm");
+    });
+
+    it("implements addListener", function (this: RealmContext) {
+      //@ts-expect-error Navigator could be defined in some environments.
+      if (typeof navigator !== "undefined" && /Chrome/.test(navigator.userAgent)) {
+        // eslint-disable-line no-undef
+        // FIXME: async callbacks do not work correctly in Chrome debugging mode
+        return Promise.resolve();
+      }
```
kraenhansen, line 218: Since we don't support nor run these tests in chrome debugging mode anymore, this could probably just be deleted.
```suggestion
```

integration-tests/tests/src/tests/results.ts
kraenhansen: Sure - but it will fail on the `bindgen` branch (since we're injecting a class between the user provided constructor and the user-facing object). I think simply doing an instanceOf check would suffice:

```suggestion
      expect(results[0]).instanceOf(TestObject);
```

kraenhansen (approved):

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1060 @@
...
+      const realm = new Realm({ schema: [PersonObjectSchema, DefaultValuesSchema, TestObjectSchema] });
+
+      realm.write(function () {
+        realm.create("PersonObject", { name: "Ari", age: 10 });
+        realm.create("PersonObject", { name: "Tim", age: 11 });
+        realm.create("PersonObject", { name: "Bjarne", age: 12 });
+        realm.create("PersonObject", { name: "Alex", age: 12, married: true });
+      });
+
+      expect(realm.objects("PersonObject").filtered("truepredicate").length).equals(4);
```
takameyer, line 471: This is probably also covered in the query tests.  Might be able to refactor this out.

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1060 @@
...
+      const oids: BSON.ObjectId[] = [];
+      const uuids: BSON.UUID[] = [];
+      for (let i = 0; i < 3; i++) {
+        decimals.push(Decimal128.fromString(`${i}`));
+        oids.push(new ObjectId(`0000002a9a7969d24bea4cf${i}`));
+        uuids.push(new UUID(`f7adc509-49be-466f-b3d5-449696d9a60${i}`));
+      }
+
+      realm.write(function () {
+        realm.create("BasicTypesObject", [
```
takameyer, line 622: This will probably need to be refactored as well.  I think we are deprecating creating objects with an array.  It's also not very readable 😅

integration-tests/tests/src/tests/results.ts
```diff
@@ -0,0 +1,1061 @@
...
+        results.max("foo");
+      }).throws("Property 'foo' does not exist on object 'NullableBasicTypesObject'");
+      expect(() => {
+        results.sum("foo");
+      }).throws("Property 'foo' does not exist on object 'NullableBasicTypesObject'");
+      expect(() => {
+        results.avg("foo");
+      }).throws("Property 'foo' does not exist on object 'NullableBasicTypesObject'");
+    });
+  });
```
takameyer, line 871: Should add a new line between `describe` statements

takameyer (approved):
Just a few suggestions for future refactor.  Otherwise, LGTM 💯

PR #5329, opened 2023-01-26
New Cluster Deployment Strategy

takameyer:
## What, How & Why?
We suspect that the amount of time a cluster deployment needs to settle is greater than the time it takes to setup an integration tests.  This change prepares the cluster deployments ahead of time for each integration tests during the build phase.  

If this works, it may not be the solution we want to merge, but it could be an insight into how reliable clusters are after deployment.  If we do merge this, we should refactor to share the matrix between the three jobs (cluster-prep, integration-tests, cluster-cleanup) in order to keep the workflow maintainable.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -182,10 +182,48 @@ jobs:
           path: |
             dependencies.list
             ${{ matrix.variant.artifact-path }}
+  cluster-prep:
+    name: Prepare Clusters for Integration Tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        variant:
+          - { key: linux-node }
+          - { key: linux-electron-main }
+          - { key: linux-electron-render }
+          #- { key: windows-node }
+          #- { key: windows-electron-main }
+          #- { key: windows-electron-render }
+          - { key: darwin-electron-main }
+          - { key: darwin-electron-render }
+          - { key: darwin-node }
+          - { key: android-rn }
+          - { key: ios-rn }
+          #- { key: ios-rn-catalyst }
+    steps:
+    - name: Deploy Cluster for ${{ matrix.variant.key }}
+      uses: realm/ci-actions/mdb-realm/deployApps@fac1d6958f03d71de743305ce3ab27594efbe7b7
+      id: deploy-mdb-apps
+      with:
+        realmUrl: ${{ secrets.REALM_QA_BASE_URL }}
+        projectId: ${{ secrets.ATLAS_QA_PROJECT_ID }}
+        atlasUrl: ${{ secrets.ATLAS_QA_BASE_URL}}
+        apiKey: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}
+        privateApiKey: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
+        differentiator: ${{ matrix.variant.key }}-${{ github.run_id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -182,10 +182,48 @@ jobs:
...
+      with:
+        realmUrl: ${{ secrets.REALM_QA_BASE_URL }}
+        projectId: ${{ secrets.ATLAS_QA_PROJECT_ID }}
+        atlasUrl: ${{ secrets.ATLAS_QA_BASE_URL}}
+        apiKey: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}
+        privateApiKey: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
+        differentiator: ${{ matrix.variant.key }}-${{ github.run_id }}
+    - name: Save Cluster Name to File
+      run: |
+          echo ${{ steps.deploy-mdb-apps.outputs.clusterName }} > cluster-name.txt
```
kneth, line 217: Do you know where `cluster-name.txt` is located on the file system?

takameyer: I forgot the `txt` in the `cat` command below.  That will probably fix it.
I'm guessing this just lands in the `github.workspace`.  Should be pretty stable, but we will see.

.github/workflows/pr-realm-js.yml
```diff
@@ -182,10 +182,48 @@ jobs:
...
+    - name: Deploy Cluster for ${{ matrix.variant.key }}
+      uses: realm/ci-actions/mdb-realm/deployApps@fac1d6958f03d71de743305ce3ab27594efbe7b7
+      id: deploy-mdb-apps
+      with:
+        realmUrl: ${{ secrets.REALM_QA_BASE_URL }}
+        projectId: ${{ secrets.ATLAS_QA_PROJECT_ID }}
+        atlasUrl: ${{ secrets.ATLAS_QA_BASE_URL}}
+        apiKey: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}
+        privateApiKey: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
+        differentiator: ${{ matrix.variant.key }}-${{ github.run_id }}
```
kraenhansen, line 214: Should use `clusterName` instead (a hash of run id, attempt id and the key.

.github/workflows/pr-realm-js.yml
```diff
@@ -385,12 +418,34 @@ jobs:
...
+    steps:
+    - uses: realm/ci-actions/mdb-realm/cleanup@800a0234bcbe9073e7a07a66c864d54db3688e56
+      if: ${{always()}}
+      with:
+        realmUrl: ${{ secrets.REALM_QA_BASE_URL }}
+        atlasUrl: ${{ secrets.ATLAS_QA_BASE_URL}}
+        projectId: ${{ secrets.ATLAS_QA_PROJECT_ID }}
+        apiKey: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}
+        privateApiKey: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
+        differentiator: ${{ matrix.variant.key }}-${{ github.run_id }}
```
kraenhansen, line 451: This could be a step in the "integration-test" job.

PR #5330, opened 2023-01-26
[bindgen] iOS support

kraenhansen:
## What, How & Why?

This adds iOS support to the new SDK.

I had to move podspec and the source-code referenced by it into the SDK's package, to enabled RN auto-linking.


integration-tests/environments/react-native/Gemfile.lock
```diff
@@ -91,10 +91,10 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  cocoapods (~> 1.11, >= 1.11.2)
+  cocoapods (~> 1.11, >= 1.11.3)
 
 RUBY VERSION
-   ruby 2.7.5p203
+   ruby 2.7.6p219
```
kraenhansen, line 97: This got updated as I ran `bundle install` in the RN test app - I verified the change that triggered this, was already committed separately.

packages/realm/package.json
```diff
@@ -2,6 +2,26 @@
...
+    "asyncstorage",
+    "rocksdb",
+    "leveldb",
+    "realm"
+  ],
+  "author": {
+    "name": "Realm",
+    "email": "help@realm.io",
+    "url": "https://realm.io"
+  },
```
kraenhansen, line 24: These are simply copied from the root package and had to be added because the Podspec reference them.

integration-tests/environments/react-native/Gemfile.lock
takameyer: Probably could just add this to the gitignore.  Otherwise I'm gonna have to keep downloading new version of ruby when I want to run the integration tests

packages/realm/react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -0,0 +1,184 @@
...
+  [handlers removeObject:handler];
+}
+
+RCT_REMAP_METHOD(emit, emitEvent
+                 : (NSString *)eventName withObject
+                 : (id)object) {
+  for (RealmReactEventHandler handler in [_eventHandlers[eventName] copy]) {
+    handler(object);
+  }
+}
```
kraenhansen, line 103: I actually don't think this is used / needed anymore 🤔

integration-tests/environments/react-native/ios/RealmReactNativeTests.xcodeproj/project.pbxproj
```diff
@@ -270,8 +270,8 @@
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "set -e\nPLATFORM_NAME=ios\nWITH_ENVIRONMENT=\"../../../../node_modules/react-native/scripts/xcode/with-environment.sh\"\nREACT_NATIVE_XCODE=\"../../../../node_modules/react-native/scripts/react-native-xcode.sh\"\n\n/bin/sh -c \"$WITH_ENVIRONMENT $REACT_NATIVE_XCODE\"\n";
+			shellPath = "/bin/sh -l";
+			shellScript = "set -e\nPLATFORM_NAME=ios\nWITH_ENVIRONMENT=\"../../../../node_modules/react-native/scripts/xcode/with-environment.sh\"\nREACT_NATIVE_XCODE=\"../../../../node_modules/react-native/scripts/react-native-xcode.sh\"\nexport PROJECT_ROOT=\"$PROJECT_DIR/..\"\n/bin/sh -c \"$WITH_ENVIRONMENT $REACT_NATIVE_XCODE\"\n";
```
takameyer, line 274: Was this intentional? Just worried about the possibility of this being unintentionally wiped after a RN upgrade

packages/realm/RealmJS.podspec
```diff
@@ -0,0 +1,70 @@
+# coding: utf-8
+require 'json'
```
takameyer, line 2: Should we remove this file from the root project?  Or will that be done later?

packages/realm/react-native/ios/RealmReact/RealmReact.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef void (^RealmReactEventHandler)(id message);
+
+@interface RealmReact : NSObject
+
+- (void)addListenerForEvent:(NSString *)eventName handler:(RealmReactEventHandler)handler;
```
takameyer, line 29: I think this is legacy code from the old react native test infrastructure.  I bet we can safely remove these.

react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -153,6 +153,7 @@ - (void)setBridge:(RCTBridge *)bridge {
 
           auto &rt = *static_cast<facebook::jsi::Runtime *>(bridge.runtime);
           auto exports = jsi::Object(rt);
+          rt.global().setProperty(rt, "__RealmFuncs", exports);
```
takameyer: This file is probably not being invoked anymore, or?  I assume the `react-native` directory will go away soon.

takameyer (approved):
Just a few questions.  Other than that LGTM and good job 🎉

integration-tests/environments/react-native/ios/RealmReactNativeTests.xcodeproj/project.pbxproj
kraenhansen: Yeah. I've contributed a patch upstream to change the default to be more friendly towards mono-repos: https://github.com/facebook/react-native/pull/35970

kraenhansen: But with [the changes I did to `.xcode.env`](https://github.com/realm/realm-js/pull/5330/files#diff-0369cbaa30691bc3629a31a58b70cf9b8a6a909cb2f486446315eb93d553b055R11-R12) I can actually revert the update of the shellPath.

takameyer: Cool!  Hopefully it gets approved :)

packages/realm/RealmJS.podspec
kraenhansen: Yeah. I don't know - we could do git moves instead of straight up copies. Perhaps that would ease merging in the future.

packages/realm/react-native/ios/RealmReact/RealmReact.mm
takameyer: Just saw this comment.  They can be removed.  These were used by the legacy react native test project.

kraenhansen: I removed it ..

react-native/ios/RealmReact/RealmReact.mm
kraenhansen: Yeah. I moved it and actually that line was not needed anyway 🙈

packages/realm/RealmJS.podspec
kraenhansen: I git moved it instead.

packages/realm/react-native/ios/RealmReact/RealmReact.mm
```diff
@@ -42,7 +42,6 @@ @interface RealmReact () <RCTBridgeModule, RCTInvalidating>
 @end
 
 @implementation RealmReact {
```
takameyer, line 44: I think the removed method declarations are still existing in the `RealmReact.h`

kraenhansen: Oh yeah - thanks!

takameyer (approved):

PR #5332, opened 2023-01-26
Cleanup Clusters and Apps every night

takameyer:
## What, How & Why?
Currently this was only a manual workflow.  We should just run this every night.  Alternatively, we should find out why we end up having so many apps deployed that never get cleaned up after their tests. (Probably due to test failures)

I also updated the cleanup task.  This was very outdated and slow, since it was using the realm-cli to delete apps rather than fetch.  This should now be much faster.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/cleanup-clusters.yml
+++ b/.github/workflows/cleanup-clusters.yml
@@ -1,6 +1,8 @@
-name: Wipe all clusters and apps on dev
+name: Wipe all clusters and apps
 
 on:
+  schedule:
+    - cron: "0 6 * * *"
   workflow_dispatch:
 jobs:
   main:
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-clusters.yml
```diff
@@ -1,21 +1,23 @@
-name: Wipe all clusters and apps on dev
+name: Wipe all clusters and apps
 
 on:
+  schedule:
+    - cron: "0 6 * * *"
```
kneth, line 5: I think it is unlike anyone from the team is working at that point but maybe we should do the clean up a bit earlier to be a bit more safe 😴

kneth (approved):

.github/workflows/cleanup-clusters.yml
takameyer: I have updated it for 2am 😪

PR #5337, opened 2023-01-27
use correct baseurl for tests

JacobOscarGunnarsson:
test `throws on non existing app` failed because I used a hardcoded url in the config which worked locally but not in the CI

### DIFF
--- a/integration-tests/tests/src/tests/sync/app.ts
+++ b/integration-tests/tests/src/tests/sync/app.ts
@@ -20,6 +20,7 @@ import { ObjectId } from "bson";
 import { expect } from "chai";
 import { importAppBefore } from "../../hooks";
 import { generatePartition } from "../../utils/generators";
+import { getUrls } from "../../utils/import-app";
 
 const TestObjectSchema: Realm.ObjectSchema = {
   name: "TestObject",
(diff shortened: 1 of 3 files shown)


integration-tests/tests/src/utils/import-app.ts
```diff
@@ -25,7 +25,7 @@ export type ErrorResponse = { message: string; appId: never };
 export type ImportResponse = { appId: string; message: never };
 export type Response = ImportResponse | ErrorResponse;
 
-function getUrls() {
+export function getUrls() {
```
takameyer, line 28: We should probably move this into its own file.  Now that `getUrls` is an export, it really doesn't have much to do with `import-app`.  Alternatively, we could consider setting defaults on the environment and doing away with this function altogether.

@JacobOscarGunnarsson This is just a refactor consideration, you can still merge this without said change, but maybe write a todo

takameyer (approved):
Just a couple of considerations.  Otherwise, LGTM 💪🏼

integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -73,7 +74,8 @@ describe("App", () => {
     afterEach(async () => {
       Realm.clearTestState();
     });
-    const conf = { id: "smurf", baseUrl: "http://localhost:9090" };
+    const urls = getUrls();
+    const conf = { id: "smurf", baseUrl: urls.baseUrl };
```
kraenhansen, line 78:
```suggestion
    const smurfAppConfig = { id: "smurf", baseUrl: urls.baseUrl };
```

```suggestion
    const missingAppConfig = { id: "smurf", baseUrl: urls.baseUrl };
```

kraenhansen (approved):

PR #5338, opened 2023-01-27
One Big Cluster Deployment Strategy

takameyer:
## What, How & Why?
So many clusters can cause us to reach the limit quite fast.  This strategy deploys one dedicated cluster that all tests can talk to.  Will be interesting to see if this makes any difference in test stability.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen (approved):
Looks about right 🤞

.github/workflows/pr-realm-js.yml
```diff
@@ -202,38 +219,37 @@ jobs:
...
-          differentiator: ${{ steps.differentiator.outputs.name }}
+          clusterSize: "M10"
 
-      - name: Checkout code
-        uses: actions/checkout@v2
-        with:
-          submodules: "recursive"
+      # We set it from the output from the above deployment, in case we rerun this test, which will create a new cluster name
+      - name: Set Cluster Name to Environment
+        run: echo "CLUSTER_NAME=${{ steps.deploy.outputs.clusterName }}" >> $GITHUB_ENV
```
kraenhansen, line 252: What will set this in case of a regular (non-retry) run?

takameyer: The above deploy step.  If the cluster has already been deployed, it will just set the output to that name.

kraenhansen: Ahh okay. What's the need to store it in an env? Couldn't the test step just use the output directly?

takameyer: No need 🤷🏼 Just realized it was broken for the retry case and came up with a solution with the current setup.  We can use it directly

kneth (approved):
Looks good to me

PR #5340, opened 2023-01-27
Create Default Provider and Hooks for `@realm/react`

### LINKED ISSUE #5292: Export a default Realm context from `@realm/react`
### Problem

It's annoying for users to have to create a Realm context (calling `createRealmContext`) before accessing the `useRealm`, `useQuery` and `useObject` hooks:

```tsx
import { createRealmContext } from "@realm/react";
const { RealmProvider, useRealm } = createRealmContext();
// use the useRealm hook
```

### Solution

Instead we should create a default Realm context and export its members as named exports, making the average use of the library much simpler:

```tsx
import { RealmProvider, useRealm } from "@realm/react";
// use the useRealm hook
```

We should make sure to document the advanced use-case of working with multiple Realm's (calling `createRealmContext`) too.

### Alternatives

_No response_

I would like to have it but have a workaround

Atlas Device Sync

takameyer:
## What, How & Why?
* Creates a default context and exports the results.
* Provide the tsdoc directly on the function exports.
* This makes the single realm use case much simpler to start with.  For multi realm usage, the `createRealmContext` is still available.

This closes #5292

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/index.tsx
+++ b/packages/realm-react/src/index.tsx
@@ -25,7 +25,7 @@ import { createRealmProvider } from "./RealmProvider";
 
 type RealmContext = {
   /**
-   * Returns a Context Provider component that is required to wrap any component using
+   * The Provider component that is required to wrap any component using
    * the Realm hooks.
    *
    * @example
@@ -37,7 +37,7 @@ type RealmContext = {
    *   };
    *
    *   return (
-   *     <RealmProvider path={"data.realm"} sync={syncConfig}>
+   *     <RealmProvider schema={[Task, User]} path={"data.realm"} sync={syncConfig}>
    *       <App/>
    *     </RealmProvider>
    *   )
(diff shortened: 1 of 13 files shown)


packages/realm-react/src/index.tsx
```diff
@@ -24,79 +24,9 @@ import { createUseObject } from "./useObject";
...
-  /**
-   * Returns the instance of the {@link Realm} configured by `createRealmContext`
-   *
-   * @example
-   * ```
-   * const realm = useRealm();
-   * ```
-   *
-   * @returns a realm instance
-   */
```
kraenhansen, line 99: I don't think this should be removed. It should still be there for people that need multiple nested Realm providers.

packages/realm-react/src/index.tsx
```diff
@@ -141,6 +71,87 @@ export const createRealmContext: (realmConfig?: Realm.Configuration) => RealmCon
...
+ * }
+ * ```
+ * @param props - The {@link Realm.Configuration} for this Realm defaults to
+ * the config passed to `createRealmProvider`, but individual config keys can
+ * be overridden when creating a `<RealmProvider>` by passing them as props.
+ * For example, to override the `path` config value, use a prop named `path`
+ * e.g., `path="newPath.realm"`
+ * an attribute of the same key.
+ */
+const RealmProvider = defaultContext.RealmProvider;
```
kraenhansen, line 102: Might as well `export` right here?
```suggestion
export const RealmProvider = defaultContext.RealmProvider;
```

kraenhansen (approved):
I only have a few suggestions.

packages/realm-react/src/index.tsx
takameyer: I'm curious how we can sanely remove this duplication and also have better documentation.

packages/realm-react/src/index.tsx
takameyer: Sure, no reason not to.

PR #5347, opened 2023-01-30
Test Migration: Notification Tests

gagik:
This is moving `async-tests.js` which is actually notifications/listeners tests.

### DIFF
--- a/integration-tests/tests/src/tests/notifications.ts
+++ b/integration-tests/tests/src/tests/notifications.ts
@@ -48,15 +48,12 @@ describe("Notifications", () => {
   describe("Realm notifications", () => {
     openRealmBeforeEach({ schema: [TestObject] });
 
+    /**
+     * TODO: only "change" event is being tested, not "schema".
+     */
     it("should fire on change", function (this: RealmContext, done) {
-      this.realm.addListener("change", (sender, event) => {
-        expect(sender).deep.equals(this.realm);
+      this.realm.addListener("change", (_, event) => {
         expect(event).equals("change");
-
-        const objects = this.realm.objects(TestObject);
-
-        expect(objects.length).equals(1);
-        expect(objects[0].doubleCol).equals(42);
         done();
       });
 
@@ -66,19 +63,10 @@ describe("Notifications", () => {
     });
 
     it("should fire on multiple changes", function (this: RealmContext, done) {
-      this.realm.addListener("change", (sender, event) => {
-        expect(sender).deep.equals(this.realm);
+      this.realm.addListener("change", (_, event) => {
         expect(event).equals("change");
-
-        const objects = this.realm.objects(TestObject);
         runCount++;
-        if (runCount === 1) {
-          expect(objects.length).equals(1);
-          expect(objects[0].doubleCol).equals(1);
-        } else if (runCount == 2) {
-          expect(objects.length).equals(3);
-          expect(objects[1].doubleCol).equals(2);
-          expect(objects[2].doubleCol).equals(3);
+        if (runCount >= 2) {
           done();
         }
       });
@@ -177,7 +165,6 @@ describe("Notifications", () => {
       expect(changes.insertions).deep.equals(changesOnRun[runCount - 1].insertions);
       expect(changes.oldModifications).deep.equals(changesOnRun[runCount - 1].oldModifications);
       expect(changes.newModifications).deep.equals(changesOnRun[runCount - 1].newModifications);
-      // expect(changes.modifications).deep.equals(changesOnRun[runCount - 1].modifications);
       expect(changes.deletions).deep.equals(changesOnRun[runCount - 1].deletions);
 
       if (runCount >= changesOnRun.length) {
@@ -217,7 +204,8 @@ describe("Notifications", () => {
         expectCollectionChangesOnEveryRun(done, [
           { insertions: [], newModifications: [], oldModifications: [], deletions: [] },
           { insertions: [0, 1, 2], newModifications: [], oldModifications: [], deletions: [] },
-          { insertions: [], newModifications: [0, 2], oldModifications: [0, 2], deletions: [] },
+          { insertions: [], newModifications: [0], oldModifications: [0], deletions: [] },
+          { insertions: [], newModifications: [0, 1], oldModifications: [0, 2], deletions: [1] },
         ]),
       );
 
@@ -228,8 +216,13 @@ describe("Notifications", () => {
       });
 
       this.realm.write(() => {
+        testObjects[0].doubleCol = 15;
+      });
+
+      this.realm.write(() => {
         testObjects[0].doubleCol = 10;
         testObjects[2].doubleCol = 30;
+        this.realm.delete(testObjects[1]);
       });
     });
 
--- a/integration-tests/tests/src/tests/notifications.ts
+++ b/integration-tests/tests/src/tests/notifications.ts
@@ -81,15 +81,23 @@ describe("Notifications", () => {
       });
     });
 
-    it("implements removeListener", function (this: RealmContext) {
+    it("implements removeListener", function (this: RealmContext, done) {
       const handler = () => {
-        const objects = this.realm.objects(TestObject);
         runCount++;
         if (runCount === 1) {
-          expect(objects.length).equals(1);
-          expect(objects[0].doubleCol).equals(1);
+          this.realm.removeListener("change", handler);
+
+          // This listener should be the only one that is fired.
+          this.realm.addListener("change", () => done());
+
+          this.realm.write(() => {
+            this.realm.create("TestObject", { doubleCol: 2 });
+          });
         } else {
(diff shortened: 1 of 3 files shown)


integration-tests/tests/src/tests/results.ts
```diff
@@ -171,48 +171,6 @@ describe("Results", () => {
       }).throws("Realm object is from another Realm");
     });
 
-    it("implements addListener", function (this: RealmContext) {
```
gagik, line 174: I figured this is worth removing now... Much cleaner way to deal with this with `done` too.

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+    onEnd: () => void | Mocha.Done,
+    changesOnRun: Realm.CollectionChangeSet[],
+  ) => {
+    return (_: Realm.Collection<Realm.Object>, changes: Realm.CollectionChangeSet) => {
+      runCount++;
+
+      expect(changes.insertions).deep.equals(changesOnRun[runCount - 1].insertions);
+      expect(changes.oldModifications).deep.equals(changesOnRun[runCount - 1].oldModifications);
+      expect(changes.newModifications).deep.equals(changesOnRun[runCount - 1].newModifications);
+      // expect(changes.modifications).deep.equals(changesOnRun[runCount - 1].modifications);
```
gagik, line 180: Should I run an assert on modifications? I could also compare it with the new/oldModifications, whichever was the one this actually represents.

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+    },
+  };
+}
+
+describe("Notifications", () => {
+  let runCount = 0;
+
+  beforeEach(() => (runCount = 0));
+
+  describe("Realm notifications", () => {
```
gagik, line 48: This might need some more tests, i.e. different list functions etc., I actually added a couple so every category covered the same amount of functions since it was simple to do so. I can add a TODO. entry if necessary.

kraenhansen (approved):
Great job! I have only minor comments.

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+    it("should fire on change", function (this: RealmContext, done) {
+      this.realm.addListener("change", (sender, event) => {
+        expect(sender).deep.equals(this.realm);
+        expect(event).equals("change");
+
+        const objects = this.realm.objects(TestObject);
+
+        expect(objects.length).equals(1);
+        expect(objects[0].doubleCol).equals(42);
+        done();
```
takameyer, line 60: TIL: there's a done() function

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+  let runCount = 0;
+
+  beforeEach(() => (runCount = 0));
+
+  describe("Realm notifications", () => {
+    openRealmBeforeEach({ schema: [TestObject] });
+
+    it("should fire on change", function (this: RealmContext, done) {
+      this.realm.addListener("change", (sender, event) => {
+        expect(sender).deep.equals(this.realm);
```
takameyer, line 53: I guess `sender` in this case is a `realm`?  Perhaps rename `sender` to be more clear.

Also interested to see if `expect(sender).deep.equals({})` passes.

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+        expect(event).equals("change");
+
+        const objects = this.realm.objects(TestObject);
+
+        expect(objects.length).equals(1);
+        expect(objects[0].doubleCol).equals(42);
```
takameyer, line 59: Kind of confused about the intention of these assertions (i'm sure this is just copied).  I think just making sure the callback gets called is enough of a test.

takameyer: Could also be later refactored

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+        this.realm.create("TestObject", { doubleCol: 2 });
+      });
+    });
+
+    it("should fire on update", function (this: RealmContext, done) {
+      testObjects.addListener(
+        expectCollectionChangesOnEveryRun(done, [
+          { insertions: [], newModifications: [], oldModifications: [], deletions: [] },
+          { insertions: [0, 1, 2], newModifications: [], oldModifications: [], deletions: [] },
+          { insertions: [], newModifications: [0, 2], oldModifications: [0, 2], deletions: [] },
```
takameyer, line 220: Would be good to see the difference between `newModifications` and `oldModifications`.  This could be done by doing a `delete` or `insert` combined with a modification in a single transaction.

takameyer (approved):
Mostly just potential refactoring suggestions.  LGTM 👍🏼

integration-tests/tests/src/tests/notifications.ts
gagik: me too! kinda cool

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+      });
+
+      this.realm.removeListener("change", handler);
+
+      this.realm.write(() => {
+        this.realm.create("TestObject", { doubleCol: 2 });
+      });
+
+      expect(runCount).equal(1);
+    });
```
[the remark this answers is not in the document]
gagik: yeah there is some async weirdness that can happen here, the "proper" test could be to do some kind of timed out done? i.e. give 1 second (or less, like 500ms) to make sure the listener does not get called and consider that good enough?

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+    properties: {
+      doubleCol: "double",
+    },
+  };
+}
+
+describe("Notifications", () => {
+  let runCount = 0;
+
+  beforeEach(() => (runCount = 0));
```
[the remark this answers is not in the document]
gagik: one reason I do this is also to make the runCount in scope for the helper functions so this would need some more changes if I do that.

integration-tests/tests/src/tests/notifications.ts
```diff
@@ -0,0 +1,463 @@
...
+  });
+
+  /**
+   * Helper function to use as a generic handler for collection listeners.
+   * @param onEnd Callback to run once the listener reaches the end of the changes array. If the callback is Mocha.Done, it will ensure that
+   * the listener was run exactly [changesOnRun.length] times
+   * @param changesOnRun An array of Realm.CollectionChangeSet.
+   * The handler will assert equality of listener's changes with the next element of the array on every run.
+   */
+  const expectCollectionChangesOnEveryRun = (
```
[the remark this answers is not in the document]
gagik: so related to the runCount comment as well, this is also there to be able to access the runCount in the correct scope. this may be a little hacky in that sense, not sure

integration-tests/tests/src/tests/notifications.ts
gagik: I actually wanted to ask about comparing realms as I was not sure how to actually do this. I called it sender because this is what the addListener function calls its callback's parameters, so perhaps it is worth renaming it there too, I did find it confusing.

gagik: `        expect(sender).deep.equals(this.realm);` does indeed pass... Should I compare their schemas? I think that's what I did originally, I am not sure what's best here.

takameyer: It's not a very valuable assertion in any case.  We register a listener when anything on the realm changes, and then compare if the realm is the same.  Can probably just remove it 🤷🏼

integration-tests/tests/src/tests/notifications.ts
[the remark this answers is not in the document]
kraenhansen: I would rather await something async than relying on timeouts, as they tend to add up when running the entire test suite.

integration-tests/tests/src/tests/notifications.ts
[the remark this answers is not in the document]
kraenhansen: Ahh. Makes sense then - I didn't notice that.

kraenhansen (approved):

integration-tests/tests/src/tests/notifications.ts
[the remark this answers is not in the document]
kraenhansen: > Mocha errors if done is called multiple times.

While this is true, I wouldn't rely on this, because it messes up the report. First Mocha thinks the test pass but then it fails the test. It's ultimately a race between the second call to `done` and the test command exiting green, which could lead to unstable tests.

PR #5348, opened 2023-01-30
Migrate object-tests to TS

JacobOscarGunnarsson:


integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1297 @@
 ////////////////////////////////////////////////////////////////////////////
 import { expect } from "chai";
 import Realm from "realm";
-
+import { Buffer } from "buffer";
```
kraenhansen, line 20: Unfortunately, this won't work when running on React Native.

integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1297 @@
...
+      await firstChange;
+      obj.removeListener(listener);
 
-      expect(john._id.equals(_id)).equals(true);
-      expect(john.name).equals("John Doe");
-      expect(john.age).equals(43);
+      this.realm.write(function () {
+        obj["stringCol"] = "foobar";
+      });
+      await sleep(500);
```
kraenhansen, line 1256: I would like to avoid using sleeps, but instead await a promise that resolves when the event fires. This would be a good candidate for a refactor in the future.

integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1320 @@
...
+
+      expect(() => {
+        //@ts-expect-error TYPEBUG: __to_object does not exist on realm.
+        this.realm.__to_object(null);
+      }).throws(TypeError);
+
+      expect(() => {
+        //@ts-expect-error TYPEBUG: __to_object does not exist on realm.
+        this.realm.__to_object(undefined);
+      }).throws(TypeError);
```
kraenhansen, line 1118: These and the `__to_boolean` will go await on the `bindgen` branch anyway.

integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1320 @@
...
-
+import { Buffer } from "buffer";
 import { IPerson, Person, PersonSchema } from "../schemas/person-and-dogs";
 import {
   IPerson as IPersonWithId,
   Person as PersonWithId,
   PersonSchema as PersonSchemaWithId,
 } from "../schemas/person-and-dog-with-object-ids";
+import { openRealmBeforeEach } from "../hooks";
+import { expectArraysEqual, expectSimilar } from "../utils/comparisons";
```
kraenhansen, line 28: In a future refactor, we could consider finding ways to avoid relying on these special comparisons.

kraenhansen (approved):

integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1323 @@
...
+        }
 
-      // Expect only one instance of 'PersonSchemaWithId' in db after all updates
-      const persons = realm.objects(PersonSchemaWithId.name);
-      expect(persons.length).equals(1);
+        function tryAssign(name: string, value: any) {
+          //@ts-expect-error TYPEBUG: indexing with string is not allowed by typesystem.
+          const prop = AllTypesSchema.properties[name];
+          const type = typeof prop == "object" ? prop.type : prop;
+          //@ts-expect-error TYPEBUG: indexing with string is not allowed by typesystem.
```
takameyer, line 696: If we used a class-based model it might not complain

integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1323 @@
...
+          expect(john.age).equals(42);
+        });
+      });
     });
 
-    it("can be updated", () => {
-      const realm = new Realm({ schema: [PersonSchemaWithId] });
-      const _id = new Realm.BSON.ObjectId();
+    describe("with primary key", () => {
+      openRealmBeforeEach({ schema: [PersonWithId] });
```
takameyer, line 523: I'm wondering if we should just inline define the schema in the `openRealmBeforeEach`.  This would make it clear at the top of the test what properties have been set without scrolling to the top of the test.  Possible refactoring/discussion

integration-tests/tests/src/tests/objects.ts
```diff
@@ -17,244 +17,1323 @@
...
+        return this.realm.create<ITestObject>("TestObject", { doubleCol: 1 });
+      });
+
+      const schema = obj.objectSchema();
+      expect(schema.name).equals(TestObjectSchema.name);
+      expectArraysEqual(Object.keys(schema.properties), Object.keys(TestObjectSchema.properties));
+      expect((schema.properties.doubleCol as Realm.ObjectSchemaProperty).type).equals("double");
+    });
+  });
+  //TODO move this to notofications test
```
takameyer, line 1166: Yes, please do 😄 Alternatively, if the notification tests are already covering the same methods, then we can just remove the code 🗑️

takameyer (approved):
Looks good.  Just a couple refactoring considerations.

takameyer:
There's a [linting error](https://github.com/realm/realm-js/actions/runs/4062403946/jobs/6993420106#step:6:352) that needs to be addressed before this can be merged.

integration-tests/tests/src/tests/objects.ts
JacobOscarGunnarsson: I'll leave this as is for now (this error exists in multiple test files already and it could be worth doing them altogether in a single commit). If you guys haven't discussed it already I think it'd be worth considering wether we should support this syntax in TS also or if we should encourage consumers to use the class based model instead.

integration-tests/tests/src/tests/objects.ts
JacobOscarGunnarsson: yeah that is definitely a pro, my counterargument though is that if we inline the definition, then for methods like this
image
we would have to replace `PersonSchema.name` with its raw value (unless there's no other way to do that). 

Think I'm biased as Well since I use "gd" to go to definition via vim and "ctrl + o" to jump back to previous location in file

integration-tests/tests/src/tests/objects.ts
JacobOscarGunnarsson: I moved them to `notifications.ts` if you feel like they don't cohere too well with existing tests in that file, feel free to ping me and I can rewrite the object-notifications

takameyer (approved):

.github/workflows/pr-realm-js.yml
```diff
@@ -320,7 +321,7 @@ jobs:
 
       - name: Create Mocha Env
         id: mocha-env
-        run: echo "::set-output name=name::syncLogLevel=warn,realmBaseUrl=${{ secrets.REALM_QA_BASE_URL }},mongodbClusterName=${{ steps.deploy-mdb-apps.outputs.clusterName }},privateKey=${{ secrets.ATLAS_QA_PRIVATE_API_KEY }},publicKey=${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}"
+        run: echo "::set-output name=name::syncLogLevel=warn,longTimeout=${{ env.LONG_TIMEOUT }},realmBaseUrl=${{ secrets.REALM_QA_BASE_URL }},mongodbClusterName=${{ steps.deploy-mdb-apps.outputs.clusterName }},privateKey=${{ secrets.ATLAS_QA_PRIVATE_API_KEY }},publicKey=${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}"
```
kraenhansen, line 324: I wonder why this is being added here, as it's already on the target branch: https://github.com/realm/realm-js/blob/test-migration-master/.github/workflows/pr-realm-js.yml#L329

Does this PR need a rebase?

kraenhansen (approved):

PR #5351, opened 2023-01-31
Retry Mechanism for App Importer

### LINKED ISSUE #5342: Create Retry Mechanism for App Importer
Currently we are getting intermittent failures from the Atlas Admin API.  It is possible that the time between app deployment and configuration is too short.  Most of the time it works, but when it doesn't it's quite frustrating.  A retry mechanism could probably solve a few of these intermittent issues.  

* Create generic network request wrapper if one does not already exist
* Create constants for retry counts and retry period
* Implement retries for any network requests made by the app importer

takameyer:
## What, How & Why?
* Wrap global fetch with `this.fetch`
* Failed requests will retry 5 times at a 1 second interval

This closes #5342

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-app-importer/src/AppImporter.ts
+++ b/packages/realm-app-importer/src/AppImporter.ts
@@ -132,6 +132,12 @@ type App = {
   environment: string;
 };
 
+type FetchArgs = Parameters<typeof fetch>;
+type FetchReturn = ReturnType<typeof fetch>;
+
+const RETRY_COUNT = 5;
+const RETRY_TIMEOUT = 1000;
+
 export class AppImporter {
   private readonly baseUrl: string;
   private readonly credentials: Credentials;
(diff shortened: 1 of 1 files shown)


packages/realm-app-importer/src/AppImporter.ts
```diff
@@ -162,6 +168,30 @@ export class AppImporter {
     }
   }
 
+  private async fetch(url: FetchArgs[0], init?: FetchArgs[1]): Promise<FetchReturn> {
```
kraenhansen, line 171:
```suggestion
  private async fetch(url: FetchArgs[0], init?: FetchArgs[1], retries = 0): Promise<FetchReturn> {
```

I would suggest implementing this as a recursive function with a `retries` counting up, to avoid having two calls to `fetch` in the same function body.

kraenhansen (commented):
On a general note (I think I mentioned elsewhere), I would have loved this to be some preflight request polling the server instead of adding retry logic to all requests.

In its current form, this PR makes a deterministic error case take longer to fail, which might be very annoying for someone debugging the integration tests (calling into this) locally with broken arguments

PR #5354, opened 2023-01-31
adjust path for windows ia32

iamer:
Fix ccache for Windows 32bit compilation


takameyer:
@iamer [This step](https://github.com/realm/realm-js/pull/5354/files#diff-63a513d26453362aa63e7f2a51f5e83c112a18561d1de82473faa5cf2bcbf0d8R59-R64) is preventing us from seeing if the `ccache` hits or not, as the prebuild is just being restored and the entire build step is skipped.  Wondering if we really need this step, as we can now allow on the `ccache` to build faster.
In any case, I need a way to see this is working.

iamer:
I removed the prebuild install step in a separate commit

takameyer:
@iamer Can we verify that the 32-bit builds work on `windows-latest`?  Other than that LGTM

PR #5355, opened 2023-01-31
Parallelize IOS platform builds

iamer:


kraenhansen:
I just noticed that the xcframework contains a `ios-arm64-armv7` when I build it locally, but a `ios-arm64` in the archive from CI.
I wonder if this needs to be built on an Apple silicon host?

PR #5356, opened 2023-01-31
Sync-based UUID, Dictionary tests

gagik:


takameyer (changes requested):
I would prefer that these were in separate files.  It would not be clear to me that `sync/types.ts` would include tests for `uuid`, `set` and `dictionary`.  
In general, it would be a breath of fresh air if our test files were shorter 😂

takameyer (approved):
Looks great!  Thanks 💯

PR #5360, opened 2023-02-01
Adding a `longTimeout` method to `Suite` and `Context`

kraenhansen:
## What, How & Why?

This PR adds:
- A `longTimeout` method to `Suite` and `Context` which expose a timeout configurable by the harness, to specify what "long" is. It defaults to 10s when running locally, but 5 minutes when ran on CI.
- Names to the hooks, making it easier to determine what hook is timing out from the error logged to console.
- Types and documentation for the context variables currently available / in-use by the test suite.

### DIFF
--- a/integration-tests/tests/src/hooks/authenticate-user-before.ts
+++ b/integration-tests/tests/src/hooks/authenticate-user-before.ts
@@ -18,7 +18,7 @@
 import { Credentials } from "realm";
 
 export function authenticateUserBefore(): void {
-  before(async function (this: AppContext & Partial<UserContext>) {
+  before("authenticateUserBefore", async function (this: AppContext & Partial<UserContext>) {
     if (this.app) {
       this.user = this.app.currentUser || (await this.app.logIn(Credentials.anonymous()));
     } else {
--- a/integration-tests/tests/src/hooks/authenticate-user-before.ts
+++ b/integration-tests/tests/src/hooks/authenticate-user-before.ts
@@ -18,7 +18,7 @@
 import { Credentials } from "realm";
 
 export function authenticateUserBefore(): void {
-  before("authenticateUserBefore", async function (this: AppContext & Partial<UserContext>) {
+  before(authenticateUserBefore.name, async function (this: AppContext & Partial<UserContext>) {
     if (this.app) {
       this.user = this.app.currentUser || (await this.app.logIn(Credentials.anonymous()));
     } else {
--- a/integration-tests/tests/src/index.ts
+++ b/integration-tests/tests/src/index.ts
@@ -18,6 +18,11 @@
 
 console.log("Loading Realm Integration Tests");
 
+/**
+ * Use the `longTimeout` context variable to override this.
+ */
+const DEFAULT_LONG_TIMEOUT = 10 * 1000; // 10s
+
 if (!global.fs) {
   throw new Error("Expected 'fs' to be available as a global");
 }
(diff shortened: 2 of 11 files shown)


integration-tests/tests/src/hooks/authenticate-user-before.ts
```diff
@@ -18,7 +18,7 @@
 import { Credentials } from "realm";
 
 export function authenticateUserBefore(): void {
-  before(async function (this: AppContext & Partial<UserContext>) {
+  before("authenticateUserBefore", async function (this: AppContext & Partial<UserContext>) {
```
elle-j, line 21: To prevent typos and make it easier to refactor if having to rename this in the future, perhaps `authenticateUserBefore.name` would be better to use than `"authenticateUserBefore"`? (Same for the other occurrences where the name is passed.)

integration-tests/tests/src/typings.d.ts
```diff
@@ -29,7 +29,80 @@ interface path {
 
 type Require = (id: string) => unknown;
 
-type Environment = Record<string, unknown>;
+type Environment = Record<string, unknown> & {
+  longTimeout?: number;
```
elle-j, line 33: We could add a comment to it (feel free to modify it).
```suggestion
  /** Set the number of milliseconds to use for the long timeout. */
  longTimeout?: number;
```

elle-j (approved):
This will be very helpful (and make things a bit cleaner in our tests)!

integration-tests/tests/src/typings.d.ts
kraenhansen: I added docs for all the other variables, but forgot to add it for the one I was actually adding 🙈

elle-j: 😛

JacobOscarGunnarsson (approved):
LGTM!

takameyer:
Late to the party.  I'm wondering if it would be possible to set a default somewhere at the top level of the test harness?  If that was set to 10 seconds, then we could avoid calling `this.timeout` or `this.longTimeout` altogether.

integration-tests/tests/src/index.ts
```diff
@@ -44,7 +49,22 @@ afterEach(() => {
...
 // Using `require` instead of `import` here to ensure the Mocha globals (including `skipIf`) are set
 
-describe("Test Harness", () => {
+describe("Test Harness", function (this: Mocha.Suite) {
+  /**
+   * @see [typings.d.ts](./typings.d.ts) for documentation.
+   */
+  function longTimeout(this: Mocha.Context | Mocha.Suite) {
+    this.timeout(environment.longTimeout || DEFAULT_LONG_TIMEOUT); // 5 minutes
+  }
```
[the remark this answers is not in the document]
kraenhansen: The `mocha` API expose a method which needs to be called to set the timeout.

kraenhansen:
> If that was set to 10 seconds

I don't think thats a good default for the average async test, it would prolong the running of a test suite with unresolved promises too much in my opinion.

takameyer:
@kraenhansen So there are tests that are waiting to fail in order to pass?  Then we should probably remove this large timeout in CI, as this is waisting a lot of time.

kraenhansen:
> So there are tests that are waiting to fail in order to pass?

I sure don't hope so. That wasn't what I meant. I just do think more than 2 sec timeout when debugging locally would be a good DX when iterating async tests.

kraenhansen:
With this PR we effectively have the default of 2 secs for any test and a "longer" (defaults to 10s when running locally) we can use for any test that involves the server.

PR #5365, opened 2023-02-02
test migration:  open-behaviour

JacobOscarGunnarsson:


integration-tests/tests/src/tests/sync/open-behavior.ts
```diff
@@ -0,0 +1,658 @@
...
+      const user = await this.app.logIn(returningUserCredentials);
+      const config: Realm.Configuration = {
+        schema: [DogForSyncSchema],
+        sync: {
+          user,
+          partitionValue,
+          //@ts-expect-error TYPEBUG: cannot access const enum at runtime
+          _sessionStopPolicy: "immediately",
+          //@ts-expect-error TYPEBUG: cannot access const enum at runtime
+          existingRealmFileBehavior: { type: "downloadBeforeOpen" },
```
kraenhansen, line 197: I suspect we need a TODO to fix this?

integration-tests/tests/src/tests/sync/open-behavior.ts
```diff
@@ -0,0 +1,658 @@
...
+      return promise;
+    });
+
+    openPromise
+      .then(() => {
+        expect(true).to.be.false("Realm was opened after being canceled.");
+      })
+      .catch((err) => {
+        expect(true).to.be.false("An error was thrown after open was canceled: " + err.message);
+      });
```
kraenhansen, line 529: You might as well throw here ... I see this pattern a few places below.

```suggestion
    const unexpectedPromise = openPromise
      .then(() => {
        throw new Error("Realm was opened after being canceled.");
      })
      .catch((err) => {
        throw new Error("An error was thrown after open was canceled: " + err.message);
      });
```

That being said, this will become unhandled promise rejections, so the result of this promise (`unexpectedPromise`) should probably be included in the `Promise.race` below.

integration-tests/tests/src/tests/sync/open-behavior.ts
```diff
@@ -0,0 +1,658 @@
...
+          user,
+          partitionValue,
+          //@ts-expect-error TYPEBUG: cannot access const enum at runtime
+          _sessionStopPolicy: "immediately",
+          //@ts-expect-error testing invalid value for newRealmFileBehaviour
+          newRealmFileBehavior: { type: "openLocalRealm", timeOut: "bar" },
+        },
+      }),
+    ).throws;
+    await user.logOut();
```
kraenhansen, line 656: Note for refactor: If the test fails, this will never happen and if other tests are relying on this, it should probably be put into an `afterEach` hook instead.

kraenhansen (approved):
There's quite a bit of refactoring that we're pushing in front of us here.

One thing that caught my is is the existing `Realm.open` both throws and return rejected promises when failing. The new SDK [will only return rejected promises](https://github.com/realm/realm-js/blob/bindgen/packages/realm/src/ProgressRealmPromise.ts#L132-L134).

integration-tests/tests/src/tests/sync/open-behavior.ts
JacobOscarGunnarsson: exactly! The idea is that every place labeled with TYPEBUG is something the type system (assuming I haven't missed something) emits an error for. The idea is when bindgen is rebased on this the `ts-expect-error` will either remain the same, meaning that the error still exists, or it will get a red squiggle, which would imply that the type bug has been resolved

integration-tests/tests/src/tests/sync/open-behavior.ts
```diff
@@ -0,0 +1,658 @@
...
+    };
+
+    expect(Realm.exists(config)).to.be.false;
+
+    const realm = await Realm.open(config);
+
+    expect(realm.path).to.not.be.undefined;
+
+    realm.close();
+    await user.logOut();
```
takameyer, line 77: We should probably create a `logOutAfterEach` action.  Seems possible one could forget to do this and then there would still be an open user session on the next test.

integration-tests/tests/src/tests/sync/open-behavior.ts
```diff
@@ -0,0 +1,658 @@
...
+            timeOutBehavior: "openLocalRealm",
+          },
+        },
+      };
+      await Realm.open(config);
+
+      await user.logOut();
+
+      // Wait for the timeout to run.  This used to crash, since it opens a local realm with a logged out user.
+      await new Promise<void>((resolve) => {
```
takameyer, line 495: This could be replaced with `await sleep(1000)`

takameyer: There are other places in this test doing the same

takameyer (approved):
Some refactoring opportunities.  Otherwise LGTM

integration-tests/tests/src/tests/sync/open-behavior.ts
JacobOscarGunnarsson: think a separate hook like that would be good yeah, but for now `Realm.ClearTestState` both logs out all users and removes existing realms if I'm not mistaken, so should be good for now

PR #5369, opened 2023-02-03
Adding JSC variant when running integration tests

kraenhansen:
## What, How & Why?

This ensures we're testing on JSC when running our integration tests.

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -164,7 +164,7 @@ jobs:
             ${{ matrix.variant.artifact-path }}
 
   integration-tests:
-    name: Integration tests (${{matrix.variant.target}}) for ${{ matrix.variant.environment }} on ${{ matrix.variant.os }}
+    name: Integration tests (${{matrix.variant.target}}) for ${{ matrix.variant.environment }} on ${{ matrix.variant.os }} ${{ matrix.variant.engine || '' }}
     needs: [build]
     env:
       REALM_DISABLE_ANALYTICS: 1
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -200,7 +200,7 @@ jobs:
     steps:
       - name: Generate Cluster Differentiator
         id: differentiator
-        run: echo "::set-output name=name::${{matrix.variant.os}}-${{matrix.variant.target}}-${{matrix.variant.environment}}-${{env.GITHUB_RUN_ATTEMPT}}"
+        run: echo "::set-output name=name::${{matrix.variant.os}}-${{matrix.variant.target}}-${{matrix.variant.environment}}-${{matrix.variant.engine}}-${{env.GITHUB_RUN_ATTEMPT}}"
 
       - uses: realm/ci-actions/mdb-realm/deployApps@fac1d6958f03d71de743305ce3ab27594efbe7b7
         id: deploy-mdb-apps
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -294,8 +296,15 @@ jobs:
       - name: Install pods
         if: ${{ matrix.variant.environment == 'react-native' && matrix.variant.os == 'ios'}}
         working-directory: integration-tests/environments/${{ matrix.variant.environment }}
+        env:
+          USE_HERMES: ${{ matrix.variant.engine == 'hermes' && '0' || '1' }}
```
takameyer, line 300: Why '0' || '1'?

.github/workflows/pr-realm-js.yml
```diff
@@ -191,8 +191,10 @@ jobs:
           - { os: darwin, target: "test:main", runner: macos-latest, environment: electron }
           - { os: darwin, target: "test:renderer", runner: macos-latest, environment: electron }
           - { os: darwin, target: test, runner: macos-latest, environment: node }
-          - { os: android, target: "test:android", runner: macos-latest, environment: react-native, arch: "armeabi-v7a" }
-          - { os: ios, target: "test:ios", runner: macos-latest, environment: react-native, arch: "ios" }
+          - { os: android, target: "test:android", runner: macos-latest, environment: react-native, arch: "armeabi-v7a", engine: "hermes" }
```
takameyer, line 194: The Differentiator below needs to be updated to be unique for the new matrix entries

takameyer (changes requested):
I think we need to refactor our cluster deployment strategy before merging this (plus the differentiator is invalid for the new steps).  We are going to hit the cluster limit more often.

.github/workflows/pr-realm-js.yml
kraenhansen: Because we want `USE_HERMES` to equal `1` when engine is 'hermes' and GitHub Actions doesn't have a ternary operator. But actually - looking at this again, I messed up the order 🤦

.github/workflows/pr-realm-js.yml
takameyer: Also we should include the engine in the title of the action

takameyer (approved):
LGTM.  Though we are probably going to hit our cluster limit quite often now.

kraenhansen:
I'll be rebasing this on the new `bindgen` and probably await #5523 too.

.github/workflows/pr-realm-js.yml
```diff
@@ -294,8 +296,15 @@ jobs:
...
         if: ${{ matrix.variant.environment == 'react-native' && matrix.variant.os == 'ios'}}
         working-directory: integration-tests/environments/${{ matrix.variant.environment }}
+        env:
+          USE_HERMES: ${{ matrix.variant.engine == 'hermes' && '1' || '0' }}
         run: npx pod-install
 
+      - name: Set hermesEnabled
+        working-directory: integration-tests/environments/${{ matrix.variant.environment }}
+        if: ${{ matrix.variant.os == 'android' }}
+        run: echo "hermesEnabled=${{ matrix.variant.engine == 'hermes' }}\n" >> android/gradle.properties
```
kraenhansen, line 306: I don't think this is correct anymore - this is how we do it in the install tests: https://github.com/realm/realm-js/blob/main/install-tests/react-native/patches/jsc-build.gradle.patch .. or perhaps we can use a `GRADLE_OPTS` environment variable.

kraenhansen:

PR #5353, opened 2023-01-31
Make `BaseSubscriptionSet` type extend `DeprecatedReadonlyArray`

elle-j:
## What, How & Why?
Deprecates the array-like behavior of the `BaseSubscriptionSet` type.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 📝 Update `COMPATIBILITY.md`
* [ ] 🚦 Tests
* [ ] 🔀 Executed flexible sync tests locally if modifying flexible sync
* [ ] 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)
* [ ] 📱 Check the React Native/other sample apps work if necessary
* [ ] 📝 Public documentation PR created or is not necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [x] typescript definitions file is updated
* [ ] jsdoc files updated

PR #5374, opened 2023-02-06
Remove ability to use index operator on `BaseSubscriptionSet`

elle-j:
## What, How & Why?
Removes the ability to do `realm.subscriptions[0]`. If users need this functionality, they will need to do `[...realm.subcriptions][0]`.

Note:
* This aligns with the [deprecation](https://github.com/realm/realm-js/pull/5353) of `ReadonlyArray`.
* In the merged Flexible Sync PR, this functionality was still possible. What was not possible was accessing JS array methods.

## ☑️ ToDos
* [x] 🚦 Tests

### DIFF
--- a/packages/realm/src/app-services/BaseSubscriptionSet.ts
+++ b/packages/realm/src/app-services/BaseSubscriptionSet.ts
@@ -81,9 +81,9 @@ const PROXY_HANDLER: ProxyHandler<BaseSubscriptionSet> = {
  *
  * SubscriptionSets can only be modified inside a {@link SubscriptionSet.update} callback.
  *
- * The SubscriptionSet is an iterable; thus, if absolutely needed, the contained
- * {@link Subscription}s can be accessed in `for-of` loops or spread into an `Array`
- * for access to the ECMAScript Array API, e.g. `[...realm.subscriptions][0]`.
+ * The SubscriptionSet is an iterable; thus, the contained {@link Subscription}s can be
+ * accessed in `for-of` loops or spread into an `Array` for access to the ECMAScript
+ * Array API, e.g. `[...realm.subscriptions][0]`.
  */
 export abstract class BaseSubscriptionSet {
   /**@internal */
--- a/packages/realm/src/app-services/BaseSubscriptionSet.ts
+++ b/packages/realm/src/app-services/BaseSubscriptionSet.ts
@@ -56,19 +56,6 @@ export enum SubscriptionsState {
 
 const DEFAULT_PROPERTY_DESCRIPTOR: PropertyDescriptor = { configurable: true, enumerable: true, writable: false };
 const PROXY_HANDLER: ProxyHandler<BaseSubscriptionSet> = {
-  get(target, prop) {
-    if (Reflect.has(target, prop)) {
-      return Reflect.get(target, prop);
-    }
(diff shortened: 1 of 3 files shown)


kraenhansen (approved):

kraenhansen:
It does make the tests a little more complicated to maintain.

Perhaps they could benefit from some destructuring themselves, ex:
```
const [sub0, sub1, sub2] = subscriptions;
```

packages/realm/src/app-services/BaseSubscriptionSet.ts
```diff
@@ -56,19 +56,6 @@ export enum SubscriptionsState {
 
 const DEFAULT_PROPERTY_DESCRIPTOR: PropertyDescriptor = { configurable: true, enumerable: true, writable: false };
 const PROXY_HANDLER: ProxyHandler<BaseSubscriptionSet> = {
```
RedBeard0531, line 58: Can we remove the whole Proxy wrapping now?

elle-j: We still need it to enable the spread operator.

packages/realm/src/app-services/BaseSubscriptionSet.ts
```diff
@@ -202,11 +179,6 @@ export abstract class BaseSubscriptionSet {
     return subscription ? (new Subscription(subscription) as Subscription) : null; // TODO: Remove the type assertion into Subscription
   }
 
-  /**
-   * Enables index accessing when combined with the proxy handler's `get()` method.
```
elle-j, line 206: Creating a thread here to more easily discuss removing index accessing:

Previous comments:
> @kraenhansen
> It does make the tests a little more complicated to read / maintain.
>
> Perhaps they could benefit from some destructuring themselves, ex:
>
> `const [sub0, sub1, sub2] = subscriptions;`

> @RedBeard0531 
> Or use the findByName() API any time we need to be able to refer to a specific subscription.

Yeah I noticed particularly when refactoring the tests that the accessing by index felt more convenient. If having to use `findByName()`, users would have to always name their subscriptions.

Any more thoughts on this?

RedBeard0531: Names are stable identifiers. Indexes are not. Users shouldn't have any expectation that the index<->subscription mapping is stable over time. If they need to refer to a subscription later, that is what names are for.

elle-j: True. Seems a bit difficult to know how convenient [0] would be for users.. Or how inconvenient not having it would be?

kneth (approved):

packages/realm/src/app-services/BaseSubscriptionSet.ts
kraenhansen: I suspect it's mostly in the context of these tests that it gets complicated. Building an app I would probably store a reference to the subscription and use that to remove it later.

That being said, (and just to make it obvious) having en index accessor is not a prerequisite for the snippet I posted.

JacobOscarGunnarsson (approved):
LGTM

takameyer (approved):
LGTM

PR #5375, opened 2023-02-06
Improve install tests (setting `newArchEnabled` and `hermesEnabled` correctly on Android)

kraenhansen:
## What, How & Why?

We had a bug in the way the install tests configured the Android test to run with Hermes disabled.
Also, we now test throwing in code called from C++, because that has historically been (and is currently) broken (because of https://github.com/facebook/react-native/issues/36052).

This also makes it so PRs updating the install test workflow won't post and spam the team channel anymore.

### DIFF
--- a/.github/workflows/install-test-react-native.yml
+++ b/.github/workflows/install-test-react-native.yml
@@ -18,7 +18,7 @@ defaults:
 
 jobs:
   install:
-    name: Install Test on React Native ${{ toJSON(matrix) }}
+    name: Install Test on React Native
     runs-on: macos-12
     timeout-minutes: 120
     strategy:
@@ -53,6 +53,9 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
+      - id: set-summary
+        run: |
+          node -p "Object.entries(${{ toJSON(matrix) }}).map(([k,v]) => k + ': ' + v).join(', ')" >> $GITHUB_STEP_SUMMARY
 
       - name: ccache
         uses: hendrikmuhs/ccache-action@v1.2
--- a/.github/workflows/install-test-react-native.yml
+++ b/.github/workflows/install-test-react-native.yml
@@ -18,6 +18,7 @@ defaults:
 
 jobs:
   install:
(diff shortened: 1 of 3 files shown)


install-tests/react-native/cli.ts
```diff
@@ -182,18 +182,20 @@ yargs(hideBin(process.argv))
...
+      const localPropertiesContent =
+        "\n# Install test overwrites below\n\n" +
+        Object.entries(localProperties)
+          .map(([k, v]) => `${k}=${v}`)
+          .join("\n");
 
-      console.log(`Writing local.properties to ${localPropertiesPath}`);
-      fs.writeFileSync(localPropertiesPath, localPropertiesContent);
+      console.log(`Appending gradle properties to ${localPropertiesPath}`);
+      fs.appendFileSync(localPropertiesPath, localPropertiesContent);
```
kraenhansen, line 198: This fixes configuring hermes and new architecture.

.github/workflows/install-test-react-native.yml
```diff
@@ -18,6 +18,7 @@ defaults:
 
 jobs:
   install:
+    name: Install Test on React Native ${{ matrix.platform == 'ios' && 'iOS' || 'Android' }} realm@${{ matrix.realm-version }}, react-native@${{ matrix.react-native-version }} new architecture ${{ matrix.new-architecture && 'enabled' || 'disabled' }} running ${{ matrix.engine }}
```
kraenhansen, line 21: Making the name of the job a bit more readable:

install-tests/react-native/patches/App.js
```diff
@@ -33,13 +37,29 @@ const App = () => {
...
+
     // Read the persons out of the database again
     const message =
       "Persons are " +
       realm
         .objects("Person")
         .map((p) => p.name)
-        .join(", ");
+        .join(", ") +
+      suffix;
```
takameyer, line 62: If i'm understanding this right, this will output a "!" when successful?  Is that meaningful?

kraenhansen: > Is that meaningful?

Not really 🙂 I just need some string to round-trip and expect.

PR #5378, opened 2023-02-06
[bindgen] Migrated `tsm` → `tsx`

kraenhansen:
## What, How & Why?

I recently learned that the community seem to be converging on another TS runtime.
It wasn't a big change, so I figured we might as well change this to a safer bet - it also seem a bit less verbose to use in shebangs.
See https://github.com/privatenumber/ts-runtime-comparison for context.


elle-j (approved):
LGTM

RedBeard0531:
👍

Looks like this might solve our issue with node 18.

PR #5382, opened 2023-02-06
[bindgen] Use `node-fetch` as network on Node

kraenhansen:
## What, How & Why?

It seems like implementing network on Node using the built-in `node:http` and `node:https` modules, are a less ideal match when implementing MongoDB watch. This is an attempt to swap it for `node-fetch` which provide a `ReadableStream` API to the request body.


packages/realm/src/app-services/MutableSubscriptionSet.ts
```diff
@@ -104,7 +104,7 @@ export class MutableSubscriptionSet extends BaseSubscriptionSet {
    * @param query A {@link Results} instance representing the query to remove a subscription to.
    * @returns `true` if the subscription was removed, `false` if it was not found.
    */
-  remove(query: Results<unknown>): boolean {
+  remove(query: Results<any>): boolean {
```
kraenhansen, line 107: While in the tests, I saw that this API was a bit too restrictive.

kraenhansen:
Closing now that https://github.com/realm/realm-js/tree/lj/bindgen/collection-watch got rebased on this branch.

PR #5390, opened 2023-02-08
create database name that depends on either local or remote server

JacobOscarGunnarsson:
locally we use the hardcoded "test-database" to validate our mongoclient tests, when running it remote however the database name will be unique to make sure that tests run in isolation. The import app function must therefor return a database name depending on the context


integration-tests/tests/src/utils/import-app.ts
```diff
@@ -92,7 +92,8 @@ export async function importApp(
 ): Promise<{ appId: string; baseUrl: string; databaseName: string }> {
   const { baseUrl, appImporterUrl } = getUrls();
```
JacobOscarGunnarsson, line 94: we could add an optional `databaseName` to `importApp` and `importAppBefore` but I'm not sure if that's the cleanest solution..

PR #5391, opened 2023-02-08
test migration: session tests

JacobOscarGunnarsson:
This file is also quite big meaning that there's probably a lot of tests that could be moved to separate files, I've decided to postpone it for now and focus more on just moving the tests


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 ### Internal
 * Upgraded Realm Core from v13.1.1 to v13.1.2. ([#5174](https://github.com/realm/realm-js/issues/5174))
 <<<<<<< HEAD
+<<<<<<< HEAD
```
kraenhansen, line 36: Looks like a bad merge.

integration-tests/tests/src/tests/queries.ts
```diff
@@ -126,7 +126,7 @@ const expectQueryResultValues = (
     expect(expectedResults).to.deep.equal(
```
kraenhansen, line 126: The PR is titled "session tests" - were these changes to the queries tests intentional?

JacobOscarGunnarsson:
PR should now contain the correct modifications 🤞

takameyer (approved):

kraenhansen (approved):

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -0,0 +1,1415 @@
...
+          progressCalled = true;
+        }),
+        new Promise(() => {
+          sleep(5000, false);
+        }),
+      ]);
+      expect(progressCalled).to.be.true;
+    });
+
+    it("removing progress notification does not invoke callback again", async function (this: AppContext) {
```
JacobOscarGunnarsson, line 321: same here, it used `seedDataWithExternalUser` before. Thoughts about removing this?

integration-tests/tests/src/utils/sleep.ts
```diff
@@ -22,6 +22,8 @@ declare function setTimeout(cb: () => void, ms: number): void;
  * @param ms For how long should the promise be pending?
  * @returns A promise that returns after `ms` milliseconds.
  */
-export function sleep(ms = 1000): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, ms));
+export function sleep(ms = 1000, shouldResolve = true): Promise<void> {
+  return new Promise((resolve, reject) => {
+    setTimeout(shouldResolve ? resolve : reject, ms);
+  });
 }
```
kraenhansen, line 29: I don't like this change as it will reject passing a `Timeout` which is doesn't extend `Error`.

I would rather that we introduce another function:

```suggestion
export function sleep(ms = 1000, shouldResolve = true): Promise<void> {
  return new Promise((resolve, reject) => {
    setTimeout(shouldResolve ? resolve : reject, ms);
  });
}

export async function timeout(ms = 1000) {
  await sleep(ms);
  throw new Error(`Timed out after ${ms}ms`);
}
```

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -0,0 +1,1415 @@
...
+      const partition = generatePartition();
+      const { config } = await getSyncConfWithUser(this.app, partition);
+      let progressCalled = false;
+      await Promise.race([
+        Realm.open(config).progress(() => {
+          progressCalled = true;
+        }),
+        new Promise(() => {
+          sleep(5000, false);
+        }),
```
kraenhansen, line 316: No need to wrap this in a promise constructor.
Also, since the callback body is neither awaiting nor returning the promise, this is effectively a promise that never resolves but calls `sleep` which might become an unhandled promise rejection.

```suggestion
        sleep(5000, false),
```

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -0,0 +1,1415 @@
...
+        expect(Realm.App.Sync._hasExistingSessions(this.app)).to.be.true;
+        realm.close();
+
+        // Wait for the session to finish
+        return async () => {
+          for (let i = 50; i >= 0; i--) {
+            if (!Realm.App.Sync._hasExistingSessions(this.app)) {
+              return;
+            }
+            sleep(100);
```
kraenhansen, line 665: A promise like this must be awaited.

```suggestion
            await sleep(100);
```

integration-tests/realm-apps/with-db/services/mongodb/config.json
```diff
@@ -8,7 +8,7 @@
       "partition": {
         "key": "partition",
         "type": "string",
-        "required": true,
+        "required": false,
```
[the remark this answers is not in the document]
JacobOscarGunnarsson: I see your point here. The other two suggestions I have then is to either add partition values to the objects we create in this test (which I believe only incurs a lot of noise since the partition value is not tested). Or I can create a new app to use with sync and an optional partition key, but I feel like that would also bloat the codebase with another app.. any other suggestions? 🤔

takameyer:
> I don't like this change as it will reject passing a `Timeout` which is doesn't extend `Error`.

I would rather that we introduce another function:

```suggestion
export function sleep(ms = 1000, shouldResolve = true): Promise<void> {
  return new Promise((resolve, reject) => {
    setTimeout(shouldResolve ? resolve : reject, ms);
  });
}

export async function timeout(ms = 1000) {
  await sleep(ms);
  throw new Error(`Timed out after ${ms}ms`);
}
``` 

Wonder if we can come up with a better name than `timeout`.  Don't want this to be confused with `this.timeout`.

takameyer (approved):
A few considerations.  Once resolved, LGTM

integration-tests/tests/src/tests/sync/realm-conversions.ts
```diff
@@ -0,0 +1,307 @@
...
+        if (destination.sync) {
+          await realmDst.syncSession?.downloadAllServerChanges();
+        }
+        expect(realmDst.objects("Dog").length).equals(5);
+        realmDst.close();
+      });
+    }
+
+    afterEach(async function (this: AppContext & ConversionTestContext) {
+      const configSrc = this.currentTest?.ctx?.configSrc;
```
gagik, line 144: Not sure if there is a better way to retreive this information / accomplish... this seems like a rather ugly way to do so... but it does work.

tests/js/session-tests.js
```diff
@@ -1,1593 +0,0 @@
...
-              oldState === Realm.App.Sync.ConnectionState.Connected &&
-              newState === Realm.App.Sync.ConnectionState.Disconnected
-            ) {
-              resolve();
-            }
-          });
-          realm.close();
-        });
-      });
-  },
```
kraenhansen, line 583: @JacobOscarGunnarsson this seems to be missing from the migrated tests.

kraenhansen: Or is it this? https://github.com/realm/realm-js/pull/5391/files#diff-0ab9adc1bbb7aee926da3074f9a75569f592434ba0585601cb9321ff5410a7d6R383-R404

PR #5397, opened 2023-02-09
Add move/remove/swap methods on List

### LINKED ISSUE #3324: Implement Realm.List.{move, swap, remove}
Realm Object Store provides three methods currently not exposed in RJS: move, swap and remove.

References:

* [https://github.com/realm/realm-object-store/blob/v10/src/list.hpp#L83](https://github.com/realm/realm-object-store/blob/v10/src/list.hpp#L83)

* https://github.com/realm/realm-object-store/blob/v10/src/list.hpp#L86

papafe:
This closes #3324

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## vNext (TBD)
 
 ### Enhancements
-- Added 3 new methods on lists ([#3324](https://github.com/realm/realm-js/issues/3324)):
-  - `list.remove(index)`: removes the element of the list at the specified index.
-  - `list.move(from, to)`: moves one element of the list from one index to another.
-  - `list.swap(index1, index2)`: swaps the positions of the elements of the list at two indexes.
+* Added 3 new methods on lists ([#3324](https://github.com/realm/realm-js/issues/3324)):
+  * `list.remove(index)`: removes the element of the list at the specified index.
+  * `list.move(from, to)`: moves one element of the list from one index to another.
+  * `list.swap(index1, index2)`: swaps the positions of the elements of the list at two indexes.
 
 ### Fixed
 * Fix no notification for write transaction that contains only change to backlink property. ([realm/realm-core#4994](https://github.com/realm/realm-core/issues/4994), since v10.8.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+- Added 3 new methods on lists ([#3324](https://github.com/realm/realm-js/issues/3324)):
+  - `list.remove(index)`: removes the element of the list at the specified index.
+  - `list.move(from, to)`: moves one element of the list from one index to another.
+  - `list.swap(index1, index2)`: swaps the positions of the elements of the list at two indexes.
 
 ### Fixed
 * Fix no notification for write transaction that contains only change to backlink property. ([realm/realm-core#4994](https://github.com/realm/realm-core/issues/4994), since v10.8.0)
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -23,8 +23,8 @@ import { openRealmBefore } from "../hooks";
 
 type Item<T = unknown> = { list: Realm.List<T> };
 
-describe.only("List", () => {
-  describe.only("with unconstrained (mixed) values", () => {
+describe("List", () => {
+  describe("with unconstrained (mixed) values", () => {
     openRealmBefore({
       schema: [
         {
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -1,0 +1,103 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { expect } from "chai";
+import { Realm } from "realm";
+
+import { openRealmBefore } from "../hooks";
+
+type Item<T = unknown> = { list: Realm.List<T> };
+
+describe.only("List", () => {
+  describe.only("with unconstrained (mixed) values", () => {
+    openRealmBefore({
+      schema: [
+        {
+          name: "Item",
+          properties: { list: { type: "list", objectType: "mixed" } },
+        },
+      ],
+    });
+
+    it("supports remove", function (this: RealmContext) {
+      const { list } = this.realm.write(() =>
+        this.realm.create<Item>("Item", {
+          list: [2, 5, 8, 14, 57],
+        }),
+      );
+
+      expect([...list]).deep.equals([2, 5, 8, 14, 57]);
+
+      this.realm.write(() => {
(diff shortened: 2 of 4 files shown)


papafe:
There are a couple of things that I am not sure about, so this is still a draft to get early feedback. 

What kind of behaviour do we want for out of bounds indexes input? It seems that to align on the expectations of JS developers we could have different behaviour for the methods:
- `remove`: it will be a no-op to remove something out of bounds, so that it acts similar to `shift` and `pop`
- `move`/`swap`: those two will throw an exception for out of bounds inputs
Do we agree on this?

I have created a stub for the `list` test suite, looking at what was done for sets and dictionaries. I am unsure of what level of testing do we want to have with this. For instance, do we want to test these methods with multiple kind of list types (int, string, ...) or we just keep it simple?

packages/realm/src/List.ts
```diff
@@ -279,4 +279,38 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
     }
     return result;
   }
+
+  remove(index: number) {
+    assert.inTransaction(this.realm);
+    assert.number(index, "index");
+    const { internal } = this;
```
kraenhansen, line 286: I'd probably just inline `this.internal` as it's only used twice below.

packages/realm/src/List.ts
```diff
@@ -279,4 +279,38 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
...
+
+  move(from: number, to: number) {
+    assert.inTransaction(this.realm);
+    assert.number(from, "index");
+    assert.number(to, "to");
+
+    //TODO Need to decide if out-of-bounds indexes will be a no-op or an exception
+    const { internal } = this;
+
+    internal.move(from, to);
```
kraenhansen, line 303: This I would definitely not de-structure, because it makes `internal` appear twice instead of once.

```suggestion
    this.internal.move(from, to);
```

kraenhansen (commented):
Great approach, I like it.

Regarding the testing situation, we've talked about codifying a testing strategy and some sort of taxonomy such that we'll increate the likelihood to cover the entire API surface, while not maintaining the same tests twice. Iterating a collection is one example - should that live in a List test or would it make more sense to group that to increase code reuse? Similarly for object creation, should that be tested in an "objects.ts" or "realm.ts"? It creates objects but the `create` method hangs off the `Realm` instance. It would be great if we had principle decisions made on this that we could simply refer to.

The tests you've made this far seems valuable 👍  There are tests being migrated, which our `bindgen` branch will soon be rebased on-top-of: https://github.com/realm/realm-js/blob/test-migration-master/integration-tests/tests/src/tests/list.ts

integration-tests/tests/src/tests/list.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import { expect } from "chai";
+import { Realm } from "realm";
+
+import { openRealmBefore } from "../hooks";
+
+type Item<T = unknown> = { list: Realm.List<T> };
+
+describe.only("List", () => {
+  describe.only("with unconstrained (mixed) values", () => {
```
kraenhansen, line 27: Although convenient, I've seen green tests on CI where there were actually regressions on multiple occasions because someone merged a PR but had forgotten to remove their `.only` statements.

I'd suggest [using `--grep`](https://github.com/realm/realm-js/tree/bindgen/integration-tests#running-the-tests-during-development) instead, as that won't accidentally be committed.

integration-tests/tests/src/tests/list.ts
```diff
@@ -0,0 +1,103 @@
...
+      });
+
+      expect([...list]).deep.equals([14, 5, 8, 2, 57]);
+
+      this.realm.write(() => {
+        list.swap(2, 1);
+      });
+
+      expect([...list]).deep.equals([14, 8, 5, 2, 57]);
+    });
```
takameyer, line 101: We should also test what happens when the index is out of range

packages/realm/src/List.ts
takameyer: It's also being done elsewhere outside of this PR (see `splice` above for example).

takameyer:
Seems correct so far!  Just need to decide what to do with those out of bound indexes.  I think we should throw.

packages/realm/src/List.ts
kraenhansen: Yeah. And I'm all for that, because it has 9 occurrences of internal it simplifies the code. In this case it doesn't serve that purpose.

integration-tests/tests/src/tests/list.ts
papafe: Makes sense! I can imagine it happening. 
If that's the case probably it will be worth adding a check on CI too, so it doesn't happen again 😁

packages/realm/src/List.ts
```diff
@@ -279,4 +279,38 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
     }
     return result;
   }
+
+  remove(index: number) {
+    assert.inTransaction(this.realm);
+    assert.number(index, "index");
+    const { internal } = this;
+
+    //TODO I am not sure exactly if we expect a no-op for out of bounds operations or an exception
```
kneth, line 288: You could argue that out-of-bound should be a no-op as it is more JSish. But as the methods here are not inspired by JS Array, I think we should throw if out of bound.

packages/realm/src/List.ts
```diff
@@ -279,4 +279,38 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
...
+
+    internal.remove(index);
+  }
+
+  move(from: number, to: number) {
+    assert.inTransaction(this.realm);
+    assert.number(from, "index");
+    assert.number(to, "to");
+
+    //TODO Need to decide if out-of-bounds indexes will be a no-op or an exception
```
kneth, line 300: I am in favor of checking

kneth (approved):

integration-tests/tests/src/tests/list.ts
papafe: Just added

packages/realm/src/List.ts
papafe: Added

packages/realm/src/List.ts
```diff
@@ -279,4 +279,38 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
...
   }
+
+  remove(index: number) {
+    assert.inTransaction(this.realm);
+    assert.number(index, "index");
+    const { internal } = this;
+
+    //TODO I am not sure exactly if we expect a no-op for out of bounds operations or an exception
+    assert(index >= 0, "Index cannot be smaller than 0");
+    assert(index < internal.size, "Index cannot be bigger than the size of the list");
```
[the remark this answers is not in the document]
papafe: Fixed

packages/realm/src/List.ts
papafe: Makes sense. I've removed the destructuring.

papafe:
I have addressed all the comments (I think) and I have added the changelog. This is ready to merge in my opinion

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+- Added 3 new methods on lists ([#3324](https://github.com/realm/realm-js/issues/3324)):
+  - `list.remove(index)`: removes the element of the list at the specified index.
+  - `list.move(from, to)`: moves one element of the list from one index to another.
+  - `list.swap(index1, index2)`: swaps the positions of the elements of the list at two indexes.
```
takameyer, line 7:
```suggestion
* Added 3 new methods on lists ([#3324](https://github.com/realm/realm-js/issues/3324)):
  * `list.remove(index)`: removes the element of the list at the specified index.
  * `list.move(from, to)`: moves one element of the list from one index to another.
  * `list.swap(index1, index2)`: swaps the positions of the elements of the list at two indexes.
```

takameyer (approved):
Looks great :)

CHANGELOG.md
takameyer: Just to be consistant.  I think at some point we landed on "*" for bullets in the changelog.

PR #5400, opened 2023-02-09
[bindgen] Temporarily disable tests relying on non-queryable fields to throw

elle-j:
## What, How & Why?

BaaS was recently updated to now allow fields not explicitly listed as queryable to be queryable when development mode is enabled (the fields are automatically added to the queryable fields list).

Our tests were using non-queryable fields as a way to trigger a `SubscriptionsState.Error`, but until we have a solution for triggering such errors in our tests, they are now disabled in order to make the CI green.

This will be reenabled when #5402 is done.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-app-importer/src/AppImporter.ts
+++ b/packages/realm-app-importer/src/AppImporter.ts
@@ -54,9 +54,14 @@ type LoginResponse = {
 
 type AppConfig = {
   name: string;
+  sync?: SyncConfig;
   security?: {
     allowed_request_origins?: string[];
   };
+};
+
+type SyncConfig = {
+  development_mode_enabled?: boolean;
 };
 
 type ErrorResponse = {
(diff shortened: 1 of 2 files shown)


packages/realm-app-importer/src/AppImporter.ts
```diff
@@ -327,9 +332,8 @@ export class AppImporter {
     }
   }
 
-  private async enableDevelopmentMode(groupId: string, appId: string) {
+  private async applySyncConfig(groupId: string, appId: string, config: SyncConfig) {
```
kraenhansen, line 335: 🙏  thanks

kraenhansen (approved):

kneth:
@elle-j Should we target `master` instead of `bindgen`?

elle-j:
The same changes have already been applied to `master` 👍  (see #5401)

takameyer (approved):

kneth (approved):

PR #5413, opened 2023-02-10
support JSC versions without BigInt

RedBeard0531:
Mostly looking for review of the `Int64` API exposed in typescript.ts and implemented in node-wrapper.ts, along with the changes to the SDK to use that in place of direct operations on `bigint`s. Feel free to review the implementation in jsi.ts of course.

### DIFF
--- a/packages/bindgen/spec.yml
+++ b/packages/bindgen/spec.yml
@@ -64,6 +64,7 @@ typeAliases:
   IndexRange: std::pair<count_t, count_t>
   AppCredentialsToken: std::string
   Email: std::string
+  SchemaVersion: int64_t # is uint64_t in C++ but -1 is used as a sentinel.
 
 templates: # value tells you how many arguments it takes ('*' means any number)
   util::Optional: 1
@@ -372,7 +373,7 @@ records:
         default: false
       schema: util::Optional<std::vector<ObjectSchema>>
       schema_version:
-        type: uint64_t
+        type: SchemaVersion
         default: -1
       schema_mode:
         type: SchemaMode
@@ -680,8 +681,8 @@ classes:
 
   ObjectStore:
     staticMethods:
-      get_schema_version: '(group: Group) -> uint64_t'
-      set_schema_version: '(group: Group, version: uint64_t)'
+      get_schema_version: '(group: Group) -> SchemaVersion'
+      set_schema_version: '(group: Group, version: SchemaVersion)'
       #    verify_no_migration_required: '(changes: std::vector<SchemaChange>)'
       #needs_migration: '(changes: std::vector<SchemaChange>) -> bool'
       #verify_valid_additive_changes:
@@ -692,7 +693,7 @@ classes:
       #verify_compatible_for_immutable_and_readonly: '(changes: std::vector<SchemaChange>)'
       #verify_no_changes_required: '(changes: std::vector<SchemaChange>)'
 
-      # apply_schema_changes: '(group: Transaction&, uint64_t: schema_version, target_schema: Schema&, target_schema_version: uint64_t, mode: SchemaMode, changes: std::vector<SchemaChange>, migration_function: () -> void)'
+      # apply_schema_changes: '(group: Transaction&, schema_version: SchemaVersion, target_schema: Schema&, target_schema_version: SchemaVersion, mode: SchemaMode, changes: std::vector<SchemaChange>, migration_function: () -> void)'
       # apply_additive_changes: '(group: Group&, changes: std::vector<SchemaChanges>, update_indexes: bool)'
       # table_for_object_type
       # table_for_Schema_type
@@ -790,12 +791,12 @@ classes:
         #   sig: '(thread: ThreadSafeReference, scheduler: std::shared_ptr<Scheduler>) -> SharedRealm'
       get_synchronized_realm: '(config: RealmConfig) -> SharedAsyncOpenTask'
       #make_shared_realm: '(config: Realm::Config, version: util::Optional<VersionID>, coordinator: std::shared_ptr<_impl::RealmCoordinator>) -> SharedRealm'
-      get_schema_version: '(config: const RealmConfig&) const -> uint64_t'
+      get_schema_version: '(config: const RealmConfig&) const -> SchemaVersion'
 
     properties:
       config: const RealmConfig&
       schema: const std::vector<ObjectSchema>&
-      schema_version: uint64_t
+      schema_version: SchemaVersion
       is_in_transaction: bool
       is_frozen: bool
       is_in_migration: bool
@@ -826,7 +827,7 @@ classes:
       read_group: () -> Group&
       duplicate: () -> TransactionRef
 
-      update_schema: '(schema: std::vector<ObjectSchema>, version: uint64_t, migration_function: Nullable<std::function<(old_realm: SharedRealm, new_realm: SharedRealm, new_schema_handle: IgnoreArgument<Schema&>) -> void>>, initialization_function: Nullable<std::function<(realm: SharedRealm) -> void>>, in_transaction: bool) -> void'
+      update_schema: '(schema: std::vector<ObjectSchema>, version: SchemaVersion, migration_function: Nullable<std::function<(old_realm: SharedRealm, new_realm: SharedRealm, new_schema_handle: IgnoreArgument<Schema&>) -> void>>, initialization_function: Nullable<std::function<(realm: SharedRealm) -> void>>, in_transaction: bool) -> void'
       enable_wait_for_change: ()
       wait_for_change: () -> bool
       wait_for_change_release: ()
--- a/packages/bindgen/src/templates/jsi.ts
+++ b/packages/bindgen/src/templates/jsi.ts
@@ -75,7 +75,7 @@ function pushRet<T, U extends T>(arr: T[], elem: U) {
 class JsiAddon extends CppClass {
   exports: string[] = [];
   classes: string[] = [];
-  injectables = ["ArrayBuffer", "Float", "UUID", "ObjectId", "Decimal128", "EJSON_parse", "EJSON_stringify"];
+  injectables = ["Long", "ArrayBuffer", "Float", "UUID", "ObjectId", "Decimal128", "EJSON_parse", "EJSON_stringify"];
   mem_inits: CppMemInit[] = [];
 
   props = new Set<string>();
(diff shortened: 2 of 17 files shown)


packages/bindgen/spec.yml
```diff
@@ -680,8 +681,8 @@ classes:
 
   ObjectStore:
     staticMethods:
-      get_schema_version: '(group: Group) -> uint64_t'
-      set_schema_version: '(group: Group, version: uint64_t)'
+      get_schema_version: '(group: Group) -> SchemaVersion'
+      set_schema_version: '(group: Group, version: SchemaVersion)'
```
elle-j, line 685: Not specific to this line, but rather [line 696](https://github.com/realm/realm-js/blob/b3ce947bea5d4b4f29270b09c2b14add85fbe6cf/packages/bindgen/spec.yml#L696) which is commented out.

Could be good to update those as well right?
From (A):
```yml
# apply_schema_changes: '(group: Transaction&, uint64_t: schema_version, target_schema: Schema&, target_schema_version: uint64_t, mode: SchemaMode, changes: std::vector<SchemaChange>, migration_function: () -> void)'
```
To (B):
```yml
# apply_schema_changes: '(group: Transaction&, schema_version: SchemaVersion, target_schema: Schema&, target_schema_version: SchemaVersion, mode: SchemaMode, changes: std::vector<SchemaChange>, migration_function: () -> void)'
```

Note: Also fixed the structure of the 2nd param in (A).

packages/realm/src/assert.ts
```diff
@@ -71,14 +71,14 @@ assert.boolean = (value: unknown, target?: string): asserts value is boolean =>
...
+assert.bigInt = (value: unknown, target?: string): asserts value is binding.Int64 => {
+  if (binding.Int64.isInt(value)) {
     throw new TypeAssertionError("a bigint", value, target);
   }
 };
 
-assert.numberOrBigInt = (value: unknown, target?: string): asserts value is number => {
-  if (typeof value !== "number" && typeof value !== "bigint") {
+assert.numberOrBigInt = (value: unknown, target?: string): asserts value is number | binding.Int64 => {
+  if (typeof value !== "number" && binding.Int64.isInt(value)) {
```
elle-j, line 81:
```suggestion
  if (typeof value !== "number" && !binding.Int64.isInt(value)) {
```

elle-j (approved):
Very interesting! LGTM, just minor suggestions.

packages/bindgen/spec.yml
RedBeard0531: done. Although we will likely end up just deleting the whole `ObjectStore` class, since I don't think it is used at all.

packages/realm/src/assert.ts
RedBeard0531: D'oh! Luckily, it looks like neither `assert.bigInt` nor `assert.numberOrBigInt` are actually used anywhere...

RedBeard0531: I double checked all usages of `isInt`. Those were the only ones that were using `!==` before. The rest were correctly converted.

elle-j: Alright 👍  Yeah they probably weren't used 😛

packages/bindgen/src/templates/jsi.ts
```diff
@@ -946,6 +951,52 @@ class JsiCppDecls extends CppDecls {
...
+        body: `
+          if (${this.addon.get()}->m_on_hermes)
+              return jsi::BigInt::fromUint64(_env, val);
+
+          return ${this.addon.accessCtor("Long")}.callAsConstructor(_env, {
+              jsi::Value(int(val)), // low
+              jsi::Value(int(val >> 32)), // high
+              jsi::Value(true), // unsigned
+          });
+        `,
```
kraenhansen, line 998: If at all practical, I would love to see snippets like these maintained outside of a .ts file to get better editor support. I do see however it relies on values being injected into it, so that might not be practical here.

packages/realm/src/Realm.ts
```diff
@@ -147,7 +147,6 @@ type ObjectSchemaExtra = {
 
 // Using a set of weak refs to avoid prevention of garbage collection
 const RETURNED_REALMS = new Set<binding.WeakRef<binding.Realm>>();
-const NOT_VERSIONED = 18446744073709551615n;
```
kraenhansen, line 150: Nice to get rid of this 🙂

packages/realm/src/TypeHelpers.ts
```diff
@@ -30,22 +30,26 @@ import {
...
+    Int16Array,
+    Uint16Array,
+    Int32Array,
+    Uint32Array,
+    Float32Array,
+    Float64Array,
+    // These will not be present on old versions of JSC without BigInt support.
+    safeGlobalThis.BigInt64Array,
+    safeGlobalThis.BigUint64Array,
+  ].filter((ctor) => ctor !== undefined),
```
kraenhansen, line 51: Since they're added to a set, I wonder if it's even worth bothering to filter?

packages/realm/src/assert.ts
```diff
@@ -71,14 +71,14 @@ assert.boolean = (value: unknown, target?: string): asserts value is boolean =>
   }
 };
 
-assert.bigInt = (value: unknown, target?: string): asserts value is bigint => {
-  if (typeof value !== "bigint") {
+assert.bigInt = (value: unknown, target?: string): asserts value is binding.Int64 => {
```
kraenhansen, line 74: I wonder if it makes sense to update this name as well?
```suggestion
assert.int64 = (value: unknown, target?: string): asserts value is binding.Int64 => {
```

packages/realm/src/assert.ts
```diff
@@ -71,14 +71,14 @@ assert.boolean = (value: unknown, target?: string): asserts value is boolean =>
...
+  if (!binding.Int64.isInt(value)) {
     throw new TypeAssertionError("a bigint", value, target);
```
kraenhansen, line 76: Is this error message still correct in your opinion?

packages/realm/src/assert.ts
```diff
@@ -71,14 +71,14 @@ assert.boolean = (value: unknown, target?: string): asserts value is boolean =>
...
   }
 };
 
-assert.numberOrBigInt = (value: unknown, target?: string): asserts value is number => {
-  if (typeof value !== "number" && typeof value !== "bigint") {
+assert.numberOrBigInt = (value: unknown, target?: string): asserts value is number | binding.Int64 => {
+  if (typeof value !== "number" && !binding.Int64.isInt(value)) {
     throw new TypeAssertionError("a number or bigint", value, target);
```
kraenhansen, line 82: Same comment as ☝️

packages/realm/src/schema/normalize.ts
```diff
@@ -116,7 +116,7 @@ export function normalizeObjectSchema(arg: RealmObjectConstructor | ObjectSchema
 
   const { name, primaryKey, asymmetric, embedded, properties } = arg;
   assert(name.length > 0, "Invalid schema for unnamed object: 'name' must be specified.");
-  const primaryKeyFieldIsMissing = primaryKey && !Object.hasOwn(properties, primaryKey);
```
kraenhansen, line 119: Any particular reason for this change? Do you fear it won't be available on the older versions of JSC we support? If so, we should consider downgrading the `target` & `libs` of the ts-config.json.

kraenhansen (approved):

packages/bindgen/src/templates/jsi.ts
RedBeard0531: I initially started writing the first two in the jsi helpers header, but then moved them here when I realized that I wanted to use the `getPropId` mechanism to cache the `high` and `low` property constructions. I don't have a great answer here unfortunately. This seems like the least bad option, and isn't that bad in practice.

packages/realm/src/Realm.ts
RedBeard0531: Yeah. It is really just the unsigned 64-bit equivalent of -1 (same as the maximum uint64). I figured -1 makes a better sentinel value.

packages/realm/src/assert.ts
RedBeard0531: Hard to say since they aren't used for anything at the moment. I think it depends on how they end up being used. Maybe it would be best to just remove them for now, and we can re-add when needed?

packages/realm/src/schema/normalize.ts
RedBeard0531: Not just _fear_ that it won't be available. It _isn't_ available and tests were failing without this change.

I'm not sure what should be responsible for doing our polyfilling. It feels a bit like something that should be handled by metro bundler of the whole application when it knows the precise deployment target rather than just the library. In some cases we _do_ want to use newer features than are available everywhere as long as they are available wherever they are used.

packages/realm/src/TypeHelpers.ts
RedBeard0531: Without the filter, you'd have `undefined` in the set on JSC, which would be an issue in https://github.com/realm/realm-js/blob/99d3fee8338885e00b5e6cb8120191d468b61437/packages/realm/src/TypeHelpers.ts#L58-L61

packages/realm/src/schema/normalize.ts
RedBeard0531: Also, it seems like JSC is a low-priority target. So we may be OK with saying that some new features won't work there, although "core functionality" like this still must. I don't know if that is possible to achieve with the ts config, but I think as long as we get our CI to include the JSC versions we care about, I think we will be fine in practice either way.

packages/realm/src/assert.ts
kraenhansen: That'd be fine by me.

packages/realm/src/schema/normalize.ts
kraenhansen: It would probably make sense to spend some more time on this, after this PR is merge.
I agree its not a high priority if we're actually testing on the lowest supported JSC version, but again - I would also like to fail faster (ideally in the editor).

If we truly want to support older runtimes we should either
- transform our code (using babel?) and potentially produce a different bundle for old JSC targets (if those is practical to distinguish from newer ones) or
- make sure we're not using the newest and greatest features of the language (which would be a bummer and sort of the direction this change is taking).

RedBeard0531: > transform our code (using babel?) and potentially produce a different bundle for old JSC targets (if those is practical to distinguish from newer ones) 

Doesn't the metro bundler already do that? I think it knows if it knows if it is targeting hermes or JSC. Maybe we just need to publish a recommended config for users on JSC?

> make sure we're not using the newest and greatest features of the language (which would be a bummer and sort of the direction this change is taking).

I agree that this probably isn't the direction we want to take. But in this case the new feature is only slightly nicer than the old one, so it wasn't worth figuring out a general solution for. I'm not trying to set a precedent here, just trying to get us working on JSC. 😄

Issue #5410, opened 2023-02-10
Create Polyfill Convenience Modules


takameyer:
### Problem

In order for React Native developers to use the MongoDB `watch` method, they are required to perform the following setup to their project:
```
// Install the following dependencies
npm install react-native-polyfill-globals text-encoding;

// Add the following lines to index.js
import { polyfill as polyfillReadableStream } from "react-native-polyfill-globals/src/readable-stream";
polyfillReadableStream();

import { polyfill as polyfillEncoding } from "react-native-polyfill-globals/src/encoding";
polyfillEncoding();

import { polyfill as polyfillFetch } from "react-native-polyfill-globals/src/fetch";
polyfillFetch();
```

### Solution

Add `react-native-polyfil-globals` and `text-encoding` as optional dependencies to Realm.
Throw a warning when they are not present when trying to use the mongodb client.
Give instructions on how to setup. 
Make the setup as easy as:

```
import {realmGlobalPolyfills} from "Realm";
realmGlobalPolyfills();
```

### Alternatives

Just document this better and have the developer do the heavy lifting.

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

Atlas App Services: Auth or Functions etc

NoatToan:
Thanks @takameyer

PR #5417, opened 2023-02-10
Support `MongoDBCollection.watch()`

### LINKED ISSUE #5271: Remote MongoDB Client Watch

elle-j:
## What, How & Why?

This adds support for monitoring a MongoDB collection for changes.

An authenticated user must first retrieve a `MongoDBCollection`:

```ts
const collection = app
  .currentUser
  .mongoClient("myService")
  .db("myDb")
  .collection<Store>("stores");
```

An asynchronous change stream is returned by `collection.watch()`, yielding `ChangeEvent`s when the Atlas collection is updated:

```ts
for await (const event of collection.watch()) {
  doSomethingWithEvent();
}
```

Exports of MongoDB types:
* They exist on the `Realm.Services.MongoDB` namespace (as deprecated) to mirror the old API.
* They are also exported as named exports.

This closes #5271.

## ☑️ ToDos
* [x] 🚦 Tests


integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -0,0 +1,188 @@
...
+              if (event.fullDocument.isLast) break;
+              expect(event.fullDocument._id).to.equal(watchedId);
+              seenIt = true;
+            }
+            expect(seenIt).to.be.true;
+          })(),
+        ]);
+      });
+
+      it("throws when the user is logged out", async function (this: AppContext & Mocha.Context) {
```
elle-j, line 163: This test needs some TLC.

* `this.app.currentUser` seem to be `null` here.
* We're not hitting the `expect.fail()` (which we shouldn't), but we're also not entering the `catch` block (which we should).
* Need to consider how to handle potential tests added after this one if this test logs the user out.
* We do get a `401` auth error if e.g. trying to log in in the `finally` clause.

packages/bindgen/src/realm_js_helpers.h
```diff
@@ -249,6 +250,10 @@ struct Helpers {
 
         return query.get_description() + ' ' + results.get_descriptor_ordering().get_description(query.get_table());
     }
+
+    static void feed_buffer(app::WatchStream& ws, const BinaryData& buffer) {
+        ws.feed_buffer({ buffer.data(), buffer.size() });
+    }
```
elle-j, line 256: Temporary helper until Core can provide an overload for `feed_buffer` (it currently only takes a `std::string_view`).

packages/realm/src/Realm.ts
```diff
@@ -1467,12 +1531,78 @@ export declare namespace Realm {
   /** @deprecated Please use named imports */
   // eslint-disable-next-line @typescript-eslint/no-namespace
   export namespace Services {
-    // TODO: Fill in once the MongoDB client has stabilized
-
+    /** @deprecated Please use named imports */
+    export type MongoDB = MongoDBType;
```
elle-j, line 1535: `MongoDB` existed both as an interface and as a namespace here on the `Service` namespace.

packages/realm/src/Realm.ts
```diff
@@ -1467,12 +1531,78 @@ export declare namespace Realm {
...
+      /** @deprecated Please use named imports */
+      export type InsertEvent<T extends Document> = InsertEventType<T>;
+      /** @deprecated Please use named imports */
+      export type InsertManyResult<IdType> = InsertManyResultType<IdType>;
+      /** @deprecated Please use named imports */
+      export type InsertOneResult<IdType> = InsertOneResultType<IdType>;
+      /** @deprecated Please use named imports */
+      export type InvalidateEvent = InvalidateEventType;
+      /** @deprecated Please use named imports */
+      export type MongoDBCollection<T extends Document> = MongoDBCollectionType<T>;
```
elle-j, line 1586: Should this be exported as a `type` here?

packages/realm/src/app-services/MongoDBCollection.ts
```diff
@@ -88,15 +119,25 @@ type UpdateOptions = {
   readonly arrayFilters?: Filter[];
 };
 
+/**
+ * A document from a MongoDB collection
+ */
+export type Document<IdType = any> = {
```
elle-j, line 125: The previous API used `any` here, perhaps we could update this to `unknown`?

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -0,0 +1,188 @@
...
+    beforeEach(async function () {
+      await deleteAllDocuments();
+    });
+
+    after(async function () {
+      await deleteAllDocuments();
+    });
+
+    describe("#watch", function () {
+      const text = "use some odd chars to force weird encoding %\n\r\n\\????>>>>";
```
takameyer, line 87: This looks similar to the upcoming tests from the [test migration](https://github.com/realm/realm-js/blob/test-migration-master/integration-tests/tests/src/tests/sync/user.ts#L426).
We will have to determine what is being covered by both.  In any case, I like that the mongo-client test are seperate from the `user` tests (I would never find them there.)

integration-tests/tests/package.json
```diff
@@ -42,7 +42,8 @@
     "chai": "4.3.6",
     "chai-as-promised": "^7.1.1",
     "concurrently": "^6.0.2",
-    "jsrsasign": "^10.6.1"
+    "jsrsasign": "^10.6.1",
+    "react-native-polyfill-globals": "^3.1.0"
```
takameyer, line 46: Shouldn't this be added to `environments/react-native/package.json`?

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -0,0 +1,188 @@
...
+import { sleep } from "../../utils/sleep";
+
+type TestDocument = {
+  _id: number;
+  text?: string;
+  isLast?: boolean;
+};
+
+describe.skipIf(environment.missingServer, "MongoDB Client", function () {
+  this.timeout(60_000); // TODO: Temporarily hardcoded until envs are set up.
```
gagik, line 32: very nit this is more of a personal stylistic preference and I actually had no idea you could do this (seems like this was added in 2021 😮) but because the number is not large and this formatting is not used anywhere in the codebase else I'd prefer just `60000`

gagik (approved):
Looks good 🚀 ! Just a minor comment

integration-tests/tests/src/tests/sync/mongo-db-client.ts
RedBeard0531: Please don't make us count digits. If you *really* don't like `60_000`, then at least do `60*1000` which is the old way to write that before digit separators were allowed.

takameyer (approved):
Just a few considerations.  We might have to look into the database name problem though.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
gagik: yeah that's fair... I guess I just didn't like how `_` was chosen for this by the language(s) but then again `.` / `,` would have been a regional mess

integration-tests/environments/react-native/src/App.js
```diff
@@ -24,6 +24,9 @@
  * @flow
  */
 
+// Polyfill needed for for `fetch` and `ReadableStream`.
```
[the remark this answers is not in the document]
elle-j: Great, we can use the syntax you have in `test-migration-master` so that we only import exactly what we need.

For the description in #5410, we could add that this variant also is a possibility:
```js
import { polyfill } from "react-native-polyfill-globals";
polyfill();
```

But I do agree that simplifying this for users would be better!

integration-tests/realm-apps/with-db-flx/services/mongodb/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "mongodb",
+  "name": "mongo-db",
```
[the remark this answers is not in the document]
elle-j: It was so that we'd use the [default service name](https://www.mongodb.com/docs/atlas/app-services/functions/context/#std-label-context-services) for mongo clients. But, I see now that the default is actually `mongodb-atlas`. So if you think it's fine for us to change from `mongodb` I can go ahead and change to `mongodb-atlas`. This will also be the default argument we're using when users call `mongoClient()`.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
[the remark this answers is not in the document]
elle-j: I usually use it for numbers >= 10,000 :) I don't think we have `longTimeout` yet in bindgen?

integration-tests/tests/package.json
elle-j: Thank for pointing this out. Getting upstream dependency conflicts when moved there, but gonna look into this.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -0,0 +1,188 @@
...
+  importAppBefore("with-db");
+  authenticateUserBefore();
+
+  const serviceName = "mongo-db";
+  const dbName = "test-database";
```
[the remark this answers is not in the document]
elle-j: Yes the implemented tests all pass. Compared to the lines of code you linked to, the `importAppBefore()` on [bindgen](https://github.com/realm/realm-js/blob/a8ef5b587ff1c8628363c216e4932997c50f5d78/integration-tests/tests/src/hooks/import-app-before.ts#L36) use:
```ts
this.app = await importApp(name, replacements);
```

integration-tests/tests/src/tests/sync/mongo-db-client.ts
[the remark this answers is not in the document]
RedBeard0531: FWIW I use the same cutoff. My eyes can "sight count" up to 3 or 4 things, but with 5 my brain needs to get involved and actually start manually counting which is a lot slower. Given how common grouping digits into [2](https://en.wikipedia.org/wiki/Crore) or 3 places is across cultures, I'm guessing this is a fairly universal human limitation.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j: The `"can watch changes correctly"` test in the test migration should test the same thing as in this file. I slightly refactored to make it more readable (dividing it into separate `it` tests rather than one big one).

integration-tests/tests/src/tests/sync/mongo-db-client.ts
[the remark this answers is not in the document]
takameyer: Yeah!  I find it cool.  I just had to quickly do some googling to figure what I was looking at.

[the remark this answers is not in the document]
takameyer: And now i've learned what a Crore is.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
takameyer: FWIW, I find these tests much more readable.  We placed a priority on functionality over readability in the test migration :-(   
So lets keep them both and replace later.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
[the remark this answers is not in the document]
takameyer: Inside the `importApp` function, you will see some logic for [generating a database name](https://github.com/realm/realm-js/blob/test-migration-master/integration-tests/tests/src/utils/import-app.ts#L99).  I did some refactoring to this code in the `test-migration-master` so that the generated db-name was available in `this.database`.  We should probably get those upstreamed sooner than later.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
kneth: We have the same challenge with Realm Web: https://github.com/realm/realm-js/blob/master/packages/realm-web-integration-tests/src/remote-mongodb-service.test.ts#L417

Hopefully @kraenhansen remembers why

packages/bindgen/src/realm_js_helpers.h
kneth: Is there a core issue for it? If so, a link would remind us to check from time to time. If not, please create one.

kneth (approved):

integration-tests/tests/src/tests/sync/mongo-db-client.ts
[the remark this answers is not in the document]
elle-j: Yeah that'd be great!

packages/bindgen/src/realm_js_helpers.h
elle-j: @RedBeard0531 will create one once he goes through all our helpers here to see if anything else could be simplified from Core's side as well.

RedBeard0531 (approved):

packages/realm/src/app-services/User.ts
```diff
@@ -288,26 +296,30 @@ export class User<
...
   /**
-   * Returns a connection to the MongoDB service.
+   * @param serviceName The name of the MongoDB service to connect to. (Default: "mongo-db")
+   * @returns A connection to the MongoDB service.
    *
    * @example
-   * let blueWidgets = user.mongoClient('myClusterName')
+   * let blueWidgets = user.mongoClient()
    *                       .db('myDb')
    *                       .collection('widgets')
```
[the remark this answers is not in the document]
elle-j: That could be good, thanks. Because otherwise the document type will be `Document<any>`, and showing users that they can specify it is probably helpful.

integration-tests/realm-apps/with-db-flx/services/mongodb/config.json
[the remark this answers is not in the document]
kraenhansen: > the default is actually `mongodb-atlas` -- good find 🤔 I do remember seeing that before and I wonder why that was not the default on realm-web then.

packages/realm/src/Realm.ts
```diff
@@ -1245,6 +1276,39 @@ type ApiKeyType = ApiKey;
...
+type MongoDBDatabaseType = MongoDBDatabase;
+type NewDocumentType<T extends Document> = NewDocument<T>;
+type OperationTypeType = OperationType;
+type RenameEventType = RenameEvent;
+type ReplaceEventType<T extends Document> = ReplaceEvent<T>;
+type UpdateType = Update;
+type UpdateDescriptionType = UpdateDescription;
+type UpdateEventType<T extends Document> = UpdateEvent<T>;
+type UpdateOptionsType = UpdateOptions;
+type UpdateResultType<IdType> = UpdateResult<IdType>;
```
kraenhansen, line 1310: Unfortunately, these need their type argument defaults too.

packages/realm/src/Realm.ts
```diff
@@ -1483,12 +1547,78 @@ export declare namespace Realm {
...
+      /** @deprecated Please use named imports */
+      export type Update = UpdateType;
+      /** @deprecated Please use named imports */
+      export type UpdateDescription = UpdateDescriptionType;
+      /** @deprecated Please use named imports */
+      export type UpdateEvent<T extends Document> = UpdateEventType<T>;
+      /** @deprecated Please use named imports */
+      export type UpdateOptions = UpdateOptionsType;
+      /** @deprecated Please use named imports */
+      export type UpdateResult<IdType> = UpdateResultType<IdType>;
```
kraenhansen, line 1620: These might their type argument defaults too.

elle-j: Are you saying we should introduce defaults to all type arguments? (I used what the "old" types were using)

JacobOscarGunnarsson (approved):
LGTM

packages/realm/src/Realm.ts
kraenhansen: No sorry. I mis remembered for some reason 👍

integration-tests/realm-apps/with-db-flx/services/mongodb/config.json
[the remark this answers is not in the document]
elle-j: Update: I've reverted this back to `mongodb` and also changed `mongoClient()` to *not* have a default argument, but rather let the user be explicit about which service to connect to.

> ..the default on realm-web

P.S. Even though realm-web's internal helper function has a default argument, it doesn't seem like the public facing method actually documents that a default argument exists. Not in the docs and not by the TS type:
```ts
// Public facing
public mongoClient(serviceName: string): Realm.Services.MongoDB {
  return createMongoDBRemoteService(this.fetcher, serviceName);
}

// Helper
export function createService(fetcher: Fetcher, serviceName = "mongo-db"): MongoDBService {
  return { db: createDatabase.bind(null, fetcher, serviceName) };
}
```

kraenhansen (approved):

PR #5423, opened 2023-02-13
[bindgen] Refactored network to rely on `@realm/network-transport`

kraenhansen:
## What, How & Why?

We needed a more featureful network transport implementation. Specifically, the MongoDB `watch` method needs to read the response body as an iterable `ReadableStream`.

This refactors the platform specific network transport to simply use the existing `@realm/network-transport` package, that we know already works for the legacy SDK.


integration-tests/tests/package.json
```diff
@@ -2,17 +2,18 @@
...
     "./node": "./src/node/index.ts"
   },
   "private": true,
   "scripts": {
-    "start": "CONTEXT=appTemplatesPath=../realm-apps,$CONTEXT node start.js --watch",
+    "start": "CONTEXT=appTemplatesPath=../realm-apps,$CONTEXT node start.mjs --watch",
     "start:missing-server": "CONTEXT=missingServer,$CONTEXT npm start",
     "start:performance": "CONTEXT=missingServer,performance,integration=false,$CONTEXT npm start",
-    "test": "CONTEXT=appTemplatesPath=../realm-apps,$CONTEXT node start.js --exit",
+    "test": "CONTEXT=appTemplatesPath=../realm-apps,$CONTEXT node start.mjs --exit",
```
kraenhansen, line 16: I had to switch the `tests` package to a module to make it load the right `@realm/network-transport` bundle. (at least I think that was why).

packages/realm-network-transport/package.json
```diff
@@ -2,13 +2,13 @@
...
+  "module": "./dist/bundle.mjs",
   "types": "./dist/bundle.d.ts",
-  "react-native": "./dist/bundle.react-native.es.js",
+  "react-native": "./dist/bundle.react-native.mjs",
   "browser": {
-    "./dist/bundle.cjs.js": "./dist/bundle.dom.cjs.js",
-    "./dist/bundle.es.js": "./dist/bundle.dom.es.js",
+    "./dist/bundle.js": "./dist/bundle.dom.js",
+    "./dist/bundle.mjs": "./dist/bundle.dom.mjs",
     "./dist/bundle.d.ts": "./dist/bundle.dom.d.ts"
```
kraenhansen, line 12: These are the file extensions we use in other packages.

packages/realm-network-transport/src/IterableReadableStream.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This method ensures that a ReadableStream is iterable, wraps in a proxy with the {@link Symbol.asyncIterator} if missing.
+ * @param response
+ * @returns An async iterator.
+ */
+export function makeStreamIterable(stream: ReadableStream): ReadableStream {
+  if (Symbol.asyncIterator in stream) {
+    return stream;
+  } else if ("getReader" in stream) {
+    return new Proxy(stream, READABLE_STREAM_HANDLER);
```
kraenhansen, line 50: This is to gently polyfill a missing `asyncIterator` on the `ReadableStream`. We experienced with the legacy SDK that this might not always be set.

packages/realm/rollup.config.mjs
```diff
@@ -53,7 +57,7 @@ export default [
         tsconfig: "src/node/tsconfig.json",
       }),
     ],
-    external: ["bson", "debug", "node:module", "node:fs", "node:path", "node:http", "node:https"],
+    external: ["bson", "debug", "node-fetch", "node:module", "node:fs", "node:path"],
```
kraenhansen, line 60: We're bundling in the `@realm/network-transport`, so it won't be a dependency of the package, with the risk of mismatching versions.

packages/realm/rollup.config.mjs
```diff
@@ -64,7 +68,8 @@ export default [
     },
     plugins: [
       nodeResolve({
-        resolveOnly: ["path-browserify"],
+        mainFields: ["react-native", "module", "main"],
```
kraenhansen, line 71: Ensuring the `react-native` main field is used when importing the `@realm/network-transport` package.

elle-j (approved):
LGTM! Makes `watch()` work like a charm.

packages/realm/src/app-services/MutableSubscriptionSet.ts
```diff
@@ -104,7 +104,7 @@ export class MutableSubscriptionSet extends BaseSubscriptionSet {
    * @param query A {@link Results} instance representing the query to remove a subscription to.
    * @returns `true` if the subscription was removed, `false` if it was not found.
    */
-  remove(query: Results<unknown>): boolean {
+  remove(query: Results<any>): boolean {
```
takameyer, line 107: Why change this to `any`?

kraenhansen: It should have probably been a part of a separate PR.

This is the error I was seeing in the flexible sync integration tests:

Screenshot 2023-02-13 at 12 49 08

kraenhansen: We have quite a few TS issues in the flexible sync tests, so perhaps we should add a task to go over those and figure out the root cause of this in the process?

takameyer: Yeah, let's make an issue for it 👍🏼

PR #5431, opened 2023-02-14
Enforce PascalCase for generated enum constants

### LINKED ISSUE #5373: Enforce PascalCase convention for BindGen-generated enum constants
*(This is only an internal change, not affecting any public facing APIs)*

Currently, methods and record members emitted by the binding generator is transformed into camelCase (e.g. `get_name()` -> `getName()`) to avoid mental overhead. However, generated enum constants use inconsistent naming conventions.

(1) Enforce generated enum constants to use PascalCase (as this is the most common convention used in the realm package).

Generated enums which are not using PascalCase (see `packages/realm/generated/ts/core.ts`):
* **AuthProvider** (e.g. `ANONYMOUS`, `FACEBOOK`)
* **LoggerLevel** (e.g. `all`, `trace`)
* **HttpMethod** (e.g. `get`, `post`)
* **ProgressDirection** (e.g. `upload`, `download`)
* **WatchStreamState** (e.g. `NEED_DATA`, `HAVE_EVENT`)

(2) Add and apply an ESLint rule to enforce the conventions:
* Generated methods and record members: camelCase
* Generated enum constants: PascalCase

elle-j:
## What, How & Why?

Enforces enum constants generated by the Binding Generator to use PascalCase naming convention.

This is only an internal change (types available on our `binding` export).

This closes #5373.

## ☑️ ToDos
* [x] 🚦 Run tests

### DIFF
--- a/packages/bindgen/src/bound-model.ts
+++ b/packages/bindgen/src/bound-model.ts
@@ -621,8 +621,7 @@ export function bindModel(spec: Spec): BoundSpec {
     struct.fields = Object.entries(fields).map(([name, field]) => {
       const type = resolveTypes(field.type);
       // Optional and Nullable fields are never required.
-      const required =
-        field.default === undefined && !(type.isNullable() || type.isOptional());
+      const required = field.default === undefined && !(type.isNullable() || type.isOptional());
       return new Field(name, field.cppName ?? name, type, required, field.default);
     });
   }
--- a/packages/realm/src/schema/from-binding.ts
+++ b/packages/realm/src/schema/from-binding.ts
@@ -36,7 +36,7 @@ const TYPE_MAPPINGS: Record<binding.PropertyType, PropertyTypeName | null> = {
   [PropertyType.Mixed]: "mixed",
   [PropertyType.ObjectId]: "objectId",
   [PropertyType.Decimal]: "decimal128",
-  [PropertyType.UUID]: "uuid",
+  [PropertyType.Uuid]: "uuid",
   [PropertyType.Array]: "list",
   [PropertyType.Set]: "set",
   [PropertyType.Dictionary]: "dictionary",
--- a/packages/bindgen/src/js-passes.ts
+++ b/packages/bindgen/src/js-passes.ts
@@ -18,6 +18,7 @@
 import {
   BoundSpec,
   Class,
+  Enumerator,
   Field,
   InstanceMethod,
   Method,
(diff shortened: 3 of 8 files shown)


packages/bindgen/src/bound-model.ts
```diff
@@ -621,8 +621,7 @@ export function bindModel(spec: Spec): BoundSpec {
     struct.fields = Object.entries(fields).map(([name, field]) => {
       const type = resolveTypes(field.type);
       // Optional and Nullable fields are never required.
-      const required =
-        field.default === undefined && !(type.isNullable() || type.isOptional());
+      const required = field.default === undefined && !(type.isNullable() || type.isOptional());
```
elle-j, line 624: This is merely a lint fix.

kraenhansen (approved):

elle-j:
It was discussed in last week's round table and we thought it'd be good to align this with the bindgen goal of "avoiding mental overload" by using a common naming convention 👍

kneth (approved):

packages/realm/src/schema/from-binding.ts
```diff
@@ -36,7 +36,7 @@ const TYPE_MAPPINGS: Record<binding.PropertyType, PropertyTypeName | null> = {
   [PropertyType.Mixed]: "mixed",
   [PropertyType.ObjectId]: "objectId",
   [PropertyType.Decimal]: "decimal128",
-  [PropertyType.UUID]: "uuid",
+  [PropertyType.Uuid]: "uuid",
```
gagik, line 39: This makes sense to me but it is worth noting that the BSON library refers to `UUID` as `UUID` and provides both `ObjectId` and `ObjectID` as aliases. Not terribly important but seems a little counterintuitive to diverge from it in our internal naming.

takameyer (approved):
Looks great, just perhaps we should decide if we should make an exception for `UUID`

packages/realm/src/schema/from-binding.ts
elle-j: Good point 👍  Yeah I'd be interested to hear what the team thinks :)

gagik: @realm/realm-js-team good opportunity to test GitHub teams

RedBeard0531: This is an enumerator, not the type. I don't think it is a good idea to have one-of exceptions like that in code generated from the spec. Either we should say we want to use a rule to convert all names to a standard style, or we should just use the names as they are defined in the spec.

gagik (approved):

JacobOscarGunnarsson (approved):

PR #5436, opened 2023-02-15
Fix "Invalid Realm passed to bind_to_context()" when clearing test state

kraenhansen:
## What, How & Why?

This fix an issue provoked when an async open task resolved after `Realm.clearTestState` had been called, which would result in a termination of the process, yielding:

> Invalid Realm passed to bind_to_context()

This happened a lot in our integration tests, when a test awaiting a `Realm.open` call would timeout and `Realm.clearTestState` would be called from an after hook.

### DIFF
--- a/integration-tests/tests/src/tests/sync/open.ts
+++ b/integration-tests/tests/src/tests/sync/open.ts
@@ -1,0 +1,38 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { expect } from "chai";
+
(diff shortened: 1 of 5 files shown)


src/js_realm.hpp
```diff
@@ -30,8 +30,11 @@
 #include "js_observable.hpp"
 #include "platform.hpp"
 #include "realm/binary_data.hpp"
+#include "realm/util/functional.hpp"
+#include <exception>
```
kraenhansen, line 34: I'm not 100% sure this is strictly needed. Perhaps it brings `std::make_exception_ptr`?

kneth (approved):

kraenhansen:
Do I want to fix the legacy tests? 🤔

src/js_realm.hpp
RedBeard0531: Yup https://en.cppreference.com/w/cpp/error/make_exception_ptr

It was probably added automatically by clangd. It isn't strictly necessary since I'm sure one of the other headers transitively includes it, but it is generally considered a good habit to include what you use.

kneth:
@kraenhansen 

> Do I want to fix the legacy tests?

We are almost done with the migration, and I think you should simply disable those tests.

integration-tests/tests/src/tests/sync/open.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { expect } from "chai";
+
+import { authenticateUserBefore, importAppBefore } from "../../hooks";
+
+describe("Realm.open on a sync Realm", () => {
+  importAppBefore("with-db-flx");
+  authenticateUserBefore();
+
+  it("fails gracefully when canceled while opening", async function (this: UserContext) {
```
JacobOscarGunnarsson, line 27: These tests seem to fail consistently when I run electron tests locally. I'm not sure but could it be a race condition between the realm opening and the `cancel()` method being invoked? :thinking: 

image

PR #5437, opened 2023-02-15
Moved `closeThisRealm` and simplified `closeRealm`

kraenhansen:
## What, How & Why?

- When the opening of a Realm fails the after hook would also fail, which I think isn't ideal. Instead we should consider a `realm` missing from the context a way of signaling that the Realm was never opened or already closed.
- I don't think the utils were the right place for `closeThisRealm` as it required `this` being a Mocha context and was only used by the "open-realm-before.ts". So I simply moved it into the calling source file.
- I also changed the `closeRealm` to not have to take a config and instead use the path of the Realm to construct a mocked config pointing to the right file (even in the case of synced realms).

### DIFF
--- a/integration-tests/tests/src/tests/sync/flexible.ts
+++ b/integration-tests/tests/src/tests/sync/flexible.ts
@@ -268,9 +268,7 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
 
           async function openRealm(user: Realm.User, extraConfig: ExtraConfig = {}) {
             const config = getSuccessConfig(user, extraConfig);
-            const realm = await Realm.open(config);
-
-            return { realm, config };
+            return Realm.open(config);
           }
 
           async function testSuccess(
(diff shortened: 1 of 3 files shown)


integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -268,37 +268,34 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
 
           async function openRealm(user: Realm.User, extraConfig: ExtraConfig = {}) {
             const config = getSuccessConfig(user, extraConfig);
-            const realm = await Realm.open(config);
-
-            return { realm, config };
+            return Realm.open(config);
```
JacobOscarGunnarsson, line 271: could you elaborate to me why `await` no longer is needed for opening this realm? Should the function still be async if that's no longer the case?

JacobOscarGunnarsson (approved):
LGTM

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: i'll chime in.  The `await` here was redundant.  The code consuming this function assumes a promise anyway and would `await` the result.

takameyer (approved):
Great stuff!  LGTM :100:

integration-tests/tests/src/tests/sync/flexible.ts
JacobOscarGunnarsson: I see, appreciate the explanation! :D

PR #5443, opened 2023-02-16
Update `useUser` to never return `null`

### LINKED ISSUE #4973: The implementation of `useUser` should never return `undefined`
The `useUser` hook would return `undefined` if the user isn't logged in.  This should actually never be `undefined`, since it isn't possible to render the application without a logged in user (at least when using sync).
Since it is typed to sometimes be `undefined`, the underlying code must always check if it isn't, which isn't very clean.  The `UserProvider` and `useUser` should be refactored to always return a `user`.

takameyer:
## What, How & Why?
Remove need for optional chaining when accessing methods and properties on `user`.

This closes #4973

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/example/app/AppSync.tsx
+++ b/example/app/AppSync.tsx
@@ -26,7 +26,7 @@ export const AppSync: React.FC = () => {
   }, [realm, result]);
 
   const handleLogout = useCallback(() => {
-    user?.logOut();
+    user.logOut();
   }, [user]);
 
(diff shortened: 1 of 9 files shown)


elle-j (approved):
Nice :)

PR #5447, opened 2023-02-16
[bindgen] Adding `prebuild` and configuring it

kraenhansen:
## What, How & Why?

This adds prebuild of the Node.js native module for separate architectures.

For context, I've included an excerpt from an internal discussion we had on Slack:

---

I just learned that `uv.h` is not covered by the ABI stability of node-api: https://nodejs.org/dist/latest-v18.x/docs/api/n-api.html#implications-of-abi-stability
In our current SDK we use a pre v7 version of cmake-js, which incorrectly assumes we're building for a specific node version and incorrectly expose uv.h to object-store and risking crashes.
I noticed this when I upgraded to cmake-js v7.1.1 on master where it correctly detects that we want to compile for node-api ([because of this check](https://github.com/cmake-js/cmake-js/commit/1c78b2a235dac33034471ce1a69f8f5aabcd80de#diff-08774a7d53fdb8a05172d1228432d364ac1abb093df05efcfacaa713e321438aR17)) and I start getting:

```
object-store/util/uv/scheduler.hpp:24:10: fatal error: 'uv.h' file not found
```

I believe this means that we're risking crashes of our current native binary, because we're depending on the ABI of libuv, which we can't assume is stable.

I see the following options:
- Implement a scheduler on top of [the node-api primitives for async operations](https://github.com/nodejs/node-addon-api/blob/main/doc/async_operations.md) - if that's even possible? @fealebenpae suggested making a scheduler using the `Napi::ThreadSafeFunction` API.
- Include our own copy of libuv and somehow integrate it with that provided by node.js (but the integration code will be exposed to the same risks if prebuilt).
- Avoid upgrading cmake-js or build for node (instead of napi) and cross our fingers and hope it won't break in future versions of node :crossed_fingers:

---

This PR takes the latter approach by avoiding to include the

### DIFF
--- a/packages/realm/package.json
+++ b/packages/realm/package.json
@@ -34,11 +34,14 @@
     "test": "mocha",
     "test:types": "tsc --project type-tests/tsconfig.json",
     "build": "rollup -c",
-    "build:watch": "rollup -c --watch"
+    "build:watch": "rollup -c --watch",
+    "prebuild": "prebuild --runtime node -- --directory ../bindgen",
+    "install": "prebuild-install --runtime node || echo 'Failed to download prebuild for Realm'"
   },
   "dependencies": {
     "debug": "^4.3.4",
-    "node-fetch": "^3.3.0"
+    "node-fetch": "^3.3.0",
+    "prebuild-install": "^7.1.1"
   },
   "peerDependencies": {
     "bson": "^4",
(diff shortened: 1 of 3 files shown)


takameyer (approved):
Looks good.  Just curious how this avoids importing `uv.h`

packages/realm/package.json
```diff
@@ -71,5 +76,12 @@
     "type": "git",
     "url": "https://github.com/realm/realm-js.git",
     "directory": "packages/realm"
+  },
+  "binary": {
+    "module_name": "realm",
+    "module_path": "generated/ts",
+    "host": "https://static.realm.io",
+    "package_name": "realm-v{version}-napi-v6-{platform}-{arch}.tar.gz",
+    "remote_path": "realm-js-prebuilds/{version}"
```
kraenhansen, line 85: Because we're leaving out the `api_versions` from this object, we signal to cmake-js that we're not targeting napi explicitly, which ensures we get the uv.h in our include path.

kraenhansen: > Just curious how this avoids importing uv.h
@takameyer this is where the magic happens 🪄

takameyer: Thank you  🤩

PR #5451, opened 2023-02-18
fix remainder of failing integration tests

JacobOscarGunnarsson:
This PR fixes:

- checks for correct error messages in tests depending on platform
- moves invalid string indexing on results and lists to separate tests. They're also more elaborated on the expected behaviour
- adds a suggestion to change the `select` helper method
- updates readme for setting up react-native local tests

### DIFF
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -522,6 +522,24 @@ describe("Lists", () => {
   });
   describe("subscripts", () => {
     openRealmBeforeEach({ schema: [LinkTypeSchema, TestObjectSchema, PrimitiveArraysSchema] });
+    it("invalid object access yield correct values", function (this: RealmContext) {
+      this.realm.write(() => {
+        const obj = this.realm.create<ILinkTypeSchema>("LinkTypesObject", {
+          objectCol: { doubleCol: 1 },
+          objectCol1: { doubleCol: 2 },
+          arrayCol: [{ doubleCol: 3 }],
+        });
+
+        //React native returns an empty object upon invalid indexing.
+        if (environment.reactNative) {
+          //@ts-expect-error TYPEBUG: indexing by string on results is not allowed typewise
+          expect(Object.keys(obj?.arrayCol[""]).length).equals(0);
+        } else {
+          //@ts-expect-error TYPEBUG: indexing by string on results is not allowed typewise
+          expect(obj?.arrayCol[""]).to.be.undefined;
+        }
+      });
+    });
     it("support getters", function (this: RealmContext) {
       let obj!: ILinkTypeSchema;
       let prim!: PrimitiveArrays;
@@ -564,10 +582,6 @@ describe("Lists", () => {
       expect(obj?.arrayCol[1].doubleCol).equals(4);
       expect(obj?.arrayCol[2]).equals(undefined, "1");
       expect(obj?.arrayCol[-1]).equals(undefined, "2");
-      //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-      expect(obj?.arrayCol[""] === undefined).to.be.true;
-      //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-      expect(obj?.arrayCol["foo"] === undefined).to.be.true;
 
       expect(obj.arrayCol1[0].doubleCol).equals(5);
       expect(obj.arrayCol1[1].doubleCol).equals(6);
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -565,9 +565,9 @@ describe("Lists", () => {
       expect(obj?.arrayCol[2]).equals(undefined, "1");
       expect(obj?.arrayCol[-1]).equals(undefined, "2");
       //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-      expect(obj?.arrayCol[""]).equals(select({ reactNative: {}, default: undefined }), "3");
+      expect(obj?.arrayCol[""] === undefined).to.be.true;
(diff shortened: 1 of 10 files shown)


integration-tests/tests/src/utils/select.ts
```diff
@@ -24,14 +24,14 @@ type PlatformValues<T> = Partial<Record<keyof KnownEnvironment, T>> & {
  * @param platformValues containing value for each targeted platform.
  * @returns the value for the platform the test is running on.
  */
```
JacobOscarGunnarsson, line 26: this change may not be necessary, I realised when doing some testing that if `default == undefined` this helper function won't work. But then I'd like ism input on this part as having `any` as return type is not necessarily the best..

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -353,7 +354,9 @@ describe("SessionTest", () => {
     afterEach(() => Realm.clearTestState());
     it("can set custom logging function", async function (this: AppContext) {
       // setting a custom logging function must be done immediately after instantiating an app
-      const app = new Realm.App(this.app.id);
```
JacobOscarGunnarsson, line 356: if I remember this failed in electron saying that the app could not be found

integration-tests/tests/src/tests/list.ts
```diff
@@ -521,6 +522,24 @@ describe("Lists", () => {
...
+        const obj = this.realm.create<ILinkTypeSchema>("LinkTypesObject", {
+          objectCol: { doubleCol: 1 },
+          objectCol1: { doubleCol: 2 },
+          arrayCol: [{ doubleCol: 3 }],
+        });
+
+        //React native returns an empty object upon invalid indexing.
+        if (environment.reactNative) {
+          //@ts-expect-error TYPEBUG: indexing by string on results is not allowed typewise
+          expect(Object.keys(obj?.arrayCol[""]).length).equals(0);
```
kraenhansen, line 536: In my opinion:
- the long term solution is to figure out why this is happening. This behavior shouldn't be platform dependent.
- the short term solution could be creating an issue to track this and disable the test for now, instead of encoding an unexpected behavior as a success.

integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -108,9 +108,10 @@ describe("App", () => {
       const credentials = Realm.Credentials.anonymous();
       await expect(app.logIn(credentials)).to.be.rejectedWith(
         select({
-          reactNative: "Network request failed: Could not connect to the server",
-          default:
+          reactNative: new RegExp("Network request failed.*connect"),
+          default: new RegExp(
```
kraenhansen, line 112: Why are there not simple strings? If they must be regexp consider using the `/ /` syntax.

integration-tests/tests/src/tests/sync/open-behavior.ts
```diff
@@ -541,7 +541,8 @@ describe("OpenBehaviour", function () {
       await openPromise2;
       throw new Error("openPromise2 should have been rejected..");
     } catch (err: any) {
-      expect(err.message).equals("Operation canceled");
+      //platforms either return "Operation canceled" or "Operation Canceled"
+      expect((err.message as string).toLowerCase()).equals("operation canceled");
```
kraenhansen, line 545: Would it be easy to find the place this differs and solve it or is this thrown by the runtime?

integration-tests/tests/src/utils/select.ts
```diff
@@ -24,14 +24,14 @@ type PlatformValues<T> = Partial<Record<keyof KnownEnvironment, T>> & {
...
-export function select<T>(platformValues: PlatformValues<T>): T {
+export function select<T>(platformValues: PlatformValues<T>): any {
```
kraenhansen, line 27: Why this change? It'll be harder to determine a type mismatch in the calling code.

integration-tests/tests/src/utils/select.ts
gagik: What do you mean by `default == undefined`? I am unsure how this affects `select`'s type. If this is about the automatic derivation of the type from given values, you can manually specify i.e. `select<String | undefined>({default: "test"}})` so there is no need for `any` as a return type.

integration-tests/tests/src/tests/sync/open-behavior.ts
JacobOscarGunnarsson: Can't find this anywhere in the project so my assumption is that it's thrown by the runtime, especially since the message varies

integration-tests/tests/src/tests/sync/app.ts
JacobOscarGunnarsson: The reactNative one must be a regex. I skimmed through the documentation and thought that the `new RegExp` was the easiest readable declaration, but this is just based on my flavour so I can go with `/ /` as well

gagik: Why does it need to be regex? Could it not just be `"Network request failed"`?

JacobOscarGunnarsson: yeah fair, thought it was too generic at first, but in hindsight, my regex doesn't provide much more detailed information

takameyer (approved):
LGTM

integration-tests/tests/src/tests/sync/open-behavior.ts
takameyer: I was able to find some generic error codes within the `boost` library in the `ios/Pods`, which contained an "Operation canceled" string.  Seems the message is out of our hands and is indeed platform specific.

integration-tests/tests/src/tests/list.ts
kraenhansen: @takameyer was this addressed?

kraenhansen: Never mind, I updated my UI and found the change I needed to see 👍😅

PR #5452, opened 2023-02-18
fix: optional-chain node release and version

manniL:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-web/src/node/index.ts
+++ b/packages/realm-web/src/node/index.ts
@@ -31,8 +31,8 @@ const environment: Environment = {
     return null;
   },
 
-  platform: process.release.name || "node",
-  platformVersion: process.versions.node,
+  platform: process?.release?.name || "node",
+  platformVersion: process?.versions?.node || 'unknown',
 
   TextDecoder,
 };


manniL:
@cla-bot check

kneth:
@manniL Thank you for your contribution.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 * Fixed diverging history in flexible sync if writes occur during bootstrap to objects that just came into view. ([realm/realm-core#5804](https://github.com/realm/realm-core/issues/5804), since v10.11.0)
 * If a client reset with recovery or discard local is interrupted while the "fresh" Realm is being downloaded, the sync client may crash with a `MultipleSyncAgents` exception ([realm/realm-core#6217](https://github.com/realm/realm-core/issues/6217), since v10.15.0)
 * Online compaction may cause a single commit to take a long time. ([realm/realm-core#6245](https://github.com/realm/realm-core/pull/6245), since v11.3.0-rc.0)
+* Use optional chaining to infer Node release and version to avoid crashes in non-node environments
```
kneth, line 25: The entry should go into `packages/realm-web/CHANGELOG.md`

kneth (approved):
I have only a small wish: adding the change log entry to Realm Web's change log.

kraenhansen:
We want to be able to call any valid Node.js API in the `packages/realm-web/src/node/index.ts` file and a separate entry point for CloudFlare workers is probably a better solution to this.

I would be interesting to learn if the CloudFlare workers have their own "main field" / [conditional definition](https://nodejs.org/api/packages.html#community-conditions-definitions) that they're looking for. It doesn't seem so from [the first result I got](https://developers.cloudflare.com/workers/wrangler-legacy/webpack/#sensible-defaults) Googling this.

CHANGELOG.md
manniL: Will fix ☺️

manniL:
@kraenhansen Good comment, though having a "universally working" version would be even more ideal IMO (even though might not be possible).

> As such, I suggest we refactor this to produce a bundle for the workerd environment instead of modifying the exiting node bundle.

How would next steps look like?

PR #5461, opened 2023-02-20
allow useQuery to apply filter before listener is added

levipro:
## What, How & Why?
I am submitting this change for review because I would like to apply a filter before the listener for `useQuery` is applied. In some instances (ie: your filter doesn't change) it makes sense to apply a filter and have that be a requirements for a new render.

In the current implementation you can apply a filter in a `useMemo` call but even if your filter is constant you will still get re-renders even when objects outside the filter are changed or added.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -19,6 +19,7 @@ const SomeComponent = () => {
 }
 ```
 >NOTE: If your app is using multiple Realms, then you should continue using `createRealmContext`
+* `useQuery` can now have a filter applied within the hook. Use this when your filter does not change across re-renders. Applying the filter within the hook means that only changes found on filtered objects will trigger a re-render.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


takameyer:
Hi @levipro, thanks for the PR!

We have been giving some thought about `useQuery`.  It was originally thought to have an object argument that would be something like: `{filter: { filterString, filterArguments}, sorted: "propertyToSort"}}`.  This just didn't feel correct and was also limiting functionality.

What I would like to implement would be something like:
```
 const tasks = useQuery("Task", (tasks) => {
   return tasks.filtered(“ownerId == $0”, userId);
 }, [userId]);
 ```
 
 This would also allow for the filter to be updated and redone if any of the dependencies change.  It also allows for calling `sorted`.  Any thoughts?

levipro:
@takameyer I like that, I have implemented that in #5465

Let me know what you think there.

takameyer:
Closing this in favor of https://github.com/realm/realm-js/pull/5465

PR #5465, opened 2023-02-20
allow useQuery to filter or sort a collection by using a callback

levipro:
## What, How & Why?
allow useQuery to filter or sort a collection by using a callback mehod that gets momoized based on the dependency array

This fixes: #5471

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/useQuery.tsx
+++ b/packages/realm-react/src/useQuery.tsx
@@ -21,6 +21,16 @@ import { useEffect, useReducer, useMemo, useRef } from "react";
 import { createCachedCollection } from "./cachedCollection";
 import { symbols } from "@realm/common";
 
+type RealmResults<T> = Realm.Results<T & Realm.Object>
+type RealmObjectType<T> = string | ({ new (...args: any): T } & Realm.ObjectClass);
+type UseQueryArgs<T> = [
+    RealmObjectType<T>,
+  ] | [
+    RealmObjectType<T>,
+    (collection: RealmResults<T>) => RealmResults<T>,
+    any[],
+  ];
+
 /**
  * Generates the `useQuery` hook from a given `useRealm` hook.
  *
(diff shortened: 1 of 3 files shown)


takameyer:
At first glance it looks great, and thanks again for creating the PR!  We will need to write some tests and get the process started for updating the documentation before this gets merged.  Will update here on any progress.

packages/realm-react/src/useQuery.tsx
```diff
@@ -29,25 +39,33 @@ import { symbols } from "@realm/common";
  */
 export function createUseQuery(useRealm: () => Realm) {
   return function useQuery<T>(
-    type: string | ({ new (...args: any): T } & Realm.ObjectClass),
-  ): Realm.Results<T & Realm.Object> {
+    ...[type, collectionCallback, dependencies]: UseQueryArgs<T>
```
takameyer, line 42:
```suggestion
    type: RealmObjectType<T>, 
    query: QueryCallback<T> = (collection) => collection, 
    deps: DependencyList = [queryCallback]
    ): RealmResults<T> {
```
Type the arguments directly.  This will make the typings much cleaner in the code implementing `useQuery`, while also showing `query` and `deps` are optional.

I've also renamed the arguments to what I feel is more consistent with the naming conventions of basic hooks from React.

packages/realm-react/src/useQuery.tsx
```diff
@@ -29,25 +39,33 @@ import { symbols } from "@realm/common";
...
+  ): RealmResults<T> {
     const realm = useRealm();
 
+    const collectionCallbackMemo = useMemo(() => {
+      if (!!collectionCallback) {
+        return collectionCallback;
+      }
+      return (collection: RealmResults<T>) => collection;
+    }, dependencies || []);
```
takameyer, line 51:
```suggestion
        const queryResult = useMemo(() => {
          return query(realm.objects(type));
        }, deps)
```
With the defaults, this memo becomes simpler.  I also think we can call the callback directly and use the result directly in the lines below.
Also with the change above, deps include the `query` so that if that callback changes, the results are recalculated.

takameyer (changes requested):
I've added some requests to clean the code and typings up a bit.  The documentation for `useQuery` also needs to be documented.

packages/realm-react/src/useQuery.tsx
levipro: Should the default for `deps` not just be an empty array? Where is `queryCallback` coming from?

takameyer: Typo, `queryCallback` is `query`.
It should be an empty array, but perhaps it would be good to spread `query` into `deps` in the `useMemo`

levipro: So one goal I had here was to make it so that the end user did not have to define their `query` function with `useCallback`

The `query` param will change on each render since it will be defined inline by the user, but it is memoized internally within the `useQuery` hook based on the dependency array the user passes. Therefore in a design like that we should not rely on `query` to be a stable object

packages/realm-react/src/useQuery.tsx
levipro: If you are going to use realm inside this useMemo then `realm` and `type` will need to be dependencies along side `deps`

I think this complicates the use of the dependency array here and if the goal is to have a memoized `queryResult` it should be in a separate `useMemo` that uses the queryCallback as a dependency like this

```js
    const queryCallback = useMemo(() => {
      // query is unstable so we memoize it here
      return query;
    }, deps);

    const queryResult = useMemo(() => {
      return queryCallback(realm.objects(type));
    }, [type, realm, queryCallback]);
    
    ...
    
    const { collection, tearDown } = useMemo(() => {
      return createCachedCollection({
        collection: queryResult,
        realm,
        updateCallback: forceRerender,
        updatedRef,
      });
    }, [queryResult]);
```

packages/realm-react/src/useQuery.tsx
takameyer: Good point 👍🏼

packages/realm-react/src/useQuery.tsx
```diff
@@ -29,25 +34,38 @@ import { symbols } from "@realm/common";
...
+    deps: DependencyList = [],
+  ): RealmResults<T> {
     const realm = useRealm();
 
+    const queryCallback = useMemo(() => {
+      // We want the user of this hook to be able pass in the `query` function inline (without the need to `useCallback` on it)
+      // This means that the query function is unstable and will be a redefined on each render of the component where `useQuery` is used
+      // Therefore we use the `deps` array to memoize the query function internally, and only use the returned `queryCallback`
+      return query;
+    }, deps);
```
takameyer, line 48: I think this could be simplified.
```suggestion
      // We want the user of this hook to be able pass in the `query` function inline (without the need to `useCallback` on it)
      // This means that the query function is unstable and will be a redefined on each render of the component where `useQuery` is used
      // Therefore we use the `deps` array to memoize the query function internally, and only use the returned `queryCallback`
    const queryCallback = useCallback(query, deps);
```

takameyer (approved):
This looks good so far.  I'm going to merge this into a new branch and add some tests as well as update examples.  We are doing a few renovations in `master` right now, so I'll need to be able to rebase.

Really appreciate your contribution!  Excited to get this live.

PR #5466, opened 2023-02-21
[bindgen] SDK packaging

kraenhansen:
## What, How & Why?

This moves the Android source files needed to bootstrap the library into the SDK package to make them available for the dependent apps.
It also updates the package.json with conditional exports as speced here: https://nodejs.org/api/packages.html#conditional-exports


packages/realm/package.json
```diff
@@ -22,19 +22,38 @@
...
-  "types": "dist/bundle.d.ts",
   "exports": {
-    ".": "./dist/bundle.mjs",
-    "./scripts/submit-analytics": "./scripts/submit-analytics.mjs"
+    ".": {
+      "types": "./dist/bundle.d.ts",
+      "node": "./dist/bundle.node.mjs",
+      "require": "./dist/bundle.node.js",
+      "react-native": "./dist/bundle.react-native.mjs"
+    },
```
elle-j, line 31: Neat 😎

elle-j (approved):
LGTM

scripts/build-android.js
```diff
@@ -20,6 +20,7 @@ const commandLineArgs = require("command-line-args");
 const fs = require("fs-extra");
 const path = require("path");
 const exec = require("child_process").execFileSync;
+const packageRoot = path.resolve(__dirname, "..");
```
takameyer, line 23: Should we move the build scripts into `packages/realm`?

takameyer: Also this will put the build artifacts in `react-native` instead of `packages/realm/react-native`.  I think we want to actually delete the root project `react-native` folder and have everything within `packages/realm/react-native`, right?

takameyer (changes requested):
The Android build still lands in the root `react-native` directory.  That should be addressed.

scripts/build-android.js
kraenhansen: > move the build scripts into `packages/realm`?

I was thinking about that. I see pros and cons to it. It probably makes sense to increase cohesion as these will only be used to build the SDK native parts. But then again, the CmakeLists.txt live in the bindgen package right now, which sort of serves the same purpose.

One my thought I had was that I want to refactor and join the build scripts into a single CLI that can invoke cmake when building for Node.js, iOS or Android with one (or more) platform and arch flags.

takameyer:
> One my thought I had was that I want to refactor and join the build scripts into a single CLI that can invoke cmake when building for Node.js, iOS or Android with one (or more) platform and arch flags. 

Also a good idea.  It just seems to me that these scripts are only called from within `packages/realm`, so they could be transfered up.  In any case, if we wrapped it into a single cli, it would still end up being a dependency of `packages/realm`.

kraenhansen:
> The Android build still lands in the root react-native directory. That should be addressed.

@takameyer It was less baked than I remembered 🙂 But it's fixed now. Thanks!

takameyer (approved):
LGTM!

PR #5473, opened 2023-02-21
Add `MongoDBCollection` tests

### LINKED ISSUE #5464: Implement tests for `MongoDBCollection` methods
Tests are needed for `MongoDBCollection` methods such as `find()`, `inserMany()`, etc.

elle-j:
## What, How & Why?

Adds tests for `MongoDBCollection` methods such as `find()`, `findOne()`, `insertMany()`, etc.

This closes #5464.

## ☑️ ToDos
* [x] 🚦 Tests


packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -15,6 +15,7 @@
 // limitations under the License.
 //
 ////////////////////////////////////////////////////////////////////////////
```
elle-j, line 17: The changes in this file pertain to a bug with handling a difference between `nullptr` and the string `"null"`, as well as propagating errors thrown by the callback passed to the bindings `make_network_transport()`, into the promise chain.

Thanks to @RedBeard0531 for helping fix this interesting bug! 🐛 🚀

packages/realm/src/app-services/MongoDBCollection.ts
```diff
@@ -456,7 +456,7 @@ export class MongoDBCollection<T extends Document> {
    * @param options Additional options to apply.
    * @returns The document found before updating it.
    */
-  findOneAndUpdate(filter: Filter = {}, update: Update, options: FindOneAndModifyOptions = {}): Promise<T | null> {
+  findOneAndUpdate(filter: Filter, update: Update, options: FindOneAndModifyOptions = {}): Promise<T | null> {
```
elle-j, line 459: Removed the default arg since the 2nd arg is non-optional.

packages/realm/src/app-services/MongoDBCollection.ts
```diff
@@ -477,11 +477,7 @@ export class MongoDBCollection<T extends Document> {
    * @param options Additional options to apply.
    * @returns The document found before replacing it.
    */
-  findOneAndReplace(
-    filter: Filter = {},
-    replacement: unknown,
-    options: FindOneAndModifyOptions = {},
-  ): Promise<T | null> {
+  findOneAndReplace(filter: Filter, replacement: unknown, options: FindOneAndModifyOptions = {}): Promise<T | null> {
```
elle-j, line 480: Removed the default 1st arg since the 2nd arg is non-optional.

packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -94,21 +95,27 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
     // Copied from lib/utils.js.
     // TODO consider importing instead.
     // Might be slightly faster to make dedicated wrapper for 1 and 2 argument forms, but unlikely to be worth it.
-    function _promisify(func) {
+    function _promisify(nullAllowed, func) {
```
kraenhansen, line 98: This change reminds me that it would be great to get this code moved to a separate file some day to get linting and editor support.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -93,6 +93,438 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
...
+          // Filter all docs with `_id` > `insertedId1`.
+          { $match: { _id: { $gt: insertedId1 } } },
+          // Return a single doc (`_id: null`) with property `count` set to the number of filtered docs.
+          { $group: { _id: null, count: { $sum: 1 } } },
+          // Remove the `_id` field from the result.
+          { $project: { _id: false } },
+        ]);
+        expect(result).to.deep.equal([{ count: 2 }]);
+
+        // Note:
```
kneth, line 516: It is documented in `integration-tests/baas/README.md` but please consider to copy steps 1-4 into it.

kneth (approved):

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -93,6 +93,438 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
   });
 
   describe("MongoDBCollection", function () {
+    const insertedId1 = 1;
+    const insertedId2 = 2;
+    const insertedId3 = 3;
+    const insertedText = "Test document";
+    const nonExistentId = 100;
+
+    async function insertThreeDocuments(): Promise<void> {
```
gagik, line 102: I think this could be refactored into a beforeEach hook, it's used in the start of most of the statements so will make this cleaner.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -93,6 +93,438 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
   });
 
   describe("MongoDBCollection", function () {
+    const insertedId1 = 1;
```
gagik, line 96: Is there a benefit of having `insertedId1 = 1; insertedId2 = 2;` instead of just using the `1, 2...`?

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -93,6 +93,438 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
...
+
+        const docs = await collection.find({ _id: { $gt: insertedId1 } });
+        expect(docs).to.have.length(2);
+        for (const doc of docs) {
+          expect(doc._id > insertedId1).to.be.true;
+        }
+      });
+
+      it("returns empty array if collection is empty", async function (this: AppContext & Mocha.Context) {
+        const docs = await collection.find();
```
gagik, line 145: You will probably want to move this to a separate describe if you do go down the `beforeEach` route, rest seem to use it.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -93,6 +93,438 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
...
+          { $match: { _id: { $gt: insertedId1 } } },
+          // Return a single doc (`_id: null`) with property `count` set to the number of filtered docs.
+          { $group: { _id: null, count: { $sum: 1 } } },
+          // Remove the `_id` field from the result.
+          { $project: { _id: false } },
+        ]);
+        expect(result).to.deep.equal([{ count: 2 }]);
+
+        // Note:
+        // If getting `Error: exec: "assisted_agg": executable file not found in $PATH`:
```
gagik, line 517: I think I remember @papafe adding this to the README when resolving issues when installing baas in general, is this only relevant with these tests or is this a more global issue?

gagik (approved):
Nice! Just minor comments

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j: If we have them as variables rather than hardcoded in each test I think it makes the tests more readable since it would hint we're dealing with an id that has already been inserted. Same for the `nonExistentId` variable. Also, it's a lot easier to change the values if ever needed.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j: 👍 Yeah was contemplating that at first. I agree that most tests do use it, but I think there may be enough of the other tests (that don't use it as the very first statement) that it justifies having them where it's needed. Otherwise, some tests would (like you're pointing out in another comment) need to be moved to another describe test suite.

I'll think some more about this and see if having it in a beforeEach could be a better choice. I also like seeing "insertThreeDocuments" in the test so the rest makes more sense as well, e.g. when expecting a length to be `3`.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j: [Done](https://github.com/realm/realm-js/pull/5475) ✅

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j: Yeah thankfully I asked Kenneth about this early enough that he remembered Ferdinando had just encountered this 😄 

It concerns only the aggregation pipeline. So if skipping this test, it wouldn't be needed.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -93,6 +93,438 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
...
+    });
+
+    describe("#updateOne", function () {
+      const updatedText = "Updated text";
+
+      it("updates specific document", async function (this: AppContext & Mocha.Context) {
+        await insertThreeDocuments();
+
+        const result = await collection.updateOne({ _id: insertedId3 }, { $set: { text: updatedText } });
+        expect(result).to.deep.equal({ matchedCount: 1, modifiedCount: 1 });
```
JacobOscarGunnarsson, line 367: one suggestion here could also be to fetch the updated document and validate that it contains the new value for `text`

JacobOscarGunnarsson: there are other cases in the file where we rely on `expect(result).to.deep.equal(matchedcount..., modifiedCount...)`. I don't know how petty we want to be but I think it would be worth adding those extra checks as they're not too difficult to write nor imply too much overhead.

JacobOscarGunnarsson (approved):

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j: Thanks, I'll add that :) (I'll go ahead and apply that to e.g. deletions as well, checking that the fetched doc is `null`).

integration-tests/tests/src/tests/sync/mongo-db-client.ts
takameyer: If you added it to the `describe` for the cases that do call `insertThreeDocuments`, it would be fine.

takameyer (approved):
LGTM!  Thanks for adding this!

integration-tests/tests/src/tests/sync/mongo-db-client.ts
gagik: How about this? 

```ts
describe("#count", function () {
  describe("after insertions", function () {
       beforeEach(async function (...) => {
        await insertThreeDocuments();
       });

    // Test (B)
    it("returns number of documents using query selector", ...) {
      // ...
    });
  });
  describe("special cases", function () {
            // Test (C)
            it("returns zero when there are no matches", ...) {

              // Never calls the method

              const count = await collection.count({ _id: nonExistentId });
              expect(count).to.equal(0);
            });
  });
});
```
Of course wording can be changed, but this would be quite organised

elle-j: Do you mean having separate `beforeEach`s for each function tested that need to (e.g. `#count`, `#find`)? I may also be misinterpreting the example, but it looks like (for the case of `#count`) the `describe("after insertions", ..)` would only contain 1 `it` test. Not sure if having a `beforeEach` for a suite with 1 test is needed? Again, correct me if my interpretation is off 😛

PR #5478, opened 2023-02-22
Andrew/app-importer/rules

takameyer:
## What, How & Why?
BaaS rules endpoint changed for flexible sync.  This makes the tests green again.

* Remove any other rules and apply a general rule that allows everything
* Change the app importer to update the default rule
* Introduces a couple of changes I needed to make to the baas test server package to make the tests pass (via #5483).
* Adds a conditional to the URL used when creating rules, to fall back to the old behaviour for the "HTTP" service (via #5483).

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/default.json
```diff
@@ -0,0 +1,14 @@
+{
+	"roles": [{
+			"name": "defaultRole",
```
kraenhansen, line 3: Or is this a new "default" name? Also - I'd expect this to match the name of the file to match all the other config files.
```suggestion
			"name": "default",
```

packages/realm-app-importer/src/AppImporter.ts
```diff
@@ -351,10 +351,7 @@ export class AppImporter {
         const configFilePath = path.join(servicesDir, serviceDir, "config.json");
         if (fs.existsSync(configFilePath)) {
           const config = this.loadJson(configFilePath);
-          console.log("Creating service: ", config.name);
```
kraenhansen, line 354: Don't we want this anymore? I mean, we have all the other logs still.

kraenhansen (approved):

integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/test_data.Dog.json
```diff
@@ -1,35 +0,0 @@
...
-            "realm_id": {
-                "bsonType": "long"
-            }
-        },
-        "required": [
-            "_id",
-            "name"
-        ],
-        "title": "Dog"
-    }
```
kraenhansen, line 34: These schemas are probably not needed anymore, since the apps are in dev-mode. But I wanted to ensure that it wasn't just deleted by accident? 🙂

integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/default.json
takameyer: Hmm...I guess it doesn't make a big difference, but I can make the change to make it consistant.

integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/test_data.Dog.json
takameyer: I decided to be bold and see if the test apps functioned and passed without these schemas.  They did 💪🏼  So they have been removed to prevent confusion.

takameyer:
Looks like there's still work to be done to get the `realm-web-integration-tests` to work

kraenhansen:
You probably have to update the docker image: https://github.com/realm/realm-js/blob/master/.github/workflows/pr-realm-web.yml#L56 (and fix any issues that might arise from that 😞)

integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/default.json
takameyer: I renamed the files instead

integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/defaultRole.json
```diff
@@ -0,0 +1,14 @@
+{
```
JacobOscarGunnarsson, line 1: Just want to double check, we're not able to remove the permission rule in `services/mongodb/config.json`? Might be the case that it's only allowed in flexible sync configurations but thought it'd be worth to ask.

JacobOscarGunnarsson (approved):
LGTM

integration-tests/realm-apps/pv-int-tests/services/mongodb/rules/defaultRole.json
takameyer: Correct, it only applies to `flexible sync`.  I attempted to remove these permissions for partition based as well and the API returned an error saying the required fields were missing.

Issue #4934, opened 2022-09-19
Export only the BSON types available for `Realm.Object` properties on `Realm.BSON`


kraenhansen:
We're currently re-exporting all of the `bson` package API on `Realm.BSON`.
To simplify our exports and make it easier for users to find the types that are actually available for properties, I suggest we export only the types we currently support: `ObjectId`, `Decimal128` and `UUID`.

kraenhansen:
We should wait for TypeScript-first models to land before fixing this, as it introduce a `Realm.Types` namespace where the BSON types will be available.

PR #5479, opened 2023-02-22
Fix integration tests type errors on bindgen

gagik:
## What, How & Why?

Fixes all type errors in our integration tests pre-test migration.

This closes https://github.com/realm/realm-js/issues/5486

## ☑️ ToDos


packages/realm/src/bson.ts
```diff
@@ -27,6 +27,8 @@ import * as bson from "bson";
 export namespace BSON {
   export const ObjectId = bson.ObjectId;
   export type ObjectId = bson.ObjectId;
+  export const ObjectID = bson.ObjectID;
```
gagik, line 30: Do we want to do this? Some tests did use the `bson.ObjectID` alias

packages/realm/src/Realm.ts
```diff
@@ -822,7 +822,7 @@ export class Realm {
   /**
    * Deletes the provided Realm object, or each one inside the provided collection.
    */
-  delete(subject: RealmObject | RealmObject[] | List | Results): void {
+  delete<T = DefaultObject>(subject: RealmObject<T> | RealmObject<T>[] | List<T> | Results<T>): void {
```
gagik, line 825: I am not sure if this is the right way to go about this, especially with regards to List/Results, will need more testing, but this is for the case where you pass a interface returned by `Realm.create<IDog>` and then try `realm.delete(dog)`

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -254,8 +254,7 @@ describe("Dictionary", () => {
       expect(stringifiedAndParsed).deep.equals({ dict: values });
     });
 
-    // TODO: Unskip once https://github.com/realm/realm-core/issues/4805 is fixed
-    it.skip("throws a meaningful error if accessed after deletion", function (this: RealmContext) {
+    it("throws a meaningful error if accessed after deletion", function (this: RealmContext) {
```
gagik, line 257: This is the test that necessitated change of how `.delete` type is being handled right now, see below:

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -183,6 +202,7 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
 
       it("accepts { flexible: undefined } and a partition value", function () {
```
gagik, line 203: Should this test exist? I guess it doesn't hurt.

integration-tests/tests/src/typings.d.ts
```diff
@@ -16,7 +16,6 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-type Realm = import("realm").Realm;
 type RealmObject = import("realm").Object;
```
gagik, line 19: should all these global imports be removed? seems to be a legacy approach

packages/realm/src/Realm.ts
gagik: I noticed you added `any` to remove before, should it be something like `RealmObject<any>` instead?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -29,14 +29,26 @@
...
+} from "realm";
 
 import { authenticateUserBefore, importAppBefore, openRealmBeforeEach } from "../../hooks";
-import { DogSchema, IPerson, PersonSchema } from "../../schemas/person-and-dog-with-object-ids";
+import { Dog, IPerson, Person, PersonSchema } from "../../schemas/person-and-dog-with-object-ids";
 import { closeAndReopenRealm, closeRealm } from "../../utils/close-realm";
 import { expectClientResetError } from "../../utils/expect-sync-error";
 
-const FlexiblePersonSchema = { ...PersonSchema, properties: { ...PersonSchema.properties, nonQueryable: "string?" } };
+class FlexiblePerson extends Person {
```
gagik, line 47: I figured it'd be nice to just use class-based schemas in these tests. Not sure if I should just not extend Person and move all the information here, since this seems to be our preference generally.

integration-tests/tests/src/schemas/person-and-dog-with-object-ids.ts
```diff
@@ -24,7 +24,7 @@ export interface IPerson {
   _id: Realm.BSON.ObjectId;
   name: string;
   age: number;
-  friends: Realm.List<IPerson>;
+  friends: Realm.List<Person>;
```
gagik, line 27: Doing this as a quickfix for now... pending our discussions regarding this

integration-tests/tests/tsconfig.json
```diff
@@ -9,7 +9,7 @@
 		"resolveJsonModule": true,
 		"outDir": "dist",
 		"lib": [
-			"es2018"
+			"es2020"
```
gagik, line 12: Fixes:
```
../../node_modules/@thi.ng/api/typedarray.d.ts:112:10 - error TS2583: Cannot find name 'BigUint64Array'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2020' or later.
```

integration-tests/tests/src/tests/queries.ts
```diff
@@ -76,8 +76,11 @@ describe("Realm Query Language", () => {
         t: new Date("2004-02-26T10:15:02"),
       });
     });
+    //@ts-expect-error Test about to change
```
elle-j, line 79: Could you fill me in on `Test about to change`? :)

integration-tests/tests/src/tests/sync/asymmetric.ts
```diff
@@ -78,6 +78,7 @@ describe.skipIf(environment.missingServer, "Asymmetric sync", function () {
     it("an asymmetric schema cannot be queried through '_objectForObjectKey()'", function (this: RealmContext) {
       expect(() => {
         // A valid objectKey is not needed for this test
+        //@ts-expect-error Should not be correct usage
```
elle-j, line 81: Not sure if `Internal method` may be more appropriate ts expect err message here? (Like the other ones you added). Since from TS's side, the error is regarding accessing an internal method, while the test itself already describes the incorrect usage part.

integration-tests/tests/src/tests/sync/client-reset.ts
```diff
@@ -305,18 +316,20 @@ function getSchema(useFlexibleSync: boolean) {
...
-          {
-            schema: getSchema(useFlexibleSync),
-            sync: {
-              _sessionStopPolicy: SessionStopPolicy.Immediately,
-              ...(useFlexibleSync ? { flexible: true } : { partitionValue: getPartitionValue() }),
-              user: this.user,
-              clientReset: {
-                mode: ClientResetMode.Manual,
-              },
+        const config: ConfigurationWithSync = {
```
elle-j, line 319: Nice that you extracted this :)

integration-tests/tests/src/tests/sync/asymmetric.ts
takameyer, line 81:
```suggestion
        //@ts-expect-error Internal method
```

integration-tests/tests/src/tests/sync/flexible.ts
elle-j: 🚀

packages/realm/src/Realm.ts
takameyer: I think this needs an overloaded definition to better depict these types, similar to `create` above.  I don't think `any` is the answer

packages/realm/src/bson.ts
takameyer: I wish it didn't exist, but we might not have a choice.

takameyer (approved):
Looks good!  We might want to bring some of those open questions up in the group.
And I'm sure there will also be a lot more of this type of work once the migrated tests land in `bindgen`.  Might be a good opportunity to implement whatever is the answer to said open questions.

integration-tests/tests/src/tests/sync/flexible.ts
elle-j: It's good to keep that, it also validates that [this function](https://github.com/realm/realm-js/blob/f14917b31d5caa8dabe5814f011cc04421c6052d/packages/realm/src/app-services/SyncConfiguration.ts#L354) does its job correctly.

kraenhansen:
On a general note. I would like to see green tests before we refactor (which I do believe some of these change are) these tests.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1456,7 +1453,11 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
      */
     async function addPersonAndWaitForUpload(realm: Realm): Promise<{ person: IPerson; id: BSON.ObjectId }> {
       const person = realm.write(function () {
-        return realm.create<IPerson>(FlexiblePersonSchema.name, { _id: new BSON.ObjectId(), name: "Tom", age: 36 });
+        return realm.create<FlexiblePerson>(FlexiblePerson, {
+          _id: new BSON.ObjectId(),
+          name: "Tom",
```
elle-j, line 1458: Did Tom write this? 😄

packages/realm/src/bson.ts
kraenhansen: We used to export all of `bson` here, but we agreed [in the technical design](https://docs.google.com/document/d/1mju1TRQAiA0CxXHMs8MYRnvUhqUdkCrh6g-Rb3OUEHk/edit#bookmark=id.krw7alyf7g6j) that we wanted to break this.

```suggestion
```

elle-j (approved):
Awesome, it's going to be really nice with these type errors gone!

integration-tests/tests/src/tests/queries.ts
gagik: Yes haha, I just know this file will be completely revamped the moment we merge test-migrations so did not want to do anything about it, should have been clearer in my comment.

packages/realm/src/Realm.ts
gagik: I guess this is also a question of should we be encouraging this? `IDog` as an interface isn't necessarily managed by Realm, it would probably be good to discourage it in terms of type-checking...

packages/realm/src/bson.ts
gagik: Is there a place I could document this (minor) breaking change? I am mentioning https://github.com/realm/realm-js/issues/4934 just for future reference.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1394,23 +1390,24 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
           const { subs } = addSubscriptionForPerson(this.realm);
 
           await subs.update((mutableSubs) => {
-            expect(mutableSubs.removeByObjectType(DogSchema.name)).to.equal(0);
+            // TODO: should this not support removeByObjectType(Dog)?
+            expect(mutableSubs.removeByObjectType(Dog.schema.name)).to.equal(0);
```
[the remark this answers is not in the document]
gagik: We have this functionality with i.e. `create` so I do think this should be possible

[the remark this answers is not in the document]
elle-j: Yeah, it could provide a nicer consistency with both `create` and `objects` :)

kneth (approved):
Nice clean up!

JacobOscarGunnarsson (approved):
LGTM

kneth:
It will be complicated to rebase this branch so I am closing the PR

Issue #3365, opened 2020-10-28
Enums declared in types are missing implementations


kraenhansen:
## Goals
Use values from enums in code.

## Expected Results
```typescript
const Realm = require("realm");
assert(typeof Realm.ConnectionState === "object");
assert(typeof Realm.ConnectionState.Disconnected === "string");
```

## Actual Results
The enums (except `UpdateMode`) declared by the types are all undefined:

* `Realm.SessionStopPolicy`
* `Realm.OpenRealmBehaviorType`
* `Realm.OpenRealmTimeOutBehavior`
* `Realm.ConnectionState`
* `Realm.NumericLogLevel`

## Steps to Reproduce & Code Sample
1. Open up the Node.js REPL
1. Require Realm `const Realm = require("realm")`
1. Try accessing the enum `Realm.ConnectionState.Disconnected`
1. `Uncaught TypeError: Cannot read property 'Disconnected' of undefined` 

## Version of Realm and Tooling
* Realm JS SDK Version: ^10.0.1
* Node or React Native: Tested on Node.js (it's most likely also an issue in React Native).
* Client OS & Version: N/A
* Which debugger for React Native: None

PR #5484, opened 2023-02-23
Ensure that Realm enums are accessible

JacobOscarGunnarsson:
Small tests to make sure that the issue in [#3365](https://github.com/realm/realm-js/issues/3365) is resolved

### DIFF
--- a/integration-tests/tests/src/tests/enums.ts
+++ b/integration-tests/tests/src/tests/enums.ts
@@ -28,46 +28,46 @@ import {
 describe("Enums", function () {
   describe("ConnectionState", function () {
     it("is accessible", function () {
-      expect(typeof ConnectionState === "object");
-      expect(typeof ConnectionState.Disconnected === "string");
-      expect(typeof ConnectionState.Connecting === "string");
-      expect(typeof ConnectionState.Connecting === "string");
+      expect(ConnectionState).to.be.an("object");
+      expect(ConnectionState.Disconnected).equals("disconnected");
+      expect(ConnectionState.Connecting).equals("connecting");
+      expect(ConnectionState.Connected).equals("connected");
     });
   });
   describe("SessionStopPolicy", function () {
     it("is accessible", function () {
-      expect(typeof SessionStopPolicy === "object");
-      expect(typeof SessionStopPolicy.AfterUpload === "string");
-      expect(typeof SessionStopPolicy.Immediately === "string");
-      expect(typeof SessionStopPolicy.Never === "string");
+      expect(SessionStopPolicy).to.be.an("object");
+      expect(SessionStopPolicy.AfterUpload).equals("after-upload");
+      expect(SessionStopPolicy.Immediately).equals("immediately");
+      expect(SessionStopPolicy.Never).equals("never");
     });
   });
   describe("OpenRealmBehaviourType", function () {
     it("is accessible", function () {
-      expect(typeof OpenRealmBehaviorType === "object");
-      expect(typeof OpenRealmBehaviorType.DownloadBeforeOpen === "string");
-      expect(typeof OpenRealmBehaviorType.OpenImmediately === "string");
+      expect(OpenRealmBehaviorType).to.be.an("object");
+      expect(OpenRealmBehaviorType.DownloadBeforeOpen).equals("downloadBeforeOpen");
+      expect(OpenRealmBehaviorType.OpenImmediately).equals("openImmediately");
     });
   });
   describe("OpenRealmTimeOutBehaviour", function () {
     it("is accessible", function () {
-      expect(typeof OpenRealmTimeOutBehavior === "object");
-      expect(typeof OpenRealmTimeOutBehavior.OpenLocalRealm === "string");
-      expect(typeof OpenRealmTimeOutBehavior.ThrowException === "string");
+      expect(OpenRealmTimeOutBehavior).to.be.an("object");
+      expect(OpenRealmTimeOutBehavior.OpenLocalRealm).equals("openLocalRealm");
+      expect(OpenRealmTimeOutBehavior.ThrowException).equals("throwException");
     });
   });
   describe("NumericLogLevel", function () {
     it("is accessible", function () {
-      expect(typeof NumericLogLevel === "object");
-      expect(typeof NumericLogLevel.All === "number");
-      expect(typeof NumericLogLevel.Debug === "number");
-      expect(typeof NumericLogLevel.Detail === "number");
-      expect(typeof NumericLogLevel.Error === "number");
-      expect(typeof NumericLogLevel.Fatal === "number");
-      expect(typeof NumericLogLevel.Info === "number");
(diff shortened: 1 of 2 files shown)


integration-tests/tests/src/tests/enums.ts
```diff
@@ -0,0 +1,73 @@
...
+  OpenRealmBehaviorType,
+  OpenRealmTimeOutBehavior,
+  SessionStopPolicy,
+} from "realm";
+
+describe("Enums", function () {
+  describe("ConnectionState", function () {
+    it("is accessible", function () {
+      expect(typeof ConnectionState === "object");
+      expect(typeof ConnectionState.Disconnected === "string");
```
JacobOscarGunnarsson, line 32: Could be redundant to check all enum values, feel free to add thoughts about this.

elle-j: I'd vote for keeping them (i.e. testing all values). Even though the issue mentioned asserting that the value is a string, we'd get more accurate tests if we compare it to the actual string value (e.g. `disconnected`).

On a smaller note, perhaps we want to structure them more toward "natural language"? E.g.:
```ts
// This will pass for e.g. arrays as well.
expect(typeof ConnectionState).to.equal("object");

// This will not pass for arrays.
expect(ConnectionState).to.be.an("object");

expect(ConnectionState.Disconnected).to.equal("disconnected");
expect(ConnectionState.Connecting).to.equal("connecting");
// Fixed typo from `Connecting` to `Connected`
expect(ConnectionState.Connected).to.equal("connected");
```

integration-tests/tests/src/tests/enums.ts
```diff
@@ -0,0 +1,73 @@
...
+      expect(typeof ConnectionState.Connecting === "string");
+      expect(typeof ConnectionState.Connecting === "string");
```
elle-j, line 34: Small typo (should be `Connected`)

elle-j (commented):
Nice :D Got some suggestions 👍

integration-tests/tests/src/tests/enums.ts
JacobOscarGunnarsson: yeah went through this a bit too fast 🤦  using mocha as extensively as possible is what we've already done in `test-migration-master` so should aim for that! Thanks for pointing it out

elle-j: Np 👍  (perhaps `to.deep.equal()` could test all those cases at once as well)

elle-j (approved):
😎

integration-tests/tests/src/tests/enums.ts
```diff
@@ -28,46 +28,46 @@ import {
...
-      expect(typeof NumericLogLevel.Debug === "number");
-      expect(typeof NumericLogLevel.Detail === "number");
-      expect(typeof NumericLogLevel.Error === "number");
-      expect(typeof NumericLogLevel.Fatal === "number");
-      expect(typeof NumericLogLevel.Info === "number");
-      expect(typeof NumericLogLevel.Off === "number");
-      expect(typeof NumericLogLevel.Trace === "number");
-      expect(typeof NumericLogLevel.Warn === "number");
+      expect(NumericLogLevel).to.be.an("object");
+      expect(NumericLogLevel.All).equals(0);
```
gagik, line 62: alternatively you could possibly (but I am not sure if this will work?) do a deep equal with an object i.e.
```ts
expect(NumericLogLevel).deep.equals({
          All: 0,
          Trace: 1
           ...
});
```
This might be better as well in case in the future we add a new value to the enum, as it would cause the test to fail, which it wouldn't with the current approach. Same for other enums. Again, not sure if this will work how I think it'll work.

gagik (approved):
Nice, just a minor suggestion

integration-tests/tests/src/tests/enums.ts
gagik: Just realized @elle-j had the same idea

JacobOscarGunnarsson: I'll give it a shot, was a bit concerned that the syntax would make it less readable but the point with adding new values is very valid!

JacobOscarGunnarsson:
I added the `deep.equals`, I'm a but torn on how it looks with the `NumericLogLevel`, but it might be worth it

integration-tests/tests/src/tests/enums.ts
```diff
@@ -0,0 +1,92 @@
...
+      expect(OpenRealmTimeOutBehavior).to.deep.equal({
+        OpenLocalRealm: "openLocalRealm",
+        ThrowException: "throwException",
+      });
+    });
+  });
+  describe("NumericLogLevel", function () {
+    it("is accessible", function () {
+      expect(NumericLogLevel).to.be.an("object");
+      expect(NumericLogLevel).to.deep.equals({
```
elle-j, line 70: This first line shouldn't be needed when you're doing deep equals to an object. Btw, does it still work with having the `s`at the end of `equal`? (I'd vote for where you're using `to.deep.equal()`) :)

gagik: Yeah, most of these "to" "be" "equal/equals" are just meaningless syntactic sugar that does not change its behavior. So it could be `expect(NumericLogLevel).deep.equals` or `expect(NumericLogLevel).to.deep.equal`. I like both, I have generally been using the former for less use of this chaining but the latter does read nicer.

integration-tests/tests/src/tests/enums.ts
```diff
@@ -0,0 +1,87 @@
...
+    it("is accessible", function () {
+      expect(OpenRealmTimeOutBehavior).to.deep.equal({
+        OpenLocalRealm: "openLocalRealm",
+        ThrowException: "throwException",
+      });
+    });
+  });
+  describe("NumericLogLevel", function () {
+    it("is accessible", function () {
+      expect(NumericLogLevel).to.deep.equal({
```
gagik, line 65: Just checked and we do use `deep.equals` 161 trimes and `to.deep.equal` 9 times... with that in mind I am leaning towards being consistent... sorry for leading you wrong way

JacobOscarGunnarsson: I like consistency so I appreciate that you lifted this :D

kneth (approved):

integration-tests/tests/src/tests/enums.ts
kraenhansen: > Could be redundant to check all enum values.

I'm probably feeling this a bit. If these tests fail it'll either be because we (hopefully consciously) changed the enums or the TS compiler is broken (in which case this will be the least of our worries). These tests are in risk of becoming outdated if we add new values to the enum.

The main reason we had these issues on `master` was because we declared types separately from implementation (in JS), so I probably don't feel the need for this to be a part of our test suite. But then again - they are written now, so they don't hurt too much to keep 😅

PR #5487, opened 2023-02-23
Added overload to `Object.linkingObjects`

### LINKED ISSUE #5326: `linkingObjects()` TS type inference + class-based models

Always

### Description

The `Realm.Object.linkingObjects()` method, when given a class-based model as a type argument, does not infer the correct TypeScript type for the collection that it returns.

The method does not allow you to specify the class-based model, though this would be a welcome improvement and in line with the API for `Realm.objects`, `useObject`, etc.

```ts
// ✅ The useObject hook correctly infers the `Post` type based on the class I pass in
const post = useObject(Post, _id)!; // Post & Realm.Object<Post, never>

// ⛔️ This doesn't work - the first arg must be a string - but it would be ideal
const user = post.linkingObjects(User, 'posts')[0]!; // User & Realm.Object<User, never>

// 😵‍💫 This runs but it uses the class _name_ as a string and doesn't infer the correct return type
const user = post.linkingObjects('User', 'posts')[0]!; // Realm.Object<unknown, never>

// 😵‍💫 This runs and assigns a half correct type but still uses class name + a type cast
const user = post.linkingObjects<User>('User', 'posts')[0]!; // User & Realm.Object<unknown, never>
```

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

1. Create a `Realm.Object` class with a relationship to another class.
   ```typescript
    class Foo extends Realm.Object {
      _id: Realm.BSON.ObjectId,
      name: string;
      bars: Realm.List<Bar>
      
      static schema = {
        name: 'Foo',
        properties: {
          _id: 'objectId',
          name: 'string',
          bars: 'Bar[]',
        },
        primaryKey: '_id',
      };
    }
    
    class Bar extends Realm.Object {
      _id: Realm.BSON.ObjectId,
      name: string;
      
      static schema = {
        name: 'Bar',
        properties: {
          _id: 'objectId',
          name: 'string',
        },
        primaryKey: '_id',
      };
    }

   await Realm.open({ schema: [Foo.schema, Bar.schema] })
   ```
2. Add objects that instantiate the relationship
   ```typescript
   realm.write(() => {
     const foo = realm.create(Foo, { _id: new Realm.BSON.ObjectId(), name: "Winnie the Foo Bear" })
     const bar1 = realm.create(Bar, { _id: new Realm.BSON.ObjectId(), name: "Honey" })
     const bar2 = realm.create(Bar, { _id: new Realm.BSON.ObjectId(), name: "Smile" })
     foo.bars.push(bar1)
     foo.bars.push(bar2)
   })
   ```
3. Get a collection of backlinks for the relationship using `linkingObjects()`
   ```typescript
   const bars = realm.objects(Bar);
   const bar0 = bars[0]!
   // ERROR: Argument of type 'typeof User' is not assignable to parameter of type 'string'.
   const linkingFoos = bar0.linkingObjects(Foo, "bars")!;

   // BUG: TS Type is not correct
   const linkingFoos: Realm.Results<Realm.Object<unknown, never>> = bar0.linkingObjects('Foo', "bars")!;
   ```

### Version

11.4.0

Local Database only

### Are you using encryption?

No

macOS Monterey (12.6.2)

### Build environment

_No response_

### Cocoapods version

_No response_

papafe:
This closes #5326

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
 ## vNext (TBD)
 
 ### Enhancements
-* Add an overload to `Dictionary.set` method that takes two arguments, a `key` and a `value`. ([#4286](https://github.com/realm/realm-js/issues/4286))  
+
+* Added an overload to `Object.linkingObject` method that takes type of the linking object as an input instead of its string name ([#5326](https://github.com/realm/realm-js/issues/5326))  
+Example usage:
+```typescript
+let linkedObjects = john.linkingObjects(Person, "friends");
+``` 
+* Added an overload to `Dictionary.set` method that takes two arguments, a `key` and a `value`. ([#4286](https://github.com/realm/realm-js/issues/4286))
 Example usage:
 ```typescript
 realm.write(() => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## vNext (TBD)
 
 ### Enhancements
-
 * Added an overload to `Object.linkingObject` method that takes type of the linking object as an input instead of its string name ([#5326](https://github.com/realm/realm-js/issues/5326))  
 Example usage:
 ```typescript
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
-* Added an overload to `Object.linkingObject` method that takes type of the linking object as an input instead of its string name ([#5326](https://github.com/realm/realm-js/issues/5326))  
+* Added an overload to `Object.linkingObjects` method that takes type of the linking object as an input instead of its string name ([#5326](https://github.com/realm/realm-js/issues/5326))  
 Example usage:
 ```typescript
 let linkedObjects = john.linkingObjects(Person, "friends");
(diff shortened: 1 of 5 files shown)


kraenhansen (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 ## vNext (TBD)
 
 ### Enhancements
-* Add an overload to `Dictionary.set` method that takes two arguments, a `key` and a `value`. ([#4286](https://github.com/realm/realm-js/issues/4286))  
+* Added an overload to `Object.linkingObject` method that takes type of the linking object as an input instead of its string name ([#5326](https://github.com/realm/realm-js/issues/5326))
```
```suggestion
* Added an overload to `Object.linkingObjects` method that takes type of the linking object as an input instead of its string name ([#5326](https://github.com/realm/realm-js/issues/5326))  
```

integration-tests/tests/src/tests/objects.ts
```diff
@@ -30,6 +30,42 @@ describe("Realm objects", () => {
     Realm.clearTestState();
   });
 
+  describe("Methods", () => {
+    it("linkingObject works with both string and type input", () => {
```
kneth, line 34:
```suggestion
    it("linkingObjects works with both string and type input", () => {
```

kneth (approved):
Minor comments only

takameyer (approved):

PR #5494, opened 2023-02-24
Expose Config::cancel_waits_on_nonfatal_error publicly

papafe:
This PR exposes `Config::cancel_waits_on_nonfatal_error`. When set, upload/download waits are canceled on any sync error (like a timeout) and not just fatal ones. 

Unfortunately we can't test this properly until:
- https://github.com/realm/realm-core/issues/6301 is fixed. At the moment core rejects all the "non sensible" timeouts, and we need to set a very low one to simulate reliably a connection issue
- We expose `SyncClientConfig.timeouts`(https://github.com/realm/realm-core/blob/e83515f014ff53b0d27c59b73773d33b4252e891/src/realm/object-store/sync/sync_manager.hpp#L97) so that we can set the timeouts properly. This was not exposed on JS so I've not added this for now

For these reasons the test is skipped. I have tested it manually setting a `connect_timeout` of 1ms in core (https://github.com/realm/realm-core/blob/e83515f014ff53b0d27c59b73773d33b4252e891/src/realm/object-store/sync/impl/sync_client.hpp#L54). 

I have also added `AppConfiguration.defaultRequestTimeout`, that was present in master but missing in the bindgen branch.

Another note. I've slightly modified the tests in two ways:
- I've extracted the methods for creating a sync or a local config so that they can be used without the need to necessarily open a realm too
- I've created a new `describe` in `flexible.ts` to distinguish the tests that need a realm to be opened before, and the others that don't need it, like the flexible sync configuration tests. 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Added `SyncConfiguration.cancelWaitOnNonFatalError`. When set to true, all async operations (such as opening the realm with `await Realm.open`) will fail when a non-fatal error, such as a timeout, occurs.
```
papafe, line 4: Not sure if this is formulated properly 😁 
- I am unsure wether that's a correct terminology for a promise (it fails?)
- Also I don't know if there are other methods/promises that could fail because of this new parameter. In .NET we have also methods to asynchronously waiting for all downloads/uploads, but it seems that they don't exist in JS.

integration-tests/tests/src/utils/open-realm.ts
```diff
@@ -38,43 +38,60 @@ export async function openRealm(
...
+  } as unknown as Configuration;
+}
+
+export function createLocalConfig(partialConfig: LocalConfiguration = {}): Configuration {
+  const path = getRandomPathAndNonce().path;
+
+  return { ...partialConfig, path };
+}
+
+function getRandomPathAndNonce(): { path: string; nonce: string } {
```
JacobOscarGunnarsson, line 91: In master we have a file `/src/utils/generators.ts` that deals with generating random values. Think it might make sense to move this method to it. If so it could make sense to add a TODO at least for this PR (not sure if you have the necessary changes from master since it require the bindgen rebase to be done).

CHANGELOG.md
JacobOscarGunnarsson: `I am unsure wether that's a correct terminology for a promise (it fails?)`

I guess that would be `reject`? But personally I think it's intuitive as it is now, might need input form other people as well..

JacobOscarGunnarsson (approved):
Just minor comments that shouldn't be treated as blockers, just suggestions. Otherwise LGTM

CHANGELOG.md
```suggestion
* Added configuration option `SyncConfiguration.cancelWaitOnNonFatalError`. Set to true, all async operations (such as opening the Realm with `Realm.open`) will fail when a non-fatal error, such as a timeout, occurs.  
```

kneth (approved):

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -159,6 +169,7 @@ export function toBindingSyncConfig(config: SyncConfiguration): binding.SyncConf
     customHttpHeaders,
     flxSyncRequested: !!flexible,
     ...parseClientResetConfig(clientReset, onError),
+    cancelWaitsOnNonfatalError: cancelWaitOnNonFatalError,
```
kraenhansen, line 172: Is this difference in name intentional? Could it be the same to avoid confusion?

```suggestion
    cancelWaitsOnNonfatalError: cancelWaitOnNonFatalError,
```

papafe: It's intentional, I thought that it sounded a little better without the "s", but I don't have a strong opinion about, I'll change it

integration-tests/tests/src/utils/open-realm.ts
papafe: Added a note, thanks!

integration-tests/tests/src/utils/open-realm.ts
```diff
@@ -38,43 +38,61 @@ export async function openRealm(
...
-    return { config, realm };
+  // TODO: This should probably be done in Realm.open()
+  if (config.sync?.flexible) {
+    await realm.subscriptions.waitForSynchronization();
   }
+
+  return { config, realm };
+}
+
+export function createSyncConfig(partialConfig: SyncedConfiguration = {}, user: User): Configuration {
```
elle-j, line 70: If this function, as well as `createLocalConfig`, are only used in this file you can remove `export`. If they're needed outside of this file, you can add `/** @internal */` above their declarations 👍

elle-j (approved):
Great, thank you 🙏

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -115,6 +115,7 @@ export type BaseSyncConfiguration = {
   /** @internal */
   _sessionStopPolicy?: SessionStopPolicy;
   clientReset?: ClientResetConfig;
+  cancelWaitsOnNonFatalError?: boolean;
```
elle-j, line 118: Question is if we want to use `Nonfatal` (instead of `NonFatal`) like Core instead?

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -159,6 +169,7 @@ export function toBindingSyncConfig(config: SyncConfiguration): binding.SyncConf
     customHttpHeaders,
     flxSyncRequested: !!flexible,
     ...parseClientResetConfig(clientReset, onError),
+    cancelWaitsOnNonfatalError: cancelWaitsOnNonFatalError,
```
kraenhansen, line 172: I agree with @elle-j here - I was actually referring to `Nonfatal` vs `NonFatal` too, we should pick one or upstream an issue / PR to rename it in core if we don't like the name.

CHANGELOG.md
takameyer: > methods to asynchronously waiting for all downloads/uploads

We have `realm.syncSession.downloadAllServerChanges` and `realm.syncSession.uploadAllLocalChanges`.  Each take a timeout as a param

takameyer: https://www.mongodb.com/docs/realm-sdks/js/latest/Realm.App.Sync.Session.html

packages/realm/src/app-services/SyncConfiguration.ts
papafe: This was not exposed before, so we can do as we want with naming. I thought that `Nonfatal` seems an error compared to `NonFatal` if we want to keep camel case. I'm not strongly opposed though 😁

integration-tests/tests/src/utils/open-realm.ts
papafe: At the moment `createSyncConfig` is used in another file, while `createLocalConfig` isn't. I have extracted these two methods from the bigger method that was there before that was also creating a realm, as sometimes we just need a configuration. For this reason I have decided to export both anyways, as I expect they could be useful in some cases. 

Regarding `@internal`, this is the integration-test package, I suppose we don't need anything here, because this is not exposed to final users, am I wrong?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -124,15 +125,23 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+        //This needs to be revisited when:
+        //- https://github.com/realm/realm-core/issues/6301 is fixed. At the moment core rejects all the "non sensible" timeouts, and we need to set a very low one to simulate reliably a connection issue
+        //- We expose `SyncClientConfig.timeouts`(https://github.com/realm/realm-core/blob/e83515f014ff53b0d27c59b73773d33b4252e891/src/realm/object-store/sync/sync_manager.hpp#L97) so that we can set the timeouts properly.
+        const config = createSyncConfig({ sync: { cancelWaitsOnNonFatalError: true, flexible: true } }, this.user);
+
+        const openRealm = async () => {
+          await Realm.open(config);
+        };
+
+        await expect(openRealm()).to.be.rejected;
```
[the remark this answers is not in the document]
papafe: I think that the message here is not the important part. I've decided to keep it out also because the message itself could change when all the preconditions I've talked about are satisfied and we decide to test this properly. At the moment this is more of a "reminder" to add a test at some point when we can actually do it properly.

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -242,10 +253,21 @@ function parseRecoverOrDiscardUnsyncedChanges(clientReset: ClientResetRecoverOrD
...
+    onError,
+    customHttpHeaders,
+    clientReset,
+    flexible,
+    cancelWaitOnNonFatalError,
+  } = config;
 
   assert.instanceOf(user, User, "'user' on realm sync configuration");
+  if (cancelWaitOnNonFatalError != undefined) {
+    assert.boolean(cancelWaitOnNonFatalError, "`cancelWaitOnNonFatalError` on sync configuration");
```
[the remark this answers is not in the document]
papafe: Good eye!

packages/realm/src/app-services/App.ts
```diff
@@ -137,7 +142,7 @@ export class App {
   constructor(configOrId: AppConfiguration | string) {
     const config: AppConfiguration = typeof configOrId === "string" ? { id: configOrId } : configOrId;
     assert.object(config, "config");
-    const { id, baseUrl, app } = config;
+    const { id, baseUrl, app, timeout } = config;
     assert.string(id, "id");
```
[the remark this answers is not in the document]
papafe: Yes, it was missing so I wasn't validating anything. But I see your point that it makes sense to start somwehre.

integration-tests/tests/src/tests/sync/flexible.ts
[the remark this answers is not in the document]
elle-j: Alright :) Was thinking more so that we can be more certain that it's throwing because of the right reason. But because of several preconditions like you said, it may be harder.

integration-tests/tests/src/utils/open-realm.ts
elle-j: I stand corrected! :) Annotation not needed in the integration tests 👍

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -124,15 +125,23 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
-      flexible: true,
-    },
-  });
 
   describe("Configuration", function () {
     describe("flexible sync Realm config", function () {
+      it.skip("respects cancelWaitOnNonFatalError", async function () {
+        this.timeout(2_000);
+        //This needs to be revisited when:
+        //- https://github.com/realm/realm-core/issues/6301 is fixed. At the moment core rejects all the "non sensible" timeouts, and we need to set a very low one to simulate reliably a connection issue
```
takameyer, line 134: Is there a ticket for us to remember to revisit this?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -124,15 +125,23 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
-    schema: [FlexiblePersonSchema, DogSchema],
-    sync: {
-      flexible: true,
-    },
-  });
 
   describe("Configuration", function () {
     describe("flexible sync Realm config", function () {
+      it.skip("respects cancelWaitOnNonFatalError", async function () {
+        this.timeout(2_000);
```
takameyer, line 132: Is there a reason to shorten the timeout here?

takameyer (approved):
LGTM

packages/realm/src/app-services/SyncConfiguration.ts
elle-j: `Nonfatal` is not grammatically wrong, but I agree that `NonFatal` reads better, we also tend to use `non-<some word>` in our docs so using uppercase `F` seems consistent with that at least. Regarding Core, they use different spellings around the code base (e.g. `non_null`, `nonnull`, `NonNull`).

packages/realm/src/app-services/App.ts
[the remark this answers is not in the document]
papafe: Added

integration-tests/tests/src/tests/sync/flexible.ts
papafe: The default behaviour is to not cancel on non fatal errors (like a timeout). So in case the test fails the sync client will continue to retry connecting forever. And if we kept the 60 seconds timeout the test will wait the full 60 seconds before failing if the test doesn't pass.

packages/realm/src/app-services/SyncConfiguration.ts
kraenhansen: I don't care either way, I think they should just be the same externally as internally to avoid confusion (which is more important than a grammar nit in my opinion).

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: Great reason then 👍🏼

takameyer:
We should also signal to the documentation team to add this to the relevant pages.

packages/realm/src/app-services/SyncConfiguration.ts
papafe: Mmmh, I don't know if I agree with this. I think that the core team has much more leeway in the naming of the variables, given that they don't need to expose those to developers, and can more or less decide internally how they want to do it. 
Instead we should try to keep somehow consistent how we want to name our variables and should not depend on how core wants to do it. The reason why we have this "problem" is because the binding generator is creating the variable names based on core, but that's something that we decided to put in between. Besides, we are not even forced to keep the same name that is coming from core (like Swift, that decide to change it). In this case I thought the name made sense (even though it's kinda long) and I just kept it more or less.

papafe:
> We should also signal to the documentation team to add this to the relevant pages.
> 
> Also is this being depicted in our tsdoc? There should be some internal API documentation as well. I'm still a bit fuzzy about how this is done in `bindgen`

This is a good point. It seems that in the bindgen repo the docs are just declared together with the rest, but they seem to be missing from the realm configuration types. @kraenhansen @elle-j do you maybe know more about this? Are they somewhere else or just missing entirely for now?

packages/realm/src/app-services/SyncConfiguration.ts
kraenhansen: Your decision 🙂

kraenhansen (approved):

kraenhansen:
> they seem to be missing from the realm configuration types

We still have a task outstanding of going though the jsdocs and migrating documentation into our tsdoc strings through all types and APIs. That's probably why they're missing. We've been trying to move them as we did the implementation, but we've missed a few places.

papafe:
I've created a follow up task in order to remember to re-enable the test when the prerequisites are met: https://github.com/realm/realm-js/issues/5509

PR #5497, opened 2023-02-24
Adding Browser/WASM template for binding generator

nhachicha:
WIP adding browser template for binding gen.

- Build the browser target.
```
cd packages/realm
./build-browser-target.sh
```

- Run the sample react app
```
cd packages/web-example
npm install
npm start 
```
observe the output in the Dev console


packages/bindgen/CMakeLists.txt
```diff
@@ -193,14 +194,38 @@ if(DEFINED CMAKE_JS_VERSION)
...
+    )
+
+    target_sources(realm-js PRIVATE browser_init.cpp ${CMAKE_JS_SRC} ../../src/browser/platform.cpp)
+
+    add_custom_command(
+        TARGET realm-js-wasm
+        COMMAND ${CMAKE_COMMAND} -E copy
+                ${CMAKE_CURRENT_BINARY_DIR}/realm-js-wasm.js ${CMAKE_CURRENT_BINARY_DIR}/realm-js-wasm.wasm
+                ${TYPESCRIPT_OUTPUT_DIR}        
+    )
```
RedBeard0531, line 227: For the other targets, I've been having them produce their output in the correct output directory directly rather than copying. We should be consistent and either do the same here, or switch the others to copy. I think you need to do something like this change I recently made to node.cmake: https://github.com/realm/realm-js/commit/74f5e2ec6a3ea2b4304ab648c52138879d0b2635#diff-0e67af0e05d8343adec84207c67cc80a294337638fd517c415b7f1b5d72212d0R19

packages/bindgen/package.json
```diff
@@ -13,7 +13,8 @@
     "test": "mocha",
     "generate:spec-schema": "typescript-json-schema tsconfig.json RelaxedSpec --include src/spec/relaxed-model.ts --out generated/spec.schema.json --required",
     "build": "cmake-js build --debug",
-    "rebuild": "cmake-js rebuild --debug"
+    "rebuild": "cmake-js rebuild --debug",
+    "build-browser": "emcmake cmake -B build.browser -G Ninja && cd build.browser && ninja"
```
RedBeard0531, line 17:
```suggestion
    "build-browser": "emcmake cmake -B build.browser -G Ninja && cmake --build build.browser"
```

There are some benefits to using `cmake --build` if we switch to using `-G "Ninja Multi-Config" which I think we will want to (but not quite yet).

packages/bindgen/spec.yml
```diff
@@ -287,12 +287,12 @@ enums:
...
-    values:
-      - NEED_DATA
-      - HAVE_EVENT
-      - HAVE_ERROR
+  # WatchStreamState:
+  #   cppName: app::WatchStream::State
+  #   values:
+  #     - NEED_DATA
+  #     - HAVE_EVENT
+  #     - HAVE_ERROR
```
RedBeard0531, line 295: Why? This is needed for watch support.

Actually, we should look into reverting all of the changes in this file.

packages/bindgen/CMakeLists.txt
fealebenpae, line 227:
```suggestion
    set_target_properties(realm-js-wasm PROPERTIES
        RUNTIME_OUTPUT_DIRECTORY ${TYPESCRIPT_OUTPUT_DIR}
    )
```
This should achieve the same effect, without hardcoding the output files explicitly. You might need to set `LIBRARY_OUTPUT_DIRECTORY` also, but I think `RUNTIME` should be sufficient.

packages/bindgen/CMakeLists.txt
```diff
@@ -193,14 +194,38 @@ if(DEFINED CMAKE_JS_VERSION)
...
+        COMMAND ${NODE_FORCE_COLOR} ${CMAKE_CURRENT_SOURCE_DIR}/realm-bindgen.ts ${SPEC_ARGS} --template browser --output ${CMAKE_CURRENT_BINARY_DIR}
+        VERBATIM
+        MAIN_DEPENDENCY src/templates/browser.ts
+        DEPENDS
+            ${SPEC_FILE}
+            SpecJsonSchema
+            ${TS_FILES}
+    )
+
+    target_sources(realm-js PRIVATE browser_init.cpp ${CMAKE_JS_SRC} ../../src/browser/platform.cpp)
```
fealebenpae, line 220:
```suggestion
    target_sources(realm-js PRIVATE browser_init.cpp ../../src/browser/platform.cpp)
```
`CMAKE_JS_SRC` is only relevant to Node.js.

packages/bindgen/CMakeLists.txt
```diff
@@ -193,14 +194,38 @@ if(DEFINED CMAKE_JS_VERSION)
         OUTPUT node_init.cpp
         COMMAND ${NODE_FORCE_COLOR} ${CMAKE_CURRENT_SOURCE_DIR}/realm-bindgen.ts ${SPEC_ARGS} --template node --output ${CMAKE_CURRENT_BINARY_DIR}
         VERBATIM
-        MAIN_DEPENDENCY src/templates/node.ts
+        MAIN_DEPENDENCY src/templates/browser.ts
```
fealebenpae, line 197: Should this be reverted?

packages/bindgen/src/realm_js_browser_helpers.h
```diff
@@ -0,0 +1,32 @@
...
+    
+    return out;
+}
+REALM_NOINLINE inline void toEmscriptenException(const std::exception_ptr& e) noexcept
+{
+    try {
+        std::rethrow_exception(e);
+    }
+    catch (const std::exception& e) {
+        return emscripten::val(e.what()).throw_();
```
RedBeard0531, line 25: While this might work for now, we should at least add a `// TODO throw an actual Error object`. Maybe that is easy to do with `EM_JS`?

packages/bindgen/src/templates/jsi.ts
```diff
@@ -287,7 +287,7 @@ function convertPrimToJsi(addon: JsiAddon, type: string, expr: string): string {
     case "std::exception_ptr":
       return `toJsiException(_env, ${expr})`;
     case "std::error_code":
-      return `toJsiErrorCode(_env, ${expr})`;
+      return `toJsiException(_env, ${expr})`;
```
RedBeard0531, line 290: ???

I would be surprised if this actually compiles.

packages/bindgen/src/templates/browser.ts
```diff
@@ -0,0 +1,897 @@
...
+
+import { doJsPasses } from "../js-passes";
+
+const emscripten_call_args = new CppVar("const emscripten::val", "args");
+
+function tryWrap(body: string) {
+  return `try {
+                ${body}
+            } catch (const std::exception& ex) {
+                emscripten::val(ex.what()).throw_();
```
fealebenpae, line 43: We shouldn't be throwing raw strings as errors, but that can be changed to use Core's unified exception handling later.

packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -130,20 +143,25 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
...
     // that depend on the symbol not changing.
     const symb = `_${cls.rootBase().jsName}_Symbol`;
 
     if (!cls.base) {
       // Only root classes get symbols and constructors
-      both(`const ${symb} = Symbol("Realm.${cls.jsName}.external_pointer");`);
+      all(`const ${symb} = Symbol("Realm.${cls.jsName}.external_pointer");`);
+      browserLines.push(
+        `const _${cls.jsName}_registery = new FinalizationRegistry(nativeModule.${cls.jsName}_deleter);`,
+      );
```
RedBeard0531, line 157: I think you need to do the finalization registry for every class, not just base classes. It will be a bit tricky though because you need to only do the registration in the most-derived class's constructor. I suggest making the constructors look something like this:

```js
constructor(ptr, calledFromSubclass) {
  ${cls.base ? 'super(ptr, true)', `this[${symb}] = ptr;`}
  if (!calledFromSubclass) _${cls.jsName}_registery.register(this, ptr);
}
```

I doubt that the extra branch matters much, but if we want to avoid it where we know it isn't needed, we can do `${cls.subclasses.length == 0 ? '' : 'if (!calledFromSubClass)} _${cls.jsName}_registery.register(this, ptr);`. Might give us a slight size savings, and isn't too much extra complication, so might be worth it.

packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -166,27 +186,36 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
         ...params,
         asyncSig ? "_cb" : [],
       ].flat();
-      let call = `${native}(${args})`;
+      let call = `${native}([${args}])`;
       if (asyncSig) call = `_promisify(_cb => ${call})`;
-      body += `
+      const promisify_definition = `
```
RedBeard0531, line 191: This should be called `method_definition` or just `definition`, since it isn't always promisified.

packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -130,20 +143,25 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
...
+      );
 
-      body += `constructor(ptr) { this[${symb}] = ptr};`;
+      body += `constructor(ptr) { this[${symb}] = ptr; };`;
+      body_browser += `constructor(ptr) { this[${symb}] = ptr; _${cls.jsName}_registery.register(this, ptr);};`;
     }
 
     // This will override the extractor from the base class to do a more specific type check.
-    body += `
+    const extract_definition = `
```
RedBeard0531, line 164: Use `camelCase` in js code.

packages/realm/rollup.config.mjs
```diff
@@ -92,6 +93,38 @@ export default [
...
+      format: "es",
+      sourcemap: true,
+    },
+    plugins: [
+      nodeResolve({
+        mainFields: ["browser", "module", "main"],
+        resolveOnly: ["@realm/network-transport", "path-browserify"],
+      }),
+      // We need to use `commonjs` because of "path-browserify"
+      commonjs(),
```
RedBeard0531, line 109: It seems confusing that we need to use `commonjs()` when doing `format: "es"`, which I think is the opposite of cjs. @kraenhansen can you look into this and suggest either removing it or wording for a comment that explains why this makes sense even when not output cjs?

packages/realm/rollup.config.mjs
```diff
@@ -92,6 +93,38 @@ export default [
...
+        },
+      }),
+      typescript({
+        tsconfig: "src/browser/tsconfig.json",
+        noEmitOnError: true,
+        outputToFilesystem: true,
+      }),
+      copy({
+        targets: [{ src: "./generated/ts/realm-js-wasm.wasm", dest: "./dist/" }],
+      }),
```
RedBeard0531, line 124: Can we instead just generate to the right place? I think this file is now being copied twice after generation which seems silly.

src/browser/platform.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+    throw std::runtime_error("Realm for browser does not support this method.");
+}
+void JsPlatformHelpers::print(const char* fmt, ...)
+{
+    va_list vl;
+    va_start(vl, fmt);
+    printf(fmt, vl);
+    va_end(vl);
+}
```
RedBeard0531, line 62: I think you can delete this, since it is only used by other C++ code that isn't used on the browser.

packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -59,8 +62,16 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
     // We know that node always has real WeakRefs so just use them.
     export const WeakRef = global.WeakRef;
   `);
+  browserLines.push(`
+    /*global window, FinalizationRegistry*/
+    import Module from "./realm-js-wasm.js";
+    const nativeModule = await Module(); // loading WASM 
+    nativeModule.browserInit();
```
fealebenpae, line 69: I think this is a good spot to inject things like `UUID` and `Decimal128` on `Module` and recall them with `val::module_property()` in C++ instead of relying on the injectables mechanic.

RedBeard0531 (commented):
Posting initial feedback before reviewing the main browser.ts template file. I also skipped over the code in packages/realm since it looks like it isn't actually intended to be merged, and web-example since I know basically nothing about react.

@kraenhansen does it make sense to commit this change to package-lock.json or should we revert it for now. It seems like it is a full rewrite of that file. I never know how to handle that and usually just `git checkout package-lock.json` before committing.

packages/bindgen/src/templates/node-wrapper.ts
RedBeard0531: I'd rather we stay consistent between the engines for now, unless there is a good reason to diverge. Is there some benefit to injecting them this way rather than using `injectInjectables`?

RedBeard0531:
@nhachicha Is it possible to revert your bump of the core submodule for now? I think we are planning on doing the bump soon in the bindgen branch anyway, so I'd rather not do it here too A) so that we don't hit an extra merge conflict and B) to keep this review focused on what is needed to add emscripten/browser support.

packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -0,0 +1,221 @@
...
+      all(`const ${native} = nativeModule.${method.id};`);
+      // TODO consider pre-extracting class-typed arguments while still in JIT VM.
+      const asyncSig = method.sig.asyncTransform();
+      const params = (asyncSig ?? method.sig).args.map((a) => a.name);
+      const args = [
+        method.isStatic ? [] : `this[${symb}]`, //
+        ...params,
+        asyncSig ? "_cb" : [],
+      ].flat();
+      let call = `${native}([${args}])`;
```
RedBeard0531, line 189: I didn't notice this change at first. Please change this back to just `(${args})`. We shouldn't be making a new JS Array every time we call a native function!

nhachicha: It's easier to access the args consistently this way in the generated C++ (we don't have a JS engine abstraction like in node and JSI that facilitate accessing the args). Also passing var args is not supported AFAIK from the `Embind` perspective.

packages/bindgen/src/templates/node-wrapper.ts
nhachicha: `browserInit();` exists solely to compensate for the lack of "engine registration phase" (i.e `NODE_API_NAMED_ADDON(realm_cpp, RealmAddon)` or  `realm_jsi_init` _which is called from_ `Java_io_realm_react_RealmReactModule_install`) which will trigger the creation of the `RealmAddon` instance.

Registering the JS type should still be done via `injectInjectables` IMO to keep the separation of concerns ...

packages/bindgen/src/templates/browser.ts
nhachicha: I'll add a TODO to use unified Core exception 👍

nhachicha:
@RedBeard0531 regarding 

> Posting initial feedback before reviewing the main browser.ts template file. I also skipped over the code in packages/realm since it looks like it isn't actually intended to be merged, and web-example since I know basically nothing about react.
> 
> @kraenhansen does it make sense to commit this change to package-lock.json or should we revert it for now. It seems like it is a full rewrite of that file. I never know how to handle that and usually just `git checkout package-lock.json` before committing.

note that I `npm ejected` the sample app that @kraenhansen created to access the webpack configuration which allows us to work around an issue see  https://github.com/realm/realm-js/pull/5497/files#diff-37e3ecdc2a16d68439cd2d7f3cbf0c0e14b2ac20fc650cb0f1764768e264c921R191-R193 
(adding `experiments: { topLevelAwait: true}`)
This has caused all the extra files you see here (and the update to the `package-lock`)

packages/bindgen/src/templates/node-wrapper.ts
RedBeard0531: You don't need to use varargs in C++. When generating the C++, we know exactly how many arguments are going to be passed, so we can just generate a function taking that many args. You can choose between naming the args `arg1, arg2,....` so you just need to change `arg[${i}]` to `arg${i}`, or you can just use the arg names from the spec (`a.name`, with `'_self'` for the `this` parameter in non-static methods). Slight preference for using the meaningful names, but not a big deal either way.

nhachicha:
@RedBeard0531 

> @nhachicha Is it possible to revert your bump of the core submodule for now? I think we are planning on doing the bump soon in the bindgen branch anyway, so I'd rather not do it here too A) so that we don't hit an extra merge conflict and B) to keep this review focused on what is needed to add emscripten/browser support.

As discussed on slack, we should probably keep this PR as a feature branch until we have Emscripten support merged in Core/master.

packages/bindgen/src/templates/node-wrapper.ts
RedBeard0531: You won't be able to check the number of passed args for error handling, but that is probably fine. I think we can get rid of that everywhere since it is probably unnecessary now that we have the generated JS code to ensure that we always pass the correct number of arguments (although some may be `undefined`)

packages/realm/rollup.config.mjs
nhachicha: With the change introduced in https://github.com/realm/realm-js/pull/5497#discussion_r1121464986 the `realm-js-wasm.was & realm-js-wasm.js` will be generated from the `bindgen` without an extra copy step in cmake. These files will be under `packages/realm/generated` which is already used by other templates (`node`/`react-native`). 
Rollup is already bundling `realm-js-wasm.js` but it's not aware of the wasm file, hence this manual copy to `dist`

packages/bindgen/src/templates/jsi.ts
nhachicha: typo sorry

nhachicha:
superseded by https://github.com/realm/realm-js/pull/5525

PR #5498, opened 2023-02-24
[bindgen] `browser` target and web example app

kraenhansen:
## What, How & Why?

- Adds a web (a name we've historically picked to avoid confusing with a data browser like Realm Studio) / browser target.
- Instantiates a create-react-app typescript template as a package in the mono repository (automagically sets up linking to ease development).
- Creates a `binding-mock.mjs` which should be deleted and the rollup config updated to point to the proper output from the binding generator.

The app builds and lunches, the SDK browser target also builds, but is reporting a lot of missing exports from the mocked binding.


nhachicha:
This can be closed now, as it's merged in https://github.com/realm/realm-js/pull/5525

PR #5501, opened 2023-02-26
fix call to Platform.isPad and isTV

levipro:
## What, How & Why?
Fix error `TypeError: false is not a function, js engine: hermes`

Platform.isPad and isTV were being called as a function but both are booleans https://reactnative.dev/docs/platform#ispad-ios

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -160,9 +160,9 @@ module.exports = {
         let deviceName = "unknown";
         let deviceVersion = "unknown";
         if (Platform.OS === "ios") {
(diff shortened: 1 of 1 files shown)


kneth:
@levipro Thank you for taking the first to find the root cause. It looks like we have fixed it at the same time: https://github.com/realm/realm-js/pull/5491

PR #5507, opened 2023-02-27
Support SSL Sync Configuration

### LINKED ISSUE #5485: Implement SSL Sync config
We most likely have to implement (and test) passing SSL configuration to the sync client.

This is how it's currently being transformed:

* [https://github.com/realm/realm-js/blob/bindgen/src/js_sync.hpp#L1433-L1451](https://github.com/realm/realm-js/blob/bindgen/src/js_sync.hpp#L1433-L1451)
* [https://github.com/realm/realm-js/blob/bindgen/src/js_sync.hpp#L482-L572](https://github.com/realm/realm-js/blob/bindgen/src/js_sync.hpp#L482-L572)

One alternative to this is to mark them deprecated on v11 and adopt platform networking (where Core would rely on the platforms SSL config) instead.

elle-j:
## What, How & Why?

When opening a synced realm, users can provide an SSL configuration containing either:
* (A) A callback function used to accept or reject the server's "SSL" (TLS) certificate.
* (B) A path to a trust certificate to hand off the validation.
* (C) Choose not to validate the server's certificate (intended for non-production use).

Examples:

```ts
// (A) Custom validation.
const ssl: SSLConfiguration = {
  validate: true, // (Default)
  validateCertificates: (verifyObject: SSLVerifyObject) => {
    
    // Inspect the SSLVerifyObject and validate certificate..

    // Accept server's certificate:
    return true;

    // Or reject it:
    // return false;
  },
};

// (B) Hand off the validation.
const ssl: SSLConfiguration = {
  validate: true,
  certificatePath: "path/to/trust/certificate",
};

// (C) Do not validate.
const ssl: SSLConfiguration = {
  validate: false,
};

  schema: [MySchema],
  sync: {
    ssl,
    // ...
  },
});
```

This closes #5485.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests (Located in [PR 5694](https://github.com/realm/realm-js/pull/5694))


packages/bindgen/src/realm_js_helpers.h
```diff
@@ -254,6 +254,14 @@ struct Helpers {
...
         ws.feed_buffer({ buffer.data(), buffer.size() });
     }
+
+    static auto make_ssl_verify_callback(std::function<bool(const std::string& server_address, int server_port,
+                                         std::string_view pem_data, int preverify_ok, int depth)> callback) {
+        return [callback = std::move(callback)](const std::string& server_address, uint16_t server_port,
+                                                const char* pem_data, size_t pem_size, int preverify_ok, int depth) {
+            return callback(server_address, server_port, { pem_data, pem_size }, preverify_ok, depth);
+        };
+    }
```
elle-j, line 264: 1. [line 258] Take the JS-wrapped version of the callback as the argument. This callback will (when invoked) call the user-provided callback for validating certificates.
2. [line 260] Return the CPP (Core-friendly) version of the callback (stored in the binding).
3. [line 262] When the CPP callback is invoked, invoke the JS-wrapped callback and return its result.

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -148,20 +182,31 @@ export type SyncConfiguration = FlexibleSyncConfiguration | PartitionSyncConfigu
...
+    clientValidateSsl: ssl?.validate,
+    sslTrustCertificatePath: ssl?.certificatePath,
+    sslVerifyCallback: ssl?.validateCertificates
+      ? binding.Helpers.makeSslVerifyCallback(toSSLVerifyCallbackWithListArguments(ssl.validateCertificates))
+      : undefined,
     ...parseClientResetConfig(clientReset, onError),
   };
 }
 
+/** @internal */
```
elle-j, line 204: On second thought, don't think this annotation should be needed since the function is not exported. However, there are several non-exported functions in this file using this annotation.

packages/bindgen/spec.yml
```diff
@@ -632,6 +637,9 @@ classes:
       # This is added due to DescriptorOrdering not being exposed
       get_results_description: '(results: const Results&) -> StringData'
       feed_buffer: '(ws: WatchStream&, buffer: BinaryData)'
+      # Converts char* and length to a combined string_view argument to callback.
+      # TODO: Consider making preverify_ok a bool.
```
takameyer, line 641: I really dislike the name `preverify_ok`, but I think that's out of our hands, or?

takameyer (approved):
LGTM.  Would be great to see some tests, but I think that's not so straightforward.  I assume you have tested this manually?

packages/bindgen/spec.yml
elle-j: Yeah that's for Core 👍

elle-j:
@takameyer, you can look at the test file of the [separated wip PR](https://github.com/realm/realm-js/pull/5694). We decided to split up the PRs as with the Jira issues (v12 and final v12). Implementation and tests can only be confirmed once that PR is opened.

packages/bindgen/src/realm_js_helpers.h
```diff
@@ -265,6 +266,16 @@ struct Helpers {
...
+
+    static auto make_ssl_verify_callback(std::function<bool(const std::string& server_address, int server_port,
+                                                            std::string_view pem_data, int preverify_ok, int depth)>
+                                             callback)
+    {
+        return [callback = std::move(callback)](const std::string& server_address, uint16_t server_port,
+                                                const char* pem_data, size_t pem_size, int preverify_ok, int depth) {
+            return callback(server_address, server_port, std::string_view(pem_data, pem_size), preverify_ok, depth);
+        };
+    }
```
elle-j, line 278: 1. [signature]: Take the JS-wrapped version of the callback as the argument. This callback will (when invoked) call the user-provided callback for validating certificates.
2. [1st `return`]: Return the CPP (Core-friendly) version of the callback (stored in the binding).
3. [2nd `return`]: When the CPP callback is invoked, invoke the JS-wrapped callback and return its result.

takameyer:
Getting the following error when building locally and on the `darwin arm64` build job in GitHub Actions:
```
clang: error: unable to execute command: Segmentation fault: 11
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.3.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
clang: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /var/folders/r7/nrrb917s7859md1_13pxplwm0000gp/T/node_init-5555d2.cpp
clang: note: diagnostic msg: /var/folders/r7/nrrb917s7859md1_13pxplwm0000gp/T/node_init-5555d2.sh
clang: note: diagnostic msg: Crash backtrace is located in
clang: note: diagnostic msg: /Users/andrew.meyer/Library/Logs/DiagnosticReports/clang_<YYYY-MM-DD-HHMMSS>_<hostname>.crash
clang: note: diagnostic msg: (choose the .crash file that corresponds to your crash)
clang: note: diagnostic msg: 

********************
```

takameyer (changes requested):
We need to fix the build issues before this is merged.

packages/bindgen/spec.yml
```diff
@@ -427,6 +427,11 @@ records:
       custom_http_headers: 
         type: 'std::map<std::string, std::string>'
         default: {}
+      client_validate_ssl:
+        type: bool
+        default: true
+      ssl_trust_certificate_path: util::Optional<std::string>
+      ssl_verify_callback: Nullable<SSLVerifyCallback>
```
kneth, line 434: Should it be `off_thread`?

RedBeard0531: It is. See line 642.

elle-j: Not sure if it can be marked as `off_thread` here.

As this is the `SSLVerifyCallback`:

```yml
SSLVerifyCallback:
  cppName: std::function<SyncConfig::SSLVerifyCallback>
```

And having `ssl_verify_callback` as:

```yml
SyncConfig:
  fields:
    # ...
    ssl_verify_callback: Nullable<SSLVerifyCallback>
```

It's unsure how/if it should be applied. These are 2 examples of other how other methods use it:

```yml
should_compact_on_launch_function: 'Nullable<std::function<(total_bytes: uint64_t, used_bytes: uint64_t) off_thread -> bool>>'
error_handler: 'Nullable<std::function<(session: SharedSyncSession, error: SyncError) off_thread -> void>>'
```

We do however need it for our helper `make_ssl_verify_callback`:

```yml
classes:
  Helpers:
    staticMethods:
      # ...
      make_ssl_verify_callback: '(callback: (server_address: const std::string&, server_port: int, pem_data: std::string_view, preverify_ok: int, depth: int) off_thread -> bool) -> SSLVerifyCallback'
```

But @RedBeard0531 could chime in here 🙁

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -106,12 +106,77 @@ export type ClientResetConfig =
...
+
+type SSLVerifyCallbackWithListArguments = (
+  serverAddress: string,
+  serverPort: number,
+  pemCertificate: string,
+  preverifyOk: number, // Acts as `SSLVerifyObject.acceptedByOpenSSL`
+  depth: number,
+) => boolean;
+
+export type SSLVerifyObject = {
```
kraenhansen, line 144: Not for this PR, but I think `SSLVerificationRequest` would probably be a better name for this.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -286,11 +286,22 @@ export class SyncSession {
   // TODO: Figure out a way to avoid passing a mocked app instance when constructing the User.
   get config(): SyncConfiguration {
     const user = new User(this.internal.user, mockApp);
-    const { partitionValue, flxSyncRequested, customHttpHeaders } = this.internal.config;
+    const { partitionValue, flxSyncRequested, customHttpHeaders, clientValidateSsl, sslTrustCertificatePath } =
```
kraenhansen, line 289: Does the `clientValidateSsl` and `sslTrustCertificatePath` have sane defaults when coming from core? (i.e. not just an empty string or something).

kraenhansen (approved):

packages/realm/src/app-services/SyncConfiguration.ts
elle-j: Or perhaps `Response` instead of `Request`? Since this is the object coming from the server.

We could definitely rethink this for v13 to not break the [earlier type](https://github.com/realm/realm-js/blob/a956a74a0845df99b6a4322627bc1c88a11ba749/types/index.d.ts#L133) for v12.

kraenhansen: > rethink this for v13 to not break the [earlier type](https://github.com/realm/realm-js/blob/a956a74a0845df99b6a4322627bc1c88a11ba749/types/index.d.ts#L133) for v12.

Yep :shipit:

packages/realm/src/app-services/SyncSession.ts
elle-j: This will be whatever the user has provided as their `validate` (binding: `clientValidateSsl`) and `certificatePath` (binding: `sslTrustCertificatePath`).

```ts
export type SSLConfiguration = {
  validate?: boolean;
  certificatePath?: string;

  // ...
};
```

If you don't provide a value for `certificatePath`, the value can be undefined as in [this test](https://github.com/realm/realm-js/blob/244d77ddac0abf55a4a53691d8c184e6940b6677/integration-tests/tests/src/node/ssl.ts#L99). `validate` defaults to `true`.

PR #5512, opened 2023-02-28
Fixed some types definition in bindgen

papafe:
Fixed some types definition on bindgen branch so that methods won't give type errors when being used with `RealmObject` with or without specifying the concrete type, similar to what has been done in https://github.com/realm/realm-js/pull/5487 with `Object.linkingObjects`


packages/realm/src/Realm.ts
```diff
@@ -18,6 +18,7 @@
 
 import {
   AggregatePipelineStage,
+  AnyRealmObject,
```
elle-j, line 21: Perhaps not a part of this PR, but should this also be exported to users via named export? Looks like it's missing from `packages/realm/src/index.ts`. @kraenhansen

elle-j (approved):
LGTM

packages/realm/src/Realm.ts
elle-j: Ah okay, even if the public APIs use that type in the signatures?

PR #5513, opened 2023-02-28
allow useQuery to filter or sort a collection by using a callback

### LINKED ISSUE #5471: Add argument to `useQuery` to allow result to be filtered and sorted
### Problem

Currently when using `useQuery`, the result must be passed into a `useMemo` in order to get a memoized filtered or sorted result.  This is also not particularly useful, as when the result of `useQuery` is re-rendered, the memo is called again as well as any filtered or sorted method calls.

For example

```tsx
  const user = useUser();
  const userId = user.id;
  const result = useQuery(Task);

  const tasks = useMemo(() => result.filtered(“ownerId == $0”, userId).sorted('createdAt'), [result, userId]);
```

### Solution

A more elegant solution would be to have a callback within `useQuery` which would allow the filtered and sorted functions to be applied within `useQuery` itself.  It could also take a dependency array to determine when to reset the filtered or sorted method calls.

```tsx
 const tasks = useQuery("Task", (tasks) => {
   return tasks.filtered(“ownerId == $0”, userId);
 }, [userId]);
```

### Alternatives

_No response_

Would be a major improvement

Local Database only

takameyer:
## What, How & Why?
This adds the ability to filter and sort the return value of `useQuery`, before it arrives in the component.  This ensures that `filtered` and `sorted` are not called again when `useQuery` triggers a re-render.

This closes #5471

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/example/app/AppSync.tsx
+++ b/example/app/AppSync.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect, useMemo} from 'react';
+import React, {useCallback, useEffect} from 'react';
 import {useApp, useUser} from '@realm/react';
 import {Pressable, StyleSheet, Text} from 'react-native';
 
@@ -15,15 +15,21 @@ export const AppSync: React.FC = () => {
   const realm = useRealm();
   const user = useUser();
   const app = useApp();
-  const result = useQuery(Task);
-
-  const tasks = useMemo(() => result.sorted('createdAt'), [result]);
+  const [showDone, setShowDone] = React.useState(false);
+  const tasks = useQuery(
+    Task,
+    collection =>
+      showDone
+        ? collection.sorted('createdAt')
+        : collection.filtered('isComplete == false').sorted('createdAt'),
+    [showDone],
+  );
 
   useEffect(() => {
     realm.subscriptions.update(mutableSubs => {
       mutableSubs.add(result);
     });
-  }, [realm, result]);
+  }, [realm, tasks]);
 
   const handleLogout = useCallback(() => {
     user.logOut();
@@ -32,7 +38,12 @@ export const AppSync: React.FC = () => {
   return (
     <>
       <Text style={styles.idText}>Syncing with app id: {app.id}</Text>
-      <TaskManager tasks={tasks} userId={user?.id} />
+      <TaskManager
+        tasks={tasks}
+        userId={user?.id}
+        setShowDone={setShowDone}
+        showDone={showDone}
+      />
       <Pressable style={styles.authButton} onPress={handleLogout}>
         <Text
           style={styles.authButtonText}>{`Logout ${user?.profile.email}`}</Text>
--- a/example/app/AppSync.tsx
+++ b/example/app/AppSync.tsx
@@ -27,7 +27,7 @@ export const AppSync: React.FC = () => {
 
   useEffect(() => {
     realm.subscriptions.update(mutableSubs => {
-      mutableSubs.add(result);
+      mutableSubs.add(tasks);
     });
   }, [realm, tasks]);
 
--- a/example/app/AppSync.tsx
+++ b/example/app/AppSync.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect} from 'react';
+import React, {useCallback, useEffect, useState} from 'react';
 import {useApp, useUser} from '@realm/react';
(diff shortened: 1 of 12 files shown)


example/app/AppSync.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect, useMemo} from 'react';
+import React, {useCallback, useEffect} from 'react';
```
elle-j, line 1: Only if you want to omit `React.` on line 18.
```suggestion
import React, {useCallback, useEffect, useState} from 'react';
```

example/app/components/TaskManager.tsx
```diff
@@ -28,7 +31,10 @@ export const TaskManager: React.FC<{
       // of sync participants to successfully sync everything in the transaction, otherwise
       // no changes propagate and the transaction needs to start over when connectivity allows.
       realm.write(() => {
-        return new Task(realm, description, userId);
+        return realm.create(Task, {
+          description,
+          userId: userId ? userId : 'blah',
```
elle-j, line 36: (super minor) Only if we want `'blah'` when `userId` is `undefined`/`null` (not falsy)
```suggestion
          userId: userId ?? 'blah',
```

packages/realm-react/CHANGELOG.md
```diff
@@ -7,24 +7,36 @@
...
-const SomeComponent = () => {
-	const items = useQuery(Item)
+  Example:
+	```tsx
+	const SomeComponent = () => {
+	  const user = useUser();
+		const items = useQuery(Item,
+			(res) => res.filtered(`owner_id == "${user?.id}"`).sorted('createdAt'),
+			 [user]
+		);
```
elle-j, line 19: Added closing brace and made indents consistent.
```suggestion
	const SomeComponent = () => {
	    const user = useUser();
	    const items = useQuery(Item,
	        (res) => res.filtered(`owner_id == "${user?.id}"`).sorted('createdAt'),
	        [user]
	    );
	};
```

packages/realm-react/README.md
```diff
@@ -155,10 +155,11 @@ import {useQuery} from '@realm/react';
 const Component = () => {
   // ObjectClass is a class extending Realm.Object, which should have been provided in the Realm Config.
   // It is also possible to use the model's name as a string ( ex. "Object" ) if you are not using class based models.
-  const collection = useQuery(ObjectClass);
-
-  // The methods `sorted` and `filtered` should be wrapped in a useMemo.
-  const sortedCollection = useMemo(() => collection.sorted(), [collection]);
+  const sortedCollection = useQuery(ObjectClass, (collection) => {
+    // The methods `sorted` and `filtered` should be passed as a `query` function
+    // any variables that are dependencies for this should be placed in the dependency array
```
elle-j, line 160:
```suggestion
    // The methods `sorted` and `filtered` should be passed as a `query` function.
    // Any variables that are dependencies of this should be placed in the dependency array.
```

packages/realm-react/src/__tests__/useQueryRender.test.tsx
```diff
@@ -405,7 +405,7 @@ describe.each`
         await new Promise((resolve) => setTimeout(resolve, 10));
         const id = i;
         testRealm.write(() => {
-          testRealm.create(Item, { id, name: `${id}` }, Realm.UpdateMode.Modified);
+          return new Item(testRealm, { id, name: `${id}` });
```
elle-j, line 408: Would it make sense to test both using `new` and `create` you think?

elle-j (approved):
Nice, added minor things 👍

example/app/components/TaskManager.tsx
takameyer: Oh whoops.  I was definitely should make that a nicer string :D

packages/realm-react/src/__tests__/useQueryRender.test.tsx
takameyer: That should be tested in the integration tests.  I'm only concerned about the hooks in these tests, so i'll just stick to our current best practices (which is to use `create`

elle-j: 🙂

takameyer:
Merging this, but up for retrospective comments later.

PR #5520, opened 2023-03-02
Remove Proxy Symbol Workaround for Flexible Sync

takameyer:
## What, How & Why?
According to the workaround comments, the symbol injection for detecting proxy collections in flexible sync is no longer needed.  I have removed the workaround and successfully tested it in the example application.  This PR reverts the changes from #4541.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen:
If there's no harm in keeping it in, I'd rather that we don't touch `master` unless we really need to backport some bugfix. This feels like a simplification/ enhancement which does have some potential to break stuff if it doesn't work on some ancient version of an os / configuration.

takameyer:
@kraenhansen It's currently blocking me

kraenhansen (approved):

kraenhansen (changes requested):
Are the deletion of `lib/mutable-subscription-set.js` intentional? With that we can't release a v11?

takameyer:
> Are the deletion of lib/mutable-subscription-set.js intentional?

Yes, it was added only for this symbol hack. See https://github.com/realm/realm-js/pull/4541/files

kraenhansen (approved):

PR #5524, opened 2023-03-02
Fixed some types definition in bindgen

papafe:
Fixed some types definition on bindgen branch so that methods won't give type errors when being used with RealmObject with or without specifying the concrete type, similar to what has been done in https://github.com/realm/realm-js/pull/5487 with Object.linkingObjects

(had to close the previous PR because I had some issues with rebasing, sorry for that)


takameyer (changes requested):
I just tested this out and it looks as if this fixes using a class with `realm.create`.  Could we write a test for this to avoid a future regression?

packages/realm/src/Realm.ts
```diff
@@ -786,13 +787,17 @@ export class Realm {
...
     mode: UpdateMode.All | UpdateMode.Modified | boolean,
   ): RealmObject<T> & T;
-  create<T extends RealmObject>(type: Constructor<T>, values: RealmInsertionModel<T>, mode?: UpdateMode.Never): T;
-  create<T extends RealmObject>(
+  create<T extends AnyRealmObject = RealmObject & DefaultObject>(
+    type: Constructor<T>,
+    values: RealmInsertionModel<T>,
+    mode?: UpdateMode.Never,
+  ): T;
+  create<T extends AnyRealmObject = RealmObject & DefaultObject>(
```
kraenhansen, line 795: Is the default really needed here? I imagine it'll be better to leave it out and rely on type inference from the first argument.

packages/realm/src/Realm.ts
```diff
@@ -889,11 +894,11 @@ export class Realm {
    * @since 0.14.0
    */
   objectForPrimaryKey<T = DefaultObject>(type: string, primaryKey: T[keyof T]): (RealmObject<T> & T) | null;
-  objectForPrimaryKey<T extends RealmObject = RealmObject & DefaultObject>(
+  objectForPrimaryKey<T extends AnyRealmObject = RealmObject & DefaultObject>(
```
kraenhansen, line 897: Same as my comment above on the default value here.

papafe:
> I just tested this out and it looks as if this fixes using a class with `realm.create`. Could we write a test for this to avoid a future regression?
@takameyer I've discussed about this with @kraenhansen, and probably the best approach here would be to enable the tests to fail when there are type errors, but we can't do it now without fixing all the pre-existing errors too. 
I've modified the class definition for `Person` and `Dog`, this will definitely give a type error in `create` and the other methods if we have a regression, so that will be covered when at some point type errors will be considered as test fails.

PR #5525, opened 2023-03-02
Nh/wasm/rebased bindgen

nhachicha:
Rebase of https://github.com/realm/realm-js/pull/5497 on top of current `bindgen` branch


packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -25,12 +25,15 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
   const spec = doJsPasses(bindModel(rawSpec));
   const reactLines = [];
   const nodeLines = [];
-  function both(content: string) {
+  const browserLines = [];
```
kraenhansen, line 28: I don't like this pattern and the way it's going. This file should be refactored before this PR lands to separate node from react-native. @RedBeard0531 we've talked about this before and I believe you had a plan for it?

Issue #3593, opened 2021-02-15
Web: Named "AppTypes" generics on the App constructor


kraenhansen:
## Goals

In it's current for the `App` constructor takes generic type arguments enabling developers to "lock down" the type of the "functions factory", user custom data and profile data. As this pattern grows however, it's annoying for developers to have to provide these as positional type arguments to the `App` class. Instead we should support "named" generic type aka. a single `AppTypes` generic type which enumerates the individual types used throughout the app.

Below is a proff of concept:

```typescript
type BaseFunctionsFactory = {
    callFunction(name: string, args: unknown[]): Promise<unknown>;
};

type DefaultFunctionsFactory = BaseFunctionsFactory & {
    [name: string]: (...args: unknown[]) => Promise<unknown>;
};

type BaseAppTypes = {
    FunctionsFactory: unknown,
    CustomUserDataType: unknown,
    ProfileDataType: unknown
};

type AppTypes<T extends Partial<BaseAppTypes>> = T & BaseAppTypes;

declare class User<T extends BaseAppTypes> {
    get functions(): T["FunctionsFactory"];
    get data(): T["CustomUserDataType"];
}

class App<T extends BaseAppTypes> {
    currentUser: User<T> | null = null;
}

// The user can declare their own app's types, wrapping them in AppTypes to allow defaults to be applied for (sub)types that is not declared.
type MyAppTypes = AppTypes<{
    FunctionsFactory: {
        echo: () => Promise<void>
    },
}>;

const app = new App<MyAppTypes>();
if (app.currentUser) {
    app.currentUser.functions.echo();
}
```

PR #5526, opened 2023-03-02
Get realm/react to build in bindgen

### LINKED ISSUE #5490: Fix Example App in Bindgen Branch
The `example/` application needs to be updated to adhere to the migration to npm workspaces and the restructuring of the `realm` project.

takameyer:
## What, How & Why?
Get realm/react build in bindgen:
* Update typings to pass for both v11 and bindgen
* Update realm to accept generics for User and App

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm/src/app-services/App.ts
```diff
@@ -187,24 +189,24 @@ export class App {
     return new EmailPasswordAuth(internal);
   }
 
-  public get currentUser(): User | null {
+  public get currentUser(): User<FunctionsFactoryType, CustomDataType> | null {
```
kraenhansen, line 192: Just a thought: Instead of typing out this user type multiple times, it could make sense to add another type argument to `App` based on the other arguments, to make a local `AppUser` type alias.

I'm thinking something like: https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwKrJDAPAFQD54BvAKHgvgAcYB7AMwEsIQAuePAblIF9TTQSOImip4AQSpV8AGglTM2eAF4MWXISJlK8AK7r2kqopjc+QA

Again - just a thought.

kraenhansen (approved):

kraenhansen:
Great idea to isolate the helpers 👍 Perhaps those functions deserve a comment on their origin?

packages/realm/src/app-services/App.ts
takameyer: I do think that makes sense, but is that a breaking change?  I would have to play it out.  Perhaps my assumptions are wrong or I'm not completely getting the picture

elle-j (approved):
LGTM 🚀 😃

packages/realm/src/app-services/App.ts
RedBeard0531: Why are we only supplying 2 of the 3 parameters to `User<...>`? Maybe `App` should be templated on `UserType extends AnyUser = User` instead those details and just return `UserType` here?

packages/realm-react/src/cachedCollection.ts
```diff
@@ -73,15 +74,15 @@ type CachedCollectionArgs<T> = {
...
+export function createCachedCollection<T extends Realm.Object<any>>({
   collection,
   realm,
   updateCallback,
   updatedRef,
   objectCache = new Map(),
   isDerived = false,
-}: CachedCollectionArgs<T>): { collection: Realm.Collection<T>; tearDown: () => void } {
-  const cachedCollectionHandler: ProxyHandler<Realm.Collection<T & Realm.Object>> = {
+}: CachedCollectionArgs<T>): { collection: Realm.Results<T> | Realm.List<T>; tearDown: () => void } {
```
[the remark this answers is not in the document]
kraenhansen: It could use OrderedCollection but I think one of Andrew's goals are to keep the package backwards compatible with v11 too.

packages/realm-react/src/useObject.tsx
```diff
@@ -31,23 +27,25 @@ type PrimaryKey = Parameters<typeof Realm.prototype.objectForPrimaryKey>[1];
  * @returns useObject - Hook that is used to gain access to a single Realm object from a primary key
  */
 export function createUseObject(useRealm: () => Realm) {
-  return function useObject<T>(
+  function useObject<T>(type: string, primaryKey: T[keyof T]): (T & Realm.Object<T>) | null;
+  function useObject<T extends Realm.Object>(type: { new (...args: any): T }, primaryKey: T[keyof T]): T | null;
```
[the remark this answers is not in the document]
kraenhansen: I'd be interested to know the reason for that. I know you hit this with the bindings types, but when would you expect end-users to get a benefit from us using `abstract new`?

packages/realm/src/app-services/App.ts
kraenhansen: > Maybe `App` should be templated on `UserType`

(How) is that different from what I suggested above?

packages/realm-react/src/useObject.tsx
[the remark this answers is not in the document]
RedBeard0531: well for one thing, Realm.Object should be an `abstract class`. And to some extent, once we disallow `new T()`, so should all classes inheriting from it.

packages/realm/src/app-services/App.ts
RedBeard0531: If I understand correctly, you are proposing to add an additional parameter. I am instead suggesting to replace all of the parameters with a single `UserType` parameter.

kraenhansen: Okay, yeah - that would be a breaking change though, which we should probably postpone to v13?

We're tracking a proposal to collapse the type signature on `App` which feels related to me: https://github.com/realm/realm-js/issues/3593

PR #5528, opened 2023-03-02
Using `declare` in class-based models in the integration tests

kraenhansen:
## What, How & Why?

The update for the integration tests targeting `es2022` the default for [`useDefineForClassFields`](https://www.typescriptlang.org/tsconfig#useDefineForClassFields) changed to `true`, which [breaks creating objects via the constructor of a class-based model](https://github.com/realm/realm-js/issues/5527).

### DIFF
--- a/packages/babel-preset/index.js
+++ b/packages/babel-preset/index.js
@@ -1,6 +1,6 @@
 ////////////////////////////////////////////////////////////////////////////
 //
-// Copyright 2021 Realm Inc.
+// Copyright 2023 Realm Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


takameyer (approved):
LGTM.  Is this a breaking change for users though?  We might want to update our documentation to use `declare` if that's important.

packages/babel-preset/index.js
```diff
@@ -0,0 +1,37 @@
...
+/* eslint-env node */
+
+module.exports = () => {
+  return {
+    // TODO: Add @realm/babel-plugin here
+    overrides: [
+      {
+        plugins: [
+          [
+            require("@babel/plugin-transform-flow-strip-types"),
```
RedBeard0531, line 28: Shouldn't we use the TS plugin rather than flow? https://babeljs.io/docs/babel-plugin-transform-typescript#allowdeclarefields

RedBeard0531 (changes requested):

packages/babel-preset/index.js
kraenhansen: I thought so too, but that's not the one that complained though ... 🤔

kraenhansen: This override is patching the plugin which is pulled in via https://github.com/realm/realm-js/blob/bindgen/integration-tests/environments/react-native/babel.config.js#L19 (which I expect any React Native app to use) .. but we should probably assert that it's already loaded as a plugin. To be honest I find the Babel config override and preset APIs are super confusing and would love some help on that at some point.

kraenhansen:
> Is this a breaking change for users though?

I wouldn't say so. This is a breaking change introduced when targeting `es2022`. We should probably find a solution to work around it (which is why I created https://github.com/realm/realm-js/issues/5527). But for now, I wanted to just update the integration tests to use it.

> We might want to update our documentation to use declare if that's important.

Yes - I think that would be a good change to make. But that will require users to apply our Babel preset (introduced in this PR). We should definitely document that .. if / when we choose to publish it on NPM.

kraenhansen:

PR #5529, opened 2023-03-02
Allow optional primary keys

kraenhansen:
## What, How & Why?

The schema parser was implemented to reject optional primary keys, but that's allowed by core and was supported on v11.


elle-j (approved):
LGTM! :)

kneth (approved):
Realm Java and Realm Kotlin permit it (but as I recall Realm Swift doesn't), and I think we should permit it too.

PR #5530, opened 2023-03-02
Fetch `statusText` polyfill

kraenhansen:
## What, How & Why?

This provides a polyfill for the `statusText` on a `Response` returned by `fetch`.

The React Native fetch polyfill fails to provide this.
See https://github.com/react-native-community/fetch/issues/14.

### DIFF
--- a/packages/realm-network-transport/src/DefaultNetworkTransport.ts
+++ b/packages/realm-network-transport/src/DefaultNetworkTransport.ts
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import { makeRequestBodyIterable } from "./IterableReadableStream";
+import { deriveStatusText } from "./status-text";
 import type {
   NetworkTransport,
   Request,
@@ -35,6 +36,29 @@ export class DefaultNetworkTransport implements NetworkTransport {
   public static DEFAULT_HEADERS = {
     "Content-Type": "application/json",
   };
+
+  private static createTimeoutSignal(timeoutMs: number | undefined) {
+    if (typeof timeoutMs === "number") {
+      const controller = new DefaultNetworkTransport.AbortController();
+      // Call abort after a specific number of milliseconds
+      const timeout = setTimeout(() => {
+        controller.abort();
+      }, timeoutMs);
+      return {
+        signal: controller.signal,
+        cancelTimeout: () => {
+          clearTimeout(timeout);
+        },
+      };
+    } else {
+      return {
+        signal: undefined,
+        cancelTimeout: () => {
+          /* No-op */
+        },
+      };
+    }
+  }
 
   constructor() {
     if (!DefaultNetworkTransport.fetch) {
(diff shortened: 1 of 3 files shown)


packages/realm-network-transport/src/DefaultNetworkTransport.ts
```diff
@@ -36,6 +37,29 @@ export class DefaultNetworkTransport implements NetworkTransport {
...
+    } else {
+      return {
+        signal: undefined,
+        cancelTimeout: () => {
+          /* No-op */
+        },
+      };
+    }
+  }
+
```
kraenhansen, line 62: I just moved this up to be a `static` member, since it didn't need access to the instance.

packages/realm-network-transport/src/status-text.ts
```diff
@@ -0,0 +1,77 @@
...
+  501: "Not Implemented",
+  502: "Bad Gateway",
+  503: "Service Unavailable",
+  504: "Gateway Timeout",
+  505: "HTTP Version Not Supported",
+  506: "Variant Also Negotiates",
+  507: "Insufficient Storage",
+  508: "Loop Detected",
+  510: "Not Extended",
+  511: "Network Authentication Required",
```
kraenhansen, line 72: See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

takameyer: Can you add that as a comment in the code?

packages/realm-network-transport/src/status-text.ts
```diff
@@ -0,0 +1,77 @@
...
+  409: "Conflict",
+  410: "Gone",
+  411: "Length Required",
+  412: "Precondition Failed",
+  413: "Payload Too Large",
+  414: "URI Too Long",
+  415: "Unsupported Media Type",
+  416: "Range Not Satisfiable",
+  417: "Expectation Failed",
+  418: "I'm a teapot",
```
takameyer, line 55: 🫖

takameyer (approved):
Nice find!  LGTM

packages/realm-network-transport/src/status-text.ts
kraenhansen: Done 👍

packages/realm-network-transport/src/status-text.ts
elle-j: > The HTTP 418 I'm a teapot client error response code indicates that the server refuses to brew coffee because it is, permanently, a teapot. A combined coffee/tea pot that is temporarily out of coffee should instead return 503.

-- [Hyper Text Coffee Pot Control Protocol](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418)

packages/realm/src/app-services/User.ts
```diff
@@ -268,11 +268,11 @@ export class User<
       serviceName,
     );
 
-    const { body, ok, status, statusText } = await network.fetch(request);
-    assert(ok, `Request failed: ${statusText} (${status})`);
-    assert(body, "Expected a body in the response");
+    const response = await network.fetch(request);
+    assert(response.ok, () => `Request failed: ${response.statusText} (${response.status})`);
```
elle-j, line 272: Could you fill me in on why the 2nd arg needs to be a function here?

elle-j (approved):
LGTM!

packages/realm/src/app-services/User.ts
kraenhansen: It doesn't actually need to, but I figured it made sense to compute the error message only when the assertion has actually failed. In general - this overload can be used to avoid generating the error message if the assertion doesn't fail.

elle-j: Thanks for explanation 👍

PR #5535, opened 2023-03-03
Fix dynamic primary key initialization

takameyer:
## What, How & Why?
Default functions weren't working for primary key.  This fixes the issue.

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


integration-tests/tests/src/tests/schema.ts
```diff
@@ -44,6 +44,32 @@ describe("Realm schema", () => {
...
+                type: "int",
+                default: () => 42,
+              },
+            },
+          },
+        ],
+      });
+
+      const test = realm.write(() => {
+        return realm.create<Test>("Test", {});
```
kraenhansen, line 70: This won't fail your test (although it should).

```suggestion
        return realm.create<Test>("Test", { dynamic: 0 });
```

takameyer: Why not?  it doesn't equal 42

packages/realm/src/Object.ts
```diff
@@ -185,7 +185,11 @@ export class RealmObject<T = DefaultObject> {
     const table = binding.Helpers.getTable(realm.internal, tableKey);
     if (primaryKey) {
       const primaryKeyHelpers = properties.get(primaryKey);
-      const primaryKeyValue = values[primaryKey];
+      const defaultValue = primaryKeyHelpers.default;
+      let primaryKeyValue = values[primaryKey];
+      if (!primaryKeyValue && typeof defaultValue !== "undefined") {
```
kraenhansen, line 190: I'd say this would render incorrect behaviour if called with a falsy value (`0`, `false`, `""`, etc.), in which case the default should not override.

We could either replace this with a check for undefined (which I think is the current behaviour on v11):

```suggestion
      if (primaryKeyValue === undefined && typeof defaultValue !== "undefined") {
```

Or alternatively a check to see if the value was provided in the values (as an enumerable property) - but that's probably a bit less flexible on callers:

```suggestion
      if (!(primaryKey in values) && typeof defaultValue !== "undefined") {
```

kraenhansen (changes requested):

integration-tests/tests/src/tests/schema.ts
takameyer: image
With your change it fails

kraenhansen: Thats unrelated (and because you're missing a call to `Realm.clearTestState`).

kraenhansen: What I wanted to show was that a falsy value would be replaced by the default.

takameyer: I ran the single test, but good call in adding `Realm.clearTestState`

integration-tests/tests/src/tests/schema.ts
```diff
@@ -44,6 +44,32 @@ describe("Realm schema", () => {
...
+            primaryKey: "dynamic",
+            properties: {
+              dynamic: {
+                type: "int",
+                default: () => 42,
+              },
+            },
+          },
+        ],
+      });
```
kraenhansen, line 67: Any reason this can't use the `openRealmBefore` hook?

integration-tests/tests/src/tests/schema.ts
takameyer: Ah...you are right.  I added a test to check this as well with the change you provided.  Falsey values now work as well.

integration-tests/tests/src/tests/schema.ts
takameyer: I've refactored the tests

integration-tests/tests/src/tests/schema.ts
```diff
@@ -17,34 +17,47 @@
...
-              },
+    openRealmBefore({
+      schema: [
+        {
+          name: "Test",
+          primaryKey: "primary",
+          properties: {
+            primary: {
+              type: "int",
+              default: () => 42,
```
RedBeard0531, line 37: Can you make this something that returns different values on each call, like an incrementing counter? That would make sure we are actually dynamically calling it each time, rather than evaluating it once to pick a single default.

packages/realm/src/Object.ts
```diff
@@ -185,7 +185,11 @@ export class RealmObject<T = DefaultObject> {
     const table = binding.Helpers.getTable(realm.internal, tableKey);
     if (primaryKey) {
       const primaryKeyHelpers = properties.get(primaryKey);
-      const primaryKeyValue = values[primaryKey];
+      const defaultValue = primaryKeyHelpers.default;
+      let primaryKeyValue = values[primaryKey];
+      if (primaryKeyValue === undefined && typeof defaultValue !== "undefined") {
```
RedBeard0531, line 190: A few things:

1. we shouldn't waste time looking up `primaryKeyHelpers.default` unless `primaryKeyValue` is undefined.
2. why mix `=== undefined` with `typeof x !== "undefined"` style checks. We should probably have a rule about which is prefered and stick to it. (I propose `=== undefined` unless specially looking for a global variable, but even then I might prefer `globalThis.blah === undefined` for explicitness).
3. I think you can just remove `&& typeof defaultValue !== "undefined"` entirely without affecting the behavior of this code.

integration-tests/tests/src/tests/schema.ts
```diff
@@ -17,34 +17,47 @@
...
 
-      expect(test.dynamic).to.equal(42);
+      expect(test.primary).to.equal(42);
+      expect(test.value).to.equal(13);
+
+      const staticTest = realm.write(() => {
+        return realm.create<Test>("Test", { primary: 0, value: 0 });
+      });
+      expect(staticTest.primary).to.equal(0);
+      expect(staticTest.value).to.equal(0);
```
RedBeard0531, line 60: Consider introducing extra scoping blocks for each sub-test so that you can just name `staticTest` `test` (or better yet `obj` or `inserted`).

integration-tests/tests/src/tests/schema.ts
```diff
@@ -17,34 +17,47 @@
...
+          name: "Test",
+          primaryKey: "primary",
+          properties: {
+            primary: {
+              type: "int",
+              default: () => 42,
+            },
+            value: {
+              type: "int",
+              default: () => 13,
```
RedBeard0531, line 41: We should test simple value-defaults as well, both for PK and non-PK (since they unfortunately have separate code paths). Is that covered by other tests?

elle-j (approved):
LGTM!

integration-tests/tests/src/tests/schema.ts
elle-j: I agree, or we could have separate `it` tests instead.

integration-tests/tests/src/tests/schema.ts
takameyer: Done

packages/realm/src/Object.ts
takameyer: Good point.  I've simplified the logic to be a more straightforward

integration-tests/tests/src/tests/schema.ts
takameyer: I've split up the tests and also renamed `test` to `obj`

integration-tests/tests/src/tests/schema.ts
takameyer: It is covered now

PR #5539, opened 2023-03-04
Green builds on CI via NPM workspaces

kraenhansen:
## What, How & Why?

This is the first stab at making CI build again.


integration-tests/environments/react-native/package.json
```diff
@@ -50,7 +56,8 @@
     "test:ios": {
       "command": "npm run common",
       "dependencies": [
-        "pod-install"
+        "pod-install",
```
takameyer, line 59: I was looking at the `pod-install` script and noticed there are no `output`s defined.  Maybe we should define them to make local testing faster

takameyer (approved):
:shipit:

PR #5541, opened 2023-03-06
Fixing `@realm/common` React Native bundle

kraenhansen:
## What, How & Why?

This fix the bundling of `@realm/common`, `@realm/network-transport` and `realm` to prefer the "react-native" and "browser".


packages/realm-common/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "./dist/bundle.cjs.js",
   "module": "./dist/bundle.es.js",
   "types": "./dist/bundle.d.ts",
+  "react-native": "./dist/bundle.dom.es.js",
```
takameyer, line 8: Seems odd that react-native would source `dom` file.  Does this just happen to work because they are not using conflicting modules?

kraenhansen: Well - actually. Thinking more about this, I could use "browser" here instead.

takameyer (approved):

PR #5542, opened 2023-03-06
Replacing `turbo` with `wireit`

kraenhansen:
## What, How & Why?

Since our packages doesn't have the exact same scripts and interdependencies, I found `wireit` to be a lot easier to configure.


packages/realm-web/package.json
```diff
@@ -12,15 +12,31 @@
...
+      "files": ["rollup.config.mjs", "src/**/*.ts", "src/**/tsconfig.json", "tsconfig.json", "types"],
+      "output": ["dist"]
+    },
+    "generate-types": {
+      "command": "tsc --project src/dom/tsconfig.json --declaration --emitDeclarationOnly --declarationDir types/generated",
+      "dependencies": ["update-types"],
+      "files": ["src/**/*.ts", "src/**/tsconfig.json", "tsconfig.json"],
+      "output": ["types/generated"]
+    },
+    "update-types": {
```
takameyer, line 34: Should this have a dependency on `package/realm` types being generated?

packages/realm/package.json
```diff
@@ -49,14 +49,62 @@
...
+      "dependencies": [
+        "../realm-network-transport:build",
+        "build:bindgen"
+      ],
+      "files": ["rollup.config.mjs", "src/**/*.ts", "tsconfig.json", "src/**/tsconfig.json"],
+      "output": ["dist/bundle.*.js", "dist/bundle.*.mjs"]
+    },
+    "build:bindgen": {
+      "command": "npm run build --prefix ../bindgen",
+      "files": ["../../src/node/**", "../bindgen/src/**/*.ts", "../bindgen/src/**/*.h"],
```
takameyer, line 80: I guess this would be for a later cleanup task, but should we move anything we depend on to build for `node` out of `../../src/node`?

takameyer (commented):
Great work here.  I was having some thought's about how to really make the DX even simpler.
The idea would be that on a clean checkout `npm install` build absolutely everything.  This would require:
* preinstall in the root to `git submodule update --recursive --init`
* prepare scripts in all packages that run `build`

I think if all the dependencies are setup right, this should work.  Perhaps it's a bit much, but I think it's worth considering.

kraenhansen:
> The idea would be that on a clean checkout npm install build absolutely everything.

While this would be ideal for us developers, it will complicate our CI setup (I know you can --ignore-scripts, but that also skips scripts that for example downloads prebuilt binaries for some CLIs etc.).

How about not building on install but have a single "build" NPM script in the root package that depends on building all the other sub-packages? That way developer has to write (only) two commands to install and build while we can selectively invoke build, lint or test commands on CI?

takameyer:
> How about not building on install but have a single "build" NPM script in the root package that depends on building all the other sub-packages? 

This seems like a good compromise.  In any case, I think we should prioritize DX over CIX.  You can always program the CI to do whatever you want.  It's much more difficult to program a developers intuition

packages/realm-web/package.json
kraenhansen: The `update-types` script is copying the files from the `./types` of the root directory, so these are static files that isn't generated. This needs to be fixed at some point (as we want those files gone), but I'd say we should probably wait for that until we know the future of the WASM binding.

package.json
```diff
@@ -4,6 +4,22 @@
...
+      "dependencies": [
+        "./packages/realm:build",
+        "./packages/babel-plugin:build",
+        "./packages/mocha-reporter:build",
+        "./packages/realm-react:build",
+        "./packages/realm-tools:build",
+        "./packages/realm-web:build",
+        "./integration-tests/tests:build"
+      ]
+    }
```
kraenhansen, line 21: @takameyer this is the build script you requested. It only mentions the "leaf" packages as the other packages are transitively built.

takameyer (approved):
:shipit:

PR #5543, opened 2023-03-06
`@realm/react` don't run `tsc` on lint

kraenhansen:
## What, How & Why?

Running `tsc` as part of linting depends on `realm` already being built, which is unfortunate.
I removed this and added an option to fail the rollup on a TypeScript error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,13 +7,16 @@
       "name": "@realm/root",
       "license": "apache-2.0",
       "workspaces": [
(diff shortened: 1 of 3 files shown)


takameyer (commented):
Is there a way we can still do type checking?

kraenhansen:
It'll happen as you build the package.

takameyer (approved):
LGTM!

PR #5552, opened 2023-03-09
Renaming `master` → `main`

kraenhansen:
## What, How & Why?

This aims to replace all mentions of the `master` branch with `main` after our recent rename, in markdown files, workspaces and source-code.


takameyer (approved):
Great catch.  We may also want to signal to the documentation team to update any links that may be directed at master (maybe there are none, but I'm not sure).

Issue #5547, opened 2023-03-07
`npm install realm` broken? keeps hanging up at `reify:realm: timing...` (see screenshot)


GitMurf:
### How frequently does the bug occur?

Always

### Description

I am having trouble doing an npm install of current version as well as tried plenty of older versions as well. They all hang up at the screenshot below:

![image](https://user-images.githubusercontent.com/64155612/223328592-017c6bb0-9cb6-429e-855d-d2d37ff4d3c5.png)

### Can you reproduce the bug?

Always

### Reproduction Steps

`npm install realm`

### Version

Latest and 10.22

### What services are you using?

Local Database only

### Are you using encryption?

No

### Platform OS and version(s)

Windows 11

### Build environment

Which debugger for React Native: ..

GitMurf:
I don't know if this is it or not but when I kill the install it gives the following and appears it may be erroring out at `node scripts/submit-analytics.js`?

![image](https://user-images.githubusercontent.com/64155612/223329823-35773e3b-d4bb-46da-98c8-78a254870879.png)

kneth:
@GitMurf I am not sure if we will get any useful information but can you try to run `npm install -d verbose realm`?

Which version of node do you use? We have an unresolved [issue with node 19](https://github.com/realm/realm-js/issues/5136).

GitMurf:
> @GitMurf I am not sure if we will get any useful information but can you try to run `npm install -d verbose realm`?
> 
> Which version of node do you use? We have an unresolved [issue with node 19](https://github.com/realm/realm-js/issues/5136).

I am indeed using the "Current Version" v19.7 of node along with current version of NPM v9.6. Are there any solutions at this point other than downgrading to v18.xx?

GitMurf:
> @GitMurf I am not sure if we will get any useful information but can you try to run `npm install -d verbose realm`?

Here is the end of the verbose results. Again the only thing I can get out of it is that something seems to be hanging with the "submit-analytics.js" step?

![image](https://user-images.githubusercontent.com/64155612/223494332-4ebf0b28-d2ff-4cbe-b794-094dc28371fb.png)

kneth:
I have a fix in https://github.com/realm/realm-js/pull/5554

pachewise:
@kneth thanks for fixing this issue. Do y'all have a release planned soon for this, or will we have to build from source / use node@18 to be able to install realm?

kneth:
The plan is to get a release out this week

[...]

PR #5554, opened 2023-03-09
Prevent node 19 to hang under installation

### LINKED ISSUE #5136: node 19 hangs when installing Realm

Sometimes

### Description

Using node 19, `npm install realm` seems to hang.

### Stacktrace & log output

_No response_

Not yet

### Reproduction Steps

```sh
nvm use 19
mkdir foobar
cd foobar
npm init -y
npm install realm
```

### Version

11.2.0

Local Database only

### Are you using encryption?

No, not using encryption

MacOS, Linux

### Build environment

_No response_

### Cocoapods version

_No response_

kneth:
## What, How & Why?

node 19 changes default behaviour of https.get() so we set it to the old behaviour

This closes #5136 and #5547

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


scripts/submit-analytics.js
```diff
@@ -44,6 +44,7 @@
 const fs = require("fs");
 const fse = require("fs-extra");
 const path = require("path");
+const https = require("https");
```
elle-j, line 47: Should we localize this import to only `submitAnalytics()` instead?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 * It is not allowed to specify `deleteRealmIfMigrationIsNeeded` and sync. This can lead to error messages like `Schema validation failed due to the following errors`. ([#5548](https://github.com/realm/realm-js/issues/5548), v10.12.0)
+* Installation will hang when using Node 19. ([#5136](https://github.com/realm/realm-js/issues/5136), since v10.13.0)
```
elle-j, line 11: Possible reformulation:
```suggestion
* Installation will no longer hang when using Node 19. ([#5136](https://github.com/realm/realm-js/issues/5136), since v10.13.0)
```

scripts/submit-analytics.js
```diff
@@ -215,30 +216,9 @@ async function collectPlatformData(packagePath = getProjectRoot()) {
 }
 
 /**
- * Send collected analytics data to Realm's servers over HTTPS
- * @param  {Object} payload analytics info
+ * Collect and Send analytics data to MongoDB over HTTPS
```
elle-j, line 219:
```suggestion
 * Collect and send analytics data to MongoDB over HTTPS
```

scripts/submit-analytics.js
```diff
@@ -259,7 +239,25 @@ async function submitAnalytics(dryRun) {
     return;
   }
 
-  await Promise.all([dispatchAnalytics(payload)]);
+  return new Promise((resolve, reject) => {
+    // node 19 turns on keep-alive by default and it will make the https.get() to hang
+    // https://github.com/realm/realm-js/issues/5136
+    https.globalAgent = new https.Agent({ keepAlive: false });
```
elle-j, line 245: If we move the `https` import to this function, then it may be safer to override the `globalAgent`. E.g. thinking of a scenario where `https` is global in this file and by passing `{ keepAlive: false }` would set the other non-specified `option` values to their defaults right? (And perhaps override any changes made to these somewhere else in this file). Although, I assume this is an unlikely scenario. But localizing the `https` might still be favorable.

elle-j (approved):
No more infinite wait 🕐 😃

takameyer (approved):
LGTM.  Do we have tests activated for this?

scripts/submit-analytics.js
```diff
@@ -259,7 +239,25 @@ async function submitAnalytics(dryRun) {
...
+    const requestUrl = getAnalyticsRequestUrl(payload);
+
+    https
+      .get(requestUrl, (res) => {
```
nirinchev, line 249: Rather than overriding the global agent here, why not pass the custom agent to the `get` request options:

```suggestion
    const requestUrl = getAnalyticsRequestUrl(payload);

    https
      .get(requestUrl, { agent:  new https.Agent({ keepAlive: false }) }, (res) => {
```

kneth:
@takameyer 

> Do we have tests activated for this?

Only my local test. Please note that node 19 is not in tier 1.

scripts/submit-analytics.js
kneth: I have accepted the suggestion by @nirinchev which is basically doing what you are saying.

takameyer:
> Please note that node 19 is not in tier 1.

I think it would be great if we built a system of checking on the latest releases of the platforms we run on so that we are ahead of the breaking changes.  But that's something for another PR

fukemy:
can u fix for 11.0.0-rc.0?

PR #5550, opened 2023-03-08
Upgrade to Realm Core v13.6.0 (on `main`)

kraenhansen:
## What, How & Why?

This closes #5544 by following the PRs which upgraded Core on `v11` and migrating this to the new SDK.

All tests which previously passed are still passing (but there are still some outstanding test that run on `v11` which are not being imported, so we might want to fix those by picking the changes made to the tests in https://github.com/realm/realm-js/pull/5239)

PR #5565, opened 2023-03-15
Adopting to change in `SyncManager#path_for_realm`

kraenhansen:
## What, How & Why?

This fix the currently broken tests.
The upgrade of Core needs to merged first: https://github.com/realm/realm-js/pull/5550


elle-j (approved):
🎉

PR #5570, opened 2023-03-17
Bundle on CI

kraenhansen:
## What, How & Why?

Our v11 SDK didn't need a step to build / bundle TypeScript code, but our new SDK does.
This PR adds a "bundle" job to the PR workflow which will run the `bundle` script in our root package, which will in turn depend on the build and bundle scripts for all sub-packages producing outputs derived from TypeScript source-files.

This also adjusts the prepare and publish workflows for releasing the new SDK.

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -10,6 +10,31 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  bundle:
+    name: Bundle TypeScript
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+      - name: Setup node version
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: npm
+      - name: Install dependencies
+        # Ignoring scripts to prevent a prebuilt from getting fetched / built
+        run: npm ci --ignore-scripts
+      - name: Bundle all packages
+        run: npm run bundle
+      - name: Upload dist artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: realm-js-bundles
+          path: |
+            packages/*/dist
+
   build:
     name: Build for ${{ matrix.variant.os }} ${{ matrix.variant.arch }}
     runs-on: ${{ matrix.variant.runner }}
@@ -47,7 +72,6 @@ jobs:
         with:
           node-version: 16
           cache: npm
-          cache-dependency-path: package-lock.json
 
       - name: Get NPM cache directory
         id: npm-cache-dir
@@ -204,7 +228,7 @@ jobs:
 
   integration-tests:
     name: Integration tests (${{matrix.variant.target}}) for ${{ matrix.variant.environment }} on ${{ matrix.variant.os }}
-    needs: [build, ios-xcframework]
+    needs: [bundle, build, ios-xcframework]
     env:
       REALM_DISABLE_ANALYTICS: 1
       MOCHA_REMOTE_TIMEOUT: 60000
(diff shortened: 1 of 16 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -47,7 +75,6 @@ jobs:
         with:
           node-version: 16
           cache: npm
-          cache-dependency-path: package-lock.json
```
kraenhansen, line 50: This is the default value - so there's no need to provide it.

.github/workflows/pr-realm-js.yml
```diff
@@ -171,7 +198,7 @@ jobs:
         with:
           name: realm-js-prebuilds
           path: |
-            dependencies.list
+            README.md
```
kraenhansen, line 201: I think the dependencies.list will go away soon, so I don't want the upload to rely on its existence.

.github/workflows/publish-release.yml
```diff
@@ -81,7 +91,7 @@ jobs:
         secret_key: ${{ secrets.AWS_S3_SECRET_ACCESS_KEY }}
 
     - name: Upload prebuilds to S3
-      run: s3cmd put --recursive --acl-public prebuilds/realm-* s3://${{ secrets.PREBUILDS_S3_BUCKET_NAME }}/${{ steps.get-version.outputs.version }}/
+      run: s3cmd put --recursive --acl-public packages/realm/prebuilds/realm-* s3://${{ secrets.PREBUILDS_S3_BUCKET_NAME }}/${{ steps.get-version.outputs.version }}/
```
kraenhansen, line 94: These are now stored in `packages/realm`

.github/workflows/prepare-release.yml
```diff
@@ -41,13 +41,6 @@ jobs:
       - name: Update dependencies.list
         run: sed -i 's/^VERSION=.*/VERSION=${{ steps.update-changelog.outputs.new-version }}/' dependencies.list
 
-      - name: Update xcode project
```
takameyer, line 44: No longer necessary?

package.json
```diff
@@ -18,13 +19,19 @@
...
+        "bundle",
+        "./packages/realm:build"
+      ]
+    },
+    "bundle": {
+      "dependencies": [
+        "./packages/realm:bundle",
         "./packages/babel-plugin:build",
         "./packages/mocha-reporter:build",
+        "./packages/realm-app-importer:build",
```
takameyer, line 31: At some point we should align on the name of this script, as it is `bundle` in `realm` and `build` everywhere else.

packages/realm-react/package.json
```diff
@@ -17,7 +17,6 @@
       "command": "rollup --config",
       "dependencies": [
         "../realm:bundle",
-        "../realm:build:node",
```
takameyer, line 20: This would be a dependency for testing

takameyer (approved):
LGTM.  Just a few clarifying questions and remarks

.github/workflows/prepare-release.yml
kraenhansen: We don't have that file anymore.

package.json
kraenhansen: Yeah. I agree, its a little strange to have different names there. It's just that the convention I've seen elsewhere is "build", but we want to separate this in our packages.
We should probably rename this from "build" to "bundle" in the packages that doesn't have native code 👍

packages/realm-react/package.json
kraenhansen: Yeah. We should probably add the `test` script to `wireit` too 👍

package.json
takameyer: I think generally in pure JS produces, build means bundle.  I guess for our case, the term `bundle` is more transparant.

packages/realm-react/package.json
takameyer: It will happen soon, by me, when I get some realm/react work done

.github/workflows/prepare-release.yml
takameyer: Great!

PR #5572, opened 2023-03-18
Expose syncRootDirectory on App configuration

### LINKED ISSUE #5571: Expose `SyncClient.base_file_path` on `Realm.App`

kneth:
## What, How & Why?

This closes #5571

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests
*~ [ ] 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)~

* [x] jsdoc files updated


kraenhansen:
Are there any precedence for this in other SDKs? If not, I'd suggest we use a name closer to that of core. I.e. use "base" instead of "root" and "path" instead of "directory":

```
syncBasePath
```

kneth:
> Are there any precedence for this in other SDKs?

I have only checked Realm Java, so yes ;-)

integration-tests/tests/src/node/path.ts
```diff
@@ -36,24 +38,52 @@ const schema = {
...
     expect(realm.path).to.equal(filename);
     realm.close();
     Realm.deleteFile({ path: filename });
   });
 });
 
+describe.skipIf(environment.missingServer, `app configuration of root directory (flexible sync)`, async function () {
+  const { appId, baseUrl } = await importApp("with-db-flx");
+
+  it.only("directory and file created where expected", async function () {
```
elle-j, line 64: Is the `only` intentional? (Thinking it may have been intended to be used temporarily?)
```suggestion
  it("directory and file created where expected", async function () {
```

src/js_app.hpp
```diff
@@ -238,13 +252,9 @@ void AppClass<T>::constructor(ContextType ctx, ObjectType this_object, Arguments
...
 
-    auto realm_file_directory = default_realm_file_directory();
-    ensure_directory_exists_for_file(realm_file_directory);
-
-    SyncClientConfig client_config;
-    client_config.base_file_path = realm_file_directory;
-    client_config.metadata_mode = SyncManager::MetadataMode::NoEncryption;
-    client_config.user_agent_binding_info = get_user_agent();
+    // ensure_directory_exists_for_file(client_config.base_file_path);
+    util::try_make_dir(client_config.base_file_path);
```
elle-j, line 256: Should `ensure_directory_exists_for_file` be commented out? Is it more or less tested by `util::try_make_dir`?

elle-j (approved):
LGTM, minor comments 👍

integration-tests/tests/src/node/path.ts
```diff
@@ -36,24 +38,52 @@ const schema = {
...
+    const tmpdir = getAbsolutePath();
+    expect(fs.exists(tmpdir)).to.be.false;
+
+    const app = new Realm.App({ id: appId, baseUrl, syncRootDirectory: tmpdir });
```
takameyer, line 68: I would've expected `syncRootDirectory` to be set on the `Realm`.  Is that how other sdks are doing it as well? 

 Also, what would happen when the `path` is set on the `Realm` to something outside of the `syncRootDirectory`?

takameyer (commented):
Just a couple questions.  Is there a ticket for the docs team to update the mongodb docs?  Should we also update the jsdocs?

src/js_app.hpp
kneth: We don't need it - I'll remove it

integration-tests/tests/src/node/path.ts
kneth: > I would've expected syncRootDirectory to be set on the Realm. Is that how other sdks are doing it as well?

It is a configuration parameter for the sync client which is initialised when calling the `Realm.App` constructor. If on the Realm configuration, it will be too late. Moreover, the metadata directory is _per_ app and not _per_ Realm.

* [DotNet - baseFilePath](https://www.mongodb.com/docs/realm-sdks/dotnet/latest/reference/Realms.Sync.AppConfiguration.html#Realms_Sync_AppConfiguration_BaseFilePath)
* [Java - syncRootDirectory](https://www.mongodb.com/docs/realm-sdks/java/latest/io/realm/mongodb/AppConfiguration.Builder.html#syncRootDirectory(java.io.File))

> Also, what would happen when the path is set

For absolute path, the Realm file is stored outside the sync root directory and metadata within. For relative path, both metadata and Realm file will be stored relative to the sync root directory.

takameyer: Makes sense!  Thanks for the clarification

takameyer (approved):

kraenhansen (approved):

PR #5592, opened 2023-03-20
Download the assisted_agg binary

kraenhansen:
## What, How & Why?

This change downloads the assisted_agg binary as part of starting the server, simplifyiung starting the BaaS test server. We verified with the BaaS team that this is not considered confidential.


takameyer (approved):
:shipit:

PR #5595, opened 2023-03-20
Enable all tests after bad rebase

kraenhansen:
## What, How & Why?

When rebasing `bindgen` onto `main` we introduced a bad merge as tests on `v11` was enumerated from a different file than that on `bindgen`, which resulted in many tests not being enabled.

This PR corrects this and fix all the remaining failures.


packages/realm/src/TypeHelpers.ts
```diff
@@ -83,7 +83,10 @@ export type TypeOptions = {
 // "Only Realm instances are supported." (which should probably have been "RealmObject")
 // instead of relying on the binding to throw.
 export function mixedToBinding(realm: binding.Realm, value: unknown): binding.MixedArg {
-  if (typeof value === "undefined") {
+  if (typeof value === "string" || typeof value === "number" || typeof value === "boolean" || value === null) {
```
kraenhansen, line 86: There might be other types we should fast-track?

packages/realm/src/Configuration.ts
```diff
@@ -243,6 +243,7 @@ export function validateConfiguration(config: unknown): asserts config is Config
   }
   if (sync !== undefined) {
     assert(!onMigration, "The realm configuration options 'onMigration' and 'sync' cannot both be defined.");
+    assert(!inMemory, "The realm configuration options 'inMemory' and 'sync' cannot both be defined.");
```
elle-j, line 246: We need to check for `undefined` here since `inMemory` is a boolean. (The above `!onMigration` works because it's a function.)
```suggestion
    assert(inMemory === undefined, "The realm configuration options 'inMemory' and 'sync' cannot both be defined.");
```

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -80,7 +97,18 @@ export class ProgressRealmPromise implements Promise<Realm> {
...
         const { bindingConfig } = Realm.transformConfig(config);
+        // Construct an async open task
         this.task = binding.Realm.getSynchronizedRealm(bindingConfig);
+        // If the promise handle gets rejected, we should cancel the open task
+        // to avoid consuming a thread safe reference which is no longer registered
+        this.handle.promise.catch(() => {
+          if (this.task) {
+            this.task.cancel();
+          }
+        });
```
elle-j, line 108: An alternative (minor thing)
```suggestion
        this.handle.promise.catch(() => this.task?.cancel());
```

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -97,35 +125,18 @@ export class ProgressRealmPromise implements Promise<Realm> {
             }
             return realm;
           })
-          .then(this.handle.resolve, this.handle.reject);
+          .then(this.handle.resolve, (err) => {
+            assert.undefined(err.code, "Update this to use the error code instead of matching on message");
+            if (err instanceof Error && err.message === "Sync session became inactive") {
+              // This can happen when two async tasks are opened for the same Realm and one gets canceled
+              this.rejectAsCanceled();
```
elle-j, line 132: Interesting!

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -158,6 +171,48 @@ export class ProgressRealmPromise implements Promise<Realm> {
...
+      if (timeOutBehavior === OpenRealmTimeOutBehavior.ThrowException) {
+        // Make failing the timeout, reject the promise
+        this.timeoutPromise.catch(this.handle.reject);
+      } else if (timeOutBehavior === OpenRealmTimeOutBehavior.OpenLocalRealm) {
+        // Make failing the timeout, resolve the promise
+        this.timeoutPromise.catch((err) => {
+          if (err instanceof TimeoutError) {
+            const realm = new Realm(config);
+            this.cancelAndResetTask();
+            this.handle.resolve(realm);
```
elle-j, line 190: Question, does the realm need to be created before the call to `cancelAndResetTask()`?

packages/realm/src/PropertyMap.ts
```diff
@@ -31,7 +31,10 @@ export class PropertyMap {
   private objectSchemaName: string | null = null;
   private initialized = false;
   private mapping: Record<string, PropertyHelpers | undefined> = {};
-  private nameByColumnKey: Map<binding.ColKey, string> = new Map();
+  /**
+   * Note: Cannot key by the binding.ColKey directly, as this is `Long` on JSC (which does not compare equal)
```
elle-j, line 35: Also, could you clarify `which does not compare equal`?
```suggestion
   * Note: Cannot get key by the binding.ColKey directly, as this is `Long` on JSC (which does not compare equal)
```

packages/realm/src/TypeHelpers.ts
```diff
@@ -83,7 +83,10 @@ export type TypeOptions = {
...
+    // Fast track pass through for the most commonly used types
+    return value;
+  } else if (typeof value === "undefined") {
```
elle-j, line 89:
```suggestion
  } else if (value === undefined) {
```

packages/realm/src/assert.ts
```diff
@@ -63,6 +63,11 @@ assert.number = (value: unknown, target?: string): asserts value is number => {
   assert(typeof value === "number", () => new TypeAssertionError("a number", value, target));
 };
 
+assert.numericString = (value: unknown, target?: string): asserts value is number => {
```
elle-j, line 66: Is it intended for the value to be converted and used as a number after calling this function?

elle-j (approved):
LGTM with some suggestions and questions 🙂

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -586,41 +572,24 @@ describe("SessionTest", () => {
...
-        .then(
-          () => {
-            throw new Error("Download did not time out");
-          },
-          (e) => {
-            expect(e).equals("Downloading changes did not complete in 1 ms.");
-          },
-        );
+        }),
+      ).is.rejectedWith("Downloading changes did not complete in 1 ms.");
```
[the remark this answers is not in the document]
elle-j: Ah okay I see! (p.s. I just updated the first line in my suggestion by adding `await`)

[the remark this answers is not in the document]
kraenhansen: Argh ... I can't hold back, I had to park this on a separate PR to keep this one easier to review: https://github.com/realm/realm-js/pull/5691

integration-tests/tests/src/utils/mock-function.ts
```diff
@@ -0,0 +1,52 @@
...
+export function createMockFunction<ArgsType extends Array<unknown>, ReturnType>(
+  expectedCalls?: Array<ReturnType>,
+): MockFunction<ArgsType, ReturnType> {
+  function mock(...args: ArgsType): ReturnType {
+    mock.calls.push({ args });
+    mock.nextCall.resolve({ args });
+    // Create a new handle for the next promise
+    mock.nextCall = createPromiseHandle();
+    if (Array.isArray(expectedCalls)) {
+      if (mock.calls.length - 1 > expectedCalls.length) {
```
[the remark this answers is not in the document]
kraenhansen: Hmm. I think you're right - this shouldn't be `- 1` thanks. I should probably also ensure that's tested.

That being said, I actually ended up not using this anyway, so I'll move it to a different PR.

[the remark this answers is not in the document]
kraenhansen: I've parked in a different PR https://github.com/realm/realm-js/pull/5692

integration-tests/tests/src/utils/promise-handle.test.ts
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { expect } from "chai";
+import { createPromiseHandle } from "./promise-handle";
+
+describe("PromiseHandle", () => {
```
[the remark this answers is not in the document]
kraenhansen: I've added an asynchronous test for PromiseHandle 👍

packages/realm/src/ProgressRealmPromise.ts
kraenhansen: Not necessarily, I'd think the order doesn't matter. It would probably be cleaner to turn that around and group the creation of the `Realm` with resolving the `PromiseHandle` 👍 
```suggestion
            this.cancelAndResetTask();
            const realm = new Realm(config);
            this.handle.resolve(realm);
```

packages/realm/src/PropertyMap.ts
kraenhansen: I actually think "Cannot key by" is correct. It's talking about key'ing a record / object by some sort of value. At least that's an expression I've seen before 🙂
```suggestion
   * Note: Cannot key by the binding.ColKey directly, as this is `Long` on JSC (which does not pass equality checks like `bigint` does)
```

kraenhansen: > which does not compare equal

What I meant is that two instances of `Long` representing the same logical value are not `===` equal.

packages/realm/src/assert.ts
kraenhansen: Oh ... good catch. This was a copy-paste-o.
```suggestion
assert.numericString = (value: unknown, target?: string) => {
```

packages/realm/src/PropertyMap.ts
elle-j: Thanks 👍

packages/realm/src/assert.ts
elle-j: Okay np 🙂  It may still be helpful to have `asserts value is string` instead though?

packages/realm/src/assert.ts
```diff
@@ -63,6 +63,11 @@ assert.number = (value: unknown, target?: string): asserts value is number => {
   assert(typeof value === "number", () => new TypeAssertionError("a number", value, target));
 };
 
+assert.numericString = (value: unknown, target?: string) => {
```
kraenhansen, line 66:
```suggestion
assert.numericString = (value: unknown, target?: string): asserts value is string => {
```

packages/realm/src/assert.ts
kraenhansen: I agree 👍 and pushed a commit with that.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -317,7 +317,7 @@ describe.skipIf(environment.missingServer, "User", () => {
           .logIn(Realm.Credentials.emailPassword({ email: validEmail, password: validPassword }))
           .catch((err) => {
             expect(err.message).equals("invalid username/password");
-            expect(err.code).equals(50);
+            expect(err.code).equals(4349);
```
takameyer, line 320: Why did the error code change?  Was this a core upgrade?

takameyer (approved):
LGTM!

packages/realm/src/TypeHelpers.ts
elle-j: This looks good, which other types were you thinking of in that case?

integration-tests/tests/src/tests/sync/user.ts
kraenhansen: Yep.

takameyer: I guess that's a breaking change then :D

kraenhansen: Not as I see it. These are not documented, so technically not a part of our public API as per semantic versioning.

takameyer: Mmm...true.  Although if someone were to have code relying on this error code, it would need to be updated.

takameyer: In any case, this doesn't block this PR from being merged.  It was just a bit of a concern I had.

kraenhansen:
While this didn't fix all issues when running on CI it's better than what we had. I'll adress the remaining failures in followup PRs.

PR #5599, opened 2023-03-21
Add support for compensating writes

papafe:
This PR adds a new error class for compensating writes errors.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


packages/realm/src/errors.ts
```diff
@@ -138,3 +141,52 @@ export class ClientResetError extends SyncError {
...
+/**
+ * An error class that indicates that one or more object changes have been reverted by the server.
+ * This can happen when the client creates/updates objects that do not match any subscription, or performs writes on
+ * an object it didn't have permission to.
+ */
+export class CompensatingWriteError extends SyncError {
+  /**
+   * The array of compensating writes performed by the server.
+   */
+  public compensatingWrites: CompensatingWriteErrorInfo[] = [];
```
kraenhansen, line 154: This name seems a bit verbose.

```suggestion
  public infos: CompensatingWriteErrorInfo[] = [];
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +399,87 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
   });
 
+  describe("Errors", () => {
+    it("raises compensating writes errors", async function () {
+      const person1Id = new BSON.ObjectId("0000002a9a7969d24bea4cf5");
+      const person2Id = new BSON.ObjectId("0000002a9a7969d24bea4cf6");
+      const dogId = new BSON.ObjectId("0000002a9a7969d24bea4cf7");
+
+      const promise = new Promise<void>((resolve, _) => {
+        (async () => {
```
kraenhansen, line 409: Why the inner body? This could result in an unhandled promise rejection.

```suggestion
      const promise = new Promise<void>(async (resolve, _) => {
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +399,87 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+          const errorCallback: ErrorCallback = (_, error) => {
+            expect(error).instanceOf(CompensatingWriteError);
+            expect(error.code).to.equal(231);
+            expect(error.isFatal).to.be.false;
+            expect(error.message).to.contain(
+              "Client attempted a write that is outside of permissions or query filters; it has been reverted",
+            );
+
+            const cpe = error as CompensatingWriteError;
```
kraenhansen, line 418: It's a bit annoying that the types of `expect` doesn't declare a type assertion.

You could do this yourself with to avoid the rename:

```suggestion
            if (!(error instanceof CompensatingWriteError)) {
              throw new Error("Expected a CompensatingWriteError");
            }
```

After which `error` will be `CompensatingWriteError`.

packages/realm/src/errors.ts
kraenhansen: Alternatively.

```suggestion
  public writes: CompensatingWriteErrorInfo[] = [];
```

kraenhansen (commented):
I have just a

integration-tests/tests/src/tests/sync/flexible.ts
papafe: I agree on awaiting right away, I will fix it.

Regarding why the inner body, I did it because of this linting rule: https://eslint.org/docs/latest/rules/no-async-promise-executor. I was reading that having it defined as async could cause some issue, like the fact that if it throws, then the promise will never be resolved. After doing some testing I think that this probably behaves in the same way as the version that the linter complains about. 

What do you think, should I just ignore the linter or do you have any other suggestion about how to do this?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * None
 
 ### Enhancements
+* Added new error class `CompensatingWriteError`, that that indicates that one or more object changes have been reverted by the server.
```
kneth, line 7:
```suggestion
* Added new error class `CompensatingWriteError`, which indicates that one or more object changes have been reverted by the server. 
```

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * None
 
 ### Enhancements
+* Added new error class `CompensatingWriteError`, that that indicates that one or more object changes have been reverted by the server. 
+This can happen when the client creates/updates objects that do not match any subscription, or performs writes on an object it didn't have permission to ([#5599](https://github.com/realm/realm-js/pull/5599)).
```
kneth, line 8:
```suggestion
This can happen when the client creates/updates objects that do not match any subscription, or performs writes on an object it didn't have permission to access. ([#5599](https://github.com/realm/realm-js/pull/5599))
```

packages/realm/src/errors.ts
```diff
@@ -138,3 +141,45 @@ export class ClientResetError extends SyncError {
     };
   }
 }
+
+/**
+ * An error class that indicates that one or more object changes have been reverted by the server.
```
kneth, line 146: Consider to update if my suggestion to `CHANGELOG.md` change is accepted

kneth (approved):
I have added a couple of suggestions, and they shouldn't block you in any way. Great to see compensating writes coming to a SDK near me :smile:

packages/realm/src/Types.ts
```diff
@@ -18,6 +18,8 @@
 
 import { BSON, Dictionary, List, RealmSet, Results } from "./internal";
 
+export type Mixed = unknown;
```
papafe, line 21: @kraenhansen before I was importing `Mixed` from `.` in `error.ts`, but you correctly suggested to import it from `./internal`, but it's not available there. So I added it here for now, but I am not sure of what's a good place for that. Directly `internal`?

elle-j: When importing it in `error.ts`, try importing `Types` from `./internal` instead. Was `Mixed` not available on `Types`?

packages/realm/src/errors.ts
```diff
@@ -92,7 +93,9 @@ export class TimeoutError extends Error {
 
 /** @internal */
 export function fromBindingSyncError(error: binding.SyncError) {
-  if (error.isClientResetRequested) {
+  if (error.systemError.code == 231) {
```
elle-j, line 96:
```suggestion
  if (error.systemError.code === 231) {
```

packages/realm/src/errors.ts
```diff
@@ -138,3 +140,45 @@ export class ClientResetError extends SyncError {
...
+/**
+ * An error class that indicates that one or more object changes have been reverted by the server.
+ * This can happen when the client creates/updates objects that do not match any subscription, or performs writes on
+ * an object it didn't have permission to access.
+ */
+export class CompensatingWriteError extends SyncError {
+  /**
+   * The array of compensating writes performed by the server.
+   */
+  public infos: CompensatingWriteInfo[] = [];
```
elle-j, line 153: I'd suggest using e.g. `info` or `writes`.
```suggestion
  public writes: CompensatingWriteInfo[] = [];
```

packages/realm/src/errors.ts
```diff
@@ -138,3 +140,45 @@ export class ClientResetError extends SyncError {
...
+
+  /** @internal */
+  constructor(error: binding.SyncError) {
+    super(error);
+    if (error.compensatingWritesInfo) {
```
elle-j, line 158: I believe `error.compensatingWritesInfo` is always an array right? If so, the if condition will always be true. You can check the length instead 👍 
```suggestion
    if (error.compensatingWritesInfo.length) {
```

packages/realm/src/errors.ts
```diff
@@ -138,3 +140,45 @@ export class ClientResetError extends SyncError {
...
+      error.compensatingWritesInfo.forEach((element) => {
+        this.infos.push({ objectName: element.objectName, reason: element.reason, primaryKey: element.primaryKey });
+      });
```
elle-j, line 161: Not sure if you actually have to create a separate object. If not: 
```ts
for (const write of error.compensatingWritesInfo) {
  // IF RENAMING `infos` TO `writes`:
  this.writes.push(write);
}
```

Otherwise:
```ts
for (const write of error.compensatingWritesInfo) {
  // IF RENAMING `infos` TO `writes`:
  this.writes.push({ ...write });
}
```

P.s. `forEach()` is quite slow.

packages/realm/src/errors.ts
```diff
@@ -138,3 +140,45 @@ export class ClientResetError extends SyncError {
...
 }
+
+/**
+ * An error class that indicates that one or more object changes have been reverted by the server.
+ * This can happen when the client creates/updates objects that do not match any subscription, or performs writes on
+ * an object it didn't have permission to access.
+ */
+export class CompensatingWriteError extends SyncError {
+  /**
+   * The array of compensating writes performed by the server.
```
elle-j, line 151: Should there be a distinction between?:
* (A) The information about an object that caused the compensating write error.
* (B) A compensating write.

If so, one reformulation could be
```suggestion
   * The array of information about each object that caused the compensating write.
```

packages/realm/src/errors.ts
```diff
@@ -138,3 +140,45 @@ export class ClientResetError extends SyncError {
...
+    }
+  }
+}
+
+/**
+ * The details of a compensating write performed by the server.
+ */
+export type CompensatingWriteInfo = {
+  /**
+   * The type of the object affected by the compensating write.
```
elle-j, line 171: Is the following a more accurate description?
```suggestion
   * The type of the object that caused the compensating write.
```

packages/realm/src/errors.ts
```diff
@@ -138,3 +140,45 @@ export class ClientResetError extends SyncError {
...
+   */
+  objectName: string;
+
+  /**
+   * The reason for the compensating write.
+   */
+  reason: string;
+
+  /**
+   * The primary key of the object affected by the compensating write.
```
elle-j, line 181: Same as previous comment.
```suggestion
   * The primary key of the object that caused the compensating write.
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +399,88 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
     });
   });
 
+  describe("Errors", () => {
```
elle-j, line 402: We might want to make this description more specific since we're testing errors in other suites as well 👍 
```suggestion
  describe("Compensating writes", () => {
```

elle-j (approved):
Great and thanks 🙏 Added suggestions 👍

packages/realm/src/Types.ts
papafe: I can do that, but then I'll need to use it as `Types.Mixed`. Does it make sense?

packages/realm/src/errors.ts
papafe: Yes, it should be an array. My idea here is just to check that it's not null or undefined

packages/realm/src/errors.ts
papafe: Makes sense! And good to know about foreach 😁

packages/realm/src/errors.ts
takameyer, line 153: `info` is plural.  I do, however, think this name is a bit amiguous.  Perhaps, `writeErrors` would suffice.

packages/realm/src/errors.ts
takameyer, line 151:
```suggestion
   * The array of compensating write failures performed by the server.
```

takameyer: As proven by the tests, only the failed writes are included in the error.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +399,86 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+    it("raises compensating writes errors", async function () {
+      const person1Id = new BSON.ObjectId("0000002a9a7969d24bea4cf5");
+      const person2Id = new BSON.ObjectId("0000002a9a7969d24bea4cf6");
+      const dogId = new BSON.ObjectId("0000002a9a7969d24bea4cf7");
+
+      await new Promise<void>((resolve, _) => {
```
takameyer, line 408: Does this even need to be wrapped in a promise?  We could just await the realm.open and the upload.  Mocha has a `done()` argument to the test method that can be included in the error callback.

papafe: When the upload is finished it's not guaranteed that the error is raised, that's why I did it like this.

Thinking about it... I can maybe add a wait for the download after the upload, and maybe that would work.

takameyer: https://mochajs.org/#asynchronous-code

papafe: Thanks for the link! That's probably what I need

takameyer: I actually learned about the `done()` argument when reviewing the test migrations.  Very cool feature for async code.

packages/realm/src/errors.ts
takameyer: ah, I have misunderstood what CompensatingWrite actually means.  The compensating write, is the server telling the client to remove those entries that it didn't have the rights to perform.  The comment is correct then.

packages/realm/src/errors.ts
takameyer: Ok, so as we talked about above, these aren't failed writes, they are actually writes performed after the server has determined they are not allowed.  In that case, probably just renaming the property to `writes` would be clear.

integration-tests/tests/src/tests/sync/flexible.ts
papafe: It seems I can't use this, because of `Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.`
It seems because the body of the method is async, then it's implicitly a promise.

takameyer: Ok, was worth a shot.

packages/realm/src/errors.ts
takameyer: Looks like @elle-j and I had the same idea

elle-j: I think a more specific description of what this array is is: `The array of information about each object that caused the compensating write.`. So I don't believe it's actually an array of writes, just the info about what caused the compensating write. Core's more verbose naming `CompensatingWriteErrorInfo` disambiguates this, but for our shorter field name, I think all of these work: `errorInfo`, `info` (since it's within the context of a `CompensatingWriteError`), `writeErrors`.

packages/realm/src/Types.ts
papafe: Discussed with @kraenhansen , and I will use `PrimaryKey` as the type here, as it makes sense

packages/realm/src/errors.ts
nirinchev: Could also go with `revertedObjects` or something similar.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +399,86 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
     });
   });
 
+  describe("Errors", () => {
+    it("raises compensating writes errors", async function () {
+      const person1Id = new BSON.ObjectId("0000002a9a7969d24bea4cf5");
+      const person2Id = new BSON.ObjectId("0000002a9a7969d24bea4cf6");
+      const dogId = new BSON.ObjectId("0000002a9a7969d24bea4cf7");
```
nirinchev, line 406: Why are these hardcoded?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +399,86 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+            const luigi = realm.create<IPerson>(FlexiblePersonSchema.name, {
+              _id: new BSON.ObjectId(),
+              name: "Luigi",
+              age: 20,
+            });
+            realm.create<IPerson>(FlexiblePersonSchema.name, { _id: new BSON.ObjectId(), name: "Mario", age: 22 });
+            realm.create<IDog>(DogSchema.name, { _id: new BSON.ObjectId(), name: "Oldy", age: 6, owner: luigi });
+          });
+
+          await realm?.syncSession?.uploadAllLocalChanges();
```
nirinchev, line 476: Why?all?these?question?marks? 😄

nirinchev (commented):
This is not a problem with this PR, but it looks like we don't expose the log url in the sync error. We should probably do it as the server logs are quite useful to understand what exactly is failing.

integration-tests/tests/src/tests/sync/flexible.ts
papafe: Just to have an order of the objects, I could have done the same with names probably :)

integration-tests/tests/src/tests/sync/flexible.ts
papafe: They're fun to use 😁

elle-j: Looks like it should only be needed after `syncSession`.

takameyer: Optional chaining!   It replaces this code:
```tsx
if( realm && realm.syncSession && realm.uploadAllLocalChanges){
  await realm.syncSession.uploadAllLocalChanges();
}
```

packages/realm/src/errors.ts
papafe: https://github.com/realm/realm-js/issues/5609 created an issue for this, there's already too many comments on this PR 😁

integration-tests/tests/src/tests/sync/flexible.ts
papafe: No, just the second one, I think it's fine :)

papafe:
@elle-j @kraenhansen @takameyer @kneth 
I've just tried to follow all your recommendations so it'll be great if you could give a final (?) look:
- The final name for the array property is `writeErrors`, I hope we can all agree with that in some way 😁 
- I'm using `promiseHandle` for the test, instead of that horrible promise with async body
- I've changed the type of the primary key from `Mixed` to `PrimaryKey` and added an assert method to verify it's of the correct type

packages/realm/src/assert.ts
```diff
@@ -76,6 +76,17 @@ assert.symbol = (value: unknown, target?: string): asserts value is symbol => {
...
+assert.primaryKey = (value: unknown, target?: string): asserts value is PrimaryKey => {
+  assert(
+    value === null ||
+      typeof value === "number" ||
+      typeof value === "string" ||
+      value instanceof BSON.UUID ||
+      value instanceof BSON.ObjectId,
+    () => new TypeAssertionError("a primary key", value, target),
+  );
+};
```
kraenhansen, line 88: 🥇

packages/realm/src/errors.ts
```diff
@@ -138,3 +141,45 @@ export class ClientResetError extends SyncError {
...
+  public writeErrors: CompensatingWriteInfo[] = [];
+
+  /** @internal */
+  constructor(error: binding.SyncError) {
+    super(error);
+    for (const write of error.compensatingWritesInfo) {
+      const primaryKey = write.primaryKey;
+      assert.primaryKey(primaryKey);
+      this.writeErrors.push({ objectName: write.objectName, reason: write.reason, primaryKey: primaryKey });
+    }
```
kraenhansen, line 163:
```suggestion
    for (const { objectName, primaryKey, reason } of error.compensatingWritesInfo) {
      assert.primaryKey(primaryKey);
      this.writeErrors.push({ objectName, reason, primaryKey });
    }
```

papafe: I need to remind myself this is javascript...
Fixed it

integration-tests/tests/src/tests/sync/flexible.ts
nirinchev: You could generate them and then sort them manually. My concern is that if you're running the test multiple times against the same app/cluster, we could observe unexpected behavior due to old data.

elle-j (approved):

integration-tests/tests/src/tests/sync/flexible.ts
takameyer: TypeScript will be defensive about this and have a type error if you do not have a check.

takameyer:
:shipit: ShipIt

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -396,6 +398,85 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+          _id: new BSON.ObjectId(),
+          name: "Luigi",
+          age: 20,
+        });
+        realm.create<IPerson>(FlexiblePersonSchema.name, { _id: new BSON.ObjectId(), name: "Mario", age: 22 });
+        realm.create<IDog>(DogSchema.name, { _id: new BSON.ObjectId(), name: "Oldy", age: 6, owner: luigi });
+      });
+
+      await realm.syncSession?.uploadAllLocalChanges();
+      await callbackHandle;
```
kraenhansen, line 476: Looking at this, we should probably change this API to return something that can actually be awaited directly 🤔  but you have to await the `promise` property on the handle instead of the handle itself.

```suggestion
      await callbackHandle.promise;
```

kraenhansen (commented):
There's just one thing missing 😬

packages/realm/src/errors.ts
```diff
@@ -138,3 +141,44 @@ export class ClientResetError extends SyncError {
...
+/**
+ * An error class that indicates that one or more object changes have been reverted by the server.
+ * This can happen when the client creates/updates objects that do not match any subscription, or performs writes on
+ * an object it didn't have permission to access.
+ */
+export class CompensatingWriteError extends SyncError {
+  /**
+   * The array of information about each object that caused the compensating write.
+   */
+  public writes: CompensatingWriteInfo[] = [];
```
kraenhansen, line 154: 💙

integration-tests/tests/src/tests/sync/flexible.ts
nirinchev: Wait, why is `Realm.syncSession` nullable but `Realm.subscriptions` not?

takameyer: My guess is that `subscriptions` can just return an empty array, even if sync isn't enabled.  `syncSession` should return null if sync isn't configured

takameyer (approved):

integration-tests/tests/src/tests/sync/flexible.ts
papafe: Understandable, just fixed it.

PR #5610, opened 2023-03-22
Fix linting on CI

kraenhansen:
## What, How & Why?

This solves running the linters on CI.

## ☑️ ToDos
* [x] 📝 Changelog entry

### DIFF
--- a/packages/realm-react/package.json
+++ b/packages/realm-react/package.json
@@ -10,7 +10,7 @@
     "build": "wireit",
     "start": "npm run build -- --watch",
     "test": "jest --forceExit",
-    "lint": "eslint --ext .tsx . && tsc --noEmit"
+    "lint": "eslint --ext .tsx ."
   },
   "wireit": {
     "build": {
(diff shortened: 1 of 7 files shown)


packages/realm-react/package.json
```diff
@@ -10,7 +10,7 @@
     "build": "wireit",
     "start": "npm run build -- --watch",
     "test": "jest --forceExit",
-    "lint": "eslint --ext .tsx . && tsc --noEmit"
```
kraenhansen, line 13: @takameyer please note this change. Since we're running the rollup bundler with `noEmitOnError` I don't see a need to invoke the transpiler (which would fail on CI since `realm` isn't yet bundled) when linting.

kraenhansen: The strange thing is that I do remember already removing this at some other point in time / PR 🤔

takameyer: Yeah...I also remember that, but not sure where/when that was.

takameyer (approved):
Looks fine to me, however it appears we have unlocked some new linter errors we need to take care of!

kraenhansen:
> What I find strange is that the linter workflow passes, but github is showing it's own linting errors on dictionary,

I believe these are warnings, not errors.

PR #5622, opened 2023-03-23
Update wireit configs (renaming build → bundle + bindgen updates)

kraenhansen:
## What, How & Why?

This updates the wireit configs, renaming rollups from "build" to "bundle".


packages/realm/package.json
```diff
@@ -73,72 +73,64 @@
...
-    "build:bindgen": {
-      "command": "cmake-js build --debug --directory ../bindgen",
+    "bindgen:configure": {
+      "command": "cmake-js configure --debug --directory ../bindgen",
+      "dependencies": ["../..:submodules"]
+    },
+    "bindgen:build:node": {
+      "command": "cmake --build ../bindgen/build --target realm-js-node",
+      "dependencies": ["bindgen:configure"]
+    },
```
kraenhansen, line 98: @RedBeard0531 can I get your 👀 on this?

kraenhansen: I wonder if there are some files and outputs I can use to prevent the configure task to run 🤔

packages/realm/package.json
```diff
@@ -51,12 +51,12 @@
     "test": "wireit",
     "test:types": "tsc --project type-tests/tsconfig.json",
     "lint": "eslint --ext .js,.mjs,.ts .",
-    "build": "wireit",
+    "build": "echo 'Run bundle, build:node, build:ios or build:android or other scripts depending on these instead' && exit 1",
```
kraenhansen, line 54: This change is potentially controversial. I don't see a use for building for all platforms now that tests invoke the platform specific build commands transitively and on CI it's also split into separate runners.

packages/realm/package.json
RedBeard0531: Hmm. One option if we are willing to force `cmake-js` to use the `Ninja` generator is to make the output be the `build.ninja`. Then, rather than using `cmake-js configure` you can use `cmake-js build --target build.ninja`. Assuming that wireit treats tasks with no input files as "always dirty", it will always invoke ninja to check if build.ninja needs to be updated or not, and  in the common case, it will very quickly exit with nothing to do.

Integrating multiple build systems is always tricky. Especially if you care about both correctness and incremental builds (which I think we do). The simplest thing will always be to have the "outer" build system unconditionally invoke the "inner" build system and let it handle deciding if it needs to do anything or not. Ideally the outer build system has a way to detect if the output files have changed *after* running the inner build system, so that it can only rerun downstream tasks when needed. We should only deviate from the "simple" solution if it brings enough benefit to outweigh the complexity and risks.

packages/realm/package.json
takameyer: That's a good change.  I constantly find myself having to cancel that task.  The `ios` build script doesn't respond immediately to `ctrl-c`

packages/realm/package.json
```diff
@@ -74,72 +74,64 @@
...
-    },
-    "rebuild:bindgen": {
-      "command": "cmake-js rebuild --debug --directory ../bindgen",
-      "dependencies": [
-        "../..:submodules"
+      ],
+      "files": [
+        "../bindgen/spec.yml",
+        "../bindgen/js_spec.yml",
+        "../bindgen/src"
```
RedBeard0531, line 125: Consider excluding `src/templates` and then re-including the specific template used by the command. That matches the cmake behavior and avoids rerunning one generator when you only modify another.

Ditto for all similar tasks.

packages/realm/package.json
kraenhansen: > always be to have the "outer" build system unconditionally invoke the "inner" build system

As I don't declare files and outputs, this is the current (default) behaviour.

PR #5632, opened 2023-03-24
Added `SyncError.logUrl`

### LINKED ISSUE #5609: Expose log url in SyncError
We need to expose log url in `SyncError`, as it's missing now.

papafe:
This PR exposes `SyncError.logUrl`, that contains the URL pointing to the server log related to the sync error. 

As I was there:
- I added docs for `SyncError`
- I deprecated `SyncError.isFatal`, as it seems it's something that is only used internally and should probably not be exposed to SDK

This closes #5609 
## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * None
 
 ### Enhancements
+* Exposed `SyncError.logUrl`, that contains the URL to the server log related to the sync error. ([#5609](https://github.com/realm/realm-js/issues/5609))
 * Added a new error class `CompensatingWriteError` which indicates that one or more object changes have been reverted by the server. 
 This can happen when the client creates/updates objects that do not match any subscription, or performs writes on an object it didn't have permission to access. ([#5599](https://github.com/realm/realm-js/pull/5599))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None
 
 ### Enhancements
-* Exposed `SyncError.logUrl`, that contains the URL to the server log related to the sync error. ([#5609](https://github.com/realm/realm-js/issues/5609))
+* Exposed `SyncError.logUrl` which contains the URL to the server log related to the sync error. ([#5609](https://github.com/realm/realm-js/issues/5609))
 * Added a new error class `CompensatingWriteError` which indicates that one or more object changes have been reverted by the server. 
 This can happen when the client creates/updates objects that do not match any subscription, or performs writes on an object it didn't have permission to access. ([#5599](https://github.com/realm/realm-js/pull/5599))
 
--- a/integration-tests/tests/src/tests/sync/flexible.ts
+++ b/integration-tests/tests/src/tests/sync/flexible.ts
@@ -412,7 +412,7 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
 
       const errorCallback: ErrorCallback = (_, error) => {
         expect(error.code).to.equal(231);
-        expect(error.isFatal).to.be.false;
+        expect(error.logUrl).to.be.not.empty;
         expect(error.message).to.contain(
           "Client attempted a write that is outside of permissions or query filters; it has been reverted",
         );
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * None
 
 ### Enhancements
+* Exposed `SyncError.logUrl`, that contains the URL to the server log related to the sync error. ([#5609](https://github.com/realm/realm-js/issues/5609))
```
elle-j, line 7:
```suggestion
* Exposed `SyncError.logUrl` which contains the URL to the server log related to the sync error. ([#5609](https://github.com/realm/realm-js/issues/5609))
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -412,7 +412,7 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
 
       const errorCallback: ErrorCallback = (_, error) => {
         expect(error.code).to.equal(231);
-        expect(error.isFatal).to.be.false;
+        expect(error.logUrl).to.be.not.empty;
```
elle-j, line 415:
```suggestion
        expect(error.logUrl).to.not.be.empty;
```

packages/realm/src/errors.ts
```diff
@@ -103,11 +103,30 @@ export function fromBindingSyncError(error: binding.SyncError) {
...
+  /**
+   * A numeric code representing the error.
+   */
   public code: number;
+
+  /**
+   * A string representing the error category.
+   */
   public category: string;
-  public isFatal: boolean;
```
elle-j, line 110: Removing `isFatal` would be a breaking change though right?

packages/realm/src/errors.ts
elle-j, line 1: I like docs 😃 But from past comments on my own additions to some docs, I believe these could be seen as redundant and should be left out? At least the types in the descriptions. E.g. not saying "string".,

packages/realm/src/errors.ts
```diff
@@ -103,11 +103,30 @@ export function fromBindingSyncError(error: binding.SyncError) {
...
+
+  /**
+   * The URL to the associated server log, if any.
```
elle-j, line 123: When you're saying "if any", do you mean that the URL could be empty? I interpret "if any" as if this field could be `undefined`.

elle-j (commented):
👍  Mainly wondering if the removal of `isFatal` is a breaking change we don't want to add now?

packages/realm/src/errors.ts
papafe: You're correct, I saw it wrong, I thought we added it in the new version

packages/realm/src/errors.ts
papafe: I have to agree you on this ....

packages/realm/src/errors.ts
papafe: Most of the time `SyncError` can be caused by the server (for compensating writes, client resets and other errors), but also by the realm client itself (for example `illegal-realm-path` https://github.com/realm/realm-core/blob/11aa84592fc049b2a5492f80d725f7a202ca0a87/src/realm/sync/protocol.hpp#L302). In the last case there is no server logs because the error is raised locally.

papafe: And yes, my bad, this can be empty but not undefined. I will update the docs to make it clearer.

packages/realm/src/errors.ts
kraenhansen: Yeah. If we want to do this, it could be deprecated here and the removal go into v13.

kraenhansen: Oh. I saw now that you just moved it down in the file 👍

kraenhansen (approved):

packages/realm/src/errors.ts
papafe: Yep, I added the deprecated tag now

packages/realm/src/errors.ts
elle-j: Alrighty thanks! 👍

kneth (approved):

PR #5636, opened 2023-03-24
Update how to run pod install

kneth:
Using `bundle` seems to be a good idea

### DIFF
--- a/integration-tests/environments/react-native/README.md
+++ b/integration-tests/environments/react-native/README.md
@@ -21,8 +21,9 @@ npx lerna bootstrap --scope @realm/react-native-tests --include-dependencies
 For iOS environments run 
 
 ```bash
+cd ios
 bundle install
-cd ios && bundle exec pod install
+bundle exec pod install
 ```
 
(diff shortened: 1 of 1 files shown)


integration-tests/environments/react-native/README.md
```diff
@@ -21,8 +21,9 @@ npx lerna bootstrap --scope @realm/react-native-tests --include-dependencies
 For iOS environments run 
 
 ```bash
+cd ios
```
kraenhansen, line 24: Are you use the first command should be run from "ios" too? It might traverse the filetree to look for the gemlock file 🤔

elle-j: `bundle install` was originally outside of the ios directory, but I asked if it should be inside instead due to React Native's [setup page](https://reactnative.dev/docs/environment-setup).

kraenhansen: Oh okay .. that works too then 😄

kraenhansen (approved):

elle-j (approved):

PR #5648, opened 2023-03-27
Enable garbage collection between tests

kraenhansen:
## What, How & Why?

This change forces GC between tests when running them from `integration-tests/environments/node`.
I added this to tests when they were running via `integration-tests/tests` a while back, as it made makes the tests more deterministic and helped debug some lifetime related issues.

Having this is good because it make the tests behave more deterministic (as it eliminates races related to GC), but at the same time it hides some lifetime related issues (suchs as realms not being closed by core if a sync session is kept alive due to an SDK level object owning a shared pointer to it).

In the future, we should probably
- make this more configurable and run the tests on CI with and without GC between tests.
- investigate enabling this for the other supported platforms.

### DIFF
--- a/integration-tests/environments/node/index.mjs
+++ b/integration-tests/environments/node/index.mjs
@@ -21,6 +21,11 @@ import { Client } from "mocha-remote-client";
 import fs from "fs-extra";
 import path from "path";
 
+import v8 from "node:v8";
+import vm from "node:vm";
+
+v8.setFlagsFromString("--expose_gc");
+
 const client = new Client({
   title: `Node.js v${process.versions.node} on ${os.platform()}`,
   async tests(context) {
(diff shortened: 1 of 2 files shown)


integration-tests/environments/node/index.mjs
```diff
@@ -21,13 +21,19 @@ import { Client } from "mocha-remote-client";
 import fs from "fs-extra";
 import path from "path";
 
+import v8 from "node:v8";
+import vm from "node:vm";
+
+v8.setFlagsFromString("--expose_gc");
```
kraenhansen, line 27: I got some inspiration for this from https://github.com/legraphista/expose-gc/blob/master/function.js

elle-j (approved):
🙏  👍

kneth (approved):

PR #5650, opened 2023-03-28
Fix `mocha` dependencies

kraenhansen:
## What, How & Why?

- Removes peer dependencies on `mocha` which are now redundant as NPM version ([>=7](https://nodejs.org/es/blog/npm/peer-dependencies)) install these silently.
- Adds an `engines` property to the `package.json` to ensure we all use the same major versions of NPM and Node when developing.
- Rebuilds the package-lock.json, because it was in a broken state that would incorrectly install `mocha@v5.4.0` into `integration-tests/node_modules/mocha` which would break parsing of the `mocharc.json` in the `integration-tests/tests` as that's supported only by newer versions of Mocha.
- Reverts the upgrade of the `package-lock.json`'s version (as version 3 is emitted by NPM v9).


package.json
```diff
@@ -42,19 +42,10 @@
...
-    "sqlite",
-    "asyncstorage",
-    "rocksdb",
-    "leveldb",
-    "realm"
-  ],
+  "engines": {
+    "node": "^18",
+    "npm": "^8"
+  },
```
kraenhansen, line 48: These are the active LTS versions.

elle-j (approved):
Thanks! Regarding "...as newer versions of NPM install these silently" in the description, it may also be helpful to specify from which version this is true?

PR #5651, opened 2023-03-28
Keep an `IterableWeakRefs` of `SyncSession`s to reset when clearing test state

kraenhansen:
## What, How & Why?

This adds clearing of `SyncSession` internal shared pointers when clearing test state between tests.
It makes our tests behave more predictably, since these shared pointers would otherwise keeps sessions and therefore realms open across tests.

### DIFF
--- a/packages/realm/src/IterableWeakRefs.ts
+++ b/packages/realm/src/IterableWeakRefs.ts
@@ -24,38 +24,27 @@ import { binding } from "./internal";
  */
 export class IterableWeakRefs<T extends object> {
   private internal: Set<binding.WeakRef<T>>;
-  constructor(values?: readonly T[] | null) {
-    this.internal = new Set(values ? values.map((value) => new binding.WeakRef(value)) : undefined);
-  }
-
-  /**
-   * Deletes all weak references that no longer points to an object
-   */
-  clean() {
-    for (const ref of this.internal) {
-      if (!ref.deref()) {
-        this.internal.delete(ref);
-      }
-    }
+  constructor() {
+    this.internal = new Set();
   }
 
   add(value: T) {
+    // This is a good time to get rid of entries that has been GC'ed
+    this.clean();
     this.internal.add(new binding.WeakRef(value));
     return this;
   }
 
-  clear() {
-    this.internal.clear();
-  }
-
-  delete(value: T): boolean {
-    // This is a good time to get rid of entries that has been GC'ed
+  delete(value: T) {
     this.clean();
     const ref = this.find(value);
     if (ref) {
-      return this.internal.delete(ref);
+      this.internal.delete(ref);
     }
-    return false;
+  }
+
+  clear() {
+    this.internal.clear();
   }
 
   *[Symbol.iterator](): IterableIterator<T> {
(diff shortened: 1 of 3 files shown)


packages/realm/src/app-services/SyncSession.ts
```diff
@@ -273,6 +290,7 @@ export class SyncSession {
   /** @internal */
   constructor(internal: binding.SyncSession) {
     this._internal = internal;
+    SyncSession.instances.add(new binding.WeakRef(this));
```
RedBeard0531, line 293: Won't this grow forever in the non-test case? I don't see anything that removes dead WeakRefs from `instances` outside of tests.

kraenhansen: Good catch. I could add the removal to `resetInternal`? That way the closing a Realm would at least remove it.
If we had the `FinalizationRegistry` available on all platforms, that might also be used, but I'm not completely sure that we do.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -273,13 +287,16 @@ export class SyncSession {
...
+    SyncSession.instances.add(this);
   }
 
   /**@internal*/
   resetInternal() {
     if (!this._internal) return;
     this._internal.$resetSharedPtr();
     this._internal = null;
+    // No need to remember this instance, now that we've already resat the internal
+    SyncSession.instances.delete(this);
```
RedBeard0531, line 299: This makes the `resetAllInternals` loop O(n^2) because `delete()` is secretly linear. Given that you are calling `clear()` there is no need to also delete everything.

Ditto for the loop over realms in `Realm.clearTestState()` because closing a realm also calls resetInternal on its SyncSession.

PS- This is why I don't like the idea of trying to look like a well known type, while having a very different performance profile.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -262,6 +263,19 @@ const CONNECTION_LISTENERS = new Listeners<ConnectionNotificationCallback, Liste
...
+  /**
+   * Resets the internal shared pointer of all instances returned since this message was last called.
+   * @internal
+   */
+  public static resetAllInternals() {
+    for (const session of SyncSession.instances) {
+      session.resetInternal();
+    }
+    SyncSession.instances.clear();
+  }
```
kraenhansen, line 277:
```suggestion
  public static resetAllInternals() {
    for (const session of SyncSession.instances) {
      // `resetInternal` will remove it from the collection internally
      session.resetInternal();
    }
  }
```

RedBeard0531: Isn't that still O(N^2)? It also seems better to use `clear()` to wipe everything out in one go, rather than individually removing each entry from the set (even ignoring the fact that delete is linear). Clear is usually O(1), while deleting everything is usually O(N), but in this case is O(N^2).

packages/realm/src/app-services/SyncSession.ts
RedBeard0531: It's not just the `clean()` call. `delete()` is inherently O(N) because it calls `find()`. One way around that would be to have the `IterableWeakRefs` hold something like `interface WeakFromThis<T> extends T { weakThis: WeakRef<T> }`, then delete could do `this.internal.delete(val.weakThis)`. But at that point I think you might as well just use `Set<WeakRef<T>>` again.

packages/realm/src/IterableWeakRefs.ts
```diff
@@ -72,8 +61,12 @@ export class IterableWeakRefs<T extends object> {
    */
   private find(value: T) {
     for (const ref of this.internal) {
-      if (ref.deref() === value) {
+      const other = ref.deref();
+      if (other === value) {
         return ref;
```
kraenhansen, line 66: I know we're potentially not making it all the way through the list here.

kraenhansen:
Closing this in favour of https://github.com/realm/realm-js/pull/5689

PR #5656, opened 2023-03-29
Throwing errors when using symbols as keys of dictionary

### LINKED ISSUE #5604: Add a test expecting an error when using a `Symbol` as key in a `Dictionary`

papafe:
This closes #5604 

## ☑️ ToDos
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -176,11 +176,11 @@ describe("Dictionary", () => {
         expect(() => {
           //@ts-expect-error We need to verify using a symbol as a key throws
           item.dict.set(sym, "value1");
-        }).to.throw("Input key cannot be a symbol");
+        }).to.throw("Symbols cannot be used as keys of a dictionary");
 
         expect(() => {
           item.dict.set({ [sym]: "value2" });
-        }).to.throw("Input object cannot contain symbol keys");
+        }).to.throw("Symbols cannot be used as keys of a dictionary");
 
         expect(() => {
           //@ts-expect-error We need to verify using a symbol as a key throws
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -167,6 +167,26 @@ describe("Dictionary", () => {
       expect(Object.keys(item.dict).length).deep.equals(3);
     });
 
+    it("throws when using Symbol as key", function (this: RealmContext) {
+      const sym = Symbol("testSymbol");
+
+      this.realm.write(() => {
+        const item = this.realm.create<Item>("Item", {});
+
+        expect(() => {
+          item.dict.set(sym, "value1");
+        }).to.throw("Input key cannot be a symbol");
+
+        expect(() => {
+          item.dict.set({ [sym]: "value2" });
+        }).to.throw("Input object cannot contain symbol keys");
+
+        expect(() => {
+          item.dict[sym] = "value3";
+        }).to.throw("Symbols cannot be used as keys of a dictionary");
+      });
+    });
+
     it("set/remove methods return the dictionary", function (this: RealmContext) {
       const item = this.realm.write(() => this.realm.create<Item>("Item", {}));
 
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -174,6 +174,7 @@ describe("Dictionary", () => {
         const item = this.realm.create<Item>("Item", {});
 
         expect(() => {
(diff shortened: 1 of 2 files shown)


kraenhansen (approved):
I have a small suggestion, but it can merge as is.

packages/realm/src/Dictionary.ts
```diff
@@ -232,34 +233,37 @@ export class Dictionary<T = unknown> extends Collection<string, T, [string, T],
...
+   * @param elementsOrKey The element to add or the key of the element to add.
+   * @param value The value of the element to add.
+   * @throws {@link AssertionError} If not inside a write transaction, if using symbol as keys, or if any value violates type constraints.
+   * @returns The dictionary.
    * @since 10.6.0
    */
   set(elementsOrKey: string | { [key: string]: T }, value?: T): this {
+    assert(typeof elementsOrKey != "symbol", "Input key cannot be a symbol");
+    const elements = typeof elementsOrKey == "string" ? { [elementsOrKey]: value } : elementsOrKey;
+    assert(Object.getOwnPropertySymbols(elements).length == 0, "Input object cannot contain symbol keys");
```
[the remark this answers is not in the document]
papafe: Makes sense... I was wondering though, in this case wouldn't the error message sound a little awkward for someone that is using the `set(key, value)` override?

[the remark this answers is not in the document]
kraenhansen: I don't think so 🤔 You're providing a single symbol as the key, but the statement (an invariant using "keys" in plural) is still correct.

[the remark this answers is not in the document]
papafe: You're right, I think I mixed up the lines in the suggestion 😁  My bad!

packages/realm/src/Dictionary.ts
```diff
@@ -57,6 +57,7 @@ const PROXY_HANDLER: ProxyHandler<Dictionary> = {
       internal.insertAny(prop, toBinding(value, undefined));
       return true;
     } else {
+      assert(typeof prop != "symbol", "Symbols cannot be used as keys of a dictionary");
```
elle-j, line 60:
```suggestion
      assert(typeof prop !== "symbol", "Symbols cannot be used as keys of a dictionary");
```

packages/realm/src/Dictionary.ts
```diff
@@ -232,34 +233,36 @@ export class Dictionary<T = unknown> extends Collection<string, T, [string, T],
...
-   * @throws {@link AssertionError} If not inside a write transaction or if value violates type constraints
-   * @returns The dictionary
+   * @param elementsOrKey The element to add or the key of the element to add.
+   * @param value The value of the element to add.
+   * @throws {@link AssertionError} If not inside a write transaction, if using symbol as keys, or if any value violates type constraints.
+   * @returns The dictionary.
    * @since 10.6.0
    */
   set(elementsOrKey: string | { [key: string]: T }, value?: T): this {
+    const elements = typeof elementsOrKey == "object" ? elementsOrKey : { [elementsOrKey]: value };
```
elle-j, line 260:
```suggestion
    const elements = typeof elementsOrKey === "object" ? elementsOrKey : { [elementsOrKey]: value };
```

kraenhansen (approved):

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -167,6 +167,28 @@ describe("Dictionary", () => {
...
     });
 
+    it("throws when using Symbol as key", function (this: RealmContext) {
+      const sym = Symbol("testSymbol");
+
+      this.realm.write(() => {
+        const item = this.realm.create<Item>("Item", {});
+
+        expect(() => {
+          //@ts-expect-error We need to verify using a symbol as a key throws
```
elle-j, line 177: Not a huge deal but I think it would be fine for ts-expect-error messages to be formulated more toward the specific ts issue and less so to surrounding logic (the test name in this case also describes the purpose). At least this is what I've noticed in the rest of the code base 👍 
```suggestion
          //@ts-expect-error Testing invalid key.
```

If changing this, there's 1 more occurrence further below.

elle-j (approved):
😃

packages/realm/src/Dictionary.ts
papafe: I need to pummel the 3 equals in my mind....

packages/realm/src/Dictionary.ts
kraenhansen: I'm a bit surprised the [`eqeqeq`](https://eslint.org/docs/latest/rules/eqeqeq) lint rule isn't enabled by the recommended settings.

PR #5658, opened 2023-03-29
Unit test workflow using NPM workspaces

kraenhansen:
## What, How & Why?

This PR updates the unit test workflow to use NPM workspaces instead of Lerna.


takameyer (approved):
Great Idea.  Looks good!

PR #5628, opened 2023-03-23
Printing calling stack frame when accessing global `Realm`

kraenhansen:
## What, How & Why?

This will improve the warning logged when someone access the global `Realm` from their code, by including the source-code frame reading it:

```
Your app is relying on a Realm global, which will be removed in realm-js v13,
 please update your code to ensure you import Realm via a named import:

 import { Realm } from "realm"; // For ES Modules
 const { Realm } = require("realm"); // For CommonJS
 
 This is happening at <anonymous> (/Users/kraen.hansen/Projects/realm-js/integration-tests/tests/src/tests/sync/realm.ts:384:26)
```

PR #5660, opened 2023-03-29
Adding a `THROW_ON_GLOBAL_REALM` flag

kraenhansen:
## What, How & Why?

As an alternative to https://github.com/realm/realm-js/pull/5628, this is a proposal to add a `THROW_ON_GLOBAL_REALM` which will throw when the global Realm is accessed. This gives a better experience on React Native since the error will result in a red-box with the stack being symbolicated and clickable:

### The initial warning

Screenshot 2023-03-29 at 15 17 45

### After setting the flag

```typescript
import { flags } from "realm";
flags.THROW_ON_GLOBAL_REALM = true;
```

Clicking the calling function (shown by the arrow) will open the offending piece of sourcecode.

Screenshot 2023-03-29 at 15 23 12


kraenhansen:
@takameyer that would be a breaking change that we want to avoid. The idea is for this flag and  the global `Realm` to be removed completely at `v13`.

takameyer (approved):
LGTM :shipit:

PR #5661, opened 2023-03-29
Adds a deprecation to the global `WeakRef` constructor

kraenhansen:
## What, How & Why?

Our SDK depends on the runtime types for the "ES2022" lib, since most of these are available in end-user projects, but this includes `WeakRef` which is currently not exposed on React Native. This adds a deprecation to the symbol to help us avoid relying on it in the SDK.

Screenshot 2023-03-29 at 16 23 52

### DIFF
--- a/packages/realm/src/runtime.d.ts
+++ b/packages/realm/src/runtime.d.ts
@@ -42,7 +42,7 @@ declare module "buffer" {
   type Buffer = Uint8Array;
 }
 
-/** @deprecated React Native doesn't provide this */
+/** @deprecated React Native doesn't provide this, use our `binding.WeakRef` polyfill instead. */
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 declare interface WeakRef<T extends object> {}
 /** @deprecated React Native doesn't provide this */
--- a/packages/realm/src/runtime.d.ts
+++ b/packages/realm/src/runtime.d.ts
@@ -41,3 +41,13 @@ declare const console: Console;
 declare module "buffer" {
   type Buffer = Uint8Array;
 }
+
+/** @deprecated React Native doesn't provide this */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+declare interface WeakRef<T extends object> {}
+/** @deprecated React Native doesn't provide this */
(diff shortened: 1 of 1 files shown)


packages/realm/src/runtime.d.ts
```diff
@@ -41,3 +41,13 @@ declare const console: Console;
 declare module "buffer" {
   type Buffer = Uint8Array;
 }
+
+/** @deprecated React Native doesn't provide this */
```
elle-j, line 45: Would it make sense to add that we should use `binding.WeakRef` if needed? Or is it not the right place for it?

elle-j (approved):

packages/realm/src/runtime.d.ts
kraenhansen: It could make sense 👍 
```suggestion
/** @deprecated React Native doesn't provide this, use our `binding.WeakRef` polyfill instead. */
```

packages/realm/src/runtime.d.ts
```diff
@@ -42,7 +42,7 @@ declare module "buffer" {
   type Buffer = Uint8Array;
 }
 
-/** @deprecated React Native doesn't provide this */
```
RedBeard0531: You may want to @internal this to ensure it never ends up in our .d.ts file and causes problems for users.

RedBeard0531: Actually, nevermind. That applies to everything in this file, so we must be handling this somehow already.

kraenhansen: Yeah. Nothing imports the file, it's included via the `tsconfig`:

https://github.com/realm/realm-js/blob/5d00d08103a20e3682b6196419dee46d43daa31b/packages/realm/tsconfig.json#L21

RedBeard0531: But doesn't that get used by rollup when producing our bundle.d.ts?

RedBeard0531: Or I guess rollup uses its own `inputs` rather than the tsconfig's `include`? I could still see this causing issues via ambient inclusion, but if these types aren't showing up in the .d.ts, I guess it is fine. Just something to keep an eye out for.

PR #5664, opened 2023-03-29
Introducing `IterableWeakSet` and using it to close Realms when clearing test state

kraenhansen:
## What, How & Why?

As a prerequisite for a refactor of #5651 I wanted to add a primitive we could use for both situations.
Note: I wanted to get this up for review before I got too invested in writing unit tests for it too.

### DIFF
--- a/packages/realm/src/IterableWeakRefs.ts
+++ b/packages/realm/src/IterableWeakRefs.ts
@@ -22,7 +22,7 @@ import { binding } from "./internal";
  * An internal iterable set of {@link binding.WeakRef} objects wrapping objects of type {@link T}.
  * @internal
  */
-export class IterableWeakSet<T extends object> {
+export class IterableWeakRefs<T extends object> {
   private internal: Set<binding.WeakRef<T>>;
   constructor(values?: readonly T[] | null) {
     this.internal = new Set(values ? values.map((value) => new binding.WeakRef(value)) : undefined);
--- a/packages/realm/src/IterableWeakSet.ts
+++ b/packages/realm/src/IterableWeakSet.ts
@@ -1,0 +1,128 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { binding } from "./internal";
+
+/**
+ * Implements an iterable `WeakSet`.
+ * @internal
+ */
+export class IterableWeakSet<T extends object> implements Set<T> {
+  private internal: Set<binding.WeakRef<T>>;
+  constructor(values?: readonly T[] | null) {
+    this.internal = new Set(values ? values.map((value) => new binding.WeakRef(value)) : undefined);
+  }
+
+  /**
+   * Deletes all weak references that no longer points to an object
+   */
+  clean() {
+    for (const ref of this.internal) {
+      if (!ref.deref()) {
+        this.internal.delete(ref);
+      }
+    }
+  }
+
+  get size() {
+    return this.internal.size;
+  }
+
+  get [Symbol.toStringTag]() {
+    return IterableWeakSet.name;
+  }
+
+  forEach(cb: (value: T, value2: T, set: Set<T>) => void, thisArg?: unknown) {
+    this.internal.forEach((valueRef, valueRef2) => {
+      const value = valueRef.deref();
+      const value2 = valueRef2.deref();
+      if (value && value2) {
+        cb.call(thisArg, value, value2, this);
+      }
+    }, this);
+  }
+
+  add(value: T) {
+    this.internal.add(new binding.WeakRef(value));
+    return this;
+  }
+
+  clear() {
+    this.internal.clear();
+  }
+
+  delete(value: T): boolean {
+    // This is a good time to get rid of entries that has been GC'ed
+    this.clean();
+    const ref = this.find(value);
+    if (ref) {
+      return this.internal.delete(ref);
+    }
+    return false;
+  }
+
+  has(value: T): boolean {
+    const ref = this.find(value);
+    return ref ? true : false;
+  }
+
+  *entries(): IterableIterator<[T, T]> {
+    for (const [valueRef, valueRef2] of this.internal.entries()) {
+      const value = valueRef.deref();
+      const value2 = valueRef2.deref();
+      if (value && value2) {
+        yield [value, value2];
+      }
+    }
+  }
+
+  *keys(): IterableIterator<T> {
+    for (const keyRef of this.internal.keys()) {
+      const key = keyRef.deref();
+      if (key) {
+        yield key;
+      }
+    }
+  }
+
+  *values(): IterableIterator<T> {
+    for (const valueRef of this.internal.values()) {
+      const value = valueRef.deref();
+      if (value) {
+        yield value;
+      }
+    }
+  }
+
+  *[Symbol.iterator](): IterableIterator<T> {
+    for (const value of this.values()) {
+      yield value;
+    }
+  }
+
+  /**
+   * Finds the weak reference given a value
+   */
+  private find(value: T) {
+    for (const ref of this.internal) {
+      if (ref.deref() === value) {
+        return ref;
+      }
+    }
+  }
+}
--- a/packages/realm/src/IterableWeakSet.ts
+++ b/packages/realm/src/IterableWeakSet.ts
@@ -19,10 +19,10 @@
 import { binding } from "./internal";
 
 /**
- * Implements an iterable `WeakSet`.
+ * An internal iterable set of {@link binding.WeakRef} objects wrapping objects of type {@link T}.
  * @internal
  */
-export class IterableWeakSet<T extends object> implements Set<T> {
+export class IterableWeakSet<T extends object> {
   private internal: Set<binding.WeakRef<T>>;
   constructor(values?: readonly T[] | null) {
     this.internal = new Set(values ? values.map((value) => new binding.WeakRef(value)) : undefined);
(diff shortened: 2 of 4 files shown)


packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import { binding } from "./internal";
+
+/**
+ * Implements an iterable `WeakSet`.
+ * @internal
+ */
+export class IterableWeakSet<T extends object> implements Set<T> {
+  private internal: Set<binding.WeakRef<T>>;
```
kraenhansen, line 26: As you mentioned in the other PR, I might have used an array here instead. That could theoretically speed up deletion, but I'd need O(n) to find the object in any case, so I opted for something closer to the API it implements outwards.

kraenhansen: I also like that I can delete while iterating without having to keep track of the shifting offsets as I need to do with an array.

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.internal = new Set(values ? values.map((value) => new binding.WeakRef(value)) : undefined);
+  }
+
+  /**
+   * Deletes all weak references that no longer points to an object
+   */
+  clean() {
+    for (const ref of this.internal) {
+      if (!ref.deref()) {
+        this.internal.delete(ref);
```
kraenhansen, line 37: I also like that I can delete while iterating without having to keep track of the shifting offsets as I need to do with an array.

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  clear() {
+    this.internal.clear();
+  }
+
+  delete(value: T): boolean {
+    // This is a good time to get rid of entries that has been GC'ed
+    this.clean();
+    const ref = this.find(value);
```
RedBeard0531, line 72: I don't think this will work since value will never equal a weakref in the set

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+   */
+  clean() {
+    for (const ref of this.internal) {
+      if (!ref.deref()) {
+        this.internal.delete(ref);
+      }
+    }
+  }
+
+  get size() {
```
elle-j, line 42:
```suggestion
  get size(): number {
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+    return this.internal.size;
+  }
+
+  get [Symbol.toStringTag]() {
```
elle-j, line 46:
```suggestion
  get [Symbol.toStringTag](): string {
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+export class IterableWeakSet<T extends object> implements Set<T> {
+  private internal: Set<binding.WeakRef<T>>;
+  constructor(values?: readonly T[] | null) {
+    this.internal = new Set(values ? values.map((value) => new binding.WeakRef(value)) : undefined);
+  }
+
+  /**
+   * Deletes all weak references that no longer points to an object
+   */
+  clean() {
```
elle-j, line 34:
```suggestion
  clean(): void {
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  get size() {
+    return this.internal.size;
+  }
+
+  get [Symbol.toStringTag]() {
+    return IterableWeakSet.name;
+  }
+
+  forEach(cb: (value: T, value2: T, set: Set<T>) => void, thisArg?: unknown) {
```
elle-j, line 50:
```suggestion
  forEach(cb: (value: T, value2: T, set: Set<T>) => void, thisArg?: unknown): void {
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.internal.forEach((valueRef, valueRef2) => {
+      const value = valueRef.deref();
+      const value2 = valueRef2.deref();
+      if (value && value2) {
+        cb.call(thisArg, value, value2, this);
+      }
+    }, this);
+  }
+
+  add(value: T) {
```
elle-j, line 60:
```suggestion
  add(value: T): this {
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.internal.add(new binding.WeakRef(value));
+    return this;
+  }
+
+  clear() {
```
elle-j, line 65:
```suggestion
  clear(): void {
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+    const ref = this.find(value);
+    if (ref) {
+      return this.internal.delete(ref);
+    }
+    return false;
+  }
+
+  has(value: T): boolean {
+    const ref = this.find(value);
+    return ref ? true : false;
```
elle-j, line 81: If you think this makes it less readable, feel free to ignore.
```suggestion
    return !!ref;
```

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+  *[Symbol.iterator](): IterableIterator<T> {
+    for (const value of this.values()) {
+      yield value;
+    }
+  }
+
+  /**
+   * Finds the weak reference given a value
+   */
+  private find(value: T) {
```
elle-j, line 121:
```suggestion
  private find(value: T): binding.WeakRef<T> {
```

elle-j (approved):
Interesting 🙂 Most suggestions are adding explicit return types 👍

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  *keys(): IterableIterator<T> {
+    for (const keyRef of this.internal.keys()) {
+      const key = keyRef.deref();
+      if (key) {
+        yield key;
+      }
+    }
+  }
```
RedBeard0531, line 101:
```suggestion
  keys(): IterableIterator<T> {
    return this.values();
  }
```

or remove because there is no reason to call this method.

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { binding } from "./internal";
+
+/**
+ * Implements an iterable `WeakSet`.
+ * @internal
+ */
+export class IterableWeakSet<T extends object> implements Set<T> {
```
RedBeard0531, line 25: This is a bit of a weird type. I don't think it is correct to say it `implements Set<T>`, even if it has the right signatures. In addition to the O(1) vs O(N) issue, it also has the wrong semantics because `add(x)` always adds even if there is already a value equal to `x` in the set (due to wrapping in a new WeakRef). Consider:

```js
const mySet = new IterableWeakSet();
const someObj = {};
mySet.add(someObj);
mySet.add(someObj); // should be a no-op but isn't!
assert(mySet.size == 1); // Boom!
```

packages/realm/src/IterableWeakSet.ts
RedBeard0531: But you still have O(n) here.

packages/realm/src/IterableWeakSet.ts
RedBeard0531: The suggestion to use an Array only applied in the limited context of the other review where you never did partial removals. Doing partial removals with an Array is much more complicated, and something that makes sense with a Set. But in the other review your weren't actually benefiting from the "setness".

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.internal.add(new binding.WeakRef(value));
+    return this;
+  }
+
+  clear() {
+    this.internal.clear();
+  }
+
+  delete(value: T): boolean {
+    // This is a good time to get rid of entries that has been GC'ed
```
RedBeard0531, line 70: Maybe `find()` should do this since it is derefing everything anyway and it could speed up later calls to find.

RedBeard0531 (commented):
I don't love this type, for reasons I go into in more detail in the inline comments. I think we'd be better off integrating this with whatever this is being used for to produce a purpose-built tool. It seems like what you want is a way to call a method on all instances of a type that are still alive, but without keeping them alive longer. So lets build _that_ thing directly.

It may need some polyfilling to support platforms without `FinalizationRegistry`, but I think it should be possible.

packages/realm/src/IterableWeakSet.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  get [Symbol.toStringTag]() {
+    return IterableWeakSet.name;
+  }
+
+  forEach(cb: (value: T, value2: T, set: Set<T>) => void, thisArg?: unknown) {
+    this.internal.forEach((valueRef, valueRef2) => {
+      const value = valueRef.deref();
+      const value2 = valueRef2.deref();
```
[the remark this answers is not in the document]
kraenhansen: Sure. I just saw no need to make that assumption.
One alternative is to just throw as not implemented as I'm not using the method.

packages/realm/src/Realm.ts
```diff
@@ -261,19 +259,20 @@ export class Realm {
...
     // Close any realms not already closed
-    for (const weakRealm of RETURNED_REALMS) {
-      const realm = weakRealm.deref();
+    for (const realm of this.internals) {
       if (realm && !realm.isClosed) {
         realm.close();
       }
     }
-    RETURNED_REALMS.clear();
+    this.internals.clear();
```
RedBeard0531, line 275:
```suggestion
    Realm.internals.clear();
```

While technically correct, using `this` in a static method is a bit weird, unless you are specifically trying to get dynamic binding with subclasses.

packages/realm/src/IterableWeakRefs.ts
```diff
@@ -0,0 +1,80 @@
...
+  add(value: T) {
+    this.internal.add(new binding.WeakRef(value));
+    return this;
+  }
+
+  clear() {
+    this.internal.clear();
+  }
+
+  delete(value: T): boolean {
```
RedBeard0531, line 51: This and `clean()` are also unused. Do you have a plan to use them in a PR soon?

RedBeard0531 (approved):

packages/realm/src/IterableWeakRefs.ts
kraenhansen: Yep. I have another PR which depends on `delete`, but I could probably make `clean` private. It's called by `delete` in an attempt to partially metigate the unbounded growing of the internal set.

PR #5675, opened 2023-03-31
[v11] Suppress logging

kneth:
## What, How & Why?

The new core logger is noisy

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/src/js_realm.hpp
+++ b/src/js_realm.hpp
@@ -31,7 +31,9 @@
 #include "platform.hpp"
 #include "realm/binary_data.hpp"
 #include "realm/util/functional.hpp"
+#include "realm/util/logger.hpp"
 #include <exception>
+#include <memory>
 #include <stdexcept>
 #include <string>
 #include <set>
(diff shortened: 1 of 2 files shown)


papafe (approved):

elle-j (approved):
Yees! 😄

src/js_realm.hpp
```diff
@@ -598,6 +600,9 @@ inline typename T::Function RealmClass<T>::create_constructor(ContextType ctx)
         realm::disable_sync_to_disk();
     }
 
+    std::shared_ptr<util::NullLogger> logger = std::make_shared<util::NullLogger>();
+    util::Logger::set_default_logger(std::move(logger));
```
kraenhansen, line 604: Wouln't it make more sense to set the default log level?

kneth: `NullLogger` is setting it to `off` so no need

kraenhansen: But with this, users can't set the log level to something else and get an output?

kraenhansen: Fair.

kneth:
@kraenhansen The sync logger can still be used when the general logger is silent.

With sync log level = 'all':

Screenshot 2023-04-12 at 10 49 07

With sync log level = 'warn':

Screenshot 2023-04-12 at 10 49 07

src/js_realm.hpp
kneth: See my comment below

kraenhansen:
> With sync log level = 'warn'

Those two screenshots look identical to me.
I any way, I'm glad you verified this change can coexist without interfering with the existing behaviour.

PR #5678, opened 2023-04-01
Use src instead of d.ts for types in `@realm/react`

takameyer:
## What, How & Why?
The rollup plugin for type definitions was creating a very ugly output was unreadable for `@realm/react`.
For now, we will revert to using the src as the type definitions until we find a better way to generate the d.ts.

Before:
image

After:
image

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/package.json
+++ b/packages/realm-react/package.json
@@ -4,7 +4,7 @@
   "description": "React specific hooks and implementation helpers for Realm",
   "type": "module",
   "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "types": "src/index.tsx",
   "source": "src/index.tsx",
   "scripts": {
     "bundle": "wireit",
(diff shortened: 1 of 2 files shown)


kraenhansen:
Do you have any example of another package declaring it's TS source-code via the `types` field? If you don't like the rollup renames, I'd suggest using `tsc` to build the type declarations and include those instead of the source-code 👍

takameyer:
>  I'd suggest using tsc to build the type declarations 

I've spent a significant amount of time trying this out.  For some reason, it produces an `index.d.ts` that doesn't export the module.  it declares a module named `index`, which contains all of my exports, but they are only within the `index` scope.  I think I'll revert this change for now, but I'd like to dig into this again.

PR #5679, opened 2023-04-01
Fix Types for `useQuery` and `useObject`

takameyer:
## What, How & Why?
There were type errors when using models defined such as:
```
class Person extends Realm.Object<Person>{
```

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm-react/src/useObject.tsx
```diff
@@ -29,6 +29,7 @@ import { CollectionCallback, getObjectForPrimaryKey, getObjects } from "./helper
 export function createUseObject(useRealm: () => Realm) {
   function useObject<T>(type: string, primaryKey: T[keyof T]): (T & Realm.Object<T>) | null;
   function useObject<T extends Realm.Object>(type: { new (...args: any): T }, primaryKey: T[keyof T]): T | null;
+  function useObject<T extends Realm.Object<T>>(type: { new (...args: any): T }, primaryKey: T[keyof T]): T | null;
```
kraenhansen, line 32: I'd say it's simpler to change the overload above to `<T extends AnyRealmObject>` (which would be equivalent of `<T extends Realm.Object<any>>`)?

kraenhansen: That's what `Realm.objects` does anyway: https://github.com/realm/realm-js/blob/b7b4c6f7cb276ab47acf59ed2b263cd28bbd3bfa/packages/realm/src/Realm.ts#L982

takameyer: Yes, that is a better way.  Thanks!

PR #5683, opened 2023-04-01
[ISSUE_5032][ANDROID]  fix Error: Missing Realm constructor. When New Architecture is enabled

### LINKED ISSUE #5032: Error: Missing Realm constructor. When New Architecture is enabled

All the time

### Description

Error: Missing Realm constructor. Did you run "pod install" on react-native run-android. I try to build app on MacMini.
**Everything works fine if Hermes is enabled and New Architecture is disabled.** But, as soon as I turn on the new architecture, the application crashes after launch with such an error. 
The app is built from the newest Reactnative example.

my package:
 "react": "18.1.0",
    "react-native": "0.70.3",
"realm": "11.0.0",

I also tried to clear the build folder, put different versions of ream-11-rc-2 and so on

### Stacktrace & log output

```shell
ERROR  Error: Missing Realm constructor. Did you run "pod install"? Please see https://docs.mongodb.com/realm/sdk/react-native/install/ for troubleshooting, js engine: hermes
 LOG  Running "test-project" with {"fabric":true,"initialProps":{"concurrentRoot":true},"rootTag":41}
```

Yes, always

### Reproduction Steps

1. Init latest React-native 70 fabric example (NO expo)
2. Add latest realm package to project (11.0.0)
3. Hermes ON (now by default), New Architecture ON
4. react-native run-android
> Successfully build
x Crush on startup app

### Version

v11.0.0

Local Database only

### Are you using encryption?

No, not using encryption

Android 10

### Build environment

All from React-native 0.70.3 newest template.
Ruby 2.7.6

### Cocoapods version

_No response_

zabutok:
## What, How & Why?

This closes #5032.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm/react-native/android/src/main/java/io/realm/react/RealmReactPackage.java
+++ b/packages/realm/react-native/android/src/main/java/io/realm/react/RealmReactPackage.java
@@ -22,8 +22,10 @@ import com.facebook.react.ReactPackage;
 import com.facebook.react.TurboReactPackage;
 import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.module.annotations.ReactModule;
 import com.facebook.react.module.model.ReactModuleInfo;
 import com.facebook.react.module.model.ReactModuleInfoProvider;
+import com.facebook.react.turbomodule.core.interfaces.TurboModule;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


zabutok:
@cla-bot check

kraenhansen:
Thanks for your contribution. We want to support building with the new architecture enabled, but we also want to support building with it disabled. Specifically I want us to adopt the multi sourceset approach outlined in this tutorial: https://github.com/react-native-community/RNNewArchitectureLibraries/tree/feat/back-turbomodule#android-sourceset

zabutok:
@kraenhansen From my side I can confirm what is works for both flags when architecture enabled and disabled. Also I am not so good in new realm implementation but as I can understood realm can't be used without turbo modules with Hermes. At least it is not so easy. In my own project it works for both implementation.

packages/realm/react-native/android/src/main/java/io/realm/react/RealmReactPackage.java
```diff
@@ -40,19 +42,30 @@ public NativeModule getModule(String name, final ReactApplicationContext reactCo
 
     @Override
     public ReactModuleInfoProvider getReactModuleInfoProvider() {
+        Class<? extends NativeModule>[] moduleList =
+                new Class[] {
+                        RealmReactModule.class, RealmReactModule.class,
+                };
```
kraenhansen, line 48: Seems a bit complicated to use an array and loop, since we only have one module class? Is this code copied from somewhere?

zabutok: Not copied but i was trying to reproduce `react-native-reanimated` implementation because their implementation works pretty well

zabutok: Actually yes I think I can simplify it

kraenhansen:
I've verified this works both with the new architecture enabled and disabled.
Thanks a lot for your contribution @zabutok!

kraenhansen:
We just released this as [v12.0.0-alpha.2](https://github.com/realm/realm-js/releases/tag/v12.0.0-alpha.2).
Thanks again for your contribution.

PR #5684, opened 2023-04-02
Delete pr-realm-react.yml

kraenhansen:
## What, How & Why?

Now that we have [another workflow](https://github.com/realm/realm-js/blob/main/.github/workflows/package-unit-tests.yml) running these tests, I believe we can delete this.


takameyer (approved):
Yes, missed this.

PR #5689, opened 2023-04-03
Clear test state flag

kraenhansen:
## What, How & Why?

Instead of introducing a memory leak when running in production, this PR introduce a flag set from the tests to signal the SDK that it's allowed to store `WeakRef` instances into `Set`s without worrying about the memory leak, which would have happened if `clearTestState` was never called.

This also removes the `IterableWeakRefs`, as we no longer need something mimicking a `WeakSet`.

This also introduce the reset of SyncSession shared pointers and canceling unresolved async open tasks when clearing test state.


integration-tests/tests/src/index.ts
```diff
@@ -22,6 +22,8 @@ import { flags } from "realm";
 
 // TODO: Refactor tests to disable this
 flags.ALLOW_VALUES_ARRAYS = true;
+// We need this to call Realm.cleanTestState()
+flags.CLEAN_TEST_STATE = true;
```
elle-j, line 26: Typo? `CLEAR_` would align with the method name (rather than `CLEAN_`) 👍

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -57,6 +58,21 @@ function determineBehavior(config: Configuration, realmExists: boolean): OpenBeh
...
+  /**
+   * Cancels all unresolved `ProgressRealmPromise` instances.
+   * @internal
+   */
+  public static cancelAll() {
+    for (const promiseRef of ProgressRealmPromise.instances) {
+      const promise = promiseRef.deref();
+      if (promise) {
+        promise.cancel();
+      }
```
elle-j, line 72: Alternative
```suggestion
      promiseRef.deref()?.cancel();
```

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -262,6 +263,23 @@ const CONNECTION_LISTENERS = new Listeners<ConnectionNotificationCallback, Liste
 });
 
 export class SyncSession {
+  private static instances = new Set<binding.WeakRef<SyncSession>>();
```
elle-j, line 266: We have the `@internal` on the instances on `ProgressRealmPromise`. Did we decide which one we wanted to converge on?

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -262,6 +263,23 @@ const CONNECTION_LISTENERS = new Listeners<ConnectionNotificationCallback, Liste
...
+   * Resets the internal shared pointer of all instances returned since this message was last called.
+   * @internal
+   */
+  public static resetAllInternals() {
+    assert(flags.CLEAN_TEST_STATE, "Set the flag.CLEAN_TEST_STATE = true before calling this.");
+    for (const sessionRef of SyncSession.instances) {
+      const session = sessionRef.deref();
+      if (session) {
+        session.resetInternal();
+      }
```
elle-j, line 278: Alternative
```suggestion
      sessionRef.deref()?.resetInternal();
```

packages/realm/src/flags.ts
```diff
@@ -26,4 +26,9 @@ export const flags = {
    * Helps finding places where the app is depending on the now deprecated way of using the package.
    */
   THROW_ON_GLOBAL_REALM: false,
+  /**
+   * When enabled, calling `Realm.cleanTestState` will be callable.
```
elle-j, line 30:
```suggestion
   * When enabled, calling `Realm.clearTestState` will be callable.
```

integration-tests/tests/src/index.ts
```diff
@@ -22,6 +22,8 @@ import { flags } from "realm";
 
 // TODO: Refactor tests to disable this
 flags.ALLOW_VALUES_ARRAYS = true;
+// We need this to call Realm.cleanTestState()
```
elle-j, line 25:
```suggestion
// We need this to call Realm.clearTestState()
```

elle-j (approved):
Nice addition!

integration-tests/tests/src/index.ts
kraenhansen: Oh my. Thanks.

kraenhansen: I've updated this and hopefully all other occurrences 🤞

integration-tests/tests/src/index.ts
```diff
@@ -22,8 +22,8 @@ import { flags } from "realm";
 
 // TODO: Refactor tests to disable this
 flags.ALLOW_VALUES_ARRAYS = true;
-// We need this to call Realm.cleanTestState()
-flags.CLEAN_TEST_STATE = true;
+// We need this to call Realm.clearTestState()
+flags.CLEAR_TEST_STATE = true;
```
RedBeard0531, line 26:
```suggestion
flags.ENABLE_CLEAR_TEST_STATE = true;
```

Or `ALLOW`. The point is that setting the flag doesn't by itself clear the test state, it mearly *enables* (or allows) the test state to be cleared.

Also, reading this now, I think `clearTestState()` is slightly misnamed. Really it should be something like `clearStateAfterTest()` because it doesn't clear the _test_ state, it tries to reset the general state of the world. But I don't suggest changing that right now.

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -57,6 +58,19 @@ function determineBehavior(config: Configuration, realmExists: boolean): OpenBeh
 }
 
 export class ProgressRealmPromise implements Promise<Realm> {
+  /** @internal */
+  private static instances = new Set<binding.WeakRef<ProgressRealmPromise>>();
+  /**
+   * Cancels all unresolved `ProgressRealmPromise` instances.
+   * @internal
+   */
+  public static cancelAll() {
```
RedBeard0531, line 67: It seems weird to have a method in the middle of your instance variable declarations. Move this down?

packages/realm/src/app-services/SyncSession.ts
RedBeard0531: FWIW, there is less (but still not 0) benefit from keeping the `private static` members in the .d.ts vs the `private` instance members.

packages/realm/src/flags.ts
```diff
@@ -26,4 +26,9 @@ export const flags = {
    * Helps finding places where the app is depending on the now deprecated way of using the package.
    */
   THROW_ON_GLOBAL_REALM: false,
+  /**
+   * When enabled, calling `Realm.clearTestState` will be callable.
+   * This is disabled by default, mainly because the data-structures needed to support this, introduce minor memory leaks and are not intended for production use.
```
RedBeard0531, line 31:
```suggestion
   * This is disabled by default, mainly because the data-structures needed to support this, introduce minor memory leaks if clearTestState() is not called regularly and are not intended for production use.
```

packages/realm/src/flags.ts
```diff
@@ -26,4 +26,9 @@ export const flags = {
    * Helps finding places where the app is depending on the now deprecated way of using the package.
    */
   THROW_ON_GLOBAL_REALM: false,
+  /**
+   * When enabled, calling `Realm.clearTestState` will be callable.
```
RedBeard0531, line 30:
```suggestion
   * Enables calling internal, test-only functions like `Realm.clearTestState`.
```

* better to use active rather than passive voice where possible.
* the effects of this flag are already not limited to that function, so we might as well not lie.

RedBeard0531 (approved):

integration-tests/tests/src/index.ts
kraenhansen: I opted for `ALLOW` to align with the other existing flag.

And I agree, the function name is confusing. It's internal and can easily be renamed in another PR 👍

packages/realm/src/ProgressRealmPromise.ts
kraenhansen: I consider it "normal" to have all static members and methods before any instance methods and properties.

RedBeard0531: Fair enough. Could you at least separate the sections with a blank line then?

packages/realm/src/flags.ts
kraenhansen: I added this.

packages/realm/src/flags.ts
kraenhansen: I added this.

PR #5694, opened 2023-04-03
Add SSL sync config tests

### LINKED ISSUE #5693: Test SSL Sync config

elle-j:
## What, How & Why?

Adds tests for [PR 5507](https://github.com/realm/realm-js/pull/5507).

### Notes for Running Tests Locally

The tests cannot be run on Darwin platforms due to OpenSSL. These are some necessary local changes needed if running these tests on a Debian `linux_arm64` Node image and the `CMAKE_HOST_SYSTEM_PROCESSOR` is not `x86_64`:
* Uninstall the npm `clang-format` package (from `package.json` and `packages/realm/bindgen/vendor/realm-core/package.json`).
* Install `clang-format` using `apt-get install clang-format`.
* Update `packages/realm/bindgen/vendor/realm-core/bindgen/src/formatter.ts` in order for it to use the command `clang-format <args>` instead of `npx clang-format <args>`:
```diff
 export function executeCommand(cwd: string, command: string, ...args: string[]) {
+  // Replace `npx clang-format ...` with `clang-format ...`.
+  if (command === "npx" && args[0] === "clang-format") {
+    command = args.shift()!;
+  }
+

   // .. rest is unchanged ..
 }
```
  * Those ☝️ changes were needed due to the following error if using the npm package:
```
Error: This module doesn't bundle the clang-format executable for your platform. (linux_arm64)
Consider installing it with your native package manager instead.
```
* Update `packages/realm/bindgen/CMakeLists.txt` to skip using our provided toolchain and use whatever the host system uses.
```diff
     elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
-      if(NODE_ARCH STREQUAL "arm")
-        set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/armv7-linux-gnueabihf.toolchain.cmake")
-        # due to how multiarch works in debian this is needed to link to the correct system libraries
-        set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
-      elseif(NODE_ARCH STREQUAL "arm64")
-        set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/aarch64-linux-gnu.toolchain.cmake")
-        set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
-      elseif(NODE_ARCH STREQUAL "x64")
-        set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/x86_64-linux-gnu.toolchain.cmake")
-      endif()
-      add_link_options(-fuse-ld=gold)
-      add_link_options(-Wl,-z,noexecstack)
-      add_link_options(-Wl,--warn-execstack)
+        # Pick our toolchains only if we're on an x86_64 system
+        # (as in CI), otherwise use whatever the system is using.
+        if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64")
+            if(NODE_ARCH STREQUAL "arm")
+                set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/armv7-linux-gnueabihf.toolchain.cmake")
+                # due to how multiarch works in debian this is needed to link to the correct system libraries
+                set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
+            elseif(NODE_ARCH STREQUAL "arm64")
+                set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/aarch64-linux-gnu.toolchain.cmake")
+                set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
+            elseif(NODE_ARCH STREQUAL "x64")
+                set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/x86_64-linux-gnu.toolchain.cmake")
+            endif()
+        endif()
+        add_link_options(-fuse-ld=gold)
+        add_link_options(-Wl,-z,noexecstack)
+        add_link_options(-Wl,--warn-execstack)
     endif()
 endif()
```

This closes #5693.

## ☑️ ToDos
* [x] 🚦 Tests


integration-tests/tests/src/node/ssl.ts
```diff
@@ -0,0 +1,253 @@
...
+          // otherwise, if it's higher up in the chain and accepted by OpenSSL, they recommend
+          // returning `true`.
+          return isServerCertificate ? !!x509.checkHost(verifyObject.serverAddress) : true;
+        } else {
+          // If the certificate is not accepted by OpenSSL, Core recommends using an independent
+          // verification step. That step is represented here by checking the dates.
+          // (TODO: Add extra validation)
+          const now = new Date();
+          const isValid = now >= new Date(x509.validFrom) && now <= new Date(x509.validTo);
+          return isServerCertificate ? isValid && !!x509.checkHost(verifyObject.serverAddress) : isValid;
```
elle-j, line 174: As mentioned in the comment, I'm currently using the `validFrom/To` of the certificate as the extra validation step.

Not sure if the following should pass but it returns `false` for all certificates passed to this callback:
```ts
// Verify that the certificate was signed by the given public key.
const verified = x509.verify(x509.publicKey);
```

If comparing against e.g. `x509.fingerprint256` or another known certificate, we'd had to know where those are so that we can hardcode it here, but I haven't found any info on that.

integration-tests/tests/src/node/ssl.ts
elle-j, line 1: Not currently tested:
* When a string is passed as the `certificatePath` property.
* For this, we'd need known trusted certificates.

integration-tests/tests/src/node/ssl.ts
```diff
@@ -0,0 +1,253 @@
...
+  User,
+} from "realm";
+
+import { closeRealm } from "../utils/close-realm";
+import { createPromiseHandle } from "../utils/promise-handle";
+import { importAppBefore } from "../hooks";
+import { PersonSchema } from "../schemas/person-and-dog-with-object-ids";
+
+// IMPORTANT:
+// * Can only run on non-Apple machines, otherwise tests will await forever.
```
kneth, line 39: Do we know the root cause on Apple machines?

kneth (approved):

integration-tests/tests/src/node/ssl.ts
elle-j: On apple, [this no-op method](https://github.com/realm/realm-core/blob/34086a385127595167203786c03fca181c303b8b/src/realm/sync/network/network_ssl.cpp#L1182) is called.

integration-tests/tests/src/node/ssl.ts
```diff
@@ -0,0 +1,253 @@
...
+
+describe.skipIf(environment.missingServer, "SSL Configuration", function () {
```
takameyer, line 41: Perhaps we can also figure out a way to `skipIf` the users machine is a mac?

elle-j: I've tried locally on my Mac now and for MacOS we can import `node:os` and use:
```js
describe.skipIf(platform() === "darwin" || environment.missingServer, ..)
```

takameyer: That's great!  Let's do that 🙂

elle-j: ✅

integration-tests/tests/src/node/ssl.ts
```diff
@@ -0,0 +1,257 @@
...
+import { importAppBefore } from "../hooks";
+import { PersonSchema } from "../schemas/person-and-dog-with-object-ids";
+
+// IMPORTANT:
+// * Can only run on non-Apple machines, otherwise tests will await forever.
+function isMacOS() {
+  return platform() === "darwin";
+}
+
+describe.skipIf(isMacOS() || environment.missingServer, "SSL Configuration", function () {
```
kraenhansen, line 45:
```suggestion

describe.skipIf(platform() === "darwin" || environment.missingServer, "SSL Configuration", function () {
```

papafe:
I realised I'm still a reviewer on this. Does this still need to be open?

elle-j:
@papafe, yes it should still be open. Just updated (removed) you in the reviewers list 👍

kraenhansen:
Possibly for a follow-up PR, but I wondered if we could update the formatter.ts file to check if clang-format got downloaded from NPM before calling it and check (possibly via `which`) if another clang-format exists in the `PATH` 🤔 In other words, make it more robust to this scenario, out of the box.

On the other hand, we might require that the binding is generated and SDK is built on macOS and then only have instructions on how to run the tests from within an Arm64 Linux container, using a packed version of the `realm` package. Anecdotally, I doubt that all our build scripts are Windows friendly ...

elle-j:
Regarding checking if another clang-format exists: One of the issues is that even if another clang-format exists (the one installed by the native package manager) and we drop the `npx` command, we still get the error. Uninstalling the npm package seemed to resolve that.

PR #5697, opened 2023-04-04
Fix `realm.create` types and 'requiredProperties' on Realm.Object constructor

takameyer:
## What, How & Why?
The partial type was not added for the default creation mode.  This would cause a type error when creating an object without its optional types defined.
The typings for the Realm.Object constructor didn't make use of 'requiredProperties', which was available in v11.  Required properties will show a type error if said properties are not defined on creation (only works with constructor, not create).

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/Realm.ts
+++ b/packages/realm/src/Realm.ts
@@ -802,17 +802,15 @@ export class Realm {
    *       should not be used.
    * @returns A {@link RealmObject} or `undefined` if the object is asymmetric.
    */
-  create<T = DefaultObject>(type: string, values: RealmInsertionModel<T>, mode?: UpdateMode.Never): RealmObject<T> & T;
   create<T = DefaultObject>(
     type: string,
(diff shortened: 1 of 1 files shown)


kraenhansen:
On first glance this looks reasonable. It seems @realm/react needs an update to adopt to this?

takameyer:
@kraenhansen I don't think it does, but there seems to be a error in one of my rebases.  I'll fix that quick to make sure the tests pass.
Guess my biggest questions with this PR, is if the `create` `values` need to be typed differently based on the `mode`.  I'm not sure if I've somehow misunderstood something.  In v11, for the `never` mode the `values` are of typed `Unmanaged<T>` https://github.com/realm/realm-js/blob/main/types/index.d.ts#L1281-L1297

elle-j (commented):
I'm personally not certain of the effects of this change, so would be a bit of a reach for me as of now to approve or not.

PR #5698, opened 2023-04-05
Improve default logger for realm/react

takameyer:
## What, How & Why?
The default logger could be a bit smarter.  This will print a contextual log message with the level and use either `console.error`, `console.warn` and `console.log` based on the relevant level.  This also sets a default log level to WARN.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 import { isEqual } from "lodash";
 import React, { createContext, useContext, useLayoutEffect, useRef, useState } from "react";
-import Realm from "realm";
+import Realm, { LogLevel } from "realm";
 
 /**
  * Create a context containing the Realm app.  Should be accessed with the useApp hook.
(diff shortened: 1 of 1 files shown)


kraenhansen:
Why not add this to the `realm` package directly? I think a good default logger would be valuable more generally.

takameyer:
> Why not add this to the realm package directly?

I've wanted to do this for a while, so I have no argument against not doing it directly.

takameyer:
> @realm/react won't override the default

The unified logger will be at a different entry point and this would turn into a no-op.  So this will just be backwards compatible. I want to still ensure v11 users get the advantages of the `@realm/react` updates.

kraenhansen:
As I read the last comment on the unified logger issue, the `Realm.App.Sync.setLogger` will change to call `Realm.setLogger` and with this PR merged `@realm/react` will override whatever default the `realm` package injects on import, right? What am I missing? To me this change will interfere with the future default behavior of `realm@v12`.

takameyer:
@kraenhansen How would you recommend we support both versions of logging?  With this change, users wouldn't even have to update `realm` to get default logger.  After unified logging is realised, we can make this support both. 

> Realm.App.Sync.setLogger will change to call Realm.setLogger

This will be a breaking change then right?  We should probably make this a v13 change if `Realm.App.Sync.setLogger` doesn't work anymore.

packages/realm-react/src/AppProvider.tsx
```diff
@@ -36,8 +36,49 @@ type AppProviderProps = Realm.AppConfiguration & {
   logger?: (level: Realm.App.Sync.NumericLogLevel, message: string) => void;
 };
 
+// This is the default log level that would be useful for React Native.
+const DEFAULT_LOG_LEVEL: LogLevel = "warn";
+
+// Since the logger provides the numeric log level, we need to convert it to a
+// string for the log message.
+const getLogLevelString = (level: Realm.App.Sync.NumericLogLevel) => {
```
elle-j, line 44: `NumericLogLevel.All` and `NumericLogLevel.Off` as well 👍

elle-j (approved):
I like the split off the various console methods!

takameyer:
I may be closing this, as the effort @papafe is currently doing will probably nullify this, as well as the `logLevel` and `logger` properties of the `AppProvider`.  Part of me does want to release this, just so people that haven't updated to `v12` can reap the benefits, although I'm unsure if this comes into conflict with the recent core versions.

PR #5700, opened 2023-04-05
Align Realm Object Class Types with v11

takameyer:
## What, How & Why?
The new typings for the Realm Object Class require `schema`.  This unfortunately breaks the typings for users of the realm babel plugin.  This commit makes `schema` optional and adds a runtime assertion if it's not set.

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm/src/schema/types.ts
```diff
@@ -27,7 +27,7 @@ export type DefaultObject = Record<string, unknown>;
 export type Constructor<T = unknown> = { new (...args: any): T };
 export type RealmObjectConstructor<T extends RealmObject = RealmObject> = {
   new (...args: any): T;
-  schema: ObjectSchema;
+  schema?: ObjectSchema;
```
elle-j, line 30: Adding a comment to why this is needed would also be useful for our future selves 🙂

elle-j (approved):

packages/realm/src/schema/types.ts
takameyer: Good point!  Done 👍🏼

kraenhansen:
But I see why this needs to change ... It's pretty sad that we can't modify types from the babel plugin 🥺

takameyer:
> I see why this needs to change

Alternatively we could make ts-plugin, but that would probably be overkill.

PR #5708, opened 2023-04-05
Fix return type of `App.allUsers`

### LINKED ISSUE #5671: Return a record from `App.allUsers` instead of an array (v12)
`App.allUsers` is [intended](https://github.com/realm/realm-js/blob/v11/types/app.d.ts#LL289C13-L289C23) to return a record with the users ids as the keys, but is [currently](https://github.com/realm/realm-js/blob/fa519267504f0a4f898073349f0b2e7aa68452e1/packages/realm/src/app-services/App.ts#L220-L222) returning an array of users in v12.

elle-j:
## What, How & Why?

`App.allUsers` was returning an array of users but now returns a record with the `User.id` as the key and the `User` as the value as the [intended type](https://github.com/realm/realm-js/blob/715fa2dd70263c6ea86a84793d7c2f196db537ed/types/app.d.ts#L289).

This closes #5671.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Updated tests

### DIFF
--- a/packages/realm/src/app-services/App.ts
+++ b/packages/realm/src/app-services/App.ts
@@ -217,8 +217,13 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType 
     return currentUser ? User.get(currentUser) : null;
   }
 
-  public get allUsers(): User<FunctionsFactoryType, CustomDataType>[] {
-    return this.internal.allUsers.map((user) => User.get(user));
+  public get allUsers(): Readonly<Record<string, User<FunctionsFactoryType, CustomDataType>>> {
+    const result: Record<string, User<FunctionsFactoryType, CustomDataType>> = {};
+    for (const user of this.internal.allUsers) {
+      result[user.identity] = User.get(user);
+    }
+
+    return result;
   }
 
   public switchUser(): unknown {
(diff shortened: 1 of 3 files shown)


packages/realm/src/app-services/App.ts
```diff
@@ -217,8 +217,13 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
-  public get allUsers(): User<FunctionsFactoryType, CustomDataType>[] {
-    return this.internal.allUsers.map((user) => User.get(user));
+  public get allUsers(): Readonly<Record<string, User<FunctionsFactoryType, CustomDataType>>> {
+    const result: Record<string, User<FunctionsFactoryType, CustomDataType>> = {};
+    for (const user of this.internal.allUsers) {
+      result[user.identity] = User.get(user);
+    }
+
+    return result;
   }
```
elle-j, line 227: Note: `realm-web` uses `Object.fromEntries()`. Feel free to comment on what is preferred here.

packages/realm/src/app-services/App.ts
```diff
@@ -217,8 +217,13 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
 
-  public get allUsers(): User<FunctionsFactoryType, CustomDataType>[] {
-    return this.internal.allUsers.map((user) => User.get(user));
+  public get allUsers(): Readonly<Record<string, User<FunctionsFactoryType, CustomDataType>>> {
+    const result: Record<string, User<FunctionsFactoryType, CustomDataType>> = {};
+    for (const user of this.internal.allUsers) {
+      result[user.identity] = User.get(user);
+    }
+
+    return result;
```
kraenhansen, line 226: A more functional alternative - your call 🙂 

```suggestion
    return Object.fromEntries(this.internal.allUsers.map(user => {
      return [user.identity, User.get(user)];
    }))
```

kraenhansen (approved):

packages/realm/src/app-services/App.ts
kraenhansen: Lol - I just saw this now 🤦 If I have to choose, I probably like `fromEntries` because it's shorter (fewer lines and doesn't need the type declared on `result`) and I generally like a pure function over iterating and mutating an object (the `result` in this case).

elle-j: Gotcha, despite the extra iteration from needing to create the array (`.map`) passed to `fromEntries`?

kraenhansen (approved):

takameyer (approved):
LGTM :shipit:

PR #5718, opened 2023-04-12
Update usage of `assert.never()`

elle-j:
## What, How & Why?

Updates return type of `assert.never()`:
* From: `asserts value is never`
* To: `never`

This change allows TS to detect unreachable code after calling `assert.never()` without explicitly throwing.

Example old usage:
```ts
throw assert.never(eventName, "eventName");
```

Updated usage:
```ts
assert.never(eventName, "eventName");
// Still unreachable
```

## ☑️ ToDos
--


takameyer (approved):
LGTM!

PR #5719, opened 2023-04-12
Auth Hooks

takameyer:
## What, How & Why?
This introduces the hooks `useAuth` and `useEmailPasswordAuth`.  Included are:

* Tests to prove the returned methods on the hooks update status and report errors and success
** NOTE: I am not testing complete functionality of all the methods as this requires a lot of heavy lifting in our test harness and is frankly out of the scope of this project.
* Update the example app
* Update unit-test CI workflow to work with baas
* Fix type errors in app-importer, which is used in the tests for auth hooks
* TODO: Update Readme and Changelog

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm-react/src/__tests__/helpers.ts
```diff
@@ -0,0 +1,52 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2020 Realm Inc.
```
kneth, line 3:
```suggestion
// Copyright 2023 Realm Inc.
```

packages/realm-react/src/useEmailPasswordAuth.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { useApp } from "./AppProvider";
+import { AuthError, AuthResult, OperationState } from "./types";
+import { Realm, App, Object, User, Credentials } from "realm";
+
+interface UseEmailPasswordAuth {
+  /**
+   * Convenience function to login a user with an email and password - users
+   * could also call `logIn(Realm.Credentials.emailPassword(email, password)).
+   *
+   * TODO: Does it make sense to have this convenience function? Should we add
```
kneth, line 29: I think it makes sense as we are saving the user from writing some trivial code.

example/app/components/LoginScreen.tsx
```diff
@@ -3,53 +3,23 @@ import {View, Text, StyleSheet, TextInput, Pressable} from 'react-native';
...
-  None,
-  Loading,
-  LoginError,
-  RegisterError,
-}
+import {useAuth, useEmailPasswordAuth} from '@realm/react';
 
 export const LoginScreen = () => {
-  const app = useApp();
+  const {result: loginResult, logInWithEmailPassword} = useAuth();
```
kneth, line 9: So much code can be removed 😮 😎

kneth (approved):

packages/realm-react/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 ## vNext (TBD)
 
 ### Enhancements
+* Add authentication hooks, `useAuth` and `useEmailPasswordAuth`
+	Usage example:
+	```tsx
```
kneth, line 6: Gentle reminder to add an example

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+};
+```
+
+### `logInWithJWT`
+Log in with a JSON Web Token (JWT).
+```tsx
+const {logInWithJWT, result} = useAuth();
+
+const performLogin = () => {
+  const token = authorizeWithCustomerProvider(); //user defined function
```
kneth, line 407:
```suggestion
  const token = authorizeWithCustomerProvider(); // user defined function
```

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+};
+```
+
+### `logInWithGoogle`
+Log in Google.
+```tsx
+const {logInWithGoogle, result} = useAuth();
+
+const performLogin = () => {
+  const token = getGoogleToken(); //user defined function
```
kneth, line 418:
```suggestion
  const token = getGoogleToken(); // user defined function
```

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+};
+```
+
+### `logInWithApple`
+Log in with Apple.
+```tsx
+const {logInWithApple, result} = useAuth();
+
+const performLogin = () => {
+  const token = getAppleToken(); //user defined function
```
kneth, line 429:
```suggestion
  const token = getAppleToken(); // user defined function
```

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+};
+```
+
+### `logInWithFacebook`
+Log in with Facebook.
+```tsx
+const {logInWithFacebook, result} = useAuth();
+
+const performLogin = () => {
+  const token = getFacebookToken(); //user defined function
```
kneth, line 440:
```suggestion
  const token = getFacebookToken(); // user defined function
```

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
```diff
@@ -0,0 +1,241 @@
...
+      });
+      ({ appId } = await importApp(config.config));
+    });
+    it("can register and login with email/password, just by calling register", async () => {
+      const { result } = renderEmailPasswordAuth(appId, baseUrl);
+
+      await act(async () => {
+        result.current.register({ email: "test@test.com", password: "password" });
+        await waitFor(() => {
+          expect(result.current.result.pending).toEqual(true);
```
kneth, line 59: I suggest that we test that `results.current.result.success` is false as it is an invariant that they cannot be true simultaneously. Likewise in the following assertions.

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
```diff
@@ -0,0 +1,241 @@
...
+        disabled: false,
+      });
+      ({ appId } = await importApp(config.config));
+    });
+    it("logIn", async () => {
+      const { result } = renderEmailPasswordAuth(appId, baseUrl);
+      await act(async () => {
+        result.current.logIn({ email: "test@test.com", password: "password" });
+        await waitFor(() => {
+          expect(result.current.result.pending).toBeTruthy();
```
kneth, line 141: Why not `toEqual(true)`?

kneth (approved):

example/app/components/LoginScreen.tsx
```diff
@@ -3,53 +3,21 @@ import {View, Text, StyleSheet, TextInput, Pressable} from 'react-native';
...
-    }
-  }, [email, password, setAuthState, app]);
+
+  const handleLogin = useCallback(() => {
+    logInWithEmailPassword({email, password});
+  }, [logInWithEmailPassword, email, password]);
+
+  const handleRegister = useCallback(() => {
+    register({email, password});
+  }, [register, email, password]);
```
kraenhansen, line 20: I wonder if it makes sense to wrap these in `useCallback` or not ... I actually don't know how big a callback needs to be for it to be considered best practice to simply inline an arrow function instead 🤔

example/app/components/LoginScreen.tsx
```diff
@@ -71,42 +39,41 @@ export const LoginScreen = () => {
...
-      {authState === AuthState.RegisterError && (
+      {result.error && result.error.operation === AuthOperationName.LogIn && (
         <Text style={[styles.error]}>
-          There was an error registering, please try again
+          There was an error logging in, please try again{' '}
         </Text>
       )}
 
+      {result.error &&
+        result.error.operation === AuthOperationName.Register && (
```
kraenhansen, line 55: Is this a good place to use the `?.` operator?

```suggestion
      {result.error?.operation === AuthOperationName.Register && (
```

example/index.js
```diff
@@ -2,7 +2,6 @@
  * @format
  */
 
-import 'react-native-get-random-values';
```
kraenhansen, line 5: Don't we want to promote adding this? Or is it no longer needed?

example/sync.config.js
```diff
@@ -1,6 +1,6 @@
 export const SYNC_CONFIG = {
   // Set `enabled` to `true` to enable sync.
-  enabled: false,
+  enabled: true,
   // Add your Realm App ID here if sync is enabled.
-  appId: '<Your App ID>',
+  appId: 'example-oazfs',
```
kraenhansen, line 5: Was this committed by accident?

packages/realm-app-importer/src/AppConfigBuilder.ts
```diff
@@ -37,7 +37,7 @@ export type SyncConfig = {
 export type ServiceConfig = {
   name: string;
   type: string;
-  config: DisabledSyncConfig | PartitionSyncConfig | FlexibleSyncConfig;
```
kraenhansen, line 40: Instead of deleting this, I think https://github.com/realm/realm-js/blob/main/packages/realm-app-importer/src/AppConfigBuilder.ts#L126 should be renamed to `DisabledSyncConfig`.

packages/realm-react/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
...
+	```
```
kraenhansen, line 7: Looks like an example is missing 🙂

packages/realm-react/src/AppProvider.tsx
```diff
@@ -18,11 +18,43 @@
 import { isEqual } from "lodash";
 import React, { createContext, useContext, useLayoutEffect, useRef, useState } from "react";
 import Realm from "realm";
+import { AuthOperationName, AuthResult, OperationState } from "./types";
+
+type AppContextValue = {
+  app: Realm.App | null;
+};
```
kraenhansen, line 25: Any reason for packing this in an object now that it's a single property again?

```suggestion
type AppContextValue = Realm.App | null;
```

packages/realm-react/src/AppProvider.tsx
```diff
@@ -18,11 +18,43 @@
...
 /**
  * Create a context containing the Realm app.  Should be accessed with the useApp hook.
  */
-const AppContext = createContext<Realm.App | null>(null);
+const AppContext = createContext<AppContextValue>({
+  app: null,
+});
+
+type AppOpStateValue = {
+  authOperationStateHook: [AuthResult, React.Dispatch<React.SetStateAction<AuthResult>>] | null;
```
kraenhansen, line 35: -||- Any reason for wrapping this in an object?

packages/realm-react/src/AppProvider.tsx
```diff
@@ -18,11 +18,43 @@
...
+};
+
+const AuthOpStateContext = createContext<AppOpStateValue>({
```
kraenhansen, line 38: I'd suggest avoiding the abbreviation here and "state context" seem a bit redundant here.

```suggestion
const AuthOperationContext = createContext<AppOpStateValue>({
```

packages/realm-react/src/__tests__/helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+const credentials = getCredentials();
+
+const importer = new AppImporter({
+  baseUrl: realmBaseUrl,
+  credentials,
+});
+
+export async function importApp(config: AppConfig): Promise<{ appId: string }> {
+  const { appId } = await importer.importApp(config);
+  return { appId };
```
kraenhansen, line 74:
```suggestion
  return importer.importApp(config);
```

packages/realm-react/src/__tests__/useAuth.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// These will all fail since we haven't configured any of the authentication protocols
+// The tests for the authentication methods themselves should be written elsewhere
+describe("useAuth", () => {
+  describe("all methods are callable and report a state", () => {
+    let appId: string;
+    beforeAll(async () => {
+      const config = new AppConfigBuilder("test-app");
+      ({ appId } = await importApp(config.config));
```
kraenhansen, line 42: Funky 🎶 🙂

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
```diff
@@ -0,0 +1,241 @@
...
+////////////////////////////////////////////////////////////////////////////
+import React from "react";
+import { AppProvider } from "../AppProvider";
+import { waitFor, renderHook, act } from "@testing-library/react-native";
+
+import { AppConfigBuilder } from "@realm/app-importer";
+import { App, EmailPasswordAuth } from "realm";
+import { useEmailPasswordAuth } from "../useEmailPasswordAuth";
+import { OperationState } from "../types";
+import { baseUrl, importApp, testAuthOperation } from "./helpers";
```
kraenhansen, line 26: Since these are not used by any other files, it might be better to simply inline them into this file?

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
kraenhansen, line 141: I suspect there might be a race condition here, if the error happens so fast that it never sees the `pending` state?

packages/realm-react/src/types.ts
```diff
@@ -0,0 +1,93 @@
...
+  LogInWithFunction = "logInWithFunction",
+  LogOut = "logOut",
+  Register = "register",
+  Confirm = "confirm",
+  ResendConfirmationEmail = "resendConfirmationEmail",
+  RetryCustomConfirmation = "retryCustomConfirmation",
+  SendResetPasswordEmail = "sendResetPasswordEmail",
+  ResetPassword = "resetPassword",
+  CallResetPasswordFunction = "callResetPasswordFunction",
+  None = "none",
```
kraenhansen, line 37: Wouldn't it make more sense to represent this as `undefined` or `null`?

```suggestion
```

packages/realm-react/src/__tests__/helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+    expectedResult();
+  });
+}
+
+const { realmBaseUrl = "http://localhost:9090" } = process.env;
+
+export const baseUrl = realmBaseUrl;
+
+function getCredentials(): Credentials {
+  const { publicKey, privateKey, username = "unique_user@domain.com", password = "password" } = process.env;
```
kraenhansen, line 49: Still feel these names are a bit too general for the `process.env` namespace, would like an `baas` prefix or similar .. also environment variables tend to be "ALL_CAPS".

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+  error // Error based object or undefined
+}
+```
+
+This can be used to manage the state of the current login operation.
+
+### `logIn`
+Log in with a Realm.Credentials instance. This allows login with any authentication mechanism supported by Realm. If this is called when a user is currently logged in, it will switch the user.  Typically the other methods from `useAuth` would be used.
+```tsx
+const {logIn, result} = useAuth();
```
kraenhansen, line 364: Perhaps you want to return something which renders the result too? Like a conditional spinner when `result.pending` or something.

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+   * operation is called.
+   */
+  error // Error based object or undefined
+}
+```
+
+This can be used to manage the state of the current login operation.
+
+### `logIn`
+Log in with a Realm.Credentials instance. This allows login with any authentication mechanism supported by Realm. If this is called when a user is currently logged in, it will switch the user.  Typically the other methods from `useAuth` would be used.
```
kraenhansen, line 362:
```suggestion
Log in with a Realm.Credentials instance. This allows login with any authentication provider supported by Atlas App Services. If this is called when a user is currently logged in, it will switch the user.  Typically the other methods from `useAuth` would be used.
```

kraenhansen: > Typically the other methods from `useAuth` would be used.

I don't think I understand this.

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+  logInWithApiKey(key);
+};
+```
+
+### `logInWithEmailPassword`
+Log in with Email/Password.
+```tsx
+const {logInWithEmailPassword, result} = useAuth();
+const [email, setEmail] = useState();
+const [password, setPassword] = useState();
```
kraenhansen, line 394:
```suggestion
const [email, setEmail] = useState("");
const [password, setPassword] = useState("");
```

packages/realm-react/README.md
kraenhansen, line 407: Alternatively take this as an argument for the function. Just a thought.

```suggestion
const performLogin = (token) => { // a token provided by some other authentication service
```

packages/realm-react/src/useAuth.tsx
```diff
@@ -0,0 +1,196 @@
...
+ */
+interface UseAuth {
+  /**
+   * Log in with a {@link Realm.Credentials} instance. This allows login with any
+   * authentication mechanism supported by Realm.
+   * If this is called when a user is currently logged in, it will switch the user.
+   *
+   * @returns A `Realm.User` instance for the logged in user.
+   */
+  logIn(credentials: Realm.Credentials): Promise<Realm.User | void>;
```
kraenhansen, line 45: In which was will this (or any of the other methods) return a promise of `void`? 🤔

packages/realm-react/src/useAuthOperation.tsx
```diff
@@ -0,0 +1,72 @@
...
+  operationName,
+  onSuccess = () => undefined,
+}: {
+  operation: (...args: Args) => Promise<Result | void>;
+  operationName: AuthOperationName;
+  onSuccess?: (...args: Args) => void;
+}) {
+  const [result, setResult] = useAuthResult();
+
+  return useCallback<(...args: Args) => ReturnType<typeof operation>>(
```
kraenhansen, line 34: I wouldn't expect this to have any (positive) effect, since the `operation` is not memoized from the caller, it will invalidate this memoization on every render.

packages/realm-react/src/useAuthOperation.tsx
```diff
@@ -0,0 +1,72 @@
...
+        })
+        .catch((error) => {
+          const authError = new AuthError(operationName, error);
+          setResult({
+            state: OperationState.Error,
+            pending: false,
+            success: false,
+            error: authError,
+            currentOperation: operationName,
+          });
```
kraenhansen, line 67:
```suggestion
          setResult({
            state: OperationState.Error,
            pending: false,
            success: false,
            error: new AuthError(operationName, error),
            currentOperation: operationName,
          });
```

packages/realm-react/src/AppProvider.tsx
```diff
@@ -18,11 +18,43 @@
...
+
+type AuthOpStateProviderProps = { children: React.ReactNode };
+
+const AuthOpStateProvider: React.FC<AuthOpStateProviderProps> = ({ children }) => {
+  const [authOpResult, setAuthOpResult] = useState<AuthResult>({
+    state: OperationState.NotStarted,
+    pending: false,
+    success: false,
+    error: undefined,
+    currentOperation: AuthOperationName.None,
```
kraenhansen, line 50: Since there are only supposed to be a single operation in flight at any given time, I think `current` is redundant.

```suggestion
    operation: AuthOperationName.None,
```

packages/realm-react/src/useEmailPasswordAuth.tsx
```diff
@@ -0,0 +1,190 @@
...
+   */
+  logIn(args: { email: string; password: string }): Promise<Realm.User | void>;
+
+  /**
+   * Register a new user. By default this will login the newly registered user
+   * after they have been successfully registered, unless the
+   * `loginAfterRegister` property is `false`.
+   *
+   * @returns A `Realm.User` instance for the logged in user if
+   * `loginAfterRegister` is not `false`, `void` otherwise.
```
kraenhansen, line 52: Small improvement (which could be considered as a future enhancement): It might make sense to provide overloads for these two call patterns. That way a user wouldn't have to null check the returned promise value before accessing it if they don't explicitly provide `loginAfterRegister: false`.

kraenhansen (approved):
Only minor suggestions.
Great work!! 🚀

.github/workflows/package-unit-tests.yml
```diff
@@ -19,27 +19,50 @@ jobs:
...
+
+      - name: Set baas env
+        if: ${{matrix.variant.use-baas}}
+        run: echo "realmBaseUrl=${{ steps.baas-config.outputs.url }}" >> $GITHUB_ENV
+
       # Install the root package to get dev-dependencies
       # (--ignore-scripts to avoid downloading or building the native module)
-      - run: npm ci --ignore-scripts
-      - run: npm test --workspace ${{ matrix.workspace }}
+      - run: npm ci
```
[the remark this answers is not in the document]
takameyer: It was needed, as it wasn't doing any of the `wireit` parts associated with running `npm install`

example/app/components/LoginScreen.tsx
takameyer: It's not critical, but more of a best practice sort of thing.  Any function that's passed into a component should generally be wrapped in `useCallback` to ensure we aren't pushing an unneeded render into a child.  I think it's less about the performance of creating a function again, as it is holding onto a stable reference on re-renders

example/index.js
takameyer: Is this no longer needed?  I definitely don't want an example to start with warnings about this.

example/sync.config.js
takameyer: 🙈

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
takameyer: They are used in `useAuth` as well

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
takameyer: Can you expand on this?  Where would the race condition be?  In the test, or the lib?

packages/realm-react/src/types.ts
takameyer: I could live with `null`

packages/realm-react/src/__tests__/helpers.ts
takameyer: We should address this globally then, as this is not unique to here

packages/realm-react/README.md
takameyer: Neither do I 😅

packages/realm-react/README.md
takameyer: Guess it depends on what is calling this.  Honestly, I think most of these examples should be clarified in the mdb docs with a bit more context.

packages/realm-react/src/useAuth.tsx
takameyer: If it fails

.github/workflows/package-unit-tests.yml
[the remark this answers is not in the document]
kraenhansen: Perhaps we could instead make those invocations explicit instead of relying on `postinstall` scripts on CI?
This was added to ensure we're not accidentally running tests against a prebuilt binary downloaded from NPM.

example/index.js
kraenhansen: > Is this no longer needed?

I think it is, that's why I'm curious why you suggest removing it? 😄

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
kraenhansen: Oh okay - I didn't realise that.

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
kraenhansen: In the test. `waitFor` will call the callback and poll until the callback no longer throws, but if the network is fast, it might go straight into the `error` state before it gets into `pending`, which would leave the test hanging indefinitely.

packages/realm-react/src/__tests__/helpers.ts
kraenhansen: This is the first place they're read from `process.env` .. in the integration tests they are read from `environment` (I know - not the best name, but `context` was sort of taken by Mocha) which is passed through the `MOCHA_REMOTE_CONTEXT` environment variable.

packages/realm-react/src/useAuth.tsx
kraenhansen: That should reject the promise? Not resolve it with `undefined`?

packages/realm-react/src/useAuthOperation.tsx
kraenhansen: Actually .. this should re-throw instead.

```suggestion
          const authError = new AuthError(operationName, error);
          setResult({
            state: OperationState.Error,
            pending: false,
            success: false,
            error: authError,
            currentOperation: operationName,
          });
          throw authError;
```

packages/realm-react/src/useAuth.tsx
takameyer: Then would the correct type be `Promise<Resul.User | AuthError>`?

example/index.js
takameyer: Sorry, I read this as your suggestion :D I did not intend to remove it.  Unsure how that happened.  Good find!

packages/realm-app-importer/src/AppConfigBuilder.ts
takameyer: Sounds good to me.

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
takameyer: On one point they should be consistent.  `toBeTruthy` would guarantee that the value resolves to true in an `if` statement.

takameyer: It's set to `pending` right before the operation is invoked, so I doubt it would be possible.  I could be wrong though.  In any case, for the most part we should be waiting for the result.  Although I do want to verify in the tests that `pending` is set to true after the method is invoked.  I guess this could be done though mocking rather than assuming 😂

packages/realm-react/src/useAuthOperation.tsx
takameyer: You may be right.  I'll have to do a deeper dive into `useCallback`.  I may just be being defensive here.

packages/realm-react/src/useEmailPasswordAuth.tsx
takameyer: I'll have to give this some thought.  I mostly assume these hooks would be used in the `fallback` of `UserProvider`.  In this case, the actually returned result will never get used, since the component will be demounted by then.

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
kraenhansen: > It's set to pending right before the operation is invoked

Even so, it could still fail and switch to error before the callback passed to `waitFor` is invoked (depending on the implementation of `waitFor`).

packages/realm-react/src/__tests__/helpers.ts
kraenhansen: > I could also read them out of `CONTEXT`

That would make the tests align more with the integration tests 👍

packages/realm-react/src/useAuth.tsx
kraenhansen: Nope. Just `Promise<Result.User>` there's no generic type for the value of the rejection.

packages/realm-react/src/useAuthOperation.tsx
kraenhansen: > We shouldn't throw in hooks

You'll be throwing inside a `then`/ `catch` callback, which doesn't bubble into the calling hook, it will simply reject the resulting promise.

packages/realm-react/src/useAuth.tsx
takameyer: I'll do some reviewing here on how promises should reject in the error case.  Throwing is unsafe in a hooks environment.

packages/realm-react/src/useEmailPasswordAuth.tsx
```diff
@@ -0,0 +1,190 @@
...
+   */
+  register(args: {
+    email: string;
+    password: string;
+    // Defaults to true so that a newly registered user is immediately logged
+    // in.
+    loginAfterRegister?: boolean;
+  }): Promise<Realm.User | void>;
```
elle-j, line 60: Regarding `loginAfterRegister`, I think it could potentially create confusion having this default to `true` since the function has additional semantics than the name (it also logs in). For apps using Realm today, I'd assume that many users set up their apps so that after `register()` is successfully called, they also call `logIn()`. Even if this hook now provides an option to not have to explicitly perform the login step (which I think is great), the call will look the same (`register(credentials)`) but the semantics will differ.

Some alternatives are:
1. Provide `register()` without the `loginAfterRegister` and have it just be responsible for registration.
2. Provide `register()` with `loginAfterRegister` (or `logInAfterRegister`) default set to `false`.
3. Provide (a) `registerAndLogIn()` as well as (b) `register()` without the `loginAfterRegister`.

I think I'd prefer Alt. 3 (or some variation of that name). It also aligns with all the other convenience methods where we differentiate them via separate methods rather than an option param.

I'd be interested to hear your thoughts on this.

packages/realm-react/src/useEmailPasswordAuth.tsx
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * The {@link AuthResult} states returned from this hook are "global" for all
+ * components under a given AppProvider, as only one operation can be in progress
+ * at a given time (i.e. we will store the states on the context). This means that,
+ * for example, multiple components can use the `useAuth` hook to access
+ * `result.pending` to render a spinner when login is in progress, without
+ * needing to pass that state around or store it somewhere global in their app
+ * code.
+ */
+interface UseEmailPasswordAuth {
```
elle-j, line 37: The docs for many of the methods on this interface says `@returns void` but should be a `Promise<void>` 👍

packages/realm-react/src/useEmailPasswordAuth.tsx
```diff
@@ -0,0 +1,190 @@
...
+    operationName: AuthOperationName.CallResetPasswordFunction,
+  });
+
+  const resetPassword = useAuthOperation({
+    operation: ({ password, token, tokenId }) => app.emailPasswordAuth.resetPassword({ password, token, tokenId }),
+    operationName: AuthOperationName.ResetPassword,
+  });
+
+  const logOut = useAuthOperation({
+    operation: () => (app.currentUser ? app.currentUser.logOut() : Promise.resolve()),
```
elle-j, line 174: Same suggestion a my earlier one.
```suggestion
    operation: async () => app.currentUser?.logOut(),
```

elle-j (approved):
It was really great to read through this PR and get more familiarized with these upcoming hooks! Awesome work and can't wait for users to get to leverage them 🚀

.github/workflows/package-unit-tests.yml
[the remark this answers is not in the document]
takameyer: Testing this again now and it seems the `--ignore-scripts` was not necessary.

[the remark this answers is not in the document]
kraenhansen: But have you tried making changes to the native code? I suspect that it'll download the prebuilt binary from NPM instead of building from source.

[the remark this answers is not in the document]
takameyer: Sorry, my previous comment was misleading.  Everything works fine with `--ignore-scripts` so I have added it back in.

example/app/components/LoginScreen.tsx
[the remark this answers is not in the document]
takameyer: That's a good idea.  I'll just inline the methods where they are invoked.  They are already wrapped in a useCallback internally

.github/workflows/package-unit-tests.yml
```diff
@@ -19,27 +19,50 @@ jobs:
...
-          - '@realm/network-transport'
-          - '@realm/babel-plugin'
-          - '@realm/react'
-    name: ${{ matrix.workspace }} unit tests
+        variant:
+          - {workspace: realm}
+          - {workspace: '@realm/bindgen'}
+          - {workspace: '@realm/network-transport'}
+          - {workspace: '@realm/babel-plugin'}
+          - {workspace: '@realm/react', use-baas: true}
```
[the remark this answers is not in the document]
takameyer: Done

packages/realm-react/src/useAuthOperation.tsx
takameyer: Good catch.  I'll update the callers so that they provide a memoized operation to this.

example/app/components/LoginScreen.tsx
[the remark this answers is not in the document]
takameyer: I'm trying to think of a way to do this where these functions don't update.  If I make it argument based, then below I will have to do 
```
onPress={() => handleRegister({email, password})}
```
which will also reinstatiate on every state change.  If the state were implicitly passed to the `handleRegister`, this would work, but unfortunately, that's not how it works :D

example/app/AppNonSync.tsx
```diff
@@ -6,6 +6,7 @@ import {TaskManager} from './components/TaskManager';
 import {useQuery} from '@realm/react';
 
 export const AppNonSync = () => {
+  debugger;
```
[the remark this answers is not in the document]
takameyer: Yep

example/app/components/LoginScreen.tsx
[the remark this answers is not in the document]
takameyer: As per our discussion, since there is no possible way to avoid having this function reinstantiate, we will go the path of less characters and just inline the function to the `onPress` event.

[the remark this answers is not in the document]
elle-j: Agreed 🙂

takameyer: I had discussed this with LJ.  If this were a web application with a form submit event, then we could setup a nice `useCallback` the was memoized and could get the form field values from the event, rather than the state.  Since this isn't possible in RN without writing our own formlike library (or using something else like Formik), using inline arrow functions is the path of least resistance.

example/app/AppWrapperSync.tsx
```diff
@@ -19,7 +20,15 @@ export const AppWrapperSync: React.FC<{
...
           <RealmProvider
             schema={schemas}
-            sync={{flexible: true, onError: error => console.error(error)}}>
+            sync={{
+              flexible: true,
+              existingRealmFileBehavior: {
+                type: OpenRealmBehaviorType.DownloadBeforeOpen,
+                timeOut: 1000,
+                timeOutBehavior: OpenRealmTimeOutBehavior.OpenLocalRealm,
+              },
```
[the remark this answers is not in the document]
takameyer: We do [now](https://github.com/realm/realm-js/issues/5887)

example/app/AppWrapperSync.tsx
```diff
@@ -19,7 +20,15 @@ export const AppWrapperSync: React.FC<{
...
+              onError: (_, error) => console.error('sync error', error),
```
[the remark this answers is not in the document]
takameyer: I guess we can remove this now with the default logger

packages/realm-react/src/__tests__/useAuth.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    });
+    it("logIn", async () => {
+      const { result } = renderAuth(appId, baseUrl);
+      await testAuthOperation({
+        authOperation: () => result.current.logIn({ email: "test@test.com", password: "password" }),
+        result,
+        expectedResult: () => expect(result.current.result.error).toBeDefined(),
+      });
```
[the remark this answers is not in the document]
takameyer: > the Atlas App is not set up with the auth providers?

Yes, no auth providers have been setup.  

> is expect(... .error).toBeDefined() meant to be changed afterward or do we expect an error here?

We expect to see an error in the result.  If afterwards, an auth method is called again, the error will be reset to null.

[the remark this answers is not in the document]
takameyer: Basically the test expects the operations `result` state to settle into an `expectedResult`.

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+```tsx
+const {logOut, result} = useAuth();
+
+const performLogOut = () => {
+  logOut();
+};
+```
+
+## `useEmailPasswordAuth`
+This hook is similar to `useAuth`, but specifically offers operations around Email/Password authentication.  This includes methods around resetting passwords and confirming users.  It returns the same `result` object as `useAuth`.
```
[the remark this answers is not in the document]
takameyer: I believe I intended to, but got distracted.  Good catch!

packages/realm-react/README.md
```diff
@@ -323,8 +323,149 @@ const SomeComponent = () => {
...
+   * if `state === OperationState.Error`, and will be cleared each time the
+   * operation is called.
+   */
+  error // Error based object or undefined
+}
+```
+
+This can be used to manage the state of the current login operation.
+
+### `logIn`
```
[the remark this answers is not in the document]
takameyer: 1. As I am assuming this is all done within the `fallback` of `UserProvider`, the result of logIn would never get a chance to be used, as the application would then render as soon as `currentUser` is set on the `Realm.App`.  But perhaps for the example, it would be fine to show some sort of logging of the result.
2. Perhaps I could just wrap it in a `useEffect` that fires immediately on render, for this example.

packages/realm-react/CHANGELOG.md
takameyer: I think i'll just link to the README section, as adding usage examples here would be a bit overwhelming

packages/realm-react/CHANGELOG.md
takameyer: Adding a link to the readme section

packages/realm-react/src/useAuth.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Log in with the Anonymous authentication provider.
+   *
+   * @returns A `Realm.User` instance for the logged in user.
+   * @see https://www.mongodb.com/docs/atlas/app-services/authentication/anonymous/
```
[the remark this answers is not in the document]
takameyer: Was inspired by the React hooks, which also have a `@see`

packages/realm-react/src/useAuth.tsx
```diff
@@ -0,0 +1,196 @@
...
+   */
+  logInWithJWT(token: string): Promise<Realm.User | void>;
+
+  /**
+   * Log in with Google.
+   *
+   * @returns A `Realm.User` instance for the logged in user.
+   * @see https://www.mongodb.com/docs/atlas/app-services/authentication/google/
+   */
+  logInWithGoogle(params: { idToken: string } | { authCode: string }): Promise<Realm.User | void>;
```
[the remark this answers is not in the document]
takameyer: I am mirroring what is set at the SDK level.  Perhaps I should think of a way to do this generically, so that if at some point the underlying Credential changes, the hooks would still function.  I do agree that the naming is inconsistent and could be misleading.

packages/realm-react/src/useAuth.tsx
```diff
@@ -0,0 +1,196 @@
...
+    operationName: AuthOperationName.LogInWithFacebook,
+  });
+
+  const logInWithFunction = useAuthOperation({
+    operation: (payload: Record<string, unknown>) => app.logIn(Realm.Credentials.function(payload)),
+    operationName: AuthOperationName.LogInWithFunction,
+  });
+
+  const logOut = useAuthOperation({
+    operation: () => (app.currentUser ? app.currentUser.logOut() : Promise.resolve()),
```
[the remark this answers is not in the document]
takameyer: I'll have to try this out.  Good idea

packages/realm-react/src/useEmailPasswordAuth.tsx
takameyer: Glad you brought this up, as I have been a bit torn on this as well.  It is not clear from the method that some side effect is happening by default.  I think I would actually lean towards option 3.  Then it is clear what is being attempted. 
One thought I'm having now though, is that if the LogIn fails, it may be unclear that the registration was successful.  There would be no roll-back in this case.  So maybe option 1 is just the better way to go to ensure the business logic can be setup exactly as the developer intends.

packages/realm-react/README.md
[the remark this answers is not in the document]
elle-j: Yeah wrapping in `useEffect` might be suitable 👍 Do what you think makes most sense 🙂

packages/realm-react/src/__tests__/useAuth.test.tsx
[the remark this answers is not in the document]
elle-j: Thanks for the explanation! 🙂

packages/realm-react/src/useAuth.tsx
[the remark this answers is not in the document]
elle-j: Ah okay gotcha!

packages/realm-react/src/useAuth.tsx
```diff
@@ -0,0 +1,196 @@
...
+  });
+
+  const logInWithApiKey = useAuthOperation({
+    operation: (key: string) => app.logIn(Realm.Credentials.apiKey(key)),
+    operationName: AuthOperationName.LogInWithApiKey,
+  });
+
+  const logInWithEmailPassword = useAuthOperation({
+    operation: (params: { email: string; password: string }) =>
+      app.logIn(Realm.Credentials.emailPassword(params.email, params.password)),
```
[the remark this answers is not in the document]
elle-j: That [method](https://github.com/realm/realm-js/blob/afb84f0a4426db0ed1e10632855c776bbd5a017b/packages/realm/src/app-services/Credentials.ts#LL68C5-L68C5) should be overloaded, so should work :)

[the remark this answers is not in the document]
takameyer: Ah!  I was mislead by the mouse over documentation!  Well then, I will definitely use the object :D

packages/realm-react/src/useEmailPasswordAuth.tsx
elle-j: Interesting point you mentioned with the "failed login but successful registration" scenario. I think Option 3 can still allow users to set up custom logic (providing `registerAndLogIn()` as well as `logIn()` and `register()`), while still catering to the use case of `registerAndLogIn()`. Since registering and logging in would throw different error messages, users could also differentiate between the errors in their `catch` clauses.

With our recent discussion regarding email confirmation, perhaps the more common case in production is to actually perform confirmation logic after registration before logging in.

takameyer: I'm going to pull the trigger on option 3.  The idea with doing both at the same time, was to make it even quicker to get started, but I don't think it's really worth the cost here.  We can always decide to add a combined version later, but we can't back out if we do it now.

elle-j: Good reasoning, go for it 👍

packages/realm-react/src/useAuthOperation.tsx
```diff
@@ -0,0 +1,72 @@
...
+            state: OperationState.Success,
+            pending: false,
+            success: true,
+            error: undefined,
+            currentOperation: operationName,
+          });
+          onSuccess(...args);
+          return res;
+        })
+        .catch((error) => {
```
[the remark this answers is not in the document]
takameyer: I came to the conclusion that automatic login after register was too side-effecty.  I believe it would only be nice for the templates, but not for a "real-world" authentication experience.  With this being removed, the `onSuccess` is no longer needed.

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
takameyer: As per Kræns suggestion, we won't test against pending, as this could be unstable and a race condition.  Rather we will wait for the expected ended result.

packages/realm-react/src/AppProvider.tsx
takameyer: I'll also set this to a single property

packages/realm-react/src/AppProvider.tsx
```diff
@@ -18,11 +18,43 @@
...
+  });
+  return (
+    <AuthOpStateContext.Provider value={{ authOperationStateHook: [authOpResult, setAuthOpResult] }}>
```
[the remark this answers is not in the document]
takameyer: I'm going to get this released first and then look at this: https://github.com/realm/realm-js/issues/5897

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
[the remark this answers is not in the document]
takameyer: Good call.  I've added this, but I think i've also uncovered a bug in bindgen :D Without recreating an instance of the `App`, the currentUser seems to just go null after the second login.  In the new instance, the user object is set.

packages/realm-react/src/useAuth.tsx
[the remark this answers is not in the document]
takameyer: Also added this to the EmailPasswordAuth

packages/realm-react/src/useAuth.tsx
takameyer: Now everything returns void :100:

packages/realm-react/src/useAuthOperation.tsx
takameyer: As discussed, we will not throw, but we will process the rejection instead of catching the exception

packages/realm-react/src/__tests__/useEmailPasswordAuth.test.tsx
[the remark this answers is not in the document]
elle-j: Investigation time 🔍

packages/realm-react/src/AppProvider.tsx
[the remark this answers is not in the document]
elle-j: Sounds good 👍

PR #5722, opened 2023-04-12
Use `SchemaParseError` for invalid schemas

### LINKED ISSUE #5198: Create a `SchemaParseError` that the new parser should use.
The new schema parser from #5178 is currently throwing either an `Error` or `AssertionError` for invalid schemas. A `SchemaParseError` that extends `Error` would provide a more suitable exception to throw.

Notes:
* Wherever `assert()` is used, the error object will have to be passed to it (rather than just the message).
* This change may require integrating with the new updates in Core error handling.

elle-j:
## What, How & Why?

Rather than throwing an `AssertionError` or `Error` when opening a realm with invalid schemas, a `SchemaParseError` is introduced to use instead.

This closes #5198.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (small update to assert instance of the error)


packages/realm/src/index.ts
elle-j, line 1: I also exposed `AssertionError` and `TypeAssertionError` since there mentioned in various places in our TS docs as being thrown.

takameyer: In my opinion, if those errors ever land in user space, then they should be exported.  This way users can type their catch statements and possible extend these if needed.

takameyer (approved):
Sounds like a good change to me.  Look great!

packages/realm/src/index.ts
kraenhansen: I agree 👍

elle-j: I'll update the constructors as `@internal` as well.

papafe (approved):

packages/realm/src/index.ts
elle-j: Updates:
* [Marked](https://github.com/realm/realm-js/pull/5722/commits/4d13b24540a5919283a65517653702c3ea77bae3) constructors and some fields as `@internal`. Bundle is now:
```ts
declare class AssertionError extends Error {
}
declare class TypeAssertionError extends AssertionError {
}
declare class SchemaParseError extends Error {
}
```

* [Wrapped](https://github.com/realm/realm-js/pull/5722/commits/a859b457bf2066b10c40007cc4098a41af3b0e35) the validation part in a `try-catch` to rethrow errors as `SchemaParseError` so that this error also applies to invalid JS data types. (The validation part threw a mix of Error, TypeError, TypeAssertionError, or AssertionError depending on how assert was called.)

packages/realm/src/schema/validate.ts
elle-j, line 1: The only change in this file:
* Wrapped the `validateObjectSchema` body in a try-catch to rethrow errors as SchemaParseError so that this error also applies to invalid JS data types. (The validation part threw a mix of Error, TypeError, TypeAssertionError, or AssertionError depending on how assert was called.)

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None
 
 ### Enhancements
-* None
+* Opening a realm with invalid schemas will throw a `SchemaParseError` rather than an `AssertionError` or `Error`.
```
kneth, line 7:
```suggestion
* Opening a Realm with invalid schemas will throw a `SchemaParseError` rather than an `AssertionError` or `Error`. ([#5198](https://github.com/realm/realm-js/issues/5198))
```

kneth (approved):

CHANGELOG.md
elle-j: Thanks!

packages/realm/src/schema/normalize.ts
```diff
@@ -387,16 +388,39 @@ function assertNotUsingShorthand(input: string | undefined, info: PropertyInfo):
...
+}
+
 /**
  * Throw an error caused by an invalid property schema.
  */
-function error(info: PropertyInfo, message: string): never {
-  // TODO: Create a SchemaParseError that extends Error?
-  throw new Error(errMessage(info, message));
+function throwError(info: PropertyInfo, message: string): never {
+  throw new SchemaParseError(propSchemaErrorMessage(info, message));
```
kraenhansen, line 423: I find it a bit strange that we have a function which is just responsible for throwing 🤔

packages/realm/src/schema/normalize.ts
```diff
@@ -387,16 +388,39 @@ function assertNotUsingShorthand(input: string | undefined, info: PropertyInfo):
...
+  return `Invalid schema for ${displayName}: ${message}`;
+}
+
+/**
+ * Generate an error caused by an invalid object schema.
+ */
+function objectError(objectName: string, message: string): () => SchemaParseError {
+  return () => new SchemaParseError(objectSchemaErrorMessage(objectName, message));
+}
+
```
kraenhansen, line 418: Did you consider making each of these cases subclasses the `SchemaParseError` instead?

packages/realm/src/schema/validate.ts
```diff
@@ -73,50 +74,58 @@ export function validateRealmSchema(realmSchema: unknown): asserts realmSchema i
...
+        if (!isUsingShorthand) {
+          validatePropertySchema(objectName, propertyName, propertySchema);
+        }
       }
     }
+  } catch (err: any) {
+    // Rethrow as SchemaParseError rather than a mix of Error, TypeError,
+    // TypeAssertionError, or AssertionError.
+    throw new SchemaParseError(err.message);
   }
```
kraenhansen, line 129: I believe this pattern is better as it won't risk getting the `message` off something that wasn't an object.

```suggestion
  } catch (err) {
    if (err instanceof Error) {
      // Rethrow as SchemaParseError rather than a mix of Error, TypeError,
      // TypeAssertionError, or AssertionError.
      throw new SchemaParseError(err.message);
    } else {
      throw err;
    }
  }
```

packages/realm/src/schema/normalize.ts
elle-j: Good idea!

We could have these:

```ts
export class SchemaParseError extends Error {
  /** @internal */
  constructor(message: string) {
    super(message);
  }
}

export class ObjectSchemaParseError extends SchemaParseError {
  /** @internal */
  constructor(objectName: string, message: string) {
    const displayName = objectName ? `object '${objectName}'` : "unnamed object";
    super(`Invalid schema for ${displayName}: ${message}`);
  }
}

export class PropertySchemaParseError extends SchemaParseError {
  /** @internal */
  constructor(objectName: string, propertyName: string, message: string) {
    super(`Invalid type declaration for property '${propertyName}' on '${objectName}': ${message}`);
  }
}
```

Having the specific error class be responsible for the message formatting can be useful here. In that case, I could try to modify the [validation part](https://github.com/realm/realm-js/blob/4fb7b0f3c4d79da4a7e3360e719e8e9b2d01b7bb/packages/realm/src/schema/validate.ts#L128) to also throw the more specific one.

PR #5727, opened 2023-04-13
Added core logging 

### LINKED ISSUE #2546: Support Core logging
## Goals

Things don't always go as expected. Developers should have a way to enable a sort of trace logging with Realm so they can see what it's doing in a production application. Without logging we have to go through a lot of extra work to narrow down performance issues.

## Expected Results

There should be a way to enable logging, either by configuring Realm, or for performance reasons packaging up the logging version as a different NPM dependency.

## Actual Results

The only way to debug what Realm is doing is to guess at what might be causing a problem and remove it. 

## Steps to Reproduce

Have someone that is not familiar with Realm trying to make a repro to demo a bug. It takes a lot longer than it should.

## Code Sample

Not relevant

- Realm JS SDK Version: Any
- Node or React Native: Any
- Client OS & Version: Any

## Note

This issue was originally raised as #2499, I'm splitting it out because multiple separate and unconnected issues were discovered in exploring there.

papafe:
This PR adds support for the new core logging feature. 

This closes #2546

NOTE: The test could need to be revisited after https://github.com/realm/realm-core/pull/6524 is merged

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests
* [x]  Deprecate previous sync logging feature and create ticket to remove it in v13
* [x]  Add default logger callback (similar to https://github.com/realm/realm-js/pull/5698/files)


packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,71 @@
...
+  [binding.LoggerLevel.Warn, "warn"],
+  [binding.LoggerLevel.Error, "error"],
+  [binding.LoggerLevel.Fatal, "fatal"],
+  [binding.LoggerLevel.Off, "off"],
+]);
+
+/** @internal */
+export function fromBindingLoggerLevelToLogLevel(arg: binding.LoggerLevel): LogLevel {
+  return translationTable.get(arg) ?? "all";
+}
```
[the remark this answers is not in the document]
papafe: Thanks for the tip! From what I can understand, it seems that with `Record` I get the advantage that the compiler will complain if I'm not handling all the cases. 
Is that the reason why you suggested it or is there something else?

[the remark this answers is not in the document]
elle-j: That too but also I don't think you'll need the ordering guarantee that Map is good for if you're just using it here. Although, I believe indexing by string will work and not generate a TS error. You can try with `translationTable["0"]`.

packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,91 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { assert, binding } from "./internal";
+
+export type LogLevel = "all" | "trace" | "debug" | "detail" | "info" | "warn" | "error" | "fatal" | "off";
```
elle-j, line 21: Thoughts on providing some helper levels?
```suggestion
export type LogLevel = "all" | "trace" | "debug" | "detail" | "info" | WarnLogLevel | ErrorLogLevel | "off";
export type ErrorLogLevel = "error" | "fatal";
export type WarnLogLevel = "warn";
```

packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,91 @@
...
+  Object.entries(translationTable).map(([key, val]) => [val, Number(key)]),
+) as unknown as Record<LogLevel, binding.LoggerLevel>;
+
+/** @internal */
+export function fromBindingLoggerLevelToLogLevel(arg: binding.LoggerLevel): LogLevel {
+  return translationTable[arg];
+}
+
+const consoleErrorLevels: LogLevel[] = ["error", "fatal"];
+const consoleWarnLevels: LogLevel[] = ["warn"];
```
elle-j, line 74: This suggestion is combined with my previous one if we provide the additional log level helpers.
```suggestion
const consoleErrorLevels: ErrorLogLevel[] = ["error", "fatal"];
const consoleWarnLevels: WarnLogLevel[] = ["warn"];
```

In this case, you'd have to change e.g. `consoleErrorLevels.includes(logLevel)` to `consoleErrorLevels.includes(logLevel as ErrorLogLevel)`, so not sure how beneficial this would be for us.

packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/** @internal */
+export const defaultLogger: LoggerCallback = function (logLevel: LogLevel, message: string) {
```
elle-j, line 77:
```suggestion
export function defaultLogger(logLevel: LogLevel, message: string): LoggerCallback {
```

packages/realm/src/app-services/Sync.ts
```diff
@@ -82,13 +56,16 @@ export class Sync {
   /** @deprecated Please use named imports */
   static NumericLogLevel = NumericLogLevel;
 
+  /** @deprecated Please use Realm.setLogLevel. Will be removed in v13.0.0. */
```
elle-j, line 59: Mainly changed adding a `@link` so that users can navigate to the new type. (Swapped the two sentences just for consistency 👍 )
```suggestion
  /** @deprecated Will be removed in v13.0.0. Please use {@link Realm.setLogLevel}. */
```

packages/realm/src/app-services/Sync.ts
```diff
@@ -82,13 +56,16 @@ export class Sync {
...
   static setLogLevel(app: App, level: LogLevel) {
     const numericLevel = toBindingLoggerLevel(level);
     app.internal.syncManager.setLogLevel(numericLevel);
   }
+
+  /** @deprecated Please use Realm.setLoggerCallback. Will be removed in v13.0.0. */
```
elle-j, line 65:
```suggestion
  /** @deprecated Will be removed in v13.0.0. Please use {@link Realm.setLoggerCallback}. */
```

elle-j (approved):
Great addition! I have some suggestions 👍

integration-tests/tests/src/tests/shared-realms.ts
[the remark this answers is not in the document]
papafe: I am not sure I understand what you mean here. What this is doing is disabling the default logger (I need to add default to that string...), so that we don't get anything printed in the console. 
We have another ticket created that should allow to print to console only in case the test is failing: https://github.com/realm/realm-js/issues/5742

integration-tests/tests/src/tests/shared-realms.ts
```diff
@@ -22,7 +22,57 @@ import { Realm, List } from "realm";
...
+        .and.to.not.contain("trace");
+      logs = [];
+
+      Realm.setLogLevel("warn");
+      realm.write(() => realm.create("Person", { name: "Alice" }));
+      expect(logs).to.be.empty;
+
+      Realm.setLogLevel("error");
+      realm.write(() => realm.create("Person", { name: "Alice" }));
+      expect(logs).to.be.empty;
```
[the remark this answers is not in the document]
papafe: I don't think we need separate tests, otherwise we are probably verifying that the whole logger works, even at the core level. We just need to be sure the callback is called, and get a confirmation that we can change the log level. I think we can assume that core verified that all the levels work. 

Regarding "warn", "error" and so on, I was thinking about it, but it seems that those are actually raised only when using sync (like with a compensating write), and I thought it would have been too much to create a sync test just to verify the logger (also connected to what I've said before). I'm open to change my mind though 😁

packages/realm/src/Logger.ts
papafe: If I make this change than the function would need to return a `LoggerCallback` instead of being itself a `LoggerCallback`

packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,91 @@
...
+  Info = 4,
+  Warn = 5,
+  Error = 6,
+  Fatal = 7,
+  Off = 8,
+}
+
+export type Logger = (level: NumericLogLevel, message: string) => void;
+
+export type LoggerCallback = (level: LogLevel, message: string) => void;
```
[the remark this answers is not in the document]
papafe: I think it would make sense to rename `Realm.setLoggerCallback` to `Realm.setLogger`, so it's more consistent with what we had before. 
Regarding the callback, unfortunately we still use the `Logger` type for the sync logger, so I can't use that type

packages/realm/src/Logger.ts
papafe: This makes sense, but I think it will make the code much more difficult to read. Also even with this we can't be sure we're covering all the error levels (or warn levels) with the arrays.

packages/realm/src/Logger.ts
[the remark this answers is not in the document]
kraenhansen: I think `LoggerCallback` and `Realm.setLogger` makes sense 👍

packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,91 @@
...
+  [binding.LoggerLevel.Info]: "info",
+  [binding.LoggerLevel.Warn]: "warn",
+  [binding.LoggerLevel.Error]: "error",
+  [binding.LoggerLevel.Fatal]: "fatal",
+  [binding.LoggerLevel.Off]: "off",
+};
+
+const inverseTranslationTable: Record<LogLevel, binding.LoggerLevel> = Object.fromEntries(
+  Object.entries(translationTable).map(([key, val]) => [val, Number(key)]),
+) as unknown as Record<LogLevel, binding.LoggerLevel>;
```
kraenhansen, line 66: I'd like to iterate on this with you. I think it'd be possible to use the type parameters on the `fromEntries`, `entries` and `map` to achieve this and avoid `as unknown`.

packages/realm/src/Logger.ts
kraenhansen: Or perhaps `satisfies WarnLogLevel[]`.
```suggestion
const consoleErrorLevels = ["error", "fatal"] satisfies LogLevel[];
const consoleWarnLevels = ["warn"] satisfies LogLevel[];
```

kraenhansen (approved):

packages/realm/src/Logger.ts
elle-j: My bad, I read that one wrong!

packages/realm/src/Logger.ts
papafe: Sure! Let me know when you want to go over it together

packages/realm/src/Logger.ts
papafe: @kraenhansen I wrote this as a comment to another related suggestion, but I'm reluctant to go this route (defining `WarnLogLevel` and `ErrorLogLevel`), because it will make the code more difficult to read, and we won't get much out of it.

integration-tests/tests/src/tests/shared-realms.ts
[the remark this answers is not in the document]
elle-j: Alright, what I meant was that I only saw the other log levels being used in the test except for the ones mentioned above.

packages/realm/src/Logger.ts
papafe: I actually discovered that we don't need the `unknown` 😁 
Before I didn't have the conversion to `number`, and that was the issue

packages/bindgen/src/realm_js_helpers.h
```diff
@@ -194,6 +194,15 @@ struct Helpers {
     using LoggerFactory = std::function<std::shared_ptr<util::Logger>(util::Logger::Level)>;
     using LogCallback = std::function<void(util::Logger::Level, const std::string& message)>;
     static LoggerFactory make_logger_factory(LogCallback&& logger)
+    {
```
papafe, line 197: @RedBeard0531 I followed your suggestion of creating a `make_logger` method that is being used by the older `make_logger_factory` method. 
I just wanted to ask if it makes sense the way that I've done it.

packages/realm/src/Logger.ts
```diff
@@ -0,0 +1,91 @@
...
+export const defaultLogger: LoggerCallback = function (logLevel: LogLevel, message: string) {
+  const formattedLogMessage = `[${logLevel}] ${message}`;
+  /* eslint-disable no-console */
+  if (consoleErrorLevels.includes(logLevel)) {
+    console.error(formattedLogMessage);
+  } else if (consoleWarnLevels.includes(logLevel)) {
+    console.error(formattedLogMessage);
+  } else {
+    console.log(formattedLogMessage);
+  }
```
kraenhansen, line 86: Actually .. since you're not using these arrays elsewhere nor exporting them, I would suggest just inlining a few equality checks instead. These levels are not something we expect to change in the future.

I also suggest calling `console.warn` for the "warn" level.

```suggestion
  if (logLevel === "error" || logLevel === "fatal") {
    console.error(formattedLogMessage);
  } else if (logLevel === "warn") {
    console.warn(formattedLogMessage);
  } else {
    console.log(formattedLogMessage);
  }
```

packages/realm/src/Logger.ts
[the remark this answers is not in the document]
kneth: I agree

kneth (approved):

PR #5734, opened 2023-04-14
Rename `local-mongodb` to enable app replacements

kraenhansen:
## What, How & Why?

This fixes consistent app import failures on CI.

### DIFF
--- a/integration-tests/realm-apps/with-custom-function/secrets.json
+++ b/integration-tests/realm-apps/with-custom-function/secrets.json
@@ -1,3 +1,3 @@
 {
-    "local-mongodb-uri": "mongodb://localhost:26000"
+    "mongodb_uri": "mongodb://localhost:26000"
 }
--- a/integration-tests/realm-apps/with-custom-function/functions/customAuthentication/source.js
+++ b/integration-tests/realm-apps/with-custom-function/functions/customAuthentication/source.js
@@ -21,7 +21,7 @@
 
 exports = async function (loginPayload) {
   // Get a handle for the app.users collection
-  const users = context.services.get("local-mongodb").db("app").collection("users");
+  const users = context.services.get("mongodb").db("app").collection("users");
 
   // Parse out custom data from the FunctionCredential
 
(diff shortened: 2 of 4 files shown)


kraenhansen:
This seem to fix tests for node and RN iOS 👍

integration-tests/realm-apps/with-custom-function/secrets.json
```diff
@@ -1,3 +1,3 @@
 {
-    "local-mongodb-uri": "mongodb://localhost:26000"
+    "mongodb_uri": "mongodb://localhost:26000"
```
elle-j, line 2: Just wondering if the app importer might be relying on `local-mongodb-uri` given these other occurrences:
* [with-jwt/secrets.json](https://github.com/realm/realm-js/blob/9bdd7cc384f3be4c5f4afae40927c692bb7d0a4b/integration-tests/realm-apps/with-jwt/secrets.json#L3)
* [with-db/secrets.json](https://github.com/realm/realm-js/blob/9bdd7cc384f3be4c5f4afae40927c692bb7d0a4b/integration-tests/realm-apps/with-db/secrets.json#L2)
* [with-db-flex/secrets.json](https://github.com/realm/realm-js/blob/9bdd7cc384f3be4c5f4afae40927c692bb7d0a4b/integration-tests/realm-apps/with-db-flx/secrets.json#L2)
* [with-db-flex/services/mongodb/config.json](https://github.com/realm/realm-js/blob/9bdd7cc384f3be4c5f4afae40927c692bb7d0a4b/integration-tests/realm-apps/with-db-flx/services/mongodb/config.json#L11)

elle-j (approved):
Thanks for the fix 👍 Just left a question regarding the rename.

integration-tests/realm-apps/with-custom-function/secrets.json
kraenhansen: Actually .. Thinking more about it, it should probably not have been renamed.

PR #5738, opened 2023-04-17
Improving Electron tests

kraenhansen:
## What, How & Why?

Improves running the Electron tests:
- Producing a cjs bundle of the app importer.
- Upgrading Electron to v24.
- Loading realm in the renderer process still fails.

### DIFF
--- a/integration-tests/environments/electron/app/main.js
+++ b/integration-tests/environments/electron/app/main.js
@@ -42,6 +42,8 @@ app.on("ready", () => {
     mainWindow = new BrowserWindow({
       show: false,
       webPreferences: {
+        nodeIntegration: true,
+        contextIsolation: false,
         enableRemoteModule: true,
         preload,
       },
--- a/integration-tests/environments/electron/app/main.js
+++ b/integration-tests/environments/electron/app/main.js
@@ -19,6 +19,9 @@
 // This file is pretty much a copy of https://github.com/electron/electron-quick-start/blob/master/main.js
 const electron = require("electron");
 const { app, BrowserWindow } = electron;
+// Needed to forward console logs
+const remote = require("@electron/remote/main");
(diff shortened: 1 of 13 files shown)


packages/bindgen/src/templates/node-wrapper.ts
```diff
@@ -52,10 +52,10 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
     };
   `);
   nodeLines.push(`
-    /*global global*/
+    /*global global, require */
     import { createRequire } from 'node:module';
-    const require = createRequire(import.meta.url);
-    const nativeModule = require("./realm.node");
+    const nodeRequire = typeof require === 'function' ? require : createRequire(import.meta.url);
```
elle-j, line 57: Unsure if this is being linted.
```suggestion
    const nodeRequire = typeof require === "function" ? require : createRequire(import.meta.url);
```

elle-j (approved):
🙂

integration-tests/environments/electron/app/main.js
```diff
@@ -19,6 +19,9 @@
 // This file is pretty much a copy of https://github.com/electron/electron-quick-start/blob/master/main.js
 const electron = require("electron");
 const { app, BrowserWindow } = electron;
+// Needed to forward console logs
+const remote = require("@electron/remote/main");
+remote.initialize();
```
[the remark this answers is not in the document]
kraenhansen: Only when executed from the `@electron/remote`, which this is not.

packages/bindgen/src/templates/node-wrapper.ts
kraenhansen: Unfortunately, it's a limitation of the TS in TS code, that it isn't being linted 😞

PR #5740, opened 2023-04-17
Start separating JS-specific from SDK-agnostic parts of bindgen

RedBeard0531:
This is the first step to prepare for moving the SDK-agnostic parts of bindgen to the core repo. It moves the JS-specific bits to `packages/realm/bindgen` and the agnostic parts to `packages/realm/bindgen/vendor/bindgen-lib`. 

The main remaining TODOs to have a clean split are:
* The CMakeLists.txt file needs to be split up with the library providing a cmake function that the SDK's CMakeLists.txt can use to generate a binding from a template file.
* The js-specific header files need to move up (easier once cmake split is done)
* ~~The "low-level" `tests/test.ts` needs to move. I tried doing that, but ran into some `tsx` + modules pain.~~ (decided to delete it instead)

I'd like to get this step merged sooner rather than later, then tackle those issues once this baseline is established.

NOTE TO REVIEWERS: I suggest reviewing the individual commits, since each handles one specific step along with an explanation of what it is doing, rather than just looking at the combined stack of changes.


kraenhansen:
Looks like some GHA related stuff needs an update. Would you like me to take a look at that?

kraenhansen:
Regarding the rename of [realm_js_helpers.h to just realm_helpers.h](https://github.com/realm/realm-js/pull/5740/commits/183a96894780f4d2ee909164b731a85218b19cb3) I'm wondering if this is actually what we want.

I understand that most SDKs adopting the binding generator would probably want / need the helpers we've implemented so far. At the same time I've also seen them as escape hatches that was the domain of our team. I.e. an abstraction on Realm Core that we didn't need to coordinate and could break at our will other SDK teams nor the core team. Perhaps we just need to create another helpers.h for our team (I know we have the two engine specific ones as well).

packages/realm/bindgen/vendor/bindgen-lib/package.json
```diff
@@ -6,8 +6,10 @@
   "bin": {
     "realm-bindgen": "./realm-bindgen.ts"
   },
+  "exports": {
+      "./*": "./src/*"
```
kraenhansen, line 10: I'd prefer if we didn't declare the entire internal implementation of the bindgen package as public / exported to the SDK specific code. Instead I'd suggest we export a single `index.ts` which exports the parts we actually want exposed. This will help us iterate the binding generator codebase without breaking the generators between SDKs.

RedBeard0531: I'm fine with that as a follow-up, but I don't want to do that in this commit. I'm not 100% sure what we want to be public, but I suspect it is close to 90% of the code, so it may not be worth making a fine-grained distinction here.

kraenhansen: > not 100% sure what we want to be public

We don't have to be, for now we should just export whatever the JS generators need. Now, you're effectively declaring everything public 🤷

packages/realm/package.json
```diff
@@ -185,16 +187,18 @@
       "command": "node ../../scripts/build-android.js",
       "files": [
         "../../src/android/**",
-        "../bindgen/src/**/*.ts",
-        "../bindgen/src/**/*.h"
+        "bindgen/src/**/*.ts",
+        "bindgen/vendor/bindgen-lib/src/**/*.ts",
+        "bindgen/vendor/bindgen-lib/src/**/*.h"
```
elle-j, line 192: Will there not be any `.h` files under `"bindgen/src/**/"`?

elle-j (approved):
Looks good 🙂

packages/realm/package.json
RedBeard0531: There will be, but I haven't moved them yet. I'll do that soon.

packages/realm/bindgen/vendor/bindgen-lib/package.json
RedBeard0531: If 90% of the code is public (my rough guess), I think the benefit from making the remaining 10% private is fairly low. The main benefit IMO is that it would let you have a single import in the consumer rather than several.

RedBeard0531:
> Regarding the rename of [realm_js_helpers.h to just realm_helpers.h](https://github.com/realm/realm-js/pull/5740/commits/183a96894780f4d2ee909164b731a85218b19cb3) I'm wondering if this is actually what we want.
> 
> I understand that most SDKs adopting the binding generator would probably want / need the helpers we've implemented so far. At the same time I've also seen them as escape hatches that was the domain of our team. I.e. an abstraction on Realm Core that we didn't need to coordinate and could break at our will other SDK teams nor the core team. Perhaps we just need to create another helpers.h for our team (I know we have the two engine specific ones as well).

(unfortunately GH doesn't support properly replying to comments not on code 😡)

We can always add our own sdk-specific helpers as needed. That became supported once I added support for multiple spec files. But _this_ file seems like it contains things that will be helpful to _all_ SDKs (note that it is described in the main spec.yml), so I would want it to be shared.

packages/realm/bindgen/vendor/bindgen-lib/CMakeLists.txt
```diff
@@ -148,7 +150,7 @@ set(GENERATED_JS_FILES
     ${TYPESCRIPT_OUTPUT_DIR}/native-rn.mjs
 )
 
-set(NPX ${NODE_FORCE_COLOR} npx)
+set(NPX ${NODE_FORCE_COLOR} npx --prefix .) # without a prefix, npx changes the current directory and that is really confusing and hard to debug
```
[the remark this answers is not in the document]
RedBeard0531: Meh. Maybe, but it is describing the behavior that we are opting-out of, so it probably isn't that useful to know about, since its hard to imagine someone wanting to go back to it.

RedBeard0531:
@kraenhansen could you take another look? The first 5 commits are the same as before, but rebased on main. The last 3 are new since you reviewed.

There are a handful of test failures, but I don't _think_ they could be related to what I'm doing here, since if these changes were broken, I'd expect build or load failures, rather than a few specific test failures.

packages/realm/bindgen/.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
+  "root": true,
```
kraenhansen, line 2: While I understand and appreciate you want this to be an example for other teams, I think having this a root config is not really in the best interest of this team. We'd want a change to the eslint config in the repo root to have an effect the code in this package.

packages/realm/bindgen/vendor/bindgen-lib/.eslintrc.json
```diff
@@ -1,7 +1,10 @@
 {
+  "root": true,
   "parser": "@typescript-eslint/parser",
   "extends": [
-    "plugin:@typescript-eslint/recommended"
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:prettier/recommended"
```
kraenhansen, line 7: These should probably be added as dependencies to the package as well, so it doesn't rely on a parent package providing them.

kraenhansen:
You might also want to delete the `test` script from the `package.json` of the bindgen-lib and turn the "generate json schema" into a regular (non-wireit) script 🤔

RedBeard0531:
> You might also want to delete the `test` script from the `package.json` of the bindgen-lib and turn the "generate json schema" into a regular (non-wireit) script 🤔

Does anything actually use the scripts and wireit tasks from bindgen-lib? If not, I'd rather just remove them since core will invoke it from cmake rather than nodejs.

packages/realm/bindgen/vendor/bindgen-lib/.eslintrc.json
RedBeard0531: I'm going to do a pass on that package.json to make sure it really declares everything it needs as part of moving it to the core repo. I'll defer this until then, since I'm sure ill miss some stuff.

PR #5761, opened 2023-04-24
React Native 0.71.4 → 0.71.7

kraenhansen:
## What, How & Why?

Upgrading React Native to 0.71.7 using https://react-native-community.github.io/upgrade-helper/?from=0.71.4&to=0.71.7 in the root package and the integration test app. I considered doing the upgrade in the example app, but since this is an instance of our templates, I'd rather allow for a follow-up template upgrading those and the example app.


kneth (approved):

package.json
```diff
@@ -83,7 +83,7 @@
     "eslint-plugin-react": "^7.31.10",
     "eslint-plugin-react-native": "^4.0.0",
     "node-addon-api": "^6.0.0",
-    "react-native": "0.71.4",
+    "react-native": "0.71.7",
```
elle-j, line 86: We have some more occurrences of `0.71.4` and `0.71.5` in various package.json. Should we update those too perhaps?

elle-j (approved):

package.json
kraenhansen: As the description mentions:
>I considered doing the upgrade in the example app, but since this is an instance of our templates, I'd rather allow for a follow-up template upgrading those and the example app.

Are there other occurrences outside of the examples and our templates you're thinking of?

elle-j: Our `realm` and `realm-react` packages.

kraenhansen: God catch 🙂 I'll push that directly to main 🙈

PR #5766, opened 2023-04-25
Fixing a missing simulator type on CI

kraenhansen:
## What, How & Why?

This is a follow up to https://github.com/realm/realm-js/pull/5761 - which introduced an error related to a missing simulator device type on CI.


kraenhansen:
Merging as it passed on CI (and I introduced the issue in the first place anyway).

PR #5769, opened 2023-04-26
switch to consuming bindgen from realm-core repo

RedBeard0531:
Now we are tracking https://github.com/realm/realm-core/tree/bindgen which has what used to be `packages/realm/bindgen/vendor/bindgen-lib` at `bindgen`. There is still some separating to do, such as moving the wireit tasks out of realm-core's bindgen package.json, and making that package.json freestanding, but I want to lock in this progress first.

(Adding Yavor and Nikola mostly as an FYI)


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "docs/jsdoc-template"]
 	path = docs/jsdoc-template
 	url = https://github.com/realm/realm-jsdoc.git
-[submodule "vendor/realm-core"]
-	path = vendor/realm-core
+[submodule "packages/realm/bindgen/vendor/realm-core"]
```
RedBeard0531, line 4: `git mv` didn't change this line and I wasn't 100% sure if I should, but I decided to do it manually. Let me know if I shouldn't have.

packages/realm/bindgen/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ if(DEFINED CMAKE_JS_VERSION)
     include(NodeJSTargets)
 endif()
 
+# TODO is this option dead? Do we want to support compiling with it set to OFF?
```
fealebenpae, line 81: We never ended up implementing building from Core prebuilds so we might as well kill it.

RedBeard0531:
@kraenhansen  How do I tell GHA to not lint every js file in realm-core? https://github.com/realm/realm-js/actions/runs/4807896465/jobs/8557223717?pr=5769#step:5:26. I think we either want it to lint nothing, or only lint the bindgen dir.

kraenhansen:
> How do I tell GHA to not lint every js file in realm-core?

@RedBeard0531 I've pushed a commit adding a .lintignore file to the `realm/bindgen` directory 👍

RedBeard0531:
I pulled this back to a draft. I'm going to move the package.json up to the root of realm-core, and that will have some effects on how we consume it from realm-js.

RedBeard0531:
Out of draft now that https://github.com/realm/realm-core/pull/6546 has been merged into the realm-core:bindgen branch. There is still a weird linting issue that @kraenhansen said he would look into, but I think this is generally ready to review.

packages/realm/bindgen/CMakeLists.txt
kraenhansen: I'd like to keep that option open as a potential optimization when we'd eventually start exploring moving back to deferring compilation to the developer's machine, instead of shipping a prebuilt binary for RN (to remove our strong dependency on the jsi.h)

.gitmodules
kraenhansen: It worked wonderfully when I pulled and did a `git submodule update --init --recursive` locally.

packages/realm/bindgen/package.json
```diff
@@ -3,6 +3,6 @@
   "private": true,
   "type": "module",
   "workspaces": [
-    "packages/realm/bindgen/vendor/realm-core/bindgen/"
+    "packages/realm/bindgen/vendor/realm-core/"
```
kraenhansen, line 6: Interesting. Any particular reason this needed to be moved to the root?

kraenhansen (approved):

packages/realm/bindgen/CMakeLists.txt
RedBeard0531: I think this setup is wrong either way. This (used to) let you import a binary realm-core lib, but then still build the realm-js C++ code. I think we need to defer building both, rather than just realm-core. So I propose removing this option for now*, to make room for reintroducing deferred building correctly later

*Not _right now_ though. I'd rather land this PR, and have separate work to clean up the CML.txt file now that we no longer need to be concerned with merging changes with the non-bindgen version.

packages/realm/bindgen/package.json
RedBeard0531: Without that `npm run foo` and `npx bar` would only work when run from a subdirectory of `bindgen`. I thought it would be best if they worked when run from anywhere in the realm-core repo. https://github.com/realm/realm-core/commit/3bad9efef965406781de97daf4354b4181404627 _should_ solve this for commands run by cmake in the `$repo_root/build` directory, but it seemed better to just make those commands work from anywhere in the repo.

In theory, I could have made a root package.json, with bindgen as a subworkspace, but that seems silly for realm-core which isn't really a JS project.

RedBeard0531: `npx bar` in particular would have the odd behavior of "working", but using the latest version of `bar` and its deps, rather than considering the versions from `package.json`

packages/realm/bindgen/CMakeLists.txt
kraenhansen: I'm fine with that 👍 Prebuilding Core with a particular NDK does introduce a dependency on the stdlib ABI as we've noticed before, so I'm fine with doing this correctly later.

packages/realm/bindgen/package.json
kraenhansen: Yeah. When not invoked through an NPM script in the package.json, `npx` use a "globally" installed package.

RedBeard0531:
I was going to delete the whole `/vendor` directory, but then I noticed the [`/vendor/.npmignore` file](https://github.com/realm/realm-js/blob/ms/bindgen-from-core/vendor/.npmignore). Does anyone know if we need something like this in `/packages/realm/bindgen/vendor`?

kraenhansen:
> Does anyone know if we need something like this in /packages/realm/bindgen/vendor?

Running a `pack` in `packages/realm` (`npm pack --dry-run`) yields:

```
npm notice === Tarball Contents === 
npm notice 3.4kB  RealmJS.podspec                                                         
npm notice 8.5kB  package.json                                                            
npm notice 977B   react-native.config.js                                                  
npm notice 1.8kB  react-native/android/build.gradle                                       
npm notice 52.3kB react-native/android/gradle/wrapper/gradle-wrapper.jar                  
npm notice 230B   react-native/android/gradle/wrapper/gradle-wrapper.properties           
npm notice 5.1kB  react-native/android/gradlew                                            
npm notice 2.4kB  react-native/android/gradlew.bat                                        
npm notice 38B    react-native/android/settings.gradle                                    
npm notice 127B   react-native/android/src/main/AndroidManifest.xml                       
npm notice 4.6kB  react-native/android/src/main/java/io/realm/react/RealmReactModule.java 
npm notice 2.4kB  react-native/android/src/main/java/io/realm/react/RealmReactPackage.java
npm notice 6.6kB  react-native/android/src/main/java/io/realm/react/util/SSLHelper.java   
npm notice 109B   react-native/android/src/main/java/io/realm/react/Version.java          
npm notice 8.8MB  react-native/android/src/main/jniLibs/arm64-v8a/librealm.so             
npm notice 5.6MB  react-native/android/src/main/jniLibs/armeabi-v7a/librealm.so           
npm notice 9.8MB  react-native/android/src/main/jniLibs/x86_64/librealm.so                
npm notice 9.4MB  react-native/android/src/main/jniLibs/x86/librealm.so                   
npm notice 886B   react-native/ios/RealmReact/RealmReact.h                                
npm notice 5.2kB  react-native/ios/RealmReact/RealmReact.mm                               
npm notice 150B   react-native/ios/RealmReact/RealmReact.xcconfig                         
npm notice 829B   scripts/submit-analytics.mjs
```

Which doesn't include anything from the `realm/bindgen` directory, because it isn't listed in `files`. I believe this is fine for now, but we probably have to add it back into the `files` (as well as an `.npmignore` file) once we start supporting compiling from source again.

PR #5770, opened 2023-04-26
Fix mongodb client tests to avoid crosstalks

kraenhansen:
## What, How & Why?

This refactors the MongoDB Client tests to rely on a generated database name stored on the test context rather than a const name that would result in crosstalking between tests being ran in parallel on CI.
Note: I've sat the base-branch to `kh/reuse-imported-apps` since that is a dependency of this fix, but I expect that to merge before this PR.

### DIFF
--- a/integration-tests/tests/src/tests/sync/mongo-db-client.ts
+++ b/integration-tests/tests/src/tests/sync/mongo-db-client.ts
@@ -555,14 +555,8 @@ describe.skipIf(environment.missingServer, "MongoDB Client", function () {
         expect(result).to.deep.equal([{ count: 2 }]);
 
         // Note:
-        // If getting `Error: exec: "assisted_agg": executable file not found in $PATH`:
-        //  1) Download the file `assisted_agg` (if on Mac) or `libmongo.so` (if on Linux) and add it
-        //     to your PATH (see https://github.com/10gen/baas/blob/master/etc/docs/onboarding.md).
-        //  2) Load the PATH variable to the terminal window used for starting the BaaS server.
-        //  3) Allow the file to be executable (run: chmod +x your/path/to/assisted_agg).
-        //  4) When running the test again, Mac will block execution of the file. Then (for Mac) go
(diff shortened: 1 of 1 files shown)


integration-tests/tests/src/tests/sync/mongo-db-client.ts
```diff
@@ -29,64 +29,50 @@ type TestDocument = {
...
+type CollectionContext = { collection: MongoDBCollection<TestDocument> };
+type TestContext = CollectionContext & AppContext & UserContext & Mocha.Context;
+
+const serviceName = "mongodb";
+const collectionName = "test-collection";
+
 describe.skipIf(environment.missingServer, "MongoDB Client", function () {
   this.timeout(60_000); // TODO: Temporarily hardcoded until envs are set up.
   importAppBefore(buildAppConfig("with-flx").anonAuth().flexibleSync());
+  authenticateUserBefore();
```
elle-j, line 41: If we add tests after the very last `it` test, it will probably be problematic to use this `authenticateUserBefore()` since it only logs in 1 time (`before`). I previously logged in the user `beforeEach` test because we're also logging out the user in the last test. So for future tests, the user will not be logged in if using `authenticateUserBefore()`. Since the previous login helper just returns `app.currentUser` if it's already logged in, I don't think there should be that much overhead having it.

integration-tests/tests/src/tests/sync/mongo-db-client.ts
elle-j, line 1: In the `describe("#aggregate", ...)` block, there's an old comment about the `assisted_agg` file. You can replace those comments to something like:
```
// Note:
// If getting an error on Mac blocking execution of a file called `assisted_agg`, go
// to `System Settings > Privacy & Security`, find blocked files, then allow `assisted_agg`.
```

elle-j (approved):
Thanks! Consider the comment about `authenticateUserBefore()` 👍

PR #5772, opened 2023-04-27
Add Flexible Sync subscribe/unsubscribe APIs

### LINKED ISSUE #5760: Flex Sync Subscribe API

elle-j:
### Additions:
* Experimental APIs: Adds the possibility to subscribe and unsubscribe directly to and from a `Results` instance via `Results.subscribe()` and the `Results.unsubscribe()`.
  * A `SubscriptionOptions` can be passed to `Results.subscribe()` which has been extended to include:
    * A `WaitForSync` behavior specifying how to wait or not wait for subscribed objects to be downloaded before the returned promise resolves.
    * A maximum waiting timeout before resolving.
* Adds the instance method `MutableSubscriptionSet.removeUnnamed()` allowing for the removal of unnamed subscriptions only.

### Example `subscribe()`:
```js
const peopleOver20 = await realm
  .objects("Person")
  .filtered("age > 20")
  .subscribe({
    name: "peopleOver20",
    behavior: WaitForSync.FirstTime, // Default
    timeout: 2000,
  });
// ...
peopleOver20.unsubscribe();
```

### Example `removeUnnamed()`:
```js
await realm.subscriptions.update((mutableSubs) => {
  mutableSubs.removeUnnamed();
});
```

This closes #5760.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,21 @@
 * Improve performance of equality queries on a non-indexed mixed property by about 30%. ([realm/realm-core#6506](https://github.com/realm/realm-core/pull/6506))
 * Improve performance of rolling back write transactions after making changes.  ([realm/realm-core#6513](https://github.com/realm/realm-core/pull/6513))
 * Extended `PropertySchema.indexed` with the `full-text` option, that allows to create an index for full-text search queries.  ([#5755](https://github.com/realm/realm-js/issues/5755))
+* Added APIs to facilitate adding and removing subscriptions ([#5772](https://github.com/realm/realm-js/pull/5772)):
+  * Enabled subscribing and unsubscribing directly to and from a `Results` instance via `Results.subscribe()` (asynchronous) and `Results.unsubscribe()` (synchronous).
+  * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before returning from `Results.subscribe()`.
+  * Extended `SubscriptionOptions` to take a `WaitForSync` behavior and a maximum waiting timeout before returning from `Results.subscribe()`.
+  * Added an optional `unnamedOnly` parameter to `MutableSubscriptionSet.removeAll()` allowing the removal of only unnamed subscriptions.
+  ```javascript
+  const peopleOver20 = await realm
+    .objects("Person")
+    .filtered("age > 20")
+    .subscribe({
+      name: "peopleOver20",
+      behavior: WaitForSync.FirstTime, // Default
+      timeout: 2000,
+    });
+  ```
 
 ### Fixed
 * Fix a stack overflow crash when using the query parser with long chains of AND/OR conditions. ([realm/realm-core#6428](https://github.com/realm/realm-core/pull/6428), since v10.11.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@
 * Extended `PropertySchema.indexed` with the `full-text` option, that allows to create an index for full-text search queries.  ([#5755](https://github.com/realm/realm-js/issues/5755))
 * Added APIs to facilitate adding and removing subscriptions. ([#5772](https://github.com/realm/realm-js/pull/5772))
   * Experimental APIs: Enabled subscribing and unsubscribing directly to and from a `Results` instance via `Results.subscribe()` (asynchronous) and `Results.unsubscribe()` (synchronous).
-  * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before returning from `Results.subscribe()`.
-  * Extended `SubscriptionOptions` to take a `WaitForSync` behavior and a maximum waiting timeout before returning from `Results.subscribe()`.
+    * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before resolving the promise returned from `Results.subscribe()`.
+    * Extended `SubscriptionOptions` to take a `WaitForSync` behavior and a maximum waiting timeout before returning from `Results.subscribe()`.
   * Added the instance method `MutableSubscriptionSet.removeUnnamed()` for removing only unnamed subscriptions.
   ```javascript
   const peopleOver20 = await realm
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 * Improve performance of equality queries on a non-indexed mixed property by about 30%. ([realm/realm-core#6506](https://github.com/realm/realm-core/pull/6506))
 * Improve performance of rolling back write transactions after making changes.  ([realm/realm-core#6513](https://github.com/realm/realm-core/pull/6513))
 * Extended `PropertySchema.indexed` with the `full-text` option, that allows to create an index for full-text search queries.  ([#5755](https://github.com/realm/realm-js/issues/5755))
-* Added APIs to facilitate adding and removing subscriptions ([#5772](https://github.com/realm/realm-js/pull/5772)):
+* Added APIs to facilitate adding and removing subscriptions. ([#5772](https://github.com/realm/realm-js/pull/5772))
   * Enabled subscribing and unsubscribing directly to and from a `Results` instance via `Results.subscribe()` (asynchronous) and the experimental `Results.unsubscribe()` (synchronous).
   * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before returning from `Results.subscribe()`.
   * Extended `SubscriptionOptions` to take a `WaitForSync` behavior and a maximum waiting timeout before returning from `Results.subscribe()`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,10 @@
 * Improve performance of rolling back write transactions after making changes.  ([realm/realm-core#6513](https://github.com/realm/realm-core/pull/6513))
 * Extended `PropertySchema.indexed` with the `full-text` option, that allows to create an index for full-text search queries.  ([#5755](https://github.com/realm/realm-js/issues/5755))
 * Added APIs to facilitate adding and removing subscriptions ([#5772](https://github.com/realm/realm-js/pull/5772)):
-  * Enabled subscribing and unsubscribing directly to and from a `Results` instance via `Results.subscribe()` (asynchronous) and `Results.unsubscribe()` (synchronous).
+  * Enabled subscribing and unsubscribing directly to and from a `Results` instance via `Results.subscribe()` (asynchronous) and the experimental `Results.unsubscribe()` (synchronous).
   * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before returning from `Results.subscribe()`.
   * Extended `SubscriptionOptions` to take a `WaitForSync` behavior and a maximum waiting timeout before returning from `Results.subscribe()`.
-  * Added an optional `unnamedOnly` parameter to `MutableSubscriptionSet.removeAll()` allowing the removal of only unnamed subscriptions.
+  * Added the instance method `MutableSubscriptionSet.removeUnnamed()` for removing only unnamed subscriptions.
   ```javascript
   const peopleOver20 = await realm
     .objects("Person")
@@ -29,6 +29,8 @@
       behavior: WaitForSync.FirstTime, // Default
       timeout: 2000,
     });
+  // ...
+  peopleOver20.unsubscribe();
   ```
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


packages/realm/src/Results.ts
```diff
@@ -106,6 +109,48 @@ export class Results<T = unknown> extends OrderedCollection<T> {
     }
   }
 
+  /**
+   * Add the query (represented by this {@link Results} instance) to the set of active subscriptions.
+   * The query will be joined via an `OR` operator with any existing queries for the same type.
```
elle-j, line 114: This doc is similar to `MutableSubscriptionSet.add()`.

packages/realm/src/app-services/MutableSubscriptionSet.ts
```diff
@@ -141,17 +169,38 @@ export class MutableSubscriptionSet extends BaseSubscriptionSet {
...
+
+  /**
+   * Remove all subscriptions from the SubscriptionSet.
+   *
+   * @param unnamedOnly Whether to only remove unnamed/anonymous subscriptions. (Default: `false`)
+   * @returns The number of subscriptions removed.
+   */
+  removeAll(unnamedOnly = false): number {
+    if (unnamedOnly) {
+      return this.removeByPredicate((subscription) => !subscription.name);
```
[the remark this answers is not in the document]
kneth: I think it is worth discussing with a broader audience if the empty string should be banned as name.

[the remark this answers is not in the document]
kraenhansen: I'd say this is more correct:

```suggestion
      return this.removeByPredicate((subscription) => typeof subscription.name === "undefined");
```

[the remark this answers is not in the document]
elle-j: Yeah we could definitely check for `undefined`, but was thinking more if we should treat empty-string names like unnamed or not.

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@
 * Opening a Realm with invalid schemas will throw a `SchemaParseError` (or one of its subtypes `ObjectSchemaParseError` and `PropertySchemaParseError`) rather than an `AssertionError` or `Error`. ([#5198](https://github.com/realm/realm-js/issues/5198))
 * Enable multiple processes to operate on an encrypted Realm simultaneously. ([realm/realm-core#1845](https://github.com/realm/realm-core/issues/1845))
 * Added support for a sync configuration option to provide an `SSLConfiguration` with a custom function for validating the server's SSL certificate. [#5485](https://github.com/realm/realm-js/issues/5485)
+* Added APIs to facilitate adding and removing subscriptions [#5772](https://github.com/realm/realm-js/pull/5772):
```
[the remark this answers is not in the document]
kneth: I don't have a strong preference. Since the PR has more info, it is fewer clicks for our users.

kneth, line 10:
```suggestion
* Added APIs to facilitate adding and removing subscriptions. ([#5772](https://github.com/realm/realm-js/pull/5772)
```

packages/realm/src/app-services/MutableSubscriptionSet.ts
```diff
@@ -36,7 +56,18 @@ export interface SubscriptionOptions {
    * Adding a subscription with the same name and query is always a no-op.
    */
   throwOnUpdate?: boolean;
-}
+  /**
+   * Specifies how to wait or not wait for subscribed objects to be downloaded.
+   */
+  behavior?: WaitForSync;
+  /**
+   * The maximum number of milliseconds to wait for objects to be downloaded.
```
kneth, line 64:
```suggestion
   * The time (in milliseconds) to wait for objects to be downloaded.
```

kneth (approved):

takameyer (approved):
LGTM!  Just have a few pondering questions/considerations

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1570,6 +1592,139 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+          expect(this.realm.subscriptions.state).to.equal(SubscriptionsState.Complete);
+
+          // Subscribing the second time should return without waiting.
+          await peopleOver10.subscribe({ behavior: WaitForSync.FirstTime });
+          expect(this.realm.subscriptions.state).to.equal(SubscriptionsState.Pending);
+
+          expect(this.realm.subscriptions).to.have.length(1);
+        });
+
+        it("waits for objects to sync the first time only by default", async function (this: RealmContext) {
```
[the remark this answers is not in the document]
cmelchior: It forces the API to be always online, which kinda goes against Realm being an offline-first database.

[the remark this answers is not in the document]
takameyer: Ah ok, that's a great point!  Hopefully we'll have this documented to help our users not fall into this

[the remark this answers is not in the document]
elle-j: Thanks for chiming in @cmelchior 🙂 Sidenote, when opening a synced realm, our default behavior for new and existing realms is `"downloadBeforeOpen"`. Perhaps `"openImmediately"` would also be more inline with offline-first?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1570,6 +1592,139 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+        });
+
+        it("does not unsubscribe multiple times", async function (this: RealmContext) {
+          const peopleOver10 = await this.realm.objects(FlexiblePersonSchema.name).filtered("age > 10").subscribe();
+          expect(this.realm.subscriptions).to.have.length(1);
+
+          let unsubscribed = peopleOver10.unsubscribe();
+          expect(unsubscribed).to.be.true;
+
+          unsubscribed = peopleOver10.unsubscribe();
```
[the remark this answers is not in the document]
elle-j: I agree that it may not be as clear a use case for what a user would do with the return value. But just to add some specificity to what you said, it returns whether it was unsubscribed which could return `false`, e.g. if the query was not subscribed to beforehand.

The Kotlin SDK is facing some issues implementing this and there are discussions if this particular API should just be removed for now. Users would then use the currently available API via `MutableSubscriptionSet`.

packages/realm/src/app-services/MutableSubscriptionSet.ts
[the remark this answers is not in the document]
kraenhansen: > treat empty-string names like unnamed.

Does core support an empty string as name? If yes, we should allow that too.

integration-tests/tests/src/tests/sync/flexible.ts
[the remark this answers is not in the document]
elle-j: Update: It now returns `void`.

packages/realm/src/app-services/MutableSubscriptionSet.ts
kneth: I suggest that it should be `The maximum time (in milliseconds) ...`.

packages/realm/src/Results.ts
```diff
@@ -1,6 +1,6 @@
 ////////////////////////////////////////////////////////////////////////////
 //
-// Copyright 2022 Realm Inc.
+// Copyright 2023 Realm Inc.
```
kneth, line 3: We typically don't update the year in order to establish [prior art](https://en.wikipedia.org/wiki/Prior_art)

packages/realm/src/Results.ts
```diff
@@ -104,6 +114,47 @@ export class Results<T = unknown> extends OrderedCollection<T> {
     }
   }
 
+  /**
+   * Add this query result to the set of active subscriptions. The query will be joined
+   * via an `OR` operator with any existing queries for the same type.
+   *
+   * @param options Options to use when adding this subscription (e.g. a name or wait behavior).
+   * @returns A promise that resolves to this {@link Results} instance.
```
kneth, line 122: We should add `@experimental` to the API documentation

kneth (approved):
It still looks good to me. I have only minor comments.

packages/realm/src/Results.ts
elle-j: Thanks! Forgot to add it to `subscribe()` as well

packages/realm/src/Results.ts
elle-j: Very good to know, thanks for pointing it out 🙂

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1571,6 +1624,198 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+          await peopleOver10.subscribe();
+          expect(this.realm.subscriptions.state).to.equal(SubscriptionSetState.Complete);
+
+          await peopleOver10.subscribe();
+          expect(this.realm.subscriptions.state).to.equal(SubscriptionSetState.Pending);
+
+          expect(this.realm.subscriptions).to.have.length(1);
+        });
+
+        it("waits for objects to sync the first time only for different 'Results' instances", async function (this: RealmContext) {
```
papafe, line 1658: Maybe the test title is a little confusing for me, but this means that `WaitForSync` applies to the query and not to the results, am I correct?

packages/realm/src/Results.ts
```diff
@@ -104,6 +114,48 @@ export class Results<T = unknown> extends OrderedCollection<T> {
...
+    }
+
+    return this;
+  }
+
+  /**
+   * Unsubscribe from this query result.
+   * @experimental This API is experimental and may change or be removed.
+   */
+  unsubscribe(): void {
```
papafe, line 149: Should we specify that this is non-blocking?

papafe (approved):
Looks good! And nice test coverage

packages/realm/src/Results.ts
elle-j: I think the non-blocking part is already documented via the `void` (rather than `Promise<void>`) return type. But it wouldn't hurt to add to the TS docs that it will not wait for the update to complete, I'll add that 🙂

CHANGELOG.md
```diff
@@ -14,6 +14,23 @@
 * Added support for a sync configuration option to provide an `SSLConfiguration` with a custom function for validating the server's SSL certificate. ([#5485](https://github.com/realm/realm-js/issues/5485))
 * Improve performance of equality queries on a non-indexed mixed property by about 30%. ([realm/realm-core#6506](https://github.com/realm/realm-core/pull/6506))
 * Improve performance of rolling back write transactions after making changes.  ([realm/realm-core#6513](https://github.com/realm/realm-core/pull/6513))
+* Added APIs to facilitate adding and removing subscriptions. ([#5772](https://github.com/realm/realm-js/pull/5772))
+  * Experimental APIs: Enabled subscribing and unsubscribing directly to and from a `Results` instance via `Results.subscribe()` (asynchronous) and `Results.unsubscribe()` (synchronous).
+  * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before returning from `Results.subscribe()`.
```
kraenhansen, line 19:
```suggestion
  * Added a `WaitForSync` enum specifying whether to wait or not wait for subscribed objects to be downloaded before resolving the promise returned from `Results.subscribe()`.
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -1397,6 +1399,19 @@ describe.skipIf(environment.missingServer, "Flexible sync", function () {
...
+            addSubscription(this.realm, this.realm.objects(FlexiblePersonSchema.name).filtered("age > 10"));
+            const { subs } = addSubscriptionForPerson(this.realm, { name: "" });
+            expect(subs).to.have.length(2);
+
+            await subs.update((mutableSubs) => {
+              expect(mutableSubs.removeByName("")).to.be.true;
+            });
+
+            expect(subs).to.have.length(1);
+            expect([...subs][0].queryString).to.equal("age > 10");
```
kraenhansen, line 1413: This would be equivalent and perhaps a bit simpler?

```suggestion
            const [ sub ] = subs;
            expect(sub.queryString).to.equal("age > 10");
```

kraenhansen (approved):

integration-tests/tests/src/tests/sync/flexible.ts
elle-j: This is due to how we determine if we should wait for sync or not. If, for instance, the subscription is not named, we check if it already exists *by query* like you said.

I'll modify the test description and can also add some comments in the test 👍

integration-tests/tests/src/tests/sync/flexible.ts
elle-j: That can be a small but good refactor we could add in a separate PR (perhaps for our test refactor ticket), because the previous pattern is also used in other places so would be good to change all of them together.

packages/realm/src/Results.ts
elle-j: I've added some clarification to the TS docs ✅

PR #5783, opened 2023-05-01
Test against a server running via docker exposed via Ngrok

kraenhansen:
## What, How & Why?

This PR make the "PR workflow" use a BaaS running via Docker spun up via the [test-server.yml](https://github.com/realm/realm-js/blob/main/.github/workflows/test-server.yml) workflow, when running tests.

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -240,7 +240,6 @@ jobs:
       REALM_DISABLE_ANALYTICS: 1
       MOCHA_REMOTE_TIMEOUT: 60000
       LONG_TIMEOUT: 300000 # 5 minutes
-      REALM_BASE_URL: realm-js-test-server-${{ github.run_id }}-${{ github.run_attempt }}.ngrok.io
       MOCHA_REMOTE_REPORTER: mocha-github-actions-reporter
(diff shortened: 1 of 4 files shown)


elle-j (approved):
🚀

.github/workflows/pr-realm-js.yml
```diff
@@ -279,6 +275,19 @@ jobs:
           node-version: 16
           cache: npm
 
+      - name: Generate server configuration
+        id: baas-config
+        run:
+          suffix=$(node -p 'Math.floor(Math.random()*Number.MAX_SAFE_INTEGER)');
+          subdomain="realm-js-test-server-${{ github.run_id }}-${{ github.run_attempt }}-${suffix}";
+          echo "subdomain=${subdomain}" >> $GITHUB_OUTPUT;
```
takameyer, line 283: does `${subdomain}` resolve correctly?  Shouldn't it be `${{subdomain}}`?  The output of the job shows `${subdomain}` instead of the value of "subdomain".  Maybe it's working, but it doesn't appear to.

takameyer (approved):
It's amazing this works!  Wish we would have done it this way in the first place.  💯

kneth (approved):

PR #5786, opened 2023-05-02
[v11] Test against a server running via docker exposed via Ngrok

kraenhansen:
## What, How & Why?

This is a backport of https://github.com/realm/realm-js/pull/5783 to the `v11` branch.


elle-j (approved):

kneth:
I don't think it is worth doing.

kraenhansen:
@kneth if this is because we don't care too much about tests passing for v11, perhaps we should also delete the [/deploy-cluster workflow](https://github.com/realm/realm-js/blob/main/.github/workflows/deploy-cluster.yml), to free up some costs.

PR #5790, opened 2023-05-04
Fix currentUser

takameyer:
## What, How & Why?
The `currentUser` does not persist on app creation.  This PR will address this issue.

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/tests/sync/user.ts
+++ b/integration-tests/tests/src/tests/sync/user.ts
@@ -23,7 +23,7 @@ import {
   randomVerifiableEmail,
 } from "../../utils/generators";
 import { KJUR } from "jsrsasign";
-import { UserState } from "realm";
+import { Realm, UserState } from "realm";
 
 import { buildAppConfig } from "../../utils/build-app-config";
 
@@ -491,18 +491,18 @@ describe.skipIf(environment.missingServer, "User", () => {
           exports = async function (loginPayload) {
             // Get a handle for the app.users collection
             const users = context.services.get("mongodb").db("app").collection("users");
-          
+
             // Parse out custom data from the FunctionCredential
-          
+
             const { username, secret } = loginPayload;
-          
+
             if (secret !== "v3ry-s3cret") {
               throw new Error("Ah ah ah, you didn't say the magic word");
             }
             // Query for an existing user document with the specified username
-          
+
             const user = await users.findOne({ username });
-          
+
             if (user) {
               // If the user document exists, return its unique ID
               return user._id.toString();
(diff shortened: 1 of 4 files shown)


integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -574,4 +574,20 @@ describe.skipIf(environment.missingServer, "User", () => {
       await expect(user.functions.error()).to.be.rejectedWith("function not found: 'error'");
     });
   });
+  describe("currentUser", () => {
+    importAppBefore(buildAppConfig("with-custom-function").anonAuth());
```
kraenhansen, line 578:
```suggestion
    importAppBefore(buildAppConfig("with-anon-auth").anonAuth());
```

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -574,4 +574,20 @@ describe.skipIf(environment.missingServer, "User", () => {
...
+
+    it("persists currentUser on opening the app", async function (this: AppContext & RealmContext) {
+      const credentials = Realm.Credentials.anonymous();
+      const user = await this.app.logIn(credentials);
+      const appId = this.app.id;
+      //@ts-expect-error Normally you wouldn't do this
+      delete this.app;
+      this.app = new Realm.App(appId);
```
kraenhansen, line 586: Wouldn't it be simpler to just have a new local app instance here?
```suggestion
      const newApp = new Realm.App(appId);
```

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -574,4 +574,20 @@ describe.skipIf(environment.missingServer, "User", () => {
...
+
+      const currentUser = this.app.currentUser;
+      expect(currentUser).not.to.be.null;
+      expect(user.id).to.equal(currentUser?.id);
```
kraenhansen, line 590: Wouldn't it be simpler to just have a new local app instance here?

```suggestion
      const newApp = new Realm.App(appId);

      const currentUser = newApp.currentUser;
      expect(currentUser).not.to.be.null;
      expect(user.id).to.equal(currentUser?.id);
```

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -23,7 +23,7 @@ import {
   randomVerifiableEmail,
 } from "../../utils/generators";
 import { KJUR } from "jsrsasign";
-import { UserState } from "realm";
+import Realm, { UserState } from "realm";
```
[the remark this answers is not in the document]
takameyer: Force of habit.  Should probably refactor that whole test for the named import (which was a bit out of scope with what I was doing)

packages/realm/src/app-services/App.ts
```diff
@@ -130,8 +133,12 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
 
   /** @internal */
-  public static getAppByUser(userInternal: binding.SyncUser): App {
-    const app = this.appByUserId.get(userInternal.identity)?.deref();
+  public static getAppByUser(userInternal: binding.SyncUser, currentApp?: AnyApp): App {
+    if (currentApp) {
+      App.appByUserId.set(userInternal.identity, new binding.WeakRef(currentApp));
+      return currentApp;
+    }
+    const app = App.appByUserId.get(userInternal.identity)?.deref();
```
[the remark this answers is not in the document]
takameyer: True, it is a bit of an unexpected side effect.  I'll give it some thought.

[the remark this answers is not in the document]
elle-j: Nice 🙂

packages/realm/src/app-services/App.ts
```diff
@@ -131,13 +134,20 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
     if (!app) {
       throw new Error(`Cannot determine which app is associated with user (id = ${userInternal.identity})`);
     }
     return app;
   }
 
+  /** @internal */
+  public static setAppByUser(userInternal: binding.SyncUser, currentApp: AnyApp): void {
+    if (currentApp) {
```
kraenhansen, line 146: I can't imagine this ever being falsy?

elle-j: The check could probably be removed since it's an internal method. I think it just got added since that code was previously used where the `currentApp` as optional 👍

PR #5792, opened 2023-05-05
Export Storage

divyam234:
export storage classes to final bundle to implement different storage policy for example KV storage of cf workers


divyam234:
@cla-bot check

kneth:
@divyam234 Any reason for closing the PR?

divyam234:
@kneth I don't think storage implementation is needed for KV storage of cf workers or edge-config of vercel  which I thought earlier. One can simply serialise user object to save login state in these storages.

PR #5796, opened 2023-05-08
Updated core to v 13.10.1

papafe:

PR #5797, opened 2023-05-08
Added support for full text search

### LINKED ISSUE #5755: Expose minimal full text search

papafe:
This closes #5755

Depends on updating core to the latest release (https://github.com/realm/realm-js/pull/5796). The update has already been merged in this branch.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests


CHANGELOG.md
```diff
@@ -7,7 +7,11 @@
...
 * Enable multiple processes to operate on an encrypted Realm simultaneously. ([realm/realm-core#1845](https://github.com/realm/realm-core/issues/1845))
 * Added `Realm.setLogger`, that allows to setup a single static logger for the duration of the app lifetime. Differently from the now deprecated sync logger (that was setup with `Sync.setLogger`), this new one will emit messages coming also from the local database, and not only from sync. It is also possible to change the log level during the whole duration of the app lifetime with `Realm.setLogLevel`. ([#2546](https://github.com/realm/realm-js/issues/2546))
+* Added support for a sync configuration option to provide an `SSLConfiguration` with a custom function for validating the server's SSL certificate. [#5485](https://github.com/realm/realm-js/issues/5485)
+* Extended `PropertySchema.indexed` with the `fulltext` option, that allows to create an index for full-text search queries.  [#5755](https://github.com/realm/realm-js/issues/5755)
```
kneth, line 11:
```suggestion
* Extended `PropertySchema.indexed` with the `fulltext` option, that allows to create an index for full-text search queries.  ([#5755](https://github.com/realm/realm-js/issues/5755))
```

kneth (approved):
Please create an issue to add support for the new index type in the Babel plugin.

kraenhansen (approved):

packages/realm/src/schema/to-binding.ts
```diff
@@ -99,10 +99,16 @@ export function toBindingPropertySchema(name: string, schema: CanonicalPropertyS
   const result: BindingProperty = {
     name,
     type: toBindingPropertyType(schema),
-    isIndexed: schema.indexed,
     objectType: schema.objectType && schema.objectType in TYPE_MAPPINGS ? undefined : schema.objectType,
     linkOriginPropertyName: schema.property,
   };
+
+  if (schema.indexed === "fulltext") {
```
takameyer, line 106: Is `fulltext` a single word?  I would expect this to be either `full-text` or `fullText`.

packages/realm/src/schema/types.ts
```diff
@@ -84,14 +84,21 @@ export type CanonicalRealmSchema = CanonicalObjectSchema[];
  */
 export type CanonicalObjectSchemaProperty = CanonicalPropertySchema;
 
+/**
+ * The type of index on a property:
+ * - 'true' enables a regular index
+ * - `fulltext` enables a full-text search index and can only be applied to string properties.
```
elle-j, line 90:
```suggestion
 * - `true` enables a regular index.
 * - `"fulltext"` enables a full-text search index and can only be applied to string properties.
```

packages/realm/src/schema/validate.ts
```diff
@@ -164,7 +164,10 @@ export function validatePropertySchema(
       assert.string(property, `'${propertyName}.property' on '${objectName}'`);
     }
     if (indexed !== undefined) {
-      assert.boolean(indexed, `'${propertyName}.indexed' on '${objectName}'`);
+      assert(
+        typeof indexed === "boolean" || indexed === "fulltext",
+        `"indexed" needs to be either a boolean or "fulltext"`,
```
elle-j, line 169: The change of error message should have caused [this validation test](https://github.com/realm/realm-js/blob/d9df13ff87a4c26e8e2eac188927ac4873ebc37f/packages/realm/src/tests/schema-validation.test.ts#L233C5-L240) to fail. You can run from the root `npm test --workspace realm`. Even with this change, the test should fail but the test need to be updated by changing the last part of the expected error message. (Can make a small PR against this branch to add minor things.)
```suggestion
        `Expected '${propertyName}.indexed' on '${objectName}' to be a boolean or 'fulltext'.`,
```

integration-tests/tests/src/tests/queries.ts
```diff
@@ -131,6 +150,65 @@ const expectQueryResultValues = (
 };
 
 describe("Queries", () => {
+  describe("Full text search", () => {
```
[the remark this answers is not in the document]
papafe: Added

packages/realm/src/schema/to-binding.ts
papafe: Usually it's written as "full-text". I didn't think too much about it, but I think it makes sense to write it like that

elle-j (commented):
Great new feature support 🙂 Left some comments. There are minor updates to things not included in this PR that I can add in another PR that targets this branch.

elle-j (approved):
Looks awesome! The only thing you may want to wait with is the spec changes. You could coordinate with Mathias on that 👍

packages/realm/src/schema/validate.ts
```diff
@@ -149,6 +148,7 @@ export function validatePropertySchema(
   objectName: string,
   propertyName: string,
   propertySchema: unknown,
+  primaryKey: string | undefined,
```
papafe, line 151: Does this make sense? @kraenhansen @elle-j @takameyer

elle-j: Currently we have the validation part (i.e. `validate.ts`) be responsible for checking that the JS data types are correct. E.g. checking that `objectType` is a string, but not checking that `objectType` only is set when `type` is e.g. `"list"` (that's done in `normalize.ts`). So the validation is just for making sure that other functions can rely on the types being what TS says it is.

So maybe we can move the logic of checking that the combination of `"full-text"` and other props are allowed, to `normalize.ts`?

papafe: Makes total sense!

PR #5802, opened 2023-05-09
Binding resolution refactored

kraenhansen:
## What, How & Why?

This fixes part of #5576 by declaring the output of the binding generator as a virtual "realm/binding" module, which we provide a path for implementation via tsconfig `paths`: A default fallback for tests (in `packages/realm/tsconfig.json`) as well as each of the platform specific directories.

It also removes the `typescript-checker`, as we concluded this might just be deferred and will happen as as part of building the SDK anyway.


packages/realm/bindgen/src/templates/typescript.ts
```diff
@@ -162,31 +162,21 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
...
 
-  coreOut("// Enums");
-  for (const e of spec.enums) {
-    // Using const enum to avoid having to emit JS backing these
-    coreOut(`export const enum ${e.jsName} {`);
-    coreOut(...e.enumerators.map(({ jsName, value }) => `${jsName} = ${value},\n`));
-    coreOut("};");
-  }
-  coreOut(`
+  out("declare module 'realm/binding' {");
```
kraenhansen, line 168: This is one of the more important changes.

packages/realm/bindgen/src/templates/node-wrapper.ts
```diff
@@ -122,6 +119,29 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
...
+  both("// Enums");
+  for (const e of spec.enums) {
+    // This pattern is copied from the TS compiler
+    both(`export var ${e.jsName};`);
+    both(`(function (${e.jsName}) {`);
+    both(
+      ...e.enumerators.map(({ jsName, value }) => `${e.jsName}[${e.jsName}["${jsName}"] = ${value}] = "${jsName}";`),
+    );
+    both(`})(${e.jsName} || (${e.jsName} = {}));`);
+  }
```
kraenhansen, line 143: Probably the most controversial change, but not a large price to pay to avoid relying on code replacements from the Rollup config.

packages/realm/bindgen/src/templates/typescript.ts
RedBeard0531: I think you are missing a word after "without"

packages/realm/bindgen/src/templates/typescript.ts
```diff
@@ -210,13 +200,21 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
     }
   `);
 
+  out("// Enums");
+  for (const e of spec.enums) {
+    // Using const enum to avoid having to emit JS backing these
+    out(`export enum ${e.jsName} {`);
```
RedBeard0531, line 206: You lost the `const` in `const enum`

packages/realm/bindgen/src/templates/node-wrapper.ts
RedBeard0531: const enums don't generate _any_ js code: https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM72MgrgWzgMQgjgG8BYAKDmrgDMiAaOAIwEMom2AvSgXyA. It only adds declarations to the .d.ts file.

packages/realm/bindgen/src/templates/typescript.ts
```diff
@@ -162,31 +162,21 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
 
   const spec = doJsPasses(bindModel(rawSpec));
 
-  const coreOut = file("core.ts", "eslint", "typescript-checker");
```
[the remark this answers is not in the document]
RedBeard0531: This doesn't make sense. core.ts was a plain ts file that was `export * from "./core"` in both the wrapper and in the .d.ts file. It doesn't need to export anything from "realm/binding" because the module imported by that name will transitively re-export all of its contents in both TS declarations, and in JS runtime land.

Also, you totally can do module augmentation from a ts file. I do it in [bindgen](https://github.com/realm/realm-js/blob/main/packages/realm/bindgen/src/js-passes.ts#L95-L114). Also [like this](https://github.com/realm/realm-js/blob/main/packages/realm/bindgen/src/templates/node.ts#L586-L594)

packages/realm/bindgen/src/templates/typescript.ts
kraenhansen: This is on purpose. Having a `const enum` in a non .d.ts file made the tests fail as it didn't have values at runtime. I suppose TSC only inlines these if they're in a TS source file.

packages/realm/bindgen/src/templates/typescript.ts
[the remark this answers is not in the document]
kraenhansen: When I include `export * from "./core";` in the `declare module "realm/binding" {` block of the `native.d.mts` I get:

> Import or export declaration in an ambient module declaration cannot reference module through relative module name.ts(2439)

I'll try doing module augmentation from the `core.ts` again 🤔

packages/realm/bindgen/src/templates/node-wrapper.ts
kraenhansen: I believe `const enum` is only supposed to be used in non-declaration files.

Here's an example:

```json
{
  "name": "ts-const-enums",
  "private": true,
  "type": "module",
  "scripts": {
    "test": "tsx index.ts"
  },
  "dependencies": {
    "tsx": "^3.12.7",
    "typescript": "^5.0.4"
  }
}
```

```ts
// index.ts
import { Foo } from "./enums";
console.log(Foo.bar);
```

```ts
// enums.d.ts
export const enum Foo {
  bar = "bar"
}
```

```ts
// enums.js
// note how we provide no implementation
```

Running `npm test` yeilds:

```
/Users/kraen.hansen/Projects/ts-const-enums/index.ts:1
import { Foo } from "./enums";
         ^

SyntaxError: The requested module './enums' does not provide an export named 'Foo'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:527:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

Even if we do provide a dummy implementation of `Foo` in `enums.js`:

```js
export const Foo = {};
```

This simply prints `undefined`.

packages/realm/bindgen/src/templates/typescript.ts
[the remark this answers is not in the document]
kraenhansen: > Also, you totally can do module augmentation from a ts file. I do it in [bindgen](https://github.com/realm/realm-js/blob/main/packages/realm/bindgen/src/js-passes.ts#L95-L114). Also [like this](https://github.com/realm/realm-js/blob/main/packages/realm/bindgen/src/templates/node.ts#L586-L594)

Those are examples of augmentation of the module declaration, not the implementation. We'd need the latter to implement `core.ts` since it provides the implementation too?

packages/realm/bindgen/src/templates/node-wrapper.ts
RedBeard0531: Maybe it behaves differently for string vs number `const enum`s? When I tested locally with this patch, everything seems to work:

```diff
diff --git a/packages/realm/bindgen/src/templates/node-wrapper.ts b/packages/realm/bindgen/src/templates/node-wrapper.ts
index c66cf638c..364d7d840 100644
--- a/packages/realm/bindgen/src/templates/node-wrapper.ts
+++ b/packages/realm/bindgen/src/templates/node-wrapper.ts
@@ -131,17 +131,6 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
     }
   `);

-  both("// Enums");
-  for (const e of spec.enums) {
-    // This pattern is copied from the TS compiler
-    both(`export var ${e.jsName};`);
-    both(`(function (${e.jsName}) {`);
-    both(
-      ...e.enumerators.map(({ jsName, value }) => `${e.jsName}[${e.jsName}["${jsName}"] = ${value}] = "${jsName}";`),
-    );
-    both(`})(${e.jsName} || (${e.jsName} = {}));`);
-  }
-
   const injectables = [
     "Long",
     "ArrayBuffer",
diff --git a/packages/realm/bindgen/src/templates/typescript.ts b/packages/realm/bindgen/src/templates/typescript.ts
index 3210f1f5c..0091d5f34 100644
--- a/packages/realm/bindgen/src/templates/typescript.ts
+++ b/packages/realm/bindgen/src/templates/typescript.ts
@@ -203,7 +203,7 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
   out("// Enums");
   for (const e of spec.enums) {
     // Using const enum to avoid having to emit JS backing these
-    out(`export enum ${e.jsName} {`);
+    out(`export const enum ${e.jsName} {`);
     out(...e.enumerators.map(({ jsName, value }) => `${jsName} = ${value},\n`));
     out("};");
   }
```

Of course, I've often wondered whether we should be using `enum` or `const enum` for these. `const enum` may provide some perf and size benefits, but I don't know how much that matters given the JIT in node and the bytecode compiler for hermes. But it certainly shouldn't be slower, and I don't think we've needed the features of a non-const enum yet (eg mapping numeric values to/from string names at runtime)

packages/realm/bindgen/CMakeLists.txt
RedBeard0531, line 1: Delete line 132 if you remove core.ts.

packages/realm/bindgen/src/templates/node-wrapper.ts
kraenhansen: When I apply this, the SDK unit test fails with:

```
✖ ERROR: TypeError: Cannot read properties of undefined (reading 'Get')
    at <anonymous> (/Users/kraen.hansen/Projects/realm-js/packages/realm/src/platform/network.ts:35:23)
    at Object.<anonymous> (/Users/kraen.hansen/Projects/realm-js/packages/realm/src/platform/network.ts:62:17)
```

kraenhansen:
I've pushed a commit pointing core to https://github.com/realm/realm-core/pull/6604, which I'll change once the Core PR has been merged.

packages/realm/bindgen/src/templates/typescript.ts
kraenhansen: Thanks - I updated the comment 👍

packages/realm/bindgen/src/templates/node-wrapper.ts
```diff
@@ -20,14 +20,15 @@ import { bindModel, Property } from "@realm/bindgen/bound-model";
...
 export function generate({ spec: rawSpec, file }: TemplateContext): void {
   const spec = doJsPasses(bindModel(rawSpec));
   const reactLines = [];
   const nodeLines = [];
-  function both(content: string) {
-    reactLines.push(content);
-    nodeLines.push(content);
+  function both(...content: string[]) {
+    reactLines.push(...content);
+    nodeLines.push(...content);
```
RedBeard0531, line 31: This change now seems unnecessary. But also probably harmless. 🤷

packages/realm/bindgen/src/templates/node-wrapper.ts
```diff
@@ -20,14 +20,15 @@ import { bindModel, Property } from "@realm/bindgen/bound-model";
 import { TemplateContext } from "@realm/bindgen/context";
 
 import { doJsPasses } from "../js-passes";
+import { eslintFormatter } from "../formatters/eslint-formatter";
```
RedBeard0531, line 23: I'm not sure it is worth introducing a new directory for just this one file. Personally, since the formatters are tiny, I would just put all of them in a single `../formaters.ts` file, even if we had more than one.

But at the very least, lets remove some redundancy in `src/formatters/eslint-formatter.ts`. If you want to keep one file per formatter, I'd go either `src/formatters/eslint.ts` or just `src/eslint-formatter.ts`. Slight preference for the latter if you don't want just `src/formatters.ts`

packages/realm/bindgen/src/templates/node.ts
```diff
@@ -31,6 +31,7 @@ import {
   Pointer,
   Template,
 } from "@realm/bindgen/bound-model";
+import { clangFormatter } from "@realm/bindgen/formatter";
```
RedBeard0531, line 34: It feels a bit weird to call this `clangFormatter` since the tool is named `clang-format`, and the tool called `clang` doesn't do any formatting. I admit that `clangFormatFormatter` sounds ridiculous though. Maybe we should just call the formatters by the name of the tool with no suffix, so just `clangFormat` and `eslint`?

packages/realm/bindgen/src/templates/typescript.ts
```diff
@@ -162,7 +149,7 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
...
+  const coreOut = file("core.ts", eslintFormatter);
```
RedBeard0531, line 152:
```suggestion
  // This file is imported by the virtual path "realm/binding/core" in TS.
  const coreOut = file("core.ts", eslintFormatter);
```

packages/realm/bindgen/src/templates/typescript.ts
RedBeard0531: I still don't love that this file is picked up by the `src/*.ts` glob and injects its contents into a virtual module rather than being found via an `import` path. But I don't currently have a great idea for an alternative if we want to move away from rollup's replace plugin.

packages/realm/bindgen/src/templates/typescript.ts
```diff
@@ -216,7 +205,7 @@ export function generate({ spec: rawSpec, file }: TemplateContext): void {
 
   out("// Opaque types (including Key types)");
   for (const { jsName } of (spec.opaqueTypes as NamedType[]).concat(spec.keyTypes)) {
-    out.lines("/** Using an empty enum to express a nominal type */", `export declare enum ${jsName} {}`);
+    out.lines("/** Using an empty enum to express a nominal type */", `export enum ${jsName} {}`);
```
RedBeard0531, line 208:
```suggestion
    out.lines("/** Using an empty enum to express a nominal type */", `export const enum ${jsName} {}`);
```

We should probably consider some of the alternative mechanisms we discovered for opaque types at some point. But at the very least, we shouldn't imply that there is any actual value with that name.

packages/realm/src/binding.ts
```diff
@@ -16,13 +16,13 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import { IndexSet, Int64, ObjKey, Timestamp } from "../generated/ts/native.mjs";
+import { IndexSet, Int64, ObjKey, Timestamp } from "realm/binding";
 
 /** @internal */
-export * from "../generated/ts/native.mjs"; // core types are transitively exported.
+export * from "realm/binding";
```
RedBeard0531, line 22: I know we just spend a long time figuring out how to make the re-export of every type from `core.ts` work through `native.d.mts`, but what if we just ... didn't. Do you think it would actually be cleaner to have this file do the `export * from "../generated/ts/core";` rather than doing that in the wrapper and the .d.ts? That might let you remove the virtual `realm/binding/core` path.

Feel free to ignore if you are just tired of fucking around with that file 😁

packages/realm/src/platform/node/fs.ts
```diff
@@ -19,8 +19,8 @@
 import { existsSync, mkdirSync, readdirSync, rmSync, unlinkSync } from "node:fs";
 import { dirname, isAbsolute, join } from "node:path";
 
-import { inject } from "../platform/file-system";
-import { extendDebug } from "../debug";
+import { inject } from "../file-system";
+import { extendDebug } from "../../debug";
```
RedBeard0531, line 23: We should consider defining a virtual path `REALM/` or `src/` so that imports look the same regardless of where you are in the directory structure. It is really unfortunate that when you move a file around, you need to modify its own imports, not just the imports _of_ it. This is pretty common in C++ projects to always use "fully qualified" paths in #includes, so you use `#include "realm/object-store/object.hpp"` rather than `#include "./object.hpp"` if you are in `realm/object-store/blah.hpp`.

packages/realm/tsconfig.json
RedBeard0531, line 1: Is this file only used for editors and bundle.d.ts generation?

RedBeard0531 (approved):
Comments are just suggestions. Everything looks correct.

packages/realm/tsconfig.json
kraenhansen: I believe so.

packages/realm/bindgen/src/templates/node-wrapper.ts
kraenhansen: Moved it one level up again 👍

packages/realm/src/platform/node/fs.ts
kraenhansen: I think that's a good idea, but I'll leave it for a future PR 👍 Ideally I would also like to find a way to enforce that we import through the `internal` module instead of relative paths (for all other files than the platform specific).

packages/realm/bindgen/src/templates/typescript.ts
RedBeard0531: You mean like [we already do for classes](https://github.com/realm/realm-js/blob/63c2fcb315f95fa89ecf9daccf4755a9d973f9bc/packages/realm/bindgen/src/templates/typescript.ts#L256-L257)? Yes, possibly something like that. I wish there was a way to tell TS to model the behavior of `Object.create(null)` correctly, so that we could create a class that really had no known properties, but that doesn't seem possible right now.

kraenhansen:
I'll merge now, as the two failing "Test Coverage" scripts seem very unrelated to the change I'm proposing.

PR #5807, opened 2023-05-11
SDK metrics

kneth:
## What, How & Why?

Aligning with other Realm SDKs what is submitted

- [x] SDK metrics (post-installation script)
- [ ] Sync connection parameters

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* ~[ ] jsdoc files updated~


scripts/submit-analytics.js
```diff
@@ -62,15 +64,12 @@ const getAnalyticsRequestUrl = (payload) =>
...
+ *
+ * @param {string} data
+ * @returns base64 encoded SHA256 of data
  */
-function sha256(data) {
-  let hash = require("crypto").createHash("sha256");
-  hash.update(data);
-  return hash.digest("hex");
-}
+const sha256 = (data) => createHmac("sha256", Buffer.from("Realm is great")).update(data).digest().toString("base64");
```
kraenhansen, line 72: If this is a salt, I think we should extract it to a constant called `SALT`.

scripts/submit-analytics.js
```diff
@@ -122,6 +121,40 @@ function getRealmVersion() {
...
+  const root = getProjectRoot();
+  if (fse.exists(path.resolve(root, "yarn.lock"))) {
+    const yarnVersion = execSync("yarn --version").toString().trim();
+    return ["yarn", yarnVersion];
+  }
+  if (fse.exists(path.resolve(root, "package-lock.json"))) {
+    const npmVersion = execSync("npm --version").toString().trim();
+    return ["npm", npmVersion];
+  }
+  return ["unknown", "unknown"];
```
kraenhansen, line 155: I don't think this is reliable:
1. The repository might have a `yarn.lock` and and the user might have `yarn` installed, but have used npm to install
2. Even worse, the repo might have a `yarn.lock` but the user does not have `yarn` installed, in which case `execSync` will throw and the entire submit will fail.

Instead, I suggest using the `npm_config_user_agent` environment variable to determine this:
- Using `npm`: `npm/8.19.1 node/v18.9.0 darwin arm64 workspaces/false`
- Using `yarn`: `yarn/1.22.19 npm/? node/v18.9.0 darwin arm64`

kneth: Cool, I will parse `npm_config_user_agent` instead of adding an extra dependency.

scripts/submit-analytics.js
```diff
@@ -122,6 +121,40 @@ function getRealmVersion() {
...
+ * @returns the Realm Core version as a string
+ */
+function getRealmCoreVersion() {
+  const dependenciesListPath = path.resolve(__dirname, "../dependencies.list");
+  const dependenciesList = fse
+    .readFileSync(dependenciesListPath)
+    .toString()
+    .split("\n")
+    .map((s) => s.split("="));
+  return dependenciesList.find((e) => e[0] === "REALM_CORE_VERSION")[0];
```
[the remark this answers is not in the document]
kneth: I will add a test too

scripts/submit-analytics.js
kraenhansen: A regular expression should do the trick 👍

scripts/submit-analytics.js
```diff
@@ -122,6 +124,32 @@ function getRealmVersion() {
...
+    .toString()
+    .split("\n")
+    .map((s) => s.split("="));
+  return dependenciesList.find((e) => e[0] === "REALM_CORE_VERSION")[1];
+}
+
+/**
+ * Determines if `npm` or `yarn` is used.
+ *
+ * @returns An array with two elements: method and version
```
elle-j, line 146: Not sure if it's necessary to mention what I added below. (Saw it from [this npm package](https://www.npmjs.com/package/npm-config-user-agent-parser?activeTab=readme).)
```suggestion
 * @returns An array with two elements: method and version. (The version for npm will be "?" if npm is not used).
```

elle-j (approved):
Awesome, thanks for adding this 🙂

scripts/submit-analytics.js
```diff
@@ -122,6 +124,32 @@ function getRealmVersion() {
...
+ */
+function getInstallationMethod() {
+  const userAgent = process.env["npm_config_user_agent"];
+  return userAgent.split(" ")[0].split("/");
```
kraenhansen, line 150: Although I like the simplicity, I think it's a little brittle to depend on the order of these elements 🙂

scripts/submit-analytics.js
```diff
@@ -122,6 +124,32 @@ function getRealmVersion() {
...
+ * @returns the Realm Core version as a string
+ */
+function getRealmCoreVersion() {
+  const dependenciesListPath = path.resolve(__dirname, "../dependencies.list");
+  const dependenciesList = fse
+    .readFileSync(dependenciesListPath)
+    .toString()
+    .split("\n")
+    .map((s) => s.split("="));
+  return dependenciesList.find((e) => e[0] === "REALM_CORE_VERSION")[1];
```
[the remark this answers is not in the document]
elle-j: Ah okay, that's why then 😛

scripts/submit-analytics.js
kneth: I believe it will also return version if you are using yarn

scripts/submit-analytics.js
kneth: I think it conforms to [a typical user-agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent), and it will also have "product/version" in the beginning of the string.

Looking at the document for [npm](https://docs.npmjs.com/cli/v9/using-npm/config#user-agent) and [yarn](https://classic.yarnpkg.com/lang/en/docs/cli/config/) also support this, and I don't know if I can make it less brittle.

scripts/submit-analytics.js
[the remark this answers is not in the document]
kneth: For `main` we might have to reintroduce it.

scripts/submit-analytics.js
elle-j: Good point!

nielsenko (approved):
LGTM

PR #5808, opened 2023-05-11
Update install tests to specify Ruby version

kraenhansen:
## What, How & Why?

Due to https://github.com/facebook/react-native/pull/36423 we need specify a Ruby version.


kraenhansen:
I triggered a run, which succeeded on all variants 🎉  https://github.com/realm/realm-js/actions/runs/4950952428

takameyer (approved):
Good catch.  LGTM

PR #5815, opened 2023-05-16
Hold SyncSessions with a std::weak_ptr, except while calling a method

RedBeard0531:
This avoids keeping SyncSessions alive due to lazy GC. This also matches the implementation technique of v11, which also held a weak_ptr for the same reasons.

## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/bindgen/js_spec.yml
+++ b/packages/realm/bindgen/js_spec.yml
@@ -17,3 +17,12 @@ classes:
       get_cpu_arch: () -> std::string
       # print: (const char* fmt, ...) # can't expose varargs directly. Could expose a fixed overload.
 
+  WeakSyncSession:
+    cppName: std::weak_ptr<SyncSession>
+    constructors:
+      weakCopyOf: '(shared: SharedSyncSession)'
+    methods:
+      rawDereference:
+        sig: '() const -> Nullable<SharedSyncSession>'
+        cppName: lock
+
--- a/packages/realm/bindgen/src/realm_js_jsi_helpers.h
+++ b/packages/realm/bindgen/src/realm_js_jsi_helpers.h
@@ -134,7 +134,7 @@ REALM_NOINLINE inline jsi::Value toJsiException(jsi::Runtime& env, const std::ex
 
 [[noreturn]] REALM_NOINLINE inline void throwNullSharedPtrError(jsi::Runtime& env, const char* clsName)
 {
-    throw jsi::JSError(env, util::format("Attempting to use an instanace of $1 holding a null shared_ptr. "
+    throw jsi::JSError(env, util::format("Attempting to use an instanace of %1 holding a null shared_ptr. "
                                          "Did you call $resetSharedPtr on it already?",
                                          clsName));
 }
--- a/packages/realm/src/app-services/SyncSession.ts
+++ b/packages/realm/src/app-services/SyncSession.ts
@@ -103,7 +103,6 @@ export function toBindingErrorHandler(onError: ErrorCallback) {
     const session = new SyncSession(sessionInternal);
     const error = fromBindingSyncError(bindingError);
     onError(session, error);
-    session.resetInternal();
   };
 }
 
@@ -208,7 +207,7 @@ export function toBindingClientResetMode(resetMode: ClientResetMode): binding.Cl
 }
 
 type ListenerToken = {
-  internal: binding.SyncSession;
+  weakInternal: binding.WeakSyncSession;
   token: binding.Int64;
 };
 
@@ -229,19 +228,19 @@ const mockApp = new Proxy({} as App, {
 const PROGRESS_LISTENERS = new Listeners<
   ProgressNotificationCallback,
   ListenerToken,
-  [binding.SyncSession, ProgressDirection, ProgressMode]
+  [binding.WeakSyncSession, binding.SyncSession, ProgressDirection, ProgressMode]
 >({
   throwOnReAdd: true,
-  add(callback, internal, direction, mode) {
+  add(callback, weakInternal, internal, direction, mode) {
     const token = internal.registerProgressNotifier(
       (transferred, transferable) => callback(Number(transferred), Number(transferable)),
       toBindingDirection(direction),
       mode === ProgressMode.ReportIndefinitely,
     );
-    return { internal, token };
+    return { weakInternal, token };
   },
-  remove({ internal, token }) {
-    return internal.unregisterProgressNotifier(token);
+  remove({ weakInternal, token }) {
+    weakInternal.withDeref((internal) => internal?.unregisterProgressNotifier(token));
   },
 });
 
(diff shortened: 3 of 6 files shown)


packages/realm/bindgen/js_spec.yml
```diff
@@ -17,3 +17,12 @@ classes:
       get_cpu_arch: () -> std::string
       # print: (const char* fmt, ...) # can't expose varargs directly. Could expose a fixed overload.
 
+  WeakSyncSession:
+    cppName: std::weak_ptr<SyncSession>
```
RedBeard0531, line 21: This is a _bit_ of a hack, but I think it worked out reasonably well. I wanted to keep this contained to realm-js since I don't think other SDKs need to do this, and I'd rather not expose them to `weak_ptr` unless we have to. I also didn't want to have to start thinking about how we would handle `weak_ptr` generically in when generating code, so I kept it as a special case for SyncSession.

packages/realm/bindgen/js_spec.yml
```diff
@@ -17,3 +17,12 @@ classes:
...
+    constructors:
+      weakCopyOf: '(shared: SharedSyncSession)'
+    methods:
+      rawDereference:
+        sig: '() const -> Nullable<SharedSyncSession>'
+        cppName: lock
```
RedBeard0531, line 27: None of these are directly used by code. Both are only used by the friendlier helpers in `binding.ts` at the very bottom of this PR.

packages/realm/bindgen/src/realm_js_jsi_helpers.h
```diff
@@ -134,7 +134,7 @@ REALM_NOINLINE inline jsi::Value toJsiException(jsi::Runtime& env, const std::ex
 
 [[noreturn]] REALM_NOINLINE inline void throwNullSharedPtrError(jsi::Runtime& env, const char* clsName)
 {
-    throw jsi::JSError(env, util::format("Attempting to use an instanace of $1 holding a null shared_ptr. "
+    throw jsi::JSError(env, util::format("Attempting to use an instanace of %1 holding a null shared_ptr. "
```
RedBeard0531, line 137: I don't know why by brain wants to use `$1` placeholders rather than `%1` with `util::format`, but I seem to make this typo a lot...

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -229,106 +228,90 @@ const mockApp = new Proxy({} as App, {
...
   },
-  remove({ internal, token }) {
-    internal.unregisterConnectionChangeCallback(token);
+  remove({ weakInternal, token }) {
+    weakInternal.withDeref((internal) => internal?.unregisterConnectionChangeCallback(token));
   },
 });
 
 export class SyncSession {
-  private static instances = new Set<binding.WeakRef<SyncSession>>();
```
RedBeard0531, line 266: I'm pretty sure we can kill this off now that we are using `weak_ptr` since a long-lived `SyncSession` won't keep the underlying C++ object alive anymore.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -229,106 +228,90 @@ const mockApp = new Proxy({} as App, {
...
   /** @internal */
-  private _internal: binding.SyncSession | null;
+  private weakInternal: binding.WeakSyncSession;
   /** @internal */
-  public get internal() {
-    assert(this._internal, "This SyncSession is no longer valid");
-    return this._internal;
+  public withInternal<Ret = void>(cb: (syncSession: binding.SyncSession) => Ret) {
+    return this.weakInternal.withDeref((syncSession) => {
+      assert(syncSession, "This SyncSession is no longer valid");
```
RedBeard0531, line 274: Note: this is the place that unconditionally throws from each method if the C++ object has been destroyed already. See the slack poll on #realm-js-team. This PR currently implements behavior (1) but I can easily change that if the consensus forms around a different option.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -229,106 +228,90 @@ const mockApp = new Proxy({} as App, {
...
       toBindingDirection(direction),
       mode === ProgressMode.ReportIndefinitely,
     );
-    return { internal, token };
+    return { weakInternal, token };
   },
-  remove({ internal, token }) {
-    return internal.unregisterProgressNotifier(token);
+  remove({ weakInternal, token }) {
+    weakInternal.withDeref((internal) => internal?.unregisterProgressNotifier(token));
```
RedBeard0531, line 243: Note: unlike other cases, this will *NOT* throw if the Session has been destroyed. `removing()` a listener from a destroyed session will silently do nothing. This seams reasonable to me since the listener was removed when the session was destroyed, and because `remove()` seems likely to be called as part of teardown, which may have caused the session to be destroyed already.

kraenhansen: I'm a bit curious on why `internal` is able to be undefined / null, i.e. why you need the `?.` here.

packages/realm/src/app-services/SyncSession.ts
kraenhansen: That would be great 🙏

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -229,106 +228,90 @@ const mockApp = new Proxy({} as App, {
...
-    return this._internal;
+  public withInternal<Ret = void>(cb: (syncSession: binding.SyncSession) => Ret) {
+    return this.weakInternal.withDeref((syncSession) => {
+      assert(syncSession, "This SyncSession is no longer valid");
+      return cb(syncSession);
+    });
   }
 
   /** @internal */
   constructor(internal: binding.SyncSession) {
```
kraenhansen, line 280: I wonder if it's a problem that a strong `binding.SyncSession` is ever returned to the SDK?

packages/realm/src/app-services/SyncSession.ts
kraenhansen: Let's try this and see if it becomes too painful to work with 👍

kraenhansen (approved):

packages/realm/bindgen/js_spec.yml
kraenhansen: If they're not intended for consumption by the SDK, could we find a way to declare that? Perhaps in the `binding.ts` by marking them as deprecated?

kraenhansen:
Looks like the tests are generally happy 👍

elle-j (approved):

packages/realm/src/app-services/SyncSession.ts
RedBeard0531: Do you think it is worth a comment like `// Note: not using the withInternal() helper, so internal can be null`?

packages/realm/src/app-services/SyncSession.ts
RedBeard0531: I _think_ everywhere we touch a `binding.SyncSession` it basically immediately goes through here. In particular, I was concerned about these, and considered adding a call to `weaken()` immediately after getting a `binding.SyncSession` out of native, but I realized that it would still happen at exactly the same point, so didn't seem worth it. https://github.com/realm/realm-js/blob/fb7de6e1367a48a91f1892cfa90885537450c346/packages/realm/src/app-services/Sync.ts#L74-L87

packages/realm/bindgen/js_spec.yml
```diff
@@ -17,3 +17,12 @@ classes:
       get_cpu_arch: () -> std::string
       # print: (const char* fmt, ...) # can't expose varargs directly. Could expose a fixed overload.
 
+  WeakSyncSession:
+    cppName: std::weak_ptr<SyncSession>
+    constructors:
+      weakCopyOf: '(shared: SharedSyncSession)'
```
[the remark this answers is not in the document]
RedBeard0531: Honestly, I have no idea what the right policy is for these two. This is in JS specific code, and both are describing names that are already detached from the C++ name, so it _feels_ reasonable to me to just use JS style directly. But I could also see an argument that all names in all spec files should consistently use C++ style. Does anyone have a strong opinion here?

packages/realm/bindgen/js_spec.yml
kraenhansen: > like the pattern of using `binding.ts` to augment the raw C++ APIs from TS.

I agree, that seems best.

packages/realm/src/app-services/SyncSession.ts
kraenhansen: Nah - I just had to take a step back and read it properly 👍

packages/realm/bindgen/js_spec.yml
RedBeard0531: I didn't see a good technique that would let me add annotations or comments to both the static and instance methods. If you care, and are able to find such a technique, we can always add it later.

packages/realm/src/app-services/SyncSession.ts
kraenhansen: After writing this comment, I understood that `weaken` will call the `$resetSharedPtr` which makes me less worried about a potential issue.

packages/realm/bindgen/js_spec.yml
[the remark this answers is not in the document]
kraenhansen: > all names in all spec files should consistently use C++ style

I'm leaning that way. It gives us less to consider and keeps us from accidentally transferring this habit across to the `spec.yml` when moving something between the files.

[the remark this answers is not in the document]
RedBeard0531: Sold. Done.

packages/realm/bindgen/src/realm_js_jsi_helpers.h
kneth: I have typed `$` so many times that I know I have to stop and think

packages/realm/bindgen/js_spec.yml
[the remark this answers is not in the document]
elle-j: Agree, personally I think it's easier to just stick to the same convention in all specs.

kneth (approved):

PR #5820, opened 2023-05-17
Utilize Bindgen opt-in list

### LINKED ISSUE #5275: Allow-list to narrow subset of spec.yml
We want to make it possible for the SDK to specify a subset of features to use from the spec.yml.

elle-j:
## What, How & Why?

Adds a Bindgen opt-in list (spec file) and utilizes the `opt-in` option added in the [Core PR](https://github.com/realm/realm-core/pull/6639).

The Core PR description elaborates on the expected shape of the opt-in list and other affected parts.

### Updates in this PR

**Main usage update:**
* The `BoundSpec` passed to each template's `generate()` function has an `applyOptInList()` method which we call in our `doJsPasses()`.
* Once applied, the methods and fields of the `BoundSpec` have an `isOptedInTo` boolean property that can be handled accordingly.

**New JSI, node-wrapper, and node templates usage:**
* They check for `!isOptedInTo` to skip producing bindings.

**New TS template usage:**
* It checks for `!isOptedInTo` to add deprecation annotations to the generated binding types.

### Developer requirements when updating spec

If changes are made to the general spec in Core or the JS-specific spec (i.e. addition/removal/renaming of a class, method, record, or field), that change needs to be reflected in the [opt-in list](https://github.com/realm/realm-js/blob/main/packages/realm/bindgen/js_opt_in_spec.yml).

This closes #5275 

## ☑️ ToDos
* [x] 📝 Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@
 * Aligning analytics with other Realm SDKs. You can still disable the submission by setting environment variable `REALM_DISABLE_ANALYTICS`, and you can print out what is submitted by setting the environment variable `REALM_PRINT_ANALYTICS`.
 * Disabling sync session multiplexing by default in the SDK, since Core's default changed to enabled with v13.11.0. ([#5831](https://github.com/realm/realm-js/pull/5831))
 * Upgraded Realm Core from v13.11.1 to v13.13.0. ([#5873](https://github.com/realm/realm-js/pull/5873))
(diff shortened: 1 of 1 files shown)


kneth (approved):

takameyer (approved):
LGTM!

RedBeard0531 (approved):
Nice work! Any chance you have numbers for how much this shrinks our release builds?

PR #5830, opened 2023-05-21
Overhaul SDK metrics

### LINKED ISSUE #5242: Submit analytics on the bindgen SDK
We submit analytics when users install our package.
We'll need to copy the script, update our package.json to run it on install and update the Node.js specific tests of this (mentioned in https://github.com/realm/realm-js/pull/5241) to test this with the new SDK package too.

kneth:
## What, How & Why?

This closes #5242 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated


elle-j (approved):
Nice!

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -16,6 +16,356 @@
...
+// Why are we doing this? In short, because it helps us build a better product
+// for you. None of the data personally identifies you, your employer or your
+// app, but it *will* help us understand what language you use, what Node.js
+// versions you target, etc. Having this info will help prioritizing our time,
+// adding new features and deprecating old features. Collecting an anonymized
+// application path & anonymized machine identifier is the only way for us to
+// count actual usage of the other metrics accurately. If we don’t have a way to
+// deduplicate the info reported, it will be useless, as a single developer
+// `npm install`-ing the same app 10 times would report 10 times more than another
+// developer that only installs once, making the data all but useless.
```
kraenhansen, line 30:
```suggestion
// developer that only installs once, making the data less useless.
```

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -16,6 +16,356 @@
...
+import process from "node:process";
+import https from "node:https";
+import os from "node:os";
+import console from "node:console";
+import { createHmac } from "node:crypto";
+import { Buffer } from "node:buffer";
+
+import machineId from "node-machine-id";
+import commandLineArgs from "command-line-args";
+import fse from "fs-extra";
```
kraenhansen, line 56: `fs-extra` and `command-line-args` are both missing as dependencies.

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -16,6 +16,356 @@
...
+}
+
+if (log) {
+  doLog = (msg) => console.log(msg);
+} else {
+  // eslint-disable-next-line no-unused-vars
+  doLog = () => {
+    /* don't log */
+  };
+}
```
kraenhansen, line 367: In general, I don't really like this pattern of reassigning a function - as something might capture it in a scoped variable and it won't update as expected.

In general, I would rather that we implement a (const) `log` function that does a check on the `options.log` before calling `console.log` or similar.

Or even better, that we just rely on the "debug" package, which we already have a dependency on - that way we could get rid of the `--log` runtime parameter, leaving just `--dry-run` which could be replaced with the `REALM_DISABLE_ANALYTICS` (which is already essentially doing the same anyway) - and if we did that we could drop the `command-line-args` dependency entirely. Just a thought ...

integration-tests/tests/src/node/analytics.ts
```diff
@@ -64,7 +81,10 @@ describe.skip("Analytics", () => {
     expect(data.Version).equals("1.2.3");
     expect(data.Framework).equals("electron");
     expect(data["Framework Version"]).equals("1.0.1");
```
takameyer, line 83: This test seems to be failing.  It is expecting `^1.0.0`

integration-tests/tests/src/node/analytics.ts
```diff
@@ -73,6 +93,9 @@ describe.skip("Analytics", () => {
     expect(data.Version).equals("1.2.3");
     expect(data.Framework).equals("react-native");
     expect(data["Framework Version"]).equals("1.0.1");
```
takameyer, line 95: Same test from above is failing here as well

takameyer (approved):
👍🏼

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -16,6 +16,356 @@
...
+// - An anonymous machine identifier and hashed application path to aggregate
+//   the other information on.
+
+import fs from "node:fs";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+import process from "node:process";
+import https from "node:https";
+import os from "node:os";
+import console from "node:console";
```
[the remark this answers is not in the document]
kneth: Yes, ESLint was loudly complaining.

packages/realm/scripts/submit-analytics.mjs
kneth: `command-line-args` is in the root `package.json` and I guess we should move it. And we don't see the issue with `fs-extra` in the integration tests as `integration-tests/environments/electron/package.json` has it as a dependency. I'll add both to the `realm` package's dependencies.

packages/realm/scripts/submit-analytics.mjs
kneth: Good point. I'll rewrite it to use the `debug` package and see if I can get rid of `command-line-args`

integration-tests/tests/src/node/analytics.ts
```diff
@@ -25,28 +25,30 @@ import { readFileSync } from "node:fs";
...
 type Fixture = "node" | "react-native" | "electron";
 
-// TODO: Update this to use ESM friendly APIs
-describe.skip("Analytics", () => {
+describe("Analytics", () => {
   function resolvePath(fixture: Fixture) {
     return path.resolve(__dirname, "fixtures", fixture);
   }
 
   function getRealmVersion() {
```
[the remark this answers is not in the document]
kneth: I agree, using the same implementation will make it easier to read and understand the code

PR #5832, opened 2023-05-22
WASM fixes

fealebenpae:


packages/realm/bindgen/src/templates/browser.ts
```diff
@@ -909,34 +912,25 @@ class BrowserCppDecls extends CppDecls {
...
         }        
       `);
     });
 
-    // export free functions via embind
+    // export method functions via embind
     out(`\nEMSCRIPTEN_BINDINGS(realm_c_api) {`);
-    this.free_funcs
-      .filter((free_func: CppFunc) => {
-        // FIXME:  manually exclude some function signatrue causing issue wiht embind (ex: call to deleted constructor of 'realm::ThreadSafeReference')
```
nhachicha, line 920: Are these fixed now?

nhachicha (approved):
Did you end up implementing a workaround for the `std::move`? I didn't manage to locate it within the browser template

nhachicha (approved):
Did you end up implementing a workaround for the `std::move`? I didn't manage to locate it within the browser template

fealebenpae:
@nhachicha no, the real fix was to not expose the `EMVAL_FROM`/`EMVAL_TO` conversion functions to embind - they were returning `T&`, which embind treats as plain `T` and attempts to copy, which failed for types with deleted copy constructors.

packages/realm/bindgen/src/templates/browser.ts
fealebenpae: They are excluded at an earlier level - only the functions in `this.methodFunctions` are declared inside the `EMSCRIPTEN_BINDINGS` block.

PR #5835, opened 2023-05-22
Use Object.defineProperty rather than Object.defineProperties because it is faster

RedBeard0531:
Yes, this is kinda stupid, but the results are fairly significant. And, IMO at least, the code is as clear or clearer like this.

FYI, I'm looking into how to get our list perf back up to v11 levels (or better), and this was part of that work. Once I discovered the impact of this, I figured I might as well globally replace all of `Object.defineProperties` and bank that improvement, before doing additional List/OrderedCollection-specific optimization. I'll pick those back up tomorrow.

Edit: I decided to sneak a second related commit in once I remembered that the combination of `prop!: Type` and `Object.declareProperty()` was bad news. The first commit made it more obvious that we were doing that.

Before:
```
      ✓ reads bool?[] (81,881 ops/sec, ±22%)
      ✓ reads bool?<> (108,726 ops/sec, ±14.99%)
      ✓ reads bool?{} (211,068 ops/sec, ±15.59%)
```

After first commit (using `Object.declareProperty`):
```
      ✓ reads bool?[] (90,663 ops/sec, ±29.44%)
      ✓ reads bool?<> (117,569 ops/sec, ±24.64%)
      ✓ reads bool?{} (255,618 ops/sec, ±24.08%)
```

After second commit (using `declare property`):
```
      ✓ reads bool?[] (142,468 ops/sec, ±42.28%)
      ✓ reads bool?<> (185,514 ops/sec, ±28.95%)
      ✓ reads bool?{} (316,215 ops/sec, ±27.6%)
```

## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/ClassMap.ts
+++ b/packages/realm/src/ClassMap.ts
@@ -79,46 +79,39 @@ export class ClassMap {
     const properties = [...schema.persistedProperties, ...schema.computedProperties];
     const propertyNames = properties.map((p) => p.publicName || p.name);
 
-    // Build a map of property descriptors from the properties declared in the schema
-    const descriptors: PropertyDescriptorMap = Object.fromEntries(
-      properties.map((property) => {
-        const propertyName = property.publicName || property.name;
-        const { get, set } = propertyMap.get(propertyName);
-        return [
-          propertyName,
-          {
-            enumerable: true,
-            get(this: RealmObject) {
-              return get(this[INTERNAL]);
-            },
-            set(this: RealmObject, value: unknown) {
-              set(this[INTERNAL], value);
-            },
-          },
-        ];
-      }),
-    );
+    // Set up accessors for the properties declared in the schema
+    for (const property of properties) {
+      const propertyName = property.publicName || property.name;
+      const { get, set } = propertyMap.get(propertyName);
+      Object.defineProperty(constructor.prototype, propertyName, {
+        enumerable: true,
+        get(this: RealmObject) {
+          return get(this[INTERNAL]);
+        },
+        set(this: RealmObject, value: unknown) {
+          set(this[INTERNAL], value);
+        },
+      });
+    }
 
-    descriptors[REALM] = {
+    Object.defineProperty(constructor.prototype, REALM, {
       enumerable: false,
       configurable: false,
       writable: false,
       value: realm,
-    };
-    descriptors[KEY_ARRAY] = {
+    });
+    Object.defineProperty(constructor.prototype, KEY_ARRAY, {
       enumerable: false,
       configurable: false,
       writable: false,
       value: propertyNames,
-    };
-    descriptors[KEY_SET] = {
+    });
+    Object.defineProperty(constructor.prototype, KEY_SET, {
       enumerable: false,
       configurable: false,
       writable: false,
       value: new Set(propertyNames),
-    };
-
-    Object.defineProperties(constructor.prototype, descriptors);
+    });
   }
 
   constructor(realm: Realm, realmSchema: readonly BindingObjectSchema[], canonicalRealmSchema: CanonicalRealmSchema) {
--- a/packages/realm/src/Collection.ts
+++ b/packages/realm/src/Collection.ts
@@ -56,12 +56,10 @@ export abstract class Collection<
       },
     });
     // Make the internal properties non-enumerable
-    Object.defineProperties(this, {
-      listeners: {
-        enumerable: false,
-        configurable: false,
-        writable: false,
-      },
+    Object.defineProperty(this, "listeners", {
+      enumerable: false,
+      configurable: false,
+      writable: false,
     });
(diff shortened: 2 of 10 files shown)


RedBeard0531:
Converting to draft to fix test failures. Hold off on reviewing for now.

kneth (approved):
The numbers are clear, and it makes sense to let them show the path.

Interestingly, [`Object.defineProperty()` is reported to be slow](https://bugs.chromium.org/p/v8/issues/detail?id=8447#c12) for V8. It doesn't contradict this PR and it hints that `Object.defineProperties()` is much slower than it should be.

Do we have numbers for Hermes and JSC?

RedBeard0531:
> The numbers are clear, and it makes sense to let them show the path.
> 
> Interestingly, [`Object.defineProperty()` is reported to be slow](https://bugs.chromium.org/p/v8/issues/detail?id=8447#c12) for V8. It doesn't contradict this PR and it hints that `Object.defineProperties()` is much slower than it should be.

I think that bug was before https://v8.dev/blog/faster-class-features. That post was what tipped me off that there might be some new optimizations available for `Object.defineProperties()`, and I decided to check if they weren't all available to `defineProperties()`.

> Do we have numbers for Hermes and JSC?

Not yet. I can get hermes numbers today. Would JSC numbers affect our decision at all?

kraenhansen (approved):
This is mildly frustrating. Do you have any insights into why its faster to call `Object.defineProperty`? Are you sure this is the case on Hermes & JSC too?

I would love if we could document these findings / assumptions somewhere, so we don't accidentally refactor this in the future and regress the performance gain.

RedBeard0531:
> Converting to draft to fix test failures. Hold off on reviewing for now.

Ok, fixed. The issue was that we were using `this.realm` in the constructor prior to setting the property, and I changed it to just `realm` (the argument). I also changed `this.results` to `results` in the `addListener` callback while I was at it, not because it was necessary (since that expression doesn't evaluate until after init completes), but because A) it is easier to just have the rule be to not use `this.blah` in that ctor, and B) there is no reason to use indirection through `this` there, since both `this` and `results` are eagerly bound in the `=>` lambda, so it couldn't possibly evaluate to anything else.

RedBeard0531:
I was slightly concerned that the issue with `defineProperties` was caused by hitting some perverse case when redefining properties, potentially in a different order, so I tried going back to using `defineProperties`, but with `declare prop`. The results show that both changes make sense, both separately and together (at least on node 18).

```
      ✓ reads bool?[] (116,515 ops/sec, ±34.78%)
      ✓ reads bool?<> (158,146 ops/sec, ±22.99%)
      ✓ reads bool?{} (260,956 ops/sec, ±22.42%)
```

The best results are still with `Object.defineProperty` + `declare prop`. I'll try testing with hermes soon.

RedBeard0531:
Ok, after much struggle, I now have the hermes numbers, and they are quite _interesting_, but don't suggest any changes to this PR.

Summary:
* Hermes doesn't seem to care much about `declare prop` vs `prop!`. Perf difference on that axis is within the run-to-run noise range. It is possible that it literally makes *no* difference since something in the metro/hermes chain just completely ignores it (my unconfirmed theory)
* Hermes might slightly prefer `declareProperty` for normal properties, but the difference is small and it is hard to be sure it isn't noise.
* Hermes *really* hates `declareProperties` for `Symbol` properties. They are used by Dictionary, and I think that is why there is a huge improvement in this PR for the `bool?{}` case.

`prop!` + `defineProperties` (baseline):
```
      ✓ reads bool?[] (39,783 ops/sec, ±18.25%)
      ✓ reads bool?<> (45,996 ops/sec, ±16.59%)
      ✓ reads bool?{} (51,490 ops/sec, ±16.03%)
```


`prop!` + `defineProperty` (patch 1):
```
      ✓ reads bool?[] (41,343 ops/sec, ±16.7%)
      ✓ reads bool?<> (48,760 ops/sec, ±16.66%)
      ✓ reads bool?{} (82,426 ops/sec, ±19.77%)
```

`declare prop` + `defineProperties`:
```
      ✓ reads bool?[] (40,020 ops/sec, ±14.8%)
      ✓ reads bool?<> (47,643 ops/sec, ±14.1%)
      ✓ reads bool?{} (53,016 ops/sec, ±15.75%)
```

`declare prop` + `defineProperty` (full PR):
```
      ✓ reads bool?[] (41,139 ops/sec, ±20.84%)
      ✓ reads bool?<> (48,287 ops/sec, ±17.62%)
      ✓ reads bool?{} (82,485 ops/sec, ±22.28%)
```


I would really rather not collect all of these numbers for JSC. It would take a long time, and I suspect that we wouldn't change any decisions based on them.

PR #5838, opened 2023-05-23
Update `importAppBefore` to take an AppConfiguration

kraenhansen:
## What, How & Why?

Quick change to have `importAppBefore` take an SDK-level `AppConfiguration` instead of a log level (which can be set via the mocha remote context).


takameyer (approved):
Nothing is using the `logLevel` argument, as it was only there for debugging purposes.  We used to have it activated in the CI so that we could try and report any failures with logs to the sync team.  I think an environment variable is a cleaner and more flexible option, so good improvement here.

PR #5848, opened 2023-05-26
Update RN install test to pass the `--new-architecture` runtime option

kraenhansen:
## What, How & Why?

It seems we didn't pass the new architecture matrix variable correctly to the CLI when initialising the install test app.


kraenhansen:
This is a test run showing that enabling new architecture for `realm@latest` is indeed broken: https://github.com/realm/realm-js/actions/runs/5089041790

kneth (approved):

kneth:
I have approved the PR as we want to test against the new architecture. The question is why we see `Missing Realm constructor`.

kraenhansen:
That's probably just a symptom of the native .so not loading correctly.

Issue #5737, opened 2023-04-17
Does realm support geo query such as distance, bounds etc.


somasekharkakarla:
### Problem

How to query a geo query in realm react native? i need to filter by distance, radius and bounds 

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

Local Database only

kneth:
@somasekharkakarla The short answer is "currently Realm doesn't support geospatial queries". The longer answer is that it is in progress (see https://github.com/realm/realm-core/pull/6352), but we can't give a timeline.

somasekharkakarla:
Hi @kneth 
Use case
We have geotagged items that are dense within a specific region. We want to query/filter the items within a boundary (polygon/multipolygon feature) and return the items within the boundary without implementing the filter in the code with realtime sync.

robertmiles3:
I'm currently starting an app (Realm, Swift, Device Sync) that will have GPS coordinates (points) for items in the DB and will want to query this. Given that this support is in the pipeline, what's the suggested schema for storing the coordinates?

papafe:
Hi @robertmiles3,
Actually there is a PR open that is adding initial support for geospatial queries: https://github.com/realm/realm-js/pull/5850

To make the story short, in order to store a point you need to create an embedded object that adheres to `CanonicalGeoPoint` (object with two properties: coordinates, that is an array of two or three elements, and type that needs to be "Point"):

```typescript
/**
 * Represents the array coordinates of a point. The first two required elements of the array are longitude and latitude.
 * The third and optional element is altitude, that is currently ignored in the geospatial queries calculations.
 */
export type GeoPosition = [number, number] | [number, number, number];

/**
 * Interface that satisfies the geoJSON specification for a point.
 * Any embedded object that adhere to this interface can be queried in geospatial queries.
 * Additionally, this can be used as one of the possible forms of {@link GeoPoint}.
 */
export interface CanonicalGeoPoint {
  coordinates: GeoPosition;
  type: "Point";
}
```
For example:

```typescript
class MyGeoPoint implements CanonicalGeoPoint {
  coordinates: GeoPosition;
  type = "Point" as const;

  constructor(long: number, lat: number) {
    this.coordinates = [long, lat];
  }

  static schema: ObjectSchema = {
    name: "MyGeoPoint",
    embedded: true,
    properties: {
      type: "string",
      coordinates: "double[]",
    },
  };
}

interface IPointOfInterest {
  id: number;
[...]

PR #5850, opened 2023-05-26
Add support for geospatial queries

### LINKED ISSUE #3684: Geospatial Queries
## Goals

Sort or filter results by distance from user's current location.

## Expected Results

Be able to calculate a distance from the current lat/long to the lat/long on the record in question, or at least filter by those within X distance from a specified point. See [MongoDB's Geospatial Queries](https://docs.mongodb.com/manual/geospatial-queries/).

papafe:
This closes #3684 and #5737

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


papafe:
@elle-j @kneth @kraenhansen @RedBeard0531 @takameyer This is a very rough PR to see if what I'm doing kinda makes sense, before I go too much forward.

packages/realm/src/TypeHelpers.ts
```diff
@@ -110,6 +113,10 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
   }
 }
 
+function isGeoCircle(value: any): value is GeoCircle {
+  return "distance" in value && "center" in value;
```
kraenhansen, line 117: I think we'd want to check the type of these too?

integration-tests/tests/src/tests/queries.ts
```diff
@@ -150,6 +197,92 @@ const expectQueryResultValues = (
...
+      this.realm.write(() => {
+        this.realm.create("PointOfInterest", zero);
+        this.realm.create("PointOfInterest", poiA);
+        this.realm.create("PointOfInterest", poiB);
+        this.realm.create("PointOfInterest", poiC);
+        this.realm.create("PointOfInterest", poiD);
+      });
+    });
+
+    it.only("GeoCircle basic", function (this: RealmContext) {
```
elle-j, line 269:
```suggestion
    it("GeoCircle basic", function (this: RealmContext) {
```

packages/realm/src/binding.ts
```diff
@@ -50,6 +66,23 @@ declare module "realm/binding" {
...
 }
 
+Geospatial.fromCircle = function (c: ExternalGeoCircle): Geospatial {
+  return Geospatial.makeFromCircle({
+    center: convertToBindingGeoPoint(c.center),
+    radiusRadians: c.distance,
+  });
+};
+
+function convertToBindingGeoPoint(p: ExternalGeoPoint): BindingGeoPoint {
```
elle-j, line 76: Small thing, but our functions for converting to/from a binding type are usually just called `toBinding<Type>` (e.g. `toBindingGeoPoint`) 🙂

integration-tests/tests/src/tests/queries.ts
papafe: Oh yes, this is only for now 😁 
Just a fast draft

integration-tests/tests/src/tests/queries.ts
```diff
@@ -82,6 +92,43 @@ class Story extends Realm.Object<Story> implements IStory {
   };
 }
 
+class MyGeoPoint implements IGeoPoint {
```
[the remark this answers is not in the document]
papafe: If I make this extend RealmObject, then the constructor needs to call RealmObject's constructor, that will add it to the Realm, but I don't want that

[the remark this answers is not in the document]
elle-j: Okay, so prob related to the current  inconvenience of `new` combined with being an embedded object?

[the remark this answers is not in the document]
papafe: Yep, otherwise I can't create points before I need them

takameyer, line 95: Just an idea, but I think it would be better if we provided classes that the developer could import and use in their schema, such as GeoPoint, GeoCircle, etc.  The structure of these embedded classes are very specific and could possibly be messed up (ie the `type` string).

Then the developer could just import the class, add it to their `schema` config and use the geo type in other schemas.

takameyer: The interface is a good start, since it will show a type error if `type` is not set to the correct string.  I hadn't read the last part of the tech design, but now I see that the types will be coming in core.  Hopefully the migration from this to having an internal type is not a large lift for our users.

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface IGeoPoint {
+  coordinates: IGeoPosition;
+  type: "Point";
+}
+
+export type GeoPoint =
+  //This is compatible with GeoLocationCoordinates (https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates)
+  | {
+      latitude: number;
+      longitude: number;
```
RedBeard0531, line 38:
```suggestion
      longitude: number;
      altitude?: number;
```

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type GeoCircle = {
+  center: GeoPoint;
+  distance: number;
+};
+
+//Utility class for distance conversions to radians
+export class Distance {
+  static earthRadiusKm = 6378.1;
+  static milesPerKm = 1.609344;
```
RedBeard0531, line 64: Why not give `earthRadiusMiles = 3963.2` rather than the conversion from km to miles. Also this appears to be based on the "equatorial radius" from https://en.wikipedia.org/wiki/Earth. A) I have no idea if we should use that or the mean radius, B) Why does this have truncated precision?

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,72 @@
...
+  topRight: GeoPoint;
+};
+
+export type GeoCircle = {
+  center: GeoPoint;
+  distance: number;
+};
+
+//Utility class for distance conversions to radians
+export class Distance {
```
RedBeard0531, line 62: I think this is essentially a namespace since you never actually want to construct a `Distance` instance. Maybe just export free functions `kmFromRadians`/`radiansFromKm` (and mile equivalents) rather than exporting like this?

packages/realm/src/TypeHelpers.ts
```diff
@@ -110,6 +120,61 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
   }
 }
 
+//TODO NEED TO CHANGE ALSO WHAT HAPPENS ON JSI (NOT ONLY NODE)
```
RedBeard0531, line 123: Don't forget!

Also, this comment seems misplaced here. It probably belongs on node.ts. But probably better to just implement it in jsi.ts and delete this comment rather than moving it.

packages/realm/src/TypeHelpers.ts
```diff
@@ -98,6 +102,12 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
     throw new Error(`Using a ${value.constructor.name} as Mixed value, is not yet supported`);
   } else if (Array.isArray(value)) {
     throw new TypeError("A mixed property cannot contain an array of values.");
+  } else if (isGeoCircle(value)) {
+    return binding.Geospatial.fromCircle(value);
+  } else if (isGeoBox(value)) {
+    return binding.Geospatial.fromBox(value);
+  } else if (isGeoPolygon(value)) {
+    return binding.Geospatial.fromPolygon(value);
```
RedBeard0531, line 110: For now, this feels a bit like the wrong place for this. I don't think we want to accept geo stuff every place where we accept `Mixed`. Instead, lets add this logic specifically in the query args API.

Also, I wouldn't frame it as an independent set of if/else if `ifFoo` checks with their own duplicated checks. Instead, lets properly define what we will look for when making the decision.

I suggest something like:

```ts
if (typeof(value) !== 'object') return Mixed()
if (Array.isArray(val)) return Point()
if ('type' in value) {
    if (value.type === 'Point') return Point()
    if (value.type === 'Polygon' return Polygon()
}
if ('latitude' in value) return Point()
if ('topRight' in value) return GeoBox()
if ('center' in value) return GeoCircle()
```

Alternatively, we could use classes with constructors and require the user to do `new GeoPoint(flexibleArg)` and just have a chain of `if (value instanceof GeoPoint)` style checks.

packages/realm/src/TypeHelpers.ts
```diff
@@ -110,6 +120,61 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
   }
 }
 
+//TODO NEED TO CHANGE ALSO WHAT HAPPENS ON JSI (NOT ONLY NODE)
+
+//Here I am trying to strike a balance between completeness and performance.
+//I think if we have to check everything (especially for polygon), all those type assertions will become very complex,
+//and probably also inefficient.
```
RedBeard0531, line 127: I mean, we already sort of will be doing the assertion when converting to the binding, so the perf concerns are a bit moot. But also, see above for why I don't think you want these functions anyway.

packages/realm/src/binding.ts
```diff
@@ -16,7 +16,27 @@
...
 
-import { IndexSet, Int64, ObjKey, SyncSession, Timestamp, WeakSyncSession } from "realm/binding";
+import {
+  GeoBox as ExternalGeoBox,
+  GeoCircle as ExternalGeoCircle,
+  GeoPoint as ExternalGeoPoint,
+  GeoPolygon as ExternalGeoPolygon,
+  IGeoPoint,
+  IGeoPosition,
+} from "./internal";
```
RedBeard0531, line 26: We can't have this import from `internal`. At the very least, make this an `import type` so that it has no runtime impact. But really this feels like the wrong place for this logic. At least in my opinion, this file should only depend on the binding and add some simpler helpers that are easiest to implement in JS/TS and don't depend on the rest of the SDK. For stuff like this, I think we want the normal `toBinding`-style functions that live in the SDK layer rather than in the binding layer.

packages/realm/src/binding.ts
```diff
@@ -50,6 +77,47 @@ declare module "realm/binding" {
...
+  return Geospatial.makeFromPolygon({
+    points,
+  });
+};
+
+function toBindingGeoPoint(p: ExternalGeoPoint): BindingGeoPoint {
+  if (Array.isArray(p)) {
+    return { longitude: p[0], latitude: p[1] };
+  } else if ("type" in p) {
+    return { longitude: p.coordinates[0], latitude: p.coordinates[1] };
```
RedBeard0531, line 111:
```suggestion
  if (Array.isArray(p)) {
    return { longitude: p[0], latitude: p[1], altitude: p[2] };
  } else if ("type" in p) {
    return { longitude: p.coordinates[0], latitude: p.coordinates[1], altitude: p.coordinates[2] };
```

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,72 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+export type IGeoPosition = [number, number]; //long/lat, so it's the same order as geoJSON
```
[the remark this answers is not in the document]
papafe: At the moment the altitude is ignored, so I wasn't sure if we should consider it yet or not.

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,72 @@
...
+
+//Interface that satisfies the geoJSON specification for a polygon.
+export interface IGeoPolygon {
```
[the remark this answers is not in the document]
papafe: Agree with the naming.
Regarding exporting, I think we should at least export `IGeoPoint`, because developers are expected to make their objects conform to this interface in order for queries to work. 
Regarding `IGeoPolygon`, I'm not sure if that's necessary, but probably we could keep it for completeness.

packages/realm/src/GeoSpatial.ts
papafe: Yep, we definitely can simplify the miles->radian calculation. 
Regarding the earth radius, it's coming from core (where it's said to be coming directly from src/mongo/db/geo/geoconstants.h)

packages/realm/src/binding.ts
papafe: I think it makes sense. I'll move this to `GeoSpatial.ts`

packages/realm/src/TypeHelpers.ts
papafe: Regarding where to put this code... It seems that `mixedToBinding` is used anyway only for the query arguments. 
If at some point this method is going to be used also for something else, then we could add a boolean input that would skip the geospatial discriminators

papafe: Regarding the duplicated checks... I think the only duplicated one for now is `typeof value === object and value !== null`, as we can't have GeoPoints as arguments.
I will move those two checks outside of the type predicates.

packages/realm/src/GeoSpatial.ts
papafe: Yep

papafe:
@kneth @kraenhansen @elle-j @takameyer This PR is ready for review. Some notes:
- I am not sure what is the status about exporting types, especially considering that we didn't have those geo types before
- There are a couple of tests that are skipped because there are going to be some small changes in core, but it should not make much of a difference
- I am open to suggestions regarding the naming of the new types, but don't get too crazy with that 😁

packages/realm/src/TypeHelpers.ts
```diff
@@ -99,6 +106,15 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
   } else if (Array.isArray(value)) {
     throw new TypeError("A mixed property cannot contain an array of values.");
   } else {
+    if (typeof value == "object" && value !== null) {
```
kneth, line 109:
```suggestion
    if (typeof value === "object" && value !== null) {
```

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,164 @@
...
+  } else {
+    return p;
+  }
+}
+
+function toBindingGeoPointArray(arr: GeoPoint[][]): binding.GeoPoint_Relaxed[][] {
+  return arr.map((ring) => ring.map((p) => toBindingGeoPoint(p)));
+}
+
+const earthRadiusKm = 6378.1;
```
kneth, line 145: Is it common to use International Astronomical Union's value and not the one provided by International Union of Geodesy and Geophysics?

kneth (approved):

packages/realm/src/GeoSpatial.ts
papafe: Maybe 👀

integration-tests/tests/src/tests/queries.ts
```diff
@@ -139,9 +190,9 @@ const expectQueryResultValues = (
   queryResultPairs.forEach(([expectedResults, queryString, ...queryArgs]) => {
     let results = realm.objects<any>(objectSchema);
     results = results.filtered(queryString, ...queryArgs);
-    expect(expectedResults.length).equals(results.length);
-    expect(expectedResults).to.deep.equal(
-      results.map((el) => el[propertyToCompare]),
+    expect(results.length).to.equal(expectedResults.length);
```
papafe, line 193: This uses the correct order of arguments, so that the text that is printed to console from the test is correct ("expected" vs "result")

packages/realm/src/TypeHelpers.ts
```diff
@@ -109,6 +125,20 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
...
+function isGeoBox(value: object): value is GeoBox {
+  return "bottomLeft" in value && "topRight" in value;
+}
+
+function isGeoPolygon(value: object): value is GeoPolygon {
+  return (
+    ("type" in value && value["type"] === "Polygon" && "coordinates" in value && Array.isArray(value["coordinates"])) ||
+    ("outerRing" in value && Array.isArray(value["outerRing"]))
+  );
+}
```
elle-j, line 141: Regarding `isGeoCircle()`, `isGeoBox()`, and `isGeoPolygon()`, for the `GeoPoint` properties, would it make sense to create a `isGeoPoint()` function to that each of the aforementioned functions can call to validate those values? Currently, it only checks that the field is defined (e.g. `"center" in value`).

elle-j (approved):
Great to finally be able to leverage one of the really awesome features of MongoDB! Nice work!

packages/realm/bindgen/src/templates/jsi.ts
```diff
@@ -933,6 +933,9 @@ class JsiCppDecls extends CppDecls {
                       }`,
                   )
                   .join(" ")
+              } else if (obj.instanceOf(_env,  (*RealmAddon::self->m_cls_Geospatial_ctor))) {
+                //This needs his own case because the constructor of Mixed for Geospatial requires a pointer
```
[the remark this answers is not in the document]
papafe: This thing still bites me in the butt after many years speaking english. In italian we only have masculine and feminine, no "neutral" form 😁

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  return binding.Geospatial.makeFromPolygon({
+    points,
+  });
+}
+
+function toBindingGeoPoint(p: GeoPoint): binding.GeoPoint_Relaxed {
+  if (Array.isArray(p)) {
+    return { longitude: p[0], latitude: p[1], altitude: p[2] };
```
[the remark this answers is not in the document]
papafe: Correct

CHANGELOG.md
```diff
@@ -37,6 +37,77 @@
...
+  location: new MyGeoPoint(12.558892784045568, 55.66717839648401),
+};
+
+const newYork: IPointOfInterest = {
+  id: 2,
+  location: new MyGeoPoint(-73.92474936213434, 40.700090994927415),
+};
+
+realm.create("PointOfInterest", copenhagen);
+realm.create("PointOfInterest", newYork);
```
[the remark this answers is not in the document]
papafe: You're right

integration-tests/tests/src/tests/queries.ts
```diff
@@ -150,6 +201,703 @@ const expectQueryResultValues = (
...
+        };
+
+        this.realm.write(() => {
+          this.realm.create("PointOfInterest", magasasaKodbyen);
+          this.realm.create("PointOfInterest", slurpRamen);
+        });
+
+        let circle: GeoCircle = {
+          center: norreport,
+          distance: kmToRadians(0.5), //500 meters radius
```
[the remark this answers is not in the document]
papafe: Probably you're right

packages/realm/src/TypeHelpers.ts
papafe: I had this function before (and I can retrieve it), but I was thinking that maybe it could be too much to be so strict? ("too much" from the performance point of view)

packages/realm/src/TypeHelpers.ts
```diff
@@ -109,6 +125,20 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
     return value as binding.MixedArg;
   }
 }
+function isGeoCircle(value: object): value is GeoCircle {
+  return "distance" in value && "center" in value && typeof value["distance"] == "number";
+}
```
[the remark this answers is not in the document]
papafe: What is the difference between object and record in this case?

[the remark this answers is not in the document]
elle-j: `object` is the JS type which are those values returning `true` for `value === "object"`, which includes e.g. arrays. `Record` uses the TS type which is the regular "dictionary"/"hash map" containing properties.

packages/realm/src/GeoSpatial.ts
```diff
@@ -0,0 +1,164 @@
...
+  | {
+      latitude: number;
+      longitude: number;
+      altitude?: number;
+    }
+  | CanonicalGeoPoint
+  | GeoPosition;
+
+/**
+ * Represents a circle in spherical geometry that can be used for the definition of geospatial queries.
```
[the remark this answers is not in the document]
elle-j: Yeah 😛 Wasn't completely sure about the exact meaning of "the definition of geospatial queries".

packages/realm/src/TypeHelpers.ts
elle-j: I guess it depends on what behavior Core will have or the errors it will generate if you pass along incorrect types. See what happens with some faulty types 🙂

papafe: This is a problem for us, not for them, because core requests that we create certain classes. So what will happen is that at some point during the conversion there will be a js exception popping up

CHANGELOG.md
```diff
@@ -37,6 +37,77 @@
...
+    primaryKey: "id",
+  };
+}
+
+const berlinCoordinates = new MyGeoPoint(13.397255909303222, 52.51174463251085);
+const radius = kmToRadians(500); //500 kilometers radius, equal to 0.0783932519 radians
+
+const copenhagen: IPointOfInterest = {
+  id: 1,
+  location: new MyGeoPoint(12.558892784045568, 55.66717839648401),
```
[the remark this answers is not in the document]
papafe: I agree with your point. I'll keep it like this for the tests because it's easier to work like this, but I'll make `MyGeoPoint` derive from `Realm.Object`, as I think this is what we want to suggest going forward

packages/realm/src/GeoSpatial.ts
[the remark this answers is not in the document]
papafe: Maybe "that can be used as an argument for geospatial queries"

packages/realm/src/TypeHelpers.ts
```diff
@@ -109,6 +125,20 @@ export function mixedToBinding(realm: binding.Realm, value: unknown): binding.Mi
...
+
+function isGeoBox(value: object): value is GeoBox {
+  return "bottomLeft" in value && "topRight" in value;
+}
+
+function isGeoPolygon(value: object): value is GeoPolygon {
```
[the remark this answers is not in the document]
papafe: Not all form of GeoPolygon are assignable to `Record`, so I'll keep it as `object`

CHANGELOG.md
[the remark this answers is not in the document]
papafe: Regarding the example, if I make `MyGeoPoint` derive from `Realm.Object`, would there be an easy way of constructing the object before adding it to Realm. Because like this we need to do something like:

```typescript
    const poi: IPointOfInterest = {
      id: 5,
      location: {
        coordinates: [0, 0],
        type: "Point",
      },
    };
```
What I don't like about this is that one needs to specify `Point`, but I suppose there is no way around it.

CHANGELOG.md
```diff
@@ -37,6 +37,79 @@
   // ...
   peopleOver20.unsubscribe();
   ```
+<<<<<<< HEAD
```
elle-j, line 40:
```suggestion
```

CHANGELOG.md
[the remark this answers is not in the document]
elle-j: I'm unsure of which approach we want to use in the examples for the current state of the SDK (`new` or literals). Perhaps @realm/realm-js-team could chime in. But if we do literals, then yeah it'd look like the snippet you just provided.  You could also see if adding `as MyGeoPoint` after the `location` object helps.
```ts
      location: {
        coordinates: [0, 0],
        type: "Point",
      } as MyGeoPoint,
```

packages/realm/src/GeoSpatial.ts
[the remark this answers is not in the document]
elle-j: From the [RQL docs](https://www.mongodb.com/docs/realm/realm-query-language/#expressions), the term `value` could probably be applied to this.

CHANGELOG.md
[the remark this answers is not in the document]
papafe: I just corrected this as we discussed, let's see if the others think some other solution ;)

takameyer (approved):
Looks good!  Can we get a core update first?

integration-tests/tests/src/tests/queries.ts
```diff
@@ -150,6 +201,703 @@ const expectQueryResultValues = (
...
+
+    beforeEach(function (this: RealmContext) {
+      this.realm.write(() => {
+        this.realm.create(PointOfInterest, zero);
+        this.realm.create(PointOfInterest, poiA);
+        this.realm.create(PointOfInterest, poiB);
+        this.realm.create(PointOfInterest, poiC);
+        this.realm.create(PointOfInterest, poiD);
+        this.realm.create(PointOfInterest, northPole);
+        //this.realm.create("PointOfInterest", invalid);  //TODO Need to re-enable fix in core (https://github.com/realm/realm-core/pull/6719)
```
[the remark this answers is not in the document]
papafe: It's not yet in a release, so I think we can wait for now

kraenhansen (approved):

kraenhansen:
It seems this introduce a few eslint warnings, but IMO this is safe to merge.

PR #5853, opened 2023-05-26
CommonJS bundle (without Rollup)

### LINKED ISSUE #5857: Export only a CommonJS entrypoint

kraenhansen:
## What, How & Why?

This fixed #5857 by removing rollup from the SDK package in favour of a [TypeScript projects](https://www.typescriptlang.org/docs/handbook/project-references.html) based setup which emits only CommonJS code.

This is aligned with recent discussions on the team to ensure v12.0.0 exports exactly as v11 did, to keep friction from migrating at an absolute minimum.

I ran into several problems of which some I managed to find workarounds for and others I didn't:
- The Typescript compiler would emit `.mjs` files even when `module` was set to `commonjs` which would break the package upon import. I filed [an issue](https://github.com/microsoft/TypeScript/issues/54573) for this, [created a PR](https://github.com/microsoft/TypeScript/pull/54583) and found [a temporary workaround](https://github.com/realm/realm-js/blob/kh/cjs-bundle/packages/realm/scripts/fix-dist-filenames.ts).
- Although [the Jest issue is closed](https://github.com/jestjs/jest/issues/9771), I found problems from Jest lacking support of export conditions out of the box and had to implement a custom resolver. I later transitioned into exporting via a separate package in an attempt to work around this.
- I was planning to rely on `bundleDependencies` to pull @realm/network-transport (and a new `@realm/binding` package) into the SDK's .tar.gz before uploading it to NPM. But this is not supported when using NPM workspaces (yet) :meow_sad: https://github.com/npm/cli/issues/3466


kraenhansen:
Woups - looks like I need to find a solution for re-enabling test coverage 🤔

elle-j (approved):

packages/realm/bindgen/CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ file(GLOB_RECURSE SDK_TS_FILES
 list(FILTER SDK_TS_FILES EXCLUDE REGEX "templates/[^/]*\.ts$")
 
 set(JS_SPEC_FILE ${SDK_DIR}/bindgen/js_spec.yml)
-set(TYPESCRIPT_OUTPUT_DIR ${SDK_DIR}/generated/ts)
+set(TYPESCRIPT_OUTPUT_DIR ${SDK_DIR}/generated)
```
RedBeard0531, line 128: Shouldn't this be `src/generated`?

packages/realm/src/Realm.ts
```diff
@@ -1433,6 +1437,7 @@ export declare namespace Realm {
     CollectionChangeSet,
     /** @deprecated Please use named imports */
     CollectionType as Collection,
+    CompensatingWriteErrorType,
```
RedBeard0531, line 1440:
```suggestion
    CompensatingWriteErrorType as CompensatingWriteError,
```

packages/realm/src/cjs.ts
RedBeard0531, line 1: What is with mixing the `index.cjs.ts` and `cjs.ts` naming conventions? Can we pick one. Also do we need both index.cjs.ts and index.ts? or could they be combined?

integration-tests/tests/package.json
```diff
@@ -20,35 +20,35 @@
...
-				"../../packages/realm:bundle",
+				"../../packages/realm:build:ts",
 				"build-dependencies"
 			]
 		},
 		"coverage": {
 			"command": "nyc mocha --exit",
 			"dependencies": [
-				"../../packages/realm:bundle:coverage",
+				"../../packages/realm:build:ts:coverage",
```
takameyer, line 44: I think this is missing from `realm/package.json`

takameyer (approved):

.gitignore
```diff
@@ -113,6 +113,7 @@ coverage/
 # Generated artifacts
 /packages/realm/bindgen/vendor/bindgen-lib/generated/
 /packages/realm/generated/
+/packages/realm/src/generated/
```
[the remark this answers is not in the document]
kraenhansen: I needed to move `core.ts` into `src` when adopting TypeScript projects, because of:

> error TS6059: File '/Users/kraen.hansen/Projects/realm-js/packages/realm/generated/core.ts' is not under 'rootDir' '/Users/kraen.hansen/Projects/realm-js/packages/realm/src'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Matched by include pattern '../generated/core.ts' in '/Users/kraen.hansen/Projects/realm-js/packages/realm/src/tsconfig.json'

And I don't really consider the wrapper .mjs files and realm.node file as "src" as it's not TS and won't be moved to "dist" by tsc.

packages/realm/src/platform/node/tsconfig.json
```diff
@@ -1,18 +1,15 @@
 {
-  "extends": "../../../tsconfig.json",
+  "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
+    "outDir": "../../../dist/platform/node",
     "types": ["node"],
-    "noResolve": false,
-    "incremental": true,
```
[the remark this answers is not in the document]
kraenhansen: As I understand it, it's default for a `composite` config.

packages/realm/src/cjs.ts
kraenhansen: > do we need both index.cjs.ts and index.ts? or could they be combined?
I wanted the ensure the general code-base was ESM and the `cjs.ts` was only to export an entrypoint for the types. I'll align the naming though 👍

integration-tests/tests/package.json
kraenhansen: True - that was half baked 🙃

.gitignore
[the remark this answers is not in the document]
RedBeard0531: Hmm. I'm not sure we have a clear rule for what `src/` is. Is it where we put the "source code" for the realm package, or is it only the input for `tsc`? I had assumed the former, but it seems like you intend for it to be the later. I guess I don't feel too strongly either way, but we should be clear what our rule is.

packages/realm/src/cjs.ts
RedBeard0531: (FYI in markdown you need a blank line after the last quoted line in order to end a quote and start your reply)

> I wanted the ensure the general code-base was ESM and the cjs.ts was only to export an entrypoint for the types. I'll align the naming though

We still could get rid of index.ts and move its minimal contents to index.cjs.ts. It just seems weird to have both with the same root name. If you *really* want to keep as much as possible out of cjs files, it may be best to rename index.ts to reexport.ts or umbrella.ts.

packages/realm/bindgen/CMakeLists.txt
```diff
@@ -146,7 +147,7 @@ set(GENERATED_JS_FILES
 bindgen(
     TEMPLATE ${SDK_DIR}/bindgen/src/templates/node-wrapper.ts
     OUTPUTS ${GENERATED_JS_FILES}
-    OUTDIR ${TYPESCRIPT_OUTPUT_DIR}
+    OUTDIR ${GENERATED_OUTPUT_DIR}
```
RedBeard0531, line 150: I think you meant to change lines 144/145 instead. Right now there are out of sync.

packages/realm/bindgen/CMakeLists.txt
```diff
@@ -125,10 +125,11 @@ file(GLOB_RECURSE SDK_TS_FILES
...
 
 set(JS_SPEC_FILE ${SDK_DIR}/bindgen/js_spec.yml)
-set(TYPESCRIPT_OUTPUT_DIR ${SDK_DIR}/generated)
+set(GENERATED_OUTPUT_DIR ${SDK_DIR}/generated)
+set(TYPESCRIPT_OUTPUT_DIR ${SDK_DIR}/src/generated)
 
 set(GENERATED_TS_FILES
-    ${TYPESCRIPT_OUTPUT_DIR}/native.d.mts
+    ${GENERATED_OUTPUT_DIR}/native.d.mts
     ${TYPESCRIPT_OUTPUT_DIR}/core.ts
```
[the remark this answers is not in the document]
kraenhansen: True - I was missing a replace here 👍

takameyer:
I'm guessing we are going to close this, or?

kraenhansen:
@takameyer see my previous comment 🙂

kraenhansen:
Closing this as I've created https://github.com/realm/realm-js/issues/5894 to eventually do this migration.

PR #5866, opened 2023-06-02
OrderedCollection no longer requires a Results

RedBeard0531:
Previously it always held a `Results` object, even for the `List` and `Set` subclasses. That was expensive to construct, and possibly changed the runtime complexity. Now it will directly work with whatever native object it's subclass represents. This required a small change to the core spec file to give Results and Collections a compatible API.

Before
```
Node:
      ✓ reads bool?[] (129,694 ops/sec, ±48.46%)
      ✓ reads bool?<> (178,647 ops/sec, ±31.71%)
Hermes:
      ✓ reads bool?[] (43,417 ops/sec, ±11.36%)
      ✓ reads bool?<> (48,925 ops/sec, ±12.1%)
```

After
```
Node:
      ✓ reads bool?[] (211,569 ops/sec, ±34.5%)
      ✓ reads bool?<> (237,292 ops/sec, ±20.4%)
Hermes:
      ✓ reads bool?[] (56,075 ops/sec, ±13.68%)
      ✓ reads bool?<> (59,356 ops/sec, ±11.46%)
```
## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/OrderedCollection.ts
+++ b/packages/realm/src/OrderedCollection.ts
@@ -51,11 +51,6 @@ export type CollectionChangeCallback<T = unknown, EntryType extends [unknown, un
   changes: CollectionChangeSet,
 ) => void;
 
-/** @internal */
-export type OrderedCollectionHelpers = TypeHelpers & {
-  get(results: binding.Results, index: number): unknown;
-};
-
 const DEFAULT_PROPERTY_DESCRIPTOR: PropertyDescriptor = { configurable: true, enumerable: true, writable: true };
 const PROXY_HANDLER: ProxyHandler<OrderedCollection> = {
   // TODO: Consider executing the `parseInt` first to optimize for index access over accessing a member on the list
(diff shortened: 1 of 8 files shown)


packages/realm/src/OrderedCollection.ts
```diff
@@ -100,20 +95,28 @@ const PROXY_HANDLER: ProxyHandler<OrderedCollection> = {
...
+type OrderedCollectionInternal = binding.Results | binding.List | binding.Set;
+
 export abstract class OrderedCollection<T = unknown, EntryType extends [unknown, unknown] = [number, T]>
   extends Collection<number, T, EntryType, T, CollectionChangeCallback<T, EntryType>>
   implements Omit<ReadonlyArray<T>, "entries">
 {
   /** @internal */ protected declare realm: Realm;
-  /** @internal */ protected declare results: binding.Results;
-  /** @internal */ protected declare helpers: OrderedCollectionHelpers;
+  /** @internal */ protected declare abstract internal: OrderedCollectionInternal;
```
RedBeard0531, line 105: Alternatively, this could be made into a template param. I tried doing that, but ran into rollup/dts issues because it requires passing a `binding` type into a template parameter but `binding` is `@internal` so rollup gets confused. And I couldn't make rollup happy with removing `@internal` on `binding` because of the way we use `declare module` in `native.d.ts`. May be worth trying again once we jettison rollup if we think the template param route is clearer.

packages/realm/src/OrderedCollection.ts
```diff
@@ -132,21 +135,24 @@ export abstract class OrderedCollection<T = unknown, EntryType extends [unknown,
...
+    const itemType = internal.type; // Can probably get this passed in rather than dropping to C++.
+    const isObjectType =
+      (itemType & ~binding.PropertyType.Flags) === binding.PropertyType.Object ||
+      (itemType & ~binding.PropertyType.Flags) === binding.PropertyType.LinkingObjects;
     // Get the class helpers for later use, if available
-    const { objectType } = results;
-    const classHelpers = typeof objectType === "string" && objectType !== "" ? realm.getClassHelpers(objectType) : null;
+    const classHelpers = isObjectType && objectSchemaName ? realm.getClassHelpers(objectSchemaName) : null;
     // Make the internal properties non-enumerable
     Object.defineProperty(this, "realm", {
```
RedBeard0531, line 145: There is a large perf boost at least on node from not using `Object.defineProperty()` on these. Is it actually required? If if it is, it is likely faster to define a single property which is a plain object with all of these fields.

packages/realm/bindgen/vendor/realm-core
RedBeard0531, line 1: This is the change to make `List` and `Set` have the same shape as `Results`

RedBeard0531:
Unfortunately some of the test failures are due to https://github.com/realm/realm-core/issues/6695. We may need to convert to `Results` for some of these methods for now, or implement them in JS.

PR #5873, opened 2023-06-07
Update main to core v13.13.0

papafe:


takameyer (approved):

kneth:
Since Realm Core v13.13.0 comes with a change on how to build for Linux/ARM, I think we need to verify that the binaries are correct. Looking at the build log, I see many entries during linking like:

```
/usr/bin/ld.gold: warning: vendor/realm-core/openssl/lib/libcrypto.a(libdefault-lib-digest_to_nid.o): unknown program property type 0xc0010001 in .note.gnu.property section
```

This might be related to an earlier log entry:

```
-- Getting https://static.realm.io/downloads/dependencies/openssl/3.0.8/openssl-3.0.8-x64-linux-gnu.tar.gz...
```

Maybe it can be ignored, but I would like us to verify first.

elle-j (approved):

PR #5882, opened 2023-06-12
Only export a CJS bundle (using rollup)

kraenhansen:
## What, How & Why?

This supersedes https://github.com/realm/realm-js/pull/5853 and fixes https://github.com/realm/realm-js/issues/5857 by removing the ESM exports from the rollup config and adding CJS entrypoint files which export via `module.exports =` assignments as our v11 SDK does as well.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/enums.ts
+++ b/integration-tests/tests/src/tests/enums.ts
@@ -23,6 +23,7 @@ import {
   OpenRealmBehaviorType,
   OpenRealmTimeOutBehavior,
   SessionStopPolicy,
+  WaitForSync,
 } from "realm";
 
 describe("Enums", function () {
@@ -81,6 +82,15 @@ describe("Enums", function () {
         Error: 6,
         Fatal: 7,
         Off: 8,
+      });
+    });
+  });
+  describe("WaitForSync", function () {
+    it("is accessible", function () {
+      expect(WaitForSync).deep.equals({
+        Always: "always",
+        FirstTime: "first-time",
+        Never: "never",
       });
     });
   });
--- a/packages/realm-react/custom-resolver.cjs
+++ b/packages/realm-react/custom-resolver.cjs
@@ -1,0 +1,30 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+/* eslint-env node */
+
+module.exports = function (request, options) {
+  if (request === "react-native") {
+    // Ensure that "react-native" always resolve relative to the rootDir
+    options.basedir = options.rootDir;
+  } else if (request === "realm") {
+    // Ensure the node binding is used when testing with Jest on node
+    options.conditions = ["require", "default", "node"];
+  }
+  return options.defaultResolver(request, options);
+};
--- a/packages/realm/index.node.js
+++ b/packages/realm/index.node.js
@@ -1,6 +1,6 @@
 ////////////////////////////////////////////////////////////////////////////
 //
-// Copyright 2022 Realm Inc.
+// Copyright 2023 Realm Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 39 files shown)


packages/realm/index.node.js
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+/* eslint-disable @typescript-eslint/no-var-requires -- We're exporting using CJS assignment */
+/* eslint-env commonjs */
+
+module.exports = require("./dist/bundle.node").Realm;
```
kraenhansen, line 22: I don't love it, but it works 🙂

integration-tests/tests/src/tests/enums.ts
```diff
@@ -84,4 +85,13 @@ describe("Enums", function () {
       });
     });
   });
+  describe("WaitForSync", function () {
```
takameyer, line 88: Is this part of CJS?

packages/realm/src/Realm.ts
```diff
@@ -220,14 +223,17 @@ export class Realm {
   public static ClientResetMode = ClientResetMode;
   /** @deprecated Please use named imports */
   public static Collection = Collection;
+  public static CompensatingWriteError = CompensatingWriteError;
```
takameyer, line 226: Seems there have been more things added.  Is there a rebase missing?

takameyer (approved):
Can't wait for v13! 🧼

packages/realm/index.node.js
takameyer: At least we have proper entry points set for node and React Native.  Let's break everything in v13 💣

packages/realm/src/Realm.ts
kraenhansen: The tests are already accessing this via named imports, but since we're now simply exporting `module.exports = Realm` they need to be exposed as statics off the `Realm` class instead.

takameyer: Ok, thanks for the explanation

packages/realm-react/custom-resolver.cjs
```diff
@@ -0,0 +1,30 @@
...
+module.exports = function (request, options) {
+  if (request === "react-native") {
+    // Ensure that "react-native" always resolve relative to the rootDir
+    options.basedir = options.rootDir;
+  } else if (request === "realm") {
+    // Ensure the node binding is used when testing with Jest on node
+    options.conditions = ["require", "default", "node"];
+  }
+  return options.defaultResolver(request, options);
+};
```
kraenhansen, line 30: @takameyer to be honest, I don't understand why this worked before.

- The first branch will ensure "react-native" from `@realm/react` is loaded, instead of the peer dependency of `realm`(`realm/node_modules/react-native`), as that yielded `Invariant Violation: __fbBatchedBridgeConfig is not set, cannot invoke native modules` errors
- The second branch is to ensure the Node.js version of the SDK is loaded. If I didn't do this, I would be getting `TypeError: Cannot read properties of undefined (reading 'Helpers_get_table')` errors.

integration-tests/tests/src/tests/enums.ts
elle-j: This is me forgetting to add this enum test when doing the Flx Sync v2 PR 😉

packages/realm/rollup.config.mjs
```diff
@@ -106,7 +99,7 @@ export default [
   {
     input: "src/index.ts",
     output: {
-      file: mainExport.types,
+      file: "dist/bundle.d.ts",
```
elle-j, line 102: Is the `"./"` prefix needed?

elle-j (commented):
I believe these (in `packages/realm/src/tests`) also need to be updated:
* [list.test.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/list.test.ts)
* [milestone-4.test.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/milestone-4.test.ts)
* [milestone-5.test.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/milestone-5.test.ts)
* [platform.test.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/platform.test.ts)
* [schema-utils.test.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/schema-utils.test.ts)
* [updateschema.test.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/updateschema.test.ts)
* [utils.ts](https://github.com/realm/realm-js/blob/main/packages/realm/src/tests/utils.ts)

Other than that, LGTM 👍

packages/realm/rollup.config.mjs
kraenhansen: It doesn't seem like it 🙂

elle-j (approved):

PR #5885, opened 2023-06-13
Clean up TypeDoc implementation

krollins-mdb:
## What, How & Why?

The TypeDoc implementation from #5709 didn't do much with styling or customization. This PR overrides the default theme to come closer to the MongoDB look and feel.

Also, upgraded to the latest version of TypeDoc to fix some issues.

Unfortunately, I don't have a preview on a server. You'll need to check out my branch, build the docs, then preview locally.

To build the docs, navigate to `packages/realm-react` and then run `npm run docs`. A new docs directory will be added that contains the generated docs site.

### DIFF
--- a/packages/realm-react/typedoc.json
+++ b/packages/realm-react/typedoc.json
@@ -7,7 +7,7 @@
   "media": "./media",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  "favicon": "../../media/favicon.ico",
+  // "favicon": "../../media/favicon.ico",
   "visibilityFilters": {},
   "categorizeByGroup": false,
   "navigationLinks": {
--- a/packages/realm-react/typedoc.json
+++ b/packages/realm-react/typedoc.json
@@ -1,11 +1,16 @@
 {
+  "name": "Realm React",
   "entryPoints": ["src/index.tsx"],
   "tsconfig": "./tsconfig.json",
   "excludeInternal": true,
   "customCss": "../../typedoc/style.css",
   "media": "./media",
-  "name": "Realm React",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  "favicon": "../../media/favicon.ico"
+  "favicon": "../../media/favicon.ico",
+  "visibilityFilters": {},
+  "categorizeByGroup": false,
+  "navigationLinks": {
+    "GitHub": "https://github.com/realm/realm-js/tree/main/packages/realm-react#readme"
+  }
 }
--- a/packages/realm-react/typedoc.json
+++ b/packages/realm-react/typedoc.json
@@ -7,7 +7,6 @@
   "media": "./media",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  // "favicon": "../../media/favicon.ico",
   "visibilityFilters": {},
   "categorizeByGroup": false,
   "navigationLinks": {
--- a/packages/realm-web/typedoc.json
+++ b/packages/realm-web/typedoc.json
@@ -7,5 +7,4 @@
   "name": "Realm Web",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  // "favicon": "../../media/favicon.ico"
 }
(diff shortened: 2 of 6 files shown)


kneth:
Thanks @krollins-mdb but I have to admit that all this CSS is beyond me.

krollins-mdb:
Kræn mentioned some changes to how types are handled. Leaving this PR in draft mode until I have time to dig into those changes. I may end up closing this PR and creating a new one after I get that stuff sorted.

kraenhansen:
> Kræn mentioned some changes to how types are handled.

The main thing is: With the current configuration, [the entrypoint for the docs](https://github.com/realm/realm-js/blob/main/packages/realm/typedoc.json#L2) is the `src/index.ts` file. After the merge of https://github.com/realm/realm-js/pull/5882 this has changed, as we're no longer exporting a ESM and we simply export the `Realm` constructor via a CJS exports assignment (see [index.react-native.js](https://github.com/realm/realm-js/blob/main/packages/realm/index.react-native.js#L22), [index.node.js](https://github.com/realm/realm-js/blob/main/packages/realm/index.node.js#L22) and [types.d.cts](https://github.com/realm/realm-js/blob/main/packages/realm/types.d.cts#L19-L20)).

While this pattern of export was abundant in the past, typedoc isn't optimised for it and it results in quite a few drawbacks in the UX when the `types.d.cts` is picked as entrypoint. For one thing, the name of the `Realm` constructor doesn't seem to be picked up and since all the type exports are type aliases, they're visibly shown as links to the actual type declarations - which requires a click for users to get to the actual docs of a type. All of this is very unfortunate and something we hope to be able to resolve by transitioning to ESM.

In the meantime, I see a couple of ways forward:
- We could update the entrypoint to reflect reality and try to tweak and bend typedoc into submission. Try different configs, perhaps write a plugin, extend the template or _something_ to help improve the UX.
- We could keep the entry point pointing to an ESM file and not care too much about the correctness here. It will potentially confuse people that some types are not exported at the root level but only available in the nested namespace, when they're actually consumed. If we choose this route, we need to be even more careful that existing and future type exports are present both [on the `Realm` constructor](https://github.com/realm/realm-js/blob/main/packages/realm/src/Realm.ts#L1258-L1568) and [via named exports](https://github.com/realm/realm-js/blob/main/packages/realm/src/index.ts#L28-L182).

takameyer:
If we are just reviewing the look of the TypeDoc, I would say this is a huge improvement.  I think the type wrangling would probably be more on our turf to correct.  That being said I'm tempted to just merge this now and if we want to improve, open a new PR. Publishing something reasonable with the v12 release would be great.  We can always update it later.

packages/realm-react/typedoc.json
```diff
@@ -1,11 +1,16 @@
...
   "entryPoints": ["src/index.tsx"],
   "tsconfig": "./tsconfig.json",
   "excludeInternal": true,
   "customCss": "../../typedoc/style.css",
   "media": "./media",
-  "name": "Realm React",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  "favicon": "../../media/favicon.ico"
+  // "favicon": "../../media/favicon.ico",
```
takameyer, line 10: JSON doesn't support comments, so just delete it.
```suggestion
```

packages/realm-web/typedoc.json
```diff
@@ -7,5 +7,5 @@
   "name": "Realm Web",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  "favicon": "../../media/favicon.ico"
+  // "favicon": "../../media/favicon.ico"
```
takameyer, line 10:
```suggestion
```

packages/realm/typedoc.json
```diff
@@ -7,5 +7,5 @@
   "name": "Realm JavaScript",
   "includeVersion": true,
   "sourceLinkTemplate": "https://github.com/realm/realm-js/blob/{gitRevision}/{path}#L{line}",
-  "favicon": "../../media/favicon.ico"
+  // "favicon": "../../media/favicon.ico"
```
takameyer, line 10:
```suggestion
```

takameyer (approved):

takameyer:
@cla-bot check

takameyer:
I opened a new [PR](https://github.com/realm/realm-js/pull/5965) for this so that I can get something in shape for publishing.  Thanks for the original contribution!

takameyer:
@cla-bot check

krollins-mdb:
Thanks for taking over on this while I was on vacation, @takameyer!

PR #5890, opened 2023-06-15
Add `baseFilePath` to app config

### LINKED ISSUE #5777: Expose `baseFilePath` on the `App` configuration

kneth:
## What, How & Why?

This closes #5777 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/hooks/import-app-before.ts
+++ b/integration-tests/tests/src/hooks/import-app-before.ts
@@ -37,7 +37,7 @@ export type AppConfigurationRelaxed = {
 export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkConfig?: AppConfigurationRelaxed): void {
   // Unwrap when passed a builder directly
   if ("config" in config) {
-    return importAppBefore(config.config);
+    return importAppBefore(config.config, sdkConfig);
   }
 
   before(importAppBefore.name, async function (this: Partial<AppContext> & Mocha.Context) {
--- a/integration-tests/tests/src/hooks/import-app-before.ts
+++ b/integration-tests/tests/src/hooks/import-app-before.ts
@@ -26,7 +26,15 @@ const REALM_LOG_LEVELS = ["all", "trace", "debug", "detail", "info", "warn", "er
 
 const { syncLogLevel = "warn" } = environment;
 
-export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkConfig?: AppConfiguration): void {
+export type AppConfigurationRelaxed = {
+  id?: string;
+  baseUrl?: string;
+  timeout?: number;
+  multiplexSessions?: boolean;
+  baseFilePath?: string;
+};
+
+export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkConfig?: AppConfigurationRelaxed): void {
   // Unwrap when passed a builder directly
   if ("config" in config) {
     return importAppBefore(config.config);
@@ -39,7 +47,7 @@ export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkCo
       throw new Error("Unexpected app on context, use only one importAppBefore per test");
     } else {
       const { appId, baseUrl } = await importApp(config);
-      this.app = new Realm.App({ id: appId, baseUrl, baseFilePath: config?.baseFilePath, ...sdkConfig });
+      this.app = new Realm.App({ id: appId, baseUrl, ...sdkConfig });
 
       // Extract the sync database name from the config
       const databaseNames: string[] = config.services
--- a/integration-tests/tests/src/hooks/import-app-before.ts
+++ b/integration-tests/tests/src/hooks/import-app-before.ts
@@ -39,7 +39,7 @@ export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkCo
       throw new Error("Unexpected app on context, use only one importAppBefore per test");
     } else {
       const { appId, baseUrl } = await importApp(config);
-      this.app = new Realm.App({ id: appId, baseUrl, ...sdkConfig });
+      this.app = new Realm.App({ id: appId, baseUrl, baseFilePath: config?.baseFilePath, ...sdkConfig });
 
       // Extract the sync database name from the config
       const databaseNames: string[] = config.services
--- a/integration-tests/tests/src/node/path.ts
+++ b/integration-tests/tests/src/node/path.ts
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import { expect } from "chai";
-import Realm, { BSON } from "realm";
+import Realm, { BSON, SessionStopPolicy } from "realm";
 import path from "node:path";
 import os from "node:os";
 import { existsSync, rmSync } from "node:fs";
@@ -59,28 +59,29 @@ describe("path configuration (local)", function () {
   });
 });
 
-//describe.skipIf(environment.missingServer, `app configuration of root directory (flexible sync)`, async function () {
-describe.only(`app configuration of root directory (flexible sync)`, async function () {
+describe.skipIf(environment.missingServer, `app configuration of root directory (flexible sync)`, async function () {
+  // describe.only(`app configuration of root directory (flexible sync)`, async function () {
+  this.timeout(60_000);
   const tmpdir = getAbsolutePath();
   importAppBefore(buildAppConfig("with-flx").baseFilePath(tmpdir).anonAuth().flexibleSync());
   authenticateUserBefore();
 
   it("directory and file created where expected", async function () {
-    expect(existsSync(tmpdir)).to.be.false;
-
-    const realm = await Realm.open({
+    expect(existsSync(tmpdir)).to.be.true; // importAppBefore will create `tmpdir`
+    const realm = new Realm({
       schema: [FlexibleSchema],
       sync: {
+        // @ts-expect-error Using an internal API
+        _sessionStopPolicy: SessionStopPolicy.Immediately,
         flexible: true,
         user: this.user,
       },
-    });
+    } as Realm.Configuration);
 
     expect(existsSync(tmpdir)).to.be.true;
     expect(realm.path.startsWith(tmpdir));
 
     realm.close();
-    rmSync(tmpdir, { recursive: true });
   });
 });
 
--- a/integration-tests/tests/src/node/path.ts
+++ b/integration-tests/tests/src/node/path.ts
@@ -63,7 +63,7 @@ describe.skipIf(environment.missingServer, `app configuration of root directory 
   // describe.only(`app configuration of root directory (flexible sync)`, async function () {
   this.timeout(60_000);
   const tmpdir = getAbsolutePath();
-  importAppBefore(buildAppConfig("with-flx").baseFilePath(tmpdir).anonAuth().flexibleSync());
+  importAppBefore(buildAppConfig("with-flx").anonAuth().flexibleSync(), { baseFilePath: tmpdir });
   authenticateUserBefore();
 
   it("directory and file created where expected", async function () {
--- a/integration-tests/tests/src/node/path.ts
+++ b/integration-tests/tests/src/node/path.ts
@@ -20,6 +20,7 @@ import { expect } from "chai";
 import Realm, { BSON } from "realm";
 import path from "node:path";
 import os from "node:os";
+import { existsSync, rmSync } from "node:fs";
 
 import { importAppBefore, authenticateUserBefore } from "../hooks";
 import { buildAppConfig } from "../utils/build-app-config";
@@ -36,6 +37,7 @@ const schema = {
     value: "mixed?",
   },
 };
+const FlexibleSchema = { ...schema, properties: { ...schema.properties, nonQueryable: "string?" } };
 
 describe("path configuration (local)", function () {
   it("relative path", function () {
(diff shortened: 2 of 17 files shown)


packages/realm/bindgen/js_spec.yml
```diff
@@ -7,7 +7,7 @@ classes:
   JsPlatformHelpers:
     abstract: true
     staticMethods:
-      # set_default_realm_file_directory: '(dir: std::string)' # currently only used via android JNI
+      set_default_realm_file_directory: '(dir: std::string)'
```
elle-j, line 10: When rebasing, you'll have to add `set_default_realm_file_directory` [here](https://github.com/realm/realm-js/blob/569fbb82c2cbb036b682d95234d1a48fa0304254/packages/realm/bindgen/js_opt_in_spec.yml#L177) as well 👍

packages/realm/src/app-services/App.ts
```diff
@@ -40,24 +40,34 @@ export type AnyApp = App<any, any>;
...
   app?: LocalAppConfiguration;
 
   /**
    * The timeout for requests (in milliseconds)
+   * @since v10.0.0
    */
   timeout?: number;
+
+  /**
+   * Specify where synced Realms and metadata is stored. If not specified, the current work directory is used.
```
elle-j, line 67:
```suggestion
   * Specify where synced Realms and metadata are stored. If not specified, the current work directory is used.
```

types/app.d.ts
```diff
@@ -360,6 +360,11 @@ declare namespace Realm {
      * Specifying this will enable the server to respond differently to specific versions of specific apps.
      */
     app?: LocalAppConfiguration;
+
+    /**
+     * Specify where synced Realms and metadata is stored. If not specified, the current work directory is used.
```
elle-j, line 365:
```suggestion
     * Specify where synced Realms and metadata are stored. If not specified, the current work directory is used.
```

elle-j (approved):
🥳

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -39,7 +39,7 @@ export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkCo
       throw new Error("Unexpected app on context, use only one importAppBefore per test");
     } else {
       const { appId, baseUrl } = await importApp(config);
-      this.app = new Realm.App({ id: appId, baseUrl, ...sdkConfig });
+      this.app = new Realm.App({ id: appId, baseUrl, baseFilePath: config?.baseFilePath, ...sdkConfig });
```
takameyer, line 42: There is a lint error here, but I'm unsure of where it comes from.

```
JsPlatformHelpers.setDefaultRealmFileDirectory is not a function
```

takameyer: Also, I don't think we should be setting `baseFilePath` on `config`, but rather on `sdkConfig`.  I can see how the intention of these parameters are misleading. The `config` is for the `app-importer` to setup the local BaaS. 
 The `sdkConfig` contains the properties for `Realm.App`.  With that being said we should just spread the `sdkConfig` into the `Realm.App`.
 ```suggestion
      this.app = new Realm.App({ id: appId, baseUrl, ...sdkConfig });
```

File: packages/realm-app-importer/src/AppConfigBuilder.ts [hunk 5]
```diff
@@ -18,6 +18,7 @@
 
 export type AppConfig = {
   name: string;
+  baseFilePath?: string;
```
takameyer, line 21: This should be removed and we should be using the `AppConfiguration`.  Perhaps this is a good time to refactor the type `AppConfig` into `AtlasAppConfig` or `BaasConfig`.

packages/realm-app-importer/src/AppConfigBuilder.ts
```diff
@@ -169,6 +170,11 @@ export class AppConfigBuilder {
     return this;
   }
 
+  baseFilePath(value: string) {
```
takameyer, line 173: Also we should remove this. As per my other comments.

integration-tests/tests/src/node/path.ts
```diff
@@ -57,6 +59,32 @@ describe("path configuration (local)", function () {
   });
 });
 
+describe.skipIf(environment.missingServer, `app configuration of root directory (flexible sync)`, async function () {
+  // describe.only(`app configuration of root directory (flexible sync)`, async function () {
+  this.timeout(60_000);
+  const tmpdir = getAbsolutePath();
+  importAppBefore(buildAppConfig("with-flx").baseFilePath(tmpdir).anonAuth().flexibleSync());
```
takameyer, line 66: Let's use the `sdkConfig` param of `importAppBefore` to set the properties of the local `Realm.App`instance.  The builder parts should just be used for configuring the remote BaaS instance.
```suggestion
  importAppBefore(buildAppConfig("with-flx").anonAuth().flexibleSync(), {baseFilePath:tmpDir});
```

takameyer (changes requested):
I would like to remove `baseFilePath` from the builder for BaaS and use the `skdConfig` param instead.  We should also rename the type `AppConfig` from the `app-importer` to `BaasConfig`, or `AtlasConfig` or `RemoteAppConfig`. (Perhaps a quick question in the group what the best name is, or just pick something besides `AppConfig`).  This will hopefully remove confusion between what is the local `AppConfiguration` and what is remote.

packages/realm-app-importer/src/AppConfigBuilder.ts
kraenhansen: I agree and agree 👍

kraenhansen:
I see Andrew has added a couple of valuable comments and suggestions. I'll hold back my review until those has been integrated. Please ping me for a re-review 👍

integration-tests/tests/src/node/path.ts
kneth: The commit suggestion will not as `App.Configuration` (second argument) will require `id`. `importAppBefore()` could take a relaxed version of `App.Configuration` (only optional parameters)

integration-tests/tests/src/node/path.ts
```diff
@@ -57,6 +59,32 @@ describe("path configuration (local)", function () {
...
+        _sessionStopPolicy: SessionStopPolicy.Immediately,
+        flexible: true,
+        user: this.user,
+      },
+    } as Realm.Configuration);
+
+    expect(existsSync(tmpdir)).to.be.true;
+    expect(realm.path.startsWith(tmpdir));
+
+    realm.close();
```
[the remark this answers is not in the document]
kneth: Good catch!

integration-tests/tests/src/hooks/import-app-before.ts
kneth: I have reworked in the spirit of your comment

packages/realm-app-importer/src/AppConfigBuilder.ts
kneth: I have refactored it a bit

integration-tests/tests/src/node/path.ts
```diff
@@ -57,6 +59,32 @@ describe("path configuration (local)", function () {
...
+describe.skipIf(environment.missingServer, `app configuration of root directory (flexible sync)`, async function () {
+  // describe.only(`app configuration of root directory (flexible sync)`, async function () {
+  this.timeout(60_000);
+  const tmpdir = getAbsolutePath();
+  importAppBefore(buildAppConfig("with-flx").anonAuth().flexibleSync(), { baseFilePath: tmpdir });
+  authenticateUserBefore();
+
+  it("directory and file created where expected", async function () {
+    expect(existsSync(tmpdir)).to.be.true; // importAppBefore will create `tmpdir`
+    let config = {
```
takameyer, line 71:
```suggestion
    const config = {
```

takameyer (approved):
Looks great :shipit:

packages/realm-web/types/realm/app.d.ts
```diff
@@ -360,6 +360,11 @@ declare namespace Realm {
      * Specifying this will enable the server to respond differently to specific versions of specific apps.
      */
     app?: LocalAppConfiguration;
+
+    /**
+     * Specify where synced Realms and metadata is stored. If not specified, the current work directory is used.
+     */
+    baseFilePath?: string;
```
kraenhansen, line 367: No need to add this, as `realm-web` don't have an implementation for it and it's the only component depending on these types now.

packages/realm-app-importer/src/AppConfigBuilder.ts
kraenhansen, line 21: I (still) don't think this line is needed.

packages/realm/src/app-services/App.ts
```diff
@@ -183,12 +193,18 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
-    const { id, baseUrl, app, timeout, multiplexSessions = true } = config;
+    const { id, baseUrl, app, timeout, multiplexSessions = true, baseFilePath } = config;
     assert.string(id, "id");
     if (timeout !== undefined) {
       assert.number(timeout, "timeout");
     }
     assert.boolean(multiplexSessions, "multiplexSessions");
+    if (baseFilePath !== undefined) {
+      assert.string(baseFilePath, "baseFilePath");
+      fs.setDefaultDirectoryPath(baseFilePath);
```
kraenhansen, line 204: I don't think we should update the default path, just because someone choose to create an app instance with some random path?
```suggestion
```

packages/realm/src/app-services/App.ts
```diff
@@ -183,12 +193,18 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
+    }
+
+    fs.ensureDirectoryForFile(fs.joinPaths(fs.getDefaultDirectoryPath(), "mongodb-realm"));
```
kraenhansen, line 207:
```suggestion
    fs.ensureDirectoryForFile(fs.joinPaths(baseFilePath || fs.getDefaultDirectoryPath(), "mongodb-realm"));
```

packages/realm/src/app-services/App.ts
```diff
@@ -201,7 +217,7 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
         defaultRequestTimeoutMs: timeout ? binding.Int64.numToInt(timeout) : undefined,
       },
       {
-        baseFilePath: fs.getDefaultDirectoryPath(),
+        baseFilePath: baseFilePath || fs.getDefaultDirectoryPath(),
```
kraenhansen, line 220: Do you actually need to pass `fs.getDefaultDirectoryPath()`? I mean, core knows about this, so doesn't it default to that if `undefined` is passed as `baseFilePath`?

kraenhansen, line 220: Do you actually need to pass `fs.getDefaultDirectoryPath()`? I mean, core knows about this, so doesn't it default to that if `undefined` is passed as `baseFilePath`?

```suggestion
        baseFilePath,
```

src/android/io_realm_react_RealmReactModule.cpp
```diff
@@ -93,7 +93,7 @@ JNIEXPORT void JNICALL Java_io_realm_react_RealmReactModule_setDefaultRealmFileD
 
     // Setting the internal storage path for the application
     const char* strFileDir = env->GetStringUTFChars(fileDir, NULL);
-    realm::set_default_realm_file_directory(strFileDir);
+    realm::JsPlatformHelpers::set_default_realm_file_directory(strFileDir);
```
kraenhansen, line 96: 🤔 (how) did this work before?

packages/realm/src/app-services/App.ts
kraenhansen: Actually .. reading the code, I think Core doesn't know about this, in which case I see very little purpose in round-tripping this through C++? Since we don't expose an API to set the default to end-users, I'd suggest treading the "default directory path" as a constant.

packages/realm-app-importer/src/AppConfigBuilder.ts
kneth: You're right. I reworked it (not using it) but forgot remove it.

packages/realm-web/types/realm/app.d.ts
kneth: True, it can be removed.

packages/realm/src/app-services/App.ts
kneth: I see your point, and I think we need to find a "unified" way to set the default directory (one that works for both local and synced Realms) in the future. I'll change it so it `App.Config.baseFilePath` and `Realm.defaultDirectory` are two independent ways to specify where Realms are stored by default.

src/android/io_realm_react_RealmReactModule.cpp
kraenhansen: Never mind - I understand think I understand it better now.

kneth: Nope, but since we do late binding, you will only see if it use the method

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -230,7 +230,6 @@ async function collectPlatformData(packagePath = getProjectRoot()) {
       const electronPackageJson = JSON.parse(fs.readFileSync(electronPath, "utf-8"));
       frameworkVersion = electronPackageJson["version"];
     } catch (err) {
-      console.log("FISK 5", err);
```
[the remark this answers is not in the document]
kneth: [So long, and thanks for all the fish](https://en.wikipedia.org/wiki/So_Long,_and_Thanks_for_All_the_Fish)

PR #5898, opened 2023-06-16
Fix global `Realm` usage in integration tests

kraenhansen:
## What, How & Why?

This updates updates the warning and error emitted when accessing the global `Realm` to reflect the change to CommonJS and updates the integration tests to import accordingly.


elle-j (approved):
👍

PR #5911, opened 2023-06-20
Quick Fixes for 0.5

takameyer:
## What, How & Why?
* Remove warning about source maps by including the `src` in the package.json
* Move the comment blocks for `useAuth` and `useEmailPasswordAuth` so that they show up in editors
* Refactor the `useAuthOperation` to use the `reject` callback rather than catch.
* Remove returning a result from `useAuthOperation`.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm-react/src/useAuthOperation.tsx
+++ b/packages/realm-react/src/useAuthOperation.tsx
@@ -42,8 +42,8 @@ export function useAuthOperation<Args extends unknown[]>({
         error: undefined,
         operation: operationName,
       });
-      return operation(...args)
-        .then((res) => {
+      operation(...args).then(
+        () => {
           setResult({
             state: OperationState.Success,
             pending: false,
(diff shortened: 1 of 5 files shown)


elle-j (approved):

packages/realm-react/src/useAuthOperation.tsx
```diff
@@ -42,18 +42,17 @@ export function useAuthOperation<Args extends unknown[]>({
         error: undefined,
         operation: operationName,
       });
-      return operation(...args)
-        .then((res) => {
+      return operation(...args).then(
```
kraenhansen, line 45: I thought you didn't want to return a promise here?

takameyer: True...we could just call this.

takameyer: Fixed.

PR #5926, opened 2023-06-22
Moving all valuable docs from `/types` into the SDK

kraenhansen:
## What, How & Why?

I went through the `/types` files and moved any valuable information into the SDK and moved the `/types` directory into the `realm-web` package.

My intent is not to suggest improvements to the existing comments or code, but rather ensure we can delete the old types from the root without loosing anything valuable.


packages/realm/src/Unmanaged.ts
```diff
@@ -49,7 +47,7 @@ type RealmDictionaryRemappedModelPart<T> = {
 };
 
 /** Omits all properties of a model which are not defined by the schema */
-type OmittedRealmTypes<T> = Omit<
+export type OmittedRealmTypes<T> = Omit<
```
takameyer, line 50: Is that something we need exported?

packages/realm/src/Realm.ts
```diff
@@ -820,12 +821,12 @@ export class Realm {
    */
   create<T = DefaultObject>(
     type: string,
-    values: Partial<T> | Partial<RealmInsertionModel<T>>,
+    values: Partial<T> | Partial<Unmanaged<T>>,
```
takameyer, line 824: Does this also take `, RequiredTypes`?

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
     return this.internal.config.appId;
   }
 
+  /**
+   * Log in a user.
```
takameyer, line 213:
```suggestion
   * Log in a user.
   * @param credentials - {@link Credentials} object which determines the authentication method.
   * @returns {@link User}
```

packages/realm/src/app-services/EmailPasswordAuth.ts
```diff
@@ -86,15 +90,23 @@ export class EmailPasswordAuth {
 
   /**
    * Sends an email to the user for resetting the password.
-   * @param credential The email details to send the reset to
-   * @param credential.email - The email address of the user to re-send a confirmation for.
+   * @param details The email details to send the reset to
```
takameyer, line 93:
```suggestion
   * @param details - The email details to send the reset to
```
I like the `-` after the argument name in the `@param` directive for readability.  Most likely missing in other places.  Maybe a lint rule?

takameyer: https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.README/rules/require-hyphen-before-param-description.md

packages/realm/src/app-services/App.ts
takameyer: I also think we should explicitly define return types on our public methods (not just jsdoc and implicit).  Makes it less likely that users will have to dig through our code to figure out what's going on.

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
+   * NOTE: This irrecoverably deletes the user from the device as well as the server!
+   * @returns A promise that resolves once the user has been deleted.
+   */
   public async deleteUser(user: AnyUser) {
     await this.internal.deleteUser(user.internal);
   }
 
+  /**
+   * Adds a listener that will be fired on various user events.
+   * This includes login, logout, switching users, linking users and refreshing custom data.
```
takameyer, line 270:
```suggestion
   * This includes login, logout, switching users, linking users and refreshing custom data.
   * @param callback - Listener callback function
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
+   */
   public addListener(callback: AppChangeCallback) {
     this.listeners.add(callback);
   }
 
+  /**
+   * Removes an event listener previously added via {@link App.addListener}.
```
takameyer, line 277:
```suggestion
   * Removes an event listener previously added via {@link App.addListener}.
   * @param callback - The same callback function from {@link App.addListener}.
```

takameyer (approved):
Just a few things.  Thanks for doing this! 💪🏼

kraenhansen:
Looks like I need to update the `relam-web` workflow to use NPM workspaces too.

packages/realm/src/app-services/EmailPasswordAuth.ts
kraenhansen: I suggest leaving that for a future PR.

packages/realm/src/app-services/App.ts
kraenhansen: It wasn't my intent to improve the comments as such. Just wanted to ensure we could delete the old types from the root without loosing anything valuable. I suggest leaving this to a future PR too.

takameyer: I understand that, but I was cleaning the campground while I was reading the review.

packages/realm/src/app-services/App.ts
kraenhansen: It's a recommend setting to leave type information out of tsdoc comments as it's considered redundant with the TS type declarations.

takameyer: Sorry, my comment isn't clear.  This method (`logIn`) in particular is not showing a return type in the TS type declarations.  I believe it returns a `Promise<User>`

takameyer: > It's a recommend setting to leave type information out of tsdoc comments as it's considered redundant with the TS type declarations.

I'm fine with that if it's clear from the code.

packages/realm/src/Realm.ts
```diff
@@ -285,6 +285,7 @@ export class Realm {
 
   /**
    * Clears the state by closing and deleting any Realm in the default directory and logout all users.
+   * NOTE: Not a part of the public API and it's primarily used from the library's tests.
    * @private
    */
   public static clearTestState(): void {
```
elle-j, line 291: Regarding `clearTestState()`, do you know of the reason why this is public with the private annotation, rather than internal?

packages/realm/src/Unmanaged.ts
```diff
@@ -58,11 +56,28 @@ type OmittedRealmTypes<T> = Omit<
...
+  RequiredProperties
+>;
+
 /** Remaps realm types to "simpler" types (arrays and objects) */
 type RemappedRealmTypes<T> = RealmListsRemappedModelPart<T> & RealmDictionaryRemappedModelPart<T>;
 
 /**
- * Joins T stripped of all keys which value extends {@link Collection} and all inherited from {@link RealmObject},
- * with only the keys which value extends {@link List}, remapped as Arrays.
+ * Joins `T` stripped of all keys which value extends {@link Collection} and all inherited from Realm.Object,
```
elle-j, line 75: Should we perhaps link or backtick this?
```suggestion
 * Joins `T` stripped of all keys which value extends {@link Collection} and all inherited from {@link Realm.Object},
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
     return this.internal.config.appId;
   }
 
+  /**
+   * Log in a user.
+   */
   public async logIn(credentials: Credentials) {
```
elle-j, line 215:
```suggestion
  public async logIn(credentials: Credentials): User {
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
+   */
   public switchUser(): unknown {
     throw new Error("Not yet implemented");
   }
 
+  /**
+   * Logs out and removes a user from the client.
+   * @returns A promise that resolves once the user has been logged out and removed from the app.
+   */
   public async removeUser(user: AnyUser) {
```
elle-j, line 255:
```suggestion
  public async removeUser(user: AnyUser): Promise<void> {
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
     await this.internal.removeUser(user.internal);
   }
 
+  /**
+   * Delete the user.
+   * NOTE: This irrecoverably deletes the user from the device as well as the server!
+   * @returns A promise that resolves once the user has been deleted.
+   */
   public async deleteUser(user: AnyUser) {
```
elle-j, line 264:
```suggestion
  public async deleteUser(user: AnyUser): Promise<void> {
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
     await this.internal.deleteUser(user.internal);
   }
 
+  /**
+   * Adds a listener that will be fired on various user events.
+   * This includes login, logout, switching users, linking users and refreshing custom data.
+   */
   public addListener(callback: AppChangeCallback) {
```
elle-j, line 272:
```suggestion
  public addListener(callback: AppChangeCallback): void {
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
     this.listeners.add(callback);
   }
 
+  /**
+   * Removes an event listener previously added via {@link App.addListener}.
+   */
   public removeListener(callback: AppChangeCallback) {
```
elle-j, line 279:
```suggestion
  public removeListener(callback: AppChangeCallback): void {
```

packages/realm/src/app-services/App.ts
```diff
@@ -206,46 +209,80 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
     this.listeners.remove(callback);
   }
 
+  /**
+   * Removes all event listeners previously added via {@link App.addListener}.
+   */
   public removeAllListeners() {
```
elle-j, line 286:
```suggestion
  public removeAllListeners(): void {
```

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -201,6 +201,33 @@ export type InitialSubscriptions = {
...
+   *       rerunOnOpen: true,
+   *     },
+   *   },
+   *   // ... rest of config ...
+   * };
+   * const realm = await Realm.open(config);
+   *
+   * // At this point, the Realm will be open with the data for the initial set
+   * // subscriptions fully synchronised.
+   * ```
```
elle-j, line 229: Replace `Example:` with `@example` and removed the ```
```suggestion
   * @example
   * const config: Realm.Configuration = {
   *   sync: {
   *     user,
   *     flexible: true,
   *     initialSubscriptions: {
   *       update: (subs, realm) => {
   *         subs.add(realm.objects('Task'));
   *       },
   *       rerunOnOpen: true,
   *     },
   *   },
   *   // ... rest of config ...
   * };
   * const realm = await Realm.open(config);
   *
   * // At this point, the Realm will be open with the data for the initial set
   * // subscriptions fully synchronised.
```

elle-j (approved):
DX increase 🥳

packages/realm/src/Realm.ts
kraenhansen: Because it needs to be callable from the integration tests. All the stuff we annotate internal is completely stripped from the types from the POV of the integration tests.

elle-j: Thought we usually used `// @ts-expect-error` when having to use our internal parts (since it's still callable despite `@internal`). But helps with intellisense to not have it.

kraenhansen:
I had a go at trying to fix the `realm-web` package (building on CI etc.), but ran into some roadblocks: https://github.com/realm/realm-js/pull/5934

I'll go ahead and merge this PR as it isn't breaking more than is already broken on `main`.

Issue #5605, opened 2023-03-21
Deprecate `Realm.Object#keys()` and `Realm.Object#entries()`


kraenhansen:

elle-j:
Closing this issue as this was resolved in [#5927](https://github.com/realm/realm-js/pull/5927).

PR #5927, opened 2023-06-22
Added RealmObject.entries()

### LINKED ISSUE #5923: Implement `RealmObject#entries`
We need to add an implementation for the entries instance method on RealmObject and add tests for this as well.

papafe:
Added `RealmObject.entries()`

I am not sure this method (or `RealmObject.values()`) need docs, as they are equivalent to the default `Object`  methods. Also I'm not sure if we should add it to the changelog

This closes #5923 and #5605

## ☑️ ToDos
* [x] 🚦 Tests

### DIFF
--- a/packages/realm/src/Object.ts
+++ b/packages/realm/src/Object.ts
@@ -274,11 +274,13 @@ export class RealmObject<T = DefaultObject> {
    */
   private declare readonly [KEY_SET]: ReadonlySet<string>;
 
+  /** @deprecated Please use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys|Object.keys()}  */
   keys(): string[] {
     // copying to prevent caller from modifying the static array.
     return [...this[KEY_ARRAY]];
   }
 
+  /** @deprecated Please use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries|Object.entries()}*/
   entries(): [string, unknown][] {
     return Object.entries(this);
   }
--- a/packages/realm/src/Object.ts
+++ b/packages/realm/src/Object.ts
@@ -280,7 +280,7 @@ export class RealmObject<T = DefaultObject> {
   }
 
   entries(): [string, unknown][] {
-    throw new Error("Not yet implemented");
+    return Object.entries(this);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


kneth (approved):

kneth:
>  am not sure this method need docs

Currently the methods don't have API documentation, and I suggest a simple solution like `@see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries}` as API documentation.

kraenhansen:
I don't know if you're aware of https://github.com/realm/realm-js/issues/5605 ... But you could choose to solve that with this PR too.

kraenhansen (approved):

takameyer (approved):
:shipit:

papafe:
> I don't know if you're aware of #5605 ... But you could choose to solve that with this PR too.

Didn't know, but I've deprecated the methods and slightly modified the tests.

packages/realm/src/Object.ts
```diff
@@ -274,13 +274,15 @@ export class RealmObject<T = DefaultObject> {
    */
   private declare readonly [KEY_SET]: ReadonlySet<string>;
 
+  /** @deprecated Please use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys|Object.keys()}  */
```
elle-j, line 277: Spaces needed in between `|` for the formatting to be correct 👍 
```suggestion
  /** @deprecated Please use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys | Object.keys()}  */
```

packages/realm/src/Object.ts
```diff
@@ -274,13 +274,15 @@ export class RealmObject<T = DefaultObject> {
...
   keys(): string[] {
     // copying to prevent caller from modifying the static array.
     return [...this[KEY_ARRAY]];
   }
 
+  /** @deprecated Please use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries|Object.entries()}*/
```
elle-j, line 283:
```suggestion
  /** @deprecated Please use {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries | Object.entries()} */
```

elle-j (approved):
🥳

packages/realm/src/Object.ts
papafe: Thanks!

PR #5930, opened 2023-06-23
Fix failing `@realm/react` async unit test

kraenhansen:
## What, How & Why?

This is an attempt to fix the sporadically failing `@realm/react` test, by unmounting the test component before completing the test. This will make sure the list doesn't try to access the size of the results after the Realm has been closed (most likely as an effect of the after each hook calling `Realm.clearTestState`)


takameyer (commented):
Good idea :shipit:

takameyer (approved):
Good idea :shipit:

PR #5931, opened 2023-06-23
Add eslint rules for comments

takameyer:
## What, How & Why?
* Space after `//`
* Comment blocks start capitalized

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,9 @@
     "./header.eslintrc"
   ],
   "rules": {
-    "no-restricted-globals": ["error", "globalThis"]
+    "no-restricted-globals": ["error", "globalThis"],
+    "spaced-comment": ["error", "always", {"exceptions":["/"]}],
+    "capitalized-comments": ["error", "always", {"ignoreConsecutiveComments": true }]
   },
   "env": {
     "es2020": true
--- a/integration-tests/tests/src/typings.d.ts
+++ b/integration-tests/tests/src/typings.d.ts
@@ -20,7 +20,7 @@ type RealmObject = import("realm").Object;
 type App = import("realm").App;
 type User = import("realm").User;
 type Configuration = import("realm").Configuration;
-//type BenchmarkResult = import("@thi.ng/bench").BenchmarkResult;
+// Type BenchmarkResult = import("@thi.ng/bench").BenchmarkResult;
 
 interface fs {
   exists: (path: string) => boolean;
@@ -174,7 +174,7 @@ type RealmContext = {
 type RealmObjectContext<T = Record<string, unknown>> = {
   object: RealmObject & T;
 } & RealmContext;
-// type BenchmarkContext = {
+// Type BenchmarkContext = {
 //   result: BenchmarkResult;
 // } & Mocha.Context;
 
@@ -202,7 +202,7 @@ interface Console {
 }
 
 declare const console: Console;
-// allow import of json files
+// Allow import of json files
 declare module "*.json" {
   const value: unknown;
   export = value;
--- a/integration-tests/tests/src/utils/benchmark.ts
+++ b/integration-tests/tests/src/utils/benchmark.ts
@@ -60,7 +60,7 @@ export function describePerformance(title: string, parameters: PerformanceTestPa
     before(parameters.before);
     itPerforms(parameters.benchmarkTitle, parameters.test);
     after(function (this: BenchmarkContext) {
-      // console.log(this.summary);
+      // Console.log(this.summary);
     });
   });
 }
--- a/packages/realm/src/Configuration.ts
+++ b/packages/realm/src/Configuration.ts
@@ -41,7 +41,7 @@ export type BaseConfiguration = {
   readOnly?: boolean;
   fifoFilesFallbackPath?: string;
   sync?: SyncConfiguration;
-  /**@internal */ openSyncedRealmLocally?: true;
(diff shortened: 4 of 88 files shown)


packages/realm/src/Realm.ts
```diff
@@ -194,7 +194,7 @@ type RealmSchemaExtra = Record<string, ObjectSchemaExtra | undefined>;
 type ObjectSchemaExtra = {
   constructor?: RealmObjectConstructor;
   defaults: Record<string, unknown>;
-  // objectTypes: Record<string, unknown>;
+  // ObjectTypes: Record<string, unknown>;
```
kraenhansen, line 197: Admittedly, this line should probably be deleted, but I can see how the `capitalized-comments` rule would be annoying when commenting out code while developing and having format-on-save mess it up like this.

packages/realm/src/Configuration.ts
```diff
@@ -59,7 +59,7 @@ export type ConfigurationWithoutSync = BaseConfiguration & {
 
 export type Configuration = ConfigurationWithSync | ConfigurationWithoutSync;
 
-// export type PartitionValue = string | number | null | ObjectId | UUID;
+// Export type PartitionValue = string | number | null | ObjectId | UUID;
```
kraenhansen, line 62: Similar to my other comment

integration-tests/tests/src/utils/benchmark.ts
```diff
@@ -60,7 +60,7 @@ export function describePerformance(title: string, parameters: PerformanceTestPa
     before(parameters.before);
     itPerforms(parameters.benchmarkTitle, parameters.test);
     after(function (this: BenchmarkContext) {
-      // console.log(this.summary);
+      // Console.log(this.summary);
```
kraenhansen, line 63: Another bad fix.

integration-tests/tests/src/typings.d.ts
```diff
@@ -174,7 +174,7 @@ type RealmContext = {
 type RealmObjectContext<T = Record<string, unknown>> = {
   object: RealmObject & T;
 } & RealmContext;
-// type BenchmarkContext = {
+// Type BenchmarkContext = {
 //   result: BenchmarkResult;
 // } & Mocha.Context;
```
kraenhansen, line 179: Here as well ..

integration-tests/tests/src/typings.d.ts
```diff
@@ -20,7 +20,7 @@ type RealmObject = import("realm").Object;
 type App = import("realm").App;
 type User = import("realm").User;
 type Configuration = import("realm").Configuration;
-//type BenchmarkResult = import("@thi.ng/bench").BenchmarkResult;
+// Type BenchmarkResult = import("@thi.ng/bench").BenchmarkResult;
```
kraenhansen, line 23: And here.

kraenhansen (commented):
There's a couple of places where commented out source-code gets updated. Admittedly that code should probably be deleted, but I can see how the `capitalized-comments` rule combined with format-on-save would be super annoying at times.

.eslintrc.json
```diff
@@ -6,7 +6,9 @@
     "./header.eslintrc"
   ],
   "rules": {
-    "no-restricted-globals": ["error", "globalThis"]
+    "no-restricted-globals": ["error", "globalThis"],
+    "spaced-comment": ["error", "always", {"exceptions":["/"]}],
+    "capitalized-comments": ["error", "always", {"ignoreConsecutiveComments": true }]
```
kraenhansen, line 11: I'd say "warn" would be good enough here.

elle-j, line 11: I saw e.g. `// tslint:disable-next-line` in some tests and not sure if the capitalization will invalidate that. If so, we could add it to the [ignorePattern](https://eslint.org/docs/latest/rules/capitalized-comments#ignorepattern). But maybe those tslints can even be removed?
```suggestion
    "capitalized-comments": ["error", "always", {
      "ignorePattern": "tslint",
      "ignoreConsecutiveComments": true
    }]
```

elle-j (approved):

.eslintrc.json
```diff
@@ -6,7 +6,9 @@
     "./header.eslintrc"
   ],
   "rules": {
-    "no-restricted-globals": ["error", "globalThis"]
+    "no-restricted-globals": ["error", "globalThis"],
+    "spaced-comment": ["error", "always", {"exceptions":["/"]}],
```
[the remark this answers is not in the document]
takameyer: Good idea on `balanced`.  The expection for `/` is still needed due to our headers.  Granted, we could reformat or header to be similar to our other comment blocks

takameyer:
Thanks for the comments so far.  I have attempted some of the suggestions, but haven't gotten it in a state where I'm really happy with the results.  Our lovely header is making things a bit difficult.  Perhaps we should reformat it, although I have also gone down this road and haven't found a good alternative yet. 
Honestly, I have spent more time on this than intended 😄.  I'm going to leave this open for now and get back to it when as time permits.

.eslintrc.json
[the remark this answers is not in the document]
elle-j: Oh yeah, then we could use `exceptions` rather than `markers` if you'd like 🙂

takameyer:
I will most likely never get back to this.  More effort than it was worth.

PR #5932, opened 2023-06-23
Added App.switchUser

### LINKED ISSUE #5922: Implement `App#switchUser`
We need to add an implementation for the switchUser instance method on App and add tests for this as well.

papafe:
Added implementation for `App.switchUser`.

This closes #5922

## ☑️ ToDos
* [x] 🚦 Tests


integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -39,17 +39,17 @@ function expectIsUSer(user: Realm.User) {
   expect(user).instanceOf(Realm.User);
 }
 
-function expectIsSameUser(value: Realm.User, user: Realm.User | null) {
+function expectIsSameUser(value: Realm.User, user: Realm.User<any, any> | null) {
```
papafe, line 42: This and the following ones are just my "I don't know what I'm doing" attempt at making typescript stop complaining about the types. I'm curious to know what would be the correct approach for similar cases

packages/realm/src/app-services/App.ts
```diff
@@ -233,14 +233,28 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
     await this.internal.removeUser(user.internal);
   }
 
+  /**
+   * Deletes a user from the server. The user is also removed from the device together with their local data.
+   * If the user is logged in, they will be logged out in the process.
+   * @param user The user to delete.
+   */
   public async deleteUser(user: AnyUser) {
     await this.internal.deleteUser(user.internal);
```
kraenhansen, line 259: These doc comments will conflict with https://github.com/realm/realm-js/pull/5926

integration-tests/tests/src/tests/sync/user.ts
kneth: Hmm, when you hover over a user, the type isn't `Realm.User`. I think that's the root cause.

Screenshot 2023-06-26 at 12 51 38

kneth (approved):
I believe the type of the user must be fixed but it is unrelated to this PR.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -39,17 +39,17 @@ function expectIsUSer(user: Realm.User) {
...
   expectIsUSer(value);
```
takameyer, line 43: I think there's a typo here?  `expectIsUSer` -> `expectIsUser`

integration-tests/tests/src/tests/sync/user.ts
takameyer: We do have an `AnyUser` you could use as well.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -327,6 +327,25 @@ describe.skipIf(environment.missingServer, "User", () => {
...
       });
 
+      it("can switch user", async function (this: Mocha.Context & AppContext & RealmContext) {
+        expect(this.app.currentUser, "No users").to.be.null;
+
+        const user1 = await registerAndLogInEmailUser(this.app);
+        const user2 = await registerAndLogInEmailUser(this.app);
+
+        expect(this.app.currentUser?.id).to.equal(user2.id);
+        expect(Object.keys(this.app.allUsers)).to.have.lengthOf(2);
```
takameyer, line 337: Wonder if we have a cleanup issue here.  The user count is `7` instead of `2` in the tests.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -327,6 +327,25 @@ describe.skipIf(environment.missingServer, "User", () => {
...
+
+        await this.app.deleteUser(user2);
+
+        expect(() => this.app.switchUser(user2)).to.throw("User is no longer valid or is logged out");
+      });
+
       describe("state", () => {
         it("can fetch state when logged in with email password", async function (this: AppContext & RealmContext) {
           expect(this.app.currentUser).to.be.null;
```
takameyer, line 351: There seems to be a problem occurring in this test.  I have also seen some oddities regarding `currentUser` in my realm/react tests.  There might be an underlying issue in our implementation here.

packages/realm/src/app-services/App.ts
takameyer: Better get that merged soon so we can start improving the docs 😂

takameyer (commented):
Implementation and tests look great.  It does appear though we have some failing tests regarding user that need to be addressed.

integration-tests/tests/src/tests/sync/user.ts
papafe: Yes, I'll fix this

integration-tests/tests/src/tests/sync/user.ts
papafe: @kneth, yes, because it's using generics. 
@takameyer I can see that, but it seems that's not exported by Realm, so I'd keep it like this if it makes sense?

packages/realm/src/app-services/App.ts
```diff
@@ -249,10 +249,12 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
   }
 
   /**
-   * Switch current user, from an instance of `User` or the string id of the user.
+   * Switches the current user to the one specified in {@link user}.
+   * The new user must be currently logged in, otherwise this method will throw an error.
+   * @param user The new current user.
```
elle-j, line 254:
```suggestion
   * @param user The user to switch to.
```

packages/realm/src/app-services/App.ts
```diff
@@ -249,10 +249,12 @@ export class App<FunctionsFactoryType = DefaultFunctionsFactory, CustomDataType
...
    */
-  public switchUser(): unknown {
-    throw new Error("Not yet implemented");
+  public switchUser(user: AnyUser) {
```
elle-j, line 256:
```suggestion
  public switchUser(user: AnyUser): void {
```

elle-j (approved):
🚀

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -327,6 +326,29 @@ describe.skipIf(environment.missingServer, "User", () => {
...
+        expect(Object.keys(this.app.allUsers)).to.have.lengthOf(2);
+
+        this.app.switchUser(user1);
+
+        expect(this.app.currentUser?.id).to.equal(user1.id);
+        expect(Object.keys(this.app.allUsers)).to.have.lengthOf(2);
+
+        await this.app.deleteUser(user2);
+
+        expect(() => this.app.switchUser(user2)).to.throw("User is no longer valid or is logged out");
```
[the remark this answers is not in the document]
papafe: In general I think it makes sense, but I looked at the other method tests and they also have multiple assertions inside, so I would keep it as it is now for consistency

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -327,6 +326,29 @@ describe.skipIf(environment.missingServer, "User", () => {
         expect(didFail).to.be.true;
       });
 
+      it("can switch user", async function (this: Mocha.Context & AppContext & RealmContext) {
+        expect(this.app.currentUser, "No users").to.be.null;
+
+        expect(Object.keys(this.app.allUsers)).to.have.lengthOf(0);
```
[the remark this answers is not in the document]
papafe: I can see that. I think it's a little bit about how much we want to lay on char or  not. I don't have strong opinions 😁

integration-tests/tests/src/tests/sync/user.ts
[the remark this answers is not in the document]
elle-j: Alrighty 👍 Just thinking it may be good to take the opportunity to structure it like that from the start when it's being written from scratch 😉

PR #5938, opened 2023-06-26
Add missing dependencies to wasm example

nlarew:
## What, How & Why?

Adds missing depdencies:
- ~Polyfills the `stream` and `crypto` Node.js built-in modules~
- ~Adds the `bson` library, a peer dependency~
- Adds some missing useEffect dependencies

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/examples/example-react-task/craco.config.ts
+++ b/examples/example-react-task/craco.config.ts
@@ -18,7 +18,7 @@
 
 import { CracoConfig } from "@craco/types";
 
-export default {
+const config = {
   webpack: {
     configure(config, context) {
       const experiments = {
@@ -26,19 +26,12 @@ export default {
         topLevelAwait: true,
       }
 
-      const resolve = {
-        ...config.resolve,
-        fallback: {
-          crypto: require.resolve('crypto-browserify'),
-          stream: require.resolve('stream-browserify'),
-        }
-      }
-
       return {
         ...config,
         experiments,
-        resolve,
       };
     },
   }
 } satisfies CracoConfig;
+
+export default config;
--- a/examples/example-react-task/craco.config.ts
+++ b/examples/example-react-task/craco.config.ts
@@ -18,7 +18,7 @@
 
 import { CracoConfig } from "@craco/types";
 
-const config: CracoConfig = {
+export default {
   webpack: {
     configure(config, context) {
       const experiments = {
@@ -41,6 +41,4 @@ const config: CracoConfig = {
       };
     },
   }
-};
-
-export default config;
+} satisfies CracoConfig;
--- a/examples/example-react-task/craco.config.ts
+++ b/examples/example-react-task/craco.config.ts
@@ -21,17 +21,26 @@ import { CracoConfig } from "@craco/types";
 const config: CracoConfig = {
   webpack: {
     configure(config, context) {
-      config.experiments = {
-        topLevelAwait: true,
+      const experiments = {
         ...config.experiments,
-      };
+        topLevelAwait: true,
+      }
 
-      return config;
+      const resolve = {
+        ...config.resolve,
+        fallback: {
+          crypto: require.resolve('crypto-browserify'),
+          stream: require.resolve('stream-browserify'),
+        }
+      }
+
+      return {
+        ...config,
+        experiments,
+        resolve,
+      };
     },
-  },
-  devServer: {
-    open: "chrome",
-  },
+  }
 };
 
 export default config;
--- a/examples/example-react-task/package.json
+++ b/examples/example-react-task/package.json
@@ -27,13 +27,10 @@
   },
   "dependencies": {
     "@realm/react": "^0.5.1",
-    "bson": "^5.3.0",
-    "crypto-browserify": "^3.12.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-router-dom": "^6.11.2",
     "realm": "12.0.0-browser.2",
-    "stream-browserify": "^3.0.0",
     "web-vitals": "^2.1.4"
   },
   "devDependencies": {
--- a/examples/example-react-task/package.json
+++ b/examples/example-react-task/package.json
@@ -27,10 +27,13 @@
   },
   "dependencies": {
     "@realm/react": "^0.5.1",
+    "bson": "^5.3.0",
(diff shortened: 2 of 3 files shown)


examples/example-react-task/craco.config.ts
```diff
@@ -21,17 +21,26 @@ import { CracoConfig } from "@craco/types";
...
-      };
+        topLevelAwait: true,
+      }
 
-      return config;
+      const resolve = {
+        ...config.resolve,
+        fallback: {
+          crypto: require.resolve('crypto-browserify'),
+          stream: require.resolve('stream-browserify'),
```
elle-j, line 33: Interesting, what errors did you run in to when these were not included? And did you run the app directly in the JS repo or did you grab just the example app directory?

examples/example-react-task/package.json
```diff
@@ -27,10 +27,13 @@
   },
   "dependencies": {
     "@realm/react": "^0.5.1",
+    "bson": "^5.3.0",
```
elle-j, line 30: We do provide `bson` in the `realm` package directly. Usually it's used via e.g. `import { BSON } from "realm"`. But it'd also be interesting here to know the error you got from not have this separately.

examples/example-react-task/src/pages/LoginPage.tsx
```diff
@@ -34,7 +34,7 @@ export function LoginPage() {
     if (result.operation === AuthOperationName.Register && result.success) {
       logIn({ email, password });
     }
-  }, [result.operation, result.success]);
+  }, [result.operation, result.success, logIn, email, password]);
```
elle-j, line 37: React dependency list warnings are not pretty 😉 so I understand the addition of these. Although, functions like `logIn()` will not change behavior during runtime, but as part of rerenders elsewhere in the app its reference may change, causing `useEffect()` to run unnecessarily. Similarly, `email` and `password` would cause `useEffect()` to run on each character entered. If only running it when `result.operation` or `result.success` changes, it minimizes those calls. (That's also why just `result` itelf is not added.)

Did you experience any faulty behavior from not having those dependencies listed or only warnings?

If you think it's better for the example app to run without those warnings, then I'm fine with adding to the dependency list 👍

elle-j (commented):
Nick, thanks a lot for the updates! Just got a few questions and comments 🙂

nlarew:
Here's my reproduction & motivation for the changes

First, I install a clean copy of realm-js and set up the app:

```shell
gh repo clone realm/realm-js
cd realm-js
git fetch
git checkout nh/wasm/emscripten_target
cd examples/example-react-task
pnpm install
pnpm run start
```

At this point I get the following error & warning:

```
Module not found: Error: Can't resolve 'bson' in '/Users/nlarew/rjs-web-sync/realm-js/examples/example-react-task/node_modules/.pnpm/realm@12.0.0-browser.2/node_modules/realm/dist'

WARNING in [eslint]
src/pages/LoginPage.tsx
  Line 37:6:  React Hook useEffect has missing dependencies: 'email', 'logIn', and 'password'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```

To fix the error, I install `bson` and run the app again:

```sh
pnpm install bson
pnpm run start
```

This fixed the `bson` error but the `bson` module uses the Node `crypto` and
`stream` APIs, so now I get this error:

```
Failed to compile.

Module not found: Error: Can't resolve 'crypto' in '/Users/nlarew/rjs-web-sync/realm-js/examples/example-react-task/node_modules/.pnpm/bson@5.3.0/node_modules/bson/lib'
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "crypto": require.resolve("crypto-browserify") }'
        - install 'crypto-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "crypto": false }
```

I'd also get a similar error for `stream` but this one throws first.

They're both coming from `bson`, so maybe your upcoming changes will fix this?

To solve both, I follow the recommended advice and add a fallback to the webpack
config. In this case that's through craco:

```
pnpm i stream-browserify crypto-browserify
```

```js
import { CracoConfig } from "@craco/types";

export default {
  webpack: {
    configure(config, context) {
      const experiments = {
        ...config.experiments,
        topLevelAwait: true,
      }

      const resolve = {
        ...config.resolve,
        fallback: {
          crypto: require.resolve('crypto-browserify'),
          stream: require.resolve('stream-browserify'),
        }
      }

      return {
        ...config,
        experiments,
        resolve,
      };
    },
  }
} satisfies CracoConfig;
```

This fixes the errors! Note that I also removed the `devServer` config so that
my browser opens to `http://localhost:3000` instead of `http://localhost:3000/chrome`.

All that's left is the missing useEffect dependency lint warnings. I added the
missing dependencies and the app builds and runs without errors/warnings.

examples/example-react-task/src/pages/LoginPage.tsx
nlarew: I agree that as we have it set up we don't _actually_ need these additional dependencies - the app functions correctly either way. However in general we should declare exhaustive dependencies. This is a super lightweight effect function so I think the impact of running the effect on each character input would be negligible. It has the added benefit of the app having a clean build with no errors/warnings out of the box which I think gives devs a lot more confidence.

examples/example-react-task/craco.config.ts
elle-j: Thanks for the repro steps Nick. Still can't reproduce these errors unfortunately but will keep trying tmrw.

examples/example-react-task/src/pages/LoginPage.tsx
elle-j: Sounds good 🙂

elle-j (approved):
Great 🙂

examples/example-react-task/craco.config.ts
elle-j: Update: As per our Slack discussion, we'll remove `pnpm` instructions from the Docs and encourage only `npm` as of now.

PR #5943, opened 2023-06-27
Add bson ^5 to peerDeps in addition to ^4

elle-j:
## What, How & Why?

> **NOTE:** Wait to merge until a BSON release containing [these changes](https://github.com/mongodb/js-bson/pull/578) is out. (See [comment](https://github.com/realm/realm-js/pull/5943#issuecomment-1609398945) below)

Users of `realm@12.0.0-alpha.2` would end up with the following error:
```
Unable to resolve "bson" from "node_modules/realm/dist/bundle.react-native.mjs"
```

If using bson v5+, the issue is resolved but with the following warning:
```
warning " > realm@12.0.0-alpha.2" has incorrect peer dependency "bson@^4".
```

This PR adds the `^5` to the bson peer dependency.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


kraenhansen:
🤞 I can't remember why we needed to pin this to `v4`, I suspect I was concerned about the polyfills introduced by https://github.com/mongodb/js-bson/pull/550 (which got shipped as part of v5.0.0). Also, for context [a PR, which will vendor a couple of these polyfills has merged](https://github.com/mongodb/js-bson/pull/578) but is yet to be published, we might want to wait for the release of that and support only `v4 || v5.4.0` (or whatever that version will be).

To ensure we're testing with `bson@5` we should probably add a `devDependency` to the SDK package (in addition to the peer dependency) as well as [the integration tests package](https://github.com/realm/realm-js/blob/main/integration-tests/tests/package.json) .. and you might need to [bump the version in `realm-web`](https://github.com/realm/realm-js/blob/main/packages/realm-web/package.json#L87) to avoid conflicting versions (which would potentially prevent `bson` from being hoisted to the root). We've had issues with value comparisons before, where the test app would use one version of `bson` while the SDK used a different one and `value instanceof bson.ObjectId` would be incorrectly false.

packages/realm/package.json
```diff
@@ -264,7 +264,7 @@
     "prebuild-install": "^7.1.1"
   },
   "peerDependencies": {
-    "bson": "^4",
+    "bson": "^5 || ^4",
```
elle-j, line 267: [v11](https://github.com/realm/realm-js/blob/49f3d84a5323347ee20749b09b363a9320c00ea9/package.json#L87) is pinned to `4.4.1`.

kraenhansen:
I'm surprised that users are seeing

> Unable to resolve "bson" from "node_modules/realm/dist/bundle.react-native.mjs"

if they just allow NPM to install `bson@v4` 🤔

packages/realm/package.json
takameyer: Good catch! ⚾

takameyer (approved):

packages/realm/package.json
kneth: I pinned it to 4.4.1 since 4.5.0 broke on React Native.

kneth (approved):

packages/realm/package.json
elle-j: Thanks for clarification!

elle-j:
Update: I've added to the PR description to wait with the merge until the related BSON changes are released.

takameyer:
I've updated the `peerDependency` to be `^4 || >=5.4.0`.  This should make sure we are always compatible and not have the polyfill issue

takameyer:
test

takameyer:
So one thing i've noted, is that if we have `bson` as a `peerDependency` than it will be required in the users project.  I think this is not our intention, as this would break for most users as soon as they updated, since in older `realm` versions, `bson` was just a normal dependency.  Therefore, I think we should change it to an `optionalDependency` to ensure that if a user is using their own version of `bson`, that it is within the compatible range for `realm`.

takameyer:
The React Native test are also failing for version 5.4.0.  I believe there are missing polyfills.  We may just have to revert to pinning this back to v4 and keeping it as an internal dependency.

takameyer (changes requested):
I think we need to review this further.  Upgrading to v5 breaks the integration tests for React Native.  I've created #6017 to ensure our users don't get errors thrown when installing `realm` regarding `bson`.  This pins `bson` to v4 to ensure compatibility.  We should look into this at another time #6018

elle-j:
Closing this (instead follow #6561).

PR #5944, opened 2023-06-27
Support nested updates of objects through `realm.create` via `UpdateMode.All`

### LINKED ISSUE #5933: UpdateMode doesn't get passed recursively when creating child objects in realm v12

Always

### Description

Realm v12.0.0-alpha.2 always use `UpdateMode.Never` for children objects ignoring mode arg in create function:

https://github.com/realm/realm-js/blob/d357fafbd53f8a57b3a00a34b2db7d98b38b09a4/packages/realm/src/TypeHelpers.ts#L231-L238

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

Code:
```js
import Realm, { UpdateMode } from "realm";

const ASchema = {
  name: "A",
  primaryKey: "id",
  properties: {
    id: "int",
  },
};

const BSchema = {
  name: "B",
  properties: {
    a: "A[]",
  },
};

const realm = new Realm({
  schema: [ASchema, BSchema],
  schemaVersion: 0,
  inMemory: true,
  path: "temp_db",
});

realm.write(() => {
  realm.create("A", {
    id: 1,
  });
  realm.create("B", {
    a: [{ id: 1 }],
  }, UpdateMode.All);
});
```

### Expected
Successful creation

### Actual
`Error: Attempting to create an object of type 'A' with an existing primary key value '1'., js engine: hermes`

### Version

12.0.0-alpha.2

Local Database only

### Are you using encryption?

No

React-native; Android, IOS

### Build environment

### Cocoapods version

_No response_

kraenhansen:
## What, How & Why?

This PR implements an initial solution for applying the `UpdateMode` to all objects when passed to `Realm.create()`.

This closes #5933.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -588,6 +588,8 @@ describe("Realm.Object", () => {
       });
 
       it("can be updated recursively", function (this: Mocha.Context & RealmContext) {
+        console.log("\n==== WRITE TRANSACTION: CREATE OBJECTS ====\n");
+
         // Create two mutual friends
         const { alice, bob } = this.realm.write(() => {
           const alice = this.realm.create(PersonWithId, {
@@ -605,6 +607,8 @@ describe("Realm.Object", () => {
           bob.friends.push(alice);
           return { alice, bob };
         });
+
+        console.log("\n==== WRITE TRANSACTION: CREATE W/ UPDATE MODE ALL ====\n");
 
         // Now update them
         this.realm.write(() => {
@@ -625,6 +629,10 @@ describe("Realm.Object", () => {
           );
         });
       });
+
+      // TODO: Add tests:
+      // * Child objects should also exist in a Dictionary and Set.
+      // * Child objects should also be embedded.
     });
   });
 
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -639,8 +639,8 @@ describe("Realm.Object", () => {
         });
 
         it("can be updated recursively in lists", function (this: Mocha.Context & RealmContext) {
-          this.realm.write(() => {
-            this.realm.create(
+          const alice = this.realm.write(() => {
+            return this.realm.create(
               CollectionSchema.name,
               {
                 _id: aliceId,
@@ -661,11 +661,22 @@ describe("Realm.Object", () => {
               Realm.UpdateMode.All,
             );
           });
+
+          expect(alice._id.equals(aliceId)).to.be.true;
+          expect(alice.name).to.equal("UpdatedAlice");
+
+          const bob = alice.list[0];
+          expect(bob._id.equals(bobId)).to.be.true;
+          expect(bob.name).to.equal("UpdatedBob");
+
+          const max = bob.list[0];
+          expect(max._id.equals(maxId)).to.be.true;
+          expect(max.name).to.equal("UpdatedMax");
         });
 
         it("can be updated recursively in dictionaries", function (this: Mocha.Context & RealmContext) {
-          this.realm.write(() => {
-            this.realm.create(
+          const alice = this.realm.write(() => {
+            return this.realm.create(
               CollectionSchema.name,
               {
                 _id: aliceId,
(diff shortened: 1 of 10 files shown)


packages/realm/src/Realm.ts
```diff
@@ -857,11 +859,15 @@ export class Realm {
...
-      throw new Error("Unsupported 'updateMode'. Only 'never', 'modified' or 'all' is supported.");
+      throw new Error(
+        `Unsupported 'updateMode'. Only '${UpdateMode.Never}', '${UpdateMode.Modified}' or '${UpdateMode.All}' is supported.`,
+      );
     }
     this.internal.verifyOpen();
     const helpers = this.classes.getHelpers(type);
+    this.currentUpdateMode = mode;
     const realmObject = RealmObject.create(this, values, mode, { helpers });
+    this.currentUpdateMode = undefined;
```
kraenhansen, line 870: This will never happen if `RealmObject.create` throws.

kraenhansen:
While having a state (`currentUpdateMode` on `Realm`) might work for now, I'd suggest avoiding storing and maintaining state for operations as it has a couple of drawbacks:
- It can cause weird bugs if not properly set between calls of the operation.
  - If the reset happen at the end, it might not happen if an exception is throw.
  - If it happens in the beginning (as with your suggestion), the deep dependent code has implicit dependencies on the state being sat before it's called, which isn't easily expressed in code and you'd need to rely on documentation.
- It will make it more difficult to transition the implementation into asynchronous code (i.e. make it harder to change if we wanted to), since that might introduce concurrent calls into the deep code and it can't reasonably expect the state to originate from its particular invocation.

All in all, I'd suggest refactoring to passing the update mode through argument, possibly by adding a property to the `CreationContext`.

elle-j:
@kraenhansen, very valuable input! We've discussed various issues and options on Slack (see our channel) with your input in mind. For now we've opted to move forward with the initial solution (but using `try-finally` for setting/unsetting `currentUpdateMode`), and creating a ticket for finding a more ideal solution for a later v12 release.

packages/realm/src/TypeHelpers.ts
```diff
@@ -73,7 +72,7 @@ export function toArrayBuffer(value: unknown, stringToBase64 = true) {
 
 /** @internal */
 export type TypeHelpers<T = unknown> = {
-  toBinding(value: T, createObj?: ObjCreator): binding.MixedArg;
+  toBinding(value: T, options?: { createObj?: ObjCreator; updateMode?: UpdateMode }): binding.MixedArg;
```
takameyer, line 75: Is the `updateMode` being set somewhere?  I haven't yet seen this in any of the `toBinding` calls.

integration-tests/tests/src/tests/objects.ts
```diff
@@ -586,6 +606,113 @@ describe("Realm.Object", () => {
...
+                    },
+                  },
+                },
+              },
+              Realm.UpdateMode.All,
+            );
+          });
+        });
+
+        it("can be updated recursively in sets", function (this: Mocha.Context & RealmContext) {
```
takameyer, line 691: That's amazing that this works for Sets and Dictionaries now.  
Should we also add assertions for the updated values?  We are just testing that this doesn't throw, should we also test that it does throw?

takameyer (approved):
Great!  Just a few clarifying questions/suggestions

packages/realm/src/TypeHelpers.ts
elle-j: Not yet, but that's the intention for the more ideal solution down the line. So I kept that refactor. (I'm mentioning it in [this TODO comment](https://github.com/realm/realm-js/blob/5e2caccff9c805205939bb4932e77d44eb6748f8/packages/realm/src/TypeHelpers.ts#L264).) Will also create a ticket and link to the ticket in that comment.

integration-tests/tests/src/tests/objects.ts
takameyer: Great!  Thanks 🎉

packages/realm/src/TypeHelpers.ts
takameyer: Perfect 🙏🏼

packages/realm/src/Realm.ts
elle-j: This has been updated to use `finally`. Also, see the PR comment further below addressing your longer comment 👍

kneth (approved):

PR #5955, opened 2023-06-29
Update `ClassMap` before notifying schema change listeners

### LINKED ISSUE #5574: Re-initialize the ClassMap when schema is updated (in a listener)
When the schema of a Realm is changed during runtime, the ClassMap should be updated. The schema change can come from the sync client.

Introduce a test for this.

elle-j:
## What, How & Why?

When calling `Realm._updateSchema()` in a write transaction, the schema change listener is fired after the call into Core, but before `_updateSchema()` returns (and thus before the write transaction is completed).

The `ClassMap` instance on the Realm was previously updated/reinitialized in `_updateSchema()`, but since the listener was fired before that, the `ClassMap` had not been updated in time. This PR moves that update into the `schemaDidChange` callback, updating it before notifying schema change listeners.

This closes #5574 

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm/src/ClassHelpers.ts
+++ b/packages/realm/src/ClassHelpers.ts
@@ -26,8 +26,6 @@ import {
   binding,
 } from "./internal";
 
-type BindingObjectSchema = binding.Realm["schema"][0];
-
 type ObjectWrapper = (obj: binding.Obj) => (RealmObject & DefaultObject) | null;
 
 /**
@@ -35,8 +33,7 @@ type ObjectWrapper = (obj: binding.Obj) => (RealmObject & DefaultObject) | null;
  */
 export type ClassHelpers = {
   constructor: RealmObjectConstructor;
-  // TODO: Use a different type, once exposed by the binding
-  objectSchema: BindingObjectSchema;
+  objectSchema: binding.ObjectSchema;
   properties: PropertyMap;
   wrapObject: ObjectWrapper;
   canonicalObjectSchema: CanonicalObjectSchema;
--- a/packages/realm/src/Realm.ts
+++ b/packages/realm/src/Realm.ts
@@ -601,7 +601,7 @@ export class Realm {
   public readonly syncSession: SyncSession | null;
 
   private schemaExtras: RealmSchemaExtra = {};
-  private classes: ClassMap;
+  private classes!: ClassMap;
   private changeListeners = new RealmListeners(this, RealmEvent.Change);
   private beforeNotifyListeners = new RealmListeners(this, RealmEvent.BeforeNotify);
   private schemaListeners = new RealmListeners(this, RealmEvent.Schema);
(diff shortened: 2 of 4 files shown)


packages/realm/src/ClassHelpers.ts
```diff
@@ -26,17 +26,14 @@ import {
...
 type ObjectWrapper = (obj: binding.Obj) => (RealmObject & DefaultObject) | null;
 
 /**
  * @internal
  */
 export type ClassHelpers = {
   constructor: RealmObjectConstructor;
-  // TODO: Use a different type, once exposed by the binding
-  objectSchema: BindingObjectSchema;
+  objectSchema: binding.ObjectSchema;
```
kraenhansen, line 36: Are these completely equivalent? It might not be a problem, but I vaguely remember it having optional properties for the properties which have defaults in Core, making it suitable as a type for something passed into Core whereas it's not as ideal for something read from Core.

packages/realm/src/Realm.ts
```diff
@@ -599,7 +599,7 @@ export class Realm {
   public readonly syncSession: SyncSession | null;
 
   private schemaExtras: RealmSchemaExtra = {};
-  private classes: ClassMap;
+  private classes!: ClassMap;
```
kraenhansen, line 602: I'm almost thinking that this drawback from deferring to a method at line 688 isn't worth it. As we don't expect to have to initialize the class map more than twice and it's not too complex, I would probably suggest duplicating and inlining it instead.

kraenhansen (approved):

papafe (approved):

packages/realm/src/Realm.ts
```diff
@@ -1237,6 +1242,13 @@ export class Realm {
     return this.classes.getHelpers<T>(arg);
   }
 
+  /**
+   * @internal
+   */
+  private updateClassMap() {
+    this.classes = new ClassMap(this, this.internal.schema, this.schema);
```
kneth, line 1249: For a large schema, it is an expensive operation to change even small things. Would be possible to apply changes to the class map and not start all over each time? Maybe not in this PR but creating an issue to track it?

kneth (approved):

packages/realm/src/ClassHelpers.ts
elle-j: I believe they should be equivalent. The type of  `binding.Realm["schema"]` is `readonly binding.ObjectSchema[]` from:
```ts
type BindingObjectSchema = binding.Realm["schema"][0];
```
There's usually the `_Relaxed` convention when deviating from what's needed for Core, but do you think there are types that we use that are not using that convention?

packages/realm/src/Realm.ts
elle-j: You make a good point, I'll inline those calls 👍

P.S. would be nice to add something in function signatures that tells TS that a certain property will be set. Similar to `asserts` or `is`. That would remove the risks of having to use `!` as in the above case.

packages/realm/src/Realm.ts
elle-j: Nice observation! It'd definitely be interesting to look into what we could do to optimize this. I'll create a ticket.

PR #5975, opened 2023-07-11
Refactor post-install script to not depend on fs-extra

kneth:
## What, How & Why?

No open issue for this.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/scripts/submit-analytics.mjs
+++ b/packages/realm/scripts/submit-analytics.mjs
@@ -52,7 +52,6 @@ import { createHmac } from "node:crypto";
 import { Buffer } from "node:buffer";
 
 import machineId from "node-machine-id";
-import fse from "fs-extra";
 
 import createDebug from "debug";
 export const debug = createDebug("realm:submit-analytics");
@@ -105,7 +104,9 @@ function getProjectRoot() {
  */
 function getPackageJson(packagePath) {
   const packageJson = path.resolve(packagePath, "package.json");
-  return fse.readJsonSync(packageJson);
+  const buffer = fs.readFileSync(packageJson);
+  const asJson = JSON.parse(buffer);
+  return asJson;
 }
 
 /**
@@ -127,8 +128,7 @@ function isAnalyticsDisabled() {
 }
 
 function getRealmVersion() {
-  const packageJsonPath = path.resolve(__dirname, "../package.json");
-  const packageJson = fse.readJsonSync(packageJsonPath);
+  const packageJson = getPackageJson(__dirname);
   return packageJson["version"];
 }
 
@@ -139,7 +139,7 @@ function getRealmVersion() {
  */
 function getRealmCoreVersion() {
   const dependenciesListPath = path.resolve(__dirname, "../dependencies.list");
-  const dependenciesList = fse
+  const dependenciesList = fs
     .readFileSync(dependenciesListPath)
     .toString()
     .split("\n")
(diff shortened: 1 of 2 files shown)


packages/realm/scripts/submit-analytics.mjs
```diff
@@ -105,7 +104,9 @@ function getProjectRoot() {
  */
 function getPackageJson(packagePath) {
   const packageJson = path.resolve(packagePath, "package.json");
-  return fse.readJsonSync(packageJson);
+  const buffer = fs.readFileSync(packageJson);
```
takameyer, line 107: Where is `fs` imported from?  Is it somehow set on the global?  Or should one import from `node:fs`?

takameyer (approved):

packages/realm/scripts/submit-analytics.mjs
elle-j: It's imported from `node:fs` 👍

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -105,7 +104,9 @@ function getProjectRoot() {
...
+  const asJson = JSON.parse(buffer);
```
elle-j, line 108: 1. Won't this throw if you try to parse a Buffer rather than a string?
   * I believe you could provide an `{ encoding: ".." }` 2nd arg to `readFileSync` in order to get a string back (see [various encodings](https://github.com/nodejs/node/blob/41f70568a46efceb59bdb00a727b4fe8d858cb2b/lib/buffer.js#L719)).
3. Maybe `asObject` is a more suitable name? Or perhaps return the result directly, your call 🙂

packages/realm/scripts/submit-analytics.mjs
kneth: line 44 imports `node:fs`

elle-j (approved):
An important fix, thanks! Just left a comment that could be good to address before merge 👍

packages/realm/scripts/submit-analytics.mjs
takameyer: Ah!  I didn't see there was something above the PR fold.  Good catch here.

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -105,7 +104,8 @@ function getProjectRoot() {
  */
 function getPackageJson(packagePath) {
   const packageJson = path.resolve(packagePath, "package.json");
-  return fse.readJsonSync(packageJson);
+  const buffer = fs.readFileSync(packageJson, "utf-8");
+  return JSON.parse(buffer);
```
elle-j, line 108: Minor naming suggestions:
```js
  const packageJsonPath = path.resolve(packagePath, "package.json");
  const packageJson = fs.readFileSync(packageJsonPath, "utf-8");
  return JSON.parse(packageJson);
```

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -127,8 +127,8 @@ function isAnalyticsDisabled() {
 }
 
 function getRealmVersion() {
-  const packageJsonPath = path.resolve(__dirname, "../package.json");
-  const packageJson = fse.readJsonSync(packageJsonPath);
+  const packageJsonPath = path.resolve(__dirname, "..", "package.json");
```
kraenhansen, line 130: 🎉

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -139,7 +139,7 @@ function getRealmVersion() {
  */
 function getRealmCoreVersion() {
   const dependenciesListPath = path.resolve(__dirname, "../dependencies.list");
```
kraenhansen, line 141: I believe I deleted this file 🤔

PR #5976, opened 2023-07-11
Remove type error on OrderedCollection

takameyer:
## What, How & Why?
* TypeScript was yelling since we weren't implementing [Symbol.unscopables]
* Set to empty object to remove error (this has no negatives effects)
* See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/unscopables for more info

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/src/OrderedCollection.ts
+++ b/packages/realm/src/OrderedCollection.ts
@@ -748,4 +748,12 @@ export abstract class OrderedCollection<T = unknown, EntryType extends [unknown,
       return DEFAULT_COLUMN_KEY;
     }
   }
+
+  /**
+   * This symbol was required for implementing ReadOnlyArray, but we don't need
+   * it for anything.  It's just here to make TypeScript happy.
+   * For more reading see:
+   * @see (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/unscopables)
+   */
+  readonly [Symbol.unscopables] = {};
 }


kneth (approved):

packages/realm/src/OrderedCollection.ts
```diff
@@ -748,4 +748,12 @@ export abstract class OrderedCollection<T = unknown, EntryType extends [unknown,
       return DEFAULT_COLUMN_KEY;
     }
   }
+
+  /**
+   * This symbol was required for implementing ReadOnlyArray, but we don't need
+   * it for anything.  It's just here to make TypeScript happy.
```
kraenhansen, line 754: @takameyer I don't think this is relevant as the public facing documentation though.

takameyer: Appears this was reverted anyway.  I'll look into why

elle-j: [Reverted](https://github.com/realm/realm-js/pull/5992#event-9822726832) this prior to the RC2 release due [enumeration errors](https://github.com/realm/realm-js/actions/runs/5528893972/jobs/10086388396). Have not had time to investigate further, but you can refer to the JS Slack thread 👍

takameyer: Very curious.  Thanks for the link.  Quite interesting that it only effects certain engines and platforms.  Perhaps it's binding related.  Node seems to be fine

PR #5978, opened 2023-07-12
Sync connection parameter bundleId

kneth:
## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm/bindgen/vendor/realm-core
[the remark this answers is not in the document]
kneth: Yes, I needed to expose `bundle_id`

packages/realm/src/platform/node/device-info.ts
```diff
@@ -17,8 +17,12 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import os from "node:os";
+import process from "node:process";
 
 import { version } from "realm/package.json";
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
elle-j, line 24: What error are you getting without these? Looks like TS doesn't yell for the line right above it.

packages/realm/src/platform/react-native/device-info.ts
```diff
@@ -19,6 +19,9 @@
 import { Platform } from "react-native";
 
 import { version } from "realm/package.json";
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
elle-j, line 23: I've got the same question as the previous one 🙂

packages/realm/src/platform/node/device-info.ts
kneth: Yeah, you might be right. It was yelling when I used a JavaScript file but switching to JSON might have helped.

packages/realm/src/platform/react-native/device-info.ts
kneth: Same answer

kraenhansen:
I'd suggest writing to the `realm/package.json` instead - that's guaranteed to always be present and therefore we can rely on it being there and not have to handle the special case of the postinstall script failing to produce the file (if #6038 gets merged and analytics is disabled when installing).

PR #5989, opened 2023-07-14
[useObject] Update primaryKeyRef on change (#5620)

### LINKED ISSUE #5620: @realm/react - useObject stops updating after repeated calls with the same ID

Sometimes

### Description

I apologize if this is something I'm doing wrong, but I can't seem to find any explanations for the behavior other than `useObject`. I looked at the source code and it does some hard-to-follow reference updating, so it doesn't seem out of the realm of possibility that something is going wrong with repeated calls of the same value, although I would have expected someone else would have encountered the same problem.

Expected Results:
Pass different IDs to `useObject` throughout the lifetime of a component and get back the corresponding Realm object.

Actual Results:
The value returned by `useObject` behaves as expected the first time any particular ID is passed as an argument. But the second or third time `useObject` receives an ID that it has previously received, it fails to update the return value, continuing to return whatever the previous object was.

### Stacktrace & log output

_No response_

Sometimes

### Reproduction Steps

I encountered it in a project I'm working on and reproduced it in another repo (based on the Realm Expo template) with a really basic setup as seen below. Basically just make a FlatList where clicking on an item will render a separate component which takes the `id` as a prop and then passes it to `useObject`. If you click around, the behavior is erratic:
- If I click through all items, each item will update as expected the first time.
- Typically exactly one repeated call to `useObject` will update with the appropriate data.
- After that, the value in `useObject` seems to be mostly "stuck" wherever it is.
- Occasionally an update will occur as expected after several calls to `useObject` (eg clicking the list items a bunch in my example).

```js
const PersonItem = ({ name, onPress }) => {
  return (
    <Pressable onPress={onPress}>
      <Text>
        {name}
      </Text>
    </Pressable>
  )
}

const ViewPerson = ({ personId }) => {
  const person = useObject(Person, personId)
  return <Text>{person.personName}</Text>
}

const BrowsePeople = () => {
  const people = useQuery(Person)
  const [selected, setSelected] = useState<BSON.ObjectId>()

  return (
    <View>
      <FlatList
        data={people}
        keyExtractor={el => el._id.toHexString()}
        renderItem={(el) => 
          <PersonItem  
            name={el.item.personName}
            onPress={() => setSelected(el.item._id)}
          />}
      />
      {selected && <ViewPerson personId={selected} />}
    </View>
  )
}
```

### Version

0.4.3, 0.4.1

### Are you using encryption?

No

expo 48.0.6, react-native 0.71.4, realm 11.5.2

### Build environment

### Cocoapods version

_No response_

RS1-Project:
## What, How & Why?
Within the `useObject` hook, the internal ref `primaryKeyRef` is never updated, causing successive calls to `arePrimaryKeysIdentical` to fail. 

This causes two issues:
- If `useObject`  is called with a primary key `a`, then again with `b`, and then again with `a`, it will return the object for the primary key `b`.
- After a primary key change, `arePrimaryKeysIdentical` will always return false, potentially causing the cached object to be recreated on each render.

Updating `primaryKeyRef` after `arePrimaryKeysIdentical` fails solves these two issues, and `useObject` works as expected.

This closes #5620 

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests
* [X] 🔀 Executed flexible sync tests locally if modifying flexible sync
* [X] 📝 Public documentation PR created or is not necessary

### DIFF
--- a/packages/realm-react/src/useObject.tsx
+++ b/packages/realm-react/src/useObject.tsx
@@ -90,6 +90,8 @@ export function createUseObject(useRealm: () => Realm) {
             updatedRef,
           });
           originalObjectRef.current = originalObject;
+          // Update the primaryKeyRef, so we can check if the primaryKey has changed on the next render
+          primaryKeyRef.current = primaryKey;
         }
         return cachedObjectRef.current;
       },


RS1-Project:
@cla-bot check

takameyer:
@RS1-Project Thanks of submitting the PR.  I'll need to write some regression tests and if this solves them, we can get this merged.  
If you want to update the tests, feel free, that would help this land sooner.

takameyer:
@RS1-Project I took over the PR to add tests.  I added an additional reference update that would signal to the component that the object returned from `useObject` has been updated.  Tests are now green.
Thanks for you contribution!
https://github.com/realm/realm-js/pull/6084

PR #5998, opened 2023-07-18
Set iterator tests

takameyer:
## What, How & Why?
* Tests for `entries`, `values`, `keys`, `foreach`, and `for in`
* Comments on what fails

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/tests/set.ts
+++ b/integration-tests/tests/src/tests/set.ts
@@ -68,6 +68,64 @@ describe("Set", () => {
         const success2 = this.realm.write(() => set.delete("hello"));
         expect(success2).equals(false);
       });
+      it("is iterable", function (this: RealmContext) {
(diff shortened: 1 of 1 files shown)


kraenhansen:
Aren't this covered with https://github.com/realm/realm-js/blob/main/integration-tests/tests/src/tests/iterators.ts?

takameyer:
@kraenhansen Looks like there's a todo for `keys`, `entries` and `values`, but this is place we should test for the correct behavior.  I somehow missed it, as I only looked at `Set`.
Although it is a bit misleading that `Set` is an unordered collection that inherits `OrderedCollection`.  Guessing there were already discussions about this.

PR #6012, opened 2023-07-28
Make the react-native entry point Jest compatible

### LINKED ISSUE #6003: Fix Realm resolver for Jest in v12
Currently Jest will resolve into the React Native bundle instead of the Node bundle.  To workaround this, users would have to create their own custom resolver in their application code in order to write tests using Realm.  
This should be fixed before the v12 release as to not break any tests our users have written.

takameyer:
## What, How & Why?
* Remove the need to create a custom resolver in jest code testing realm
* This ensures there are no breaking changes in current users tests

This closes #6003

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated


packages/realm/index.react-native.js
```diff
@@ -19,4 +19,19 @@
...
+  // Define a require function that will load the node bundle
+  // otherwise, metro will preemptively load the node bundle
+  const nodeRequire = require;
+  // Jest is running, use the node bundle
+  entryPoint = nodeRequire("./dist/bundle.node");
+} else {
+  entryPoint = require("./dist/bundle.react-native");
+}
+
+module.exports = entryPoint.Realm;
```
kraenhansen, line 37: I'd suggest simply inlining this assignment. No need to store it in `entryPoint`.

packages/realm/index.react-native.js
```diff
@@ -19,4 +19,19 @@
...
+const isJest = process?.env?.JEST_WORKER_ID !== undefined;
+
+let entryPoint;
+
+if (isJest) {
+  // Define a require function that will load the node bundle
+  // otherwise, metro will preemptively load the node bundle
+  const nodeRequire = require;
+  // Jest is running, use the node bundle
+  entryPoint = nodeRequire("./dist/bundle.node");
```
kraenhansen, line 32: I'm a bit sad that we need to do this 😞 It would be better if we could find a way to keep this pure from runtime checks of the environment ... historically we've been bitten by the brittleness of these many times before.

packages/realm/index.react-native.js
```diff
@@ -19,4 +19,19 @@
 /* eslint-disable @typescript-eslint/no-var-requires -- We're exporting using CJS assignment */
 /* eslint-env commonjs */
 
-module.exports = require("./dist/bundle.react-native").Realm;
+// eslint-disable-next-line no-undef -- In React Native, process is not defined, but in Jest it is
+const isJest = process?.env?.JEST_WORKER_ID !== undefined;
```
kraenhansen, line 23: I'm a bit sad that we need to do this 😞 It would be better if we could find a way to keep this pure from runtime checks of the environment ... historically we've been bitten by the brittleness of these many times before.

takameyer: We can get around the `react-native` preset for `jest`, which is basically ensuring the `react-native` entry point is being loaded.  I did a lot of digging around this and this was the best solution I could find.  At least it's very dependent on Jest and has very little impact on the experience.

takameyer: I need to refine the comment. `process` is actually defined in `react-native`, but for some reason it is not typed.

packages/realm/index.react-native.js
takameyer: Good idea.  I was doing a lot of dancing around how metro bundled this, to ensure it didn't pre-load the node entry point.  Setting `require` to a variable was the solution.

packages/realm/index.react-native.js
kraenhansen: > which is basically ensuring the react-native entry point is [being loaded](https://github.com/facebook/react-native/blob/main/packages/react-native/jest/react-native-env.js)

As per that link, it looks to me that we could probably export the Node.js bundle via the `require` condition and avoid the runtime check?

kraenhansen: Add this line after https://github.com/realm/realm-js/blob/0b9dd8b31371d6cd4fe9e39bf06022b219756415/packages/realm/package.json#L31

```
      "require": "./index.node.js",
```

Issue #5903, opened 2023-06-17
[Expo] Realm Errors are not propagated to console


haikov:
### How frequently does the bug occur?

Always

### Description

Hey!

I'm building an app with Expo and Realm, and one of the biggest issues I'm facing with realm is inability to understand what's going wrong in case of issues with schema, query (or anything else related to Realm), since Realm always throws the same error in console which gives no details. The error looks like that:

```
 ERROR  Error: Exception in HostFunction: <unknown>

This error is located at:
    in Unknown (created by App)
    in Provider (created by App)
    in RCTView (created by View)
    in View (created by GestureHandlerRootView)
    in GestureHandlerRootView (created by App)
    in App (created by RootApp)
    in ReactNativeProfiler (created by RootApp)
    in RCTView (created by View)
    in View (created by __Sentry.TouchEventBoundary)
    in __Sentry.TouchEventBoundary (created by RootApp)
    in RootApp (created by withDevTools(RootApp))
    in withDevTools(RootApp)
    in RCTView (created by View)
    in View (created by AppContainer)
    in RCTView (created by View)
    in View (created by AppContainer)
    in AppContainer
    in main(RootComponent), js engine: hermes
```

Is this something known or maybe I'm missing some configuration step to make the output cleaner? Or is there some better way to look up the detailed stack trace?

### Can you reproduce the bug?

Always

### Version

11.9.0

### What services are you using?

Local Database only

### Are you using encryption?

No

### Platform OS and version(s)

iOS 16

### Build environment

Expo SDK 48

kraenhansen:
I tried but failed to reproduce this. I do see "Exception in HostFunction:", but I get a proper message.
Can you help reproduce this by sharing a minimal repository exhibiting this behaviour?

haikov:
Tried reproducing it with an example repo, no luck, works as expected there. So it must be something with my project setup. I'm trying to dig into that a bit more and will report back as soon as I get anything. But one more thing which I see when Realm emits an error is this weird log:

```
Error: ENOENT: no such file or directory, open '/Users/path/to/project/folder/JavaScript'
    at Object.openSync (node:fs:585:3)
    at Object.readFileSync (node:fs:453:35)
    at getCodeFrame (/Users/path/to/project/folder/node_modules/metro/src/Server.js:1004:18)
    at Server._symbolicate (/Users/path/to/project/folder/node_modules/metro/src/Server.js:1073:22)
[...]

PR #6017, opened 2023-07-31
Make BSON an optional dependency

takameyer:
## What, How & Why?
* Users trying out the release candidates of v12 reported errors being thrown for the `bson` dependency
* In v11 and earlier, `bson` was a dependency of `realm` and not a `peerDependency`.
* In v12 `bson` was added as a `peerDependency`, causing the issue.

This PR reverts the changes to v12 regarding `bson`, but also adds `bson` as an optional dependency, freezing it to a v4 if it's installed locally, since v5 is not directly compatible with `react-native`

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/package.json
+++ b/packages/realm/package.json
@@ -263,13 +263,16 @@
     }
   },
   "dependencies": {
+    "bson": "^4.7.2",
     "debug": "^4.3.4",
     "node-fetch": "^2.6.9",
     "node-machine-id": "^1.1.12",
     "prebuild-install": "^7.1.1"
   },
+  "optionalDependencies": {
+    "bson": "^4"
+  },
   "peerDependencies": {
-    "bson": "^4",
     "react-native": ">=0.71.0"
   },
   "peerDependenciesMeta": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -2982,7 +2982,6 @@
       "version": "10.5.0",
       "resolved": "https://registry.npmjs.org/@chevrotain/cst-dts-gen/-/cst-dts-gen-10.5.0.tgz",
       "integrity": "sha512-lhmC/FyqQ2o7pGK4Om+hzuDrm9rhFYIJ/AXoQBeongmn870Xeb0L6oGEiuR8nohFNL5sMaQEJWCxr1oIVIVXrw==",
-      "dev": true,
       "dependencies": {
         "@chevrotain/gast": "10.5.0",
         "@chevrotain/types": "10.5.0",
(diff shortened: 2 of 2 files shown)


packages/realm/package.json
```diff
@@ -263,13 +263,16 @@
     }
   },
   "dependencies": {
+    "bson": "^4.7.2",
     "debug": "^4.3.4",
     "node-fetch": "^2.6.9",
     "node-machine-id": "^1.1.12",
     "prebuild-install": "^7.1.1"
   },
+  "optionalDependencies": {
```
kraenhansen, line 272: I don't think this is the intended use of `optionalDependencies` as per [this description](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies):

> If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the optionalDependencies object.

I don't see a valid case where the installation of the `bson` package would fail, as this happens mostly if it fails to download or build a binary for an unsupported target OS or similar failure during a `postinstall` scripts execution.

kraenhansen: What you might be looking for is the [`peerDependenciesMeta `](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta) which can mark a peer dependency as optional, but I don't think that's what we want, since the internals of our SDK relies on it being installed.

kraenhansen:
> In v12 bson was added as a peerDependency, causing the issue.

What issue is this causing? A link would be great 👍

packages/realm/package.json
takameyer: I've been experimenting with this, and have also found that the `peerDependency` is the correct way of going about this.  I was under the assumption that `npm` wouldn't install `bson` on `npm install`, but it does.

I think in it's current form on `main` the package.json is fine.  We should stay on v4 for now, as the react-native tests fail using v5

takameyer:
After trying out `peerDependencies` and `optionalDependencies`, it seems going optional is not the correct path.  It appears the issue described [here](https://github.com/realm/realm-js/issues/5903#issuecomment-1604310038) has already been solved by other methods.  My main concern was that users would be forced to `npm install bson` in their projects, which would lead them to installing v5, which has some compatibility issues with react native. 
see https://github.com/realm/realm-js/actions/runs/5711929450/job/15474655575

packages/realm/package.json
kraenhansen: Yeah - a missing peer dependency is automatically installed since NPM v7.

PR #6032, opened 2023-08-03
Flaky tests

takameyer:
* Looking into flaky tests

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/integration-tests/tests/src/tests/sync/user.ts
+++ b/integration-tests/tests/src/tests/sync/user.ts
@@ -68,7 +68,8 @@ function removeExistingUsers(): void {
     if (this.app) {
       const users = this.app.allUsers;
       for (const userId of Object.keys(this.app.allUsers)) {
-        await this.app.removeUser(users[userId]);
+        // await this.app.removeUser(users[userId]);
+        await this.app.deleteUser(users[userId]);
       }
     }
   });
--- a/integration-tests/tests/src/tests/sync/user.ts
+++ b/integration-tests/tests/src/tests/sync/user.ts
@@ -16,7 +16,6 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 import { expect } from "chai";
-import chaiAsPromised from "chai-as-promised";
 import { importAppBefore } from "../../hooks";
 import {
   randomNonVerifiableEmail,
@@ -68,8 +67,11 @@ function removeExistingUsers(): void {
     if (this.app) {
       const users = this.app.allUsers;
       for (const userId of Object.keys(this.app.allUsers)) {
-        // await this.app.removeUser(users[userId]);
-        await this.app.deleteUser(users[userId]);
+        if (users[userId].isLoggedIn) {
+          await this.app.deleteUser(users[userId]);
+        } else {
+          await this.app.removeUser(users[userId]);
+        }
       }
     }
   });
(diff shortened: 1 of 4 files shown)


integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -67,7 +67,11 @@ function removeExistingUsers(): void {
     if (this.app) {
       const users = this.app.allUsers;
       for (const userId of Object.keys(this.app.allUsers)) {
-        await this.app.removeUser(users[userId]);
+        if (users[userId].isLoggedIn) {
+          await this.app.deleteUser(users[userId]);
+        } else {
+          await this.app.removeUser(users[userId]);
+        }
```
elle-j, line 74: Why was this change needed? From the TS docs, `removeUser()` will handle logging out the user.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -102,43 +106,60 @@ describe.skipIf(environment.missingServer, "User", () => {
...
+    // any email address is valid and registerUser() will never fail -- the test doesn't make sense
+    it.skip("invalid email, valid password", async function (this: AppContext & RealmContext) {
+      const credentials = Realm.Credentials.emailPassword({ email: invalidEmail, password: validPassword });
       await expect(this.app.logIn(credentials)).to.be.rejectedWith("invalid username/password"); // this user does not exist yet
       expect(
         this.app.emailPasswordAuth.registerUser({ email: invalidEmail, password: validPassword }),
-      ).to.be.rejectedWith(`failed to confirm user "${invalidEmail}"`);
+      ).to.eventually.be.rejectedWith(`failed to confirm user "${invalidEmail}"`);
       await expect(this.app.logIn(credentials)).to.be.rejectedWith("invalid username/password"); // this user did not register
+    });
```
elle-j, line 137: Looks like this test belongs in a "custom function" test. I agree with the comment that for an auto confirm app, the login should not be rejected.

elle-j (approved):
🟢

integration-tests/tests/src/tests/sync/user.ts
takameyer: This was an attempt to fix the test when I was unsure why a user object was being returned for something that should have failed.  The first assumption was that the user actually did exist.  This can be removed

elle-j: Ah okay, thanks for explaining!

elle-j, line 74: Suggesting this change as per [this thread](https://github.com/realm/realm-js/pull/6032#discussion_r1331456122).
```suggestion
        await this.app.removeUser(users[userId]);
```

integration-tests/tests/src/tests/sync/user.ts
kneth: We should remove the test instead of skipping it

elle-j: Yeah, you could probably delete that entire `it` test. Additionally, perhaps after the `"custom confirmation function works"` you could add the following (which is pretty much the same test):

```ts
it("custom confirmation function rejects invalid email", async function (this: AppContext & RealmContext) {
      const invalidEmail = randomNonVerifiableEmail();
      const validPassword = "password123456";

      const credentials = Realm.Credentials.emailPassword({ email: invalidEmail, password: validPassword });
      await expect(this.app.logIn(credentials)).to.be.rejectedWith("invalid username/password"); // this user does not exist yet
      expect(
        this.app.emailPasswordAuth.registerUser({ email: invalidEmail, password: validPassword }),
      ).to.eventually.be.rejectedWith(`failed to confirm user "${invalidEmail}"`);
      await expect(this.app.logIn(credentials)).to.be.rejectedWith("invalid username/password"); // this user did not register
    });
```

I know that `describe` suite is currently skipped, but at least we still keep the test around.

PR #6034, opened 2023-08-03
Fix concurrency issue with workflow

takameyer:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/package-unit-tests.yml
+++ b/.github/workflows/package-unit-tests.yml
@@ -14,6 +14,10 @@ on:
       # Run this on updates to the workflow
       - ".github/workflows/package-unit-tests.yml"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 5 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ${{ github.head_ref }}-${{ github.run_id }}
```
kraenhansen, line 10: @takameyer I believe this was taken from the official docs.

Perhaps we should use the alternative example (using `workflow` instead of `head_ref`) there? https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

takameyer: Yeah, I think something is not quite correct here, so I opened this to experiment.  I was noticing that every now and again the newest push would just not run the tests for some reason.  I'll test this out more on Monday

kraenhansen: This is an important detail from the docs I didn't initially knew nor expected:

> If you have multiple workflows in the same repository, concurrency group names must be unique across workflows to avoid canceling in-progress jobs or runs from other workflows. Otherwise, any previously in-progress or pending job will be canceled, regardless of the workflow.

takameyer:
The strategy appears to be sound.  Now any pushes to a branch will cancel the previously spawned actions, and pushes to other branches will not effect others.

PR #6043, opened 2023-08-07
Move bson to a normal dependency to prevent issues in yarn

### LINKED ISSUE #6040: v12 RC with Yarn: Unable to resolve "bson" from "node_modules/realm/dist/bundle.react-native.js"

Always

### Description

Unable to resolve "bson" from "node_modules/realm/dist/bundle.react-native.js"

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

Clean expo project, make a dev build, add     "realm": "^12.0.0-rc.3",

### Version

12.0.0-rc.3

-- select --

### Are you using encryption?

No

expo 49 0.72.3

### Build environment

### Cocoapods version

_No response_

takameyer:
## What, How & Why?
When using `yarn` as a dependency manager, `peerDependencies` are not automatically installed.  This causes errors with our `bson` dependency, as many users will not have this package as a direct dependency.  
Therefore `bson` has been removed from the `peerDependencies` and is now a direct dependency of `realm`.  This has the effect that users can't enforce which version of `bson` Realm should use, but it's a better compromise than getting runtime crashes where they are forced to add `bson` as a dependency.

This closes #6040

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@
 * None
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None
+* Fix issues with `yarn` and the `bson` dependency. ([#6040](https://github.com/realm/realm-js/issues/6040))
 
 ### Compatibility
 * React Native >= v0.71.4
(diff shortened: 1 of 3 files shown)


kraenhansen (approved):
Sad that we need to do this, but it makes sense 👍

PR #6061, opened 2023-08-11
Provide guidance when Realm binary is missing

### LINKED ISSUE #5981: Create a meaningful error message when the `Realm` binary is not bound to the JS environment
Currently when users of `v12` end up in a state where `Realm` is not compiled/available in their environment (for example: for got to `pod install` or tried `expo web` or `expo go`), the error returned is:
```
TypeError: Cannot read property 'Helpers_get_table' of undefined
```
This is misleading and can lead to frustration, as it is very unclear what the issue actually is.  We should create a check for this problem and respond with a meaningful error message such as:
```
Realm could not be found in your environment.  Make sure you have followed the setup instructions at https://www.mongodb.com/docs/realm/sdk/react-native/install/
```

We also could create a `TROUBLESHOOTING.md` which goes provides some general guidance specific to these type of errors.  If this is done, than it should be linked in instead of the mdb installation docs.

takameyer:
## What, How & Why?
It was not clear when the Realm binary was missing and how to fix it.  This provides a pre-check for the Realm binary and throws an error messaging reporting what the problem is.  A section has been added to the `README.md` which provides various tips to fix the problem

This closes #5981

image

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

* [ ] jsdoc files updated

### DIFF
--- a/packages/realm/bindgen/src/templates/node-wrapper.ts
+++ b/packages/realm/bindgen/src/templates/node-wrapper.ts
@@ -46,6 +46,9 @@ export function generate({ rawSpec, spec: boundSpec, file }: TemplateContext): v
     // TODO: Remove the need to store Realm as a global
     // @see https://github.com/realm/realm-js/issues/2126
     const nativeModule = global.__RealmFuncs;
+    if(!nativeModule) {
+      throw new Error("Could not find the Realm binary. Please consult our troubleshooting guide. (https://www.mongodb.com/docs/realm-sdks/js/latest/#md:troubleshooting-missing-binary)");
+    }
 
     export const WeakRef = global.WeakRef ?? class WeakRef {
         constructor(obj) { this.native = nativeModule.createWeakRef(obj) }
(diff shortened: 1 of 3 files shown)


kraenhansen (approved):

packages/realm/bindgen/src/templates/node-wrapper.ts
```diff
@@ -46,6 +46,9 @@ export function generate({ rawSpec, spec: boundSpec, file }: TemplateContext): v
     // TODO: Remove the need to store Realm as a global
     // @see https://github.com/realm/realm-js/issues/2126
     const nativeModule = global.__RealmFuncs;
+    if(!nativeModule) {
+      throw new Error("Could not find the Realm binary. Please consult our troubleshooting guide. (https://www.mongodb.com/docs/realm-sdks/js/latest/#md:troubleshooting-missing-binary)");
```
kraenhansen, line 50: I'm not sure, but perhaps some terminals would have an easier time constructing a link if it's not wrapped in parentheses. Just a thought.

```suggestion
      throw new Error("Could not find the Realm binary. Please consult our troubleshooting guide: https://www.mongodb.com/docs/realm-sdks/js/latest/#md:troubleshooting-missing-binary");
```

takameyer: When I tested it, it worked fine, but maybe a good idea if people need to actually manually copy and paste it out.

PR #6076, opened 2023-08-18
Add migrationOptions.resolveEmbeddedConstraints

### LINKED ISSUE #6073: Orphaned embedded objects are not deleted when migrating from Object to Embedded Object

Always

### Description

This issue has previously been noted and resolved in realm-java and realm-kotlin:

https://github.com/realm/realm-java/issues/7769
https://github.com/realm/realm-kotlin/issues/1464

Orphaned documents should be handled automatically according to this change in realm-core:
https://github.com/realm/realm-core/pull/5737

The issue occurs when migrating a schema for a local realm. Version 2 of the schema has three object models defined. Version 3 of the schema embeds each of these three object models into one another in order to take advantage of cascading deletes. When the existing realm is attempted to be migrated from version 2 to 3, the following error is received:

"Failed to open local Realm for {REALM_INSTANCE} - Error: ","args":["Cannot convert '{OBJECT}' to embedded: at least one object has no incoming links and would be deleted."]

### Stacktrace & log output

```shell
"Failed to open local Realm for {REALM_INSTANCE} - Error: ","args":["Cannot convert '{OBJECT}' to embedded: at least one object has no incoming links and would be deleted."]
```

Always

### Reproduction Steps

_No response_

### Version

11.3.1

Local Database only

### Are you using encryption?

-- select --

Electron 22.3.14

### Build environment

### Cocoapods version

_No response_

kneth:
## What, How & Why?

This closes #6073 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/migrations.ts
+++ b/integration-tests/tests/src/tests/migrations.ts
@@ -521,4 +521,75 @@ describe("Migrations", () => {
       expect(objects[1].values[2]).equals(5);
     });
   });
+
+  describe("Migrate to embedded objects", () => {
+    const ParentSchema = {
+      name: "Parent",
+      primaryKey: "id",
+      properties: {
+        id: "int",
+        child: "Child",
+      },
+    };
+
+    const ChildSchema = {
+      name: "Child",
+      embedded: false,
+      properties: {
+        value: "int",
+      },
+    };
+
+    const EmbeddedChildSchema = {
+      name: "Child",
+      embedded: true,
+      properties: {
+        value: "int",
+      },
+    };
+
+    beforeEach(() => {
+      Realm.clearTestState();
+
+      const realm = new Realm({
+        schema: [ParentSchema, ChildSchema],
+        schemaVersion: 1,
+      });
+
+      realm.write(() => {
+        const child1 = realm.create(ChildSchema.name, { value: 11 });
+        const child2 = realm.create(ChildSchema.name, { value: 22 });
+        realm.create(ParentSchema.name, { id: 1, child: child1 });
+        realm.create(ParentSchema.name, { id: 2, child: child2 });
+      });
+
+      realm.write(() => {
+        realm.delete(realm.objectForPrimaryKey(ParentSchema.name, 1));
+      });
+
+      realm.close();
+    });
+
+    it("migration should delete orphan objects", () => {
+      const realm = new Realm({
+        schema: [ParentSchema, EmbeddedChildSchema],
+        schemaVersion: 2,
+        automaticallyHandleBacklinksInMigration: true,
+      });
+      expect(realm.objects(ParentSchema.name).length).equals(1);
+      realm.close();
+    });
+
+    it("orphan objects lead to exception", () => {
+      expect(() => {
+        new Realm({
+          schema: [ParentSchema, EmbeddedChildSchema],
+          schemaVersion: 2,
+          automaticallyHandleBacklinksInMigration: false,
+        });
+      }).to.throw(
+        `Cannot convert '${EmbeddedChildSchema.name}' to embedded: at least one object has no incoming links and would be deleted.`,
+      );
+    });
+  });
 });
--- a/integration-tests/tests/src/tests/migrations.ts
+++ b/integration-tests/tests/src/tests/migrations.ts
@@ -579,7 +579,9 @@ describe("Migrations", () => {
       const realm = new Realm({
         schema: [ParentSchema, EmbeddedChildSchema],
         schemaVersion: 2,
-        autoResolveEmbeddedConstraintsInMigration: true,
+        migrationOptions: {
+          resolveEmbeddedConstraints: true,
+        },
       });
       expect(realm.objects(ParentSchema.name).length).equals(1);
       realm.close();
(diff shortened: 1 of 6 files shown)


integration-tests/tests/src/tests/migrations.ts
```diff
@@ -521,4 +521,75 @@ describe("Migrations", () => {
...
+      });
+
+      realm.close();
+    });
+
+    it("migration should delete orphan objects", () => {
+      const realm = new Realm({
+        schema: [ParentSchema, EmbeddedChildSchema],
+        schemaVersion: 2,
+        automaticallyHandleBacklinksInMigration: true,
```
jcarpe, line 577: 👏

packages/realm/src/Configuration.ts
```diff
@@ -47,6 +50,15 @@ export type BaseConfiguration = {
   disableFormatUpgrade?: boolean;
   encryptionKey?: ArrayBuffer | ArrayBufferView | Int8Array;
   onMigration?: MigrationCallback;
+  /**
+   * A flag to indicate whether Realm should resolve
+   * embedded object constraints after migration. If this is `true` then all embedded objects
+   * without a parent will be deleted and every embedded object with multiple references to it
```
jcarpe, line 56: Nit - is it "multiple"? Or would this apply even for a single reference (I think as the test implies)? May suggest saying: "...every embedded object with one or more references..."

packages/realm/src/Configuration.ts
```diff
@@ -47,6 +50,15 @@ export type BaseConfiguration = {
...
+   * will be duplicated so that every referencing object will hold its own copy of the
+   * embedded object.
+   * @default false
+   */
```
takameyer, line 60: I'm noticing that this seems to be the only configuration option that is documented.  I believe we need a follow-up ticket to document the realm configuration in code.

takameyer (approved):
LGTM :shipit:

packages/realm/src/Configuration.ts
```diff
@@ -47,6 +50,15 @@ export type BaseConfiguration = {
...
+  automaticallyHandleBacklinksInMigration?: boolean;
```
elle-j, line 61: Wondering if it'd be better to name this more in line with how you're describing it in the docs. Perhaps:
* `autoResolveEmbeddedConstraintsInMigration`
* `autoHandleEmbeddedConstraintsInMigration`
* (ps. Core uses the plural of Migration.)

Thoughts?

elle-j (approved):
Nice! Only got a naming related comment.

packages/realm/src/Configuration.ts
kneth: Yep, I noticed too. I have created https://github.com/realm/realm-js/issues/6081

packages/realm/src/Configuration.ts
kneth: I plead guilty: I copied from Kotlin :smile:

Your formulation sounds more accurate, and I will update before merging.

packages/realm/src/Configuration.ts
elle-j: Sounds good to me 👍

packages/realm/src/Configuration.ts
```diff
@@ -109,6 +130,7 @@ export function validateConfiguration(config: unknown): asserts config is Config
   }
   if (sync !== undefined) {
     assert(!onMigration, "The realm configuration options 'onMigration' and 'sync' cannot both be defined.");
+    assert(!migrationOptions, "The realm configuration options 'migrationOptions' and 'sync' cannot both be defined.")
```
takameyer, line 133:
```suggestion
    assert(!migrationOptions, "The realm configuration options 'migrationOptions' and 'sync' cannot both be defined.");
```
;)

takameyer (approved):
:shipit:

packages/realm/src/Realm.ts
```diff
@@ -547,6 +547,9 @@ export class Realm {
         encryptionKey: Realm.determineEncryptionKey(config.encryptionKey),
         syncConfig: config.sync ? toBindingSyncConfig(config.sync) : undefined,
         forceSyncHistory: config.openSyncedRealmLocally,
+        automaticallyHandleBacklinksInMigrations: config.migrationOptions
+          ? config.migrationOptions.resolveEmbeddedConstraints === true
+          : false,
```
elle-j, line 552: One alternative. (Feel free to choose whichever. If going with the one below, double check linting is happy 🙂)
```suggestion
        automaticallyHandleBacklinksInMigrations: config.migrationOptions?.resolveEmbeddedConstraints ?? false,
```

elle-j (approved):

elle-j:
Addition: Also export `MigrationOptions` in [index.ts](https://github.com/realm/realm-js/blob/4715caa2283d8903131513bb635026995168e410/packages/realm/src/index.ts#L108) and on the [Realm namespace](https://github.com/realm/realm-js/blob/4715caa2283d8903131513bb635026995168e410/packages/realm/src/Realm.ts#L1462).

PR #6079, opened 2023-08-21
Update documentation to include default option for behavior

takameyer:
## What, How & Why?
Just noticed there was no default documented while reading the new documentation for NodeJS.  I've updated here and requested a note in the mdb docs.

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/src/app-services/MutableSubscriptionSet.ts
+++ b/packages/realm/src/app-services/MutableSubscriptionSet.ts
@@ -67,6 +67,7 @@ export type SubscriptionOptions = {
   throwOnUpdate?: boolean;
   /**
    * Specifies how to wait or not wait for subscribed objects to be downloaded.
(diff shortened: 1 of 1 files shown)


packages/realm/src/app-services/MutableSubscriptionSet.ts
```diff
@@ -67,6 +67,7 @@ export type SubscriptionOptions = {
   throwOnUpdate?: boolean;
   /**
    * Specifies how to wait or not wait for subscribed objects to be downloaded.
+   * By default, the behavior is `WaitForSync.Never`.
```
elle-j, line 70:
```suggestion
   * @default WaitForSync.Never
```

elle-j (approved):
Thanks!

PR #6088, opened 2023-08-22
Add flag to keep Realm open on RealmProvider unmount

### LINKED ISSUE #6023: `RealmProvider` should have an option like `keepRealmOpen`
### Problem

Hi!

I apologize if this can be achieved with the current options that `RealmProvider` has, but I could not find how to make my use case to work with it. Please let me know if there is a way to get it working.

I'm using:
React Native 0.72.3
Realm: 11.10.1
@realm/react 0.5.1

This is my problem:

I have two realms in my app, one is public, the other is private (requires a logged user), none of them uses sync. So, I'm using `createRealmContext`. I have some RN headless tasks that will use both realms, these tasks can run on background or foreground, but when I tried to use the realms sometimes I was getting the error `Cannot access a realm that is closed`. I could not find a way to keep my realm instances independent, so that if the `RealmProvider` closes it the one being used by the headless task keeps open. What I had to do was to patch-package. This is what I changed in `node_modules/@realm/react/dist/index.js`:
```patch
--- a/node_modules/@realm/react/dist/index.js
+++ b/node_modules/@realm/react/dist/index.js
@@ -665,26 +667,6 @@ function createRealmProvider(realmConfig, RealmContext) {
                 realmRef.current = realm;
             }
         }, [realm]);
-        useEffect(() => {
-            const realmRef = currentRealm.current;
-            // Check if we currently have an open Realm. If we do not (i.e. it is the first
-            // render, or the Realm has been closed due to a config change), then we
-            // need to open a new Realm.
-            const shouldInitRealm = realmRef === null;
-            const initRealm = async () => {
-                const openRealm = await Realm.open(configuration.current);
-                setRealm(openRealm);
-            };
-            if (shouldInitRealm) {
-                initRealm().catch(console.error);
-            }
-            return () => {
-                if (realm) {
-                    realm.close();
-                    setRealm(null);
-                }
-            };
-        }, [configVersion, realm, setRealm]);
         if (!realm) {
             if (typeof Fallback === "function") {
                 return React.createElement(Fallback, null);
```
**NOTE:** This works for me because I'm not changing the Realm config on the fly. Also, I think it can lead to memory leaks.

### Solution

One solution can be to let the instances independent, but with a prop in the `RealmProvider` that prevents the realm from being closed.

### Alternatives

Another solution can be to let us pass to the `createRealmContext` the realm instance so that we can have a unique instance of the realm in the entire app, for headless tasks or just code outside of the React component tree and for the hooks provided.

Would be a major improvement

Local Database only

takameyer:
## What, How & Why?
Add flag to keep Realm open on RealmProvider unmount
* There are a few valid use cases for doing this as calling realm.close will close all open instances of said realm.

This closes #6023

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -11,14 +11,10 @@
 * None
 
 ### Compatibility
-* React Native >= v0.71.4
-* Realm Studio v14.0.0.
-* File format: generates Realms with format v23 (reads and upgrades file format v5 or later for non-synced Realm, upgrades file format v10 or later for synced Realms).
+* Realm >= 11.0.0
 
 ### Internal
-<!-- * Either mention core version or upgrade -->
-<!-- * Using Realm Core vX.Y.Z -->
-<!-- * Upgraded Realm Core from vX.Y.Z to vA.B.C -->
+* Added more documentation to provider params.
 
 ## 0.5.2 (2023-08-09)
 
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None
 
 ### Enhancements
-* Add flag to keep realm open on unmount of RealmProvider. ([#6023](https://github.com/realm/realm-js/issues/6023))
+* Add flag to keep realm open on unmount of `RealmProvider`. ([#6023](https://github.com/realm/realm-js/issues/6023))
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None
 
 ### Enhancements
-* None
+* Add flag to keep realm open on unmount of RealmProvider. ([#6023](https://github.com/realm/realm-js/issues/6023))
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/packages/realm-react/README.md
+++ b/packages/realm-react/README.md
@@ -234,6 +234,19 @@ const AppWrapper = () => {
   )
 }
 ```
+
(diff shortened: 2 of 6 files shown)


kneth (approved):

packages/realm-react/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None
 
 ### Enhancements
-* None
+* Add flag to keep realm open on unmount of RealmProvider. ([#6023](https://github.com/realm/realm-js/issues/6023))
```
elle-j, line 7:
```suggestion
* Add flag to keep realm open on unmount of `RealmProvider`. ([#6023](https://github.com/realm/realm-js/issues/6023))
```

packages/realm-react/README.md
```diff
@@ -234,6 +234,19 @@ const AppWrapper = () => {
   )
 }
 ```
+
+It may also be necessary to render multiple `RealmProvider`s of the same Realm in an app.  In this case, the flag `closeOnUnmount` can be set to `false`` to prevent both Realm instances from closing when one has been removed from the component tree.
```
elle-j, line 238: An extra space got in there.

```suggestion
It may also be necessary to render multiple `RealmProvider`s of the same Realm in an app. In this case, the flag `closeOnUnmount` can be set to `false`` to prevent both Realm instances from closing when one has been removed from the component tree.
```

packages/realm-react/src/__tests__/RealmProvider.test.tsx
```diff
@@ -68,6 +68,19 @@ describe("RealmProvider", () => {
...
   });
 
+  it("returns the configured realm with useRealm and stays open if flagged", async () => {
+    const wrapper = ({ children }: { children: React.ReactNode }) => (
+      <RealmProvider closeOnUnmount={false}>{children}</RealmProvider>
+    );
+    const { result, unmount } = renderHook(() => useRealm(), { wrapper });
+    await waitFor(() => expect(result.current).not.toBe(null));
+    const realm = result.current;
+    expect(realm).not.toBe(null);
```
elle-j, line 78: Is this line needed, given line 76?

elle-j (approved):
Great enhancement 🙂

packages/realm-react/src/__tests__/RealmProvider.test.tsx
takameyer: Not really, no.

PR #6108, opened 2023-08-29
Fix `Object.linkingObjects()` across different classes

elle-j:
## What, How & Why?

When calling `Object.linkingObjects()` on an object linking to an object of a different object type, the following error would be generated:

```
<linking object type>#<linking property> is not a relationship to '<object type>'

// Example
Manufacturer#cars is not a relationship to 'Car'
```

Our assertion in the `Object.linkingObjects()` API has been corrected as well as the `collectionHelpers` passed when instantiating the returned `Results`. (It previously used the `collectionHelpers` from the `targetProperty` rather than target itself.)

**Note:**
* Previously existing tests were using linking objects of the same object type, thus the bug was not detected.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 * Fixed toolchain on Linux. On older Linux installations, the error `GLIBC_2.34' not found (required by /home/user/MyProject/node_modules/realm/generated/ts/realm.node)` could be observed. ([#6082](https://github.com/realm/realm-js/issues/6082), since v12.0.0)
-* Fixed accessing `Realm.Object.linkingObjects()` when the origin and target are of different object types. ([#6108](https://github.com/realm/realm-js/pull/6108))
+* Fixed accessing `Realm.Object.linkingObjects()` when the origin and target are of different object types. ([#6108](https://github.com/realm/realm-js/pull/6108), since v12.0.0)
 
 ### Compatibility
 * React Native >= v0.71.4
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Fixed toolchain on Linux. On older Linux installations, the error `GLIBC_2.34' not found (required by /home/user/MyProject/node_modules/realm/generated/ts/realm.node)` could be observed. ([#6082](https://github.com/realm/realm-js/issues/6082), since v12.0.0)
+* Fixed accessing `Realm.Object.linkingObjects()` when the origin and target are of different object types. ([#6108](https://github.com/realm/realm-js/pull/6108))
```
takameyer, line 11:
```suggestion
* Fixed accessing `Realm.Object.linkingObjects()` when the origin and target are of different object types. ([#6108](https://github.com/realm/realm-js/pull/6108), since v12.0.0)
```

packages/realm/src/Object.ts
```diff
@@ -427,20 +427,36 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
...
+        return value as binding.MixedArg;
+      },
+      fromBinding(value: unknown) {
+        assert.instanceOf(value, binding.Obj);
+        return wrapObject(value);
+      },
+      // See `[binding.PropertyType.Array]` in `PropertyHelpers.ts`.
+      get(results: binding.Results, index: number) {
+        return results.getObj(index);
+      }
```
takameyer, line 452:
```suggestion
      },
```

packages/realm/src/Object.ts
```diff
@@ -427,20 +427,36 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
...
+    }
```
takameyer, line 453:
```suggestion
    };
```

takameyer (approved):
LGTM! Great fix!

kneth (approved):

PR #6115, opened 2023-09-01
Make useObject compatible with undefined objects

### LINKED ISSUE #6101: The useObject doesn't re-render when the value transitions from undefined to an object.

Always

### Description
Hi Team,

I'm using Redux to cache an array of IDs. However, when the Realm database is cleared, the re-rendering doesn't occur when the IDs are updated. I have conducted debugging, and I've observed that...

The originalObjectRef.current in the useObject sometimes returns undefined, not null. However, within the function, it's checked using a null value.

In line **81** always return undefined
But in line **85** it's **"originalObjectRef.current === null"**, when I change to **"!originalObjectRef.current"** => it's work fine

image

```
diff --git a/node_modules/@realm/react/src/useObject.tsx b/node_modules/@realm/react/src/useObject.tsx
index 3d438b1..6d5d189 100644
--- a/node_modules/@realm/react/src/useObject.tsx
+++ b/node_modules/@realm/react/src/useObject.tsx
@@ -81,7 +81,7 @@ export function createUseObject(useRealm: () => Realm) {
         // Re-instantiate the cachedObject if the primaryKey has changed or the originalObject has gone from null to not null
         if (
           !arePrimaryKeysIdentical(primaryKey, primaryKeyRef.current) ||
-          (originalObjectRef.current === null && originalObject !== null)
+          (!originalObjectRef.current && originalObject !== null)
         ) {
           cachedObjectRef.current = createCachedObject({
             object: originalObject ?? null,

```

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

"realm": "11.0.0-rc.0",
"@realm/react": "^0.5.2",
"react-native": "0.68.5",

### Version

11.0.0-rc.0

Local Database only

### Are you using encryption?

Yes

0.68.5

### Build environment

### Cocoapods version

_No response_

takameyer:
## What, How & Why?
* If somehow an object returns undefined, re-renders would be skipped
* Added tests for proving that re-renders occur when the database is cleared
* Removed 11.0.0-rc from peer-dependency

This closes #6101

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


kneth (approved):

packages/realm-react/src/__tests__/useObjectRender.test.tsx
```diff
@@ -438,6 +438,24 @@ describe("useObject: rendering objects with a Realm.List property", () => {
...
+      expect(objectChangeCounter).toHaveBeenCalledTimes(initialCount);
+
+      await act(async () => {
+        testRealm.write(() => {
+          testRealm.deleteAll();
+        });
+        forceSynchronousNotifications(testRealm);
+      });
+
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
elle-j, line 455: Out of curiosity, what do we need the timeout for here? Is it to ensure the listeners have been called?

packages/realm-react/src/useObject.tsx
```diff
@@ -138,7 +138,7 @@ export function createUseObject(useRealm: () => Realm) {
     }, [realm, type, forceRerender]);
 
     // If the object has been deleted or doesn't exist for the given primary key, just return null
-    if (object === null || object?.isValid() === false) {
+    if (!object || object?.isValid() === false) {
```
elle-j, line 141: Could also be simplified to this 👇 
```suggestion
    if (!object?.isValid()) {
```

elle-j (approved):
LGTM 🙂

packages/realm-react/src/__tests__/useObjectRender.test.tsx
takameyer: Oh, I can probably remove that.  I was having problems getting the tests to pass and I thought it was a timing issue.  In the end I was looking at the wrong mock function to base my counts off of.

takameyer, line 455:
```suggestion
```

PR #6132, opened 2023-09-14
Values of properties are replaced by default value when updating

### LINKED ISSUE #6129: Values are replaces by default value when updating

Always

### Description

When you update a record and the schema includes default values, any properties that are not explicitly provided in the update will be replaced with their default values.

[Note: This only happens after upgrading to ```v12```](https://github.com/realm/realm-js/releases/tag/v12.0.0)

### Schema

```js
{
        name: 'MySchema',
        primaryKey: 'id',
        properties: {
            id: { type: 'int' },
            fieldOne: { type: 'string' },
            fieldTwo:  { type: 'string', default: 'DEFAULT_VALUE' },
        },
}
```

### Creating the record
```js
realm.create("MySchema", {id: 1337, fieldOne: "SOME_VALUE", fieldTwo: "NOT_DEFAULT_VALUE" })
```

### Updating the record
```js
realm.create("MySchema", {id: 1337, fieldOne: "SOME_OTHER_VALUE" },  Realm.UpdateMode.All)
```

This will cause the value of the ```fieldTwo``` field to be updated to ```DEFAULT_VALUE```, even though we didn't explicitly specify this property when performing the update.

from docs:

```diff
declare enum UpdateMode {
    /**
     * Objects are only created. If an existing object exists, an exception is thrown.
     */
    Never = "never",
    /**
     * If an existing object exists, only properties where the value has actually
     * changed will be updated. This improves notifications and server side
     * performance but also have implications for how changes across devices are
     * merged. For most use cases, the behavior will match the intuitive behavior
     * of how changes should be merged, but if updating an entire object is
     * considered an atomic operation, this mode should not be used.
     */
    Modified = "modified",
    /**
+     * If an existing object is found, all properties provided will be updated,
+     * any other properties will remain unchanged.     <==========
     */
    All = "all"
}
```

### Expected behaviour
Default values should be used only when the property is missing, and this should happen exclusively when you're creating a new record that doesn't already exist.

### Version

12.1.0

Local Database only

### Are you using encryption?

Yes

iOS 16.4

kneth:
## What, How & Why?

This closes #6129 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@
 * Added `Realm.Sync.Session.reconnect()` to help force a reconnection to Atlas Device Sync. ([#6123](https://github.com/realm/realm-js/issues/6123))
 
 ### Fixed
+* Fixed values of properties being replaced by default value when updating. ([#6129](https://github.com/realm/realm-js/issues/6129), since v12.0.0)
 * Fixed that value for `Realm.schemaVersion` wasn't propagated correctly for non-existing files. ([#6119](https://github.com/realm/realm-js/issues/6119), since v12.0.0)
-* Values of properties are replaced by default value when updating. ([#6129](https://github.com/realm/realm-js/issues/6129), since v12.0.0)
 
 ### Compatibility
 * React Native >= v0.71.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Fixed that value for `Realm.schemaVersion` wasn't propagated correctly for non-existing files. ([#6119](https://github.com/realm/realm-js/issues/6119), since v12.0.0)
+* Values of properties are replaced by default value when updating. ([#6129](https://github.com/realm/realm-js/issues/6129), since v12.0.0)
 
 ### Compatibility
 * React Native >= v0.71.4
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -1335,4 +1335,38 @@ describe("Realm.Object", () => {
       expect(objFromKey?.age).equals(7);
     });
   });
+
+  // from https://github.com/realm/realm-js/issues/6129
+  describe("Primary key and property with default value", () => {
+    const PrimaryAndDefaultSchema: Realm.ObjectSchema = {
+      name: "MySchema",
+      primaryKey: "id",
+      properties: {
+        id: { type: "int" },
+        fieldOne: { type: "string" },
+        fieldTwo: { type: "string", default: "DEFAULT_VALUE" },
+      },
+    };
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * None
 
 ### Fixed
+* Values of properties are replaced by default value when updating. ([#6129](https://github.com/realm/realm-js/issues/6129), since v12.0.0)
```
elle-j, line 10:
```suggestion
* Fixed values of properties being replaced by default value when updating. ([#6129](https://github.com/realm/realm-js/issues/6129), since v12.0.0)
```

elle-j (approved):
Great fix!

integration-tests/tests/src/tests/objects.ts
```diff
@@ -1335,4 +1335,38 @@ describe("Realm.Object", () => {
...
+      primaryKey: "id",
+      properties: {
+        id: { type: "int" },
+        fieldOne: { type: "string" },
+        fieldTwo: { type: "string", default: "DEFAULT_VALUE" },
+      },
+    };
+    openRealmBeforeEach({ schema: [PrimaryAndDefaultSchema] });
+
+    [Realm.UpdateMode.All, Realm.UpdateMode.Modified].forEach((updateMode) => {
```
kraenhansen, line 1352: I'm not a fan of the `forEach` being used like this, as it's undefined behaviour if the function argument is called synchronously or not by the runtime implementation and this code depends on it being called synchronously.

```suggestion
    for (const updateMode of [Realm.UpdateMode.All, Realm.UpdateMode.Modified]) {
```

integration-tests/tests/src/tests/objects.ts
```diff
@@ -1335,4 +1335,38 @@ describe("Realm.Object", () => {
...
+      it(`Update properties (updateMode = ${updateMode})`, async function (this: Mocha.Context & RealmContext) {
```
kraenhansen, line 1353: Minor suggestion to make the output read better.

```suggestion
      it(`updates properties (updateMode = ${updateMode})`, async function (this: Mocha.Context & RealmContext) {
```

kraenhansen (approved):

packages/realm/src/Object.ts
```diff
@@ -232,7 +232,7 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
           result[propertyName] = propertyValue;
         }
       } else {
-        if (typeof defaultValue !== "undefined") {
+        if (created && typeof defaultValue !== "undefined") {
```
kraenhansen, line 235: Since there's also a `created` guard on line 240, I wonder if it would be better to turn the entire `else` block (line 234) into an `else if (created)` block, instead.

PR #6136, opened 2023-09-18
Add RN example app (Multiple Realms)

### LINKED ISSUE #5821: Example in RN: Split data to multiple Realm

elle-j:
## What, How & Why?

Adds a Netflix-like React Native example app along with the backend App Services App for users who prefer to set it up via the Realm CLI.

For more information and an overview of the project structure, see the [README](https://github.com/realm/realm-js/blob/lj/example-rn-multiple-realms/examples/rn-multiple-realms/README.md).

### Main Files to Review
* `frontend/app/`
* `README.md`

This closes #5821

## ☑️ ToDos
--


examples/rn-multiple-realms/frontend/app/screens/MoviesScreen.tsx
```diff
@@ -0,0 +1,75 @@
...
+        keyExtractor={section => section.category}
+        renderItem={({item: section}) => (
+          <MovieList
+            category={section.category}
+            movies={section.movies}
+            onItemPress={showMovieInfo}
+          />
+        )}
+        showsVerticalScrollIndicator={false}
+      />
```
elle-j, line 66: Here we're rendering nested `FlatList`s. The outer one is the vertical list with each category, and the inner one (rendered by `MovieList`) contains the horizontal list of movies for 1 category.

This could ideally by swapped for 1 list (using [`SectionList`](https://reactnative.dev/docs/sectionlist)) for performance, but `SectionList` requires `Symbol.unscopables` to be available on our `Results` (at least for TypeScript), which we have yet to implement.

kneth (approved):
I enjoyed reading the top-level `README.md`, and it should serve as an example for all our examples.

elle-j:
Thanks @kneth 🙂

examples/rn-multiple-realms/frontend/app/providers/MovieProvider.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  // The selected movie will decide which movie is shown on the
+  // `MovieInfoScreen` when a movie item is pressed from the list.
+  const [selectedMovie, setSelectedMovie] = useState<Movie | null>(null);
+
+  // Query movies of certain genres. (A helper `query()` function is
+  // defined to simplify the argument passed to `useQuery()`).
+  const query = (movies: Realm.Results<Movie>, genre: string) =>
+    movies.filtered(`'${genre}' IN genres`);
+  const action = useQuery(Movie, movies => query(movies, 'Action'));
```
takameyer, line 87: I'm wondering with this setup if when one of these `useQuery` hooks triggers a re-render, if the entire application will re-render, even if the data isn't currently in use.

examples/rn-multiple-realms/README.md
```diff
@@ -0,0 +1,268 @@
...
+You can either choose to set up your App via a CLI (this has fewer steps and is much faster since all configurations are already provided in the [backend directory](./backend/)), or via the App Services UI (steps provided below).
+
+#### Via a CLI (recommended)
+
+To import and deploy changes from your local directory to App Services you can use the command line interface:
+
+1. [Set up Realm CLI](https://www.mongodb.com/docs/atlas/app-services/cli/).
+2. In the provided [backend directory](./backend/) (the App Services App), update the following:
+    * Cluster Name
+      * Update the [`"clusterName"`](./backend/data_sources/mongodb-atlas/config.json) to the name of your cluster.
```
takameyer, line 132: I think the file path should also be shown in clear text.  I didn't realize this was a link when seeing it in plain markdown and wasn't sure which file to edit.

takameyer: image.png

examples/rn-multiple-realms/frontend/app/screens/AppInfoScreen.tsx
```diff
@@ -0,0 +1,86 @@
...
+        This example app showcases how to use different Realms in MongoDB's
+        Realm React Native SDK. All users can browse (not play) movies from
+        MongoDB's Mflix sample dataset, but only users who register with email
+        and password are able to sync, read, add, and remove movies saved to "My
+        List".
+      </Text>
+      <Text style={styles.paragraph}>Learn more at:</Text>
+      <Pressable
+        onPress={() =>
+          Linking.openURL(
```
takameyer, line 45: We should make use of [this trick](https://github.com/realm/realm-js/blob/main/example/app/components/IntroText.tsx#L24) to make these links open in the users browser instead of the simulator browser

examples/rn-multiple-realms/frontend/app/screens/MoviesScreen.tsx
takameyer: We really need to fix the `unscopables` issue.  I think the nested FlatList is fine though.  I used similar strategies before.

takameyer (approved):
Very great looking application!  This is such a cool example.  Really great work here :)

examples/rn-multiple-realms/frontend/app/providers/MovieProvider.tsx
elle-j: Yeah in this case, the children of this provider will rerender. Once the movies have been synced, the only data being updated however will be the `My List`. I tried a bunch of variants in this file to try to notice speed differences. Using e.g. multiple `useQuery()`s with individual filters (one for each genre) were better than using one `useQuery()` and calling `.filtered()` for each genre.

But more to your point, since the `MoviesNavigator` is the only component using the context value, one thing I can try to do is to move the `MovieProvider` to only wrap that, rather than wrapping `RootNavigator` 👍

examples/rn-multiple-realms/README.md
elle-j: Ah okay thanks! Will update that.

examples/rn-multiple-realms/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+1. [Deploy a free Atlas cluster](https://www.mongodb.com/docs/atlas/getting-started/#get-started-with-atlas) and create an Atlas database.
+2. [Load the Sample Mflix Dataset](https://www.mongodb.com/docs/atlas/sample-data/) into your Atlas database.
+    * Several databases and collections exist in the sample dataset, but we will only be using the `sample_mflix` database and its `movies` collection.
+
+### Set up an Atlas App Services App
+
+You can either choose to set up your App via a CLI (this has fewer steps and is much faster since all configurations are already provided in the [backend directory](./backend/)), or via the App Services UI (steps provided below).
+
+#### Via a CLI (recommended)
```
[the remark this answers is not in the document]
elle-j: Yeah I definitely think so too. Was talking to Kenneth about it today. Much better DX, less hassle, and less error prone 👍

examples/rn-multiple-realms/frontend/app/screens/AppInfoScreen.tsx
elle-j: 🙂 👌

examples/rn-multiple-realms/frontend/app/screens/MoviesScreen.tsx
elle-j: I think I've seen before that nested FlatLists may cause all list items (in the inner list) to render, so not only when they appear on the screen. In our case, this makes `useMemo()` even more important.

PR #6165, opened 2023-09-29
Override fetch for Altas App Services

### LINKED ISSUE #6159: Starting from version 12 it is not possible to override the fetch API used by SDK

Always

### Description

Previously Node.js SDK was using `@realm/network-transport` package as a provider of fetch API and it was possible to override it and use e.g. `window.fetch` API when running in an Electron renderer process. But starting from version 12.0 the SDK seems to use `node-fetch` exclusively without an option to override it.
The option to use custom fetch API was important because it allowed to control SSL validation for network operations performed by SDK, for example usage of `window.fetch` in an Electron renderer allows to trust certificates added to the system certificate store.

I see there is an option to define a custom function for SSL validation in [Sync configuration](https://www.mongodb.com/docs/realm-sdks/js/latest/types/SSLConfiguration.html) which is a good feature that could add to flexibility, however there seems to be no way to control SSL validation for requests sent by SDK to log user in. This defeats the purpose of SSL settings in the sync configuration because the SSL validation issues it should help to resolve will prevent app from getting a user session needed to sync data down.

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

1. Run Realm in Electron, in a renderer process with node integration.
2. Try to inject `window.fetch` API into realm SDK in order to trust custom certificates from the system store.

### Version

12.1.0

### Are you using encryption?

No

Windows 10, Debian 11

### Build environment

_No response_

### Cocoapods version

_No response_

kneth:
## What, How & Why?

This closes #6159 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -32,6 +32,7 @@ export type AppConfigurationRelaxed = {
   timeout?: number;
   multiplexSessions?: boolean;
   baseFilePath?: string;
+  fetchOverride?: any;
```
kraenhansen, line 35: My immediate reaction is that the "override" part of this name is redundant as all of these options override default values.
```suggestion
  fetch?: any;
```

integration-tests/tests/src/node/custom-fetch.ts
```diff
@@ -0,0 +1,66 @@
...
+describe("custom fetch", () => {
+  let called = false;
+
+  const HTTP_METHOD: Record<number, string> = {
+    [0]: "GET",
+    [1]: "POST",
+    [2]: "PUT",
+    [3]: "PATCH",
+    [4]: "DELETE",
+  };
```
kraenhansen, line 34: I wonder if it'll make more sense to export this somehow.

integration-tests/tests/src/node/custom-fetch.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const customFetch = async (request: Request) => {
+    called = true;
+    const { url, body, ...rest } = request;
+    if (typeof request.method == "number") {
+      rest.method = HTTP_METHOD[request.method];
+    }
+    const response = await fetch(url, rest);
+    return response;
+  };
```
kraenhansen, line 44: An alternative could be to implement this as a wrapping of the default `fetch` from `@realm/network-transport`.

takameyer: As the user is requesting overriding the `AbortController`, it's probably worth while being full flexibly

packages/realm/src/app-services/App.ts
```diff
@@ -140,6 +141,13 @@ export type AppConfiguration = {
    * @since 12.2.0
    */
   metadata?: Metadata;
+
+  /**
+   * Specify a custom `fetch` implementation.
+   * @since 12.3.0
+   * @experimental This API is experimental and may change or be removed.
+   */
+  fetchOverride?: unknown;
```
takameyer, line 150: This type should match the `NetworkType` from `network.ts`
```
  fetch(request: Request | binding.Request ): Promise<FetchResponse>;
```

elle-j: I think it should be similar to a [specification](https://developer.mozilla.org/en-US/docs/Web/API/fetch). Also, I'd say we should probably try to not expose bindgen details to the public API.

packages/realm/src/app-services/App.ts
```diff
@@ -277,6 +286,10 @@ export class App<
       }
     }
 
+    if (fetchOverride) {
+      inject({ fetch: fetchOverride } as NetworkType);
+    }
```
elle-j, line 291: If we decide to add a more explicit type than `any` or `unknown` to the `fetchOverride` in the app config (I think at least specifying that it should be a function could be good), then what do you think of adding an explicit check for `undefined` here? This way, we can let users know if they have provided an invalid falsy value *in addition to* an invalid truthy value. Also, we can assert that it's a function.
```suggestion
    if (fetchOverride !== undefined) {
      assert.function(fetchOverride, "fetchOverride");
      inject({ fetch: fetchOverride } as NetworkType);
    }
```

packages/realm/src/platform/network.ts
```diff
@@ -60,5 +60,5 @@ export const network: NetworkType = {
 };
 
 export function inject(injected: NetworkType) {
-  Object.freeze(Object.assign(network, injected));
+  Object.assign(network, injected);
```
elle-j, line 63: Did using `.freeze()` cause an issue here?

integration-tests/tests/src/node/custom-fetch.ts
takameyer: I think this solution is good, otherwise we would also have to provide a mechanism to override the default `AbortController` as well.  This basically lets the user override `@realm/network-transport`.
I guess the crux of the issue, is the user that reported the original issue is using `electron` and this is using the node bindings instead of the browser bindings to `fetch`.

packages/realm/src/platform/network.ts
takameyer: Unless this `inject` called multiple times, the `freeze` shouldn't be an issue.

packages/realm/src/app-services/App.ts
takameyer: Will there be unintended side effects if the `fetch` function doesn't match our specifications?

elle-j: Are you referring to our `NetworkType`? Since our TS expects the `fetch` used (whether it'd be ours or a user-provided one) to be any of the functions defined in the `NetworkType`, the preferable option (from the way it's set up atm) seems to be `fetch(request: Request): Promise<FetchResponse>;`. In that case we'd need to export `Request` and `FetchResponse`.

But, safest would probably be to ask the user of their intended use (arg and return type) and if `fetch(request: Request): Promise<FetchResponse>;` would be sufficient for them.

takameyer: The user in this case just wants to replace `node-fetch` with `window.fetch`, since they are running `realm` in a browser environment in `electron`.  

My main concern is [this line](https://github.com/realm/realm-js/blob/main/packages/realm/src/platform/network.ts#L55).  Here there is a transformation if the request happening depending on it's contents.  If a user is providing a custom `fetch`, will not doing this transformation cause an issue?

From this observation, perhaps @kraenhansen is correct in that we should directly override the `fetch` method within `@realm/network-transport` (a level deeper), and leave this one alone.  Then we would also need to provide mechanisms to override the `AbortController`.

It's probably worth while doing a manual test in React Native, to see if such an override could cause significant issues.

elle-j: The way I understood it was that the entire [fetch](https://github.com/realm/realm-js/blob/83783da13ac4616e8955a3c35f7491986f2786ce/packages/realm/src/platform/network.ts#L53) would be replaced by the user-provided implementation. Thus the line you linked to is only for when our implementation is used, which is why the `typeof` check is necessary in that case.

takameyer: Our internal code will be potentially sending two types of requests objects to the fetch in `network.ts`.  When a user overrides this, the override may have to also react to two different types of requests.  It's possible it will just work, but I don't think we have properly tested for this.

packages/realm/src/platform/network.ts
kraenhansen: Yeah - my suggestion to avoid the freeze was only if we choose to expose the `inject` function to end-users.

elle-j: Gotcha 🙂

packages/realm/src/app-services/App.ts
takameyer: That could work, but thinking about that gave me another idea.  We could do something like this:

```
let baseFetch = transport.fetch;

export const overrideFetch(fetch) = {
  baseFetch = fetch;
}

export const network: NetworkType = {
  async fetch(request) {
    debug("Requesting %O", request);
    const fetchRequest = typeof request.method === "string" ? request : toFetchRequest(request);
    const response = await baseFetch();  // before was `transport.fetch()`
    debug("Responded %O", response);
    return response;
  },
};
```

Probably should do this in a way that freezes the fetch override, and maybe rename `baseFetch` to something else.

elle-j: I like it! 😎 I'll add it to our Round Table and let's discuss more :)
```ts
export function overrideFetch(fetch: (request: Request) => Promise<FetchResponse>) {
  baseFetch = fetch;
}
```

kraenhansen: My immediate feeling is that I like the declarative API of keeping this a configuration of the `App`. Possibly providing a way to override the default values for those config values, but then again - how many developers do we expect to have more than one `App` instance per application?

I like ☝️ over this more procedural `overrideFetch` function suggestion as that minimize the state in the library, which also has the added benefit of allowing two `App` instances to fetch using different implementations concurrently.

takameyer: The intention of the `overrideFetch` suggestion keeps the configuration through `App`, not for the developer to call this directly.  The point being, to keep the internal logic of `network:fetch` (calls to `toFetchRequest`), and give the opportunity to override the `fetch` within the `network:fetch` method.
I was not proposing to remove the configuration by `App`, but to clear up that we would be overriding the internal `network:fetch` logic with the current PR state.

kraenhansen: @takameyer I don't see how that'll work. The `App` would have to call `overrideFetch`, passing in the `fetch` before every single request to the server? It'll be a shared state between two `App` instances, that would have to be constantly switched between the two.

takameyer: Then we should write this in a way that works in the case you are suggesting.  The code I wrote above was only an example, but the intention should be clear.  

In any case, I think we are misunderstanding each other and would enjoy discussing this further when you return.

PR #6169, opened 2023-10-02
Delete deploy-cluster workflow

kraenhansen:
## What, How & Why?

This workflow keeps a cluster on QA alive for v11 to test against - which is no longer needed are we're focussing work on v12 going forward. We could migrate v11 to test against the local BaaS + Ngrok as v12, but choose not to (see https://github.com/realm/realm-js/pull/5786#issuecomment-1696158441)

💥 NOTE: Once this is merged, v11 won't be able to execute integration tests


kneth (approved):

kraenhansen:
I thought a bit more about this as figured it would be better to disable the scheduled deployment than deleting the workflow completely: https://github.com/realm/realm-js/pull/6170

PR #6172, opened 2023-10-03
Refactor preexisting example app with focus on offline support

### LINKED ISSUE #5825: Example in RN: Offline login/ realm access

elle-j:
## What, How & Why?

Moves our existing React Native "Todo/Task Manager" example app from `example/` to `examples/rn-todo-list` and refactors several parts of the app. *(Only some files are relevant for review. See further below.)*

**The following has been updated/refactored:**
* It focuses on showcasing offline support and how to CRUD data.
* Created more of a separation between business and view concerns.
  * This is structurally similar to our React and Electron Todo apps.
* Updated the styles.
  * Since this example uses the same data model as our React/Wasm and Electron Todo examples, they can be demoed on these different platforms at the same time, using the same underlying App Services App.
  * To have them look and feel more cohesive, I've restyled this app to look similar.
* As earlier, it provides two apps in one (one local-only and one Device Sync enabled).
  * Which one to use is configured by the developer.
  * The previous default in this config was local-only. This PR sets Device Sync as the default.
  * (This can be reverted if needed.)
* The app has been removed from workspaces.
  * This also means it is not using our `@realm/metro-config` package.
  * If we find it more useful for us to have it be part of workspaces, I propose that we make those changes in a follow-up PR.
* A README is provided along with the backend App Services App for users who prefer to set it up via the Realm CLI.

> For more information and an overview of the project structure, see the [README](https://github.com/realm/realm-js/blob/lj/refactor-example/examples/rn-todo-list/README.md).

**Not updated:**
* Our templates remain untouched.
  * As I understand it, the templates were very similar to this example app.
  * If needed, the templates can also get an upgrade at a later point :)

### Main Files to Review
* `examples/rn-todo-list`
  * `frontend/app/`
  * `README.md`

This closes #5825

## ☑️ ToDos
--


examples/rn-todo-list/frontend/app/models/Task.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export class Task extends Realm.Object {
+  _id: BSON.ObjectId = new BSON.ObjectId();
+  description!: string;
+  isComplete: boolean = false;
+  createdAt: Date = new Date();
+  userId!: string;
+
+  static primaryKey = '_id';
+}
```
elle-j, line 43: **Discussion:**
Our other examples are not using the Babel plugin. Should we update the above data model to not use the plugin (but instead mention it in the TS docs)?

examples/rn-todo-list/frontend/sync.config.js
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+export const SYNC_CONFIG = {
+  // Set `enabled` to `true` to enable sync.
+  enabled: true,
+  // Add your Atlas App ID here if sync is enabled.
+  appId: 'YOUR_APP_ID',
+};
```
elle-j, line 24: Since this app uses two versions (a local-only and a sync one), I did not refactor this part into the convention used by the other examples (i.e. having an `ATLAS_APP_ID` variable exported in `app/atlas-app-services/config.ts`. Feel free to say if you'd rather see this following the same convention as the other apps.

Note:
Previous default for `enabled` was `false`.

examples/rn-todo-list/README.md
```diff
@@ -0,0 +1,208 @@
...
+This app opens a Realm via `RealmProvider` (see [AppSync.tsx](./frontend/app/AppSync.tsx)) and passes the configuration as props. We use `OpenImmediately` for new and existing Realm files in order to use the app while offline.
+
+> See [OpenRealmBehaviorConfiguration](https://www.mongodb.com/docs/realm-sdks/js/latest/types/OpenRealmBehaviorConfiguration.html) for possible configurations of new and existing Realm file behaviors.
+
+## Getting Started
+
+### Prerequisites
+
+* [Node.js](https://nodejs.org/)
+* [React Native development environment](https://reactnative.dev/docs/environment-setup?guide=native)
```
elle-j, line 91: For the React Native prereq here, should we also mention that it uses the React Native CLI Quickstart rather than the Expo one?

examples/rn-todo-list/frontend/__tests__/App.test.tsx
```diff
@@ -4,7 +4,10 @@
 
 import 'react-native';
 import React from 'react';
-import App from '../App';
+import {App} from '../index';
+
+// Note: import explicitly to use the types shipped with jest.
+import {it} from '@jest/globals';
```
kraenhansen, line 10: I wonder if this should be included in a `types` array in a tsconfig.json instead 🤔

elle-j: I'll look into it, this was the default when creating a new project (`npx react-native@latest init`).

examples/rn-todo-list/README.md
kneth: Yes, otherwise it might cause confusion.

examples/rn-todo-list/frontend/app/models/Task.ts
kneth: My suggestion is to leave it enabled. Our babel plugin is a package that we wish to support and maintain, and this example can be a source for bug reports from users if it doesn't work.

examples/rn-todo-list/frontend/sync.config.js
kneth: With `enabled` set to `true`, the app will fail early which can guide the user to try Device Sync - smart move :smile:

I don't see an issue that one example deviate from the other example so let's keep it as it is.

kneth (approved):
Thank you for migrating the old example so it can be included in our example collection.

examples/rn-todo-list/frontend/app/models/Task.ts
elle-j: Good point!

examples/rn-todo-list/frontend/__tests__/App.test.tsx
kraenhansen: No worries then 👍 It's better to keep aligned with the template.

takameyer (approved):
Looks good!  Still need to find a way to make this "workspace"able.  I do prefer not having to update all those xcode and android files.  Perhaps we will end up using a new package manager to solve this issue.

examples/rn-todo-list/frontend/app/AppSync.tsx
```diff
@@ -0,0 +1,89 @@
...
+              // If the user has logged in at least 1 time before, the Realm and its data will
+              // exist on disk and can be opened even when offline. We can either (a) open the
+              // Realm immediately (or first create a new empty Realm file if it does not
+              // exist before opening it) and sync the data to the device in the background
+              // (`OpenRealmBehaviorType.OpenImmediately`), or (b) wait for any non-synced
+              // data to be fully downloaded (`OpenRealmBehaviorType.DownloadBeforeOpen`).
+              // For more possible configurations of new and existing Realm file behaviors, see:
+              // https://www.mongodb.com/docs/realm-sdks/js/latest/types/OpenRealmBehaviorConfiguration.html
+              newRealmFileBehavior: {
+                type: OpenRealmBehaviorType.OpenImmediately,
```
takameyer, line 70: Ah, forgot to look over these settings.  I think I'd prefer `DownloadBeforeOpen` for the `newRealmFileBehavior`.  Otherwise it will start with an empty list and suddenly a populated list will appear (if there was something pre-existing).

elle-j: Yeah, since the first open would also have required the user to be online from logging in, that may be better 👍 Will test how that looks. Not easy at this point though to show some decent loading until we have the progress notifications project done right?

takameyer: Just the a simple spinner at this point

elle-j: `RealmProvider.fallback` to the rescue 🙏

PR #6175, opened 2023-10-05
Duplicate the README and media into the Realm Package

takameyer:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


elle-j (approved):
For some reason I couldn't apply comments or suggestions, but maybe we can take the opportunity to replace `https://realm.io/` with `https://www.mongodb.com/docs/realm/`.

PR #6176, opened 2023-10-05
New README.md strategy

takameyer:
* Move README.md into `packages/realm` and replace with a symbolic link.
* Update all relative paths in README.md to be absolute.
* Replace realm.io links

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


packages/realm-react/package.json
```diff
@@ -96,9 +96,9 @@
     "hooks"
   ],
   "author": {
-    "name": "Realm",
+    "name": "MongoDB",
```
elle-j, line 99: P.S. Some of the other package.jsons still have `"name": "Realm"`.

elle-j (approved):

packages/realm-react/package.json
takameyer: Ah, I wasn't sure if I was sold on that yet, but I guess MongoDB is the official author, so I can update that as well

PR #6185, opened 2023-10-07
Update Wasm example app

elle-j:
## What, How & Why?

Most importantly, this fixes a package-lock issue when running `npm install` by removing the example from our workspaces.

It also:
* Upgrades `@realm/react`.
* Adds and updates comments.
* Updates README.
* Minor update to UI to align with our corresponding Electron example.

**Notes:**
* This PR does not separate the example into frontend and backend directories in order to not break existing links and Doc instructions.

## ☑️ ToDos
--


takameyer (approved):
Looks good!

PR #6203, opened 2023-10-18
Update Podfile.lock after RN upgrades

elle-j:
## What, How & Why?

Updates the Podfile.lock to also use the updated RN version to prevent error when installing the pods. The error is incompatible version for the hermes engine pod.

## ☑️ ToDos
--


takameyer (approved):
Good catch!

Issue #6186, opened 2023-10-09
[@realm/react] Update the user reference from `useUser()` on each update


elle-j:
Currently, the `User` reference returned by `useUser()` will be updated only [when the ID is updated](https://github.com/realm/realm-js/blob/b95f5ec0586e2f554d8136d9cc6f2a1a5e2e54c4/packages/realm-react/src/UserProvider.tsx#L51-L56).

To more easily react to changes when e.g. the access token is refreshed or the user profile is updated, we could instead update the user on each invocation of the user listener.

This would make `useUser()` behavior similarly to `useQuery()` in terms of rerendering.

PR #6206, opened 2023-10-20
`useUser` fixes for Realm React

### LINKED ISSUE #6196: Mismatch type when passing `user`  from `useUser` as parameter

Always

### Description

As the [documentation](https://www.mongodb.com/docs/realm/sdk/react-native/manage-users/authenticate-users/) suggested, we can pass the `user` to a function. 

```typescript

// Gets a valid user access token to authenticate requests
async function getValidAccessToken(user: Realm.User) {
// An already logged in user's access token might be stale. To
// guarantee that the token is valid, refresh it if necessary.
await user.refreshCustomData();
 
return user.accessToken;
}

```

However, when passing the `user` from `useUser()`. following error display. 

```
Argument of type
User<DefaultFunctionsFactory, Record<string, unknown>, DefaultUserProfileData>
is not assignable to parameter of type
UserType<UserFunctionsType, UserCustomDataType, UserProfileDataType>
Types have separate declarations of a private property  cachedProfile 
```

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

`file1.ts`
```typescript
import Realm from 'realm';

export function foo(user: Realm.User){
  // Do something
}

```

`file2.ts`
```typescript
import {foo} from file1;

const user = useUser();
foo(user);
```

### Version

"realm": "^12.2.1",, "@realm/react": "^0.6.0",

Atlas App Services: Functions or GraphQL or DataAPI etc

### Are you using encryption?

No

Win11 23H2

### Build environment

### Cocoapods version

_No response_

takameyer:
## What, How & Why?
* Used a different type declaration generation method which solved typing issues
* Updated `UserProvider` to always update the user reference on user events
* Updated connection example to include a delete user function to test this feature

This closes #6196 and #6186

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/examples/rn-connection-and-error/frontend/app/components/Button.tsx
+++ b/examples/rn-connection-and-error/frontend/app/components/Button.tsx
@@ -64,6 +64,7 @@ const styles = StyleSheet.create({
     borderRadius: 30,
     borderColor: colors.purple,
     backgroundColor: colors.purple,
+    justifyContent: 'center',
   },
   buttonSecondary: {
     paddingVertical: 6,
--- a/packages/realm-react/package.json
+++ b/packages/realm-react/package.json
@@ -16,7 +16,7 @@
   },
   "wireit": {
     "bundle": {
-      "command": "rollup --config",
+      "command": "rollup --config && tsc -p tsconfig.types.json",
       "dependencies": [
         "../realm:bundle",
         "../realm-common:bundle"
@@ -71,7 +71,6 @@
     "react-native": "0.72.6",
     "react-test-renderer": "18.2.0",
     "realm": "*",
-    "rollup-plugin-dts": "^5.0.0",
     "ts-jest": "^29.0.5"
   },
   "peerDependencies": {
--- a/packages/realm-react/src/UserProvider.tsx
+++ b/packages/realm-react/src/UserProvider.tsx
@@ -49,9 +49,6 @@ export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback, 
 
   useEffect(() => {
     const event = () => {
-      console.log("UserProvider: user changed");
-      console.log("UserProvider: app.currentUser", app.currentUser);
-      console.log("UserProvider: app.currentUser.id", app.currentUser);
       setUser(app.currentUser);
     };
     user?.addListener(event);
--- a/packages/realm-react/src/UserProvider.tsx
+++ b/packages/realm-react/src/UserProvider.tsx
@@ -49,9 +49,10 @@ export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback, 
 
   useEffect(() => {
     const event = () => {
-      if (app.currentUser?.id != user?.id) {
-        setUser(app.currentUser);
-      }
+      console.log("UserProvider: user changed");
+      console.log("UserProvider: app.currentUser", app.currentUser);
+      console.log("UserProvider: app.currentUser.id", app.currentUser);
+      setUser(app.currentUser);
     };
     user?.addListener(event);
     app?.addListener(event);
--- a/packages/realm-react/tsconfig.types.json
+++ b/packages/realm-react/tsconfig.types.json
@@ -3,6 +3,8 @@
   "compilerOptions": {
     "declaration": true,
     "emitDeclarationOnly": true,
+    "declarationMap": true,
+    "incremental": true,
     "outDir": "./dist",
   },
   "exclude": [
--- a/packages/realm-react/tsconfig.types.json
+++ b/packages/realm-react/tsconfig.types.json
@@ -1,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "./dist",
+  },
+  "exclude": [
+    "**/__tests__/*"
+  ]
+}
--- a/examples/rn-connection-and-error/README.md
+++ b/examples/rn-connection-and-error/README.md
@@ -71,6 +71,7 @@ It specifically addresses the following points:
       * With invalid password
       * With non-existent email
 * Listening (and triggering) when a user gets logged out.
+* Listening (and triggering) when a user gets removed.
 * Listening (and triggering) when a user's tokens are refreshed.
 * Listening (and triggering) when the underlying sync session:
   * Tries to connect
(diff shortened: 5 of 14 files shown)


examples/rn-connection-and-error/frontend/app/components/Button.tsx
```diff
@@ -64,6 +64,7 @@ const styles = StyleSheet.create({
     borderRadius: 30,
     borderColor: colors.purple,
     backgroundColor: colors.purple,
+    justifyContent: 'center',
```
elle-j, line 67: I believe this may be unnecessary (no effect).

packages/realm-react/src/UserProvider.tsx
```diff
@@ -49,9 +49,10 @@ export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback,
 
   useEffect(() => {
     const event = () => {
-      if (app.currentUser?.id != user?.id) {
-        setUser(app.currentUser);
-      }
+      console.log("UserProvider: user changed");
+      console.log("UserProvider: app.currentUser", app.currentUser);
+      console.log("UserProvider: app.currentUser.id", app.currentUser);
```
elle-j, line 54:
```suggestion
```

packages/realm-react/src/UserProvider.tsx
```diff
@@ -49,9 +49,10 @@ export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback,
...
+      setUser(app.currentUser);
```
elle-j, line 55: 🙂

elle-j (commented):
Awesome to get this fix in! LGTM when [this PR](https://github.com/realm/realm-js/pull/6208) is merged into the current one 🙂

packages/realm-react/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "wireit": {
     "bundle": {
-      "command": "rollup --config",
+      "command": "rollup --config && tsc -p tsconfig.types.json",
```
kraenhansen, line 19: Should this be a separate `types` command? Since technically this isn't bundling the types now.

packages/realm-react/tsconfig.types.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "./dist",
```
kraenhansen, line 6: Consider adding https://www.typescriptlang.org/tsconfig#incremental and https://www.typescriptlang.org/tsconfig#declarationMap too.

examples/rn-connection-and-error/frontend/app/components/Button.tsx
takameyer: It was intentional.  This centers vertically within the button container.

packages/realm-react/package.json
takameyer: As our ci assumes a single command for preparing packages for release, I would be hesitant to put this in a separate command.
Otherwise, we can make a follow up PR which adds a `types` command to every package (or simply `build`, which contains `bundle` and `types`) and update our workflows to call this as well.

examples/rn-connection-and-error/frontend/app/components/Button.tsx
takameyer: Specifically "Refresh Access Token" was on two lines and this caused the height of the button on this row to change.

packages/realm-react/src/UserProvider.tsx
takameyer: Somehow I lost the change I actually wanted to commit.  This line is now
```
setUser( app.currentUser ? new Proxy(app.currentUser, {}) : null );
```

examples/rn-connection-and-error/frontend/app/components/Button.tsx
takameyer: image

takameyer: image
without change

elle-j: Ok great 🙂 Was using a wider screen 😛

packages/realm-react/src/UserProvider.tsx
elle-j: Is it worth adding a test for this btw?

takameyer: It's worth it, but I don't have time, but I'm not actually scheduled to work on this right now.  I just want things to work.  
The UserProvider needs tests, at the time of creation it wasn't clear how to deploy an app for functional testing, but now we can do this.

kneth (approved):

packages/realm-react/src/UserProvider.tsx
takameyer: https://github.com/realm/realm-js/issues/6212. For the next phase of `realm react` we have planned

elle-j: Great!

packages/realm-react/package.json
takameyer: I put this in our next roundtable.  I think you bring up a good point.

My suggestion is to create a script in each of our mono-repo packages called `prepack`, which can include whatever scripts it needs to prepare for being consumed by another project.  This could include `bundle`, `generateTypes`, `build`, etc.
Then the CI workflows would be updated to run this command instead of `bundle`.

kraenhansen: It was the idea that these inter-package task dependencies should be configured via wireit. It's much more capable as it'll cache outputs and potentially execute tasks in parallel.

packages/realm-react/src/UserProvider.tsx
elle-j: FYI, looks like the `new Proxy()` implementation was never committed.

takameyer: 🤯

takameyer: Must be a rebase issue.  Guess that's why we should write tests :D

packages/realm-react/package.json
takameyer: I don't disagree, I just want to unify on a script that specifically deals with preparing for release.

PR #6207, opened 2023-10-20
Remove `postinstall` script

elle-j:
## What, How & Why?

Our `bindgen:generate:typescript` and `bindgen:generate:wrappers` scripts rely on the `realm-bindgen` command. For Bindgen's  node dependencies in Core to be installed correctly and allow for the `realm-bindgen` command to be run, `npm install` needs to be run after `git submodule update --init --recursive` (which was previously in our `postinstall` script).

Replacing the `postinstall` with a `preinstall` does not fix this issue.

This PR removes the `postinstall` script all together, moving the responsibility of fetching the submodule to the developer:

```sh
npm install

# Now you can e.g. run tests (which will build etc. automatically).
npm run test --workspace realm # Unit tests
```

## ☑️ ToDos
--

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
     "lint": "wireit",
     "lint:fix": "wireit",
     "lint:cpp": "find src packages/realm/bindgen/src -name '*.cpp' -or -name '*.h' -or -name '*.hpp' | xargs clang-format --dry-run --Werror",
-    "lint:cpp:fix": "find src packages/realm/bindgen/src -name '*.cpp' -or -name '*.h' -or -name '*.hpp' | xargs clang-format -i",
-    "postinstall": "git submodule update --init --recursive && npm install --prefix ./packages/realm/bindgen/vendor/realm-core"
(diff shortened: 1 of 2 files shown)


kraenhansen (approved):
Any docs we need to update in https://github.com/realm/realm-js/tree/main/contrib as a part of this?

kneth (approved):

package.json
elle-j, line 1: @kraenhansen, regarding your comment if any of our contrib files also needs updating. In general I think we need to take a look at [updating](https://github.com/realm/realm-js/blob/eca95b6c45217e61206afbfab5d96eec188484ae/contrib/building.md?plain=1#L153) a lot of information when it comes to building and contributing (after `bindgen` was merged into `main`).

So unsure if any single update to those files with regard to this PR would be helpful, perhaps adding an `npm install` after [this line](https://github.com/realm/realm-js/blob/eca95b6c45217e61206afbfab5d96eec188484ae/contrib/building.md?plain=1#L153).

PR #6208, opened 2023-10-20
Add updates to the addition of `deleteUser` in example app

elle-j:
## What, How & Why?

Makes a few updates to the addition of `deleteUser` to the example app:
* Moves `deleteUser` responsibility to `useDemoSyncTriggers`.
* Updates the README with the new use case.
* Removes rollup related changes in `package-lock.json`.

## ☑️ ToDos
--

PR #6221, opened 2023-10-30
Additions to custom user data

elle-j:
## What, How & Why?

Some additions/updates to the custom user data addition to the example app:
* Adds permission role to the custom data collection to only read and write to your own.
* Updates the `onUserCreation` callback to be run as an `Application Authentication` in order to run in the context of the existing app user.
* Types the user data for type safety.
* Updates the README.

## ☑️ ToDos
--

### DIFF
--- a/examples/rn-connection-and-error/frontend/app/hooks/useDemoSyncTriggers.ts
+++ b/examples/rn-connection-and-error/frontend/app/hooks/useDemoSyncTriggers.ts
@@ -18,7 +18,7 @@
 
 import {useCallback, useEffect, useState} from 'react';
 import {BSON, ConnectionState, UserState} from 'realm';
-import {useRealm, useUser} from '@realm/react';
+import {useApp, useRealm, useUser} from '@realm/react';
 
 import {Store} from '../models/Store';
 import {logger} from '../utils/logger';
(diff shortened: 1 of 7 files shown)


examples/rn-connection-and-error/frontend/app/hooks/useDemoSyncTriggers.ts
elle-j, line 1: The changes here (moving the responsibility of deleting the user to this file) are the same as in the [previous PR](https://github.com/realm/realm-js/pull/6208). But since this one will be merged after, it probably would have overwritten the other changes.

examples/rn-connection-and-error/frontend/app/screens/StoreScreen.tsx
```diff
@@ -72,6 +84,11 @@ export function StoreScreen() {
             />
           </View>
           <View style={styles.triggers}>
+            <View style={styles.status}>
+              <Text style={styles.statusText}>
+                Team: {user.customData.team}
+              </Text>
+            </View>
```
elle-j, line 91: Nit: Just moved this above the "Connected" text.

But also, removed the `?` after `customData` as it will always return a record.

examples/rn-connection-and-error/frontend/app/hooks/useDemoSyncTriggers.ts
elle-j, line 1: The changes here and some in `StoreScreen.tsx` (i.e. moving the responsibility of deleting the user to this file) are the same as in the [previous PR](https://github.com/realm/realm-js/pull/6208). But since this one will be merged after, the previous changes probably would have been overwritten had it not been changed here as well.

takameyer (approved):
Thanks so much for putting some final polish on this!

PR #6224, opened 2023-10-31
Unify a prepack command for packages

takameyer:
## What, How & Why?
Packages were using the `bundle` script for preparing for release.  Sometimes it requires more than bundling to prepare a package.  This unifies everything under `prepack`.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/.github/workflows/publish-package-release.yml
+++ b/.github/workflows/publish-package-release.yml
@@ -97,8 +97,8 @@ jobs:
     - name: Install node modules
       run: npm ci --ignore-scripts
 
-    - name: Bundle package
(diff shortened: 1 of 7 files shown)


takameyer:
Putting this in draft, just want this here for later discussion

kraenhansen:
An alternative is to add another `prepack` script which declares a dependency on `bundle` but that shouldn't happen for the `realm` package as we bundle this in a separate job than we package (we could of course also revisit that decision).

takameyer:
Leaving as is for now.

PR #6237, opened 2023-11-07
Create required deps for `useQuery`

### LINKED ISSUE #6235: `deps` argument of `useQuery` in `@realm/react` should default to the `type` argument
## Problem

When changing the `type` in the example below, the `objects` doesn't update.
This happens because the first `type` argument of `useQuery` isn't included in the third `deps` argument by default.

```typescript
const objects = useQuery(type);
```

## Workaround

```typescript
const objects = useQuery(type, undefined, [type]);
```

takameyer:
## What, How & Why?
* This ensures the type could be updated and re-render the collection.

This closes #6235

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-react/src/useQuery.tsx
+++ b/packages/realm-react/src/useQuery.tsx
@@ -44,9 +44,13 @@ export function createUseQuery(useRealm: () => Realm) {
   function useQuery<T extends Realm.Object<any>>(
     type: string | RealmClassType<T>,
     query: QueryCallback<T> = (collection: Realm.Results<T>) => collection,
-    deps: DependencyList = [],
+    deps: DependencyList = [type],
   ): Realm.Results<T> {
     const realm = useRealm();
+
+    // We need to add the type to the deps, so that if the type changes, the query will be re-run.
+    // This will be saved in an array which will be spread into the provided deps.
+    const requiredDeps = [type];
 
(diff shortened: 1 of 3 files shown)


packages/realm-react/src/useQuery.tsx
```diff
@@ -44,10 +44,14 @@ export function createUseQuery(useRealm: () => Realm) {
   function useQuery<T extends Realm.Object<any>>(
     type: string | RealmClassType<T>,
     query: QueryCallback<T> = (collection: Realm.Results<T>) => collection,
-    deps: DependencyList = [],
+    deps: DependencyList = [type],
```
kraenhansen, line 47: With the introduction of `requiredDeps` this could continue defaulting to `[]`, otherwise this would be duplicated in the final `deps`.

kraenhansen (approved):
It could be good to get a regression test for this 🙂

packages/realm-react/src/useQuery.tsx
takameyer: Good catch, I forgot to revert this.

takameyer:
I agree there should be tests for this, but the amount of effort involved is a bit too large.  I'll create a ticket.

PR #6242, opened 2023-11-10
Setter must replace embedded object instead of add a new object

### LINKED ISSUE #6239: Updating an element within an array of embedded schema behaves differently on Realm 12 and 11

Always

### Description

Realm 11.9.0 and 12.0.0 have different behavior when updating an element in an array within an embedded schema. If a change is desired, how can an element be updated in this case?

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

I've created an example to show the difference between Realm 12.0.0 and 11.9.0.

https://github.com/andreyukD/reproduce-realmjs-issue

Starting point was `npx create-expo-app MyAwesomeRealmApp --template @realm/expo-template`

I've added a new field `items` in `Task` schema
 
Commit with v12 https://github.com/andreyukD/reproduce-realmjs-issue/commit/897316aa157562b4724c5272c0dea90953919b9d

Branch with v11, code remained unchanged, only the Realm version was updated https://github.com/andreyukD/reproduce-realmjs-issue/commits/v11

https://github.com/realm/realm-js/assets/33383393/1cb8a88d-6c76-4c45-8614-8bb30673bb95

https://github.com/realm/realm-js/assets/33383393/262b46f0-7700-4e74-ad2d-e20235292a57

### Version

12.0.0 / 11.9.0

Local Database only

### Are you using encryption?

No

macOS 14.0

### Build environment

### Cocoapods version

_No response_

kneth:
## What, How & Why?

When replacing an element in a list of embedded object, the existing object should be removed. Calling `Obj::set_embedded()` is replacing while `Obj::insert_embedded()` will create a new object and add it to the list without touching the existing element.

This closes #6239 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -2095,6 +2095,7 @@ describe("Lists", () => {
       // check that we have successfully added a Group
       expect(divisions[0]["groups"].length).equals(2);
     });
+
     it("creating standalone embedded object throws", function () {
       // creating standalone embedded object is not allowed
       this.realm.write(() => {
@@ -2103,7 +2104,9 @@ describe("Lists", () => {
         }).throws(Error);
       });
     });
+
     it("supports adding embedded objects", function () {
+      // create a parent object with two embedded child objects
       this.realm.write(() => {
         this.realm.create(HouseOwnerSchema.name, {
           name: "Ib",
@@ -2117,6 +2120,7 @@ describe("Lists", () => {
       const ib = this.realm.objectForPrimaryKey(HouseOwnerSchema.name, "Ib");
       expect(ib.addresses.length).equals(2);
 
+      // add new embedded object
       this.realm.write(() => {
         expect(() => {
           //standalone object throws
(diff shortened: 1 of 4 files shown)


integration-tests/tests/src/tests/list.ts
```diff
@@ -2126,13 +2130,23 @@ describe("Lists", () => {
         ib.addresses.push({ street: "Njalsgade", city: "Islands Brygge" });
         expect(3).equals(ib.addresses.length);
```
elle-j, line 2131:
```suggestion
        expect(ib.addresses.length).equals(3);
```

integration-tests/tests/src/tests/list.ts
```diff
@@ -2126,13 +2130,23 @@ describe("Lists", () => {
...
       });
+
+      // update/replace one embedded child object
+      this.realm.write(() => {
+        ib.addresses[0] = { street: "Nørregade", city: "Vesterled" };
+      });
+
+      expect(3).equals(ib.addresses.length);
+      expect("Nørregade").equals(ib.addresses[0].street);
```
elle-j, line 2140:
```suggestion
      expect(ib.addresses.length).equals(3);
      expect(ib.addresses[0].street).equals("Nørregade");
```

takameyer (approved):
Wow, good fix!

PR #6261, opened 2023-11-17
Restore default for client reset

kneth:
## What, How & Why?

This closes #6290

Unfortunately I cannot find an easy way to test it as the callbacks are`undefined`.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Fixed an error "Invalid schema change (UPLOAD): cannot process AddColumn instruction for non-existent table" when using automatic client reset with recovery in dev mode to recover schema changes made locally while offline. ([realm/realm-core#7042](https://github.com/realm/realm-core/pull/7042))
 * When place an embedded object would create a new object and keep the original object too. ([#6239](https://github.com/realm/realm-js/issues/6239), since v12.0.0)
 * When setting an embedded object in a `Realm.List` by index, the new object would be inserted at the end rather than replacing the existing object at the given index. ([#6239](https://github.com/realm/realm-js/issues/6239), since v12.0.0)
+* When `SyncConfiguration.clientReset` was `undefined`, no client reset mode were set which could lead an app crash with the message `m_mode != ClientResyncMode::Manual`. The default mode is `RecoverOrDiscardUnsyncedChanges` and not callbacks are defined. ([#6260](https://github.com/realm/realm-js/issues/6260), since v12.0.0)
 
 ### Compatibility
 * React Native >= v0.71.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * Fixed an error "Invalid schema change (UPLOAD): cannot process AddColumn instruction for non-existent table" when using automatic client reset with recovery in dev mode to recover schema changes made locally while offline. ([realm/realm-core#7042](https://github.com/realm/realm-core/pull/7042))
 * When place an embedded object would create a new object and keep the original object too. ([#6239](https://github.com/realm/realm-js/issues/6239), since v12.0.0)
 * When setting an embedded object in a `Realm.List` by index, the new object would be inserted at the end rather than replacing the existing object at the given index. ([#6239](https://github.com/realm/realm-js/issues/6239), since v12.0.0)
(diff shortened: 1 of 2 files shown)


packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -366,7 +366,7 @@ function toSSLVerifyCallbackWithListArguments(verifyCallback: SSLVerifyCallback)
 function parseClientResetConfig(clientReset: ClientResetConfig | undefined, onError: ErrorCallback | undefined) {
   if (!clientReset) {
     return {
-      clientResyncMode: undefined,
+      clientResyncMode: toBindingClientResetMode(ClientResetMode.RecoverOrDiscardUnsyncedChanges),
```
takameyer, line 369:
```suggestion
      clientResyncMode: toBindingClientResetMode(ClientResetMode.RecoverUnsyncedChanges),
```
I have changed my opinion on this.  Perhaps having `discard` as the default is too destructive.  Let's align with the documentation and other SDKs:
https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/#recover-unsynced-changes

packages/realm/src/app-services/SyncConfiguration.ts
```diff
@@ -366,7 +366,7 @@ function toSSLVerifyCallbackWithListArguments(verifyCallback: SSLVerifyCallback)
 function parseClientResetConfig(clientReset: ClientResetConfig | undefined, onError: ErrorCallback | undefined) {
   if (!clientReset) {
     return {
-      clientResyncMode: undefined,
+      clientResyncMode: toBindingClientResetMode(ClientResetMode.RecoverUnsyncedChanges),
```
kraenhansen, line 369: I would prefer just referring directly to the core enum here:
```suggestion
      clientResyncMode: binding.ClientResetMode.Recover,
```

kraenhansen (approved):
I have just a minor suggestion.
Also, to test this, we could add at least the mode on the config getter for the sync session: https://github.com/realm/realm-js/blob/main/packages/realm/src/app-services/SyncSession.ts#L308, but that's just an idea.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * Fixed an error "Invalid schema change (UPLOAD): cannot process AddColumn instruction for non-existent table" when using automatic client reset with recovery in dev mode to recover schema changes made locally while offline. ([realm/realm-core#7042](https://github.com/realm/realm-core/pull/7042))
 * When place an embedded object would create a new object and keep the original object too. ([#6239](https://github.com/realm/realm-js/issues/6239), since v12.0.0)
 * When setting an embedded object in a `Realm.List` by index, the new object would be inserted at the end rather than replacing the existing object at the given index. ([#6239](https://github.com/realm/realm-js/issues/6239), since v12.0.0)
+* When `SyncConfiguration.clientReset` was `undefined`, no client reset mode were set which could lead an app crash with the message `m_mode != ClientResyncMode::Manual`. The default mode is `RecoverOrDiscardUnsyncedChanges` and not callbacks are defined. ([#6260](https://github.com/realm/realm-js/issues/6260), since v12.0.0)
```
elle-j, line 17:
```suggestion
* When `SyncConfiguration.clientReset` was `undefined`, no client reset mode was set which could lead to an app crash with the message `m_mode != ClientResyncMode::Manual`. The default mode is now `RecoverUnsyncedChanges` and no callbacks are defined. ([#6260](https://github.com/realm/realm-js/issues/6260), since v12.0.0)
```

elle-j (approved):
LGTM, with just some updates to the CHANGELOG entry 🙂 

1. To add to @kraenhansen's comment:
> Also, to test this, we could add at least the mode on the config getter for the sync session: https://github.com/realm/realm-js/blob/main/packages/realm/src/app-services/SyncSession.ts#L308

I also think ☝️ is a good idea (adding the `clientReset.mode` to the returned object from `SyncSession.config`) as it more accurately reflects what the user has configured. If we add it, it may be convenient to add `newRealmFileBehavior.type` and `existingRealmFileBehavior.type` while we're at it?

2. Regarding the original error of `m_mode != ClientResyncMode::Manual`, has this been tested manually to see that it actually solves the issue?

3. P.S. the PR description mentions the wrong issue number (I believe it should be 6260).

kraenhansen (approved):

kneth:
@kraenhansen @elle-j I have created https://github.com/realm/realm-js/issues/6274 to track the suggestions about the config getter.

> Regarding the original error of m_mode != ClientResyncMode::Manual, has this been tested manually to see that it actually solves the issue?

No, the fix is done by reviewing the code - and reading the corresponding Core code.

PR #6267, opened 2023-11-21
Upgrading typescript, rollup, eslint and plugins

kraenhansen:
## What, How & Why?

While working on #6266 I tried a solution that required updating the `typescript` package (which transitively required updates of `rollup` and `eslint`). I also choose to hoist the rollup plugins to the root, since they all have peer dependencies on `rollup` which is already only present in the root package.

Because of the updates, a few places needed fixes (in the babel plugin and some rollup configs).

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
Seems this is blocked on `@react-native/eslint-config` upgrading [its dependency on `eslint-plugin-prettier`](https://github.com/facebook/react-native/blob/main/packages/eslint-config-react-native/package.json#L32) - I'll draft this for now.

takameyer (approved):

takameyer:
Appears there are some linting errors and the tests are not working.

kraenhansen:
This seems outdated now that we have @depfu running again.

Issue #6268, opened 2023-11-21
Error when using mapTo on a property of type List


GregoryEsberci:
### How frequently does the bug occur?

Always

### Description

When mapTo is used on a property of type List, an error occurs when trying to access it.

### Stacktrace & log output

```shell
ERROR  [Error: Property 'test_list' does not exist on 'Task' objects] 
    at TaskList (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:162361:21)
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:59191:43)
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:59191:43)
    at TaskScreen (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:161948:22)
    at anonymous (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:123651:27)
    at RCTSafeAreaView
    at AppNonSync
    at App
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:59191:43)
    at RCTView
    at View (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:59191:43)
    at AppContainer (http://localhost:8081/index.bundle//&platform=ios&dev=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.example.ExampleRnTodoList:59035:36)
[...]

PR #6269, opened 2023-11-21
Bug in mapTo on Realm.List

kneth:
## What, How & Why?

In #6268, `mapTo` is reported not to work on lists.  We have tests in https://github.com/realm/realm-js/blob/v11/integration-tests/tests/src/tests/alias.ts but they only verify that the schema has been transformed.

Attempts to reproduce bug:

* [x] Add test to babel plugin
* [x] Add test to use alias after schema verification (in `alias.ts`)

The bug seems to hide is the getting but only for list properties.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@
 * None
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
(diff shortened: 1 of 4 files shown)


takameyer (approved):

elle-j (approved):
Clean fix 👌

Issue #6259, opened 2023-11-16
Improve useQuery arguments so react-hooks/exhaustive-deps recognise missing deps


bimusiek:
### Problem

Currently there is no way to know if you forgot `deps` besides testing. We would love to use `react-hooks/exhaustive-deps` to recognise the missing deps in the `useQuery` function.

### Solution

If we move the `type` argument as the last one, so example usage of useQuery:
```
  const results = useQuery<ImporterAuthenticationModel>(
    (objects) =>
      objects.filtered(query, ...args).sorted([
        ['importerId', false],
        ['title', false],
      ]),
    [query, args],
    ImporterAuthenticationModel.schema.name
  );
```
Then `react-hooks/exhaustive-deps` correctly recognises missing deps (as it expects deps to be on second position and callback to be on the first one like in useEffect).

### Alternatives

No alternatives really. For our code base we copy-pasted `useQuery` source code and modified it to suit our needs, but I think it would be really beneficial to all developers using @realm/react.

Maybe alternative is to fork `react-hooks/exhaustive-deps` and make it work for `realm-hooks` :)

### How important is this improvement for you?

Fairly niche but nice to have anyway

### Feature would mainly be used with

Atlas Device Sync

kneth:
@bimusiek Thank you for the suggestion. As you can guess, we didn't design `@realm/react` with `react-hooks/exhaustive-deps` fresh in our memory :smile:

> Maybe alternative is to fork ...

Maintaining additional code isn't such a good idea, and we should avoid it.

As `@realm/react` is still early in development, I suppose the proposed change is acceptable.

takameyer:
Ah, good catch.  This would unfortunately be a breaking change, so we'll have to consider how we approach this.

kneth:
> unfortunately be a breaking change

Imho, the major version is 0, so we can allow to introduce breaking changes

PR #6272, opened 2023-11-21
Possible solution for exhaustive deps

bimusiek:
## What, How & Why?
Possible solution for exhaustive deps

This closes https://github.com/realm/realm-js/issues/6259

## ☑️ ToDos
Test it, but I cannot build the package as `realm/bindgen` is not published on NPM registry and not sure if this is something I have to build or I simply do not have access to. 

Btw, would be amazing to have README also for contributors, would save A TON of time to figure out the inner workings around buildtools, scripts and ways you have the repo setup.

### DIFF
--- a/packages/realm-react/.eslintrc.json
+++ b/packages/realm-react/.eslintrc.json
@@ -8,7 +8,12 @@
   "rules": {
     "@typescript-eslint/explicit-function-return-type": "off",
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "(^useQuery$)"
+      }
+    ],
     "jsdoc/check-tag-names": "off",
     "jsdoc/require-jsdoc": "off",
     "jsdoc/require-returns": "off",
--- a/packages/realm-react/src/RealmProvider.tsx
+++ b/packages/realm-react/src/RealmProvider.tsx
@@ -123,7 +123,7 @@ export function createRealmProvider(
       if (realmRef) {
         realmRef.current = realm;
       }
-    }, [realm]);
+    }, [realm, realmRef]);
 
     useEffect(() => {
       const realmRef = currentRealm.current;
--- a/packages/realm-react/src/__tests__/useQueryHook.test.tsx
+++ b/packages/realm-react/src/__tests__/useQueryHook.test.tsx
@@ -112,4 +112,22 @@ describe("useQueryHook", () => {
 
     expect(collection[99]).toBe(undefined);
   });
+
+  it("should support reversed order of arguments for exhausive deps", () => {
+    // Eslint will trigger warning here with
+    // 120:92  warning  React Hook useQuery has a missing dependency: 'gender'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
+    // which is correct behaviour.
+    const { result } = renderHook(
+      ({ gender }) => useQuery<IDog>((objects) => objects.filtered("gender = $0", gender), [], "dog"),
+      {
+        initialProps: { gender: "male" },
+      },
+    );
+    const collection = result.current;
+
+    expect(collection).not.toBeNull();
+    expect(collection.length).toBe(2);
+    expect(collection[0]).toMatchObject(testDataSet[0]);
+    expect(collection[1]).toMatchObject(testDataSet[3]);
+  });
 });
--- a/packages/realm-react/src/index.tsx
+++ b/packages/realm-react/src/index.tsx
@@ -218,3 +218,5 @@ export { useUser, UserProvider } from "./UserProvider";
 export * from "./useAuth";
 export * from "./useEmailPasswordAuth";
 export * from "./types";
+export * from "./useQuery";
+export * from "./useObject";
--- a/packages/realm-react/src/useQuery.tsx
+++ b/packages/realm-react/src/useQuery.tsx
@@ -25,12 +25,36 @@ type RealmClassType<T = any> = { new (...args: any): T };
 type QueryCallback<T> = (collection: Realm.Results<T>) => Realm.Results<T>;
 type DependencyList = ReadonlyArray<unknown>;
 
+function simplifyArguments<T extends Realm.Object<any>>(
+  typeOrQuery: string | RealmClassType<T> | QueryCallback<T>,
+  queryOrDeps: QueryCallback<T> | DependencyList = (collection: Realm.Results<T>) => collection,
+  depsOrType: DependencyList | string | RealmClassType<T> = [],
+): { type: string | RealmClassType<T>; query: QueryCallback<T>; deps: DependencyList } {
+  if (Array.isArray(queryOrDeps)) {
+    return {
+      type: depsOrType as unknown as string | RealmClassType<T>,
+      query: typeOrQuery as unknown as QueryCallback<T>,
+      deps: queryOrDeps as unknown as DependencyList,
+    };
+  }
+  return {
+    type: typeOrQuery as unknown as string | RealmClassType<T>,
+    query: queryOrDeps as unknown as QueryCallback<T>,
+    deps: depsOrType as unknown as DependencyList,
+  };
+}
 /**
  * Generates the `useQuery` hook from a given `useRealm` hook.
  * @param useRealm - Hook that returns an open Realm instance
  * @returns useObject - Hook that is used to gain access to a {@link Realm.Collection}
  */
 export function createUseQuery(useRealm: () => Realm) {
+  function useQuery<T>(query: QueryCallback<T>, deps: DependencyList, type: string): Realm.Results<T & Realm.Object<T>>;
+  function useQuery<T extends Realm.Object<any>>(
+    query: QueryCallback<T>,
+    deps: DependencyList,
+    type: RealmClassType<T>,
+  ): Realm.Results<T & Realm.Object<T>>;
   function useQuery<T>(
     type: string,
     query?: QueryCallback<T>,
@@ -42,10 +66,11 @@ export function createUseQuery(useRealm: () => Realm) {
     deps?: DependencyList,
   ): Realm.Results<T>;
   function useQuery<T extends Realm.Object<any>>(
-    type: string | RealmClassType<T>,
-    query: QueryCallback<T> = (collection: Realm.Results<T>) => collection,
-    deps: DependencyList = [],
+    typeOrQuery: string | RealmClassType<T> | QueryCallback<T>,
+    queryOrDeps: QueryCallback<T> | DependencyList = (collection: Realm.Results<T>) => collection,
+    depsOrType: DependencyList | string | RealmClassType<T> = [],
   ): Realm.Results<T> {
+    const { type, query, deps } = simplifyArguments(typeOrQuery, queryOrDeps, depsOrType);
     const realm = useRealm();
 
     // We need to add the type to the deps, so that if the type changes, the query will be re-run.
--- a/packages/realm-react/src/useQuery.tsx
+++ b/packages/realm-react/src/useQuery.tsx
@@ -87,6 +87,7 @@ export function createUseQuery(useRealm: () => Realm) {
     // We want the user of this hook to be able pass in the `query` function inline (without the need to `useCallback` on it)
     // This means that the query function is unstable and will be a redefined on each render of the component where `useQuery` is used
     // Therefore we use the `deps` array to memoize the query function internally, and only use the returned `queryCallback`
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     const queryCallback = useCallback(query, [...deps, ...requiredDeps]);
 
     // If the query function changes, we need to update the cachedCollection


kraenhansen:
Thanks for taking the time to look into this 👍 

> I cannot build the package as realm/bindgen is not published on NPM registry

The `@realm/bindgen` package is vendored in as a submodule of the repository: https://github.com/realm/realm-js/tree/main/packages/realm/bindgen/vendor

Unfortunately our docs on building is a bit outdated since the major changes we did in v12, but at least some of the instructions here should be applicable: https://github.com/realm/realm-js/blob/main/contrib/building.md#cloning-the-repository

Specifically, you need to initialize submodules after cloning the repo:

```
```

After that, you should be able to simply run `npm install` in the root and then run `npm test` in the `packages/realm-react` package. If that's not possible, we have something we need to fix 🙂

bimusiek:
So, trying to make it work on my machine, there are some issues and I am not sure if I should commit the changes for those, or am I doing something wrong.

My node: v20.9.0
I am using `yarn` to install everything. 

First of all, the bindgen submodule HEAD commit sha in parent repo was not commited recently so I had to checkout the master for this submodule.

# Rollup
The first issue was with rollup:
```
[!] RollupError: Node tried to load your configuration file as CommonJS even though it is likely an ES module. To resolve this, change the extension of your configuration to ".mjs", set "type": "module" in your package.json file or pass the "--bundleConfigAsCjs" flag.

Original error: Cannot use import statement outside a module
https://rollupjs.org/command-line-interface/#bundleconfigascjs
/Users/bimusiek/Developer/FreeYourMusic/realm-js/packages/realm-common/rollup.config.js:19
import typescript from "@rollup/plugin-typescript";
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1153:20)
    at Module._compile (node:internal/modules/cjs/loader:1205:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Function.Module._load (node:internal/modules/cjs/loader:938:12)
    at cjsLoader (node:internal/modules/esm/translators:284:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:234:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at ModuleLoader.import (node:internal/modules/esm/loader:316:24)
```
Changing to .mjs helped though. Then I needed to do ` assert { type: "json" }` for `package.json` import. And change `dts()` to `dts.default()` to make it work. 

# CommanderJS
```
/Users/bimusiek/Developer/FreeYourMusic/realm-js/packages/realm/bindgen/vendor/realm-core/bindgen/src/program.ts:67
  .createOption("-s, --spec <spec...>", "Path of the API specification")
   ^


TypeError: program.createOption is not a function
    at <anonymous> (/Users/bimusiek/Developer/FreeYourMusic/realm-js/packages/realm/bindgen/vendor/realm-core/bindgen/src/program.ts:67:4)
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)
```

Basically, it installs commander `2.20.3` instead of required >10. 
So I have to add to `package.json` of `realm-core`:

```
 
    "commander": "^10.0.1",
```

# CMake 
Now, from TS point everything is building as expected, but I get an error with Realm compilation:

```
Running 'clang' formatter
npx clang-format -i /Users/bimusiek/Developer/FreeYourMusic/realm-js/packages/realm/bindgen/build/node_init.cpp
✅ [../realm:bindgen:generate:typescript] Executed successfully
✅ [../realm:bindgen:generate:wrappers] Executed successfully
🏃 [../realm:bundle] Running command "rollup --config"
(node:59020) ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:59020) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time

src/platform/node/index.ts → ./dist/bundle.node.js...
(!) Circular dependencies
src/internal.ts -> src/errors.ts -> src/internal.ts
src/internal.ts -> src/assert.ts -> src/internal.ts
src/internal.ts -> src/JSONCacheMap.ts -> src/internal.ts
...and 36 more
created ./dist/bundle.node.js in 1.4s

src/platform/react-native/index.ts → ./dist/bundle.react-native.js...
(!) Circular dependencies
src/internal.ts -> src/errors.ts -> src/internal.ts
src/internal.ts -> src/assert.ts -> src/internal.ts
src/internal.ts -> src/JSONCacheMap.ts -> src/internal.ts
...and 36 more
created ./dist/bundle.react-native.js in 1.1s

src/index.ts → dist/bundle.d.ts...
[ 98%] Building CXX object CMakeFiles/realm-js.dir/node_init.cpp.o
created dist/bundle.d.ts in 655ms
✅ [../realm:bundle] Executed successfully
/Users/bimusiek/Developer/FreeYourMusic/realm-js/packages/realm/bindgen/build/node_init.cpp:22:10: fatal error: 'realm/object-store/util/bson/bson.hpp' file not found
#include <realm/object-store/util/bson/bson.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[3]: *** [CMakeFiles/realm-js.dir/node_init.cpp.o] Error 1
make[2]: *** [CMakeFiles/realm-js.dir/all] Error 2
make[1]: *** [CMakeFiles/realm-js-node.dir/rule] Error 2
make: *** [realm-js-node] Error 2
❌ [../realm:bindgen:build:node] Failed with exit status 2
```

bimusiek:
OK, all of the previous issues are my mistakes. If I keep the proper commit in submodule (as it was commited by the parent), then the build works. And by using `npm` I am getting correct `commander` version. Rollup was also fixed by using npm.

So lesson learned for future, stick with the package manager that the project is using 😄

bimusiek:
@kraenhansen I have written one test for the new way of using useQuery. Please take a look.
Not all tests are passing as it looks like I don't have access to some secret keys you guys use in your CI workflow.

Same with my local test run:
```
    FetchError: request to http://localhost:9090/api/admin/v3.0/auth/providers/local-userpass/login failed, reason:
```
I tried to setup docker container, but then again there are some AWS keys I would need (probably to offload docker to AWS?). I did not play with it longer to make it work as it was not crucial for me to improve `useQuery`. 

Also, there was another missing dependency with `RealmProvider` that I commited, but if it is wrong then tell me and I will add `eslin-disable-next-line` so it won't popup again.

bimusiek:
I have published package to [@hernas/realm-react](https://www.npmjs.com/package/@hernas/realm-react) so I am able to easily test it in our project and everything looks good now. 

I had to export `createUseQuery` and `createUseObject` due to the way our app is written. We have extensive `Realm` initialisation process in place, to make sure it handles all the quirks and issues from past few years of using Realm, so we don't want to switch to `createRealmContext` where we pass the `Realm.Configuration`. 

I don't think it is something bad. Now consumers of this library, can still use their own `Realm` initialisation code and start using `@realm/react` with it.

kraenhansen:
Thanks again for taking the time to look into this. While I'm really happy you've found a way to solve the immediate issue, I'm also afraid that the added overload could cause a bit of confusion. I'm also not a super fan of the fact that the `type` becomes the third argument to the hook.

If I understand correctly, this overload is because the eslint rule expects `deps` to be the second argument.
Could we make this overload take just two arguments, the first being an object with `query` and `type` fields and the second being the `deps`?

An alternative solution would be to create a `createQueryHook<T>(type: string)` function, which would return a `useQuery(query: QueryCallback<T>, deps?: DependencyList): Realm.Results<T>`, they users could export a bunch of these to query their data:

```typescript
export const usePersons = createQueryHook<Person>("Person");
export const useDogs = createQueryHook<Dog>("Dog");

// `createQueryHook` could even have an overload for class-based models, like some of our other APIs:
export const usePersons = createQueryHook(Person);

// And in a consuming component:
const twenty = 20;
const teenagers = usePersons(persons => persons.filtered("age < $0", twenty), [twenty]);
```

bimusiek:
@kraenhansen Interesting idea, this would work even better.

Looks much cleaner to me. 
And in our codebase we could have something like `ImporterAuthenticationModel.useHook` which would be simply `createQueryHook(ImporterAuthenticationModel)`. 

And as we already have our own `BaseModel` that all models extends from to add new functionalities to the Realm models, we can easily get hooks working then. 

Let me get on that, just to be sure, something like this:
```ts
const createQueryHook = prepareCreateQueryHook(useRealm)
const useRealmDogs = createQueryHook <Dog>("dogs");
const useRealmOwners = createQueryHook(OwnerModel); // where OwnerModel extends Realm.Object
```

But then, you want to keep legacy `useQuery` hook as it was? And export additional new `createQueryHook` / `prepareCreateQueryHook`?

kraenhansen:
Yep. Something like that. Perhaps the "prepare create" becomes a bit verbose, but you've got the gist of it. I'd like to get @takameyer's and @elle-j's thoughts on the above.

BTW: To work around the issue you're having with the Realm having to be opened outside of the provider, it might make sense for us to create a simple component (let's call it RawRealmProvider or RealmInstanceProvider) taking the Realm instance as a prop instead of a config, and use that from the current RealmProvider or perhaps simply overload the existing providers interface to take either props from config or a single realm prop. That way you don't need to call a prepareCreateQueryHook, but you can just use the defaults exported by the package.

takameyer:
I really like what was proposed.  I'll just present my spin on it.

Creating a `createModelHooks` method would be an straightforward step that we could implement right now and reap the benefits thereof.

My idea would be to make this exportable on the default `RealmProvider` and also add it to the return object of `createRealmContext`.  

The `createModelHook` could return
```tsx
import {createModelHooks} from '@realm/react';

// or
import {createModelHooks} from '../MyRealmContext';

class Person extend Realm.Object{} //some class based model

const [usePersonById, usePerson] = createModelHooks(Person);
```

Each hook returned would be bound to the realm and model and be an extension on `useObject` and `useQuery` respectively.

bimusiek:
From our perspective as users of the library, I would love to be able to use as much barebone functions as we can. We already have our own RealmProvider that creates a context for us and we would love to stick with what we have, as it is quite tailored for our needs. We support from one codebase Electron, React Native but also web, which we cannot use Realm in. 

Another long term way of doing this, would be to have realm context creation ability to accept list of models you want to work with in React.

Then we could have something like:
```ts
enum SchemaName {
    Dog = 'Dog',
    Cat = 'Cat'
}

abstract class RealmModel {
    static schemaName: SchemaName;

    static get() {
        console.log('get');
    }
}

class Dog extends RealmModel {
    static schemaName = SchemaName.Dog

    static woof() {
        console.log('woof');
    }
}

class Cat extends RealmModel {
    static schemaName = SchemaName.Cat

    static meow() {
        console.log('meow')
    }
}

type SchemaWithModels = {
    [SchemaName.Cat]: typeof Cat;
    [SchemaName.Dog]: typeof Dog
};

function createReactUtilities<Model extends typeof RealmModel>(models: Model[]) {
    type SchemaNameHooks = {
        [K in SchemaName as K extends string ? `use${K}` : never]:
        (callback: (query: SchemaWithModels[K]) => void, deps: any[]) => void
    }

    return models.reduce((hooks, model) => {
        return {
            ...hooks,
            [`use${model.schemaName}`]: (callback: any) => {
                callback(model);
            }
        }
    }, {}) as SchemaNameHooks;
}

const utils = createReactUtilities([Dog, Cat]);
utils.useCat((query) => {
    query.get();
    query.meow();
}, [])
utils.useDog((query) => {
    query.get();
    query.woof();
}, [])
```
[Playground Link Here ](https://www.typescriptlang.org/play?#code/KYOwrgtgBAygxgC2BAhgORRYUDeAoKQqAEQHsBzKAXigHIzzaAaAogYRQBdq6PPa8AXzx4UAIwDOnAE4o43OABsUEiVABKwFIogBZUgBNgi3K0JSuASzhQJiZOkzAAXLHuoMWANwiitzlY25MCcABQAlKZ+fnCkIBKkisAAdIoUobTB-OE+fsLCeEoqagxQwAAenKAGapraeobGUUQWnNa27o5YPPBIHk7JDL4tAW02AO6kpABmEc3RULHxiSlp5BmTM7Q5ZlD5IkWqUHxlldW1Wjr6Rib4fq3tdn1d2DS9Dp4pfMPmo+1YpHGczuC0WcQSSVS6VoAPG212+zwnAAngAHbDvVAAdUsnAQ12MahoIMIAG1MS9knwALquFHombHLi5IjkzqfQYUWlQenARlDQQ+PDTMAgeSWOKLaRaKp1eQAVTailxlmAEgAPASTBUqiAajy0XzphpLg0bgA+UIQRqKCSuLWk6mREkG9FuZ6fAASUwA1kT5tFSQBpKCWEDuj5OKAqKAhnXnfzSMOUAD8UAABmAJMAACQ4IOCdNQVwgYAAN2A0lpuz8oTg2kUYjkPtcoQAjmBK8jXBScXitRJg07qOaoGXSJYDEwoEZUXboyBkY7IlRR+PJwiflBpZwwNJw9abhJktKDGA4MBQqEEL6JNPD8YV6OXX4d3vwy-QclvzfSH6WKCgaZtmeYPooyRPJGWCFtydYNk2cAtguyJPgGgExPBzZWjaOzoUQwiAQR+HTjggiRDGFJerePgFEsUhQGASr+nA0pcMAcqcIqljKm0aqhKSDDTjSOyMdxx5ZsAfBXh2XaoS6MnSMiyRZBELKEApSmwqpQjTsueCibayQSQw0mdopcm7BpykhNpfhWZssw7IIulOkAA)

This is just proof of concept, so there has to be more work & thought put into it. But theoretically it is possible to infer proper types and have hooks created. In the example above, `query` has type `Cat` in `useCat` and type `Dog` in `useDog`

kraenhansen:
@bimusiek again - we really appreciate your engagement here!
No-one knows the ideal patterns for using our SDK than our users ...

I've created https://github.com/realm/realm-js/issues/6284 to track the proposal to add the creation of these class-aware hooks. We've discussed your concrete proposal on the team and we found the overload a bit too envolved (expecially the fact that the `type` is the third argument is a deal-breaker for us).

You're welcome to either
1) adopt this to the alternative of adding an overload taking an options object (with `type` and `query`) or alternatively
2) close this PR and create another one for #6284 (in which case we should probably iron out the API beforehand - as we value the time your putting into this).
3) close this PR and have us implement a fix for either #6259 or #6284 as it gets to the top of our backlog.

Would this either of those ways forward work for you?

elle-j:
Adding the following for context and discoverability:

* React has had an [issue](https://github.com/facebook/react/issues/19644) open regarding this for a while (and a [more recent one](https://github.com/facebook/react/issues/25443)).
* They also had two proposed solutions ([PR1](https://github.com/facebook/react/pull/18861), [PR2](https://github.com/facebook/react/pull/21719)) that were accidentally closed.
  * For now, it doesn't look like they plan on customizing the `additionalHooks` option.

bimusiek:
@kraenhansen I understand! :) 

I am gonna close this PR then and simply wait for you guys to find the best solution that works for everyone. We will stick with our package for now, until there is new version of @realm/react with the fixes for exhaustive deps, in whatever form they will appear.

And I would like to say, that RealmJS is going through amazing improvements and as the early users we are quite excited to be part of the ecosystem. You are doing great job with the React integration and overall developer experience.

PR #6275, opened 2023-11-22
Updates to advanced custom user data

elle-j:
## What, How & Why?

Some updates to the custom user data additions:

* Rule name updates to match the rule.
* Make Atlas functions private where possible.
* Fixed the auth trigger for creating custom user data document.
* Added trigger and function for deleting the custom user data document if the app user gets deleted.
* Replaced manual polling of connection state with using our connection listener.
* Made some smaller miscellaneous updates to code and comments.
* Updated Realm to 12.3.0 in the package.json.
* Updated the README.

## ☑️ ToDos
--


examples/rn-connection-and-error/backend/data_sources/mongodb-atlas/StoreDemo/Kiosk/rules.json
```diff
@@ -3,7 +3,7 @@
     "collection": "Kiosk",
     "roles": [
         {
-            "name": "readAndWriteAll",
+            "name": "readAndWriteSpecificStore",
```
elle-j, line 6: Updating the name to match the rule better.

examples/rn-connection-and-error/backend/functions/deleteCustomUserDataDoc.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable */
+
+/**
+ * Deletes a custom user data (`Users`) document.
+ */
+exports = async function deleteCustomUserDataDoc({ user }) {
```
elle-j, line 6: Added this to make it easier when playing around with the example and not needed to manually go and delete docs. Is probably good to showcase this as well to show what's reasonable to do in production.

examples/rn-connection-and-error/frontend/app/hooks/useDemoSyncTriggers.ts
```diff
@@ -155,38 +170,43 @@ export function useDemoSyncTriggers() {
...
-  const refreshSession = useCallback(async () => {
-    logger.info('Pausing Session...');
-    realm.syncSession?.pause();
+  const refreshSession = useCallback(() => {
+    // To refresh the session, we first pause the session, causing sync to get
+    // disconnected. Once disconnected, the `handleConnectionChange()` listener
+    // will be called, wherein we manually resume the session if `reconnectAfterDisconnect`
+    // is set to `true`. This pattern can be used to react to connection changes.
+    reconnectAfterDisconnect = true;
+    disconnect();
```
elle-j, line 187: Figured it may be good to show users how to leverage the pattern of reacting to connection state changes via the connection listener. This way users won't have to have their own implementations of checking the connection in intervals, and it could be more reliable as well. (The connection listener callback will also take care of logging the new connection state.)

To better see this refactor, you can check out [this commit](https://github.com/realm/realm-js/commit/3935cdd94ee910c17d42502fe2815fefc608e688).

examples/rn-connection-and-error/frontend/app/providers/StoreProvider.tsx
```diff
@@ -62,19 +68,28 @@ export function StoreProvider({children}: PropsWithChildren) {
...
+  // (Having `useEffect` rely on `user.customData.storeId` will not
+  // be sufficient in this case as the `storeId` is an `ObjectId`
+  // whose reference will change on each store update.)
+  const customDataStoreId = user.customData.storeId;
+  const storeIdChanged =
+    mostRecentStoreId?.toHexString() !== customDataStoreId.toHexString();
+  if (storeIdChanged) {
+    logger.info(`User's current store ID: ${customDataStoreId}`);
+    mostRecentStoreId = customDataStoreId;
+  }
```
elle-j, line 81: This merely exists to log the current store ID being used whenever it changes. Previously it was logging a bit too much as it got logged whenever something changed in the store. Since we do show the store ID in the UI header, I think removing this particular logging part altogether could also be a good idea to reduce the code.

examples/rn-connection-and-error/frontend/app/providers/StoreProvider.tsx
```diff
@@ -62,19 +68,28 @@ export function StoreProvider({children}: PropsWithChildren) {
...
+   * Adds a store. This demo app is syncing only 1 store at a time. Thus, if
+   * a store has already been created, this function will return immediately.
    */
   const addStore = useCallback(() => {
     if (store) {
       return;
     }
     realm.write(() => {
-      realm.create(Store, {});
+      realm.create(Store, {_id: new BSON.ObjectId()});
```
elle-j, line 92: Since we updated to add the store on the backend, users probably won't call this function. Could be good to show though if a developer decides to do it differently. In this case, the `_id` here must be provided.

examples/rn-connection-and-error/frontend/app/screens/StoreScreen.tsx
```diff
@@ -78,79 +73,76 @@ export function StoreScreen() {
...
+            </View>
           </View>
+        </>
+      ) : (
+        // Store isn't yet loaded, perhaps while switching stores
+        <View style={styles.createStore}>
+          <Loading />
         </View>
-      </>
+      )}
```
elle-j, line 145: Made some updates to the layout here since the UI was flickering a bit when switching the store. Or was the restructuring due to issues in refreshing the session by the way?

examples/rn-connection-and-error/frontend/app/providers/StoreProvider.tsx
takameyer: No problem.  The logs were mostly for me to verify things were working as expected, since this example was non-trivial to setup.

elle-j: Cool cool, I'll remove it then 👍

examples/rn-connection-and-error/frontend/app/providers/StoreProvider.tsx
takameyer: We could also set this in the default method of `_id` in the model itself.

takameyer: But I'm actually leaning towards just removing the method if it's not used.

examples/rn-connection-and-error/frontend/app/screens/StoreScreen.tsx
takameyer: The goal was to keep visible the buttons that may need to be called to ensure the client reset is properly handled.  For instance, if one were to refresh the session and not update the current user data, then we will be stuck in a loading screen and not able to move forward.

examples/rn-connection-and-error/frontend/app/providers/StoreProvider.tsx
elle-j: Could definitely do that too 🙂 I'll push an update

takameyer (commented):
These changes are an outstanding polish to make these examples even more useful for our users.  Thanks for taking the time to clean this up :)

examples/rn-connection-and-error/frontend/app/screens/StoreScreen.tsx
elle-j: Thanks for clarifying, updates comin'

PR #6281, opened 2023-11-24
Update PR workflow to import cert and sign Node.js binary

kraenhansen:
## What, How & Why?

This is an attempt to enable codesigning of the Node.js prebuilt binary.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
Closing this for now, since the solution was getting quite involved and it seems the established pattern is for the dependent app to perform the signing.

PR #6282, opened 2023-11-26
Fixed babel plugin add extra name property

XHMM:
## What, How & Why?

When defining class with a static property `name`, the generated js file containes two `name` fields, one is class name and another is the static property 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/babel-plugin/src/plugin/index.ts
+++ b/packages/babel-plugin/src/plugin/index.ts
@@ -402,11 +402,16 @@ function visitRealmClass(path: NodePath<types.ClassDeclaration>) {
     .map(visitRealmClassStatic)
     .filter((property) => property) as types.ObjectProperty[];
 
-  const schema = types.objectExpression([
-    types.objectProperty(types.identifier("name"), types.stringLiteral(className)),
-    types.objectProperty(types.identifier("properties"), types.objectExpression(schemaProperties)),
+  const hasStaticName = schemaStatics.some(p => p.key.name === "name");
+  const properties = [
+    types.objectProperty(types.identifier("properties"),
+    types.objectExpression(schemaProperties)),
     ...schemaStatics,
-  ]);
+  ];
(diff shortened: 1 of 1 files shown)


XHMM:
There is no doc for how to running test locally and I am not familier with `wireit`,  so I skipped writing tests

kraenhansen:
@XHMM can you share an example of code that would trigger this?

> There is no doc for how to running test locally

It should be as simple as:
- Checking out your fork of the repo
- `git submodule update --init --recursive`
- `npm install`
- `npm test --workspace @realm/babel-plugin`

XHMM:
Here is a code sample:

**schema defination using ts**:
```ts


export class RecordModel extends Realm.Object<RecordModel, "recordFrom"> {
  // @ts-ignore
  static name = "Record";
  static primaryKey = "_id";

  _id: Realm.Types.ObjectId = new Realm.BSON.ObjectId();
  @Realm.index recordAt: Realm.Types.Date = new Date();
  recordFrom!: string;
}
```

**after babel transform with current plugin:** (some unrelated code was omitted)
```js
(0, _defineProperty2.default)(RecordModel, "name", "Record");
RecordModel.primaryKey = "_id";
RecordModel.schema = {
  name: "RecordModel", // class name
  properties: {
    _id: {
      type: "objectId",
      default: function _default() {
        return new _realm.default.BSON.ObjectId();
      },
    },
    recordAt: {
      type: "date",
      default: function _default() {
        return new Date();
      },
      indexed: true,
    },
    recordFrom: { type: "string" },
  },
  name: "Record", // custom static name
  primaryKey: "_id",
};
```

As above showed , `name` property appeared twice, although this won't caused unexpected behavior due to how js language works.

This pull request just checked if user defined custom static name, it will remove the default class name

PR #6290, opened 2023-11-30
Added `skipIf` on missingServer to non-local tests

kraenhansen:
## What, How & Why?

We're already using `.skipIf(environment.missingServer` a couple of places to be able to run local tests that doesn't require the server. This add a couple more to make the tests pass when the server is stopped and the tests are ran with `MOCHA_REMOTE_CONTEXT=missingServer`.


elle-j (approved):
Good update 👍

takameyer (approved):
LGTM

PR #6291, opened 2023-11-30
Fix race condition with useObject

bimusiek:
## What, How & Why?
We had some race condition that re-render was triggered quicker than `updateRef.current` was set to `true`. 
So basically, re-render was happening but the `useObject` was returning old object, so the children component did not see the change and was not re-rendering there. 

We have just tested this fix at `@hernas/realm-js@0.6.5` and this fixed the object not being updated.

## ☑️ ToDos
* [ ] 🚦 Tests

This was caught on:
- React Native with Hermes

### DIFF
--- a/packages/realm-react/src/cachedObject.ts
+++ b/packages/realm-react/src/cachedObject.ts
@@ -111,6 +111,7 @@ export function createCachedObject({ object, realm, updateCallback, updatedRef }
 
   const cachedObjectResult = new Proxy(object, cachedObjectHandler);
   const listenerCallback: Realm.ObjectChangeCallback<any> = (obj, changes) => {
+    updatedRef.current = true;
     if (changes.deleted) {
       updateCallback();
     } else if (changes.changedProperties.length > 0) {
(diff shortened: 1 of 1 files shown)


bartoszhernas:
Race condition deep down, maybe it helps with writing tests?

On Electron the issue was not visible. On ReactNative it was visible only when the same hook that used `useObject` has been used together with other hook that was triggering rerender, and there were multiple  `useObject` hooks for the same object registered as well. 

The same hook, fetching the same object was used in 3 places, and in first place it was working fine, in two other places the hook would rerender with new properties, but the object would have the old reference, so the useMemo that used the object didn't recalculate anything.

Looking at code for cachedObject we saw that `updateCallback();` would call `forceRerender()` function from `useObject`, but at the end of the function a new `Proxy` would be returned only if `updateRef.current` would be true. Our guess turned out correct, and making sure `updateRef` is changed to true first, fixed the issue of having stale (old reference) object and having new Proxy returned, hence having all useMemo recalculated for us.

takameyer:
This change makes a lot of sense.  The ref should definitely be updated before the update method is invoked.  Good catch!  I'll test this locally and get this merged.

PR #6294, opened 2023-12-01
Double Login Tests

takameyer:
## What, How & Why?
* I wanted to ensure the double login tests would pass for realm-react.
* Everything works as they should and now we have a basic UserProvider test.
* Still can't reproduce the original issue that we are seeing in React Native.
* Making a PR as they expand the tests a bit further and get a start to the UserProvider tests.

## To the reviewer
* I do not need a deep review here, just wanted to make sure this code lives somewhere while I continue to debug a related issue.
* Had hoped these tests would help reproduce the issue outside of React Native, but sadly they pass.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-react/src/__tests__/UserProvider.test.tsx
+++ b/packages/realm-react/src/__tests__/UserProvider.test.tsx
@@ -29,6 +29,8 @@ import { App } from "realm";
 const testEmail = "test@email.com";
 const testPassword = "password";
 
+// This is used to determine the initial state for user registrations.
+// Ensures that subsequent tests do not register the test user again.
 let userRegistered = false;
 
 const Login = () => {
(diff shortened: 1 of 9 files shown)


packages/realm-react/src/__tests__/UserProvider.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { useEmailPasswordAuth } from "../useEmailPasswordAuth";
+import { baseUrl, importApp } from "./helpers";
+import { AuthOperationName } from "../types";
+import { UserProvider, useUser } from "../UserProvider";
+import { App } from "realm";
+
+const testEmail = "test@email.com";
+const testPassword = "password";
+
+let userRegistered = false;
```
elle-j, line 32: Even though the variable is updated to `true` later on, it doesn't look like the variable is (currently) needed (only passed as the initial value of `isUserRegistered`).

elle-j (approved):
Nice getting these underway 🥳 🙏

packages/realm-react/src/__tests__/UserProvider.test.tsx
takameyer: I wanted to commit a comment, but had moved to other things.  This ensures that subsequent tests do not have to register.  Only the first that renders this will.

PR #6297, opened 2023-12-04
Refactor notification / listeners tests

kraenhansen:
## What, How & Why?

This is a refactor of our listener / notification tests, adding missing coverage for some of our collection types and (hopefully) increasing the readability of the tests in the process.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


integration-tests/tests/src/typings.d.ts
```diff
@@ -192,7 +192,9 @@ declare namespace Chai {
 declare function setImmediate(cb: () => void): void;
 
 /** Rough typing of setTimeout to avoid type errors */
-declare function setTimeout(cb: (args: any[]) => void, timeout: number): any;
+declare type Timer = number;
+declare function setTimeout(cb: (args: any[]) => void, timeout: number): Timer;
+declare function clearTimeout(timer: Timer): void;
```
kraenhansen, line 197: I needed to expose the `clearTimeout` to use it in the "listener stub" util.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+          this.realm.create("Person", { name: "Alice" });
+        });
+
+        await handle;
+      });
+    });
+
+    describe("beforenotify", () => {
+      // Skipping on React Native because the callback is called one too many times
+      it.skipIf(environment.reactNative, "calls listener", async function (this: RealmContext) {
```
kraenhansen, line 176: We were already skipping this test on React Native:

https://github.com/realm/realm-js/blob/ca6abdd242a838e3d9c70a99e3a0fecf110e86e0/integration-tests/tests/src/tests/listeners.ts#L44-L45

I still don't know why this is happening and it would make sense for us to look into it at some point. I've created https://github.com/realm/realm-js/issues/6298 to track this.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+/**
+ * This suite aims to test observability of Realm, Realm.Object and all collection types.
```
takameyer, line 20: Bit nit picky, but "observable" to is a term from RxJS and means something different when I read that.  Without reading this PR, I'm not sure I would deduce that event handlers were being tested here.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+
+function expectObservableMethods(observable: Observable) {
+  expect(observable.addListener).to.be.a("function");
+  expect(observable.removeListener).to.be.a("function");
+  expect(observable.removeAllListeners).to.be.a("function");
+}
+
+function expectRealmNotifications(
+  realm: Realm,
+  eventName: "change" | "schema" | "beforenotify",
```
takameyer, line 43: I was trying to figure out what the difference between "change" and "beforenotify" is and have realized that this isn't [documented](https://github.com/realm/realm-js/blob/main/packages/realm/src/Realm.ts#L214).  Perhaps we could add this while we are at it.

takameyer (approved):
Amazing stuff here.  Really easy to read and follow.  Couple comments, but can be disregarded at will.

integration-tests/tests/src/tests/observable.ts
kraenhansen: I choose "observable" as a reference to the very general ["observer pattern"](https://en.wikipedia.org/wiki/Observer_pattern) in software design: The ability to register listeners (i.e. observers) of an object and have notifications trigger on change, which is probably also what RxJS and lots of other reactive / observable libraries are referencing.

integration-tests/tests/src/tests/observable.ts
kraenhansen: Yeah - it's not ideal. I'll leave that for a follow up PR. To be completely honest, I don't know the intended use-case for it: Perhaps we should actually just break the API and not expose this.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+) {
+  const handle = createPromiseHandle();
+  realm.addListener(
+    eventName,
+    createListenerStub(
+      handle,
+      ...expectedChangeSets.map(
+        (expectedChanges, c) => (realm: Realm, name: string, schema?: Realm.CanonicalObjectSchema[]) => {
+          expect(name).equals(eventName, `Realm change event #${c} name didn't match`);
+          expect(schema).deep.equals(expectedChanges.schema, `Realm change event #${c} schema didn't match`);
```
elle-j, line 54: The naming of `c` initially through me off a bit, thinking it may be misused as I didn't associate the name to an index. But what do you think of moving the part within `map(..)` to another function to speed up readability? Perhaps something like
```js
...expectedChangeSets.map(createRealmListenerCallback)
```
If so, then it could also apply to the other `expectXNotifications` functions.

integration-tests/tests/src/tests/observable.ts
kneth: Does it happens both on Android and iOS? I mean, the event loop is different and I wonder if it is related to core's scheduler or our code?

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+          deletions: [],
+          insertions: [0],
+          oldModifications: [],
+          newModifications: [],
+        },
+      ]);
+
+      this.realm.write(() => {
+        // collection["bob"].name = "Bobby";
+        // TODO: It seems we cannot trigger a notification when a property value changes.
```
kneth, line 680: Can we open an issue about it?

kneth (approved):

integration-tests/tests/src/tests/observable.ts
elle-j: P.S. You could use [RealmEventName](https://github.com/realm/realm-js/blob/3fa8b6baf88d385e7f62546ea7b5b677eda83983/packages/realm/src/Realm.ts#L214) here, but it doesn't seem to be exported in `index.ts`. (Noticed that our [enum](https://github.com/realm/realm-js/blob/3fa8b6baf88d385e7f62546ea7b5b677eda83983/packages/realm/src/RealmListeners.ts#L21-L25) is also not in `index.ts`. So if we do expose it, we could also use those in place of the hardcoded ones.)

If I'm not mistaken, I think one of the use cases is to react to potential runtime schema change initiated on the backend? We currently simulate this in our tests via `realm._updateSchema()`.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+        expect(changes).deep.equals(expectedChanges, `Changeset #${c} didn't match`);
+      }),
+    ),
+  );
+  return handle;
+}
+
+describe("Observable", () => {
+  // describe("App", () => {});
+
```
elle-j, line 103:
```suggestion
```

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+    describe("change", () => {
+      it("calls listener", async function (this: RealmContext) {
+        const completion = expectRealmNotifications(this.realm, "change", [{}]);
+
+        this.realm.write(() => {
+          this.realm.create("Person", { name: "Alice" });
+        });
+
+        await completion;
```
elle-j, line 128: Nice 😎

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+        });
+
+        this.realm.addListener("change", listener);
+
+        this.realm.write(() => {
+          this.realm.create("Person", { name: "Alice" });
+        });
+
+        await handle;
+      });
```
elle-j, line 150: I like the pattern taken with the `expectXNotifications` functions. Could something similar be applied to the removal tests for expecting the removal?

integration-tests/tests/src/tests/observable.ts
kraenhansen: Either that or we should clarify what the intended behaviour is. I'll follow up on this 👍

integration-tests/tests/src/tests/observable.ts
kraenhansen: Yes, it happens on Android and iOS.

integration-tests/tests/src/utils/sequence.ts
```diff
@@ -0,0 +1,35 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+/**
+ * @returns A function which will delegate calls to the callbacks in sequence or throw if called too many times.
+ */
+export function sequence<This, Args extends unknown[], R>(...callbacks: Array<(...args: Args) => R>) {
+  let i = 0;
+  return function (this: This, ...args: Args) {
+    if (i >= callbacks.length) {
+      throw new Error(`Sequence was called more than ${callbacks.length} times (call #${i})`);
```
elle-j, line 26: Nit: Maybe either `Sequence was called ${callbacks.length} or more times (call #${i})` or `Sequence was called more than ${callbacks.length} times (call #${i + 1})` is more correct?

integration-tests/tests/src/tests/observable.ts
```diff
@@ -0,0 +1,848 @@
...
+
+import { expect } from "chai";
+
+import Realm, { CollectionChangeSet, DictionaryChangeSet, ObjectChangeSet } from "realm";
+
+import { openRealmBefore, openRealmBeforeEach } from "../hooks";
+import { createListenerStub } from "../utils/listener-stub";
+import { createPromiseHandle } from "../utils/promise-handle";
+
+type Observable = Realm | Realm.Object<any> | Realm.Results<any> | Realm.List<any> | Realm.Dictionary | Realm.Set<any>;
```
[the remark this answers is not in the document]
kraenhansen: I didn't think of those, but yes - that would be great!

integration-tests/tests/src/tests/observable.ts
kraenhansen: I tend do gravitate towards inline arrow functions for simple mappings / transformations like these, especially if they're not reused.

integration-tests/tests/src/tests/observable.ts
kraenhansen: I created https://github.com/realm/realm-js/pull/6300.

integration-tests/tests/src/tests/observable.ts
kraenhansen: I agree that the `await expectXNotifications` reads really good, but I hadn't found a good abstraction for this.
I gave it another go: https://github.com/realm/realm-js/pull/6301

PR #6301, opened 2023-12-05
Introduced `expectListenerRemoval`

kraenhansen:
## What, How & Why?

As per https://github.com/realm/realm-js/pull/6297#discussion_r1415331002, I've introduced a common `expectListenerRemoval` to help test removal of listeners.


elle-j (approved):
👌

PR #6308, opened 2023-12-07
Passing actions when expecting notifications in tests

kraenhansen:
## What, How & Why?

I know I'm going in circles on this one, but while implementing the key-path filtering tests, I noticed a pattern that I think simplifies writing these async tests even further, by allowing us to interleave actions (arrow function returning `void` and taking no arguments) with the expected changeset objects, when expecting notifications.

### DIFF
--- a/integration-tests/tests/src/tests/observable.ts
+++ b/integration-tests/tests/src/tests/observable.ts
@@ -39,61 +39,127 @@ function expectObservableMethods(observable: Observable) {
   expect(observable.removeAllListeners).to.be.a("function");
 }
 
+type Action = () => void;
+type ChangeAndActions<ChangeSet extends object> = { expectedChange: ChangeSet; actions: Action[] };
+
+/**
+ * Transforms an array of changeset objects and actions into an array of objects for each changeset with all actions which immediately proceed it.
+ */
+function inlineActions<ChangeSet extends object>(changesAndActions: (ChangeSet | Action)[]) {
+  const initialActions: Action[] = [];
+  const changes: ChangeAndActions<ChangeSet>[] = [];
+
+  for (const changeOrAction of changesAndActions) {
+    if (typeof changeOrAction === "function") {
+      if (changes.length > 0) {
+        const current = changes[changes.length - 1];
+        current.actions.push(changeOrAction);
+      } else {
+        initialActions.push(changeOrAction);
+      }
+    } else {
+      changes.push({ expectedChange: changeOrAction, actions: [] });
+    }
+  }
+
+  return { initialActions, changes };
+}
+
+/**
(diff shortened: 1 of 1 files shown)


integration-tests/tests/src/tests/observable.ts
```diff
@@ -39,61 +39,127 @@ function expectObservableMethods(observable: Observable) {
...
 function expectRealmNotifications(
   realm: Realm,
   eventName: RealmEventName,
-  expectedChangeSets: { schema?: Realm.CanonicalObjectSchema[] }[],
+  changesAndActions: (Action | RealmChangeSet)[],
 ) {
   const handle = createPromiseHandle();
+
+  const { initialActions, changes } = inlineActions(changesAndActions);
+  performActions(initialActions);
```
elle-j, line 90: If these are always performed in tandem, do you see a point in moving the responsibility of calling `performActions()` out of the `expectX..` functions and perhaps to `inlineActions()`? (Also due to `initialActions` only being used as the argument.)

integration-tests/tests/src/tests/observable.ts
```diff
@@ -303,16 +372,15 @@ describe("Observable", () => {
...
-      const completion = expectObjectNotifications(this.object, [
+      await expectObjectNotifications(this.object, [
         { deleted: false, changedProperties: [] },
+        () => {
+          this.realm.write(() => {
+            this.object.name = "Bob";
+          });
+        },
         { deleted: false, changedProperties: ["name"] },
       ]);
```
elle-j, line 383: Really neat idea with being able to see the actions and their effects sequentially like this 🙂

integration-tests/tests/src/tests/observable.ts
```diff
@@ -39,61 +39,127 @@ function expectObservableMethods(observable: Observable) {
   expect(observable.removeAllListeners).to.be.a("function");
 }
 
+type Action = () => void;
+type ChangeAndActions<ChangeSet extends object> = { expectedChange: ChangeSet; actions: Action[] };
```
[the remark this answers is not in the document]
kraenhansen: It doesn't matter too much what it is, as long as it's not a `function` since that's what I use to distinguish actions from changesets as `typeof fn === "function"` for functions and `typeof changeset === "obejct"` for the changeset options 👍

integration-tests/tests/src/tests/observable.ts
kraenhansen: I tend to separate functions that transform data (the inlining here) from functions with side-effects (the expectXNotifications functions).

elle-j: Sounds good 🙂

PR #6309, opened 2023-12-08
Created binding.gyp

### LINKED ISSUE #3828: Prebuilt binaries aren't fetched when packaging an Electron app
## Goals

Using electron-builder to install native dependencies for an Electron app using Realm JS for Windows or Linux on a macOS machine.

## Expected Results

I expect Electron Builder to recognise Realm JS as having a native module which needs to be installed using `prebuild-install`.

## Actual Results

When building for Windows on a macOS machine, end users of the electron app experience the following error:

```
Uncaught Error: \\?\C:\Users\bimusiek\AppData\Local\Temp\3da4c319-5a47-4f25-9b04-8a36075b8f09.tmp.node is not a valid Win32 application.
```
CleanShot 2021-07-05 at 16 18 56@2x

This is because the realm.node binary shipped with the Windows build is actually the binary for macOS which never got replaced when the app was being packaged.

A temporary fix we needed in Realm Studio was [this line](https://github.com/realm/realm-studio/blob/channel/major-11/Jenkinsfile#L227-L228):
```
touch node_modules/realm/binding.gyp
```

My investigations this far concludes: Electron Builder (the `electron-builder` package on NPM) looks for (at least) the `binding.gyp` file when determining if a dependency of an electron app has a native module or not. As a consequence Electron builder won't fetch the correct prebuilds when building the app for Windows nor Linux on a macOS machine. The `binding.gyp` file was removed as we migrated Realm JS to `cmake-js`.

We could add the empty file to our package, but since that feels a bit like a workaround, I would love to learn of alternative ways to tell electron-builder to consider Realm JS as having native modules which needs to be fetched when packaging up the app.

This could be related to https://github.com/electron/electron-rebuild/issues/268.

## Steps to Reproduce

* Initialize an NPM package with electron@11.2.1, realm@10.5.0 and electron-builder@22.10.5.
* Try adding native dependencies (`npx electron-builder install-app-deps`)
* Notice that realm is not included.
* Add the `binding.gyp` file to Realm JS (`touch node_modules/realm/binding.gyp`)
* Try adding native dependencies again and notice that realm appears.

- Realm JS SDK Version: 10.5.0
- Node or React Native: Electron
- Client OS & Version: Tested on MacOS

kraenhansen:
## What, How & Why?

This closes #3828

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * When mapTo is used on a property of type List, an error like `Property 'test_list' does not exist on 'Task' objects` occurs when trying to access the property. ([#6268](https://github.com/realm/realm-js/issues/6268), since v12.0.0)
 * Fixed bug where apps running under JavaScriptCore on Android will terminate with the error message `No identifiers allowed directly after numeric literal`. ([#6194](https://github.com/realm/realm-js/issues/6194), since v12.2.0)
 * When an object had an embedded object as one of its properties, updating that property to `null` or `undefined` did not update the property in the database. ([#6280](https://github.com/realm/realm-js/issues/6280), since v12.0.0)
+* Fixed download of platform + arch specific prebuilt binaries when building an Electron app using `electron-builder`. ([#3828](https://github.com/realm/realm-js/issues/3828))
 
 ### Compatibility
 * React Native >= v0.71.4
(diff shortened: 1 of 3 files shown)


elle-j (approved):
Great find!

PR #6310, opened 2023-12-08
Key path filtered notifications in `Object` and `Collection`

### LINKED ISSUE #6285: Key-path filtering when adding listeners to `Collection` and `Object`

kraenhansen:
## What, How & Why?

This closes #6285 by exposing an optional third argument to the `addListener` methods.
This also updates Core to v13.24.0 to pull in the new APIs that got added to the spec.yml and depends on [one additional change](https://github.com/realm/realm-core/pull/7193) that hasn't been merged and released yet.

NOTE: Because of 30d99c0c7ca0ebc41d4afa6735b22564550ee7d4 adding a listener twice, will start throwing instead of being a no-op. We had tests for this, but it wasn't explicitly documented, so technically not a breaking change to the documented API, but it still might result in runtime errors in some consuming code-bases.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (on `Object`, `Results` and `Dictionary` - missing tests on `List` and `Set`, but they hit the same code)


integration-tests/tests/src/tests/observable.ts
```diff
@@ -404,25 +435,120 @@ describe("Observable", () => {
         },
       });
     });
+
+    describe("key-path filtered", () => {
+      it("calls listener only on relevant changes", async function (this: RealmObjectContext<Person>) {
+        await expectObjectNotifications(
+          this.object,
+          [
+            EMPTY_OBJECT_CHANGESET,
```
takameyer, line 444: Is this just to handle the initial empty event notifier?

integration-tests/tests/src/tests/observable.ts
```diff
@@ -404,25 +435,120 @@ describe("Observable", () => {
...
+              this.realm.write(() => {
+                this.object.name = "Alex";
+                this.object.age = 42;
+              });
+            },
+            {
+              deleted: false,
+              changedProperties: ["name"],
+            },
+            () => {
```
takameyer, line 456: I wonder if there's a way we could make it clearer in the utils that this action will not trigger?  Perhaps a sort of `no-op` result given.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -490,6 +604,127 @@ describe("Observable", () => {
         },
       });
     });
+
+    describe("key-path filtered", () => {
+      it("calls listener only on relevant changes", async function (this: RealmObjectContext<Person>) {
+        const collection = this.realm.objects<Person>("Person").filtered("name = $0 OR age = 42", "Alice");
+        await expectCollectionNotifications(
```
takameyer, line 611: Perhaps named params would be a bit clearer.  I find myself having to jump back and forth to the implementation to follow what all these mean.

takameyer: Something like:
```
expectCollectionNotifications(
collection,
[
 {result: EMPTY_COLLECTION_CHANGESET},
 {action: () => /*..*/, result: {deletions:[], /*etc*/ }}
 //...
],
{
 keypaths: ["name"]
}
)
```

packages/realm/package.json
```diff
@@ -31,7 +31,7 @@
     "url": "https://www.mongodb.com/docs/realm/"
   },
   "config": {
-    "anonymizedBundleId": "tkgif/+3l1e9wStGJp2TOngAK3UcQ2u7OM8ZYJU5JYo="
+    "anonymizedBundleId": "1RmJBlqbKuzyRiPm4AsdIIxe8xlRUntGcGFEwUnUh6A="
```
takameyer, line 34: I added a comment about this elsewhere, but can't find it. We should do a follow up PR to skip updating this field of `REALM_DISABLE_ANALYTICS` is set.

packages/realm/src/Collection.ts
```diff
@@ -37,14 +37,14 @@ export abstract class Collection<
 > implements Iterable<T>
 {
   /** @internal */
-  private listeners: Listeners<ChangeCallbackType, binding.NotificationToken>;
+  private listeners: Listeners<ChangeCallbackType, binding.NotificationToken, [string[] | undefined]>;
```
takameyer, line 40: Should we extract `[string[] | undefined]` into its own type?

packages/realm/src/Collection.ts
```diff
@@ -128,9 +128,9 @@ export abstract class Collection<
    * @note Adding the listener is an asynchronous operation, so the callback is invoked the first time to notify the caller when the listener has been added.
    * Thus, when the callback is invoked the first time it will contain empty arrays for each property in the `changes` object.
```
takameyer, line 129: We should add a `@param` to the jsdoc for `keyPaths` on all listeners that have implemented it.

takameyer (commented):
I have some suggestions.  Otherwise, looks good 👍🏼

integration-tests/tests/src/tests/observable.ts
takameyer: Perhaps if we are expecting this to initialize with `EMPTY_COLLECTION_CHANGESET`, then we could just assert this within the utility, as it doesn't provide much value into the test itself to know this is being asserted.

packages/realm/src/Collection.ts
kraenhansen: Yep - I must have forgotten that 😊

packages/realm/src/Object.ts
```diff
@@ -505,12 +505,12 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
    * @note Adding the listener is an asynchronous operation, so the callback is invoked the first time to notify the caller when the listener has been added.
    * Thus, when the callback is invoked the first time it will contain empty array for `changes.changedProperties`.
    */
-  addListener(callback: ObjectChangeCallback<T>): void {
+  addListener(callback: ObjectChangeCallback<T>, keyPaths?: string[]): void {
```
elle-j, line 508: Shouldn't this be `keyPaths?: string | string[]`? If so, also check other potential updates such as `ObjectListeners.ts`. Could also add some tests using the `string` alternative (in addition to the currently used `string[]` tests).
```suggestion
  addListener(callback: ObjectChangeCallback<T>, keyPaths?: string | string[]): void {
```

elle-j (commented):
Awesome getting this in so it can be tried out. Left some ideas and suggestions 👍

packages/realm/package.json
kraenhansen: I included that in #6313 👍 I'll get this change reverted ...

kraenhansen:
```suggestion
    "anonymizedBundleId": "tkgif/+3l1e9wStGJp2TOngAK3UcQ2u7OM8ZYJU5JYo="
```

packages/realm/src/Object.ts
kraenhansen: Oh yeah 🤦 I forgot all about this agian.

packages/realm/src/Collection.ts
kraenhansen: I'm fine with leaving it as-is in this particular case: I wouldn't say we need to export it, it's not duplicated all-over the code-base and don't expect it to change.

CHANGELOG.md
```diff
@@ -1,26 +1,29 @@
...
 
+### Breaking changes
+* Adding the same callback function as listener on a `Collection` or `Object` used to be undefined behavior, but now throws, which result in runtime errors that can be resolved by ensuring the a callback is only added once. ([#6310](https://github.com/realm/realm-js/pull/6310))
+
 ### Deprecations
 * None
 
 ### Enhancements
 * Exporting a `RealmEventName` type. ([#6300](https://github.com/realm/realm-js/pull/6300))
+* Added an optional third `keyPaths` argument to the `addListener` methods of `Collection` and `Object`. Use this to indicate a lower bound on the changes relevant for the listener. This is a lower bound, since if multiple listeners are added (each with their own "key paths") the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the object or collection. In other words: A listener might fire more than the key-paths specify, if other listeners with different key-paths are present. ([#6285](https://github.com/realm/realm-js/issues/6285))
```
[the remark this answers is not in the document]
kraenhansen: I added a small example 👍

integration-tests/tests/src/tests/observable.ts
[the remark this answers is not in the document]
kraenhansen: You're right. I should probably add these to make the feature even more tested.
With some of your suggestions, I do think we approach the territory of what I'd expect to be tested by Core, but it doesn't hurt to add them - especially now that writing and maintaining a listener test isn't quite as painful 👍

integration-tests/tests/src/tests/observable.ts
kraenhansen: Having to explicitly pass that, I fear we'll instead forget to pass it 🤔

integration-tests/tests/src/tests/observable.ts
kraenhansen: The `expectXNotifications` functions are already pretty magical, I fear including this in those will make them even harder to reason about as they'll start making assumptions on the behaviour of the SDK too: I like these empty changesets are explicitly listed (even if its more verbose).

kraenhansen: >  I find myself having to jump back and forth to the implementation to follow what all these mean.

Between the implementation of the `expectXNotifications` and the tests? Isn't it pretty clear once you've understand that they just assert changeset state or execute the functions? Having the `action` as part of an object with `result`, I fear someone could read it as the "action" that needs to be taken after a the `result` has been asserted. In other words, is this a pre-action or a post-action? Technically the `action` needs to fire as part of the function that asserts the previous `result` (if that makes sense).

takameyer: Most likely since I'm in the context of Github PR view, it's a little less intuitive to discern what the intention of the parameters to the `expectNotifications` functions are and how they are executed.  
>  is this a pre-action or a post-action? 

I found myself questioning that in its current form. They are quite magical, so I would be happy if we provided guidance through named parameters or through documentation.

integration-tests/tests/src/tests/observable.ts
takameyer: Then documentation would suffice.

integration-tests/tests/src/tests/observable.ts
kraenhansen: I updated the documentation of a shared utility function. I played around with named parameters, but couldn't find short names that really captured the intent without feeling ambiguous to me. I think having a list of "changes and actions" which are either asserted or executed in an ordered sequence is pretty easy to reason about, with that we don't need to explain if the `action` goes before or after the change / `result` in the same object.

integration-tests/tests/src/tests/observable.ts
kraenhansen: I updated the documentation of a shared `expectNotifications` utility function now.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -490,6 +604,127 @@ describe("Observable", () => {
...
+              this.realm.write(() => {
+                this.object.name = "Alice";
+              });
+            },
+            {
+              deletions: [],
+              insertions: [0],
+              newModifications: [],
+              oldModifications: [],
+            },
```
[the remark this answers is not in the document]
kraenhansen: I separated them out 👍

CHANGELOG.md
```diff
@@ -1,26 +1,35 @@
...
+    console.log("A car location changed");
+  }, ["location"]);
+  ```
 
 ### Fixed
 * When mapTo is used on a property of type List, an error like `Property 'test_list' does not exist on 'Task' objects` occurs when trying to access the property. ([#6268](https://github.com/realm/realm-js/issues/6268), since v12.0.0)
 * Fixed bug where apps running under JavaScriptCore on Android will terminate with the error message `No identifiers allowed directly after numeric literal`. ([#6194](https://github.com/realm/realm-js/issues/6194), since v12.2.0)
 * When an object had an embedded object as one of its properties, updating that property to `null` or `undefined` did not update the property in the database. ([#6280](https://github.com/realm/realm-js/issues/6280), since v12.0.0)
 * Fixed download of platform + arch specific prebuilt binaries when building an Electron app using `electron-builder`. ([#3828](https://github.com/realm/realm-js/issues/3828))
+* Notification listeners on a `Dictionary` would only fire when the dictionary itself changed (via inserts or deletions) but not when changes were made to the underlying objects. ([#6310](https://github.com/realm/realm-js/pull/6310))
```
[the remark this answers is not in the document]
kraenhansen:
```suggestion
* Notification listeners on a `Dictionary` would only fire when the dictionary itself changed (via inserts or deletions) but not when changes were made to the underlying objects. ([#6310](https://github.com/realm/realm-js/pull/6310), since v12.0.0)
```

packages/realm/src/Object.ts
```diff
@@ -502,15 +503,19 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
    *      console.log(` ${prop}`);
    *   });
    * })
+   * @example
+   * wine.addListener((obj, changes) => {
+   *  console.log("The wine got deleted or its brand might have changed");
+   * }, ["brand"])
```
elle-j, line 509: Regarding the `@example`'s log message, I believe it's technically also for when the listener is added. (Perhaps not necessary for the `@example` though.)

elle-j (approved):
🚀

integration-tests/tests/src/tests/observable.ts
```diff
@@ -206,16 +246,22 @@ describe("Observable", () => {
...
+      });
 
-        await completion;
+      // Skipping, because although the API would be more uniform, there is no need to throw when adding the same Realm listener twice
+      it.skip("throws on listener reuse", function (this: RealmContext) {
+        this.realm.addListener("change", noop);
+        expect(() => {
+          this.realm.addListener("change", noop);
+        }).throws("Remove callback before adding it again");
       });
```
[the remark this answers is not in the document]
kraenhansen: Makes sense :+1: I've added tests to verify that they're indeed not throwing.

kneth (approved):

integration-tests/tests/src/tests/observable.ts
```diff
@@ -404,25 +435,120 @@ describe("Observable", () => {
...
+              // Updating the age to 42 will ensure the object doesn't leave the results
+              this.realm.write(() => {
+                this.object.name = "Alex";
+                this.object.age = 42;
+              });
+            },
+            {
+              deleted: false,
+              changedProperties: ["name"],
+            },
```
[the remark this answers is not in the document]
kraenhansen: I moved this up as the second argument tot he "expectXNotification" functions.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -490,6 +604,127 @@ describe("Observable", () => {
...
+
+    describe("key-path filtered", () => {
+      it("calls listener only on relevant changes", async function (this: RealmObjectContext<Person>) {
+        const collection = this.realm.objects<Person>("Person").filtered("name = $0 OR age = 42", "Alice");
+        await expectCollectionNotifications(
+          collection,
+          [
+            EMPTY_COLLECTION_CHANGESET,
+            () => {
+              // Updating the age to 42 will ensure the object doesn't leave the results
```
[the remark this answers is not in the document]
kraenhansen: I'll leave this as is, since this was actually what I wanted to convey. The object stays in `collection`, since it match on both the name and age.

packages/realm/src/Object.ts
kraenhansen: True - but for the sake of simplicity, I suggest leaving it as-is.

takameyer (approved):

integration-tests/tests/src/tests/observable.ts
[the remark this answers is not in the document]
elle-j: Ah I see what you mean now 👍

PR #6312, opened 2023-12-11
Fix unit tests

kraenhansen:
## What, How & Why?

This fixes the failing unit tests for the `realm` package by upgrading the Node.js version to 20 and updating `tsx` CLI to the latest version.

As a part of this, I propose moving the running of the unit tests of the `@realm/bindgen` package, into the `realm-core` repo: https://github.com/realm/realm-core/pull/7199 (now merged), since this seemed easier than awaiting a Core release before merging this and it seems to be the right place to run them anyway.

### DIFF
--- a/.github/workflows/package-unit-tests.yml
+++ b/.github/workflows/package-unit-tests.yml
@@ -54,7 +54,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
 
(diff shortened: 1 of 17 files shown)


kraenhansen:
The of the "Testing Realm Web" is actually because the GHA workflow hasn't yet been migrated from Lerna to NPM workspaces and I'll just leave it as is for now.

elle-j (approved):

PR #6313, opened 2023-12-11
Read core version from submodule

kraenhansen:
## What, How & Why?

With this change we no longer have to maintain a `dependencies.list` file, since the Core version is instead read from the core submodule. I also added a fix to avoid writing the anonymized bundle id if analytics is disabled.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


takameyer (approved):
Nice :)

packages/realm/scripts/submit-analytics.mjs
```diff
@@ -330,6 +328,8 @@ async function submitAnalytics() {
     return;
   }
 
+  saveBundleId(data["Anonymized Bundle Id"]);
```
kneth, line 331: I think we need to save the bundle id before line 326 as it is needed at runtime too.

kneth (changes requested):

packages/realm/scripts/submit-analytics.mjs
kraenhansen: If this line doesn't execute on the end-users machine, it will fallback to the value already stored in the `package.json`: https://github.com/realm/realm-js/blob/main/packages/realm/package.json#L34

kneth (approved):

PR #6315, opened 2023-12-12
Adding `inspect.custom` symbol on RealmObject and Collection on Node

### LINKED ISSUE #2758: Improve the output from console.log of objects and collections on Node.js
## Goals
Using console.log during debugging is a fairly common practice and the output of calling this on an object or collection from a Realm should be informative.

## Expected Results
I would expect that logging a `Collection` of objects (such as `Results`) would give information on the type of items in the collection, the size of the collection and perhaps a preview of the first couple of elements in the collection. As a reference, calling `console.log` on an array and objects yields respectively:

```javascript
console.log(a);
// [ { name: 'Alice' }, { name: 'Bob' }, { name: 'Charlie' } ]
```

```javascript
console.log(a[0])
// { name: 'Alice' }
```

## Actual Results, Steps to Reproduce & Code Sample
```javascript
const realm = new Realm({ schema: [{ name: "Person", properties: { name: "string" } }] });
realm.write(() =>{ realm.create('Person', { name: "Alice" }); });
const persons = realm.objects("Person");
console.log(persons)
// Results {}
console.log(persons[0])
// Person { [Symbol(_external)]: [External] }
```

In addition to this, in the Node.js REPL, when executing an expression the REPL prints its value, for a collection, this prints:

```
Proxy [ Results {}, {} ]
```

All of the output above have very limited value to a user while debugging their code.

NOTE: https://nodejs.org/dist/latest-v12.x/docs/api/util.html#util_custom_inspection_functions_on_objects might come in handy when solving this.

* Realm JS SDK Version: 5.0.0
* Node or React Native: Node.js
* Client OS & Version: Tested on MacOS
* Which debugger for React Native: None

kraenhansen:
## What, How & Why?

I finally got tired of the lack of property values on objects as they're being logged and implemented a custom inspect function for `Realm.Object` and `Realm.Collection`.

This is the output of logging first a `Realm.Object` with a string, list, set and dictionary properties.

Screenshot 2023-12-12 at 21 17 31

This closes #2758.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/src/platform/node/custom-inspect.ts
+++ b/packages/realm/src/platform/node/custom-inspect.ts
@@ -1,0 +1,36 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2023 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+import { InspectOptions, inspect } from "node:util";
+
+import { Collection, RealmObject } from "../../internal";
+
+type CustomInspectFunction<T> = (this: T, depth: number, inspectOptions: InspectOptions) => void;
+
+function injectInspect<T extends object>(constructor: { prototype: T }, customInspect: CustomInspectFunction<T>) {
+  Object.assign(constructor.prototype, { [inspect.custom]: customInspect });
+}
+
+injectInspect(RealmObject, function (this, depth: number, inspectOptions: InspectOptions) {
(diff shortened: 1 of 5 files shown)


kneth (approved):
That's a goodie for our node.js users

takameyer (approved):

packages/realm/src/platform/node/custom-inspect.ts
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+import { InspectOptionsStylized, inspect } from "node:util";
+
+import { Collection, Dictionary, List, RealmObject, RealmSet } from "../../internal";
+import { Results } from "realm/binding";
```
elle-j, line 21: Preferable to use imports via `"realm"` instead?

elle-j (approved):
Way better DX 👏

packages/realm/src/platform/node/custom-inspect.ts
kraenhansen: Indeed - good catch, this is the wrong `Results`!

kraenhansen:
Merging now since the failing tests seem unrelated.

PR #6338, opened 2023-12-26
Update submodules on checkout

takameyer:
## What, How & Why?
* installs husky to implement git hooks
* create a post-checkout hook which will update submodules
* makes a better DX as it removes a manual step when building from source

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+git submodule update --recursive --init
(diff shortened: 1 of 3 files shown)


.husky/post-checkout
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
kraenhansen, line 2: I think this line deserves a comment.

takameyer: It was generated from husky, so I haven't really given any thought into what it actually does.

takameyer:
Decided to leave it as it is.  Was a valuable exercise in Cmake.

PR #6340, opened 2023-12-27
Update PR workflow to ignore unnecessary paths

elle-j:
## What, How & Why?

Adds `paths-ignore` to the pr-realm-js workflow for PRs when updating `.md` and `.png` files.

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -1,9 +1,16 @@
 name: Pull request build and test
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -4,6 +4,9 @@ on:
     branches:
       - main
   pull_request:
+    paths-ignore:
```
kraenhansen, line 7: I wonder if this would need to go into the `push` trigger above as well? 🤔

takameyer (approved):

PR #6347, opened 2023-12-29
Fix accessing `providerType` via `User.identities`

### LINKED ISSUE #6248: Realm.User.identities is returning undefined providerType

Always

### Description

When trying to access the providerType of a UserIdentity in Realm.User.identities, providerType is undefined.

https://github.com/realm/realm-js/blob/a3589c63a1d18740b4ffd7b18ca82b92a8ea33b5/packages/realm/src/app-services/User.ts#L191-L196C4

I think this could be fixed with:
realm-js/packages/realm/src/app-services/User.ts @193
```diff
- const { id, provider_type: providerType } = identity as Record<string, string>;
+ const { id, providerType } = identity as Record<string, string>;
```

because Realm.User.providerType (now deprecated) just accesses identity.providerType (not identity.provider_type)
https://github.com/realm/realm-js/blob/a3589c63a1d18740b4ffd7b18ca82b92a8ea33b5/packages/realm/src/app-services/User.ts#L141-L151C4

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

_No response_

### Version

12.3.0

### Are you using encryption?

Yes

iOS 16

### Build environment

### Cocoapods version

_No response_

elle-j:
## What, How & Why?

`providerType` was previously destructured using snake_case, but since we transform the generated types to use camelCase this caused `undefined` to always be returned.

This closes #6248 

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/sync/user.ts
+++ b/integration-tests/tests/src/tests/sync/user.ts
@@ -15,6 +15,7 @@
 // limitations under the License.
 //
 ////////////////////////////////////////////////////////////////////////////
+
 import { expect } from "chai";
 import { KJUR } from "jsrsasign";
 import Realm, { ProviderType, UserState } from "realm";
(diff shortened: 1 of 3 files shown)


integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -16,30 +16,31 @@
...
+  expect(user).to.be.an("object");
+  expect(user.accessToken).to.be.a("string");
+  expect(user.refreshToken).to.be.a("string");
+  expect(user.id).to.be.a("string");
+  expect(user.customData).to.be.an("object");
   expect(user).instanceOf(Realm.User);
+  expect(user.identities).to.be.an("array");
+  if (providerType) {
+    expect(user.identities.some((identity) => identity.providerType === providerType)).to.be.true;
+  }
```
elle-j, line 43: We could also choose to move this to a separate function to keep the `expectIsUser` focused on type checking.

kraenhansen: I'd probably inline this in the test function instead 👍

kraenhansen (approved):

integration-tests/tests/src/tests/sync/user.ts
kraenhansen: This is better IMO 👍

takameyer (approved):

PR #6348, opened 2024-01-02
`@realm/react` cleanup

kraenhansen:
## What, How & Why?

A few clean-ups of `@realm/react` as I prepare to work on #6286:
- Adding a global `setup.ts` file for the Realm specific setup before tests.
- Applying lockdown of runtime types to ensure we're not accessing runtime APIs that won't be available in the various runtimes we support.
- Sorting imports in blocks of external dependencies vs internal (relative path) imports. I used the eslint `sort-imports` rule to guide me, but I wasn't happy with the final configuration, so I didn't leave that in this PR (I might revisit this in the future).

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


packages/realm-react/jest.config.json
```diff
@@ -5,7 +5,8 @@
     "<rootDir>/src"
   ],
   "setupFilesAfterEnv": [
-    "@testing-library/jest-native/extend-expect"
+    "@testing-library/react-native/extend-expect",
+    "<rootDir>/src/__tests__/setup.ts"
```
kraenhansen, line 9: I'm adding this new setup file to handle all global setup of `Realm` so we don't need to do this in the test files themselves.

packages/realm-react/src/__tests__/UserProvider.test.tsx
```diff
@@ -15,16 +15,17 @@
 // limitations under the License.
 //
 ////////////////////////////////////////////////////////////////////////////
-import React, { useEffect, useState } from "react";
-import { AppProvider } from "../AppProvider";
-import { renderHook, waitFor } from "@testing-library/react-native";
 
+import { renderHook, waitFor } from "@testing-library/react-native";
 import { AppConfigBuilder } from "@realm/app-importer";
+import Realm, { App } from "realm";
```
kraenhansen, line 21: When I enabled `flags.THROW_ON_GLOBAL_REALM`, this file throw.

packages/realm-react/src/runtime.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+declare type Console = {
+  log(...args: unknown[]): void;
+  error(...args: unknown[]): void;
+  warn(...args: unknown[]): void;
+};
+
+declare const console: Console;
+
+declare function setImmediate(cb: (...args: unknown[]) => void);
```
kraenhansen, line 27: I suggest adding this to declare the minimal runtime APIs that we allow the package to access, ensuring we implement against APIs that are supported by both Node.js and React Native.

packages/realm-react/tsconfig.json
```diff
@@ -22,4 +26,7 @@
   "include": [
     "src"
   ],
+  "exclude": [
+    "src/__tests__"
```
kraenhansen, line 30: The `src/__tests__` now have their own (more liberal) tsconfig.

packages/realm-react/src/runtime.d.ts
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+declare type Console = {
```
takameyer, line 19: What is this for?  Type error in jest?

takameyer (approved):

packages/realm-react/src/runtime.d.ts
takameyer: Does this get picked up automatically?  Does this ensure methods like `setTimeout` need to be typed here before they are used?

packages/realm-react/src/runtime.d.ts
kraenhansen: I added a comment in the file here on GitHub. Should have done that before I tagged you for review 🙂

packages/realm-react/src/runtime.d.ts
kraenhansen: Exactly. It gets picked up because the root tsconfig.json has `src` in includes and since the file isn't using ESM import / export statements, it's declaring ambient types and the `declare` statements here, become decalrations of globals.

kraenhansen:
Merging as the failing tests are unrelated to the changes I'm making.

packages/realm-react/src/runtime.d.ts
takameyer: Ahh...so it could be renamed and still work.  Interesting trick.

PR #6353, opened 2024-01-04
Build iOS from source

takameyer:
## What, How & Why?
- Created a copy of the old build script which is invoked from the Podspec
- Modified Podspec to load libraries vs xcframework 
- Created a script phase within Podspec to invoke cmake and build libraries
- Updated supported iOS version to mirror RN (from 9 to 13. this was causing many warnings)

TODO: Something went wrong when setting ccache in the int-tests RN project.  Removed this for now.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


takameyer:
Replaced by #6396 as I couldn't resolve merge conflicts.

Issue #6355, opened 2024-01-05
Trying to reset an embedded object to an invalid one updates the original with default values if ex. is caught


elle-j:
If resetting an embedded object to an object with e.g. a missing required property, the SDK expectedly throws. If this exception is caught within the write transaction by the user, the original embedded object will still be updated but with default values.

Since our call into Core's `obj.create_and_set_linked_object()` is made prior to setting the user-provided values, the values on the embedded object have been set to default values.

Example test:
```ts
// Create a valid object with a valid embedded object (`address`).
const amy = this.realm.write(() => {
  return this.realm.create(PersonWithEmbeddedSchema.name, {
    name: "Amy",
    age: 30,
    address: { street: "Broadway" },
  });
});
expect(this.realm.objects(PersonWithEmbeddedSchema.name).length).equals(1);
expect(amy.address?.street).equals("Broadway");

// Update the embedded object with an invalid one.
this.realm.write(() => {
  // It is important to catch the exception within `realm.write()` in order to test
  // that the object creation path does not modify the object (rather than being due
  // to `realm.write()` cancelling the transaction).
  expect(() => {
    const invalidEmbeddedAddress = {};
    // @ts-expect-error Testing setting invalid type.
    amy.address = invalidEmbeddedAddress;
  }).to.throw("Missing value for property 'street'");
});
const objects = this.realm.objects<IPersonWithEmbedded>(PersonWithEmbeddedSchema.name);
expect(objects.length).equals(1);
expect(objects[0].address).to.not.be.null;
// 💥 Fails: `street` will be the default value (empty string).
expect(objects[0].address?.street).equals("Broadway");
```

PR #6356, opened 2024-01-05
Roll back invalid object creations not handled by `realm.write()`

### LINKED ISSUE #2638: Creating objects while in a manual transaction, should either create an object or throw
## Goals

Calls to `realm.create()` should either create an object or throw (while in a manual transaction).

## Expected Results

If `realm.create` throws, I expect no side-effects on the database (ie. no rows with default values created).

## Actual Results

Calling `realm.create` with missing values, while in a manual transaction throws but also creates an object with default values.

```javascript
const realm = new Realm({ schema: [ { name: "SomeClass", properties: { name: 'string', age: 'int' } } ] });
realm.beginTransaction();
realm.create('SomeClass', { age: 13 })
```

☝️ throws 
```
Error: Missing value for property 'SomeClass.name'
```

Listing the objects afterwards reveals an object was actually created:

```
realm.objects('SomeClass') // 👉 Results { [0]: RealmObject { [name]: '', [age]: 0 } }
```

- Realm JS SDK Version: 3.5.0
- Node or React Native: N/A
- Client OS & Version: N/A

elle-j:
## What, How & Why?

Trying to create an invalid object (e.g. missing required properties) via `realm.create()` or a `Realm.Object` constructor would both throw and create the object (with default values for required properties), causing the object creation to not be rolled back in the following cases:
* A) During a manual transaction if the user does not explicitly catch the exception and cancel the transaction.
* B) If the user catches the exception within the callback passed to `realm.write()` and does not rethrow it.

This PR updates it so that the individual operation is automatically rolled back.

This closes #2638

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm/src/Object.ts
+++ b/packages/realm/src/Object.ts
@@ -66,7 +66,7 @@ export enum UpdateMode {
 }
 
 /** @internal */
-export type ObjCreator = () => [binding.Obj, boolean];
+export type ObjCreator = () => [binding.Obj, boolean, binding.TableRef?];
 
 type CreationContext = {
   helpers: ClassHelpers;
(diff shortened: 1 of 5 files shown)


packages/realm/src/Object.ts
```diff
@@ -66,7 +66,7 @@ export enum UpdateMode {
 }
 
 /** @internal */
-export type ObjCreator = () => [binding.Obj, boolean];
+export type ObjCreator = () => [binding.Obj, boolean, binding.TableRef?];
```
elle-j, line 69: Given that it's now 3 values, we could choose to return an object instead to clarify the return values.

kraenhansen: I agree, this is approaching (if its not already in) the territory that needs an object 👍

In this specific case, we might not need to pass the `TableRef` here at all, but we could instead pass a "rollback", "revert" or "delete" function, taking no arguments, which will call `table.removeObject(obj.key)`.

kraenhansen (approved):
I just have a small suggestion 👍 But this can merge as is if you want to.

packages/realm/src/Object.ts
takameyer: Is the `TableRef` really optional?  Maybe i'm overlooking something, but the creator seems to always return a `TableRef`.

takameyer (commented):
Great fix!

takameyer (approved):

elle-j:
As per our team discussion, to prevent the extra oplog changes with the rollback, but still make `realm.create()` only throw *or* create an object, I'll rework the PR along with [#6359](https://github.com/realm/realm-js/issues/6359) and [#6355](https://github.com/realm/realm-js/issues/6355).

Issue #6284, opened 2023-11-27
`@realm/react` create class-aware hooks


kraenhansen:
### Problem

Users want the simples way possible to query their data, but the current `useObject` and `useQuery` APIs take the `type` as an argument, which could be simplified further.

Additionally, the current 3 positional argument layout of `useQuery` means users cannot rely on eslint rules to check for missing `deps` (see https://github.com/realm/realm-js/issues/6259).

### Solution

A possible solution to both issues above, is to expose functions to generate class / type -aware hooks, which wouldn't require users to pass the `type` argument:

### One example wrapping `useQuery` ([proposed](https://github.com/realm/realm-js/pull/6272#issuecomment-1824831128) by @kraenhansen)

```typescript
export const usePersons = createQueryHook<Person>("Person");
export const useDogs = createQueryHook<Dog>("Dog");
// ... and so on, one for every class in their schema

// `createQueryHook` could even have an overload for class-based models, like some of our other APIs:
export const usePersons = createQueryHook(Person);

// And in a consuming component:
const twenty = 20;
const teenagers = usePersons(persons => persons.filtered("age < $0", twenty), [twenty]);
```

### Another example handling both `useQuery` and `useObject` ([proposed](https://github.com/realm/realm-js/pull/6272#issuecomment-1825922173) by @takameyer)

```typescript
import {createModelHooks} from '@realm/react';

// or
import {createModelHooks} from '../MyRealmContext';

class Person extend Realm.Object{} //some class based model

const [usePersonById, usePerson] = createModelHooks(Person);
```

### A third example where the user just have to pass model classes and the hooks are generated for them ([proposed](https://github.com/realm/realm-js/pull/6272#issuecomment-1826103539) by @bimusiek)

```typescript
enum SchemaName {
    Dog = 'Dog',
    Cat = 'Cat'
}

abstract class RealmModel {
    static schemaName: SchemaName;

    static get() {
        console.log('get');
    }
}

class Dog extends RealmModel {
    static schemaName = SchemaName.Dog

    static woof() {
        console.log('woof');
    }
}

class Cat extends RealmModel {
    static schemaName = SchemaName.Cat

    static meow() {
        console.log('meow')
    }
}

type SchemaWithModels = {
    [SchemaName.Cat]: typeof Cat;
    [SchemaName.Dog]: typeof Dog
};

function createReactUtilities<Model extends typeof RealmModel>(models: Model[]) {
    type SchemaNameHooks = {
[...]

PR #6360, opened 2024-01-09
`@realm/react` key-path filtering on `useQuery` and `useObject`

### LINKED ISSUE #6286: `@realm/react` Expose keypath filtering in `useQuery` and `useObject`

kraenhansen:
## What, How & Why?

This closes #6286

## ☑️ ToDos
* [ ] Figure out a way to document the overload of `useQuery`
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ### Enhancements
 * Adding the ability to pass "options" to `useQuery` and `useObject` as an object. ([#6360](https://github.com/realm/realm-js/pull/6360))
-* Adding `keyPaths` option to the `userQuery` and `useObject` hooks, to indicate a lower bound on the changes relevant for the hook. This is a lower bound, since if multiple hooks add listeners (each with their own `keyPaths`) the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the collection or object. In other words: A listener might fire and cause a re-render more than the key-paths specify, if other listeners with different key-paths are present. ([#6360](https://github.com/realm/realm-js/pull/6360))
+* Adding `keyPaths` option to the `useQuery` and `useObject` hooks, to indicate a lower bound on the changes relevant for the hook. This is a lower bound, since if multiple hooks add listeners (each with their own `keyPaths`) the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the collection or object. In other words: A listener might fire and cause a re-render more than the key-paths specify, if other listeners with different key-paths are present. ([#6360](https://github.com/realm/realm-js/pull/6360))
 
 ### Fixed
 * Removed race condition in `useObject` ([#6291](https://github.com/realm/realm-js/issues/6291)) Thanks [@bimusik](https://github.com/bimusiek)!
--- a/packages/realm-react/CHANGELOG.md
+++ b/packages/realm-react/CHANGELOG.md
@@ -1,10 +1,19 @@
 ## vNext (TBD)
 
 ### Deprecations
-* None
+* Deprecated calling `useQuery` with three positional arguments. ([#6360](https://github.com/realm/realm-js/pull/6360))
+
+  Please pass a single "options" object followed by an optional `deps` argument (if your query depends on any local variables):
+  ```tsx
+  const filteredAndSorted = useQuery({
+    type: Object,
+    query: (collection) => collection.filtered('category == $0',category).sorted('name'),
+  }, [category]);
+  ```
 
 ### Enhancements
-* None
+* Adding the ability to pass "options" to `useQuery` and `useObject` as an object. ([#6360](https://github.com/realm/realm-js/pull/6360))
+* Adding `keyPaths` option to the `userQuery` and `useObject` hooks, to indicate a lower bound on the changes relevant for the hook. This is a lower bound, since if multiple hooks add listeners (each with their own `keyPaths`) the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the collection or object. In other words: A listener might fire and cause a re-render more than the key-paths specify, if other listeners with different key-paths are present. ([#6360](https://github.com/realm/realm-js/pull/6360))
 
 ### Fixed
 * Removed race condition in `useObject` ([#6291](https://github.com/realm/realm-js/issues/6291)) Thanks [@bimusik](https://github.com/bimusiek)!
--- a/packages/realm-react/src/__tests__/createRealmTestContext.ts
+++ b/packages/realm-react/src/__tests__/createRealmTestContext.ts
@@ -50,9 +50,9 @@ export function createRealmTestContext(rootConfig: Configuration = {}): RealmTes
     write(callback: () => void) {
       act(() => {
         context.realm.write(callback);
-        context.realm.write(() => {
-          /* Doing an empty write, since beginning a write transaction will force firing of notifications */
-        });
+        // Starting another write transaction will force notifications to fire
+        context.realm.beginTransaction();
+        context.realm.cancelTransaction();
       });
     },
   };
--- a/packages/realm-react/src/__tests__/createRealmTestContext.ts
+++ b/packages/realm-react/src/__tests__/createRealmTestContext.ts
@@ -16,16 +16,9 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import fs from "node:fs";
-import path from "node:path";
-import os from "node:os";
 import Realm, { Configuration } from "realm";
 import { act } from "@testing-library/react-native";
-
-export function randomRealmPath() {
-  const tempDirPath = fs.mkdtempSync(path.join(os.tmpdir(), "realm-react-tests-"));
-  return path.join(tempDirPath, "test.realm");
-}
+import { randomRealmPath } from "./helpers";
 
 export type RealmTestContext = {
   realm: Realm;
--- a/packages/realm-react/src/__tests__/createRealmTestContext.ts
+++ b/packages/realm-react/src/__tests__/createRealmTestContext.ts
@@ -62,7 +62,7 @@ export function createRealmTestContext(rootConfig: Configuration = {}): RealmTes
     },
     cleanup() {
       Realm.clearTestState();
-    }
+    },
   };
   return context;
 }
--- a/packages/realm-react/src/__tests__/createRealmTestContext.ts
+++ b/packages/realm-react/src/__tests__/createRealmTestContext.ts
@@ -1,0 +1,67 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2024 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import fs from "node:fs";
+import path from "node:path";
+import os from "node:os";
+import Realm, { Configuration } from "realm";
+import { act } from "@testing-library/react-native";
+
+export function randomRealmPath() {
+  const tempDirPath = fs.mkdtempSync(path.join(os.tmpdir(), "realm-react-tests-"));
+  return path.join(tempDirPath, "test.realm");
+}
+
+export type RealmTestContext = {
+  realm: Realm;
+  useRealm: () => Realm;
+  write(cb: () => void): void;
+  openRealm(config?: Configuration): void;
+};
+
+/**
+ * Opens a test realm at a random temporary path.
+ * @returns The `realm` and a `write` function, which will wrap `realm.write` with an `act` and prepand a second `realm.write` to force notifications to trigger.
+ */
+export function createRealmTestContext(rootConfig: Configuration = {}): RealmTestContext {
+  let realm: Realm | null = null;
+  const context = {
+    get realm(): Realm {
+      if (realm) {
+        return realm;
+      } else {
+        throw new Error("Call open first");
+      }
+    },
+    useRealm() {
+      return context.realm;
+    },
+    openRealm(config: Configuration = {}) {
+      realm = new Realm({ ...rootConfig, ...config, path: randomRealmPath() });
+    },
+    write(callback: () => void) {
+      act(() => {
+        context.realm.write(callback);
+        context.realm.write(() => {
+          /* Doing an empty write, since beginning a write transaction will force firing of notifications */
+        });
+      });
+    },
+  };
+  return context;
+}
--- a/packages/realm-react/src/__tests__/createRealmTestContext.ts
+++ b/packages/realm-react/src/__tests__/createRealmTestContext.ts
@@ -16,45 +16,53 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+import assert from "node:assert";
 import Realm, { Configuration } from "realm";
 import { act } from "@testing-library/react-native";
+
 import { randomRealmPath } from "./helpers";
 
 export type RealmTestContext = {
   realm: Realm;
   useRealm: () => Realm;
   write(cb: () => void): void;
-  openRealm(config?: Configuration): void;
+  openRealm(config?: Configuration): Realm;
+  cleanup(): void;
 };
 
 /**
- * Opens a test realm at a random temporary path.
- * @returns The `realm` and a `write` function, which will wrap `realm.write` with an `act` and prepand a second `realm.write` to force notifications to trigger.
+ * Opens a test realm at a randomized and temporary path.
+ * @returns The `realm` and a `write` function, which will wrap `realm.write` with an `act` and prepand a second `realm.write` to force notifications to trigger synchronously.
  */
 export function createRealmTestContext(rootConfig: Configuration = {}): RealmTestContext {
   let realm: Realm | null = null;
   const context = {
     get realm(): Realm {
-      if (realm) {
-        return realm;
-      } else {
-        throw new Error("Call open first");
-      }
+      assert(realm, "Open the Realm first");
+      return realm;
     },
     useRealm() {
       return context.realm;
     },
     openRealm(config: Configuration = {}) {
+      if (realm) {
+        // Close any realm, previously opened
+        realm.close();
+      }
       realm = new Realm({ ...rootConfig, ...config, path: randomRealmPath() });
+      return realm;
     },
     write(callback: () => void) {
       act(() => {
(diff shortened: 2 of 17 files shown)


packages/realm-react/src/index.tsx
```diff
@@ -93,6 +93,7 @@ type RealmContext = {
    * ```
    * @param type - The object type, depicted by a string or a class extending {@link Realm.Object}
    * @param primaryKey - The primary key of the desired object which will be retrieved using {@link Realm.objectForPrimaryKey}
+   * @param keyPaths - Indicates a lower bound on the changes relevant for the hook. This is a lower bound, since if multiple hooks add listeners (each with their own `keyPaths`) the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the object. In other words: A listener might fire and cause a re-render more than the key-paths specify, if other listeners with different key-paths are present.
```
kraenhansen, line 96: I've added this here, but I still need to figure out a way to document the overload on `useQuery`.

packages/realm-react/src/useQuery.tsx
```diff
@@ -60,9 +84,9 @@ export function createUseQuery(useRealm: () => Realm) {
     const updatedRef = useRef(true);
     const queryCallbackRef = useRef<QueryCallback<T> | null>(null);
 
-    // We want the user of this hook to be able pass in the `query` function inline (without the need to `useCallback` on it)
-    // This means that the query function is unstable and will be a redefined on each render of the component where `useQuery` is used
-    // Therefore we use the `deps` array to memoize the query function internally, and only use the returned `queryCallback`
+    /* eslint-disable-next-line react-hooks/exhaustive-deps -- We want the user of this hook to be able pass in the `query` function inline (without the need to `useCallback` on it)
```
kraenhansen, line 87: Turned this comment into an `eslint-disable-next-line` comment to silence the warning.

packages/realm-react/src/__tests__/createRealmTestContext.ts
```diff
@@ -0,0 +1,60 @@
...
+  useRealm: () => Realm;
+  write(cb: () => void): void;
+  openRealm(config?: Configuration): void;
+};
+
+/**
+ * Opens a test realm at a random temporary path.
+ * @returns The `realm` and a `write` function, which will wrap `realm.write` with an `act` and prepand a second `realm.write` to force notifications to trigger.
+ */
+export function createRealmTestContext(rootConfig: Configuration = {}): RealmTestContext {
```
kraenhansen, line 34: I've added this helper, to make it easier to work with Realms in the tests.

packages/realm-react/src/__tests__/createRealmTestContext.ts
```diff
@@ -0,0 +1,60 @@
...
+    openRealm(config: Configuration = {}) {
+      realm = new Realm({ ...rootConfig, ...config, path: randomRealmPath() });
+    },
+    write(callback: () => void) {
+      act(() => {
+        context.realm.write(callback);
+        // Starting another write transaction will force notifications to fire
+        context.realm.beginTransaction();
+        context.realm.cancelTransaction();
+      });
```
kraenhansen, line 56: This simplifies the test code quite a bit, ensuring `act` is called and applying the workaround to force notifications to fire.

kraenhansen:
I'm drafting this to get some early feedback, I might want to merge another PR updating our TypeScript version to > 5.0 to get [support for `@overload` in jsdocs](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#overload-support-in-jsdoc) 🤔

packages/realm-react/src/index.tsx
takameyer: Should we deprecate the previous overloads?  Then you can just document the new way and mention the old as deprecated.

takameyer (commented):
Great updates here.  I really like the helpers for the tests.
I guess the only question is, should we deprecate the old argument methods for `useQuery`?  Would be great to see that marked in consumer code so that it was apparent one should refactor.

kraenhansen:
> should we deprecate the old argument methods for useQuery?

I thought about it, but I'd rather not deprecate something, pointing to another API, that also won't be the preferred way long-term.
I hope something like https://github.com/realm/realm-js/issues/6284 is the future 🤞 Although we need to incorporate `keyPaths` in that proposal 🤔

packages/realm-react/src/index.tsx
kneth, line 96: Could you give an example?

packages/realm-react/src/useObject.tsx
```diff
@@ -58,12 +63,16 @@ export function createUseObject(useRealm: () => Realm) {
     // Ref: https://github.com/facebook/react/issues/14490
     const cachedObjectRef = useRef<null | CachedObject>(null);
 
+    /* eslint-disable-next-line react-hooks/exhaustive-deps -- Memoizing the keyPaths to avoid renders */
+    const memoizedKeyPaths = useMemo(() => keyPaths, [JSON.stringify(keyPaths)]);
```
kneth, line 67: Hard not to read "memorized" 🤣

packages/realm-react/src/__tests__/useObjectHook.test.tsx
```diff
@@ -16,81 +16,102 @@
...
+  { _id: 5, name: "River", age: 25 },
+  { _id: 6, name: "Schatzi", age: 13 },
 ];
 
-describe("useObject hook", () => {
+describe("useObject", () => {
   beforeEach(() => {
-    const realm = new Realm(configuration);
+    context.openRealm();
+    const { realm } = context;
```
elle-j, line 54: What if the `openRealm()` on the context could return the realm as well? Then in places like this one you don't have to use `context.realm` or `context.useRealm()`.

packages/realm-react/src/__tests__/useQueryHook.test.tsx
```diff
@@ -111,4 +97,74 @@ describe("useQueryHook", () => {
 
     expect(collection[99]).toBe(undefined);
   });
+
+  it("can filter objects via a query argument", () => {
```
elle-j, line 101: What do you think of updating the `..via a query option` and `..via a query argument` test names to e.g. `..via an option argument` and `..via separate type and callback arguments`. I initially interperted the "query argument" part as testing for e.g. `$0` etc in the query.

packages/realm-react/src/index.tsx
elle-j: Good idea, an `@example` could be nice. And if so, the same example could be added to the docs in the realm package part as well.

elle-j (commented):
Great updates!

packages/realm-react/src/useQuery.tsx
```diff
@@ -26,25 +26,49 @@ type RealmClassType<T = any> = { new (...args: any): T };
...
+  type: string;
+  query?: QueryCallback<T>;
+  keyPaths?: string[];
+};
+
+type QueryHookClassBasedOptions<T> = {
+  type: RealmClassType<T>;
+  query?: QueryCallback<T>;
+  keyPaths?: string[];
+};
```
elle-j, line 39: Just realized that we should probably allow `keyPaths?: string | string[]` rather than just `string[]` since we do that in the realm package.

packages/realm-react/src/__tests__/createRealmTestContext.ts
kraenhansen: > Were notifications not firing without that workaround?

Yes, but they only fire when the database advances, which happens asynchronously if not pressured by an explicit write transaction. This force the notifications to fire synchronously.

packages/realm-react/src/index.tsx
```diff
@@ -82,6 +80,44 @@ type RealmContext = {
    * @param deps - An array of dependencies that will be passed to {@link React.useMemo}
    * @returns a collection of realm objects or an empty array
    */
+
+  /**
+   * @overload
```
takameyer, line 85: In lieu of difficulties with the "overload" flag, why not just simplify the documentation down to what the desired usage is?
They types can still handle the previous usage.  My 2 cents.

packages/realm-react/src/__tests__/profileHook.tsx
```diff
@@ -0,0 +1,76 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import React, { Profiler, ProfilerOnRenderCallback } from "react";
```
takameyer, line 19: Nice usage of the Profiler here 👍🏼

packages/realm-react/src/useObject.tsx
```diff
@@ -21,18 +21,23 @@ import Realm from "realm";
 
 import { CachedObject, createCachedObject } from "./cachedObject";
 import { CollectionCallback, getObjectForPrimaryKey, getObjects } from "./helpers";
+import { UseRealmHook } from "./useRealm";
+
+export type UseObjectHook = {
+  <T>(type: string, primaryKey: T[keyof T], keyPaths?: string | string[]): (T & Realm.Object<T>) | null;
+  <T extends Realm.Object<any>>(type: { new (...args: any): T }, primaryKey: T[keyof T], keyPaths?: string | string[]): T | null;
```
takameyer, line 28: Appears we have a lint failure here and elsewhere

takameyer (approved):
Looks great 👍🏼

packages/realm-react/src/useQuery.tsx
kraenhansen: I updated the API to allow a single string too.

packages/realm-react/src/__tests__/useQueryHook.test.tsx
kraenhansen: I updated the names 🤞

packages/realm-react/src/index.tsx
kraenhansen: > the desired usage is?

Sure - but I don't know if I actually know what that is 🤔 I guess we could focus on passing an object and use this as a fix for #6259 and eventually make this more convenient via #6284?

takameyer: In my opinion, passing as an object is the desired usage.  If I would have thought of it earlier, it would be the only way to do it.

kraenhansen: If we do that, we could also `@deprecate` the existing signature for `useQuery`?

takameyer: Yes, I think that's the way forward.  We are still on version 0.x.y, so we can make some broader changes.  I think in the end it makes things more extensible if we want to add more options to the hooks

kraenhansen: I've updated the docs to focus solely on the signature where an options object is passed 👍

takameyer: The README should also be updated.  Or perhaps we should just simplify the README to point to the docs page 🤔

kraenhansen: I'm contemplating adding an overload, passing just the type as first argument and avoid deprecation of that:

```ts
const cars = useQuery(Car);
```

feels cleaner than

```ts
const cars = useQuery({ type: Car });
```

takameyer: Yeah, that's cleaner, but it's not "that" bad.

packages/realm-react/src/__tests__/useObjectHook.test.tsx
```diff
@@ -16,81 +16,124 @@
...
+      });
+      expect(renders).toHaveLength(2);
+      expect(result.current?.name).toEqual("Vince!");
+      // Update the age and don't expect a re-render
+      write(() => {
+        if (result.current) {
+          result.current.age = 5;
+        }
+      });
+      expect(renders).toHaveLength(2);
```
elle-j, line 113: If `result.current` is falsy (and we thereby don't update the age), then the test may still pass since there is no new rerender. You do check the `result.current` in the expect before `write()`, so perhaps that's enough, but we could e.g. add an `else` clause in the callback and throw an error to make sure we test that the *update* don't cause a rerender in this case. (There's one more occurrence of this I believe.)

packages/realm-react/src/__tests__/useQueryHook.test.tsx
```diff
@@ -69,20 +52,22 @@ const testDataSet = [
   { _id: 6, name: "Sadie", color: "gold", gender: "female", age: 5 },
 ];
 
-describe("useQueryHook", () => {
+describe("useQuery", () => {
+  const context = createRealmTestContext({ schema: [dogSchema] });
```
elle-j, line 56: Similar comment as before (could destructure this here instead if you prefer).

packages/realm-react/src/index.tsx
```diff
@@ -71,14 +75,25 @@ type RealmContext = {
    * @example
    * ```tsx
    * // Return all collection items
-   * const collection = useQuery(Object)
+   * const collection = useQuery({ type: Query });
```
elle-j, line 78:
```suggestion
   * const collection = useQuery({ type: Object });
```

packages/realm-react/src/index.tsx
```diff
@@ -183,14 +199,25 @@ export const useRealm = defaultContext.useRealm;
  * @example
  * ```tsx
  * // Return all collection items
- * const collection = useQuery(Object)
+ * const collection = useQuery({ type: Query });
```
elle-j, line 202:
```suggestion
 * const collection = useQuery({ type: Object });
```

packages/realm-react/src/useObject.tsx
```diff
@@ -21,18 +21,30 @@ import Realm from "realm";
...
+type AnyRealmObject = Realm.Object<any>;
+
+export type UseObjectHook = {
+  <T>(type: string, primaryKey: T[keyof T], keyPaths?: string | string[]): (T & Realm.Object<T>) | null;
+  <T extends AnyRealmObject>(
+    type: { new (...args: any): T },
+    primaryKey: T[keyof T],
+    keyPaths?: string | string[],
+  ): T | null;
+};
```
elle-j, line 36: Since we're now deprecating the usage of separate arguments for `useQuery()` in favor of an options arg, it introduces some inconsistency between that API and that of `useObject()`. Even though `useObject()` doesn't have the ESLint issue regarding dependency lists, I think we should consider allowing an options argument for that hook as well, especially since the third arg `keyPaths` has been introduced.

I'd be interested in hearing your thoughts on this @kraenhansen @takameyer .

packages/realm-react/src/useQuery.tsx
```diff
@@ -22,29 +22,59 @@ import Realm from "realm";
...
+export type QueryHookOptions<T> = {
+  type: string;
+  query?: QueryCallback<T>;
+  keyPaths?: string | string[];
+};
+
+export type QueryHookClassBasedOptions<T> = {
+  type: RealmClassType<T>;
+  query?: QueryCallback<T>;
+  keyPaths?: string[];
```
elle-j, line 40:
```suggestion
  keyPaths?: string | string[];
```

elle-j (approved):
Great addition, getting closer to letting users try this out 🚀

packages/realm-react/src/useObject.tsx
takameyer: Good consideration.  I would be in favor of using an options object for `useObject()`.

kraenhansen: True -- that would probably be more consistent and read better for some users 👍

packages/realm-react/src/index.tsx
kraenhansen: Oh my ...

packages/realm-react/src/__tests__/useObjectHook.test.tsx
kraenhansen: Good catch, this should probably just assert `result.current` instead 👍

packages/realm-react/src/useQuery.tsx
kraenhansen: Great catch!

packages/realm-react/src/useObject.tsx
kraenhansen: I've push an update including that 👍

packages/realm-react/src/__tests__/useQueryHook.test.tsx
kraenhansen: I don't think I can, because getting the `realm` from `context` needs to happen after it's been opened.

packages/realm-react/src/__tests__/useObjectHook.test.tsx
```diff
@@ -16,81 +16,124 @@
...
-    return () => {
-      realm.close();
-    };
-  }, [realm, setRealm]);
-
-  return new Realm(configuration);
-};
-
-const useObject = createUseObject(useRealm);
+const context = createRealmTestContext({ schema: [dogSchema] });
```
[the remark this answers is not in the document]
kraenhansen: I don't think I can, because getting the `realm` from `context` needs to happen after it's been opened.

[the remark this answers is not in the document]
elle-j: Ah, okay!

kraenhansen:
To get the typedocs for these overloads to shine, I think I'll need to sink a bit more time into the way we export the functions from the default contextn🤔

packages/realm-react/CHANGELOG.md
```diff
@@ -1,10 +1,19 @@
...
+  const filteredAndSorted = useQuery({
+    type: Object,
+    query: (collection) => collection.filtered('category == $0',category).sorted('name'),
+  }, [category]);
+  ```
 
 ### Enhancements
-* None
+* Adding the ability to pass "options" to `useQuery` and `useObject` as an object. ([#6360](https://github.com/realm/realm-js/pull/6360))
+* Adding `keyPaths` option to the `userQuery` and `useObject` hooks, to indicate a lower bound on the changes relevant for the hook. This is a lower bound, since if multiple hooks add listeners (each with their own `keyPaths`) the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the collection or object. In other words: A listener might fire and cause a re-render more than the key-paths specify, if other listeners with different key-paths are present. ([#6360](https://github.com/realm/realm-js/pull/6360))
```
takameyer, line 16:
```suggestion
* Adding `keyPaths` option to the `useQuery` and `useObject` hooks, to indicate a lower bound on the changes relevant for the hook. This is a lower bound, since if multiple hooks add listeners (each with their own `keyPaths`) the union of these key-paths will determine the changes that are considered relevant for all listeners registered on the collection or object. In other words: A listener might fire and cause a re-render more than the key-paths specify, if other listeners with different key-paths are present. ([#6360](https://github.com/realm/realm-js/pull/6360))
```

takameyer (approved):
LGTM.  Does this need to be in draft?  Maybe just a follow up PR for the documentation pieces.

kraenhansen:
I've merged this as is and will follow up with updates to the docs of this once I get the docs for the entire package refactored a bit 👍

PR #6364, opened 2024-01-11
Implement support for flat collections in `Mixed`

### LINKED ISSUE #6342: Store and retrieve flat lists in Mixed

elle-j:
## What, How & Why?

Adds support for storing *flat* lists and dictionaries as the underlying value of a `Mixed` property on non-synced realms. Hence, the list or dictionary can contain any of the previously allowed `Mixed` values, but not another nested collection.

Nested lists and dictionaries will be implemented as a separate PR.

### Brief Overview of Usage

```ts
class CustomObject extends Realm.Object {
  value!: Realm.Types.Mixed;

    name: "CustomObject",
    properties: {
      value: "mixed",
    },
  };
}

const realm = await Realm.open({ schema: [CustomObject] });

// Create an object with a list value as the Mixed property.
const realmObject = realm.write(() => {
  return realm.create(CustomObject, { value: [1, "hello", true] });
});

// Get all objects with a list as the Mixed `value` property
// that contains an item matching `"hello"` at index 1.
realm.objects(CustomObject).filtered(`value[1] == $0`, "hello");

// Update the Mixed property to a dictionary.
realm.write(() => {
  realmObject.value = { int: 1, string: "hello", bool: true };
});

// Get all objects with a dictionary as the Mixed `value`
// property that has a key `string` with the value `"hello"`.
realm.objects(CustomObject).filtered(`value.string == $0`, "hello");
```

Sets are not supported as a `Mixed` value.

### Notes
* Tests for the array methods on `List` will be added as part of the nested collections implementation.
* Logger related changes:
  * Breaking changes in Core requires the logger callback and `setLogLevel()` to receive a `category` as the first argument.
  * These changes will need to be looked at further as the changes I've made here were primarily in order to be able to compile.
  * E.g. we need to decide which log category to use as the default, and the categories need documentation (in Core as well I think).

This closes #6342 and #6343.

## ☑️ ToDos
* [ ] 📝 Changelog entry (could add when the feature branch is merged into `main`)
* [x] 🚦 Tests


integration-tests/tests/src/tests/notifications.ts
```diff
@@ -55,6 +58,87 @@ describe("Notifications", () => {
...
+   * The handler will assert equality of listener's changes with the next element of the array on every run.
+   */
+  const expectDictionaryChangesOnEveryRun = (
+    onEnd: () => void | Mocha.Done,
+    changesOnRun: Realm.DictionaryChangeSet[],
+  ) => {
+    return (_: Realm.Dictionary<any>, changes: Realm.DictionaryChangeSet) => {
+      runCount++;
+
+      expect(changes.insertions).members(changesOnRun[runCount - 1].insertions);
```
kraenhansen, line 103: I would probably store this `changesOnRun[runCount - 1]` in its own `const` before using it three times.

packages/realm/src/PropertyHelpers.ts
```diff
@@ -309,6 +310,61 @@ const ACCESSOR_FACTORIES: Partial<Record<binding.PropertyType, AccessorFactory>>
...
+          let index = 0;
+          for (const item of value) {
+            internal.insertAny(index++, toBinding(item));
+          }
+        } else if (value instanceof Dictionary || isPOJO(value)) {
+          obj.setCollection(columnKey, binding.CollectionType.Dictionary);
+          const internal = binding.Dictionary.make(realm.internal, obj, columnKey);
+          internal.removeAll();
+          for (const key in value) {
+            internal.insertAny(key, toBinding((value as Record<string, unknown>)[key]));
```
kraenhansen, line 358: I'm a bit surprised you need this type assertion 🤔

packages/realm/src/Realm.ts
```diff
@@ -283,13 +285,15 @@ export class Realm {
...
    * @note The log level can be changed during the lifetime of the application.
    * @since 12.0.0
    */
-  static setLogLevel(level: LogLevel) {
-    const bindingLoggerLevel = toBindingLoggerLevel(level);
-    binding.Logger.setDefaultLevelThreshold(bindingLoggerLevel);
+  static setLogLevel(category: LogCategory, level: LogLevel) {
+    // TODO: Adapting to new logging in Core. We may want to add an assertion about the
+    //       category as Core will throw if the category is not one of the valid values.
+    binding.LogCategoryRef.getCategory(category).setDefaultLevelThreshold(toBindingLoggerLevel(level));
```
kraenhansen, line 296: I wonder if this could / should happen in a separate PR.

integration-tests/tests/src/tests/notifications.ts
kraenhansen: Ditto in the collection helper above.

packages/realm/src/Realm.ts
takameyer: I was thinking that too.  This should be focused on Mixed.
This is a pretty large change from core.

packages/realm/src/Realm.ts
```diff
@@ -283,13 +285,15 @@ export class Realm {
...
    * Sets the log level.
+   * @param category - TODO: Add docs.
    * @param level - The log level to be used by the logger. The default value is `info`.
    * @note The log level can be changed during the lifetime of the application.
    * @since 12.0.0
    */
-  static setLogLevel(level: LogLevel) {
-    const bindingLoggerLevel = toBindingLoggerLevel(level);
-    binding.Logger.setDefaultLevelThreshold(bindingLoggerLevel);
+  static setLogLevel(category: LogCategory, level: LogLevel) {
```
takameyer, line 293: We should adapt this to not be a breaking change.  Category could be an optional second argument with a default value.

packages/realm/src/Realm.ts
elle-j: Yup, I also want it to be focused on Mixed. For this PR, the intent was not to include the logger changes as something to be reviewed here but were needed in order to compile, so that I can only focus on Mixed. There's more info on this in the PR description.

But as Andrew and I talked about, I could try to pull it out to a separate branch, then updating the logger will need a ticket so we can plan it 👍

packages/realm/src/Realm.ts
elle-j: We can create a ticket for going over and updating related logger changes, so that this review is not on both Mixed and the logger. (And of course handling that before the target Mixed feature branch is merged.)

kneth: Or `setLogLevel(level: LogLevel)` sets the log level for all known categories

packages/realm/src/PropertyHelpers.ts
elle-j: That was from when the if condition was slightly different, in the current state it's not needed anymore, thanks for spotting that 🙂

packages/realm/src/PropertyHelpers.ts
```diff
@@ -309,6 +310,61 @@ const ACCESSOR_FACTORIES: Partial<Record<binding.PropertyType, AccessorFactory>>
...
+            return fromBinding(binding.Dictionary.make(realm.internal, obj, columnKey));
+          }
+          return defaultGet(options)(obj);
+        } catch (err) {
+          assert.isValid(obj);
+          throw err;
+        }
+      },
+      set: (obj: binding.Obj, value: unknown) => {
+        assert.inTransaction(realm);
```
elle-j, line 344: We can also move this assertion into each if-block below (except for the `else`), in order for the `else` scenario to not call this twice as `defaultSet()` will do the same assertion.

packages/realm/src/PropertyHelpers.ts
```diff
@@ -363,3 +419,14 @@ export function createPropertyHelpers(property: PropertyContext, options: Helper
     });
   }
 }
+
+/** @internal */
+export function isPOJO(value: unknown): value is Record<string, unknown> {
```
kneth, line 424: This could be a good interview question 😄

kneth (approved):
It is amazing to see that collections in mixed is coming to JS 😍

takameyer (approved):
Great! LGTM

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -87,7 +115,71 @@ const MixedNullableSchema: Realm.ObjectSchema = {
...
+  int,
+  double,
+  d128,
+  string,
+  date,
+  oid,
+  uuid,
+  nullValue,
+  uint8Buffer,
+};
```
kraenhansen, line 175: I like that you're defining these const values instead of having them inlined in the tests 👍

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -220,23 +345,882 @@ describe("Mixed", () => {
...
+        MixedWithDefaultCollectionsSchema,
+        CollectionsOfMixedSchema,
+        EmbeddedObjectSchema,
+        TestObject,
+      ],
+    });
+
+    function expectMatchingFlatList(value: unknown) {
+      expect(value).instanceOf(Realm.List);
+      const list = value as Realm.List<any>;
```
kraenhansen, line 363: It's frustrating that `chai`'s return type doesn't `assert` here 😕 We should probably look into using a more TypeScript friendly assert library for our tests ...

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -220,23 +345,882 @@ describe("Mixed", () => {
...
+          expect(value.value).equals(unmanagedRealmObject.value);
+        } else if (key === "uint8Buffer") {
+          expectUint8Buffer(value);
+        } else {
+          expect(String(value)).equals(String(flatDictionaryAllTypes[key]));
+        }
+      }
+    }
+
+    function expectUint8Buffer(value: unknown) {
```
kraenhansen, line 398: It seems this is the convention used in the other functions above.

```suggestion
    function expectMatchingUint8Buffer(value: unknown) {
```

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -220,23 +345,882 @@ describe("Mixed", () => {
...
+      expect(value).instanceOf(ArrayBuffer);
+      expect([...new Uint8Array(value as ArrayBuffer)]).eql(uint8Values);
+    }
+
+    describe("Flat collections", () => {
+      describe("CRUD operations", () => {
+        it("can create and access a list with different types (input: JS Array)", function (this: RealmContext) {
```
kraenhansen, line 405: As this sentence gets repeated, would it make sense to put this description into a separate `describe?

```suggestion
        describe("creating and accessing a list with different types", () => {
          it("handles JS Array", function (this: RealmContext) {
```

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -220,23 +345,882 @@ describe("Mixed", () => {
...
+          this.realm.write(() => {
+            list.push(...flatListAllTypes);
+            list.push(this.realm.create(MixedSchema.name, unmanagedRealmObject));
+          });
+          expectMatchingFlatList(list);
+        });
+
+        it("inserts dictionary entries", function (this: RealmContext) {
+          const created = this.realm.write(() => this.realm.create<IMixedSchema>(MixedSchema.name, { value: {} }));
+          const dictionary = created.value as Realm.Dictionary<any>;
```
kraenhansen, line 537: Again, it would be great if `chai` just did this, but I think this could be simplified if you defined a function asserting this is a dictionary and used that instead.

```
function expectDictionary(value: unknown): asserts value is Realm.Dictionary {
  expect(value).instanceOf(Realm.Dictionary);
}
````

```suggestion
          const { value: dictionary } = this.realm.write(() => this.realm.create<IMixedSchema>(MixedSchema.name, { value: {} }));
          expectDictionary(dictionary);
```

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -220,23 +345,882 @@ describe("Mixed", () => {
...
+        });
+      });
+
+      describe("Notifications", () => {
+        let runCount = 0;
+
+        let testObjectWithList: TestObject;
+        let testObjectWithDictionary: TestObject;
+        let testList: Realm.List<any>;
+        let testDictionary: Realm.Dictionary<any>;
```
kraenhansen, line 753: Would it make sense to use the test context (setting and getting values on `this`) instead?

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -220,23 +345,882 @@ describe("Mixed", () => {
...
+            expect(changes.changedProperties).members(expectedChanges.changedProperties);
+
+            if (runCount >= changesOnRun.length) {
+              // Once runCount reaches given array, run the onEnd callback. If onEnd is Mocha.done then
+              // running it multiple times will cause tests to fail so this also ensures the listener fires
+              // exactly changesOnRun.length times.
+              onEnd();
+            }
+          };
+        };
```
kraenhansen, line 853: I believe this could be simplified by using the [src/utils/sequence.ts](https://github.com/realm/realm-js/blob/main/integration-tests/tests/src/utils/sequence.ts) or perhaps even the [src/utils/listener-stub.ts](https://github.com/realm/realm-js/blob/main/integration-tests/tests/src/utils/listener-stub.ts).

packages/realm/src/PropertyHelpers.ts
```diff
@@ -309,6 +310,61 @@ const ACCESSOR_FACTORIES: Partial<Record<binding.PropertyType, AccessorFactory>>
...
+            return fromBinding(binding.List.make(realm.internal, obj, columnKey));
+          }
+          if (mixedType === binding.MixedDataType.Dictionary) {
+            return fromBinding(binding.Dictionary.make(realm.internal, obj, columnKey));
+          }
+          return defaultGet(options)(obj);
+        } catch (err) {
+          assert.isValid(obj);
+          throw err;
+        }
```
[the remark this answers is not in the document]
kraenhansen: The `fromBinding` calls above might throw some internal core specific error that we'd like an SDK-level error for instead (if object it isn't valid).

packages/realm/src/TypeHelpers.ts
```diff
@@ -88,7 +94,11 @@ export type TypeOptions = {
 // TODO: Consider testing for expected object instance types and throw something similar to the legacy SDK:
 // "Only Realm instances are supported." (which should probably have been "RealmObject")
 // instead of relying on the binding to throw.
-export function mixedToBinding(realm: binding.Realm, value: unknown): binding.MixedArg {
+export function mixedToBinding(
+  realm: binding.Realm,
+  value: unknown,
+  { isQueryArg } = { isQueryArg: false },
```
kraenhansen, line 100: Could / should we document this?

packages/realm/src/tests/utils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { expect } from "chai";
+
+import { isPOJO } from "../PropertyHelpers";
```
kraenhansen, line 21: I'd probably rename this file to `PropertyHelpers.test.ts`.

kraenhansen (approved):
I like to see this much test vs implementation code. Great job!

integration-tests/tests/src/tests/mixed.ts
elle-j: Yeah I know, a bit strange that they don't do that. I've extracted it to a custom helper that wraps the `instanceOf` as you suggested 👍

integration-tests/tests/src/tests/mixed.ts
elle-j: I've applied updates and it was definitely a good idea to use a custom function that asserts, thanks for the suggestion 🙂

packages/realm/src/TypeHelpers.ts
elle-j: Yes 🙂

integration-tests/tests/src/tests/mixed.ts
elle-j: See [this comment](https://github.com/realm/realm-js/pull/6364#discussion_r1471222487) 👍

integration-tests/tests/src/tests/mixed.ts
kraenhansen: This PR or another, either would be fine by me, as long as it was on your radar 👍

integration-tests/tests/src/tests/mixed.ts
elle-j: Found a middle ground, I've grouped the CRUD tests into subsuites.

PR #6365, opened 2024-01-12
Removed "build-dependencies" from `@realm/integration-tests` lint

kraenhansen:
## What, How & Why?

This speeds up the lint workflow, quite significantly and we can do this because we actually don't need the native modules for linting to work.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
Merging early since I see it pass 🤞

PR #6368, opened 2024-01-13
Migrating `realm-web` workflow to NPM workspaces

kraenhansen:
## What, How & Why?

This migrates the `realm-web` workflow from Lerna to to NPM workspaces.


.github/workflows/pr-realm-web.yml
```diff
@@ -32,27 +35,22 @@ jobs:
         with:
           node-version: 20
           registry-url: https://registry.npmjs.org/
-      # Install the root package (--ignore-scripts to avoid downloading or building the native module)
-      - run: npm ci --ignore-scripts
-      # Bootstrap lerna sub-packages
-      - run: npx lerna bootstrap --scope realm-web-integration-tests --include-dependencies
+      # Install the root package
```
takameyer, line 38: I wonder if this entire workflow could be replaced by the `package-unit-tests`?

takameyer (approved):
Amazing. finally `realm-web` is catching up!

packages/realm-network-transport/src/DefaultNetworkTransport.ts
```diff
@@ -31,6 +30,8 @@ import type {
 export class DefaultNetworkTransport implements NetworkTransport {
   public static fetch: Fetch;
   public static AbortController: AbortController;
+  // A hook to transform responses before being returned from fetch
+  public static transformResponse: (response: FetchResponse) => FetchResponse = (response) => response;
```
elle-j, line 34: I know we now use this hook for transforming the response, but since the hook doesn't necessarily have to transform anything we may want to call it something else? I'm thinking e.g. in the open issue where we've discussed exposing `DefaultNetworkTransport` to allow users to add their own `fetch`, they may also use this hook.

Don't have specific name suggestions atm, perhaps something with `middleware` (as it behaves similar to that)? But yeah would be interesting to hear if you think the original name would suffice in this case 👍

elle-j (approved):
Much appreciated updates 🚀

packages/realm-network-transport/src/DefaultNetworkTransport.ts
kraenhansen: The primary need for it to be a separate package, instead of being embedded in [the "platform" abstraction of the `realm` package](https://github.com/realm/realm-js/tree/main/packages/realm/src/platform), is because we use it for both `realm` and `realm-web`: If we move forward on plans to provide an SDK for atlas app services, without native modules & sync, we can delete the `@realm/network-transport` package all together.

.github/workflows/pr-realm-web.yml
kraenhansen: Interesting thought .. it probably could. But there's something in the naming of that workflow that goes against running "integration tests" from it. For now, I think it'll be best to keep separate if we choose to merge `realm-web` into `realm` over time 🤞

takameyer: I guess my idea was any of our packages could be run by `package-unit-tests` (perhaps the naming could be improve).  I originally attempted to get `realm` and `realm-web` in there, but `realm` had too much special needs and `realm-web` needed the work you are merging here.

PR #6370, opened 2024-01-13
Migrating from `@react-native-community/eslint-config` to `@react-native/eslint-config`

kraenhansen:
## What, How & Why?

The latter has taken over as the package used by the template.
I also moved the dependency into the install tests package.json, to prevent a stale version of this from conflicting when updating the React Native apps in the repository.


kraenhansen:
Please review in retrospect. I verified this worked locally before merging.

PR #6390, opened 2024-01-16
Using BaaSaaS in the Realm JS PR workflow

kraenhansen:
## What, How & Why?

This updates the private `@realm/baas-test-server` with `baasaas` sub-commands to start, stop and list containers. The existing docker-based server is moved to a new `docker` command. Invoking the two commands are documented in the README.md of the package.

This PR also creates a JavaScript -based action, invoking the `baas-test-server` CLI (provided by our `@realm/baas-test-server` package) to start and stop the server. I had to use a "node20" action, instead of a "composite", because the latter doesn't support a "post" step, which we need to stop the server as the job ends.

I also wrote a type-safe TypeScript client for the BaaSaaS service, for good measure.

### DIFF
--- a/.github/actions/baas-test-server/action.yml
+++ b/.github/actions/baas-test-server/action.yml
@@ -1,9 +1,6 @@
 name: Start BaaS test server
 description: Uses
 inputs:
-  baasaas-key:
-    description: SDK specific API key for BaaSaaS
-    required: true
   branch:
     description: BaaS branch to use when starting the server
     default: master
--- a/.github/actions/baas-test-server/action.yml
+++ b/.github/actions/baas-test-server/action.yml
@@ -9,40 +9,15 @@ inputs:
     default: master
   githash:
     description: Specific githash to use when starting the server (this is used instead of branch if provided)
-# outputs:
-#   container-id:
-#     description: The id of the container created
-#     value: ${{ steps.baasaas-start.outputs.container-id }}
-#   baas-url:
-#     description: The url of the BaaS server created
-#     value: ${{ steps.baasaas-start.outputs.baas-url }}
-#   mongo-url:
-#     description: The url of the MongoDB instance backing the server created
-#     value: ${{ steps.baasaas-start.outputs.mongo-url }}
+outputs:
+  container-id:
+    description: The id of the container created
+  baas-url:
+    description: The url of the BaaS server created
+  mongo-url:
+    description: The url of the MongoDB instance backing the server created
 
 runs:
   using: node20
   main: start-server.js
-# runs:
-#   using: composite
-#   steps:
-#     - uses: actions/checkout@v3
-#     - uses: actions/setup-node@v3
-#       with:
-#         node-version: 20
-#         cache: npm
-#     - shell: bash
-#       run: npm ci --ignore-scripts
-#     - shell: bash
-#       run: npx baas-test-server baasaas start ${{ inputs.githash }} --branch ${{ inputs.branch }}
-#       id: baasaas-start
-#       env:
-#         BAASAAS_KEY: ${{ inputs.baasaas-key }}
-#     # Using another action to create a post step, since "composite" actions doesn't support those
-#     # See https://github.com/actions/runner/issues/1478
-#     - name: Stop server
-#       uses: TLizer/action-with-post-step@v1
-#       with:
-#         main: echo 'Nothing to do right now'
-#         post: npx baas-test-server baasaas start ${{ inputs.githash }} --branch ${{ inputs.branch }}
-
+  post: stop-server.js
--- a/.github/actions/baas-test-server/action.yml
+++ b/.github/actions/baas-test-server/action.yml
@@ -1,0 +1,44 @@
+name: Start BaaS test server
+description: Uses
+inputs:
+  baasaas-key:
+    description: SDK specific API key for BaaSaaS
+    required: true
+  branch:
+    description: BaaS branch to use when starting the server
+    default: master
+  githash:
+    description: Specific githash to use when starting the server (this is used instead of branch if provided)
+outputs:
+  container-id:
+    description: The id of the container created
+    value: ${{ steps.baasaas-start.outputs.container-id }}
+  baas-url:
+    description: The url of the BaaS server created
+    value: ${{ steps.baasaas-start.outputs.baas-url }}
+  mongo-url:
+    description: The url of the MongoDB instance backing the server created
+    value: ${{ steps.baasaas-start.outputs.mongo-url }}
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 20
+        cache: npm
+    - shell: bash
+      run: npm ci --ignore-scripts
+    - shell: bash
+      run: npx baas-test-server baasaas start ${{ inputs.githash }} --branch ${{ inputs.branch }}
+      id: baasaas-start
+      env:
+        BAASAAS_KEY: ${{ inputs.baasaas-key }}
+    # Using another action to create a post step, since "composite" actions doesn't support those
+    # See https://github.com/actions/runner/issues/1478
+    - name: Stop server
+      uses: TLizer/action-with-post-step@v1
+      with:
+        main: echo 'Nothing to do right now'
+        post: npx baas-test-server baasaas stop ${{ steps.baasaas-start.outputs.container-id }}
--- a/.github/actions/baas-test-server/action.yml
+++ b/.github/actions/baas-test-server/action.yml
@@ -9,36 +9,40 @@ inputs:
     default: master
   githash:
     description: Specific githash to use when starting the server (this is used instead of branch if provided)
-outputs:
-  container-id:
-    description: The id of the container created
-    value: ${{ steps.baasaas-start.outputs.container-id }}
-  baas-url:
-    description: The url of the BaaS server created
-    value: ${{ steps.baasaas-start.outputs.baas-url }}
-  mongo-url:
-    description: The url of the MongoDB instance backing the server created
-    value: ${{ steps.baasaas-start.outputs.mongo-url }}
+# outputs:
+#   container-id:
+#     description: The id of the container created
+#     value: ${{ steps.baasaas-start.outputs.container-id }}
+#   baas-url:
+#     description: The url of the BaaS server created
+#     value: ${{ steps.baasaas-start.outputs.baas-url }}
+#   mongo-url:
+#     description: The url of the MongoDB instance backing the server created
+#     value: ${{ steps.baasaas-start.outputs.mongo-url }}
 
 runs:
-  using: composite
-  steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-node@v3
-      with:
-        node-version: 20
-        cache: npm
-    - shell: bash
-      run: npm ci --ignore-scripts
-    - shell: bash
-      run: npx baas-test-server baasaas start ${{ inputs.githash }} --branch ${{ inputs.branch }}
-      id: baasaas-start
-      env:
-        BAASAAS_KEY: ${{ inputs.baasaas-key }}
-    # Using another action to create a post step, since "composite" actions doesn't support those
-    # See https://github.com/actions/runner/issues/1478
-    - name: Stop server
-      uses: TLizer/action-with-post-step@v1
-      with:
-        main: echo 'Nothing to do right now'
-        post: npx baas-test-server baasaas stop ${{ steps.baasaas-start.outputs.container-id }}
+  using: node20
+  main: start-server.js
+# runs:
+#   using: composite
+#   steps:
+#     - uses: actions/checkout@v3
+#     - uses: actions/setup-node@v3
+#       with:
+#         node-version: 20
+#         cache: npm
+#     - shell: bash
+#       run: npm ci --ignore-scripts
+#     - shell: bash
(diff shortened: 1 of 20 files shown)


.github/actions/baas-test-server/action.yml
```diff
@@ -0,0 +1,20 @@
...
+outputs:
+  container-id:
+    description: The id of the container created
+  baas-url:
+    description: The url of the BaaS server created
+  mongo-url:
+    description: The url of the MongoDB instance backing the server created
+
+runs:
+  using: node20
```
kraenhansen, line 18: I would actually have liked to just use a composite action here, but it doesn't support specifying a "post" step.

.github/actions/baas-test-server/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@realm/baas-test-server-action",
```
kraenhansen, line 2: We could probably rename the start-server and stop-server scripts to use an .mjs extension and avoid this package.json. On the other hand, it could come in handy to define action specific dependencies if added as a workspace package too.

.github/actions/baas-test-server/action.yml
```diff
@@ -0,0 +1,20 @@
+name: Start BaaS test server
```
kraenhansen, line 1: Note that this action is coupled to the Realm JS repo, since it depends on the `baas-test-server` CLI and its dependencies being installed ahead of being invoked.

kraenhansen:
@elle-j it seems the SSL tests for Node.js depends on a server running TLS: https://github.com/realm/realm-js/actions/runs/7548643523/job/20559519210?pr=6390

kraenhansen:
For some reason, it seems the iOS simulator isn't starting 🤔 I suspect it's unrelated.

integration-tests/baas-test-server/cli.ts
```diff
@@ -0,0 +1,204 @@
...
+
+const COMMON_ID_PREFIX = "arn:aws:ecs:us-east-1:969505754201:task/baas-container-service2/";
+
+function abbreviateId(id: string) {
+  return id.replace(COMMON_ID_PREFIX, "");
+}
+
+function expandId(id: string) {
+  return id.startsWith(COMMON_ID_PREFIX) ? id : COMMON_ID_PREFIX + id;
+}
```
kraenhansen, line 38: @mpobrien for now, abbreviating the id seems safe. Is this just because the service isn't being scaled up?

I should probably check if the if match a pattern instead of starting with a specific prefix here 🤔

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+  const response = await fetch(url, { method: "POST", headers: createHeaders(true) });
+  assert(response.ok);
+  const json = await response.json();
+  assertStartedContainer(json);
+  return json;
+}
+
+type StopContainerResponse = {
+  /* TODO: Fill in */
```
takameyer, line 68: Did you want to do that before merging this?

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -0,0 +1,225 @@
...
+  _id: string;
+  project: string;
+  order: number;
+  versionId: string;
+  taskId: string;
+  execution: number;
+  timestamp: string;
+  branch: string;
+  revision: string;
+  */
```
takameyer, line 168:
```suggestion
```

integration-tests/baas-test-server/cli.ts
```diff
@@ -0,0 +1,204 @@
...
+
+          if (GITHUB_ACTIONS) {
+            gha.notice(`Server listening on ${httpUrl}`, { title: "Started BaaS server" });
+            // Useful when called from a wrapper action
+            gha.saveState("CONTAINER_ID", container.id);
+            gha.setOutput("container-id", container.id);
+            gha.setOutput("baas-url", httpUrl);
+            gha.setOutput("mongo-url", mongoUrl);
+          }
+          // TODO: On GitHub poll the triggering workflow run
```
takameyer, line 164: Is this needed?

integration-tests/baas-test-server/cli.ts
mpobrien: Either one should work fine, they are both valid in the params to AWS. I would recommend not hardcoding the prefix anywhere in case we ever wanted to change the ECS cluster (which is where the prefix comes from). To guarantee that they will always roundtrip correctly, your best bet is to always just use the IDs exactly as they are returned from BaaSaaS, and don't do post-processing on them at all. If you just want to show an abbreviated ID for human-readable display, splitting on `"/"` and getting the last element works fine.

kraenhansen: > your best bet is to always just use the IDs exactly as they are returned from BaaSaaS

I'm always sending the exact id back to baasaas, it's mostly for display purposes when running the cli manually.

> splitting on "/" and getting the last element is safe and works fine.

Great - I'll try that instead then 👍

integration-tests/baas-test-server/README.md
```diff
@@ -1,31 +1,47 @@
...
-* When running the tests, Mac may block execution of the file. If so, go
-   to `System Settings > Privacy & Security` (for Mac), find blocked files, then allow `assisted_agg`.
-
-Run the `start` script, sit back and relax as a mongo server is started, BaaS is pulled, built and started with a proper configuration 🤞
+Run the `start` script, sit back and relax as BaaS is pulled and started.
 
 ```shell
-npm start
+npx baas-test-server docker
 ```
```
elle-j, line 25: With "Run the `start` script" I'd assume `npm start`. The current start script does run `baas-test-server`, but it may be valuable to have helper `npm run <script>` for `npx baas-test-server docker` and `npx baas-test-server baasaas start` (further below). What do you think?

integration-tests/baas-test-server/README.md
```diff
@@ -1,31 +1,47 @@
...
+```shell
+# .env
+BAASAAS_KEY=...
+```
+
+Run the `start` script, sit back and relax as BaaS is started remotely.
+
+```shell
+npx baas-test-server baasaas start
+```
```
elle-j, line 47: Same as previous comment.

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -0,0 +1,225 @@
...
+  assertStartedContainer(json);
+  return json;
+}
+
+type StopContainerResponse = {
+  /* TODO: Fill in */
+};
+
+function assertStopContainerResponse(value: unknown): asserts value is StopContainerResponse {
+  /* TODO: Fill in */
```
elle-j, line 72: `TODO` 👍

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -0,0 +1,225 @@
...
+  const response = await fetch(url, { method: "POST", headers: createHeaders(true) });
+  assert(response.ok);
+  const json = await response.json();
+  assertStopContainerResponse(json);
+}
+
+type ContainerStatus = {
+  id: string;
+  lastStatus: "RUNNING" | "PENDING" | string;
+  startedAt: string | Record<never, never>;
```
elle-j, line 87: Is there a particular scenario where `startedAt` is an object?

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -0,0 +1,225 @@
...
+
+type Images = {
+  allBranches: string[];
+  images: Record<string, undefined | Image[]>;
+};
+
+function assertImages(value: unknown): asserts value is Images {
+  assert(typeof value === "object" && value !== null);
+  const object = value as Record<string, unknown>;
+  assert(Array.isArray(object.allBranches));
```
elle-j, line 186: Do you also want to loop thru the items and assert that they're strings?

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export async function listContainers(mine = false) {
+  const url = new URL("listContainers", BAASAAS_BASE_URL);
+  if (mine) {
+    url.searchParams.append("mine", "true");
+  }
+  const response = await fetch(url, { headers: createHeaders(true) });
+  assert(response.ok, response.statusText);
+  const json = await response.json();
+  assertListContainers(json);
```
elle-j, line 223: If you want you could possibly refactor these four lines to a separate function returning the json, as I think most functions do these (with the differences being the `fetch` options and the assert function called at the end).

elle-j (approved):
Thanks for making these updates, and nice CLI 🙂 I have minor comments, but the main one is to just add the implementation for `function assertStopContainerResponse()` and `type StopContainerResponse` 👍

.github/actions/baas-test-server/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  container-id:
+    description: The id of the container created
+  baas-url:
+    description: The url of the BaaS server created
+  mongo-url:
+    description: The url of the MongoDB instance backing the server created
+
+runs:
+  using: node20
+  main: start-server.js
```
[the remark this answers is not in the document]
kraenhansen: The description of the PR mentions this, but I'll add an inline comment too.

.github/actions/baas-test-server/action.yml
kraenhansen, line 18:
```suggestion
  # Using "composite" here would be ideal, but it doesn't support a "stop" step, which we need to shutdown the server.
  using: node20
```

integration-tests/baas-test-server/baasaas.ts
kraenhansen: Yeah - I believe it's a bug in the service.
@mpobrien have you seen this too? It seems the `startedAt` can be an object, when the container is pending.

mpobrien: Yeah, I think this may be possible if the container was created but ECS did not start it yet, which is a gap of usually a few seconds. I can update the code to leave this field undefined when in this state.

mpobrien: OK, this should be fixed now as of https://github.com/10gen/baasaas/commit/5171e89423e117d27119a4a5c2e42d0b555c46ca

integration-tests/baas-test-server/baasaas.ts
kraenhansen: True.

integration-tests/baas-test-server/baasaas.ts
kraenhansen: I've changed this to make it an optional string and simply delete the property if it's not a string.

integration-tests/baas-test-server/baasaas.ts
```diff
@@ -97,10 +97,13 @@ function assertContainerStatus(value: unknown): asserts value is ContainerStatus
   assert(typeof value === "object" && value !== null);
   assert("id" in value && typeof value.id === "string");
   assert("lastStatus" in value && typeof value.lastStatus === "string");
-  assert("startedAt" in value && (typeof value.startedAt === "string" || typeof value.startedAt === "object"));
   assert("createdAt" in value && typeof value.createdAt === "string");
   assert("creatorId" in value && typeof value.creatorId === "string");
   assert("creatorName" in value && typeof value.creatorName === "string");
+  if ("startedAt" in value && typeof value.startedAt !== "string") {
+    // The service will sometimes respond with an empty object in `startedAt`
```
mpobrien, line 104: should be safe to remove this now (per other thread)

integration-tests/baas-test-server/baasaas.ts
kraenhansen: @mpobrien I'm not certain that fix actually works? Since the empty object in JS is truthy.

integration-tests/baas-test-server/baasaas.ts
kraenhansen: It was a good idea - I did the refactor and added a bit of error handling too 👍

integration-tests/baas-test-server/baasaas.ts
mpobrien: That might be an artifact of the serialization, since `this.result.startedAt` is actually a `Date`.  AFAICT the fix works for me, I was able to run `ls` and catch it at the right time with this output:
```[
  {
    "id": "arn:aws:ecs:us-east-1:969505754201:task/baas-container-service2/2db388ed48c54b37aced9cfc3a710b6f",
    "lastStatus": "RUNNING",
    "createdAt": "2024-01-17T21:40:58.474Z",
    "startedAt": "2024-01-17T21:41:00.633Z",
    "creatorId": "656f6a2f5bdfa7ee906ae64a",
    "creatorName": "...",
    "httpUrl": "...",
    "mongoUrl": "...",
    "tags": []
  },
  {
    "id": "arn:aws:ecs:us-east-1:969505754201:task/baas-container-service2/e0a341515dcc4f4bb72124366231eff7",
    "lastStatus": "RUNNING",
    "createdAt": "2024-01-17T21:22:36.399Z",
    "startedAt": "2024-01-17T21:22:37.366Z",
    "creatorId": "65a598d9a9ff26ae5fc14043",
    "creatorName": "realm-js-ci",
    "httpUrl": "...",
    "mongoUrl": "...",
    "tags": []
  },
  {
    "id": "arn:aws:ecs:us-east-1:969505754201:task/baas-container-service2/ee144edcf0a94e338353c8f7f2c5cbbe",
    "lastStatus": "PENDING",
    "createdAt": "2024-01-17T21:42:14.66Z",
    "creatorId": "656f6a2f5bdfa7ee906ae64a",
    "creatorName": "...",
    "tags": []
  }
]

PR #6396, opened 2024-01-19
iOS with core prebuilds and build from source option

takameyer:
## What, How & Why?
* Build with core prebuilds and the source code needed to bind realm-js to core, as well as RN dependencies
* Allows flexibility in the platform used as well
* Option to build core from source

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


integration-tests/environments/react-native/ios/Podfile
```diff
@@ -66,8 +66,6 @@ target 'RealmReactNativeTests' do
           # Applying https://github.com/facebook/folly/issues/1470#issuecomment-943123653
           config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', 'FOLLY_HAVE_CLOCK_GETTIME=1']
         end
-        config.build_settings['CC'] = "../../../../../scripts/ccache-clang.sh"
-        config.build_settings['CXX'] = "../../../../../scripts/ccache-clang++.sh"
```
kraenhansen, line 70: Why is this removed? It seems valuable for all other native code being built?

takameyer: It was causing problems, which at the moment I can't remember what exactly.  I can try and get it back in

packages/realm/RealmJS.podspec
```diff
@@ -35,36 +37,66 @@ Pod::Spec.new do |s|
...
-                                  '"$(PODS_ROOT)/Headers/Public/React-Core/"'
-                                  #"'#{app_path}/ios/Pods/Headers/Public/React-Core'" # Use this line instead of 👆 while linting
+                                  '"$(PODS_TARGET_SRCROOT)/react-native/ios/build/include/"',
+                                  '"$(PODS_TARGET_SRCROOT)/binding/"',
+                                  '"$(PODS_TARGET_SRCROOT)/bindgen/src/"',
+                                  '"$(PODS_TARGET_SRCROOT)/bindgen/vendor/realm-core/bindgen/src/"'
                                 ].join(' ')
                               }
+  # Create placeholders for vendored_libraries, so they are added to the xcode project
+  s.prepare_command = <<-EOS
```
kraenhansen, line 76: Instead of using a prepare command, why not simply execute the scripts right here and now?

packages/realm/scripts/ccache-clang++.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
kraenhansen, line 1: I don't get why this is moved into the SDK?

packages/realm/RealmJS.podspec
takameyer: I don't understand why not do it in the prepare command.  I want this to run exactly in the phase described [here](https://guides.cocoapods.org/syntax/podspec.html#prepare_command)

packages/realm/scripts/cmake-ios.sh
```diff
@@ -0,0 +1,17 @@
+
```
kraenhansen, line 1: What uses this?

takameyer: This can go away

packages/realm/RealmJS.podspec
kraenhansen: Oh. Re-reading this, I confused this with a [script_phases](https://guides.cocoapods.org/syntax/podspec.html#script_phases). I agree this seems to be serving the purpose 👍

integration-tests/environments/react-native/ios/Podfile
takameyer: I was able to reactivate this.

packages/realm/scripts/ccache-clang++.sh
takameyer: Good point.  I've moved this back into the root project scripts.

packages/realm/RealmJS.podspec
```diff
@@ -35,36 +37,66 @@ Pod::Spec.new do |s|
...
+
+  #Post install script
+  s.script_phase = {
+    :name => 'Retrieve libraries and headers',
+    :execution_position => :before_compile,
+    :input_file_lists => ["#{__dir__}/react-native/ios/input-files.xcfilelist"],
+    :output_file_lists => ["#{__dir__}/react-native/ios/output-files.xcfilelist"],
+    :script => <<-EOS
+    BUILD_REALM_CORE="#{build_realm_core ? '1' : '0'}"
+    CMAKE_PATH="#{CMAKE_PATH}"
```
kraenhansen, line 98: Could we store and read this using an [xcconfig](https://guides.cocoapods.org/syntax/podspec.html#pod_target_xcconfig) instead?

packages/realm/RealmJS.podspec
```diff
@@ -35,36 +37,66 @@ Pod::Spec.new do |s|
...
-  # TODO: Consider providing an option to build with the -dbg binaries instead
-  s.vendored_frameworks = 'react-native/ios/realm-js-ios.xcframework'
+  s.vendored_libraries = "react-native/ios/lib/librealm.a",
+                         "react-native/ios/lib/librealm-object-store.a",
+                         "react-native/ios/lib/librealm-parser.a",
+                         "react-native/ios/lib/librealm-sync.a"
 
   s.dependency 'React'
+
+  CMAKE_PATH = Pod::Executable::which!('cmake')
```
kraenhansen, line 88: Use `witch` instead of `witch!` to avoid throwing if it isn't found: https://rubydoc.info/gems/cocoapods/Pod/Executable#which-class_method

packages/realm/scripts/generate-ios-libs.sh
```diff
@@ -0,0 +1,69 @@
...
+
+set -e
+set -o pipefail
+
+# Start in the root directory of the project.
+cd "$(dirname "$0")/.."
+
+# Take homebrew installs as well
+CMAKE_DIRECTORY=$(dirname "$CMAKE_PATH")
+export PATH="$CMAKE_DIRECTORY:$PATH"
```
kraenhansen, line 11: We should avoid referring to `CMAKE_PATH` until we need it (`BUILD_REALM_CORE` is set).

packages/realm/scripts/generate-ios-libs.sh
```diff
@@ -0,0 +1,69 @@
...
+
+pushd lib
+
+# Debug builds add '-dbg' to the filename, but xcode has links created, so renaming is necessary
+find . -type f -name "*-dbg*" | while read -r file; do
+    # Construct new filename by removing '-dbg'
+    newfile=$(echo "$file" | sed 's/-dbg//')
+    # Rename the file
+    mv "$file" "$newfile"
+done
```
kraenhansen, line 64: I think we should either instruct Podspec to handle these too.

packages/realm/RealmJS.podspec
```diff
@@ -35,36 +37,66 @@ Pod::Spec.new do |s|
...
+
+  #Post install script
+  s.script_phase = {
+    :name => 'Retrieve libraries and headers',
+    :execution_position => :before_compile,
+    :input_file_lists => ["#{__dir__}/react-native/ios/input-files.xcfilelist"],
+    :output_file_lists => ["#{__dir__}/react-native/ios/output-files.xcfilelist"],
```
kraenhansen, line 95: Would it be possible to use the `$(PODS_TARGET_SRCROOT)` here instead?

packages/realm/scripts/generate-dummy-libs.sh
```diff
@@ -0,0 +1,14 @@
...
+
+rm -rf $LIB_DIR
+mkdir $LIB_DIR
+pushd $LIB_DIR
+
+touch librealm.a librealm-sync.a librealm-parser.a librealm-object-store.a
+popd
+
+# Wiping the build dir.  If this is being called, then this will ensure the phase script is executed again.
+rm -rf $BUILD_DIR
```
kraenhansen, line 14: This removes the build cache too, could we make this more specific?

packages/realm/react-native/ios/output-files.xcfilelist
```diff
@@ -0,0 +1,5 @@
+
```
kraenhansen, line 1:
```suggestion
```

packages/realm/RealmJS.podspec
```diff
@@ -5,6 +5,8 @@ package = JSON.parse(File.read(File.expand_path('package.json', __dir__)))
 
 app_path = File.expand_path('../..', __dir__)
 
+build_realm_core = ENV["BUILD_REALM_CORE"] == "1"
```
elle-j, line 8: Perhaps the env could be `REALM_BUILD_CORE` to be more consistent with the other envs by using `REALM` as a prefix?

packages/realm/bindgen/jsi.cmake
```diff
@@ -4,8 +4,19 @@
 # This enables building for iOS on the end-users machine,
 # where "react-native" is installed as a sibling to our package instead of being a dev-dependency of our package.
 
+if(NOT DEFINED NODE_PATH)
+set(NODE_PATH $ENV{NODE_PATH})
+    if(NOT NODE_PATH)
+    find_program(NODE_PATH node)
+    endif()
+endif()
```
elle-j, line 12:
```suggestion
if(NOT DEFINED NODE_PATH)
    set(NODE_PATH $ENV{NODE_PATH})
    if(NOT NODE_PATH)
        find_program(NODE_PATH node)
    endif()
endif()
```

packages/realm/bindgen/jsi.cmake
```diff
@@ -4,8 +4,19 @@
...
+
+if(NOT DEFINED NODE_PATH)
+    message(FATAL_ERROR "Node.js not found")
+endif()
```
elle-j, line 16: Perhaps this could be incorporated into the above `if(NOT DEFINED NODE_PATH)`.

packages/realm/RealmJS.podspec
takameyer: That would resolve to "node_modules/realm". I don't think that is the correct output for this script_phase.  
What is this staying consistent with?

packages/realm/RealmJS.podspec
takameyer: Great idea, works well

packages/realm/scripts/generate-dummy-libs.sh
takameyer: I added the `react-native/lib` directory to the output list.  Now it rebuilds as expected without the need to wipe the build folder here

takameyer:
CI is green.  Ready to review and merge

.github/workflows/pr-realm-js.yml
```diff
@@ -92,15 +93,15 @@ jobs:
...
+        uses: actions/cache@v4
         with:
           path: ${{ steps.npm-cache-dir.outputs.dir }}
           key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
 
       - name: Setup Wireit cache
-        uses: google/wireit@setup-github-actions-caching/v1
+        uses: takameyer/wireit@wireit-caching-gha
```
elle-j, line 104: Is this correct? (Same in other places)

.github/workflows/pr-realm-js.yml
```diff
@@ -293,24 +255,22 @@ jobs:
...
       - name: Restore React Native cache
-        if: ${{ matrix.variant.environment == 'react-native' }}
+        if: ${{ matrix.variant.os == 'ios' }}
         uses: actions/cache@v3
         with:
           path: '**/Pods'
-          key: ${{ runner.os }}-${{matrix.variant.environment}}-${{ hashFiles('**/Podfile.lock', './src/**', './vendor/**') }}
-          restore-keys: |
-            ${{ runner.os }}-${{matrix.variant.environment}}-
+          key: ${{ runner.os }}-${{matrix.variant.environment}}-${{ hashFiles('**/Podfile.lock', './packages/realm/binding/src/**', './packages/realm/bindgen/vendor/**', './packages/realm/RealmJS.podspec', './packages/realm/scripts/*') }}
```
elle-j, line 273: Currently with one vendor it doesn't really matter, but would it be preferable to use `'./packages/realm/bindgen/vendor/realm-core/**'` instead of `../vendor/`?

contrib/debugging-react-native.md
```diff
@@ -14,9 +14,12 @@ In your realm-js project, run the following commands to prepare for debug mode:
 ```
 # install js dependencies, but skip building the node binaries
 $ npm install --ignore-scripts
+```
+
+When running `pod install` for iOS, be sure to set the environment variable `REALM_BUILD_CORE=1`.  This will flag core to be built from source with the Debug configuration.
 
-# build the xcframework
-$ ./scripts/build-ios.sh -c Debug -s
+```
```
elle-j, line 21: Could add these to missing ones 🙂 
```suggestion
```sh
```

packages/realm/RealmJS.podspec
```diff
@@ -50,21 +57,43 @@ Pod::Spec.new do |s|
...
+  EOS
 
-  # TODO: Consider providing an option to build with the -dbg binaries instead
-  s.vendored_frameworks = 'react-native/ios/realm-js-ios.xcframework'
+  s.vendored_libraries = "react-native/ios/lib/*.a"
 
   s.dependency 'React'
+
+
+  #Post install script
```
elle-j, line 89:
```suggestion
  # Post install script
```

elle-j (approved):
Such a great effort here! Happy to have learned a lot more from the work on this project! 💪

.github/workflows/pr-realm-js.yml
takameyer: There is a bit of a thread I started in the private chat about this.  We are getting many warnings about using Node 16, which is coming from different github actions.  This one was quite trivial to fork and change (the original author hasn't touched it in 2 years)

I've cleared up a few of the issues, but there is still some others to take care of (`setup-java`, `setup-android`).  Also I had to keep the `upload-artifacts` to v3 because there's a breaking change in v4 (rabbit hole).

elle-j: Got it, thanks, I've caught up with the thread now 👍

kraenhansen, line 104: I don't think this is strictly needed here, I would rather like that we make them upgrade their action and upgrade when its available instead of forking. For that I've created an issue on their repo: https://github.com/google/wireit/issues/1039

.gitignore
```diff
@@ -97,14 +97,9 @@ coverage/
 **/realm-config
 
 #Ignore symlinked directories (otherwise on Windows they are shown as untracked files)
```
kraenhansen, line 99: It seems the file has changed too much for this comment to be relevant.
```suggestion
```

integration-tests/environments/react-native/package.json
```diff
@@ -31,16 +31,9 @@
     },
     "pod-install:simulator": {
```
kraenhansen, line 32: As we talked about on Zoom, this could be collapsed into a single "pod-install" now that it doesn't have platform specific "build" dependencies.

integration-tests/environments/react-native/package.json
```diff
@@ -116,6 +109,7 @@
     "test:ios": {
       "command": "npm run common",
       "dependencies": [
+        "../../../packages/realm:bindgen:generate:jsi",
```
kraenhansen, line 112: Shouldn't this be a dependency of the `pod-install` job instead?

packages/realm/RealmJS.podspec
```diff
@@ -5,6 +5,10 @@ package = JSON.parse(File.read(File.expand_path('package.json', __dir__)))
 
 app_path = File.expand_path('../..', __dir__)
 
+build_realm_core = ENV["REALM_BUILD_CORE"] == "1"
```
kraenhansen, line 8: Since this is only being used in a single place, would it make sense to inline it?

packages/realm/RealmJS.podspec
```diff
@@ -35,13 +39,16 @@ Pod::Spec.new do |s|
...
   s.source                 = { :http => 'https://github.com/realm/realm-js/blob/main/CONTRIBUTING.md#how-to-debug-react-native-podspec' }
 
-  s.source_files           = 'react-native/ios/RealmReact/*.mm'
+  s.source_files           = 'react-native/ios/RealmReact/*.mm',
+                             'binding/jsi/*.cpp',
+                             'binding/ios/platform.mm'
+
   s.public_header_files    = 'react-native/ios/RealmReact/*.h'
 
   s.frameworks             = uses_frameworks ? ['React'] : []
```
kraenhansen, line 54: If we can avoid '"$(PODS_ROOT)/Headers/Public/React-Core/"' in `'HEADER_SEARCH_PATHS'`, I'm left wondering if we could avoid this as well?

packages/realm/RealmJS.podspec
```diff
@@ -50,21 +57,43 @@ Pod::Spec.new do |s|
...
+
+
+  #Post install script
+  s.script_phase = {
+    :name => 'Retrieve libraries and headers',
+    :execution_position => :before_compile,
+    :input_file_lists => ["$(PODS_TARGET_SRCROOT)/react-native/ios/input-files.xcfilelist"],
+    :output_file_lists => ["$(PODS_TARGET_SRCROOT)/react-native/ios/output-files.xcfilelist"],
+    :script => <<-EOS
+    source "$PODS_TARGET_SRCROOT/scripts/generate-ios-libs.sh"
```
kraenhansen, line 96: I don't think "generate" is the best name for a script that potentially downloads 🤔

takameyer: it downloads or builds.  What verb should we use for that?

takameyer: `install`?

integration-tests/environments/react-native/package.json
takameyer: It would if this command generated the dummy files and inout list.  I think creating a script to do that would solve the problem

packages/realm/RealmJS.podspec
takameyer: Fetch, Build, Assemble, Construct, Retrieve, Synthesize, Compile, Download, Install, Resolve, Sync, Fabricate, Configure, Provision

takameyer: @realm/realm-js-team thoughts?

takameyer: @realm/realm-js-team thoughts?

kraenhansen: How about `download-or-build-ios.sh`?

.github/workflows/pr-realm-js.yml
takameyer: I'm fine with that.  Thanks for creating the issue.  Was planning on creating a PR, but wasn't sure how to make a PR against a tag :shrug:

packages/realm/scripts/generate-ios-libs.sh
```diff
@@ -0,0 +1,72 @@
...
+    cp -R _CPack_Packages/$PLATFORM_NAME/TGZ/realm-$CONFIGURATION-$CORE_VERSION-$PLATFORM_NAME/devel/* $PROJECT_ROOT/react-native/ios/build
+    popd
+else
+    # Download core prebuild and extract
+    curl -o $TAR_FILE_NAME $CDN_URL
+    # Check if the download was successful
+    if [ -f "$TAR_FILE_NAME" ]; then
+        # Extract the contents
+        tar -xzf "$TAR_FILE_NAME"
+
```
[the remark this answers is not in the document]
takameyer: What should be removed?

packages/realm/RealmJS.podspec
takameyer: Sure

packages/realm/RealmJS.podspec
takameyer: Are you suggesting removing the `uses_frameworks` check?

packages/realm/scripts/download-or-build-ios.sh
```diff
@@ -0,0 +1,72 @@
...
+    cp -R _CPack_Packages/$PLATFORM_NAME/TGZ/realm-$CONFIGURATION-$CORE_VERSION-$PLATFORM_NAME/devel/* $PROJECT_ROOT/react-native/ios/build
+    popd
+else
+    # Download core prebuild and extract
+    curl -o $TAR_FILE_NAME $CDN_URL
+    # Check if the download was successful
+    if [ -f "$TAR_FILE_NAME" ]; then
+        # Extract the contents
+        tar -xzf "$TAR_FILE_NAME"
+
```
takameyer, line 52:
```suggestion
```

packages/realm/RealmJS.podspec
```diff
@@ -50,21 +55,42 @@ Pod::Spec.new do |s|
 
   s.pod_target_xcconfig    = {
                                 # Setting up clang
-                                'CLANG_CXX_LANGUAGE_STANDARD' => 'c++17',
+                                'CLANG_CXX_LANGUAGE_STANDARD' => 'c++20',
```
kraenhansen, line 58: We're using C++20 features in the binding.

PR #6400, opened 2024-01-22
Refactor Android builds for React Native

takameyer:
## What, How & Why?
* Flexible NDK version compiling
* Use the consumers React Native version (deduced from Maven)
* Breakpoints in core code (for internal debugging)

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


packages/realm/binding/android/CMakeLists.txt
```diff
@@ -1,7 +1,5 @@
 cmake_minimum_required(VERSION 3.18.1)
 
-set(REACT_NATIVE_VERSION "0.71.0-rc.5")
```
kraenhansen, line 3: I believe we could simplify by leveragin ggit`find_package`, `find_library` to avoid the entire "Extract .so files from the React Native AAR" block below.

To my knowledge, we currently need `jsi` and `fbjni` headers. We could draw some inspiration from https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/jni/CMakeLists.txt#L43-L46.

The React Native packages can refer to these via `target_link_libraries`, see https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/jni/react/runtime/hermes/jni/CMakeLists.txt#L27

takameyer:
I'd say this is ready to review.  I've used `npm pack` locally and tested on minor versions of RN >= 0.71.0.
CI should be green if it gets past the flakes.

packages/realm/package.json
```diff
@@ -54,6 +54,7 @@
     "index.react-native.js",
     "types.d.cts",
     "react-native/android",
+    "!react-native/android/.cxx",
```
kraenhansen, line 57: I wonder if we could make the glob just above a bit more specific instead? Also, it feels a bit strange to mix this and an .npmignore file 🤔

kraenhansen (approved):

packages/realm/package.json
takameyer: Specifically this way for creating the package.  This can probably go away, since it's handled by the package.json

takameyer: I'll remove the .npmignore file.  I also specified the glob:
```
    "react-native/android/**/*{.gradle,.java,.xml}",
```

takameyer:
Merging into the iOS branch to make this easier to handle

PR #6401, opened 2024-01-22
Remove `node-fetch` dependency from SDK, introduce `@realm/fetch` and allow passing `fetch` through `AppConfiguration`

### LINKED ISSUE #6393: Remove dependencies on `node-fetch` and use Node's internal `fetch` + `AbortSignal` instead
We could most likely remove the dependency on `node-fetch` and instead rely on the [`fetch`](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch) and [`AbortSignal`](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#static-method-abortsignaltimeoutdelay) provided by Node.js from v18.

kraenhansen:
## What, How & Why?

This fixes #6393 and makes #6367 obsolete as it removes the `node-fetch` dependency, in favour of depending on the `fetch` provided by Node.js at v18 and above.

It also replaces `@realm/network-transport` with a private (not to be published on NPM) `@realm/fetch` package. The new package exports types reflecting the lowest common API for fetch across our supported platforms and adds platform specific polyfills to the fetch API. This will help ease exposing hooks for end-users to provide their own fetch implementation (ensuring it's easier to support a use-case like that outlined in #6159).

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


packages/fetch/src/types.ts
```diff
@@ -16,182 +16,174 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-export type Method = "GET" | "POST" | "DELETE" | "PUT" | "PATCH";
+export type ResponseType = "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect";
+export type RequestCredentials = "include" | "omit" | "same-origin";
+export type RequestMode = "cors" | "navigate" | "no-cors" | "same-origin";
 
-export type Headers = { [name: string]: string };
+export type RequestBody = ReadableStream | string | null; // TODO: Add TypedArray
```
takameyer, line 23: What's it take to get a `TypedArray` in here?

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,174 @@
...
-  readonly signal: AbortSignal;
-  /**
-   * Invoking this method will set this object's AbortSignal's aborted flag and signal to any observers that the associated activity is to be aborted.
-   */
-  abort(): void;
-};
+  /*
+  static error(): Response;
+  static json(data: any, init?: ResponseInit): Response;
+  static redirect(url: string | URL, status: ResponseRedirectStatus): Response;
```
takameyer, line 95: Remove this?

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,174 @@
...
+  getReader(options: { mode: "byob" }): unknown;
 
   /**
    * Cancel is used when you've completely finished with the stream and don't need any more data from it, even if there are chunks enqueued waiting to be read.
    * @return a Promise that resolves when the stream is canceled.
    * @see https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel
    */
   cancel: (reason?: string) => Promise<void>;
-} & AsyncIterable<Uint8Array>;
+}; // & AsyncIterable<T>;
```
takameyer, line 119:
```suggestion
};
```

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,174 @@
...
-  closed: Promise<boolean>;
+  closed: Promise<void>;
   /**
    * Cancel is used when you've completely finished with the stream and don't need any more data from it, even if there are chunks enqueued waiting to be read.
    * @param reason A human-readable reason for the cancellation.
    * @returns a Promise that resolves when the stream is canceled. Calling this method signals a loss of interest in the stream by a consumer.
    * @see https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/cancel
    */
-  cancel(reason?: string): Promise<string | undefined>;
+  cancel(reason?: string): Promise<unknown>; // TODO: Should have returned Promise<string | undefined, but node types conflict
```
takameyer, line 147: Is there an open issue for this?

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,174 @@
...
-  readonly redirected: boolean;
-  readonly status: number;
-  readonly statusText: string;
-  readonly type: unknown;
-  readonly url: string;
-  clone(): FetchResponse;
-}
-
-export interface FetchRequestInit<RequestBody = unknown> {
+}; //  & AsyncIterable<Uint8Array>;
```
takameyer, line 158:
```suggestion
};
```

takameyer (approved):
Nice work here!

packages/fetch/src/types.ts
kraenhansen: Defining the `TypedArray` I think. AFAIK it's a web standard.
But .. we're not actively using this from the SDK, so I didn't want to bother.

packages/fetch/src/types.ts
kraenhansen: I'd like to keep it in to signal parts of the spec, which is explicitly left out.

takameyer: Perhaps a comment explaining that would be helpful

packages/fetch/src/types.ts
kraenhansen: Again, this is explicitly left out (not simply forgotten), [if the Node types start implementing this](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1677), we can add it back in - this comment might serve as a reminder to do just that.

kraenhansen:
> Tests are failing, but seems unrelated?

The electron tests we're broken, since it didn't support ESM 🤦
Except ... since v28 it does https://www.electronjs.org/docs/latest/tutorial/esm 🎉

packages/fetch/src/types.ts
takameyer: Perhaps an explicit comment would be helpful

kneth (commented):
Can we expose `fetch` on `AppConfiguration` as we do for Realm Web?

kraenhansen:
> Can we expose fetch on AppConfiguration as we do for Realm Web?

It turned out to be pretty simple 👍 https://github.com/realm/realm-js/pull/6401/commits/6f432ab4e30dbb02becfbe62e3f91b7e09681739

packages/fetch/src/types.ts
kraenhansen: Om wrong, they were introduced in es2017.

kraenhansen:
@takameyer I've requested review since I've added functionality since your initial review.

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,177 @@
...
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+export type AnyFetch = typeof fetch<any, any, any, any>;
+
+export interface RequestInit<
+  PlatformRequestBody = RequestBody,
+  PlatformHeaders = Headers,
+  PlatformAbortSignal = AbortSignal,
+> {
   /**
    * A BodyInit object or null to set request's body.
```
elle-j, line 179: The docs of these properties describe the allowed types, but after these updates some seem to have changed. Describing the types (e.g. `a string`, `an object`) in the docs (as compared to only via the TS types) could make the docs and the actual allowed types out of sync. E.g. for this one, I believe also `ReadableStream` and `string` are allowed.
```suggestion
   * The request's body.
```

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,177 @@
...
    */
-  body?: RequestBody;
+  body?: PlatformRequestBody;
   /**
    * A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials.
    */
-  credentials?: FetchRequestCredentials;
+  credentials?: RequestCredentials;
   /**
    * A Headers object, an object literal, or an array of two-item arrays to set request's headers.
```
elle-j, line 187: Similar comment as I wrote above. The docs for this type also refers to the previously allowed `string[][]`.
```suggestion
   * The request's headers.
```

packages/realm-app-importer/src/AdminApiClient.ts
```diff
@@ -379,14 +380,11 @@ export class AdminApiClient {
       if (typeof body === "object") {
         headers["content-type"] = "application/json";
       }
-      const response = await DefaultNetworkTransport.fetch(url, {
+      const response = await fetch(url, {
         ...rest,
-        body: JSON.stringify(body),
+        body: body ? JSON.stringify(body) : undefined,
```
elle-j, line 385: Nit: Accounting for if we want empty string, `0`, and `null` responses to not be interpreted as `undefined`.
```suggestion
        body: body === undefined ? undefined : JSON.stringify(body),
```

packages/realm-web/src/App.ts
```diff
@@ -152,13 +152,13 @@ export class App<
       this._locationUrl = Promise.resolve(this.baseUrl);
     }
     this.localApp = configuration.app;
-    const { storage, transport = new DefaultNetworkTransport() } = configuration;
+    const { storage } = configuration;
```
elle-j, line 155: Looks like `storage` is only used once here, could use `configuration.storage` directly on line 166 if you want.

packages/realm-web/src/Fetcher.ts
```diff
@@ -16,41 +16,46 @@
...
-  read<T>(): Promise<{ value: T | undefined; done: boolean }>;
-  releaseLock(): void;
-};
-type ReadableStream = { getReader(): StreamReader; cancel: () => void };
+type SimpleObject = Record<string, unknown>;
 
 /**
- * @param body A possible resonse body.
+ * @param body A possible resonse.
+ * @param body.body A possible resonse body.
```
elle-j, line 36:
```suggestion
 * @param response A possible response.
 * @param response.body A possible response body.
```

elle-j (approved):
Great replacement here, interesting walkthru and PR!

integration-tests/tests/src/node/node-fetch.ts
```diff
@@ -16,15 +16,20 @@
...
+describe.skipIf(environment.missingServer, "passing node-fetch to AppConfiguration", () => {
+  importAppBefore(buildAppConfig().anonAuth());
 
-DefaultNetworkTransport.fetch = fetch as Fetch;
-DefaultNetworkTransport.AbortController = NodeAbortController as AbortController;
-DefaultNetworkTransport.transformResponse = makeRequestBodyIterable;
+  it("is supported", async function (this: AppContext) {
+    const app = new Realm.App({ id: this.app.id, baseUrl, fetch: nodeFetch });
+    const user = await app.logIn(Realm.Credentials.anonymous());
+    expect(typeof user.id).equals("string");
```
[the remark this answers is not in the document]
kraenhansen: I could probably just await the login, instead of actually expecting anything of the result 🤔

The test doesn't actually verify that the `fetch` provided will be used, I added another (more general) test for that. The main purpose of this is actually that I wanted to test that the type for the `fetch` option would accept the `node-fetch` export.

packages/realm-app-importer/src/AdminApiClient.ts
kraenhansen: The main use-case for this is actually to avoid calling stringify for empty strings (apparently cores way of signalling a missing body) and `undefined`.

packages/fetch/src/types.ts
kraenhansen:
```suggestion
}; // & AsyncIterable<T>;
// Should be [AsyncIterable](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration), but not all our supported runtimes implement this.
```

packages/fetch/src/types.ts
```diff
@@ -16,182 +16,177 @@
...
-   * Returns the AbortSignal object associated with this object.
-   */
-  readonly signal: AbortSignal;
-  /**
-   * Invoking this method will set this object's AbortSignal's aborted flag and signal to any observers that the associated activity is to be aborted.
-   */
-  abort(): void;
-};
+  /*
+  static error(): Response;
```
kraenhansen, line 93:
```suggestion
  // Not all our supported runtimes implement these statics
  // see https://developer.mozilla.org/en-US/docs/Web/API/Response

  static error(): Response;
```

integration-tests/tests/src/node/node-fetch.ts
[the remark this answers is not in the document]
elle-j: Yup I agree, the expect is likely not necessary for the test 👍

packages/fetch/src/types.ts
kraenhansen: I've comitted a comment linking to the docs.

integration-tests/tests/src/node/node-fetch.ts
kraenhansen, line 33:
```suggestion
    await app.logIn(Realm.Credentials.anonymous());
```

packages/realm-app-importer/src/AdminApiClient.ts
elle-j: Thanks for the clarification 👍 I like your suggestion 🙂

packages/fetch/src/types.ts
kraenhansen: Actually ... it seems [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel#return_value) is not following [the specification](https://streams.spec.whatwg.org/#readable-stream-cancel).

I'll have to report an issue with them, but for now I've just made this `Promise<void>`.

takameyer (approved):
Ship it! :shipit:

kneth (approved):

PR #6402, opened 2024-01-22
Update `build-android.js` to use the `ANDROID_HOME` environment variable

kraenhansen:
## What, How & Why?

According to https://developer.android.com/tools/variables `ANDROID_SDK_ROOT` is deprecated.


takameyer:
Can this be merged into https://github.com/realm/realm-js/pull/6400 ? Or does this need to be in `main` right now?

kraenhansen:
I updated my local environment to match the recommended, which is why I'd like it to merge soon 😊

takameyer (approved):

PR #6405, opened 2024-01-23
Upgrade to Realm Core v13.26.0

### LINKED ISSUE #6403: Upgrade to Realm Core 13.26.0

kneth:
## What, How & Why?

This is an upgrade of Realm Core. I have worked around the breaking change related to the implementation of `new Realm.App()` . We might want to discuss if we would like to expose app caching at a later time.

The new base URL functionality isn't exposed yet. It will be part of roaming project (yet to be decided).

Moreover, `spec.yml` needed a [small update](https://github.com/realm/realm-core/pull/7284) so the upgrade isn't as clean as I would like it.

This closes #6403 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
The changes seem fine, but I suspect the `@realm/react` unit tests caused a crash.

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -48,6 +48,7 @@ export function importAppBefore(config: AppConfig | { config: AppConfig }, sdkCo
     } else {
       const { appId, baseUrl } = await importApp(config);
       this.app = new Realm.App({ id: appId, baseUrl, ...sdkConfig });
+      this.baseUrl = baseUrl;
```
kraenhansen, line 51: I would probably just import this directly in the test, but if you want to store this in the context, it should probably be added to the `AppContext` type as well.

integration-tests/tests/src/tests/sync/user.ts
```diff
@@ -677,9 +678,10 @@ describe.skipIf(environment.missingServer, "User", () => {
       const credentials = Realm.Credentials.anonymous();
       const user = await this.app.logIn(credentials);
       const appId = this.app.id;
+
       //@ts-expect-error Wanting to prove that a completely new app instance will still return the current logged in user
       delete this.app;
-      this.app = new Realm.App({ id: appId, baseUrl: this.baseUrl });
+      this.app = new Realm.App({ id: appId, baseUrl: baseUrl });
```
kraenhansen, line 684:
```suggestion
      this.app = new Realm.App({ id: appId, baseUrl });
```

kraenhansen:
Have you had a chance to look into the crash from the `@realm/react` tests? It looks pretty deterministic.

packages/realm/bindgen/js_opt_in_spec.yml
elle-j, line 1: I read the PR description, but from the Core [spec.yml changes](https://github.com/realm/realm-core/compare/7227d6a447821c28895daa099b6c7cd4c99d461b...dcab5907983e4441de501c97ac33962b6fc85ca1#diff-9d9d60c03d9a60ce8f2a49a0bd07324beae03010bba8b280bb697fb38b0be09f) it does look like the base URL related changes are also there.

elle-j (commented):
Nice upgrade! Just have a comment about the base URL changes that would be good to address 🙂

packages/realm/bindgen/js_opt_in_spec.yml
kneth: My impression is that it should be done later but I might be wrong.

kneth:
> Have you had a chance to look into the crash from the @realm/react tests?

It is `someurl` in https://github.com/realm/realm-js/blob/main/packages/realm-react/src/__tests__/AppProvider.test.tsx#L30 - it is a malformed URL, and modifying it to `http://someurl` makes the test pass.

packages/realm/bindgen/js_opt_in_spec.yml
elle-j: Ah I see now that `base_url` was removed from the `App` and not `AppConfig` in the spec 👍

CHANGELOG.md
```diff
@@ -7,18 +7,18 @@
 * Added an optional `fetch` parameter to the `AppConfiguration`. Use this to specify a custom implementation of the `fetch` function used by the app to perform network requests.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None
+* Handle `EOPNOTSUPP` when using `posix_fallocate()` and fallback to manually consume space. This should enable android users to open a Realm on restrictive file systems. ([#6349](https://github.com/realm/realm-js/issues/6349), since v12.4.0)
```
elle-j, line 10: In the issue is said that it happened after upgrading to `v12.2.0`.

elle-j (approved):

CHANGELOG.md
kneth: You have a point, and the core change log actually says "... also an issue in some platforms before this".

kraenhansen:
I've created https://github.com/realm/realm-js/issues/6441 to track fixing the crash 👍

kraenhansen (approved):

PR #6407, opened 2024-01-25
Remove `fs-extra`

kraenhansen:
## What, How & Why?

As an alternative to #6373, this removes all use of the `fs-extra` package.
We weren't actually using it in ways that couldn't easily be replaced by `node:fs`, so I think we should just avoid it to simplify things and get one less dependency.

## In draft

It turns out that since `exists` doesn't return a `Promise` when called from `node:fs`, the libraries cannot be easily replaced, I think we should rethink the way we provide `fs` and `path` implementations to the tests: https://github.com/realm/realm-js/issues/6408

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/node/analytics.ts
+++ b/integration-tests/tests/src/node/analytics.ts
@@ -24,7 +24,7 @@ import { fileURLToPath } from "url";
 import { expect } from "chai";
 import { collectPlatformData } from "realm/scripts/submit-analytics";
 
-import fse from "fs-extra";
+import fs from "fs";
 
 // emulate old __dirname: https://flaviocopes.com/fix-dirname-not-defined-es-module-scope/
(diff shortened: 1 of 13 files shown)


integration-tests/tests/src/node/analytics.ts
```diff
@@ -39,7 +39,7 @@ describe("Analytics", () => {
 
   function getRealmVersion() {
     const packageJsonPath = path.resolve(__dirname, "../../../../packages/realm/package.json");
-    const packageJson = fse.readJsonSync(packageJsonPath);
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
```
kraenhansen, line 42: This was the only actual use of an API that isn't provided by `node:fs`.

PR #6416, opened 2024-01-26
Upgrade to core's `next-major`

kneth:
## What, How & Why?

This closes # ??? 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


packages/realm/src/Realm.ts
```diff
@@ -284,24 +290,67 @@ export class Realm {
...
    * @param loggerCallback - The callback invoked by the logger. The default callback uses `console.log`, `console.warn` and `console.error`, depending on the level of the message.
-   * @note The logger callback needs to be setup before opening the first realm.
+   * @note The logger callback needs to be setup before opening the first Realm.
    * @since 12.0.0
+   * @example
+   * Realm.setLogger(({ category, level, message }) => {
+   *   console.log(`[${category} - ${level}] ${message}`);
+   * });
    */
   static setLogger(loggerCallback: LoggerCallback) {
```
elle-j, line 337: An alternative here is also to modify the `LoggerCallback` to e.g.:
```ts
export type Logger = (level: NumericLogLevel, message: string, category?: LogCategory) => void;
```

Then this function could be:
```ts
  static setLogger(loggerCallback: LoggerCallback) {
    const logger = binding.Helpers.makeLogger((category, level, message) => {
      loggerCallback(fromBindingLoggerLevelToLogLevel(level), message, category as LogCategory);
    });
    binding.Logger.setDefaultLogger(logger);
  }
```

elle-j (commented):
Thanks for taking this on! Here are some initial comments 🙂

packages/realm/src/Realm.ts
```diff
@@ -284,24 +290,67 @@ export class Realm {
...
+      } else {
+        const level = arg;
+        Object.values(LogCategory).forEach((category) => {
+          Realm.setLogLevel({ level, category });
+        });
+      }
+    } else {
+      throw new Error(`Wrong number of arguments - expected 1, got ${arguments.length}`);
+    }
   }
```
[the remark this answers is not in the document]
elle-j: I like that reasoning 🙂 One simplification could then be:
```ts
  static setLogLevel(options: LogLevel | { category: LogCategory; level: LogLevel }) {
    const { level, category } = typeof options === "string" ? { level: options, category: LogCategory.Realm } : options;
    binding.LogCategoryRef.getCategory(category).setDefaultLevelThreshold(toBindingLoggerLevel(level));
  }
```

Or using the argument type `{ category: LogCategory; level: LogLevel }` from a named type (e.g. `type LogOptions = { category: LogCategory; level: LogLevel };`) like you're doing.

packages/realm/src/Realm.ts
```diff
@@ -284,24 +290,75 @@ export class Realm {
...
+      logger = binding.Helpers.makeLogger((_category, level, message) => {
+        cb(fromBindingLoggerLevelToLogLevel(level), message);
+      });
+    } else {
+      const cb = loggerCallback as LoggerCallback2;
+      logger = binding.Helpers.makeLogger((category, level, message) => {
+        cb({ category: category as LogCategory, level: fromBindingLoggerLevelToLogLevel(level), message });
+      });
+    }
     binding.Logger.setDefaultLogger(logger);
```
elle-j, line 362: Nice check on the function's length 🙂

To prevent the misuse of `(...args)` as you mentioned, and to remove the need for `as LoggerCallback1/2`, we could throw for the wrong number of callback args and also extract the if-conditions into two functions in e.g. `Logger.ts` (with whatever names you think are suitable):

```ts
export function isLoggerWithLevel(logger: unknown): logger is LoggerCallback1 {
  return typeof logger === "function" && logger.length === 2;
}

export function isLoggerWithOptions(logger: unknown): logger is LoggerCallback2 {
  return typeof logger === "function" && logger.length === 1;
}
```

Then the `setLogger()` function could utilize those:

```ts
  static setLogger(loggerCallback: LoggerCallback) {
    let logger: binding.Logger;

    if (isLoggerWithLevel(loggerCallback)) {
      logger = binding.Helpers.makeLogger((_, level, message) => {
        loggerCallback(fromBindingLoggerLevelToLogLevel(level), message);
      });
    } else if (isLoggerWithOptions(loggerCallback)) {
      logger = binding.Helpers.makeLogger((category, level, message) => {
        loggerCallback({
          category: category as LogCategory,
          level: fromBindingLoggerLevelToLogLevel(level),
          message,
        });
      });
    } else {
      throw new Error(`Unexpected logger passed.`);
    }

    binding.Logger.setDefaultLogger(logger);
  }
```

packages/realm/src/app-services/Sync.ts
```diff
@@ -60,7 +60,7 @@ export class Sync {
 
   /** @deprecated Will be removed in v13.0.0. Please use {@link Realm.setLogger}. */
   static setLogger(app: App, logger: Logger) {
-    const factory = binding.Helpers.makeLoggerFactory((level, message) => {
+    const factory = binding.Helpers.makeLoggerFactory((category, level, message) => {
```
elle-j, line 63: The logger received to `Sync.setLogger()` is currently (in our TS types) the one taking the numeric log level (thus, same as before). Core hasn't changed that API right?
```suggestion
    const factory = binding.Helpers.makeLoggerFactory((_, level, message) => {
```

elle-j (commented):
Nice updates! 🙂 Left some ideas and suggestions.

packages/realm/src/Realm.ts
kraenhansen: I like the suggestion to use of an assert function, but it would be even greater if it used a pure function (`loggerFromCallback` or something) returning the `logger` instead of a local mutable variable 😉

packages/realm/src/Logger.ts
```diff
@@ -20,18 +20,137 @@ import { assert, binding } from "./internal";
...
+  Server = "Realm.Sync.Server",
+  /**
+   * Log activity at the Atlas App level.
+   */
+  App = "Realm.App",
+  /**
+   * Log activity at the SDK level.
+   */
+  SDK = "Realm.SDK",
+}
```
kraenhansen, line 144: I feel the flattening of the namespace here is a bit misleading and would potentially result in future problems if two categories start using the same sub-category names. I.e. what would happen if core introduce a  `Realm.App.Network` category?

I'd rather that we keep this close to the Core API, use a string-with-dot notation and type this as a union of strings instead:

```suggestion
export type LogCategory = "Realm" | "Realm.Storage" | "Realm.Storage.Transaction" /* ... */;
```

The only drawback is that it doesn't seem to be possible [to document this using a tsdoc string](https://github.com/microsoft/tsdoc/issues/164). But we could probably document the different values on the `LogCategory` type itself or the individual methods taking the value 🤔

packages/realm/src/Realm.ts
```diff
@@ -284,24 +290,75 @@ export class Realm {
...
+
+    const setLevel = (category: LogCategory, level: LogLevel) => {
+      assert(Object.values(LogCategory).includes(category));
+      const ref = binding.LogCategoryRef;
+      const c = ref.getCategory(category);
+      c.setDefaultLevelThreshold(toBindingLoggerLevel(level));
+    };
+
+    // FIXME: don't use `arguments` but find a proper type
+    if (arguments.length === 1) {
```
kraenhansen, line 311:
```suggestion
    if (typeof _ !== "string") {
```

packages/realm/src/Realm.ts
```diff
@@ -284,24 +290,75 @@ export class Realm {
...
    * @note The log level can be changed during the lifetime of the application.
    * @since 12.0.0
+   * @example
+   * Realm.setLogLevel({ category: LogCategory.Realm, level: "all" });
    */
-  static setLogLevel(level: LogLevel) {
-    const bindingLoggerLevel = toBindingLoggerLevel(level);
-    binding.Logger.setDefaultLevelThreshold(bindingLoggerLevel);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  static setLogLevel(_: LogLevel | LogArgs) {
```
kraenhansen, line 300: I'd suggest a better name and referring to it directly instead of via `arguments[0]`.

```suggestion
  static setLogLevel(argument: LogLevel | LogArgs) {
```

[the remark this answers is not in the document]
kraenhansen: I agree ☝️ is a more clean solution.

integration-tests/tests/src/tests/shared-realms.ts
```diff
@@ -26,16 +26,19 @@ describe("SharedRealm operations", () => {
...
-        logs.push({ level, message });
-      });
+      const logger = (args: LoggerCallbackArgs) => {
+        const { category, level, message } = args;
+        logs.push({ category, level, message });
+      };
 
-      Realm.setLogLevel("all");
+      Realm.setLogger(logger);
+      Realm.setLogLevel({ category: LogCategory.Realm, level: "all" });
```
kraenhansen, line 41: If I understand the proposed changes correctly, simply passing a level to `setLogLevel` isn't deprecated, in which case I'd suggest leaving the tests as is and adding new tests for the new behaviour.

packages/realm/src/Logger.ts
```diff
@@ -40,7 +159,30 @@ export enum NumericLogLevel {
...
+/**
+ * A callback passed to `Realm.setLogger`. Arguments are passed as a POJO.
+ *
+ * @param category   - The category (origin) of the log entry.
+ * @param level      - The level of the log entry.
+ * @param message    - The message of the log entry.
+ * @since
+ */
+export type LoggerCallback2 = (args: LoggerCallbackArgs) => void;
+export type LoggerCallback = LoggerCallback1 | LoggerCallback2;
```
kraenhansen, line 185: To be honest, I'm actually not convinced the change in Core API justifies adding an overload to the callback passed to `setLogger` 🤔 Especially because we need to rely on `arguments.length` to figure out which callback the user has passed ...

Is passing the category as a third positional argument really _that_ awful?

```js
Realm.setLogger((level, message, category) => {
  // ...
});
```

PR #6437, opened 2024-01-28
"Received message ... for session iden {n} when that session never existed" bug

kraenhansen:
## What, How & Why?

This is an attempt to reliably reproduce an issue we've been experiencing on CI.


kraenhansen:
No longer relevant.

PR #6438, opened 2024-01-28
Update `pr-coverage.yml` to run only on changes to sourcefiles on PRs

kraenhansen:
## What, How & Why?

I believe it's only interesting to run this when PRs are opened/updated targeting source-files, instead of any push to any branch.


kneth (approved):
Using less of a shared resource is always a good idea

Issue #6447, opened 2024-01-30
User function arguments is wrapped in an Array


deckyfx:
### How frequently does the bug occur?

Always

### Description

Breaking issue in 12.6.0
User function argument sent to Atlas is wrapped in an Array 

the problem is in this file, (idk where is this issue in the source repo)
./node_modules/realm/dist/bundle.react-native.js
line ~7600 

```typescript
callFunction(name, ...args) {
    return this.callFunctionOnService(name, undefined, args);
}
```

As you can see, the function received spread variable arguments, but then you passed it to `callFunctionOnService` as one array variable

### Can you reproduce the bug?

Always

### Reproduction Steps

Just install realm 12.6.0 in a react-native project

### Version

12.6.0

### What services are you using?

Atlas App Services: Functions or GraphQL or DataAPI etc

### Are you using encryption?

No

### Platform OS and version(s)

WSL Ubuntu on Windows

### Build environment

Which debugger for React Native: ..
none

PR #6450, opened 2024-01-31
revert arguments when call callFunctionOnService to spread variable

deckyfx:
## What, How & Why?

in version 12+, the User function called from the client always sends arguments wrapped in Array,
I compare the code with version 11+ and found out that function `callFunction` pass the `args` to next function `callFunctionOnService` as a single array argument instead of spread arguments

This should resolve [#6447 ](https://github.com/realm/realm-js/issues/6447)

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/src/app-services/User.ts
+++ b/packages/realm/src/app-services/User.ts
@@ -289,7 +289,7 @@ export class User<
    * await doThing(a2);
    */
   callFunction(name: string, ...args: unknown[]): Promise<unknown> {
-    return this.callFunctionOnService(name, undefined, args);
+    return this.callFunctionOnService(name, undefined, ...args);
   }
 
   /** @internal */


kraenhansen (approved):
Thanks for catching this and providing a PR, ideally, I would love if we could also add a test to ensure this doesn't regress.

kraenhansen:
We do seem to have tests for this, so I'll have to look into why they're not failing https://github.com/realm/realm-js/blob/main/integration-tests/tests/src/tests/sync/user.ts#L636-L672

kraenhansen:
Unfortunately, I wasn't able to pick your change into my branch, but I updated the test and implementation in https://github.com/realm/realm-js/pull/6451 and as such I'll close this PR.

You'll be rightfully attributed in the changelog and release notes. Thanks again!

deckyfx:
np, as long as this issue is resolved, thanks for your attention <3

PR #6451, opened 2024-01-31
Fixed `User#callFunction` arguments, passing them correctly to the server

### LINKED ISSUE #6447: User function arguments is wrapped in an Array

kraenhansen:
## What, How & Why?

This closes #6447 by correctly spreading in the `args` when calling `callFunctionOnService`.

I also fixed the test that was passing, because we had wrapped the function being tested in a `parseInt`, which sadly coerced the `[123]` into `123`, making the test pass although the code was broken.

Credit for discovering and proposing a fix for the bug goes to @deckyfx - thanks again!

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm/src/app-services/utils.ts
+++ b/packages/realm/src/app-services/utils.ts
@@ -16,11 +16,15 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+import { binding } from "src/internal";
+
 /**
  * Remove entries for undefined property values.
  * @internal
  */
-export function cleanArguments(args: unknown[] | unknown): unknown[] | unknown {
+export function cleanArguments(args: unknown[]): binding.EJson[];
+export function cleanArguments(args: unknown): binding.EJson;
+export function cleanArguments(args: unknown[] | unknown) {
   if (Array.isArray(args)) {
     // Note: `undefined` elements in the array is not removed.
     return args.map(cleanArguments);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@
 * None
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


packages/realm/src/app-services/utils.ts
```diff
@@ -16,11 +16,15 @@
...
+import { binding } from "src/internal";
+
 /**
  * Remove entries for undefined property values.
  * @internal
  */
-export function cleanArguments(args: unknown[] | unknown): unknown[] | unknown {
+export function cleanArguments(args: unknown[]): binding.EJson[];
+export function cleanArguments(args: unknown): binding.EJson;
+export function cleanArguments(args: unknown[] | unknown) {
```
kraenhansen, line 27: While I was at it, I updated the signature of this, to avoid type assertions on the calling side.

kneth (approved):

PR #6457, opened 2024-02-02
Disable eslint plugin (no-duplicate-values-in-enum) as Core's PropertyType make use of it.

kneth:
## What, How & Why?

A recent upgrade of ESLint is preventing us from building. The root cause is an [enum in Realm Core with a duplicate value](https://github.com/realm/realm-core/blob/master/src/realm/object-store/property.hpp#L42-L67), and the ESLint plugin was enabled in [5.22.0](https://github.com/typescript-eslint/typescript-eslint/blob/v6.0.0/CHANGELOG.md#5220-2022-05-02)

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/.eslintrc.json
+++ b/packages/realm/.eslintrc.json
@@ -6,6 +6,7 @@
   ],
   "rules": {
     "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-duplicate-enum-values": "off",
     "no-console": "error",
     "sort-imports": ["warn", { "ignoreDeclarationSort": true }],
     "jsdoc/check-tag-names": "off",


elle-j (approved):
LGTM 👍 

As note for other reviewers: One of the duplicates (`Required`) is not used in the SDK.

packages/realm/.eslintrc.json
```diff
@@ -6,6 +6,7 @@
   ],
   "rules": {
     "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-duplicate-enum-values": "off",
```
kraenhansen, line 9: It would be nice if we had an eslint config file that was scoped to the generated source code, but .. this will work for now.

PR #6494, opened 2024-02-19
Improve running tests when server is missing

kraenhansen:
## What, How & Why?

I want to improve the default experience of running our tests.
Unfortunately, contributors to the project isn't allowed to pull the docker image and run the test server, but that shouldn't keep them from running tests at all. In its current form, its a subtly documented feature of the test runner that you can run with `missingServer` context, but I want to make this more obvious and improve the default experience (when the server is missing).

If the environment does not explicitly provide a `baseUrl` nor set `missingServer=false` the tests using `importAppBefore` will skip if the app import fails due to a connection refusal. If this happens, a warning will be printed after all tests has ran:

Screenshot 2024-02-19 at 11 38 53

- This PR also renames `realmBaseUrl` to `baseUrl` since theres only ever one base url, so there's no need to disambiguate it and the latter is the name used in the code anyway.
- It also simplifies how tests are skipped when ran with `missingServer`, to include this check in the `importAppBefore` hook as they were (almost) always used together anyway and could easily be forgotten.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
I'll merge to improve our DX tomorrow 🙈 Feel free to review in retrospect (or not 😊).

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -34,20 +46,95 @@ export type AppConfigurationRelaxed = {
...
+  } else {
+    return {
+      kind: "username-password",
+      username,
+      password,
+    };
+  }
+}
+
+const credentials = getCredentials();
```
elle-j, line 65: Looks like the function is only used once, so it may not need to be a function.
```suggestion
const credentials: Credentials =
  typeof publicKey === "string" && typeof privateKey === "string"
    ? { kind: "api-key", publicKey, privateKey }
    : { kind: "username-password", username, password };
```

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -34,20 +46,95 @@ export type AppConfigurationRelaxed = {
...
+function isConnectionRefused(err: unknown) {
+  return (
+    err instanceof Error &&
+    err.cause instanceof AggregateError &&
+    "code" in err.cause &&
+    err.cause.code === "ECONNREFUSED"
+  );
+}
+
+function printWarningBox(...lines: string[]) {
```
elle-j, line 82: We may want to utilize this elsewhere as well, could we move this into `/utils`? 🙂

integration-tests/tests/src/tests/sync/logging.ts
```diff
@@ -0,0 +1,48 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import Realm from "realm";
+import { importAppBefore } from "../../hooks";
+import { buildAppConfig } from "../../utils/build-app-config";
+
+describe("Logging", () => {
+  importAppBefore(buildAppConfig("with-pbs").anonAuth().flexibleSync());
+  afterEach(() => Realm.clearTestState());
+  // Skipped because reusing a single app across tests break this
```
elle-j, line 26:
```suggestion
```

elle-j (commented):
Nice DX improvement 👍

PR #6508, opened 2024-02-26
Improving documentation for `realm`

kraenhansen:
## What, How & Why?

After #6492 we're able to generate more clean documentation.
NOTE: This PR also introduce exports of a couple of types that we haven't historically exported, but was referenced by other public APIs.

### DIFF
--- a/packages/realm/src/Realm.ts
+++ b/packages/realm/src/Realm.ts
@@ -171,10 +171,18 @@ export class Realm {
 
   /**
    * Checks if the Realm already exists on disk.
-   * @param arg - The configuration for the Realm or the path to it.
+   * @param path - The path for a Realm.
+   * @throws An {@link Error} if anything in the provided {@link path} is invalid.
+   * @returns `true` if the Realm exists on the device, `false` if not.
+   */
+  public static exists(path: string): boolean;
+  /**
+   * Checks if the Realm already exists on disk.
+   * @param config - The configuration of a Realm.
    * @throws An {@link Error} if anything in the provided {@link config} is invalid.
    * @returns `true` if the Realm exists on the device, `false` if not.
    */
+  public static exists(config: Configuration): boolean;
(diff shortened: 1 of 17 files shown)


kneth (approved):

kraenhansen:
This failure seems unrelated.

packages/realm/src/Realm.ts
```diff
@@ -1271,6 +1317,10 @@ export namespace Realm {
   export import ObjectClass = internal.RealmObjectConstructor;
   /** @deprecated Will be removed in v13.0.0. Please use {@link internal.PropertyTypeName} */
   export import PropertyType = internal.PropertyTypeName;
+  /** @deprecated Use the another {@link internal.ClientResetMode} than {@link internal.ClientResetMode.Manual}. */
```
elle-j, line 1320: To show slightly friendlier names for users you can show just the part excluding `internal.` (e.g. as such: [previous comment](https://github.com/realm/realm-js/pull/6492#discussion_r1500544652). (Just readding it because I don't know if you saw the other reply 👍 )

packages/realm/typedoc.json
```diff
@@ -1,12 +1,12 @@
 {
-  "entryPoints": ["src/index.ts"],
-  "tsconfig": "./tsconfig.json",
+  "entryPoints": ["dist/public-types/index.d.ts"],
```
elle-j, line 2: 👍

elle-j (commented):
Thanks for making these updates!

packages/realm/src/Realm.ts
kraenhansen: I wasn't sure why you suggested that - but now it makes perfect sense. Thanks!

kraenhansen:
I've just pushed an update with your feedback: https://github.com/realm/realm-js/commit/0ad84678b9f42eb958924fa3363b4e5d29b6ce48

PR #6512, opened 2024-02-28
RJS-2739: Upgrade to Realm Core v14.1.0

### LINKED ISSUE #6499: Upgrade to Realm Core v14.1.0

kneth:
## What, How & Why?

This closes #6499 

## ☑️ ToDos

* [x] 📝 Changelog entry
* [x] 🚦 Tests


packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -751,6 +751,21 @@ describe("normalizePropertySchema", () => {
...
+        indexed: true,
+      },
+      {
+        type: "list",
+        objectType: "string",
+        indexed: true,
+        optional: false,
+      },
+      { isPrimaryKey: false },
+    );
```
kraenhansen, line 768: Seems fine, but it also seems unrelated to the upgrade.

elle-j: I added this because the upgrade also included being able to index list of strings. So this tests that the parser also allows this.

elle-j (approved):
Nice! 🙂 Left some suggestions [here](https://github.com/realm/realm-js/pull/6520) 👍

packages/realm/src/tests/schema-normalization.test.ts
kneth: It is testing an enhancement mentioned in the changelog.

packages/realm/src/Logger.ts
```diff
@@ -40,7 +167,60 @@ export enum NumericLogLevel {
  */
 export type Logger = (level: NumericLogLevel, message: string) => void;
 
-export type LoggerCallback = (level: LogLevel, message: string) => void;
+/**
+ * A callback passed to `Realm.setLogger`.
```
elle-j, line 171: Rather than tying this in the description to where it is used, I'd suggest only describing what it is (as `Realm.setLogger()` will already be referencing this).
```suggestion
 * A callback to be used as the logger.
```

packages/realm/src/Logger.ts
```diff
@@ -40,7 +167,60 @@ export enum NumericLogLevel {
...
+   */
+  level: LogLevel;
+  /**
+   * The message of the log entry.
+   */
+  message: string;
+};
+
+/**
+ * A callback passed to `Realm.setLogger`. Arguments are passed as an object.
```
elle-j, line 198: I think the argument itself can be explained by the type instead 🙂 
```suggestion
 * A callback to be used as the logger.
```

packages/realm/bindgen/CMakeLists.txt
```diff
@@ -24,17 +24,21 @@ if(DEFINED CMAKE_JS_VERSION)
...
+        add_link_options(-Wl,--warn-execstack)
       elseif(NODE_ARCH STREQUAL "x64")
         set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/vendor/realm-core/tools/cmake/x86_64-linux-gnu.toolchain.cmake")
+        add_link_options(-fuse-ld=gold)
+        add_link_options(-Wl,-z,noexecstack)
+        add_link_options(-Wl,--warn-execstack)
       endif()
-      add_link_options(-fuse-ld=gold)
-      add_link_options(-Wl,-z,noexecstack)
-      add_link_options(-Wl,--warn-execstack)
```
kraenhansen, line 37: Can you share a bit more context on why this wasn't possible / needed for "arm"?

packages/realm/src/Logger.ts
```diff
@@ -40,7 +167,60 @@ export enum NumericLogLevel {
...
+ * @since 12.7.0
+ */
+export type LoggerCallback2 = (entry: LogEntry) => void;
+export type LoggerCallback = LoggerCallback1 | LoggerCallback2;
```
elle-j, line 202:
```suggestion
/**
 * A callback to be used as the logger.
 * @since 12.7.0
 */
export type LoggerCallback = LoggerCallback1 | LoggerCallback2;
```

packages/realm/src/Realm.ts
```diff
@@ -107,27 +111,77 @@ export class Realm {
...
+   * @note The log level can be changed during the lifetime of the application.
+   * @since 12.7.0
+   * @example
+   * Realm.setLogLevel({ category: LogCategory.Realm, level: "all" });
    */
-  static setLogLevel(level: LogLevel) {
-    const bindingLoggerLevel = toBindingLoggerLevel(level);
-    binding.Logger.setDefaultLevelThreshold(bindingLoggerLevel);
+  static setLogLevel(options: LogOptions): void;
+
```
elle-j, line 132:
```suggestion
```

packages/realm/src/Realm.ts
```diff
@@ -107,27 +111,77 @@ export class Realm {
...
+    } else {
+      setLevel(arg.level, arg.category);
+    }
   }
 
   /**
    * Sets the logger callback.
    * @param loggerCallback - The callback invoked by the logger. The default callback uses `console.log`, `console.warn` and `console.error`, depending on the level of the message.
-   * @note The logger callback needs to be setup before opening the first realm.
+   * @note The logger callback needs to be setup before opening the first Realm.
```
elle-j, line 150:
```suggestion
   * @note The logger callback needs to be set up before opening the first Realm.
```

packages/realm/src/Logger.ts
```diff
@@ -20,18 +20,145 @@ import { assert, binding } from "./internal";
...
+  Server = "Realm.Sync.Server",
+  /**
+   * Log activity at the Atlas App level.
+   */
+  App = "Realm.App",
+  /**
+   * Log activity at the SDK level.
+   */
+  SDK = "Realm.SDK",
+}
```
kraenhansen, line 144: I still feel using an enum with the leaf name as key isn't the best API design and would rather that we used a string union here. The main reason being that it's possible for Core to introduce a category that we can't unambiguously express to our users.

I.e. what do we do if core introduces a "Realm.Sync.Server.Changeset" or a "Realm.App.Network" category?

packages/realm/src/Realm.ts
```diff
@@ -107,27 +111,77 @@ export class Realm {
...
+   * @since 12.0.0
+   * @deprecated Pass a callback taking a single object argument instead.
+   * @example
+   * Realm.setLogger((level, message) => {
+   *   console.log(`[${level}] ${message}`);
+   * });
+   */
+  static setLogger(loggerCallback: LoggerCallback1): void;
+
```
elle-j, line 159:
```suggestion
```

packages/realm/src/Realm.ts
```diff
@@ -107,27 +111,77 @@ export class Realm {
...
+  /**
+   * Sets the logger callback.
+   * @param loggerCallback - The callback invoked by the logger. The default callback uses `console.log`, `console.warn` and `console.error`, depending on the level of the message.
+   * @note The logger callback needs to be setup before opening the first Realm.
```
elle-j, line 163:
```suggestion
   * @note The logger callback needs to be set up before opening the first Realm.
```

packages/realm/src/Realm.ts
```diff
@@ -107,27 +111,77 @@ export class Realm {
...
+   * @since 12.0.0
+   * @example
+   * Realm.setLogger(({ category, level, message }) => {
+   *   console.log(`[${category} - ${level}] ${message}`);
+   * });
+   */
+  static setLogger(loggerCallback: LoggerCallback2): void;
+
```
elle-j, line 171:
```suggestion
```

packages/realm/src/Realm.ts
```diff
@@ -107,27 +111,77 @@ export class Realm {
...
+  /**
+   * Sets the logger callback.
+   * @param loggerCallback - The callback invoked by the logger. The default callback uses `console.log`, `console.warn` and `console.error`, depending on the level of the message.
+   * @note The logger callback needs to be setup before opening the first Realm.
```
elle-j, line 175:
```suggestion
   * @note The logger callback needs to be set up before opening the first Realm.
```

packages/realm/src/app-services/Sync.ts
```diff
@@ -46,7 +46,7 @@ export class Sync {
 
   /** @deprecated Will be removed in v13.0.0. Please use {@link Realm.setLogger}. */
   static setLogger(app: App, logger: Logger) {
-    const factory = binding.Helpers.makeLoggerFactory((level, message) => {
+    const factory = binding.Helpers.makeLoggerFactory((category, level, message) => {
```
elle-j, line 49:
```suggestion
    const factory = binding.Helpers.makeLoggerFactory((_, level, message) => {
```

elle-j (commented):
I think it was the right call to remove the `needsFileFormatUpgrade` and possibly add it as a follow-up 👍 

LGTM with some suggested updates to docs. The biggest thing to decide on is whether to change the log categories to a string union.

kraenhansen (approved):

PR #6513, opened 2024-02-28
RJS-2680: Implement support for `Mixed` data type with nested collections

### LINKED ISSUE #6344: Store and retrieve nested lists in Mixed

elle-j:
## What, How & Why?

Adds support for storing nested lists and dictionaries as the underlying value of a `Mixed` property on non-synced realms.

Sets are not supported as a `Mixed` value.

### Review notes

* This PR extends the [flat collections implementation](https://github.com/realm/realm-js/pull/6364).
  * The tests from the previous PR have been slightly reorganized to make both set of tests more cohesive.
  * That may make diffs (especially in the tests) larger without actually having changed much of the previous tests. In these cases, it may be easier to look at the file rather than the diffs, with most new tests having `"nested"` in its name.
* This PR also includes refactoring of some preexisting code to alleviate this and future implementations.
  * TLDR:
    * The JS SDK has used a pattern of injecting getters and setters (computed when the realm is opened) to each collection. The corresponding definitions of those injected functions were sometimes harder to intuitively locate, so they have been centralized to each collection class file.
    * The choice to inject accessors was due to performance reasons, by doing some computations beforehand, rather than during construction of an object or collection.
  * (Injected getters/setters for the top-level Realm Object itself has not been refactored.)
* Corresponding Core PR:
  * https://github.com/realm/realm-core/pull/7392

### Test overview

<details>
<summary>Click to expand</summary>

```
  Mixed
    Collection types
      CRUD operations
        Create and access
          List
            ✓ has all primitive types (input: JS Array)
            ✓ has all primitive types (input: Realm List)
            ✓ has all primitive types (input: Default value)
            ✓ has nested lists of all primitive types
            ✓ has nested dictionaries of all primitive types
            ✓ has mix of nested collections of all types
            ✓ inserts all primitive types via `push()`
            ✓ inserts nested lists of all primitive types via `push()`
            ✓ inserts nested dictionaries of all primitive types via `push()`
            ✓ inserts mix of nested collections of all types via `push()`
            ✓ returns different reference for each access
          Dictionary
            ✓ has all primitive types (input: JS Object)
            ✓ has all primitive types (input: JS Object w/o proto)
            ✓ has all primitive types (input: Realm Dictionary)
            ✓ has all primitive types (input: Default value)
            ✓ can use the spread of embedded Realm object
            ✓ can use the spread of custom non-Realm object
            ✓ has nested lists of all primitive types
            ✓ has nested dictionaries of all primitive types
            ✓ has mix of nested collections of all types
            ✓ inserts all primitive types via setter
            ✓ inserts nested lists of all primitive types via setter
            ✓ inserts nested dictionaries of all primitive types via setter
            ✓ inserts mix of nested collections of all types via setter
            ✓ inserts mix of nested collections of all types via `set()` overloads
            ✓ returns different reference for each access
          Results
            from List
              snapshot()
                ✓ has all primitive types
                ✓ has mix of nested collections of all types
              objects().filtered()
                ✓ has all primitive types
                ✓ has mix of nested collections of all types
            from Dictionary
              objects().filtered()
                ✓ has all primitive types
                ✓ has mix of nested collections of all types
        Update
          List
            ✓ updates top-level item via setter
            ✓ updates nested item via setter
            ✓ updates itself to a new list
            ✓ updates nested list to a new list
            ✓ does not become invalidated when updated to a new list
            - self assigns
            - self assigns nested list
          Dictionary
            ✓ updates top-level entry via setter
            ✓ updates nested entry via setter
            ✓ updates itself to a new dictionary
            ✓ updates nested dictionary to a new dictionary
            ✓ does not become invalidated when updated to a new dictionary
            - self assigns
            - self assigns nested dictionary
        Remove
          List
            ✓ removes top-level item via `remove()`
            ✓ removes nested item via `remove()`
          Dictionary
            ✓ removes top-level entry via `remove()`
            ✓ removes nested entry via `remove()`
        JS collection methods
          List
            ✓ pop()
            ✓ shift()
            ✓ unshift()
            ✓ splice()
            ✓ indexOf()
          Iterators
            ✓ values() - list
            ✓ values() - dictionary
            ✓ entries() - list
            ✓ entries() - dictionary
      Filtering
        ✓ filters by query path on list of all primitive types
        ✓ filters by query path on nested list of all primitive types
        ✓ filters by query path on dictionary of all primitive types
        ✓ filters by query path on nested dictionary of all primitive types
      Invalid operations
        ✓ throws when creating a Mixed with a set
        ✓ throws when creating a set with a list
        ✓ throws when creating a set with a dictionary
        ✓ throws when updating a list item to a set
        ✓ throws when updating a dictionary entry to a set
        ✓ throws when creating a list or dictionary with an embedded object
        ✓ throws when setting a list or dictionary item to an embedded object
        ✓ throws when setting a list or dictionary outside a transaction
        ✓ throws when setting a list item out of bounds
        ✓ throws when setting a nested list item out of bounds
        ✓ throws when assigning to list snapshot (Results)
        ✓ invalidates the list when removed
        ✓ invalidates the dictionary when removed

  Observable
    Collections in Mixed
      Collection notifications
        List
          ✓ fires when inserting, updating, and deleting at top-level
          ✓ fires when inserting, updating, and deleting in nested list
          ✓ fires when inserting, updating, and deleting in nested dictionary
          ✓ does not fire when updating object at top-level
        Dictionary
          ✓ fires when inserting, updating, and deleting at top-level
          ✓ fires when inserting, updating, and deleting in nested list
          ✓ fires when inserting, updating, and deleting in nested dictionary
          ✓ does not fire when updating object at top-level
      Object notifications
        ✓ fires when inserting, updating, and deleting in top-level list
        ✓ fires when inserting, updating, and deleting in nested list
        ✓ fires when inserting, updating, and deleting in top-level dictionary
        ✓ fires when inserting, updating, and deleting in nested dictionary
        ✓ fires when inserting, updating, and deleting in nested dictionary (using key-path)
```
</details>

### Brief overview of usage

```ts
class CustomObject extends Realm.Object {
  value!: Realm.Types.Mixed;

    name: "CustomObject",
    properties: {
      value: "mixed",
    },
  };
}

const realm = await Realm.open({ schema: [CustomObject] });

// Create an object with a dictionary value as the Mixed property, containing primitives
// and a list. (To use the properties of a Realm object, embedded Realm object, or a
// non-Realm object instance as a dictionary, you must spread the object into a new one).
const realmObject = realm.write(() => {
  return realm.create(CustomObject, {
    value: {
      num: 1,
      string: "hello",
      bool: true,
      list: [1, "hello", true /*, more nested collections */],
    },
  });
});

// Accessing the value returns the managed collection.
const dictionary = realmObject.value as Realm.Dictionary;
console.log(dictionary instanceof Realm.Dictionary);         // true
console.log(dictionary.list instanceof Realm.List);          // true
console.log((dictionary.list as Realm.List)[1] === "hello"); // true

// Get all objects with the Mixed `value` property being a dictionary
// containing the key `list` with the second item matching `"hello"`.
// (See "Filtering" tests for more alternatives.)
const objects = realm.objects(CustomObject);
let filtered = objects.filtered(`value.list[1] == $0`, "hello");
console.log(filtered.length); // 1

// Same as above query but `"hello"` can match on any index in the list.
filtered = objects.filtered(`value.list[*] == $0`, "hello");
console.log(filtered.length); // 1

// Update the Mixed property to a list.
realm.write(() => {
  realmObject.value = [1, "hello", { key: "value" }];
});
```

This closes #6344, #6345, #6245

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -307,13 +307,12 @@ describe("Dictionary", () => {
       });
     });
 
-    // This is currently not supported
-    it.skip("can store dictionary values using string keys", function (this: RealmContext) {
+    it("can store dictionary values using string keys", function (this: RealmContext) {
       const item = this.realm.write(() => {
         const item = this.realm.create<Item>("Item", {});
         const item2 = this.realm.create<Item>("Item", {});
-        item2.dict.key1 = "Hello";
-        item.dict.key1 = item2.dict;
+        item2.dict.key1 = "hello";
+        item.dict.key1 = item2;
         return item;
       });
       // @ts-expect-error We expect a dictionary inside dictionary
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -20,7 +20,6 @@ import { expect } from "chai";
 import Realm, { PropertySchema } from "realm";
 
 import { openRealmBefore, openRealmBeforeEach } from "../hooks";
-import { sleep } from "../utils/sleep";
 
 type Item<ValueType = Realm.Mixed> = {
   dict: Realm.Dictionary<ValueType>;
@@ -60,17 +59,6 @@ const DictTypedSchema: Realm.ObjectSchema = {
   },
 };
 
-const DictMixedSchema = {
-  name: "MixedDictionary",
-  properties: {
-    dict1: "mixed{}",
-    dict2: "mixed{}",
-  },
-};
-
-type IDictSchema = {
-  fields: Record<any, any>;
-};
 type ITwoDictSchema = {
   dict1: Record<any, any>;
   dict2: Record<any, any>;
@@ -598,7 +586,7 @@ describe("Dictionary", () => {
   });
 
   describe("embedded models", () => {
-    openRealmBeforeEach({ schema: [DictTypedSchema, DictMixedSchema, EmbeddedChild] });
+    openRealmBeforeEach({ schema: [DictTypedSchema, EmbeddedChild] });
     it("inserts correctly", function (this: RealmContext) {
       this.realm.write(() => {
         this.realm.create(DictTypedSchema.name, {
@@ -613,17 +601,6 @@ describe("Dictionary", () => {
       expect(dict_1.children2.num).equal(3, "We expect children2#3");
       expect(dict_2.children1?.num).equal(4, "We expect children1#4");
       expect(dict_2.children2?.num).equal(5, "We expect children2#5");
-    });
-
-    it("throws on invalid input", function (this: RealmContext) {
-      this.realm.write(() => {
-        expect(() => {
-          this.realm.create(DictMixedSchema.name, {
-            dict1: { children1: { num: 2 }, children2: { num: 3 } },
-            dict2: { children1: { num: 4 }, children2: { num: 5 } },
-          });
-        }).throws("Unable to convert an object with ctor 'Object' to a Mixed");
-      });
     });
   });
 });
--- a/integration-tests/tests/src/tests/dictionary.ts
+++ b/integration-tests/tests/src/tests/dictionary.ts
@@ -303,8 +303,9 @@ describe("Dictionary", () => {
         item.dict.key1 = item2;
         return item;
       });
-      // @ts-expect-error We expect a dictionary inside dictionary
-      expect(item.dict.key1.dict.key1).equals("hello");
+      const innerObject = item.dict.key1 as Realm.Object<Item> & Item;
+      expect(innerObject).instanceOf(Realm.Object);
+      expect(innerObject.dict).deep.equals({ key1: "hello" });
     });
 
     it("can store a reference to itself using string keys", function (this: RealmContext) {
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -772,6 +772,7 @@ describe("Lists", () => {
     openRealmBeforeEach({
       schema: [LinkTypeSchema, TestObjectSchema, PersonListSchema, PersonSchema, PrimitiveArraysSchema],
     });
+
     it("are typesafe", function (this: RealmContext) {
       let obj: ILinkTypeSchema;
       let prim: IPrimitiveArraysSchema;
@@ -870,24 +871,6 @@ describe("Lists", () => {
         testAssign("data", DATA1);
         testAssign("date", DATE1);
 
-        function testAssignNull(name: string, expected: string) {
-          //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-          expect(() => (prim[name] = [null])).throws(Error, `Expected '${name}[0]' to be ${expected}, got null`);
-          // TODO: Length should equal 1 when this is fixed: https://github.com/realm/realm-js/issues/6359
-          //       (This is due to catching the above error within this write transaction.)
-          //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-          expect(prim[name].length).equals(0);
-          // expect(prim[name].length).equals(1);
-        }
-
-        testAssignNull("bool", "a boolean");
-        testAssignNull("int", "a number or bigint");
-        testAssignNull("float", "a number");
-        testAssignNull("double", "a number");
-        testAssignNull("string", "a string");
-        testAssignNull("data", "an instance of ArrayBuffer");
-        testAssignNull("date", "an instance of Date");
-
         testAssign("optBool", true);
         testAssign("optInt", 1);
         testAssign("optFloat", 1.1);
@@ -910,7 +893,33 @@ describe("Lists", () => {
       //@ts-expect-error throws on modification outside of transaction.
       expect(() => (prim.bool = [])).throws("Cannot modify managed objects outside of a write transaction.");
     });
+
+    it("throws when assigning null to non-nullable", function (this: RealmContext) {
+      const realm = this.realm;
+      const prim = realm.write(() => realm.create<IPrimitiveArraysSchema>(PrimitiveArraysSchema.name, {}));
+
+      function testAssignNull(name: string, expected: string) {
+        expect(() => {
+          realm.write(() => {
+            // @ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
+            prim[name] = [null];
+          });
+        }).throws(Error, `Expected '${name}[0]' to be ${expected}, got null`);
+
+        // @ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
+        expect(prim[name].length).equals(0);
+      }
+
+      testAssignNull("bool", "a boolean");
+      testAssignNull("int", "a number or bigint");
+      testAssignNull("float", "a number");
+      testAssignNull("double", "a number");
+      testAssignNull("string", "a string");
+      testAssignNull("data", "an instance of ArrayBuffer");
+      testAssignNull("date", "an instance of Date");
+    });
   });
+
   describe("operations", () => {
     openRealmBeforeEach({ schema: [LinkTypeSchema, TestObjectSchema, PersonSchema, PersonListSchema] });
     it("supports enumeration", function (this: RealmContext) {
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -795,8 +795,10 @@ describe("Lists", () => {
         //@ts-expect-error TYPEBUG: type missmatch, forcecasting shouldn't be done
         obj.arrayCol = [this.realm.create<ITestObjectSchema>(TestObjectSchema.name, { doubleCol: 1.0 })];
         expect(obj.arrayCol[0].doubleCol).equals(1.0);
-        obj.arrayCol = obj.arrayCol; // eslint-disable-line no-self-assign
-        expect(obj.arrayCol[0].doubleCol).equals(1.0);
+
+        // TODO: Solve the "removeAll()" case for self-assignment.
+        // obj.arrayCol = obj.arrayCol; // eslint-disable-line no-self-assign
+        // expect(obj.arrayCol[0].doubleCol).equals(1.0);
 
         //@ts-expect-error Person is not assignable to boolean.
         expect(() => (prim.bool = [person])).throws(
@@ -874,8 +876,11 @@ describe("Lists", () => {
         function testAssignNull(name: string, expected: string) {
           //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
           expect(() => (prim[name] = [null])).throws(Error, `Expected '${name}[0]' to be ${expected}, got null`);
+          // TODO: Length should equal 1 when this is fixed: https://github.com/realm/realm-js/issues/6359
+          //       (This is due to catching the above error within this write transaction.)
           //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-          expect(prim[name].length).equals(1);
+          expect(prim[name].length).equals(0);
+          // expect(prim[name].length).equals(1);
         }
 
         testAssignNull("bool", "a boolean");
--- a/integration-tests/tests/src/tests/list.ts
+++ b/integration-tests/tests/src/tests/list.ts
@@ -709,12 +709,9 @@ describe("Lists", () => {
         expect(() => (array[0] = array)).throws(Error, "Missing value for property 'doubleCol'");
         expect(() => (array[2] = { doubleCol: 1 })).throws(
           Error,
-          "Cannot set element at index 2 out of bounds (length 2)",
+          "Requested index 2 calling set() on list 'LinkTypesObject.arrayCol' when max is 1",
         );
-        expect(() => (array[-1] = { doubleCol: 1 })).throws(
-          Error,
-          "Cannot set element at index -1 out of bounds (length 2)",
-        );
+        expect(() => (array[-1] = { doubleCol: 1 })).throws(Error, "Cannot set item at negative index -1");
 
         //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
         array["foo"] = "bar";
--- a/integration-tests/tests/src/tests/mixed.ts
+++ b/integration-tests/tests/src/tests/mixed.ts
@@ -1269,6 +1269,32 @@ describe("Mixed", () => {
             removed = this.realm.write(() => list.shift());
             expect(removed).to.be.undefined;
           });
+
+          it("unshift()", function (this: RealmContext) {
+            const { mixed: list } = this.realm.write(() => {
+              return this.realm.create<IMixedSchema>(MixedSchema.name, { mixed: [] });
+            });
+            expectRealmList(list);
+            expect(list.length).equals(0);
+
+            // Insert item into top-level list.
+            let newLength = this.realm.write(() => list.unshift({}));
+            expect(newLength).equals(1);
+            expectRealmDictionary(list[0]);
+            newLength = this.realm.write(() => list.unshift([]));
+            expect(newLength).equals(2);
+            const nestedList = list[0];
+            expectRealmList(nestedList);
+            expect(nestedList.length).equals(0);
+
+            // Insert item into nested list.
+            newLength = this.realm.write(() => nestedList.unshift("string"));
+            expect(newLength).equals(1);
+            expect(nestedList[0]).equals("string");
+            newLength = this.realm.write(() => nestedList.unshift(1));
+            expect(newLength).equals(2);
+            expect(nestedList[0]).equals(1);
+          });
         });
 
         describe("Iterators", () => {
--- a/integration-tests/tests/src/tests/mixed.ts
+++ b/integration-tests/tests/src/tests/mixed.ts
@@ -314,23 +314,24 @@ describe("Mixed", () => {
     });
 
     it("throws if nested type is an embedded object", function (this: RealmContext) {
-      this.realm.write(() => {
-        // Create an object with an embedded object property.
-        const { embeddedObject } = this.realm.create(MixedAndEmbeddedSchema.name, {
+      // Create an object with an embedded object property.
+      const { embeddedObject } = this.realm.write(() => {
+        return this.realm.create(MixedAndEmbeddedSchema.name, {
           mixed: null,
           embeddedObject: { mixed: 1 },
         });
-        expect(embeddedObject).instanceOf(Realm.Object);
-
-        // Create an object with the Mixed property being the embedded object.
-        expect(() => this.realm.create(MixedAndEmbeddedSchema.name, { mixed: embeddedObject })).to.throw(
-          "Using an embedded object (EmbeddedObject) as a Mixed value is not supported",
-        );
       });
+      expect(embeddedObject).instanceOf(Realm.Object);
+
+      // Create an object with the Mixed property being the embedded object.
+      expect(() => {
+        this.realm.write(() => {
+          this.realm.create(MixedAndEmbeddedSchema.name, { mixed: embeddedObject });
+        });
+      }).to.throw("Using an embedded object (EmbeddedObject) as a Mixed value is not supported");
+
       const objects = this.realm.objects<IMixedAndEmbedded>(MixedAndEmbeddedSchema.name);
-      // TODO: Length should equal 1 when this PR is merged: https://github.com/realm/realm-js/pull/6356
-      // expect(objects.length).equals(1);
-      expect(objects.length).equals(2);
+      expect(objects.length).equals(1);
       expect(objects[0].mixed).to.be.null;
     });
   });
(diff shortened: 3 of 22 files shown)


packages/realm/src/Set.ts
```diff
@@ -129,3 +142,42 @@ export class RealmSet<T = unknown> extends OrderedCollection<T, [T, T]> {
...
+};
+
+type SetAccessorFactoryOptions<T> = {
+  realm: Realm;
+  typeHelpers: TypeHelpers<T>;
+  isObjectItem?: boolean;
+};
+
+/** @internal */
+export function createSetAccessor<T>({
```
nirinchev, line 164: Similarly here - why don't we have a set accessor for `Set<Mixed>`?

nirinchev (commented):
Need to go through the tests, but the core changes seem to make sense to me - I flagged a few places where I have questions, but those are non-blockers, mostly stuff I'm curious about.

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -307,13 +307,12 @@ describe("Dictionary", () => {
       });
     });
 
-    // This is currently not supported
-    it.skip("can store dictionary values using string keys", function (this: RealmContext) {
+    it("can store dictionary values using string keys", function (this: RealmContext) {
```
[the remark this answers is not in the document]
elle-j: This was just a preexisting test that was skipped unnecessarily. Haven't changed those tests (except this one as the test itself was incorrect). I think the description refers to storing dictionary (`item.dict`) values using string keys (`key1`).

integration-tests/tests/src/tests/dictionary.ts
```diff
@@ -307,13 +307,12 @@ describe("Dictionary", () => {
...
       const item = this.realm.write(() => {
         const item = this.realm.create<Item>("Item", {});
         const item2 = this.realm.create<Item>("Item", {});
-        item2.dict.key1 = "Hello";
-        item.dict.key1 = item2.dict;
+        item2.dict.key1 = "hello";
+        item.dict.key1 = item2;
         return item;
       });
       // @ts-expect-error We expect a dictionary inside dictionary
```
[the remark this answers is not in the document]
elle-j: It's due to accessing unknown values, so TypeScript doesn't know about the keys.

integration-tests/tests/src/tests/list.ts
```diff
@@ -709,9 +709,12 @@ describe("Lists", () => {
         expect(() => (array[0] = array)).throws(Error, "Missing value for property 'doubleCol'");
         expect(() => (array[2] = { doubleCol: 1 })).throws(
           Error,
-          "Requested index 2 calling set() on list 'LinkTypesObject.arrayCol' when max is 1",
+          "Cannot set element at index 2 out of bounds (length 2)",
+        );
+        expect(() => (array[-1] = { doubleCol: 1 })).throws(
+          Error,
+          "Cannot set element at index -1 out of bounds (length 2)",
```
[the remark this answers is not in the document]
elle-j: Yeah that's true! I'll see if we can update these 🙂

integration-tests/tests/src/tests/results.ts
```diff
@@ -186,7 +186,7 @@ describe("Results", () => {
       expect(() => {
         //@ts-expect-error Should be an invalid write to read-only object.
         objects[-1] = { doubleCol: 0 };
-      }).throws("Index -1 cannot be less than zero.");
+      }).throws("Assigning into a Results is not supported");
```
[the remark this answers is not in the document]
elle-j: Could absolutely update that, I just decided to use the same error as we previously used for assigning into Results.

integration-tests/tests/src/tests/dictionary.ts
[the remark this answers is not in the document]
nirinchev: Gotcha - I think we need to update the clarification there because it definitely is not clarifying (at least for me).

packages/realm/bindgen/src/templates/node.ts
```diff
@@ -76,7 +76,7 @@ function pushRet<T, U extends T>(arr: T[], elem: U) {
 class NodeAddon extends CppClass {
   exports: Record<string, string> = {};
   classes: string[] = [];
-  injectables = ["Float", "UUID", "ObjectId", "Decimal128", "EJSON_parse", "EJSON_stringify"];
+  injectables = ["Float", "UUID", "ObjectId", "Decimal128", "EJSON_parse", "EJSON_stringify", "Symbol_for"];
```
kraenhansen, line 79: Technically not needed as it isn't used in the Node implementation 🙂

packages/realm/src/Dictionary.ts
```diff
@@ -322,3 +317,165 @@ export class Dictionary<T = unknown> extends Collection<string, T, [string, T],
...
+  const value = dictionary.tryGetAny(key);
+  if (value === binding.ListSentinel) {
+    const accessor = createListAccessor<T>({ realm, typeHelpers, isMixedItem: true });
+    return new List<T>(realm, dictionary.getList(key), accessor) as T;
+  }
+  if (value === binding.DictionarySentinel) {
+    const accessor = createDictionaryAccessor<T>({ realm, typeHelpers, isMixedItem: true });
+    return new Dictionary<T>(realm, dictionary.getDictionary(key), accessor) as T;
+  }
+  return typeHelpers.fromBinding(value) as T;
```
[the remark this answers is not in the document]
elle-j: For the way we're using symbols it would work. Given that the conditions are all just based on the value it may be preferable to use switch you like said. In that case, I'll have a look at the other places where I do the same thing 👍

packages/realm/src/Dictionary.ts
```diff
@@ -322,3 +317,165 @@ export class Dictionary<T = unknown> extends Collection<string, T, [string, T],
...
+  toBinding: TypeHelpers<T>["toBinding"],
+  dictionary: binding.Dictionary,
+  key: string,
+  value: T,
+): void {
+  assert.inTransaction(realm);
+
+  if (isJsOrRealmList(value)) {
+    dictionary.insertCollection(key, binding.CollectionType.List);
+    insertIntoListInMixed(value, dictionary.getList(key), toBinding);
```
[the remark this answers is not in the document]
elle-j: [Example](https://github.com/realm/realm-js/blob/e43a600f2bb9f55380e3d14d9eb281e833828c95/integration-tests/tests/src/tests/mixed.ts#L1064)

packages/realm/src/List.ts
```diff
@@ -343,3 +311,168 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
...
+function setKnownType<T>(
+  realm: Realm,
+  toBinding: TypeHelpers<T>["toBinding"],
+  isEmbedded: boolean,
+  list: binding.List,
+  index: number,
+  value: T,
+): void {
+  assert.inTransaction(realm);
+  list.setAny(index, toBinding(value, isEmbedded ? { createObj: () => [list.setEmbedded(index), true] } : undefined));
```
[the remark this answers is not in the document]
elle-j: For the known types, it's still using the same implementation as before. I've noted this difference as well between set and insert and will look at that once all new implementation is finalized, without modifying the old behavior of that path.

packages/realm/src/Results.ts
```diff
@@ -177,5 +187,30 @@ export class Results<T = unknown> extends OrderedCollection<T> {
   }
 }
 
+/**
+ * Accessor for getting items from the binding collection, as well
+ * as converting the values to and from their binding representations.
+ * @internal
+ */
+export type ResultsAccessor<T = unknown> = TypeHelpers<T> & {
```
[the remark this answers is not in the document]
elle-j: If you're referring to the `createResultsAccessor()`, that only needs to return the default getter (which handles primitives) (also true for Sets). If e.g. a user calls an API on a `List` or `Dictionary` that returns a `Results` of their content, the list/dictionary accessor will be used. (See e.g. the [accessor passed](https://github.com/realm/realm-js/blob/e43a600f2bb9f55380e3d14d9eb281e833828c95/packages/realm/src/OrderedCollection.ts#L150) to `OrderedCollection`, and [returning a Results](https://github.com/realm/realm-js/blob/e43a600f2bb9f55380e3d14d9eb281e833828c95/packages/realm/src/OrderedCollection.ts#L916)).

packages/realm/bindgen/src/templates/node.ts
elle-j: Good catch, thanks!

packages/realm/src/Dictionary.ts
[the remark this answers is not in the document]
elle-j: [Updated example](https://github.com/realm/realm-js/pull/6513/commits/70b7574a7edc76bf5c87d3e81d34d2dc72273b6d) with having a collection there prior to reassigning.

packages/realm/src/Results.ts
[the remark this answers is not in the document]
elle-j: Thanks for the interesting discussion about this in the office, I've made [some updates](https://github.com/realm/realm-js/pull/6513/commits/b5f5f04c4570df52503906b309c9791e175047e4) to make the Results and Set accessors in those cases more explicit (with the tradeoff of some duplicate code).

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -135,7 +135,7 @@ const CollectionsOfMixedSchema: ObjectSchema = {
 };
 
 const bool = true;
-const int = 123;
+const int = BigInt(123);
```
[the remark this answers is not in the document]
elle-j: Yeah TS will see this as `bigint`, but the primary reason for this change is: Since these values will be used on a property defined as `"mixed"` in the schema, all regular numbers (e.g. `123`) that we pass will be interpreted as doubles (unlike if we would set `123` to a property defined as `int` in the schema). So in order to test the [@type == 'int'](https://github.com/realm/realm-js/blob/e4d7f1b9904d56ca1cb1a2d0e8d0b2fa84d6444b/integration-tests/tests/src/tests/mixed.ts#L2049) in queries, we can use `BigInt` to pass it as an `int64` to Core.

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -145,12 +145,32 @@ const uuid = new BSON.UUID();
...
 // Realm object will be added by the individual tests after one has been created.
-const flatListAllTypes: unknown[] = [bool, int, double, d128, string, date, oid, uuid, nullValue, uint8Buffer];
-const flatDictionaryAllTypes: Record<string, unknown> = {
+const unmanagedRealmObject: IMixedSchema = { mixed: 1 };
+
+/**
+ * An array of values representing each Realm data type allowed as `Mixed`,
+ * except for a managed Realm Object, a nested list, and a nested dictionary.
+ */
+const primitiveTypesList: readonly unknown[] = [
```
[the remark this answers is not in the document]
elle-j: For numbers in Mixed, I believe we currently only pass doubles and ints to Core. (For properties with `float` in the schema, we pass a custom float binding representation.)

packages/realm/src/Set.ts
elle-j: Will refer to the related [previous thread](https://github.com/realm/realm-js/pull/6513#discussion_r1508894777) 👍

packages/realm/src/Dictionary.ts
```diff
@@ -209,12 +212,16 @@ export class Dictionary<T = unknown> extends Collection<
...
+    const itemType = toItemType(snapshot.type);
+    const { fromBinding, toBinding } = this[ACCESSOR];
+    const accessor = createResultsAccessor({ realm, typeHelpers: { fromBinding, toBinding }, itemType });
+    const results = new Results<T>(realm, snapshot, accessor);
+    const size = results.length;
 
-    const { snapshotGet } = this[ACCESSOR];
     for (let i = 0; i < size; i++) {
-      yield snapshotGet(snapshot, i);
+      yield results[i];
```
nirinchev: Maybe it's me, but isn't this a bit overly complicated? Can we do something like:

```ts
const realm = this[REALM];
const results = this[INTERNAL].values;
const itemType = toItemType(snapshot.type);
const { fromBinding, toBinding } = this[ACCESSOR];
const accessor = createResultsAccessor({ realm, typeHelpers: { fromBinding, toBinding }, itemType });
const results = new Results<T>(realm, snapshot, accessor);
return results.values();
```

Essentially, construct the managed results from `this[INTERNAL].values` and return its generator instead (it'll do the snapshotting already, making the original snapshot unnecessary).

elle-j: The generator function explicitly needs to `yield` the next value, returning the Results generator won't work. You'd have to replace `return results.values()` to the following:
```ts
for (const value of results.values()) {
  yield value;
}
```

packages/realm/src/OrderedCollection.ts
```diff
@@ -360,11 +374,17 @@ export abstract class OrderedCollection<T = unknown, EntryType extends [unknown,
...
+
     if (this.type === "object") {
       assert.instanceOf(searchElement, RealmObject);
-      return this.results.indexOfObj(searchElement[INTERNAL]);
+      return this.results.indexOfObj(searchElement[OBJ_INTERNAL]);
+    } else if (isJsOrRealmList(searchElement) || isJsOrRealmDictionary(searchElement)) {
+      // Collections are always treated as not equal since their
+      // references will always be different for each access.
+      const NOT_FOUND = -1;
+      return NOT_FOUND;
```
papafe, line 385: Why do we use a const here?

papafe (approved):
Looks good to me for what I can follow 😁

packages/realm/src/OrderedCollection.ts
elle-j: Do you mean why not just return `-1`? It's merely for readability.

papafe: I understand you want to make clear that -1 is used when the value is not found, but for me it looks a little bit strange creating a const variable and then using it straight away in the same function. Aren't we also creating a new variable unnecessarily every time the method is called?
I think a comment would suffice in this case, or maybe move the const declaration at a class or higher level. 
Just an observation though, I don't have super strong opinion on this.

elle-j: I see your point 🙂 In this case, this code only runs when the value to search for is a collection, and these two statements would likely just be replaced with returning `-1` in the end anyway. In this context I think just having `-1` (without either the variable or a comment) would not be too bad, but I think the readability increases if you do have it.

integration-tests/tests/src/tests/mixed.ts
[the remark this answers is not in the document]
kneth: Since `float` is a subset of `double` it will be difficult to create a full-proved method.

packages/realm/src/List.ts
```diff
@@ -36,27 +43,31 @@ type PartiallyWriteableArray<T> = Pick<Array<T>, "pop" | "push" | "shift" | "uns
...
-    super(realm, internal.asResults(), helpers);
+    const results = internal.asResults();
+    super(realm, results, accessor);
 
     // Getting the `objectSchema` off the internal will throw if base type isn't object
-    const baseType = this.results.type & ~binding.PropertyType.Flags;
     const isEmbedded =
-      baseType === binding.PropertyType.Object && internal.objectSchema.tableType === binding.TableType.Embedded;
+      toItemType(results.type) === binding.PropertyType.Object &&
+      internal.objectSchema.tableType === binding.TableType.Embedded;
```
kneth, line 70: swapping the order might improve performance a bit 😄

packages/realm/src/List.ts
```diff
@@ -343,3 +312,171 @@ export class List<T = unknown> extends OrderedCollection<T> implements Partially
...
+export function insertIntoListOfMixed(
+  list: List | unknown[],
+  internal: binding.List,
+  toBinding: TypeHelpers["toBinding"],
+) {
+  // TODO: Solve the "removeAll()" case for self-assignment.
+  internal.removeAll();
+
+  let index = 0;
+  for (const item of list) {
```
kneth, line 465: Is `for (const [index, item] of list.entries)` possible here?

kneth (approved):
After your introduction in person, I don't have many comments.

As we don't need much update to the API documentation, I suggest that we write a larger section in `CHANGELOG.md` when we are about to merge.

packages/realm/src/List.ts
elle-j: There's a comment above about `internal.objectSchema.tableType` throwing if we don't check for the Object property type first.

packages/realm/src/List.ts
elle-j: Yes that'd be possible with `list.entries()` 🙂 And now that I think about it, it may even be safer to do it that way because if the `list` is a `Realm.List`, then `list.entries()` would take the snapshot.

integration-tests/tests/src/tests/list.ts
```diff
@@ -871,8 +873,11 @@ describe("Lists", () => {
         function testAssignNull(name: string, expected: string) {
           //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
           expect(() => (prim[name] = [null])).throws(Error, `Expected '${name}[0]' to be ${expected}, got null`);
+          // TODO: Length should equal 1 when this is fixed: https://github.com/realm/realm-js/issues/6359
+          //       (This is due to catching the above error within this write transaction.)
           //@ts-expect-error TYPEBUG: our List type-definition expects index accesses to be done with a number , should probably be extended.
-          expect(prim[name].length).equals(1);
```
kraenhansen, line 875: Instead of patching the test here, I wonder if it'd make more sense to separate this out into another test and skip it for now? And eventually link to that from #6359.

packages/realm/bindgen/src/templates/node-wrapper.ts
```diff
@@ -141,6 +141,7 @@ export function generate({ rawSpec, spec: boundSpec, file }: TemplateContext): v
     "Decimal128",
     "EJSON_parse: EJSON.parse",
     "EJSON_stringify: EJSON.stringify",
+    "Symbol_for: Symbol.for",
```
kraenhansen, line 144: Turns out this was actually not needed on Node, since NAPI expose a method to do this.
No harm in keeping it, but its not used but the node.ts template.

packages/realm/bindgen/src/templates/jsi.ts
```diff
@@ -902,6 +912,16 @@ class JsiCppDecls extends CppDecls {
               `,
             )
             .join("\n")}
+
+          // We are returning sentinel values for lists and dictionaries in the
+          // form of Symbol singletons. This is due to not being able to construct
+          // the actual list or dictionary in the current context.
+          case realm::type_List:
+            return ${this.addon.accessCtor("Symbol_for")}.call(_env, "Realm.List");
```
kraenhansen, line 920: `accessCtor` seems ripe for a rename. Perhaps it should be closer linked to "injectable"?

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -839,78 +2714,180 @@ describe("Mixed", () => {
...
+
+        const nestedList = list[0];
+        expectRealmList(nestedList);
+        expect(nestedList.length).equals(0);
+
+        expect(() => {
+          this.realm.write(() => {
+            nestedList[0] = "primitive";
+          });
+        }).to.throw("Requested index 0 calling set() on list 'MixedClass.mixed' when empty");
```
kraenhansen, line 2810: Shouldn't this be allowed?

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -839,78 +2714,180 @@ describe("Mixed", () => {
...
+          return this.realm.create<IMixedSchema>(MixedSchema.name, { mixed: [] });
+        });
+        expectRealmList(list);
+        expect(list.length).equals(0);
+
+        expect(() => {
+          this.realm.write(() => {
+            list[0] = "primitive";
+          });
+        }).to.throw("Requested index 0 calling set() on list 'MixedClass.mixed' when empty");
```
kraenhansen, line 2779: I'm surprised this isn't allowed.

integration-tests/tests/src/tests/observable.ts
```diff
@@ -1379,304 +1387,571 @@ describe("Observable", () => {
...
-            });
-          },
-          {
-            deletions: [],
-            insertions: ["amy", "mary", "john"],
-            modifications: [],
-          },
-        ]);
-      });
+          await expectCollectionNotifications(list, undefined, [
```
kraenhansen, line 1418: It seems to me none of the new tests are using key-path filtering, is that correct / on purpose? I'd think it makes sense to exercise that for nested collections in mixed too.

integration-tests/tests/src/tests/list.ts
elle-j: Sounds good 👍 I'll add `skip` to 2 other tests that also rely on that issue being solved.

packages/realm/bindgen/src/templates/node-wrapper.ts
elle-j: Forgot to remove that when I removed it from the node template, thanks for spotting it.

integration-tests/tests/src/tests/mixed.ts
```diff
@@ -839,78 +2714,180 @@ describe("Mixed", () => {
...
         this.realm.write(() => {
-          created.value = null;
+          created.mixed = null;
         });
-        expect(created.value).to.be.null;
+        expect(created.mixed).to.be.null;
         expect(() => dictionary.prop).to.throw("This collection is no more");
       });
+
+      // If `REALM_DEBUG`, the max nesting level is 4.
```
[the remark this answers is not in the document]
elle-j: Yeah on CI that would fail, but I'm going to remove that test anyway since Core is now using the same max nesting level as for release.

integration-tests/tests/src/tests/mixed.ts
elle-j: This is where `Realm.List` differs in behavior from a JS Array. The Core APIs for *setting* will throw if `index >= size`, whereas *inserting* only throws for `index >= size + 1`. This is the same behavior as we had before for `Realm.List`.

From a JS perspective, setting a value at, say, index 5 of an empty list, we'd have to set index 0-4 to `undefined` (although this would be converted to `null` values).

integration-tests/tests/src/tests/mixed.ts
elle-j: See [comment](https://github.com/realm/realm-js/pull/6513#discussion_r1532397116).

integration-tests/tests/src/tests/observable.ts
elle-j: For collection listeners, changes to a Realm Object in the collection will not fire a notification. But for object listeners, we could e.g. test that "Changes to a nested collection fires a notification" (if we're listening on a Realm Object with the key path being the mixed property) 👍 

Currently, the above test exists where the key path is undefined. I think our preexisting key path tests take care of testing that notifications are not fired when non-specified Realm Object properties change.

Did you have any other tests in mind?

kraenhansen: > For collection listeners, changes to a Realm Object in the collection will not fire a notification.

Is this true when explicitly providing a key-path? 🤔 And is this expected? (I'm unsure)

elle-j: I've tested some of this locally, and currently the key paths option is only supported if the collection is a collection of Realm objects in the schema. So if it's a Mixed in the schema, but the underlying value is dictionary of Realm objects, it's not supported. I believe this may change in Core at a later point.

kraenhansen (approved):

packages/realm/src/Dictionary.ts
elle-j: Applied ✅

PR #6516, opened 2024-02-28
Suggested changes to `kneth/realm-core-14.0.1`

kraenhansen:
## What, How & Why?

Instead of a lot of suggested changes in the GitHub PR UI, I thought a PR would be better.

### DIFF
--- a/packages/realm/src/Logger.ts
+++ b/packages/realm/src/Logger.ts
@@ -146,7 +146,7 @@ export enum LogCategory {
 /**
  * Type for `Realm.setLogLevel`
  */
-export type LogArgs = {
+export type LogOptions = {
   level: LogLevel;
   category?: LogCategory;
 };
--- a/packages/realm/src/Logger.ts
+++ b/packages/realm/src/Logger.ts
@@ -161,27 +161,29 @@ export type Logger = (level: NumericLogLevel, message: string) => void;
 
 /**
  * A callback passed to `Realm.setLogger`.
- *
  * @param level   - The level of the log entry.
  * @param message - The message of the log entry.
  * @since 12.0.0
  * @deprecated Will be removed in v13.0.0
  */
 export type LoggerCallback1 = (level: LogLevel, message: string) => void;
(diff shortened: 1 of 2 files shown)


packages/realm/src/Logger.ts
```diff
@@ -146,7 +146,7 @@ export enum LogCategory {
 /**
  * Type for `Realm.setLogLevel`
  */
-export type LogArgs = {
+export type LogOptions = {
```
kraenhansen, line 149: We tend to use "options" to describe the type of object passed as the sole argument to a function.

packages/realm/src/Logger.ts
```diff
@@ -161,27 +161,29 @@ export type Logger = (level: NumericLogLevel, message: string) => void;
...
  * @since 12.0.0
  * @deprecated Will be removed in v13.0.0
  */
 export type LoggerCallback1 = (level: LogLevel, message: string) => void;
-export type LoggerCallbackArgs = {
+
+/**
+ * Represents an entry in the log.
+ */
+export type LogEntry = {
```
kraenhansen, line 174: I believe "entry" is semantically more precise.

packages/realm/src/Logger.ts
```diff
@@ -161,27 +161,29 @@ export type Logger = (level: NumericLogLevel, message: string) => void;
...
   level: LogLevel;
+  /** The message of the log entry. */
   message: string;
 };
 /**
- * A callback passed to `Realm.setLogger`. Arguments are passed as a POJO.
- *
- * @param category   - The category (origin) of the log entry.
- * @param level      - The level of the log entry.
- * @param message    - The message of the log entry.
```
kraenhansen, line 181: As these are not separate arguments to the callback, I've moved these descriptions into the properties on the object's type above.

packages/realm/src/Logger.ts
kneth: Yes, it reads better

kneth (approved):

PR #6518, opened 2024-02-29
RJS-2732: Support updating the `App`'s base URL via experimental export

elle-j:
## What, How & Why?

**Experimental exports/imports**
* An `experimental` directory has been added (`packages/realm/src/experimental`), intended for exporting various experimental features.

**Base URL**
* Importing `"realm/experimental/base-url"` will extend the `App` with the instance members `App.baseUrl` and `App.updateBaseUrl()`, allowing the user to retrieve and update the base URL.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (but need to add follow-ups to skipped unimplemented tests)


packages/realm/src/experimental/base-url.ts
```diff
@@ -0,0 +1,48 @@
...
+Object.defineProperty(App.prototype, "baseUrl", {
+  get(this: App) {
+    return this.internal.getBaseUrl();
+  },
+  set() {
+    throw new Error("Cannot assign the base URL, please use `switchBaseUrl()`.");
+  },
+});
+
+App.prototype.switchBaseUrl = async function (this: App, url: string) {
```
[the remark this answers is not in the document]
elle-j: 👍 Will update when everybody aligns 🙂

nirinchev (approved):
Nice! Might be worth adding a smoke test that updating the base url to a valid url also works. We could try and use `realm.mongodb.com` since that domain should still work.

elle-j:
Just reviving the old draft implementation then I'll be modifying the test and adding some more 🙂

integration-tests/tests/src/tests/experimental/base-url.ts
```diff
@@ -0,0 +1,50 @@
...
+describe.skipIf(environment.missingServer, "Base URL", () => {
+  importAppBefore(buildAppConfig("with-anon").anonAuth());
+
+  it("returns the base URL used", function (this: AppContext) {
+    expect(this.app.baseUrl).equals(originalBaseUrl);
+  });
+
+  it.skip("updates the URL", async function (this: AppContext) {
+    // TODO
+  });
```
elle-j, line 34: Will add this with a fetch mock.

integration-tests/tests/src/hooks/import-app-before.ts
```diff
@@ -39,14 +39,6 @@ export { baseUrl };
...
 const allowSkippingServerTests = typeof environment.baseUrl === "undefined" && missingServer !== false;
 
-export type AppConfigurationRelaxed = {
-  id?: string;
-  baseUrl?: string;
-  timeout?: number;
-  multiplexSessions?: boolean;
-  baseFilePath?: string;
-};
-
```
kraenhansen, line 49: 💙

integration-tests/tests/tsconfig.common.json
```diff
@@ -3,7 +3,7 @@
 		"composite": true,
 		"target": "es2022",
 		"module": "es2022",
-		"moduleResolution": "node",
+		"moduleResolution": "Bundler",
```
kraenhansen, line 6: Did you check this works with `npm test --workspace @realm/integration-tests` (where we use raw Mocha and `tsx` to transpile TypeScript)?

packages/realm/experimental/base-url.js
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+// This file is needed as a workaround for Metro unexpectedly failing
+// to use the entry point defined in the package.json `exports` field,
+// despite enabling `unstable_enablePackageExports`.
```
kraenhansen, line 21: Well - even if this wasn't a problem, we would probably need this file anyway, to workaround the fact that `unstable_enablePackageExports` is disabled by default 🤔

packages/realm/src/experimental/base-url.ts
```diff
@@ -0,0 +1,48 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { App } from "../app-services/App";
+
+declare module "../app-services/App" {
```
kraenhansen, line 21: 💪

integration-tests/environments/react-native-test-app/ios/Podfile.lock
```diff
@@ -1107,7 +1107,7 @@ PODS:
     - React-Core
     - React-jsi
   - ReactTestApp-Resources (1.0.0-dev)
-  - RealmJS (12.7.0-rc.0):
+  - RealmJS (12.7.1):
```
kraenhansen, line 1110: I wonder why this was not already on `main` 🤔

kraenhansen (approved):

packages/realm/experimental/base-url.js
kraenhansen: Also ... I know for a fact that the Metro team would be very grateful if _someone_ would set up some tests of the metro resolver (most likely copying in tests from other packages / tools). This would make a good test case if the pattern isn't already covered by one of those other tests.

kneth (approved):
Nice work!

integration-tests/environments/react-native-test-app/metro.config.js
```diff
@@ -23,6 +23,7 @@ module.exports = makeMetroConfig({
       transform: {
         experimentalImportSupport: false,
         inlineRequires: false,
+        unstable_enablePackageExports: true,
```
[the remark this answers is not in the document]
elle-j: Good point 🙂

integration-tests/tests/tsconfig.common.json
elle-j: Works ✅

packages/realm/experimental/base-url.js
elle-j: I've [updated](https://github.com/realm/realm-js/pull/6518/commits/c3064c8aeff6dcbe5656afcbaf346b96ed30dfed) the comment to incorporate being disabled by default. Feel free to suggest any additional changes you think would be relevant for this file 👍 

I'd be glad to look over some of the Metro tests and see what could be added. It does seem like their tests are covering [subpath exports](https://github.com/facebook/metro/blob/ce266ddafd5eb484168f973f84a203d8b162f842/packages/metro-resolver/src/__tests__/package-exports-test.js#L235).

integration-tests/environments/react-native-test-app/ios/Podfile.lock
elle-j: Probably just didn't get committed 🤷‍♀️

elle-j:
@kneth, for the final review there are only minor changes to look at:
* `CHANGELOG.md`
* `packages/realm/src/experimental/base-url.ts`: Per [another discussion](https://github.com/realm/realm-dart/pull/1652#discussion_r1585972733), the API `updateBaseUrl()` now accepts `null` as a way to reset the base URL to the default one (in addition to empty string).

Still missing:
* Haven't got the mock fetch to work yet, so I'll add a ticket to implement the `skip`ed tests.

kneth (approved):

PR #6520, opened 2024-02-29
Minor updates to Core upgrade

elle-j:
## What, How & Why?

Adds some feedback to the Core upgrade branch.

### DIFF
--- a/packages/realm/src/Logger.ts
+++ b/packages/realm/src/Logger.ts
@@ -192,8 +192,7 @@ export function toBindingLogger(logger: LoggerCallback) {
     return binding.Helpers.makeLogger((_, level, message) => {
       logger(fromBindingLoggerLevelToLogLevel(level), message);
     });
-  }
-  if (isLoggerWithOptions(logger)) {
+  } else {
     return binding.Helpers.makeLogger((category, level, message) => {
       logger({
         category: category as LogCategory,
--- a/packages/realm/src/Logger.ts
+++ b/packages/realm/src/Logger.ts
@@ -201,15 +201,10 @@ export function toBindingLogger(logger: LoggerCallback) {
       });
     });
   }
-  throw new Error(`Unexpected logger passed.`);
 }
 
-function isLoggerWithLevel(logger: unknown): logger is LoggerCallback1 {
-  return typeof logger === "function" && logger.length === 2;
-}
-
-function isLoggerWithOptions(logger: unknown): logger is LoggerCallback2 {
-  return typeof logger === "function" && logger.length === 1;
+function isLoggerWithLevel(logger: LoggerCallback): logger is LoggerCallback1 {
+  return logger.length === 2;
 }
 
 /** @internal */
--- a/packages/realm/src/Logger.ts
+++ b/packages/realm/src/Logger.ts
@@ -144,10 +144,18 @@ export enum LogCategory {
 }
 
 /**
- * Type for `Realm.setLogLevel`
+ * Log options to use when setting the log level.
  */
 export type LogOptions = {
+  /**
+   * The log level to be used by the logger.
+   * @default "info"
+   */
   level: LogLevel;
+  /**
+   * The category to set the log level for. If omitted, the log level
+   * is set for all categories ({@link LogCategory.Realm}).
+   */
   category?: LogCategory;
 };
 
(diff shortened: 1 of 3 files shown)


kneth (approved):

packages/realm/src/Realm.ts
```diff
@@ -123,34 +122,27 @@ export class Realm {
 
   /**
    * Sets the log level for a specific category.
+   * @param options - The log options to use.
+   * @param options.level - The log level to be used by the logger. The default value is `info`.
+   * @param options.category - The category/component to set the log level for. If omitted, log level is set for all categories.
```
kraenhansen, line 127: I'd suggest moving these to the documentation of the `LogOptions` instead: https://github.com/realm/realm-js/blob/797a39556667b87a3bc05ec89ce4e623a151d11a/packages/realm/src/Logger.ts#L146-L152

packages/realm/src/Realm.ts
```diff
@@ -123,34 +122,27 @@ export class Realm {
...
     } else {
-      const { level, category } = arg;
-      if (category) {
-        setLevel(category, level);
-      } else {
-        for (const category of Object.values(LogCategory)) {
-          setLevel(category, level);
-        }
-      }
+      setLevel(arg.level, arg.category);
```
kraenhansen, line 145: I love this simplification! 😄 I did look at that code and thought it must be possible to de-duplicate, but none presented itself. Great suggestion.

packages/realm/src/Logger.ts
```diff
@@ -186,6 +186,33 @@ export type LogEntry = {
...
+  throw new Error(`Unexpected logger passed.`);
+}
+
+function isLoggerWithLevel(logger: unknown): logger is LoggerCallback1 {
+  return typeof logger === "function" && logger.length === 2;
+}
+
+function isLoggerWithOptions(logger: unknown): logger is LoggerCallback2 {
+  return typeof logger === "function" && logger.length === 1;
+}
```
kraenhansen, line 214: Great use if type guards! I suppose you know `logger` will be some function, so you could make it even more tight, but either way would be fine 👍 

```suggestion
function isLoggerWithLevel(logger: Function): logger is LoggerCallback1 {
  return logger.length === 2;
}

function isLoggerWithOptions(logger: Function): logger is LoggerCallback2 {
  return logger.length === 1;
}
```

packages/realm/src/Logger.ts
```diff
@@ -186,6 +186,33 @@ export type LogEntry = {
...
 export type LoggerCallback = LoggerCallback1 | LoggerCallback2;
 
+/** @internal */
+export function toBindingLogger(logger: LoggerCallback) {
+  if (isLoggerWithLevel(logger)) {
+    return binding.Helpers.makeLogger((_, level, message) => {
+      logger(fromBindingLoggerLevelToLogLevel(level), message);
+    });
+  }
+  if (isLoggerWithOptions(logger)) {
```
kraenhansen, line 196: You know me ... but I wont die on that hill 😆

```suggestion
  } else if (isLoggerWithOptions(logger)) {
```

kraenhansen (approved):

packages/realm/src/Realm.ts
elle-j: Me too, thanks for catching that!

packages/realm/src/Logger.ts
elle-j: With the current implementation we actually don't know that it's a function before those are called. But the caller of `toBindingLogger()` could do an `assert()` first (I'll add that). Then those functions can assume it's a function like you said 🙂

packages/realm/src/Logger.ts
elle-j: Great, I'll take it 😉

packages/realm/src/Logger.ts
```diff
@@ -186,6 +186,32 @@ export type LogEntry = {
...
+  } else {
+    return binding.Helpers.makeLogger((category, level, message) => {
+      logger({
+        category: category as LogCategory,
+        level: fromBindingLoggerLevelToLogLevel(level),
+        message,
+      });
+    });
+  }
+  throw new Error(`Unexpected logger passed.`);
```
kraenhansen, line 204: This should be unreachable now 🤔 

```suggestion
```

packages/realm/src/Logger.ts
```diff
@@ -144,10 +144,18 @@ export enum LogCategory {
...
+  /**
+   * The log level to be used by the logger.
+   * @default "info"
+   */
   level: LogLevel;
+  /**
+   * The category to set the log level for. If omitted, the log level
+   * is set for all categories ({@link LogCategory.Realm}).
+   */
   category?: LogCategory;
```
elle-j, line 159: Didn't use a `@default` for this because it doesn't work well in combination with `{@link}`.

PR #6523, opened 2024-03-01
Adding more precise detection of Expo Go

kraenhansen:
## What, How & Why?

Many newcomers to Realm and Expo gets confused about the fact that Expo Go doesn't ship with native modules.
Instead of showing a generic error when failing to load our native module, I suggest adding an attempt to detect Expo Go and showing a more direct message, with actionable steps:

Screenshot 2024-03-01 at 21 57 20

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (manually)

### DIFF
--- a/packages/realm/src/platform/react-native/expo-go-detection.ts
+++ b/packages/realm/src/platform/react-native/expo-go-detection.ts
@@ -1,0 +1,41 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2024 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


packages/realm/src/platform/react-native/expo-go-detection.ts
```diff
@@ -0,0 +1,41 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { Platform } from "react-native";
+
+export function isExpoGo() {
```
elle-j, line 21: Does it need to be exported?
```suggestion
function isExpoGo() {
```

kraenhansen:
Coverage test is still crashing (#6589) but that's not new - I'll go ahead and merge with it.

PR #6533, opened 2024-03-05
RPM-568: Add a workflow that verifies PR title

nirinchev:
Adds a workflow that verifies that PR titles have the form of "R...-123".

### DIFF
--- a/.github/workflows/check-pr-title.yml
+++ b/.github/workflows/check-pr-title.yml
@@ -1,0 +1,22 @@
+# NOTE: This is a common file that is overwritten by realm/ci-actions sync service
+# and should only be modified in that repository.
+
+name: "Check PR Title"
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled, converted_to_draft, edited]
+
+jobs:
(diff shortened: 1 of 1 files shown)


kraenhansen:
I won't expect external contributors to reference issues by their Jira id mad as such I think this change create unnecessary noise / friction for external contributors.

As a personal note: I still consider Jira a mirror of GitHub and not the other way around, at the very least when it comes to the relationship between issues and code (which is the essence of a PR), since Jiras integration with our source-code is non-existent. The bot already comments with the Jira ID on issues as they're created, so we already have two-way traceability between the two systems.

nirinchev:
External contributors won't have GHA workflows run automatically, so they won't see the failed check until we approve the workflow run. At this point we can update the title to reference the Jira ticket or add the `no-jira-ticket` label.

Whether Jira is a mirror of github or the other way around is a philosophical discussion and I'd be happy to take that offline, but there's broad push to reference Jira tickets in PR titles/commits, so this workflow is intended to help with that.

elle-j (approved):
Thanks for adding this.

kraenhansen (approved):

PR #6535, opened 2024-03-05
Suggesting array of categories and string union for LogCategory

kraenhansen:
## What, How & Why?

This PR is a suggestion to address this comment: https://github.com/realm/realm-js/pull/6512#discussion_r1512715139

### DIFF
--- a/packages/realm/src/Realm.ts
+++ b/packages/realm/src/Realm.ts
@@ -28,6 +28,7 @@ import {
   DefaultObject,
   INTERNAL,
   InitialSubscriptions,
+  LOG_CATEGORIES,
   List,
   LogCategory,
   LogLevel,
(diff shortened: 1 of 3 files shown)


elle-j (approved):
It gives good intellisense 👌 And since the addition of duplicate names does not seem unlikely, it could be good to incorporate this change to prevent future breaking changes. It didn't look like the `LOG_CATEGORIES` array was publicly exported, which I think is probably important (to not have exported) since users may start relying on the indices which would also break with added categories.

packages/realm/src/Realm.ts
```diff
@@ -149,25 +150,25 @@ export class Realm {
...
-   * @deprecated Pass a callback taking a single object argument instead.
    * @example
-   * Realm.setLogger((level, message) => {
-   *   console.log(`[${level}] ${message}`);
+   * Realm.setLogger(({ category, level, message }) => {
+   *   console.log(`[${category} - ${level}] ${message}`);
    * });
    */
-  static setLogger(loggerCallback: LoggerCallback1): void;
+  static setLogger(loggerCallback: LoggerCallback2): void;
```
kraenhansen, line 158: I'm moving this up to improve DX by hitting this as the default overload.

Screenshot 2024-03-05 at 15 24 52

kraenhansen: But that could be considered breaking 😬

Screenshot_2024-03-05-16-54-38-91_40deb401b9ffe8e1df2f1cc5ba480b12~2

kneth (approved):

PR #6550, opened 2024-03-14
RJS-2673: Prevent flickering behavior in `RealmProvider`

### LINKED ISSUE #6333: react native: flashing of screens while realm loads. Cant navigate back

Always

### Description

I use the App, User and Realm Provider in an expo app (react native).
As soon as I turn on sync, whatever I click, am led to the home screen and cant navigate anywhere.
And its flashing like crazy whenever the realm loads (at least the loader of the realm is shown).

Its totally unusable. However, I need syncing. What goes on here?

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

_No response_

### Version

realm is version 12.3. @realm/react is 0.6.2

Atlas Device Sync

### Are you using encryption?

No

Android Emulator Pixel 7

### Build environment

I am using the expo dev client v49.0.15.

### Cocoapods version

_No response_

elle-j:
## What, How & Why?

This PR makes the internal fields of `User` non-enumerable in order to hide them during deep comparisons. It also updates the `UserProvider` to only `setUser()` when specific fields change.

A plausible reason for the flickering behavior described in #6333 is:
1. The `user` in `UserProvider` would continuously have its listener added (and removed) (in `v0.6.2`), setting the `user` to a new reference in its listener callback each time.
2. ..causing `RealmProvider` to rerender whenever the `user` is a new reference.
3. ..causing `RealmProvider` to check if the configuration has changed (`!areConfigurationsIdentical()`) and force a rerender if it has. In cases where it should not have changed, the function indicated that it had. The function uses Lodash's `isEqual()` which does a deep comparison that included the `User`'s internal fields as they were still enumerable. For objects with properties being set to functions (e.g. `User.listeners.options`), those will be different references if the user itself is a different reference.
4. ...causing the `useEffect` callback where the Realm is opened and closed in `RealmProvider` to run (setting the `realm` to the opened realm, then to `null` when unmounted). The switch between `realm` and `null` renders either the `children` or the `fallback` respectively. The repeated behavior causes the flickering.

This closes #6333

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-react/src/UserProvider.tsx
+++ b/packages/realm-react/src/UserProvider.tsx
@@ -16,7 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import React, { createContext, useContext, useEffect, useState } from "react";
+import React, { createContext, useContext, useEffect, useReducer, useState } from "react";
 import type Realm from "realm";
 
 import { useApp } from "./AppProvider";
@@ -35,6 +35,21 @@ type UserProviderProps = {
   children: React.ReactNode;
 };
 
+function userWasUpdated(userA: Realm.User | null, userB: Realm.User | null) {
+  if (!userA && !userB) {
+    return false;
+  } else if (userA && userB) {
+    return (
+      userA.id !== userB.id ||
+      userA.state !== userB.state ||
+      userA.accessToken !== userB.accessToken ||
+      userA.refreshToken !== userB.refreshToken
+    );
+  } else {
+    return true;
+  }
+}
+
 /**
  * React component providing a Realm user on the context for the sync hooks
  * to use. A `UserProvider` is required for an app to use the hooks.
@@ -42,37 +57,30 @@ type UserProviderProps = {
 export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback, children }) => {
   const app = useApp();
   const [user, setUser] = useState<Realm.User | null>(() => app.currentUser);
+  const [, forceUpdate] = useReducer((x) => x + 1, 0);
 
   // Support for a possible change in configuration
-  if (app.currentUser?.id != user?.id) {
-    setUser(app.currentUser);
+  const currentUser = app.currentUser;
+  if (userWasUpdated(user, currentUser)) {
+    setUser(currentUser);
   }
 
   useEffect(() => {
-    const event = () => {
-      let userHasChanged = false;
-      const currentUser = app.currentUser;
-      if (currentUser && user) {
-        userHasChanged =
-          currentUser.id !== user.id ||
-          currentUser.state !== user.state ||
-          currentUser.accessToken !== user.accessToken ||
-          currentUser.refreshToken !== currentUser.refreshToken;
-      } else if (currentUser || user) {
-        userHasChanged = true;
-      }
+    app.addListener(forceUpdate);
 
-      if (userHasChanged) {
-        setUser(app.currentUser);
-      }
-    };
-    user?.addListener(event);
-    app?.addListener(event);
-    return () => {
-      user?.removeListener(event);
-      app?.removeListener(event);
-    };
-  }, [user, app]);
+    return () => app.removeListener(forceUpdate);
+  }, [app]);
+
+  useEffect(() => {
+    user?.addListener(forceUpdate);
+
+    return () => user?.removeListener(forceUpdate);
+
+    /*
+      eslint-disable-next-line react-hooks/exhaustive-deps
+      -- We should depend on `user.id` rather than `user` as the ID will indicate a new user in this case.
+    */
+  }, [user?.id]);
 
   if (!user) {
     if (typeof Fallback === "function") {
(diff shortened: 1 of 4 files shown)


packages/realm-react/src/UserProvider.tsx
```diff
@@ -35,30 +35,52 @@ type UserProviderProps = {
...
  */
 export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback, children }) => {
   const app = useApp();
   const [user, setUser] = useState<Realm.User | null>(() => app.currentUser);
+  const [, forceUpdate] = useReducer((x) => x + 1, 0);
 
   // Support for a possible change in configuration
-  if (app.currentUser?.id != user?.id) {
-    setUser(app.currentUser);
+  const currentUser = app.currentUser;
```
greenafrican, line 63: Perhaps better to wrap this in a useEffect like:
```
useEffect(() => {
  const currentUser = app.currentUser;
  if (userWasUpdated(user, currentUser)) {
    setUser(currentUser);
  }
}, [app.currentUser, user]);
```

packages/realm-react/src/UserProvider.tsx
```diff
@@ -35,30 +35,52 @@ type UserProviderProps = {
...
+}
+
 /**
  * React component providing a Realm user on the context for the sync hooks
  * to use. A `UserProvider` is required for an app to use the hooks.
  */
 export const UserProvider: React.FC<UserProviderProps> = ({ fallback: Fallback, children }) => {
   const app = useApp();
   const [user, setUser] = useState<Realm.User | null>(() => app.currentUser);
+  const [, forceUpdate] = useReducer((x) => x + 1, 0);
```
greenafrican, line 60: More idiomatic to use a state setter here:
```
const [, forceUpdate] = useState(0);
const triggerUpdate = () => forceUpdate((prev) => prev + 1);
```

packages/realm-react/src/UserProvider.tsx
```diff
@@ -35,30 +35,52 @@ type UserProviderProps = {
...
-    app?.addListener(event);
-    return () => {
-      user?.removeListener(event);
-      app?.removeListener(event);
-    };
-  }, [user, app]);
+    app.addListener(forceUpdate);
+
+    return () => app.removeListener(forceUpdate);
+  }, [app]);
```
greenafrican, line 72: Add forceUpdate dep:
```
}, [app, forceUpdate]);
```

packages/realm-react/src/UserProvider.tsx
```diff
@@ -35,30 +35,52 @@ type UserProviderProps = {
...
+  useEffect(() => {
+    user?.addListener(forceUpdate);
+
+    return () => user?.removeListener(forceUpdate);
+
+    /*
+      eslint-disable-next-line react-hooks/exhaustive-deps
+      -- We should depend on `user.id` rather than `user` as the ID will indicate a new user in this case.
+    */
+  }, [user?.id]);
```
greenafrican, line 83: Add forceUpdate dep:
```
}, [user?.id, forceUpdate]);
```

packages/realm-react/src/UserProvider.tsx
elle-j: Thanks for you input 🙂 In this case, we intentionally want to avoid the `useEffect()` so as to not render stale state. Checking if the user was updated outside of the `useEffect()` and calling `setUser()` there will allow the rerender to restart without also having to rerender the children using the stale state. (You can read more about that [here](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes).)

packages/realm-react/src/UserProvider.tsx
elle-j: The `forceUpdate` is not needed here when the function is the dispatch one returned by `useReducer()`. If replacing the `useReducer()` with your suggested `useState()`/`useCallback()` combination, then it would be needed 👍

packages/realm-react/src/UserProvider.tsx
elle-j: See [previous comment](https://github.com/realm/realm-js/pull/6550#discussion_r1549491821).

packages/realm-react/src/UserProvider.tsx
elle-j: Nice to see that you also notice that we'd like to keep the function reference the same 👍

Since the suggestion splits up the steps a bit more, is it purely from an idiomatic stand point you prefer that implementation? 🙂

greenafrican: There's not much in this. I prefer the readability of not using useReducers here just to increase a counter, but it is an effective approach and stable function reference. Thanks for working on this issue 😄

kraenhansen: Note, this snippet was copied from the React documentation: https://legacy.reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdatea

kraenhansen: That being said, we could / should probably migrate this to use [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore). What do you think @elle-j - perhaps for a future PR?

elle-j: Seems like that could be something to look into for sure 👍

Issue #1099, opened 2017-06-27
Async transactions


astigsen:
Right now all write transactions are done synchronously, which is obviously not optimal as they end up blocking other events from taking place while the transaction is in progress (which involves IO).

The reason that transactions can not just be done asynchronously on background threads is that they involve javascript code (the actual code changing the data), so that code has to run on the main javascript thread. But doing a transaction is actually composed of multiple steps, some of which could done asynchronously:

1. Advance to latest version _(can be done async)_
2. Do changes _(has to be done in javascript on the main thread)_
3. Commit to disk  _(can be done async)_

**Advance to latest version**
Before starting a transaction you always have to update the state of the entire realm to the latest version. If you did not do this, the actual code being done in the transaction would risk working on stale data.

This is a potentially blocking operation, as another transaction may be in progress, so you have to wait for that to complete before you can advance to the latest version. But this waiting can be done asynchronously on a background thread, so that the javascript event loop can continue doing other work while waiting for the for the advance to complete.

**Do changes**
This is javascript code, so it has to run on the main javascript thread.

**Commit to disk**
This can again be done on a background thread so that the event loop can progress doing other stuff while it completes.

Moving these parts of the transaction process to background threads would not affect the throughput if the main constraint is transactions on the same realm (since the transactions would still have to wait for each other to complete), but it would free up the event loop to do other work while the transactions are in process, and if you are working with multiple realms they could also do transactions in parallel.

astigsen:
Note that this is different from the concept of async transactions in realm core, which is transactions that commit in a background thread but returns immediately and allows other transactions to start even if the previous one that not been fully committed to disk yet (meaning you can loose transactions not yet on disk in case of a crash).

This proposed way of doing async transactions is only async from the javascript perspective, so it is still guaranteed to be crash safe.

bigfish24:
@blagoev do you have any opinion on this?

blagoev:
In general all of the proposed is valid. We should strive to free the node event loop. I can't say on the implementation part how much time it will take to do it. Also we should measure what is the impact if any on the current implementation and how much we gain of making disk commit and updating the version to be async. If we have an exact scenario to measure against will help. 

Also there is work in progress that is kinda related to this in here: https://github.com/realm/realm-js/pull/1216
which may help measure transaction begin and commit times.

ivanquirino:
I would like to 
`realm.write()
 .then(() => { 
    realm.create(data)
    .then((newObect) => {} 
});`

[...]

PR #6552, opened 2024-03-14
Fix leaking listeners

bimusiek:
## What, How & Why?
If Realm is inTransaction and listener is created in the `setImmediate`, it won't be removed from listeners if tearDown is called quickly (which happens often if one listener triggers re-render of other listeners)

This closes https://github.com/realm/realm-js/issues/1099#issuecomment-1998188175

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm-react/src/cachedCollection.ts
+++ b/packages/realm-react/src/cachedCollection.ts
@@ -182,10 +182,12 @@ export function createCachedCollection<T extends Realm.Object<any>>({
     }
   };
 
+  let wasSetImmediate = false;
   if (!isDerived) {
     // If we are in a transaction, then push adding the listener to the event loop.  This will allow the write transaction to finish.
     // see https://github.com/realm/realm-js/issues/4375
     if (realm.isInTransaction) {
+      wasSetImmediate = true;
       setImmediate(() => {
         collection.addListener(listenerCallback, keyPaths);
       });
@@ -196,7 +198,13 @@ export function createCachedCollection<T extends Realm.Object<any>>({
 
   const tearDown = () => {
     if (!isDerived) {
-      collection.removeListener(listenerCallback);
+      if(wasSetImmediate) {
+        setImmediate(() => {
+          collection.removeListener(listenerCallback);
+        });
+      } else {
+        collection.removeListener(listenerCallback);
+      }
       objectCache.clear();
     }
   };


bimusiek:
@cla-bot check

nirinchev:
Hey @bimusiek thank you so much for the contribution. We'll try and get this reviewed shortly, but in the meantime, I was wondering if you have considered an approach similar to [C#'s cancellation tokens](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-8.0) - namely setting a flag that can be checked in the `setImmediate` callback that would skip the listener registration altogether. My understanding is that your use case involves rapid registrations/unregistrations and I can imagine there might be a moderate performance gain if we avoid the work altogether. JS is not my strongest language, but I'm thinking something that looks roughly like this:

```js
let isTornDown = false;

// ...
if (realm.isInTransaction) {
  setImmediate(() => {
    if (!isTornDown) {
      collection.addListener(listenerCallback, keyPaths);
    }
  });
}

// ...

const tearDown = () => {
  // ...
  collection.removeListener(listenerCallback);
  isTornDown = true;
};
```

bimusiek:
@nirinchev This is good idea, I will implement it soon. 

We have release the app to production with this change and it helped a lot. No more memory leaks due to listener callbacks leaks. However, the impact of this change was lesser than we thought, as we had another hook that was re-rendering a lot. Which triggered this scenario of leaking listener callback.

So at the end, this issue should not have huge impact. 

The re-rendering issue we had, was related to `Realm.BSON.ObjectID` which is changing reference even if the `hexString` value did not change. I would like to open discussion about making `ObjectID` cached so React won't rerender components if you use `ObjectID` inside the `deps`. Unfortunately this looks like deeper Realm integration and I have no idea how to approach this. 

For now, we have patched react typings to disallow `ObjectID` in the `deps` so Typescript notifies us if someone uses `ObjectID` directly. And we rewrote most of the code to operate on `hexString`. But this seems like a huge oversight if Realm wants to integrate nicely with React.

elle-j:
@bimusiek, we've added your `ObjectId` cache suggestion in [this ticket](https://github.com/realm/realm-js/issues/6583). Feel free to add anything else you think is relevant.

packages/realm-react/src/cachedCollection.ts
```diff
@@ -196,6 +197,9 @@ export function createCachedCollection<T extends Realm.Object<any>>({
 
   const tearDown = () => {
     if (!isDerived) {
+      if (setImmediateId) {
+        clearImmediate(setImmediateId);
```
elle-j, line 201: @bimusiek, please also `declare` the `clearImmediate` function in [runtime.d.ts](https://github.com/realm/realm-js/blob/main/packages/realm-react/src/runtime.d.ts), similarly to what we have for the [realm package](https://github.com/realm/realm-js/blob/fac140a239ac4df7e762e0427ad23c34d6520f14/packages/realm/src/runtime.d.ts#L22).

Also wondering if resetting the id here to `undefined` once cleared would be good.

bimusiek: Declaration added & clearing setImmediateId.

packages/realm-react/src/runtime.d.ts
```diff
@@ -25,3 +25,6 @@ declare type Console = {
 declare const console: Console;
 
 declare function setImmediate(cb: (...args: unknown[]) => void);
+
+declare type Timer = unknown;
+declare function clearImmediate(timer: Timer): void;
```
elle-j, line 30:
```suggestion
declare function clearImmediate(timer: Timer): void;

```

elle-j (approved):
Thank you for your contribution @bimusiek! We will link to your profile in the CHANGELOG entry for this fix 👍

PR #6581, opened 2024-04-02
Iterating coverage tests

kraenhansen:
## What, How & Why?

This is an attempt to fix the crashing coverage tests.


kraenhansen:
We found the root cause.

PR #6587, opened 2024-04-04
Replacing `changelog-header.sh` with a TypeScript equivalent

kraenhansen:
## What, How & Why?

Running the bash script locally errored with:

```
❯ ./scripts/changelog-header.sh
./scripts/changelog-header.sh: line 6: unexpected EOF while looking for matching `)'
```

Instead of spending time debugging it, I rewrote it in TypeScript.

### DIFF
--- a/.github/workflows/publish-package-release.yml
+++ b/.github/workflows/publish-package-release.yml
@@ -161,7 +161,6 @@ jobs:
       continue-on-error: true
 
     - name: Update Changelog
-      working-directory: 
       run: npm run prepend-changelog-header -- ${{ github.workspace }}/${{ env.PACKAGE_PATH }}/CHANGELOG.md
 
     - name: Create vNext PR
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-package-release.yml
```diff
@@ -161,8 +161,8 @@ jobs:
       continue-on-error: true
 
     - name: Update Changelog
-      working-directory: ${{ github.workspace }}/${{ env.PACKAGE_PATH }}
-      run: ${{ github.workspace }}/scripts/changelog-header.sh
+      working-directory:
```
elle-j, line 164: Is there a default when this is not defined? (And is it still needed here?)

CHANGELOG.header.md
elle-j, line 1: Do you think we should still omit `Breaking changes` from the template?

elle-j (approved):
Good initiative 🙂

.github/workflows/publish-package-release.yml
kraenhansen: Good catch, it should have been removed entirely
```suggestion
```

CHANGELOG.header.md
kraenhansen: I have no strong preference. I'd expect that to be less frequently needed than the others 🤔

PR #6592, opened 2024-04-04
RJS-2727: Re-using anonymous users by default in `realm-web`

kraenhansen:
## What, How & Why?

This closes RJS-2727 by adding the optional `reuse` argument to the anonymous credentials.
I've marked this as breaking, since this would previously default to creation of new users on subsequent logins with anonymous credentials.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/packages/realm-web/src/App.ts
+++ b/packages/realm-web/src/App.ts
@@ -202,6 +202,18 @@ export class App<
     credentials: Credentials,
     fetchProfile = true,
   ): Promise<User<FunctionsFactoryType, CustomDataType>> {
+    if (credentials.reuse) {
+      // TODO: Consider exposing providerName on "User" and match against that instead?
+      const existingUser = this.users.find((user) => user.providerType === credentials.providerType);
+      if (existingUser) {
+        this.switchUser(existingUser);
+        // If needed, fetch and set the profile on the user
+        if (fetchProfile) {
+          await existingUser.refreshProfile();
+        }
+        return existingUser;
+      }
+    }
     const response = await this.authenticator.authenticate(credentials);
     const user = this.createOrUpdateUser(response, credentials.providerType);
     // Let's ensure this will be the current user, in case the user object was reused.
--- a/packages/realm-web/src/Credentials.ts
+++ b/packages/realm-web/src/Credentials.ts
@@ -53,10 +53,11 @@ type SimpleObject = Record<string, unknown>;
 export class Credentials<PayloadType extends SimpleObject = SimpleObject> implements Realm.Credentials<PayloadType> {
   /**
    * Creates credentials that logs in using the [Anonymous Provider](https://docs.mongodb.com/realm/authentication/anonymous/).
+   * @param reuse - Reuse any existing anonymous user already logged in.
    * @returns The credentials instance, which can be passed to `app.logIn`.
    */
-  static anonymous(): Credentials<AnonymousPayload> {
-    return new Credentials<AnonymousPayload>("anon-user", "anon-user", {});
+  static anonymous(reuse = true): Credentials<AnonymousPayload> {
+    return new Credentials<AnonymousPayload>("anon-user", "anon-user", reuse, {});
   }
 
   /**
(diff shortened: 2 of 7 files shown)


packages/realm-web/src/App.ts
```diff
@@ -202,6 +202,18 @@ export class App<
     credentials: Credentials,
     fetchProfile = true,
   ): Promise<User<FunctionsFactoryType, CustomDataType>> {
+    if (credentials.reuse) {
+      // TODO: Consider exposing providerName on "User" and match against that instead?
+      const existingUser = this.users.find((user) => user.providerType === credentials.providerType);
```
elle-j, line 207: If there's an existing email/password user, and a *new* email/password user attempts a login, won't this also deem the new one as "existing"?

packages/realm-web/src/Credentials.ts
```diff
@@ -53,10 +53,11 @@ type SimpleObject = Record<string, unknown>;
 export class Credentials<PayloadType extends SimpleObject = SimpleObject> implements Realm.Credentials<PayloadType> {
   /**
    * Creates credentials that logs in using the [Anonymous Provider](https://docs.mongodb.com/realm/authentication/anonymous/).
+   * @param reuse - Reuse any existing anonymous user already logged in.
    * @returns The credentials instance, which can be passed to `app.logIn`.
    */
-  static anonymous(): Credentials<AnonymousPayload> {
-    return new Credentials<AnonymousPayload>("anon-user", "anon-user", {});
+  static anonymous(reuse = true): Credentials<AnonymousPayload> {
```
elle-j, line 59: Sidenote: For both realm and realm-web, perhaps a future overload (or replacement) with an options object would make the calling code a bit more readable 🙂

kneth, line 59: Current version is 2.0.0, and bumping to 3.0.0 might be drastic. If we say `reuse = false` and say we will switch to `reuse = true` in the future, we can avoid the bump.

The next big project for Realm Web will be enhanced MongoDB data access project, and it is - for me - unclear if it will come will any breaking changes.

kneth (approved):
If you feel strongly about the proposed breaking change, ignore my comment and merge!

packages/realm-web/src/App.ts
kraenhansen: Correct, and currently anonymous is the only credential supporting the `reuse` property.

packages/realm-web/src/Credentials.ts
nirinchev: We could also brand this as a bug fix as `reuse = true` was the intended behavior from the start and creating many users is most likely not what developers intend to do.

kraenhansen: Once reason I choose this is to align with the current API of the `realm` package:

https://github.com/realm/realm-js/blob/0bd8ee486780431594724e3ff4988aa0853e7a4e/packages/realm/src/app-services/Credentials.ts#L56

kraenhansen: I'd be okay with changing the default to `false`, but that would deviate from the "realm" package, which could potentially be confusing 🤔
https://github.com/realm/realm-js/blob/0bd8ee486780431594724e3ff4988aa0853e7a4e/packages/realm/src/app-services/Credentials.ts#L56

But I agree, we could bundle the change to `true` into a larger breaking change (possibly as part of the project to extract the App Services client).

packages/realm-web/src/App.ts
elle-j: Ah okay!

packages/realm-web/src/Credentials.ts
elle-j: Yup I figured 👍

kraenhansen: The docs from the `realm` package is a bit self-contradictory on this subject: https://www.mongodb.com/docs/realm-sdks/js/latest/classes/export_.Credentials.html#anonymous

The documentation on the function itself:
> Creates credentials for an anonymous user. These can only be used once - using them a second time will result in a different user being logged in. If you need to get a user that has already logged in with the Anonymous credentials, use [App.currentUser](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/export_.App-1.html#currentUser) or [App.allUsers](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/export_.App-1.html#allUsers).

The documentation of the optional argument:

> `reuse: boolean = true`
> Reuse any existing anonymous user already logged in.

kraenhansen: I moved it to "fixed".

PR #6593, opened 2024-04-04
RJS-2765: Upgrade BSON

### LINKED ISSUE #6561: Upgrade to BSON v6
The public types emitted for `bson.d.ts` are wrong:

```typescript
export declare namespace BSON {
    const ObjectId: typeof bson.ObjectID; // 👈 notice the capital D in ID
    type ObjectId = bson.ObjectId;
    const ObjectID: typeof bson.ObjectID;
    type ObjectID = bson.ObjectID;
    // ...
}
```

This is really weird since we're importing the right type from the namespace:

https://github.com/realm/realm-js/blob/1e1e48a0f906d471b4a046168ed8cd628ef2d70f/packages/realm/src/bson.ts#L27-L30

We should be able to upgrade to `bson@6` by now and this will fix the issue since that has removed `ObjectID` (capital D) entirely.

elle-j:
## What, How & Why?

Upgrading packages to BSON v6. There are several breaking changes ([v5](https://github.com/mongodb/js-bson/releases/tag/v5.0.0), [v6](https://github.com/mongodb/js-bson/releases/tag/v6.0.0)) in the BSON package, some of which affect us.

[WIP]
* Debugging some [null pointer exceptions](https://github.com/realm/realm-js/blob/7326f8787401fb08d8d4275a2df7f8620f1ecc24/integration-tests/tests/src/tests/sync/partition-values.ts#L96) related to [btoa](https://nodejs.org/api/buffer.html#bufferbtoadata) which is probably caused by using a Node engine <16 somewhere. The BSON package [removed the btoa shim](https://github.com/mongodb/js-bson/pull/590/commits/57248a9890d3f32d784faeb275f1ea05bf19ebe7) they were using when bumping the minimum Node version used.

This closes #6561

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
I've rebased on latest `main` to get b88a065cbd50cf1b84f381d1ab1a781d72bd4f2e and force-pushed to your branch 🤞

PR #6596, opened 2024-04-05
Removed the dependency on the clang-format package

kraenhansen:
## What, How & Why?

In alignment with https://github.com/realm/realm-core/pull/7554 this removes the root package's dependency on `clang-format`. Drafting this until the changes in the Core PR has been released.


kraenhansen:
I've included this in #6598.

Issue #6589, opened 2024-04-04
"free(): invalid pointer" when running on Ubuntu x86_64 and Node.js >= v20.12.0


kraenhansen:
> [!NOTE]
> If you're seeing "free(): invalid pointer" when running on Ubuntu x86_64, a valid workaround is downgrading Node.js to v20.11.1

### Description

See for more information https://github.com/realm/realm-core/issues/7604.

The coverage tests started crashing around March 26th, by the release of [Node.js v20.12.0](https://github.com/nodejs/node/releases/tag/v20.12.0).

Here's an example of a failing run: https://github.com/realm/realm-js/actions/runs/8552406490/job/23433443919?pr=6587

There seem to be two issues:
- The crash itself, which originates from Core (see #7604)
- The `npm run ci:coverage` process exits cleanly, which is unexpected from a crash.

I've reproduced the crash on these Realm JS commits:
- 2796745c674732bdfc45d460924c6dfefc84c720 (Realm Core 14.5.1)
- e15db6016513ac37996286abde4c44a14e98268c (Realm Core 13.26.0)
- 89687764c65f17b0a43b6a20d80a07be8063b0ce (Realm Core 13.25.1)
- 42651be546db8ea6830e9da30a11673c0bfce8c8 (Realm Core 13.25.0)

### Stacktrace & log output

```shell
free(): invalid pointer
```

kraenhansen:
This *might* have been the cause: https://github.com/realm/realm-core/issues/7434 🤞

kraenhansen:
I've created https://github.com/realm/realm-core/issues/7604 with more information on the root cause of this crash.

kraenhansen:
Regarding the uncorrect exit code when the subprocess spawed by `nyc` crashes, I've found these issues and PR that I believe is related:
- https://github.com/istanbuljs/nyc/issues/798
- https://github.com/istanbuljs/nyc/issues/1535
- https://github.com/istanbuljs/nyc/pull/1546

PR #6598, opened 2024-04-06
RJS-2792: Upgrade to Realm Core v14.5.0

kraenhansen:
## What, How & Why?

This upgrades to Realm Core v14.5.0.

In alignment with https://github.com/realm/realm-core/pull/7554 this removes the root package's dependency on `clang-format`.

Fixes https://github.com/realm/realm-js/issues/6544

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -1,0 +1,25 @@
+# This file controls the behavior of Trunk: https://docs.trunk.io/cli
+# To learn more about the format of this file, see https://docs.trunk.io/reference/trunk-yaml
+version: 0.1
+cli:
+  version: 1.21.0
+# Trunk provides extensibility via plugins. (https://docs.trunk.io/plugins)
+plugins:
+  sources:
+    - id: trunk
+      ref: v1.4.5
+      uri: https://github.com/trunk-io/plugins
+# Many linters and tools depend on runtimes - configure them here. (https://docs.trunk.io/runtimes)
+runtimes:
+  enabled:
+# This is the section where you manage your linters. (https://docs.trunk.io/check/configuration)
+lint:
+  enabled:
+    - clang-format@16.0.3
+actions:
+  disabled:
+    - trunk-announce
+    - trunk-check-pre-push
+    - trunk-fmt-pre-commit
+  enabled:
+    - trunk-upgrade-available
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -10,8 +10,8 @@ plugins:
       ref: v1.4.5
       uri: https://github.com/trunk-io/plugins
 # Many linters and tools depend on runtimes - configure them here. (https://docs.trunk.io/runtimes)
-runtimes:
-  enabled:
+# runtimes:
+#   enabled:
 # This is the section where you manage your linters. (https://docs.trunk.io/check/configuration)
 lint:
   enabled:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 > This version bumps the Realm file format to version 24. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v6.0.0, might not be upgradeable. **Only Realm Studio 15.0.0 or later** will be able to open the new file format.
 
 ### Enhancements
+* The base URL used to communicate with the Atlas App Services was updated changed from "https://realm.mongodb.com" to "https://services.cloud.mongodb.com". ([realm/realm-core#7534](https://github.com/realm/realm-core/pull/7534)).
 * Updated bundled OpenSSL version to 3.2.0. ([realm/realm-core#7303](https://github.com/realm/realm-core/pull/7303))
 * Improved performance of object notifiers with complex schemas by ~20%. ([realm/realm-core#7424](https://github.com/realm/realm-core/pull/7424))
 * Improved performance with very large number of notifiers by ~75%. ([realm/realm-core#7424](https://github.com/realm/realm-core/pull/7424))
@@ -31,6 +32,9 @@
 * Fixed opening a Realm with cached user while offline results in fatal error and session does not retry connection. ([#6554](https://github.com/realm/realm-js/issues/6554) and [#6558](https://github.com/realm/realm-js/issues/6558), since v12.6.0)
 * Fixed sorting order of strings to use standard unicode codepoint order instead of grouping similar English letters together. A noticeable change will be from "aAbBzZ" to "ABZabz". ([realm/realm-core#2573](https://github.com/realm/realm-core/issues/2573))
 * `data` and `string` are now strongly typed for comparisons and queries. This change is especially relevant when querying for a string constant on a Mixed property, as now only strings will be returned. If searching for `data` is desired, then that type must be specified by the constant. In RQL the new way to specify a binary constant is to use `mixed = bin('xyz')` or `mixed = binary('xyz')`. ([realm/realm-core#6407](https://github.com/realm/realm-core/issues/6407))
+
+* Fixed diverging history due to a bug in the replication code when setting default null values (embedded objects included) ([realm/realm-core#7536](https://github.com/realm/realm-core/issues/7536)).
+* Null pointer exception may be triggered when logging out and async commits callbacks not executed ([realm/realm-core#7434](https://github.com/realm/realm-core/issues/7434), since v12.6.0)
 
 ### Compatibility
 * React Native >= v0.71.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 > This version bumps the Realm file format to version 24. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v6.0.0, might not be upgradeable. **Only Realm Studio 15.0.0 or later** will be able to open the new file format.
 
 ### Enhancements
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 > This version bumps the Realm file format to version 24. It is not possible to downgrade to earlier versions. Older files will automatically be upgraded to the new file format. Files created by Realm JavaScript prior to v6.0.0, might not be upgradeable. **Only Realm Studio 15.0.0 or later** will be able to open the new file format.
 
 ### Enhancements
+* The base URL used to communicate with the Atlas App Services was updated changed from "https://realm.mongodb.com" to "https://services.cloud.mongodb.com". ([realm/realm-core#7534](https://github.com/realm/realm-core/pull/7534)).
```
kneth, line 7: I think it should go under "Internal"

CHANGELOG.md
```diff
@@ -32,13 +32,17 @@
 * Fixed sorting order of strings to use standard unicode codepoint order instead of grouping similar English letters together. A noticeable change will be from "aAbBzZ" to "ABZabz". ([realm/realm-core#2573](https://github.com/realm/realm-core/issues/2573))
 * `data` and `string` are now strongly typed for comparisons and queries. This change is especially relevant when querying for a string constant on a Mixed property, as now only strings will be returned. If searching for `data` is desired, then that type must be specified by the constant. In RQL the new way to specify a binary constant is to use `mixed = bin('xyz')` or `mixed = binary('xyz')`. ([realm/realm-core#6407](https://github.com/realm/realm-core/issues/6407))
 
+* Fixed diverging history due to a bug in the replication code when setting default null values (embedded objects included) ([realm/realm-core#7536](https://github.com/realm/realm-core/issues/7536)).
```
kneth, line 35:
```suggestion
* Fixed diverging history due to a bug in the replication code when setting default null values (embedded objects included). ([realm/realm-core#7536](https://github.com/realm/realm-core/issues/7536))
```

CHANGELOG.md
```diff
@@ -32,13 +32,17 @@
...
+* Null pointer exception may be triggered when logging out and async commits callbacks not executed ([realm/realm-core#7434](https://github.com/realm/realm-core/issues/7434), since v12.6.0)
```
kneth, line 36:
```suggestion
* Null pointer exception may be triggered when logging out and async commits callbacks not executed. ([realm/realm-core#7434](https://github.com/realm/realm-core/issues/7434), since v12.6.0)
```

CHANGELOG.md
```diff
@@ -32,13 +32,17 @@
...
+
 ### Compatibility
 * React Native >= v0.71.4
 * Realm Studio v15.0.0.
 * File format: generates Realms with format v24 (reads and upgrades file format v10 or later).
 
 ### Internal
 * Upgraded Realm Core from v13.26.0 to v14.4.1. ([#6499](https://github.com/realm/realm-js/issues/6499), [#6541](https://github.com/realm/realm-js/issues/6541), [#6568](https://github.com/realm/realm-js/issues/6568), and [#6572](https://github.com/realm/realm-js/issues/6572))
```
kneth, line 44:
```suggestion
* Upgraded Realm Core from v13.26.0 to v14.5.0. ([#6499](https://github.com/realm/realm-js/issues/6499), [#6541](https://github.com/realm/realm-js/issues/6541), [#6568](https://github.com/realm/realm-js/issues/6568), [#6572](https://github.com/realm/realm-js/issues/6572), and [#6599](https://github.com/realm/realm-js/issues/6599))
```

package.json
```diff
@@ -9,8 +9,9 @@
     "bundle": "wireit",
     "lint": "wireit",
     "lint:fix": "wireit",
-    "lint:cpp": "find src packages/realm/bindgen/src -name '*.cpp' -or -name '*.h' -or -name '*.hpp' | xargs clang-format --dry-run --Werror",
-    "lint:cpp:fix": "find src packages/realm/bindgen/src -name '*.cpp' -or -name '*.h' -or -name '*.hpp' | xargs clang-format -i",
+    "trunk:check": "trunk check",
```
kneth, line 12: Any reason for not calling it `lint:cpp`?

package.json
```diff
@@ -9,8 +9,9 @@
...
+    "trunk:check-all": "trunk check --all",
+    "trunk:fmt": "trunk fmt",
```
kneth, line 14: Any reason for not calling it `lint:cpp:fix`?

kneth (approved):

.trunk/trunk.yaml
```diff
@@ -0,0 +1,25 @@
...
+  version: 1.21.0
+# Trunk provides extensibility via plugins. (https://docs.trunk.io/plugins)
+plugins:
+  sources:
+    - id: trunk
+      ref: v1.4.5
+      uri: https://github.com/trunk-io/plugins
+# Many linters and tools depend on runtimes - configure them here. (https://docs.trunk.io/runtimes)
+runtimes:
+  enabled:
```
elle-j, line 14:
```suggestion
# runtimes:
#   enabled:
```

CHANGELOG.md
```diff
@@ -32,13 +32,17 @@
...
+* The base URL used to communicate with the Atlas App Services was updated changed from "https://realm.mongodb.com" to "https://services.cloud.mongodb.com". ([realm/realm-core#7534](https://github.com/realm/realm-core/pull/7534)).
```
elle-j, line 45:
```suggestion
* The base URL used to communicate with the Atlas App Services was changed from "https://realm.mongodb.com" to "https://services.cloud.mongodb.com". ([realm/realm-core#7534](https://github.com/realm/realm-core/pull/7534)).
```

package.json
elle-j: I think it might make more sense to keep the name more generic, like the old one Kenneth is referring to here, to not have the name tied to the underlying tool. Also, since `trunk` is not super obvious what it does 😉  (And if so, similarly for the `name` in the [workflow](https://github.com/realm/realm-js/blob/bb64dd0ffe1d6ed452203f09c2227ef7f331a9b8/.github/workflows/pr-linting.yml#L23).)

packages/realm/bindgen/src/templates/jsi.ts
```diff
@@ -486,7 +486,6 @@ function convertToJsi(addon: JsiAddon, type: Type, expr: string): string {
           assert.fail("Should never see AsyncResult here");
       }
       assert.fail(`unknown template ${type.name}`);
-      break;
```
elle-j, line 489: Do you want to add a `/* eslint-disable-next-line no-fallthrough */` for the removed `break` statements? Or if possible, I think it'd read better if keeping the `break` statements but instead silence the `Unreachable code detected`. (Similarly for the Node template.) What do you think?

packages/realm/bindgen/src/trunk-formatter.ts
elle-j, line 1: What do you think about making the file name less coupled to the tool used? So just `formatter.ts`. Possibly the exported variable as well?

elle-j (approved):
Nice, will be interesting to get a feel for `trunk`. Left some questions and suggestions.

packages/realm/bindgen/src/templates/jsi.ts
kraenhansen: I have no idea why this change got applied 👀

packages/realm/bindgen/src/trunk-formatter.ts
kraenhansen: We could merge the `trunk-formatter.ts` with the `eslint-formatter.ts` 🤔

package.json
kraenhansen: That would encode an implicit expectation that the `.trunk/trunk.yaml` only enables the `clang-format` linter, introducing a coupling between this name and the configuration file, which I'd like to avoid.

package.json
kraenhansen: It would encode an implicit expectation that the `.trunk/trunk.yaml` only enables linters for "cpp" files, introducing a coupling between this name and the configuration file, which I'd like to avoid.

But I agree `trunk` in itself doesn't make too much sense, how about `lint:trunk`? (lint being the semantic thing and trunk the tool - or perhaps I should just move this into the "command" for the existing "lint" script 🤔

packages/realm/bindgen/src/templates/jsi.ts
kraenhansen: It seems an update to TypeScript recognize the `break` as undead code and removes it in my editor on save, because of the `never` return-type from `assert.fail` 🤔

packages/realm/bindgen/src/trunk-formatter.ts
kraenhansen: I ended up turning this into a `return assert.fail(...)` instead 👍

package.json
elle-j: Moving it into the existing script seems like the better option then 👍

packages/realm/bindgen/src/templates/jsi.ts
elle-j: I think [this comment](https://github.com/realm/realm-js/pull/6598#discussion_r1558163311) was meant for this thread 🙂 Good solution 👍

packages/realm/bindgen/src/trunk-formatter.ts
elle-j: Was thinking that we only had one formatter now but yeah we have the eslint as well. Sounds reasonable to merge the files though so that they're grouped under "formatters".

package.json
kraenhansen: See the answer above. I've decided to call trunk in the existing `lint` and `lint:fix` instead.

kraenhansen:
Coverage test is still crashing (#6589) but that's not new - I'll go ahead and merge with it.

PR #6612, opened 2024-04-12
RJS-2784: Clear internal state to avoid crashes when React Native app is reloaded

### LINKED ISSUE #6579: React Native Android application crashing when using syncSession.addConnectionNotification

Always

### Description

The react native application is crashing when there is an active syncSession.addConnectionNotification and the app gets reloaded

### Stacktrace & log output

```shell
Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30 in tid 8357 (mqt_js), pid 8292 (om.synctutorial)
2024-03-26 17:24:24.007  8292-8292  unknown:Re...nceManager com.synctutorial                     E  destroyRootView called
2024-03-26 17:24:24.009  8292-8292  unknown:Re...nceManager com.synctutorial                     E  destroyRootView called, unmountReactApplication

Cmdline: com.synctutorial
2024-03-26 17:24:25.208  8538-8538  DEBUG                   crash_dump64                         A  pid: 8292, tid: 8357, name: mqt_js  >>> com.synctutorial <<<
2024-03-26 17:24:25.208  8538-8538  DEBUG                   crash_dump64                         A        #00 pc 00000000003bfd00  /data/app/~~kwxPndNsCq5S-a142smnIA==/com.synctutorial-Jxnmi56nwT1pLTopnsEDBg==/lib/arm64/librealm.so (BuildId: edf6f0d435afef43e7dfceba74cfdd58b0289992)
2024-03-26 17:24:25.208  8538-8538  DEBUG                   crash_dump64                         A        #01 pc 00000000003c061c  /data/app/~~kwxPndNsCq5S-a142smnIA==/com.synctutorial-Jxnmi56nwT1pLTopnsEDBg==/lib/arm64/librealm.so (BuildId: edf6f0d435afef43e7dfceba74cfdd58b0289992)
2024-03-26 17:24:25.208  8538-8538  DEBUG                   crash_dump64                         A        #02 pc 000000000041c924  /data/app/~~kwxPndNsCq5S-a142smnIA==/com.synctutorial-Jxnmi56nwT1pLTopnsEDBg==/lib/arm64/librealm.so (BuildId: edf6f0d435afef43e7dfceba74cfdd58b0289992)
2024-03-26 17:24:25.208  8538-8538  DEBUG                   crash_dump64                         A        #03 pc 000000000041d634  /data/app/~~kwxPndNsCq5S-a142smnIA==/com.synctutorial-Jxnmi56nwT1pLTopnsEDBg==/lib/arm64/librealm.so (BuildId: edf6f0d435afef43e7dfceba74cfdd58b0289992)
```

Yes, always 

### Reproduction Steps

Using this branch: https://github.com/Spring-Global/realm-react-native-sample/tree/realm-crash-callback

1. Run the application on android emulator
2. Refresh the react native app with cmd+r

### Version

realm@12.6.2

Atlas Device Sync

### Are you using encryption?

No

Android - Emulator API 34

### Build environment

### Cocoapods version

_No response_

kneth:
## What, How & Why?

This closes #6579 

During development it is common to reload a React Native app. As it is a rough process, it is important to clear the previous state - including listeners. Realm Core provides some low-level method to clear many internal data structures, and by invoking these when the SDK is initialized (loaded in the app), any state from the previous app instance is removed.

Testing is difficult, and I have only tried it manually in a app running locally on my computer.

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 * `data` and `string` are now strongly typed for comparisons and queries. This change is especially relevant when querying for a string constant on a Mixed property, as now only strings will be returned. If searching for `data` is desired, then that type must be specified by the constant. In RQL the new way to specify a binary constant is to use `mixed = bin('xyz')` or `mixed = binary('xyz')`. ([realm/realm-core#6407](https://github.com/realm/realm-core/issues/6407))
 * Fixed diverging history due to a bug in the replication code when setting default null values (embedded objects included). ([realm/realm-core#7536](https://github.com/realm/realm-core/issues/7536))
 * Null pointer exception may be triggered when logging out and async commits callbacks not executed. ([realm/realm-core#7434](https://github.com/realm/realm-core/issues/7434), since v12.6.0)
-* Fixed a bug when caused crashes when reloading React Native apps. ([#6579](https://github.com/realm/realm-js/issues/6579), since v12.0.0)
+* Fixed a bug which caused crashes when reloading React Native apps. ([#6579](https://github.com/realm/realm-js/issues/6579), since v12.0.0)
 
 ### Compatibility
 * React Native >= v0.71.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 * `data` and `string` are now strongly typed for comparisons and queries. This change is especially relevant when querying for a string constant on a Mixed property, as now only strings will be returned. If searching for `data` is desired, then that type must be specified by the constant. In RQL the new way to specify a binary constant is to use `mixed = bin('xyz')` or `mixed = binary('xyz')`. ([realm/realm-core#6407](https://github.com/realm/realm-core/issues/6407))
 * Fixed diverging history due to a bug in the replication code when setting default null values (embedded objects included). ([realm/realm-core#7536](https://github.com/realm/realm-core/issues/7536))
 * Null pointer exception may be triggered when logging out and async commits callbacks not executed. ([realm/realm-core#7434](https://github.com/realm/realm-core/issues/7434), since v12.6.0)
+* Fixed a bug when caused crashes when reloading React Native apps. ([#6579](https://github.com/realm/realm-js/issues/6579), since v12.0.0)
 
 ### Compatibility
 * React Native >= v0.71.4
--- a/packages/realm/src/Realm.ts
+++ b/packages/realm/src/Realm.ts
@@ -1360,6 +1360,10 @@ export namespace Realm {
   export import PushClient = internal.PushClient;
 }
 
-//Set default logger and log level.
+// Clear the internal state to prevent crashes when reloading on React Native
+binding.RealmCoordinator.clearAllCaches();
+binding.App.clearCachedApps();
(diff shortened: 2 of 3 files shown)


packages/realm/src/Realm.ts
```diff
@@ -1360,6 +1360,10 @@ export namespace Realm {
   export import PushClient = internal.PushClient;
 }
 
-//Set default logger and log level.
+// Clear the internal state to prevent crashes when reloading on React Native
+binding.RealmCoordinator.clearAllCaches();
+binding.App.clearCachedApps();
```
kraenhansen, line 1365: Interesting approach 👍 I suppose there's a time between the runtime being torn down until this code runs, where there could potentially fire an event which would crash the app 🤔 But ... it's probably a lot better than nothing.

I wonder if it would make sense to simply call the `Realm.shutdown` instead? That would also cancel outstanding async open tasks, etc.

Also, since this is relevant only for React Native, it might make sense to move it into the platform specific wrapper:
https://github.com/realm/realm-js/blob/5dab79824941ccbb5827603eedcf25537117d5ac/packages/realm/src/platform/react-native/index.ts#L25-L26

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 * `data` and `string` are now strongly typed for comparisons and queries. This change is especially relevant when querying for a string constant on a Mixed property, as now only strings will be returned. If searching for `data` is desired, then that type must be specified by the constant. In RQL the new way to specify a binary constant is to use `mixed = bin('xyz')` or `mixed = binary('xyz')`. ([realm/realm-core#6407](https://github.com/realm/realm-core/issues/6407))
 * Fixed diverging history due to a bug in the replication code when setting default null values (embedded objects included). ([realm/realm-core#7536](https://github.com/realm/realm-core/issues/7536))
 * Null pointer exception may be triggered when logging out and async commits callbacks not executed. ([realm/realm-core#7434](https://github.com/realm/realm-core/issues/7434), since v12.6.0)
+* Fixed a bug when caused crashes when reloading React Native apps. ([#6579](https://github.com/realm/realm-js/issues/6579), since v12.0.0)
```
elle-j, line 40:
```suggestion
* Fixed a bug which caused crashes when reloading React Native apps. ([#6579](https://github.com/realm/realm-js/issues/6579), since v12.0.0)
```

packages/realm/src/Realm.ts
kneth: > Interesting approach

No, just a hack 😄 

> simply call the `Realm.shutdown` instead

I had an issue with `Realm.shutdown` since `RealmProgressPromise` is undefined.

> move it into the platform specific wrapper

I agree.

elle-j (commented):
I've tested this as well with the RN test app (on iOS) and it seems to work. I'm not all that familiar with the ins and outs of this change so I'll leave it up to Kraen to approve 👍

kneth:
Coverage tests crashed with "free(): invalid pointer", and we should investigate it

PR #6617, opened 2024-04-15
RJS-2397: Collections in mixed - synced tests

papafe:
This PR adds tests for synced collections in mixed

## ☑️ ToDos
* [x] 🚦 Tests

### DIFF
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -69,9 +69,10 @@ export function openRealmHook(config: OpenRealmConfiguration = {}) {
 export function closeThisRealm(this: RealmContext & Mocha.Context): void {
   if (this.closeRealm) {
     this.closeRealm({ clearTestState: true, deleteFile: true });
+  } else {
+    // Clearing the test state to ensure the sync session gets completely reset and nothing is cached between tests
+    Realm.clearTestState();
   }
-  // Clearing the test state to ensure the sync session gets completely reset and nothing is cached between tests
-  Realm.clearTestState();
 }
 
 export function openRealmBeforeEach(config: OpenRealmConfiguration = {}): void {
@@ -84,8 +85,33 @@ export function openRealmBefore(config: OpenRealmConfiguration = {}): void {
   after("closeRealmAfter", closeThisRealm);
 }
 
-export function setupRealmHook() {
-  return async function openRealmHandler(this: UserContext & Mocha.Context): Promise<void> {
-    this.longTimeout();
+export async function setupRealmHook(this: AppContext & MultiRealmContext): Promise<void> {
+  this.openedInfo = [];
+
+  this.getRealm = async (config: OpenRealmConfiguration): Promise<Realm> => {
+    const user = await this.getUser(Realm.Credentials.anonymous(false));
+    const realmAndConfig = await openRealm(config, user);
+    this.openedInfo.push(realmAndConfig);
+    return realmAndConfig.realm;
   };
+
+  this.closeAllRealms = async () => {
+    this.openedInfo?.forEach(({ realm, config }) => {
+      if (!realm?.isClosed) {
+        realm.close();
+      }
+      Realm.deleteFile(config);
+    });
+
+    Realm.clearTestState();
+  };
+}
+
+export function closeMultiRealms(this: AppContext & MultiRealmContext): void {
+  this.closeAllRealms?.();
+}
+
+export function setupMultiRealmsBeforeAndAfterEach(): void {
+  beforeEach("openMultiRealmsBeforeEach", setupRealmHook);
+  afterEach("closeMultiRealmsAfterEach", closeMultiRealms);
 }
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -96,12 +96,12 @@ export async function setupRealmHook(this: AppContext & MultiRealmContext): Prom
   };
 
   this.closeAllRealms = () => {
-    this.openedInfo?.forEach(({ realm, config }) => {
-      if (!realm?.isClosed) {
-        realm.close();
-      }
-      Realm.deleteFile(config);
-    });
+    // this.openedInfo?.forEach(({ realm, config }) => {
+    //   if (!realm?.isClosed) {
+    //     realm.close();
+    //   }
+    //   Realm.deleteFile(config);
+    // });
 
     Realm.clearTestState();
   };
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -95,7 +95,7 @@ export async function setupRealmHook(this: AppContext & MultiRealmContext): Prom
     return realmAndConfig.realm;
   };
 
-  this.closeAllRealms = async () => {
+  this.closeAllRealms = () => {
     this.openedInfo?.forEach(({ realm, config }) => {
       if (!realm?.isClosed) {
         realm.close();
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -83,3 +83,9 @@ export function openRealmBefore(config: OpenRealmConfiguration = {}): void {
   before(openRealmBefore.name, openRealmHook(config));
   after("closeRealmAfter", closeThisRealm);
 }
+
+export function setupRealmHook() {
+  return async function openRealmHandler(this: UserContext & Mocha.Context): Promise<void> {
+    this.longTimeout();
+  };
+}
--- a/integration-tests/tests/src/hooks/open-realm-before.ts
+++ b/integration-tests/tests/src/hooks/open-realm-before.ts
@@ -84,34 +84,3 @@ export function openRealmBefore(config: OpenRealmConfiguration = {}): void {
   before(openRealmBefore.name, openRealmHook(config));
   after("closeRealmAfter", closeThisRealm);
 }
-
-export async function setupRealmHook(this: AppContext & MultiRealmContext): Promise<void> {
-  this.openedInfo = [];
-
-  this.getRealm = async (config: OpenRealmConfiguration): Promise<Realm> => {
-    const user = await this.getUser(Realm.Credentials.anonymous(false));
-    const realmAndConfig = await openRealm(config, user);
-    this.openedInfo.push(realmAndConfig);
-    return realmAndConfig.realm;
-  };
-
-  this.closeAllRealms = () => {
-    // this.openedInfo?.forEach(({ realm, config }) => {
-    //   if (!realm?.isClosed) {
-    //     realm.close();
-    //   }
-    //   Realm.deleteFile(config);
-    // });
-
-    Realm.clearTestState();
-  };
-}
-
-export function closeMultiRealms(this: AppContext & MultiRealmContext): void {
-  this.closeAllRealms?.();
-}
-
-export function setupMultiRealmsBeforeAndAfterEach(): void {
-  beforeEach("openMultiRealmsBeforeEach", setupRealmHook);
-  afterEach("closeMultiRealmsAfterEach", closeMultiRealms);
-}
--- a/integration-tests/tests/src/tests/sync/mixed.ts
+++ b/integration-tests/tests/src/tests/sync/mixed.ts
@@ -23,7 +23,16 @@ import { importAppBefore, authenticateUserBefore, openRealmBefore } from "../../
 import { itUploadsDeletesAndDownloads } from "./upload-delete-download";
 import { buildAppConfig } from "../../utils/build-app-config";
 
-type Value = Realm.Mixed | ((realm: Realm) => Realm.Mixed) | ((realm: Realm) => { values: Mixed; expected: Mixed });
+/**
+ * A function type that generates values to inserted and expected values used with the default tester.
+ * The input realm is necessary to create and add objects to the realm before using them in the tests.
+ * The distinction between "values" and "expected" is necessary because most tests here close the realm after using
+ * the objects in "values", so they can't be used with the default tester (that happens at a later time).
+ * "expected" contains just an object instead of a RealmObject that can be used for testing.
+ */
+type ValueAndExpectedGenerator = (realm: Realm) => { values: Mixed; expected: Mixed };
+
+type Value = Realm.Mixed | ((realm: Realm) => Realm.Mixed) | ValueAndExpectedGenerator;
 type ValueTester = (actual: Realm.Mixed, inserted: Realm.Mixed) => void;
 
 class MixedClass extends Realm.Object<MixedClass> {
--- a/integration-tests/tests/src/tests/sync/mixed.ts
+++ b/integration-tests/tests/src/tests/sync/mixed.ts
@@ -333,6 +333,7 @@ describe.only("mixed synced", () => {
     importAppBefore(buildAppConfig("with-pbs").anonAuth().partitionBasedSync());
     describeTypes(false);
   });
+
   describe.skipIf(environment.skipFlexibleSync, "flexible sync roundtrip", function () {
     this.longTimeout();
     importAppBefore(buildAppConfig("with-flx").anonAuth().flexibleSync());
@@ -360,12 +361,12 @@ describe.only("mixed synced", () => {
       });
     }
 
-    it.only("value change", async function (this: Mocha.Context & AppContext & MultiRealmContext) {
+    it.skip("value change", async function (this: Mocha.Context & AppContext & MultiRealmContext) {
       const realm1 = await this.getRealm(realmConfig);
       await setupTest(realm1, true);
 
       const valuesToInsert = realm1.write(() => {
-        return getMixedList(realm1);
+        return getNestedMixedList(realm1);
       });
 
       const obId = new Realm.BSON.ObjectID();
@@ -407,16 +408,14 @@ describe.only("mixed synced", () => {
       }
     });
 
-    it.skip("list modifications", async function (this: Mocha.Context & AppContext & MultiRealmContext) {
-      const valuesToInsert = [
-        1,
-        "string",
-        [22, "test"],
-        { key1: 10, key2: new Date(1620768552979) },
-        new Realm.BSON.ObjectID(),
-      ];
+    it.only("list modifications", async function (this: Mocha.Context & AppContext & MultiRealmContext) {
       const realm1 = await this.getRealm(realmConfig);
       await setupTest(realm1, true);
+
+      const valuesToInsert = realm1.write(() => {
+        return getNestedMixedList(realm1);
+      });
+
       const obId = new Realm.BSON.ObjectID();
       const obj1 = realm1.write(() => {
         return realm1.create(MixedClass, {
@@ -424,9 +423,11 @@ describe.only("mixed synced", () => {
           value: [],
         });
       });
+
       await realm1.syncSession?.uploadAllLocalChanges();
       const realm2 = await this.getRealm(realmConfig);
       await setupTest(realm2, true);
+
       const obj2 = await new Promise<MixedClass>((resolve) => {
         realm2
           .objects<MixedClass>("MixedClass")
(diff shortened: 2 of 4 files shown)


integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -48,35 +86,34 @@ function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
...
   function performTest(actual: Realm.Mixed, inserted: Realm.Mixed) {
-    const result = testValue(actual, inserted);
-    if (typeof result === "boolean") {
-      expect(result).equals(true, `${testValue} failed!`);
-    }
+    valueTester(actual, inserted);
   }
 
-  // TODO: This might be a useful utility
+  // TODO: This might be a useful utility  //Should we keep this around if not used?
```
kneth, line 109: If not used, it should be safe to remove it

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -16,28 +16,66 @@
...
 
 /**
  * The default tester of values.
  * @param actual The value downloaded from the server.
  * @param inserted The value inserted locally before upload.
  */
 function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
-  expect(actual).equals(inserted);
+  if (actual instanceof Realm.List) {
+    const insertedVal = inserted as Realm.Mixed[]; //TODO I should remove all of these "as" (?)
```
kneth, line 54: Or should the type for `inserted` be `any` instead of `Realm.Mixed`?

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -138,7 +161,7 @@ function describeRoundtrip({
       expect(typeof obj).equals("object");
       // Test the single value
       performTest(obj.value, this.value);
-      // Test the list of values
+      // Test the list of values  //TODO Maybe we don't need this?
```
kneth, line 164: Isn't it worth testing that what we put in is also what we get out?

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -224,12 +239,79 @@ function describeTypes(flexibleSync: boolean) {
...
+        value: nextedMixedDict,
+        useFlexibleSync: true,
+      });
+    });
+  }
 }
 
-describe("mixed", () => {
+describe("mixed synced", () => {
+  //TODO Should we keep this around?
```
kneth, line 314: As long as we support PBS, I think we should keep it

kneth (approved):

kneth (commented):
Thank you.

Do you plan to add more tests?

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -152,18 +175,18 @@ function describeRoundtrip({
 function describeTypes(flexibleSync: boolean) {
```
kraenhansen, line 175: If you rename it in one place, do it here as well, to avoid the `useFlexibleSync: flexibleSync` below.

```suggestion
function describeTypes(useFlexibleSync: boolean) {
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -224,12 +239,79 @@ function describeTypes(flexibleSync: boolean) {
...
+        data: data.buffer,
+      };
+
+      const nestedMixedList: Mixed[] = [...mixedList, mixedList, mixedDict];
+
+      const nextedMixedDict = {
+        ...mixedDict,
+        innerList: mixedList,
+        innerDict: mixedDict,
+      };
```
kraenhansen, line 284: Could you inline these below and use the `satisfies` TS keyword instead?

integration-tests/tests/src/tests/sync/mixed.ts
papafe: I agree, I'll remove it

integration-tests/tests/src/tests/sync/mixed.ts
papafe: I misunderstood a little bit the test and why the list was used. Now I understand what it was done (testing if a mixed value inside a list gets correctly around) and I'll add a similar one for a dictionary)

integration-tests/tests/src/tests/sync/mixed.ts
papafe: Agreed

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -16,28 +16,66 @@
...
+}
 
 /**
  * The default tester of values.
  * @param actual The value downloaded from the server.
  * @param inserted The value inserted locally before upload.
  */
 function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
-  expect(actual).equals(inserted);
+  if (actual instanceof Realm.List) {
```
kraenhansen, line 53: If `actual` is the wrong type (the database is broken), then this might be asserting the wrong thing and we cannot rely on the type of `inserted` since that's has the unmanaged value 🤔 I believe this was why the test functions were separate. Also it allows for more detailed testing and multiple tests per type.

integration-tests/tests/src/tests/sync/mixed.ts
papafe: That could also work. Not sure what is the preferred way for this here

integration-tests/tests/src/tests/sync/mixed.ts
papafe: Makes sense. I'll reinstantiate the local test functions that were used in the tests and keep this for the mixed testing

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: Never `any` ... `unknown` at best.

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: Never mind - I didn't notice that you were referring between these 🙈

integration-tests/tests/src/hooks/open-realm-before.ts
```diff
@@ -83,3 +84,34 @@ export function openRealmBefore(config: OpenRealmConfiguration = {}): void {
...
+  this.openedInfo = [];
+
+  this.getRealm = async (config: OpenRealmConfiguration): Promise<Realm> => {
+    const user = await this.getUser(Realm.Credentials.anonymous(false));
+    const realmAndConfig = await openRealm(config, user);
+    this.openedInfo.push(realmAndConfig);
+    return realmAndConfig.realm;
+  };
+
+  this.closeAllRealms = async () => {
```
kneth, line 98: `Realm.clearTestState()` and `Realm.shutdown()` are closing all open Realms.

papafe: From what I can understand from the body of the function, then just calling `Realm.clearTestState` would be enough here, right?

If so, maybe it's also not necessary in the hook that's already there? https://github.com/realm/realm-js/blob/6269081d4e58c116ea7800bdd18c59dc0ffd9c80/integration-tests/tests/src/hooks/open-realm-before.ts#L31-L74

papafe:
For me the work on the test is "practically" done. The roundtrip tests are all passing, no issues with those.
The issue is for the "mixed collections" tests, that even though pass when running each on their own, exhibit some strange behaviour when run together, so I need to investigate why we get this flakyness. 
But general comments about the tests/structure is appreciated 😁

integration-tests/tests/src/hooks/open-realm-before.ts
```diff
@@ -83,3 +84,34 @@ export function openRealmBefore(config: OpenRealmConfiguration = {}): void {
...
+    return realmAndConfig.realm;
+  };
+
+  this.closeAllRealms = () => {
+    // this.openedInfo?.forEach(({ realm, config }) => {
+    //   if (!realm?.isClosed) {
+    //     realm.close();
+    //   }
+    //   Realm.deleteFile(config);
+    // });
```
papafe, line 104: There is another thread in this PR that is not visible here because it's on an outdated version of this file, regarding if here using `clearTestState` is enough.

integration-tests/tests/src/typings.d.ts
```diff
@@ -153,6 +157,11 @@ type RealmContext = {
    */
   closeRealm(options?: Partial<CloseRealmOptions>): Promise<void>;
 } & Mocha.Context;
+type MultiRealmContext = {
+  openedInfo: { realm: Realm; config: Realm.Configuration }[];
+  getRealm: (config: any) => Promise<Realm>; //any should be OpenRealmConfiguration
```
papafe, line 162: I think I can't use `OpenRealmConfiguration` here because it would create a circular reference and TS didn't seem to appreciate that

papafe:
@elle-j @kraenhansen @kneth I've removed all the flaky tests and the changes I've done with those. I will add them in a separate PR, but for now I wanted to have at least the basic tests merged

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -16,28 +16,131 @@
...
+  } else if (actual instanceof MixedClass && realm) {
+    const insertedVal = realm.objects<MixedClass>("MixedClass").filtered("_id = $0", actual._id)[0];
+    expect(actual._id.equals(insertedVal._id)).equals(true);
+    defaultTester(actual.value, insertedVal.value);
+  } else {
+    expect(actual).equals(inserted);
+  }
+}
+
+async function setupTest(realm: Realm, useFlexibleSync: boolean) {
```
elle-j, line 137: Should we rename this function to something more specific like `subscribeIfFlexibleSync()` instead? Currently, that's the only responsibility of the function, so it could make it clearer.

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -48,69 +151,38 @@ function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
  * - Deletes the Realm locally
  * - Reopens and downloads the Realm
  * - Performs a test to ensure the downloaded value match the value created locally.
- * @param typeName
- * @param options
+ * @param typeName Name of the mixed type (only used for the test name)
+ * @param value The value to be used for the test, or a function to obtain it
+ * @param testValue The function used to assert equality
```
elle-j, line 156:
```suggestion
 * @param valueTester The function used to assert equality
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -48,69 +151,38 @@ function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
...
+ * @param useFlexibleSync Boolean to indicate the use of flexible sync (otherwise partition based sync will be used)
```
elle-j, line 157:
```suggestion
 * @param useFlexibleSync Whether to use flexible sync (otherwise partition based sync will be used)
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -137,78 +209,65 @@ function describeRoundtrip({
...
   const buffer = new Uint8Array([4, 8, 12, 16]).buffer;
   describeRoundtrip({
     typeName: "data",
     value: buffer,
-    testValue: (value: ArrayBuffer) => {
-      expect(value.byteLength).equals(4);
-      expect([...new Uint8Array(value)]).deep.equals([4, 8, 12, 16]);
-    },
-    flexibleSync,
+    useFlexibleSync: useFlexibleSync,
```
elle-j, line 242: When the property and the value variable share the name, you can omit the explicit set.
```suggestion
    useFlexibleSync,
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -137,78 +209,65 @@ function describeRoundtrip({
...
   });
 
   const objectId = new Realm.BSON.ObjectId("609afc1290a3c1818f04635e");
   describeRoundtrip({
     typeName: "ObjectId",
     value: objectId,
-    testValue: (value: Realm.BSON.ObjectId) => objectId.equals(value),
-    flexibleSync,
+    useFlexibleSync: useFlexibleSync,
```
elle-j, line 256:
```suggestion
    useFlexibleSync,
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -137,78 +209,65 @@ function describeRoundtrip({
...
   });
 
   const uuid = new Realm.BSON.UUID("9476a497-60ef-4439-bc8a-52b8ad0d4875");
   describeRoundtrip({
     typeName: "UUID",
     value: uuid,
-    testValue: (value: Realm.BSON.UUID) => uuid.equals(value),
-    flexibleSync,
+    useFlexibleSync: useFlexibleSync,
```
elle-j, line 263:
```suggestion
    useFlexibleSync,
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -137,78 +209,65 @@ function describeRoundtrip({
...
   });
 
   const decimal128 = Realm.BSON.Decimal128.fromString("1234.5678");
   describeRoundtrip({
     typeName: "Decimal128",
     value: decimal128,
-    testValue: (value: Realm.BSON.Decimal128) => decimal128.bytes.equals(value.bytes),
-    flexibleSync,
+    useFlexibleSync: useFlexibleSync,
```
elle-j, line 270:
```suggestion
    useFlexibleSync,
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -224,12 +283,42 @@ function describeTypes(flexibleSync: boolean) {
       result.value = result;
       return result;
     },
-    testValue: (value: MixedClass) => recursiveObjectId.equals(value._id),
-    flexibleSync,
+    valueTester: (value: MixedClass) => {
+      expect(recursiveObjectId.equals(value._id)).equals(true);
+    },
+    useFlexibleSync: useFlexibleSync,
```
elle-j, line 289:
```suggestion
    useFlexibleSync,
```

elle-j (approved):
Nice to get the roundtrip tests in 🥳 Left some ideas 👍

integration-tests/tests/src/hooks/open-realm-before.ts
```diff
@@ -69,9 +69,10 @@ export function openRealmHook(config: OpenRealmConfiguration = {}) {
 export function closeThisRealm(this: RealmContext & Mocha.Context): void {
   if (this.closeRealm) {
     this.closeRealm({ clearTestState: true, deleteFile: true });
+  } else {
+    // Clearing the test state to ensure the sync session gets completely reset and nothing is cached between tests
+    Realm.clearTestState();
   }
-  // Clearing the test state to ensure the sync session gets completely reset and nothing is cached between tests
-  Realm.clearTestState();
```
kraenhansen, line 74: What's the idea behind this change? Is this not needed if the Realm can be closed?

papafe: `this.closeRealm({ clearTestState: true, deleteFile: true });` is already calling `Realm.clearTestState`, so if we'd just call it twice otherwise

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: Did you address this? The current use of this still feel wrong to be, as the assertion performed depends on the type of the `actual`.

kneth (approved):

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -224,12 +274,42 @@ function describeTypes(flexibleSync: boolean) {
...
+        typeName: "nested dictionary",
+        value: getNestedMixedDict,
+        useFlexibleSync: true,
+      });
+    });
+  }
 }
 
-describe("mixed", () => {
+describe.only("mixed synced", () => {
```
elle-j, line 312:
```suggestion
describe("mixed synced", () => {
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,132 @@
...
 import { buildAppConfig } from "../../utils/build-app-config";
 
-type MixedClass = {
-  _id: Realm.BSON.ObjectId;
-  value: Realm.Mixed;
-  list: Realm.List<Realm.Mixed>;
-};
 type Value = Realm.Mixed | ((realm: Realm) => Realm.Mixed);
-type ValueTester = (actual: Realm.Mixed, inserted: Realm.Mixed) => void | boolean;
+type ValueTester = (actual: Realm.Mixed, inserted: Realm.Mixed, realm?: Realm) => void;
```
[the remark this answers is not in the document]
elle-j: Looks like the 3rd arg `realm` can removed instead, since it's removed from the default tester. Didn't see it used anywhere else.

integration-tests/tests/src/tests/sync/mixed.ts
kraenhansen: 🙈

integration-tests/tests/src/tests/sync/mixed.ts
papafe: I think I've fixed this. I've made the testing a little bit more complicated, but now the default tester actually uses the `inserted` value as the truth value

integration-tests/tests/src/tests/sync/mixed.ts
papafe: Even half a bitcoin would be enough 😁

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
 
 /**
  * The default tester of values.
  * @param actual The value downloaded from the server.
  * @param inserted The value inserted locally before upload.
  */
-function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
-  expect(actual).equals(inserted);
+function defaultTester(actual: unknown, inserted: unknown) {
+  if (inserted instanceof Array) {
```
elle-j, line 155:
```suggestion
  if (Array.isArray(inserted)) {
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
+    expectRealmList(actual);
+    expect(actual.length).equals(inserted.length);
+    for (let index = 0; index < actual.length; index++) {
+      defaultTester(actual[index], inserted[index]);
+    }
+  } else if (inserted != null && typeof inserted === "object" && "d128" in inserted) {
```
elle-j, line 161: We could check for an `Object` constructor to enter this for the objects you create using e.g. `{}`. Was that the intention with `d128`?
```suggestion
  } else if (inserted !== null && typeof inserted === "object" && inserted.constructor === Object) {
```

Nit: If you want to get rid off the type assertion to `Record` further below, an alternative is to pull that condition out to e.g.:
```ts
function isJsObject(value: unknown): value is Record<string, unknown> {
  return value !== null && typeof value === "object" && !Array.isArray(value) && value.constructor === Object;
}
```

Then the else-if could use:
```ts
} else if (isJsObject(inserted)) {
  // ...
  for (const key of insertedKeys) {
    defaultTester(actual[key], inserted[key]);
    // Instead of:
    // defaultTester(actual[key], (inserted as Record<string, unknown>)[key]);
  }
} // ...
```

But since it'd only be used there it's not super necessary imo.

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
+    expectRealmDictionary(actual);
+    const insertedKeys = Object.keys(actual);
```
elle-j, line 163:
```suggestion
    const insertedKeys = Object.keys(inserted);
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
+    const actualKeys = Object.keys(actual);
+    expect(insertedKeys).members(actualKeys);
```
elle-j, line 165:
```suggestion
    expect(actualKeys).members(insertedKeys);
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
+    for (const key of insertedKeys) {
+      defaultTester(actual[key], (inserted as Record<string, unknown>)[key]);
+    }
+  } else if (inserted instanceof ArrayBuffer) {
+    const actualBinaryView = new Uint8Array(actual as ArrayBuffer);
+    const insertedBinaryView = new Uint8Array(inserted as ArrayBuffer);
```
elle-j, line 171:
```suggestion
    const insertedBinaryView = new Uint8Array(inserted);
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
+    expect(actualBinaryView.byteLength).equals(insertedBinaryView.byteLength);
+    for (let index = 0; index < insertedBinaryView.length; index++) {
+      expect(actualBinaryView[index]).equals(insertedBinaryView[index]);
+    }
+  } else if (inserted != null && typeof inserted === "object" && "_id" in inserted) {
```
elle-j, line 176:
```suggestion
  } else if (inserted !== null && typeof inserted === "object" && "_id" in inserted) {
```

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,181 @@
...
+    expect(actual).instanceOf(MixedClass);
+    const actualMixed = actual as MixedClass;
+    const insertedMixed = inserted as MixedClass;
+    defaultTester(actualMixed._id, insertedMixed._id);
```
elle-j, line 180:
```suggestion
    defaultTester((actual as MixedClass)._id, inserted._id);
```

elle-j (approved):
Nice 🥳

integration-tests/tests/src/tests/sync/mixed.ts
papafe: Yes, I'm doing it because I want to differentiate between a "dictionary" mixed value and an "object" mixed value, and I didn't think of anything more clever than this 😁 
I'll add a comment.

integration-tests/tests/src/tests/sync/mixed.ts
papafe: Ouch

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -15,29 +15,179 @@
...
+}
 
 /**
  * The default tester of values.
  * @param actual The value downloaded from the server.
  * @param inserted The value inserted locally before upload.
  */
-function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
-  expect(actual).equals(inserted);
+function defaultTester(actual: unknown, inserted: unknown) {
```
kraenhansen, line 154: Since the code infers the expected type from the shape of the inserted POJO, it establishes implicit expectations from the calling code in the behavior of the function. This is especially problematic because `inserted` is `unknown` and therefore accepts any values in combination with the final `else` block, which just compares the string representation. In my opinion it's too easy for the calling code to receive a changed that will render this default tester function useless and as such I fear this will be a source of subtle bugs in the testing code that are hard to discover. Additionally, since this puts restrictions on the shape of the `inserted` values, it's potentially harder for us to test combinations of these and types we add in the future.

Do you see any way of passing the tester function explicitly from the tests?
Although potentially more verbose.

papafe: I understand your concerns. 
I suppose it depends what you mean by passing the tester function explicitly from the tests. The methods that are used to define the tests accept a tester function, and this is the one that is used when no function is passed. 
If you mean that we should pass a more specific tester function for each test then... We can do it, but it's just much more complicated, because this method not only considers all the possible mixed types (I think at least), but also collections, so it does a recursive comparison if needed.

kraenhansen: > pass a more specific tester function for each test

Exactly.

> does a recursive comparison if needed

Just an idea on the top of my head: We could implement a higher-order function to create dictionary tester functions with a specific `tester` applied on items.

```typescript
const createDictionaryTester = (itemTester: ValueTester) => (actual: unknown, inserted: unknown) {
  expectRealmDictionary(actual);
  if (typeof inserted !== "object" || inserted === null) {
    throw new Error("Expected an object of values");
  }
  for (const [insertedKey, insertedValue] of Object.entries(inserted)) {
    itemTester(actual[insertedKey], insertedValue);
  }
}
```

To test a nested dictionary `{ foo: { bar: ["baz"] } }` this be chained like:

```typescript
tester: createDictionaryTester(createDictionaryTester(stringTester))
```

papafe: I think there are a couple of things to consider:
- The values that are being used the tests look something like `[1, true, "string", ...., object, innerList, innerDicti]`. S
Thus the testing methods wouldn't look particularly pretty.
- It gets even more messy in the other PR with additional tests. In that PR I don't have a single mixed value that I'm trying to roundtrip like here, but I do a lot of modifications to the list/dictionary in order to verify it all works correctly, so I can't have one specific testing function. 
Overall I tried to make a testing function that would be generic enough to work with different kind of tests, even though it is not maybe the most specific or strict. I think that if we go the more strict route we either need to simplify the test cases (mostly considering the additional PR) or we need to greatly complicate our tester.

kraenhansen: 🤔 Yeah. I agree, there are tradeoffs here, which complicates the decision.
I wanted to mention my reservations, but I won't object to this merging as is.

kraenhansen (approved):

integration-tests/tests/src/tests/sync/mixed.ts
```diff
@@ -48,69 +200,48 @@ function defaultTester(actual: Realm.Mixed, inserted: Realm.Mixed) {
...
-        this.value = typeof value === "function" ? value(this.realm) : value;
-        this.realm.create<MixedClass>("MixedClass", {
+        if (typeof value === "function") {
+          const valueResult = value(this.realm);
+          if ("expected" in valueResult && "values" in valueResult) {
+            this.value = valueResult.values;
+            this.expected = valueResult.expected;
+          } else {
+            this.value = valueResult;
+          }
```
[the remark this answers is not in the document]
papafe: I agree it's confusing. I have added more comments (and types) to make it clearer.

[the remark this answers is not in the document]
papafe: Corrected types on this

PR #6618, opened 2024-04-15
Suggestions to the "BSON v6 upgrade"

kraenhansen:
## What, How & Why?

- Uses the rollup alias plugin to provide an exact path for the "bson" package, to avoiding a top-level await when bundling "bson" into the "realm-web" IIFE bundle.
- Updates the `realm-web-integration-tests` to consume the "realm-web" ESM bundle.

### DIFF
--- a/packages/realm-web/rollup.config.mjs
+++ b/packages/realm-web/rollup.config.mjs
@@ -25,7 +25,10 @@ import dts from "rollup-plugin-dts";
 import pkg from "./package.json" assert { type: "json" };
 
 const replacer = replace({
+  preventAssignment: true,
+  values: {
   __SDK_VERSION__: JSON.stringify(pkg.version),
+  },
 });
 
 export default [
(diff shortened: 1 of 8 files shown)


packages/realm-web/rollup.config.mjs
```diff
@@ -97,16 +104,28 @@ export default [
...
-      typescript({
-        tsconfig: "src/dom/tsconfig.json",
+      // Providing an alias for the "bson" package
+      // See https://github.com/mongodb/js-bson/pull/669#pullrequestreview-1994786536 for more context
+      alias({
+        entries: [
+          {
+            find: "bson",
+            // We're assuming the location of a .cjs next to the .mjs bundle
+            replacement: require.resolve("bson").replace(".mjs", ".mjs"),
```
kraenhansen, line 114: I tried but failed to use `import.meta.resolve("bson/lib/bson.cjs")` but this fails since the library declare no sub-path exports:
> [!] Error: Package subpath './lib/bson.cjs' is not defined by "exports" in [...]/realm-js/node_modules/bson/package.json imported from [...]/realm-js/packages/realm-web/rollup.config.mjs

PR #6619, opened 2024-04-16
Use CMake 3.29.2

kneth:
## What, How & Why?

Currently CMake on GHA Windows x64 is broken.

Reference: https://github.com/actions/runner-images/issues/9680

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -77,9 +77,9 @@ jobs:
           - { os: ios, runner: macos-latest-xlarge, arch: ios, artifact-path: packages/realm/react-native/ios/realm-js-ios.xcframework }
     steps:
       - name: Setup CMake
-        uses: jwlawson/actions-setup-cmake@v2
+        uses: lukka/get-cmake@latest
         with:
-          cmake-version: '3.29.2'
+          cmakeVersion: "3.29.2"
 
       - name: Checkout code
         uses: actions/checkout@v4
@@ -282,6 +282,11 @@ jobs:
           #- { os: ios, target: "test:ci:catalyst", runner: macos-latest, environment: react-native-test-app, arch: "catalyst" }
     timeout-minutes: 60
     steps:
+      - name: Setup CMake
+        uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: "3.29.2"
+
       - name: Checkout code
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -108,6 +108,17 @@ jobs:
...
+      - name: Setup CMake
+        if: ${{ runner.os == 'Windows' }}
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.29.2'
+
+      - name Check CMake
+        shell: bash
+        run: |
+          dir $(which cmake) >> $GITHUB_PATH
```
kraenhansen, line 120: This seems redundant? If `which` can find `cmake` why would it need to be added to the `PATH`?

From https://github.com/jwlawson/actions-setup-cmake/blob/master/README.md
> This action will update the path for your workflow to include cmake matching the platform and version requirements.

.github/workflows/pr-realm-js.yml
```diff
@@ -108,6 +108,17 @@ jobs:
         if: ${{ runner.os == 'Windows' }}
         uses: ilammy/msvc-dev-cmd@v1
 
+      - name: Setup CMake
+        if: ${{ runner.os == 'Windows' }}
```
kraenhansen, line 112: Might make sense to do this on all platforms to be resilient to regressions on other platforms?
```suggestion
```

kraenhansen (approved):

.github/workflows/pr-realm-js.yml
kneth: True, but `cmake-js` fails when spawning `cmake` with `EINVAL`

.github/workflows/pr-realm-js.yml
kneth: If it ain't broken, don't fix it :smile:

.github/workflows/pr-realm-js.yml
```diff
@@ -108,6 +108,11 @@ jobs:
         if: ${{ runner.os == 'Windows' }}
         uses: ilammy/msvc-dev-cmd@v1
 
+      - name: Setup CMake
+        uses: jwlawson/actions-setup-cmake@v2
```
kraenhansen, line 112: For what it's worth [this comment](https://github.com/actions/runner-images/issues/9680#issuecomment-2052018532) suggests `lukka/get-cmake` is faster 🤔

.github/workflows/pr-realm-js.yml
```diff
@@ -84,7 +84,7 @@ jobs:
       - name: Setup node version
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 20.11.1
```
kraenhansen, line 87: This change can be reverted once https://github.com/prebuild/prebuild/pull/325 gets merged, release and we adopt the change.

PR #6633, opened 2024-04-21
RJS-2717: Replaced lodash with lodash.isequal

### LINKED ISSUE #6458: [@realm/react] Bundle in `lodash.isequal`
Since we're only using `isEqual` from `lodash` we should avoid requiring in the entire library as it'll be difficult for end-users to shake off. Instead we could depend on the more limited [`lodash.isequal`](https://www.npmjs.com/package/lodash.isequal) package and optionally avoid excluding it to pull it into our bundle.

Ydixit007:
## What, How & Why?

Since we're only using `isEqual` from `lodash` we should avoid requiring in the entire library as it'll be difficult for end-users to shake off. Instead we could depend on the more limited `lodash.isequal` package.

This closes #6458 .

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5737,10 +5737,10 @@
       "integrity": "sha512-OvlIYQK9tNneDlS0VN54LLd5uiPCBOp7gS5Z0f1mjoJYBrtStzgmJBxONW3U6OZqdtNzZPmn9BS/7WI7BFFcFQ==",
       "dev": true
     },
-    "node_modules/@types/lodash-es": {
-      "version": "4.17.12",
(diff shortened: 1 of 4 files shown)


Ydixit007:
@cla-bot check

kraenhansen:
```
npm ERR! Missing: @types/lodash.isequal@4.5.8 from lock file
npm ERR! Missing: lodash.isequal@4.5.0 from lock file
```

You'll need to run `npm install` and commit the changes to the lock file 👍

kraenhansen (approved):

elle-j (approved):
Thanks @Ydixit007 !

PR #6637, opened 2024-04-25
Upgrade to HEAD of Realm Core's master

kneth:
## What, How & Why?

Next core release will come with a number of breaking changes. This PR is a preparation for the release, and it requires https://github.com/realm/realm-core/pull/7623 to be merged and release.

The Binding Generator has some limitations, and in conjunction with https://github.com/realm/realm-core/pull/7623 this PR works around them.

In https://github.com/realm/realm-core/pull/7300, a new user is introduced (`app::User`) which inherit from `SyncUser`. As both classes are used with `std::shared_ptr` (annotated `SharedPtrWrapper` in `spec.xml`), it is not possible to generate code. The ~solution~ hack is to downcast all objects. In most cases the downcasting has no effect but for the user classes. For user classes we will use only one - `app::User` - and downcast `SyncUser` to `app::User`.

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


packages/realm/src/Realm.ts
```diff
@@ -381,7 +381,8 @@ export class Realm {
         return Realm.normalizePath(config.path);
       } else {
         const bindingSyncConfig = toBindingSyncConfig(config.sync);
-        return config.sync.user.internal.syncManager.pathForRealm(bindingSyncConfig, config.path);
+        const path = config.sync.user.internal.pathForRealm(bindingSyncConfig, undefined);
+        return path;
```
kraenhansen, line 385: Seems fine to keep as a simple return without a local `path` variable.

```suggestion
        return config.sync.user.internal.pathForRealm(bindingSyncConfig, undefined);
```

packages/realm/src/app-services/App.ts
```diff
@@ -278,25 +279,31 @@ export class App<
...
+    this.internal = binding.App.getApp(binding.AppCacheMode.Disabled, {
+      appId: id,
+      deviceInfo: App.deviceInfo,
+      transport: createNetworkTransport(fetch),
+      baseUrl,
+      defaultRequestTimeoutMs: timeout ? binding.Int64.numToInt(timeout) : undefined,
+      baseFilePath: baseFilePath ? baseFilePath : fs.getDefaultDirectoryPath(),
+      metadataMode: metadata ? toBindingMetadataMode(metadata.mode) : binding.MetadataMode.NoEncryption,
+      customEncryptionKey: metadata?.encryptionKey,
+      syncClientConfig,
```
kraenhansen, line 298: I'd probably inline this, if there's no need for the extra local const (which also skips the need to declare the type).

```suggestion
      syncClientConfig: {
        multiplexSessions,
        userAgentBindingInfo: App.userAgent,
      }
```

packages/realm/src/app-services/User.ts
```diff
@@ -152,7 +152,7 @@ export class User<
    * @returns The user ID as a string.
    */
   get id(): string {
-    return this.internal.identity;
+    return this.internal.userId;
```
kraenhansen, line 155: I've added https://github.com/realm/realm-core/pull/7300/files#r1579400825

kraenhansen (approved):
Looks like a lot of good simplifications. If it passes tests I'm fine with merging it as part of an upgrade to Core 👍

>  In most cases the downcasting has no effect but for the user classes

🤞 We didn't test the performance of this, right? I mean - its a pretty critical code path.

packages/realm/bindgen/src/templates/jsi.ts
```diff
@@ -436,7 +436,7 @@ function convertToJsi(addon: JsiAddon, type: Type, expr: string): string {
       const inner = type.args[0];
       switch (type.name) {
         case "std::shared_ptr":
-          if (inner.kind == "Class" && inner.sharedPtrWrapped) return `JS_FROM_SHARED_${inner.name}(_env, ${expr})`;
+          if (inner.kind == "Class" && inner.sharedPtrWrapped) return `JS_FROM_SHARED_${inner.name}(_env, std::dynamic_pointer_cast<${inner.cppName}>(${expr}))`;
```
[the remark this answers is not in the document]
kneth: After consulting with @jedelbo I have change it

kneth:
> We didn't test the performance of this, right? I mean - its a pretty critical code path.

I have changed it so we don't downcast but I hope that a C++ compiler will generate very little code for 

```cpp
Foo *foo = new Foo();
auto bar = std::dynamic_pointer_cast<Foo>(foo);
```

packages/realm/src/Realm.ts
kneth: Me debugging at some point

packages/realm/src/app-services/User.ts
kneth: Good question but will not have any impact for users of the SDK if we decide to change `user_id` to `id`.

packages/realm/src/app-services/App.ts
```diff
@@ -50,8 +50,9 @@ export enum MetadataMode {
   Encryption = "encryption",
   /**
    * Do not persist {@link User} objects.
+   * @deprecated will be renamed to `InMemory`
    */
-  NoMetadata = "noMetadata",
+  NoMetadata = "inMemory",
```
elle-j, line 55: This will be an SDK breaking change if we don't allow users to keep using the old value until we bump our major. One option is to keep the old constant value (`noMetadata`) publicly, but send the new value to Core (`inMemory`).

packages/realm/src/app-services/App.ts
```diff
@@ -82,7 +83,7 @@ function toBindingMetadataMode(arg: MetadataMode): binding.MetadataMode {
 const translationTable: Record<binding.MetadataMode, MetadataMode> = {
   [binding.MetadataMode.NoEncryption]: MetadataMode.NoEncryption,
   [binding.MetadataMode.Encryption]: MetadataMode.Encryption,
-  [binding.MetadataMode.NoMetadata]: MetadataMode.NoMetadata,
+  [binding.MetadataMode.InMemory]: MetadataMode.NoMetadata, // TODO: should we rename NoMetadata to InMemory?
```
elle-j, line 86: I don't think we should rename unless this will be part of a new next major SDK version.

packages/realm/src/app-services/App.ts
```diff
@@ -278,25 +279,34 @@ export class App<
       assert.function(fetch, "fetch");
     }
 
+    const syncClientConfig: binding.SyncClientConfig_Relaxed = {
+      multiplexSessions,
+      userAgentBindingInfo: App.userAgent,
+    };
+
```
elle-j, line 286: Unused
```suggestion
```

packages/realm/src/app-services/App.ts
```diff
@@ -278,25 +279,34 @@ export class App<
...
+      defaultRequestTimeoutMs: timeout ? binding.Int64.numToInt(timeout) : undefined,
+      baseFilePath: baseFilePath ? baseFilePath : fs.getDefaultDirectoryPath(),
+      metadataMode: metadata ? toBindingMetadataMode(metadata.mode) : binding.MetadataMode.NoEncryption,
+      customEncryptionKey: metadata?.encryptionKey,
+      syncClientConfig: {
         multiplexSessions,
-      },
-    );
+        userAgentBindingInfo: App.userAgent,
+      }
```
elle-j, line 301: Linting 👍 
```suggestion
      },
```

packages/realm/src/app-services/App.ts
```diff
@@ -278,25 +279,34 @@ export class App<
...
+    });
+  }
+
+  /**
+   * @internal
+   */
+  public get syncManager(): binding.SyncManager {
+    return this.internal.syncManager;
   }
```
elle-j, line 310: Don't see any references to this, when would this get used? In `Sync.ts` I see it used directly from the internal object, but not the SDK `App` object.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -305,7 +305,7 @@ export class SyncSession {
    */
   get config(): SyncConfiguration {
     return this.withInternal((internal) => {
-      const user = new User(internal.user, mockApp);
+      const user = new User(internal.config.user, mockApp);
```
elle-j, line 308: `User` currently takes a `binding.User` as 1st arg, but `internal.config.user` returns a `binding.SyncUser`, so this won't transpile. Hopefully this is an easy fix 🤞

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -349,8 +349,9 @@ export class SyncSession {
    * Gets the User that this session was created with.
    */
   get user() {
-    return User.get(this.withInternal((internal) => internal.user));
+    return User.get(this.withInternal((internal) => internal.config.user));
```
elle-j, line 352: Similar issue here as previous comment.

elle-j (commented):
Thanks for tackling this! I've left some comments that would be good to resolve before merge 👍

packages/realm/bindgen/src/templates/jsi.ts
```diff
@@ -781,11 +781,6 @@ class JsiCppDecls extends CppDecls {
     }
 
     for (const cls of spec.classes) {
-      assert(
-        !cls.sharedPtrWrapped || (!cls.base && cls.subclasses.length == 0),
-        `We don't support mixing sharedPtrWrapped and class hierarchies. ${cls.name} requires this.`,
-      );
```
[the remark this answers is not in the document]
kneth: It will still require some additions to `realm_helpers.hpp` in core.

packages/realm/bindgen/src/templates/node.ts
```diff
@@ -709,18 +709,14 @@ class NodeCppDecls extends CppDecls {
     }
 
     for (const cls of spec.classes) {
-      assert(
-        !cls.sharedPtrWrapped || (!cls.base && cls.subclasses.length == 0),
-        `We don't support mixing sharedPtrWrapped and class hierarchies. ${cls.name} requires this.`,
-      );
```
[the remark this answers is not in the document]
kneth: Same as previous comment :-)

packages/realm/src/app-services/App.ts
kneth: I will make it non-breaking.

packages/realm/src/app-services/App.ts
kneth: Yeah, we don't need to expose it now

PR #6648, opened 2024-05-02
Additional tests for collections in mixed

papafe:
The new tests are all passing if run one their own, but there is a crash when running multiple tests. This is reported in https://github.com/realm/realm-core/issues/7657

This PR should be picked up only when the core issue has been solved.


papafe:
https://github.com/realm/realm-core/issues/7657 has been fixed, and I verified that it fixes the crashes. It has been merged into master, so we should rebase this branch onto master when possible. 

There's an additional issue though unfortunately, that was also present before but I think it was covered by the crashes. 
When running the tests on their own, they all pass but when running them together:
- The second test usually encounters a timeout. I have verified that this happens because we don't get correct notifications. In all the test methods we create a promise with `getWaiter` that gets resolved when we get a notification for the property called `value`. The timeout happens because we get two empty notifcations from core (the first one is expected);
- Resolving the promise on the second notification (even if the notification is empty) results in `Accessing object which has been invalidated or deleted`;
- Trying to use the `sleep` method, instead of the promise results in `Accessing object which has been invalidated or deleted`;

These timeouts are also flaky as they don't happen all the time. This makes the investigation quite annoying though.

PR #6649, opened 2024-05-02
Adding tests accessing `app.currentUser` from `App` and `User` listeners

kraenhansen:
## What, How & Why?

This exercise https://github.com/realm/realm-core/issues/7670 and https://github.com/realm/realm-core/issues/7183 to ensure future regressions will be easier to debug.

We should hold back on merging this, until https://github.com/realm/realm-core/issues/7670 is fixed and we've upgraded Core to include the fix.

### DIFF
--- a/integration-tests/tests/src/tests/sync/app.ts
+++ b/integration-tests/tests/src/tests/sync/app.ts
@@ -245,7 +245,7 @@ describe("App", () => {
 
       const handle = createPromiseHandle();
       const listener = () => {
-        expect(this.app.currentUser).to.not.be.null;
+        expect(this.app.currentUser).instanceOf(Realm.User);
         this.app.removeListener(listener);
         handle.resolve();
       };
@@ -269,6 +269,7 @@ describe("App", () => {
 
       const handle = createPromiseHandle();
       const listener = () => {
+        expect(this.app.currentUser).instanceOf(Realm.User);
         expect(this.app.currentUser?.id).equals(user.id);
         user.removeListener(listener);
         handle.resolve();
--- a/integration-tests/tests/src/tests/sync/app.ts
+++ b/integration-tests/tests/src/tests/sync/app.ts
@@ -236,7 +236,9 @@ describe("App", () => {
       expect(this.app.currentUser).to.be.null;
     });
 
-    it("currentUser is available from an App listener", async function (this: Mocha.Context & AppContext & RealmContext) {
+    it("currentUser is available from an App listener", async function (this: Mocha.Context &
+      AppContext &
+      RealmContext) {
       expect(this.app.currentUser).to.be.null;
 
       const credentials = Realm.Credentials.anonymous();
@@ -257,7 +259,9 @@ describe("App", () => {
       await handle;
     });
 
-    it.only("currentUser is available from a User listener", async function (this: Mocha.Context & AppContext & RealmContext) {
+    it.only("currentUser is available from a User listener", async function (this: Mocha.Context &
+      AppContext &
+      RealmContext) {
       expect(this.app.currentUser).to.be.null;
 
       const credentials = Realm.Credentials.anonymous(false);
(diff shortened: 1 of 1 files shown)


integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -235,6 +236,50 @@ describe("App", () => {
       expect(this.app.currentUser).to.be.null;
     });
 
+    it("currentUser is available from an App listener", async function (this: Mocha.Context & AppContext & RealmContext) {
```
elle-j, line 239: Lint warning
```suggestion
    it("currentUser is available from an App listener", async function (this: Mocha.Context &
      AppContext &
      RealmContext) {
```

integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -235,6 +236,50 @@ describe("App", () => {
...
+      await this.app.currentUser?.logOut();
+      expect(this.app.currentUser).to.be.null;
+
+      await handle;
+    });
+
+    it.only("currentUser is available from a User listener", async function (this: Mocha.Context & AppContext & RealmContext) {
+      expect(this.app.currentUser).to.be.null;
+
+      const credentials = Realm.Credentials.anonymous(false);
```
elle-j, line 263: Will the `false` actually make a difference here? Or can it not be guaranteed that this is the only logged in user when this test is run? (The previous test used the default argument.)

integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -235,6 +236,50 @@ describe("App", () => {
       expect(this.app.currentUser).to.be.null;
     });
 
+    it("currentUser is available from an App listener", async function (this: Mocha.Context & AppContext & RealmContext) {
+      expect(this.app.currentUser).to.be.null;
+
+      const credentials = Realm.Credentials.anonymous();
+
+      const handle = createPromiseHandle();
+      const listener = () => {
```
kneth, line 245: Will this function be called once or twice?

kraenhansen: I expect it to be called only once. It doesn't behave as the collection listeners (where it fires initially).

integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -235,6 +236,50 @@ describe("App", () => {
...
+      this.app.addListener(listener);
+      await this.app.logIn(credentials);
+
+      await this.app.currentUser?.logOut();
+      expect(this.app.currentUser).to.be.null;
+
+      await handle;
+    });
+
+    it.only("currentUser is available from a User listener", async function (this: Mocha.Context & AppContext & RealmContext) {
```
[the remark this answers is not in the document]
kraenhansen: You saw nothing ... 🙈

integration-tests/tests/src/tests/sync/app.ts
kraenhansen: I agree. It'll most likely not make a difference.

integration-tests/tests/src/tests/sync/app.ts
```diff
@@ -235,6 +236,54 @@ describe("App", () => {
...
+    it("currentUser is available from an App listener", async function (this: Mocha.Context &
+      AppContext &
+      RealmContext) {
+      expect(this.app.currentUser).to.be.null;
+
+      const credentials = Realm.Credentials.anonymous();
+
+      const handle = createPromiseHandle();
+      const listener = () => {
+        expect(this.app.currentUser).to.not.be.null;
```
elle-j, line 248: I'm thinking it may be safer to check a property on the object to be sure we're not getting e.g. an empty object back.
```suggestion
        expect(this.app.currentUser?.id).to.be.a("string");
```

elle-j (approved):
LGTM 👍

integration-tests/tests/src/tests/sync/app.ts
kraenhansen: Then checking it's an instance of `User` is probably even better 🙂

elle-j: 💪

kneth (approved):

Issue #6530, opened 2024-03-04
Upgrade `actions/upload-artifact` to v4 and refactoring `pr-realm-js` workflow


kraenhansen:
We're currently using `actions/upload-artifact` v3, which uses Node v16 and we're getting warnings on CI for this.
As per [the release notes for `actions/upload-artifact@v4`](https://github.com/actions/upload-artifact/releases/tag/v4.0.0) it's suggested to follow this guide to perform the migration: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md

In its current form, no tests are initiated before prebuilds are uploaded for all platforms, which slows down the feedback loop when pushing changes to a PR branch. Since the upload-artifacts@v4 remove a feature we're depending on for this flow (implicit merging of artefacts that share name), we should take this opportunity to refactor our workflow into jobs where a test run only depends on the upload of the prebuilt artefact that it actually needs to complete.

In addition to this, we should consider optimising which prebuilds are actually produced and which tests we run:
- By default, I'd suggest we skip integration tests on all platforms except Node.js on Linux for PRs that touch only the platform independent code of our SDK.
- In addition to this default behaviour, we could signal a desire to run tests for additional platforms by adding specific platform labels to the PR.
- On `release/...` branches (and possibly merge commits to `main`) we build all artefacts and run tests on all platforms.

kraenhansen:
https://github.com/realm/realm-js/issues/6446 seems related to this.

kraenhansen:
If we choose to refactor the workflow such that test jobs depend on yhe upload of the prebuilt artifact that it actually needs, we could probably revert https://github.com/realm/realm-js/pull/5729 as well.

PR #5053, opened 2022-11-01
Compile with C++20

RedBeard0531:
## What, How & Why?

Lets see if this works...

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 📝 Update `COMPATIBILITY.md`
* [ ] 🚦 Tests
* [ ] 🔀 Executed flexible sync tests locally if modifying flexible sync
* [ ] 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)
* [ ] 📱 Check the React Native/other sample apps work if necessary
* [ ] 📝 Public documentation PR created or is not necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [ ] typescript definitions file is updated
* [ ] jsdoc files updated

Issue #3708, opened 2021-05-03
Simplify by unifying build commands


kraenhansen:
## Goals
A more approachable developer experience.

## Expected Results
The existing build-android.js and build-ios.sh scripts could be build around a shared CLI, to simplify consumption.

A bare "build" NPM script could provide a discoverable method of building for all supported runtimes.

```
npm run build -- --runtime ios --type Debug --arch x86
```

As an example, the following NPM scripts could expose discoverable shortcuts to build for the various supported runtimes:

```
 npm run build:node
 npm run build:ios
 npm run build:android
 npm run build:node:debug
 npm run build:ios:debug
 npm run build:android:debug
```

Bonus: It would be great to be able to set an environment variable or npm config to skip building Realm JS all together.
This would help simplify the "react native" integration test jobs on GitHub Actions.

## Actual Results
These are the commands needed to build the native module to run the integration tests:

### Building the native module for Node.js in debug mode:
```
npx cmake-js -d
```

### Building the native module (xcframework) for the iOS simulator in debug mode:
```
./scripts/build-ios.sh -c Debug -s
```

### Building the native module (.so) for the Android simulator in debug mode:
```
node scripts/build-android.js --arch=x86 --build-type=Debug
```

## Version of Realm and Tooling
* Realm JS SDK Version: `develop` branch
* Node or React Native: N/A
* Client OS & Version: N/A
* Which debugger for React Native: N/A

kneth:
Should it be possible to specify if you want to build with or without sync (option `--sync` maybe)?

PR #6650, opened 2024-05-03
RJS-2101: "Build optimizations" for iOS and Android

kraenhansen:
## What, How & Why?

> [!IMPORTANT]  
> With this change, our SDK is no longer pinned to a specific ABI of JSI and we expect the next release of our SDK (after merging this PR) to be backwards and forwards compatible with many past versions of React Native.

This is my friendly takeover of #6396, by reverting the compilation of Realm Core from source and instead distributing prebuilds via our NPM package:
- For React Native Apple (`npm run prebuild-apple --workspace realm`):
  - Prebuild format is an XCFramework of Realm Core
  - Deferred compilation on the developers machine is our binding (between Realm Core and React Native Core & JSI) driven by CocoaPods & XCode.
- For React Native Android (`npm run prebuild-android --workspace realm`):
  - Prebuilds format is a CPack install directory (containing CMake config files)
  - Deferred compilation on the developers machine is our binding (between Realm Core and React Native Core & JSI) driven by Gradle calling into CMake.
- For Node.js we're able to rely on the ABI stability of N-API and we produce prebuilds containing both Realm Core and our binding (between Realm Core and N-API) (`npm run prebuild-node --workspace realm`).

This also moves all binding related code (prebuilt or not), including that previously in `packages/realm/react-native` into `packages/realm/binding` in platform specific directories.

We won't support compiling our prebuilds from from the NPM archive as users still need to `git clone` our mono-repository to produce a build from source.

It's worth mentioning that this PR introduce an internal CLI incapsulating the complexity of invoking the CMake scripts for Android and Apple platforms. I imagine we could extend this to wrap cmake-js to provide a single cohesive experience for building our prebuilds in the future: This is a replacement for the `packages/realm/scripts` for [Android](https://github.com/realm/realm-js/blob/fc633f70b69596d84f7dce5a2d7f1fac9f8e25e1/packages/realm/scripts/build-android.ts) and [iOS](https://github.com/realm/realm-js/blob/main/packages/realm/scripts/build-ios.sh) that we currently use.

This is the resulting size of our tarball:

```
npm notice === Tarball Details === 
npm notice name:          realm                                   
npm notice version:       12.9.0                                  
npm notice filename:      realm-12.9.0.tgz                        
npm notice package size:  212.5 MB                                
npm notice unpacked size: 1.1 GB                                  
npm notice shasum:        f2bc7cffc60ee97a6c00161fefe3bac05b7becae
npm notice integrity:     sha512-V8wGrC2KNZe3O[...]Q5r/SEfraxqdg==
npm notice total files:   2895
```

---

Here's a bit more context on this change and why we choose the approach we did:

As we started implementing out "React Native SDK Build Optimizations" project we realized a few things:
- Although we are generating 11K lines of C++ from the binding generator, it’s very fast to compile - most likely because it’s a single translation unit.
- This generated binding layer is the only code dependent on the React Native headers / ABI.
- We need Cmake to generate the Xcode project to compile Realm Core, which might not be installed on an app developers machine (in the correct version) but we can compile the binding layer without it.
- If we choose to leave out the binding layer from the prebuilt binary (and we can because of :point_up:) we don’t need to produce separate prebuilds for each React Native version.
- Because of :point_up: we actually don’t have to delay the download of the prebuilt until “pod install” nor build-time. We could actually include it as an xcframework into the archive that we upload to NPM, like we do now. The main drawback of this approach is the download time for users as they “npm install” our package, but that won’t be a regression in DX to what we currently have and it doesn’t seem to be a deal-breaker to anyone as-is.
- For Android, the situation is slightly more complicated, since Realm Core is dependent on ABI of the C++ Standard Library brought in by the NDK. Since the NDK version is configurable by the developer, this becomes a bit more complicated. Luckily the React Native template app (and Expo) declares a default for this NDK version, which we could rely on for the prebuild we upload to NPM, as long as we [take great care to error if developers try using this prebuild with an unexpected NDK version](https://github.com/realm/realm-js/issues/6683) and provide an easy way to build Realm Core from source, for those developers choosing a different NDK than the default.

In the context of the above, if we choose to include prebuild binaries for iOS and Android of Realm Core in the NPM package, which will be ABI independent on the React Native version (except for the caveat of NDK version, which has a default that I'd expect 95+% to leave unchanged) ... do we actually need to support compiling from source when installing the realm package via NPM as the original project scope suggests? I'd expect the need for that to be very rare (as it is right now) and it does come with drawbacks:
- We'll have to avoid hoisting shared dev-dependencies to the root package of our mono-repo
- Depending on a decision of forcing developers that wants to use this to have to npm install in their node_modules/realm we might need to include these dev-dependencies as actual dependencies of realm. Which would add to the install time of our package.
- We'll most likely have to run tests to ensure all relevant dependencies are actually available to build from source when the realm package is installed outside of the mono-repo.
- We'll have to include all of the Realm Core source-code into the repository, not a huge deal - but it all adds up.

The alternative is to ask developers that want to consume Realm JS and Realm Core from source to perform a checkout of the repository, run the relevant build script and npm pack the SDK package themselves.

Another (now closed) PR has some more context on the problem this PR is solving: https://github.com/realm/realm-js/pull/4027

---

Notes for the reviewer:
- We deleted the `packages/realm/binding/CMakeLists.txt` and instead either refer to the `CMakeLists.txt` in `bindgen/vendor/realm-core` directly or `CMakeLists.txt` files for the individual platforms, instead of having an shared CMake project used by all bindings. This, because it felt the requirements for each platform was too different across platforms to play for the complexity of having a shared project.

## ☑️ ToDos
* [x] Check the package.json `files` to ensure all files needed to compile is included.
* [x] 📝 Changelog entry
* [ ] Include before / after size of the NPM package.
* [ ] 🚦 Tests


integration-tests/environments/react-native-test-app/package.json
```diff
@@ -97,6 +101,16 @@
...
+      "clean": "if-file-deleted",
+      "files": [
+        "ios/Podfile",
+        "../../../packages/realm/RealmJS.podspec"
+      ],
+      "output": [
+        "ios/Pods",
+        "ios/Podfile.lock",
+        "ios/RealmTests.xcworkspace"
+      ],
```
kraenhansen, line 113: No need for dependencies on other scripts when testing on CI, since that's already handled by other jobs uploading artifacts.

contrib/debugging-react-native.md
```diff
@@ -1,111 +0,0 @@
-
```
kraenhansen, line 1: Deleting this entirely, since almost all the information was outdated and would have to be re-tested anyway.

packages/realm/RealmJS.podspec
```diff
@@ -35,36 +34,48 @@ Pod::Spec.new do |s|
...
+  s.public_header_files    = 'binding/apple/*.h'
 
   s.frameworks             = uses_frameworks ? ['React'] : []
 
   s.library                = 'c++', 'z', 'compression'
 
   s.pod_target_xcconfig    = {
                                 # Setting up clang
-                                'CLANG_CXX_LANGUAGE_STANDARD' => 'c++17',
+                                'CLANG_CXX_LANGUAGE_STANDARD' => 'c++20',
```
kraenhansen, line 54: We're using C++20 features in the binding.

packages/realm/bindgen/src/templates/jsi.ts
```diff
@@ -912,6 +912,12 @@ class JsiCppDecls extends CppDecls {
             )
             .join("\n")}
 
+
+          // Silence these warnings as they might be confusing to developers as they build the binding
+
+          #pragma GCC diagnostic push
+          #pragma GCC diagnostic ignored "-Wswitch"
```
kraenhansen, line 919: Adding this to silence warnings that might be confusing to our users.

packages/realm/binding/apple/RealmReact.mm
```diff
@@ -144,10 +144,10 @@ - (void)setBridge:(RCTBridge *)bridge {
             // because the work is performed before the microtask queue is flushed - see sequence
             // diagram at https://bit.ly/3kexhHm. It might be possible to further optimize this,
             // e.g. only flush the queue a maximum of once per frame, but this seems reasonable.
-            if (!waitingForUiFlush) {
-              waitingForUiFlush = true;
+            if (!self->waitingForUiFlush) {
+              self->waitingForUiFlush = true;
               [bridge jsCallInvoker]->invokeAsync(
-                  [&]() { waitingForUiFlush = false; });
+                  [&]() { self->waitingForUiFlush = false; });
```
kraenhansen, line 150: Adding this to silence warnings that might be confusing to our users.

packages/realm/scripts/submit-analytics.js
```diff
@@ -0,0 +1,27 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+/* eslint-env node */
+/* eslint-disable -- console */
+
+try {
+  const { submitAnalytics } = require("../dist/scripts/submit-analytics");
+  submitAnalytics().catch(console.error);
+} catch (err) {
+  console.error(err instanceof Error ? err.message : err);
```
kraenhansen, line 26: This only prints if loading the analytics file fails, since `submitAnalytics` returns a rejected promise if failing.

packages/realm/src/scripts/build/android.ts
```diff
@@ -0,0 +1,172 @@
+////////////////////////////////////////////////////////////////////////////
```
kraenhansen, line 1: This is the build script for the Android prebuild.

packages/realm/src/scripts/build/apple.ts
```diff
@@ -0,0 +1,181 @@
+////////////////////////////////////////////////////////////////////////////
```
kraenhansen, line 1: This is the build script for the Apple / iOS prebuild.

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -42,7 +42,6 @@
 //   the other information on.
```
kraenhansen, line 42: As I moved more scripts into TypeScript I took the time to "translate" the submit-analytics script as well.

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -56,12 +55,7 @@ import machineId from "node-machine-id";
 import createDebug from "debug";
 export const debug = createDebug("realm:submit-analytics");
 
-export { collectPlatformData };
-
-// emulate old __dirname: https://flaviocopes.com/fix-dirname-not-defined-es-module-scope/
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
-const realmPackagePath = path.resolve(__dirname, "..");
+const realmPackagePath = path.resolve(__dirname, "../..");
```
kraenhansen, line 58: It moved up a level because of the `src` directory.

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -122,17 +116,15 @@ function writePackageJson(packagePath, content) {
...
+    process.env["NODE_ENV"] === "test"
+  ) {
+    return true;
   }
-
-  // If the user has specifically opted-out or if we're running in a CI environment
-  isDisabled |= "REALM_DISABLE_ANALYTICS" in process.env || "CI" in process.env;
-
-  return isDisabled;
 }
```
kraenhansen, line 128: Refactored this to use returns instead of the more exotic `|=` operator.

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -175,15 +170,15 @@ function getInstallationMethod() {
   if (userAgent) {
     return userAgent.split(" ")[0].split("/");
   } else {
-    return "unknown";
+    return ["unknown", "?.?.?"];
```
kraenhansen, line 173: This was actually a bug I discovered during the refactor.

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -309,13 +304,11 @@ async function collectPlatformData(packagePath = getProjectRoot()) {
...
-async function submitAnalytics() {
+export async function submitAnalytics() {
   const data = await collectPlatformData();
   const payload = {
-    webHook: {
-      event: "install",
-      properties: data,
-    },
+    event: "install",
+    properties: data,
```
kraenhansen, line 311: No need to wrap this in a `webHook` object, since it was always used as `payload.webHook`.

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -349,7 +342,3 @@ async function submitAnalytics() {
       });
   });
 }
-
-if (process.argv[1] === fileURLToPath(import.meta.url)) {
-  submitAnalytics().catch(console.error);
-}
```
kraenhansen, line 355: This remains in the calling `<root>/script/submit-analytics.js`.

packages/realm/src/scripts/build/xcode.ts
```diff
@@ -0,0 +1,141 @@
...
+  "iphonesimulator",
+  "macosx",
+  // "tvos",
+  // "tvsimulator",
+  // "watchos",
+  // "watchsimulator",
+  // "xros",
+  // "xrsimulator",
+  // "maccatalyst",
+] as const;
```
kraenhansen, line 37: It should be fairly simple for us to increase support for these platforms too.
It seems to me the non-trivial part of this is to ensure we're testing on those platforms too.

packages/realm/scripts/build-node-pre-gyp.ps1
```diff
@@ -1,29 +0,0 @@
-<#
```
kraenhansen, line 1: This wasn't used anymore.

packages/realm/scripts/build-ios.sh
```diff
@@ -1,125 +0,0 @@
-#!/usr/bin/env bash
```
kraenhansen, line 1: Re-written into `src/scripts/build/apple.ts`.

packages/realm/scripts/build-android.ts
```diff
@@ -1,148 +0,0 @@
-////////////////////////////////////////////////////////////////////////////
```
kraenhansen, line 1: Re-written into `src/scripts/build/android.ts`.

packages/realm/react-native/ios/RealmReact/RealmReact.xcconfig
```diff
@@ -1,2 +0,0 @@
-// This is a dummy base configuration file so that Xcode doesn't decide to
-// inherit from the parent ReactTests project when building in a workspace
```
kraenhansen, line 2: From this comment, it doesn't seem we need this file anymore.

packages/realm/react-native/android/build.gradle
```diff
@@ -1,63 +0,0 @@
-buildscript {
```
kraenhansen, line 1: This got moved to `binding/android/build.gradle`.

packages/realm/react-native.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   dependency: {
     platforms: {
       android: {
-        sourceDir: "react-native/android",
+        sourceDir: "binding/android",
```
kraenhansen, line 26: I choose to move the `react-native/ios` and `react-native/android` into `binding` to co-locate it with all the other binding source-code.

packages/realm/package.json
```diff
@@ -374,11 +348,11 @@
   },
   "binary": {
     "module_name": "realm",
-    "module_path": "binding/build",
+    "module_path": "prebuilds/node",
```
kraenhansen, line 351: Trying to keep all prebuild binaries in `prebuilds`.

packages/realm/package.json
```diff
@@ -80,22 +79,18 @@
...
+    "prebuild-apple:simulator": "wireit",
+    "prebuild-android": "wireit",
+    "prebuild-node": "wireit",
     "build:ts": "wireit",
     "build:node": "wireit",
-    "build:node:prebuild": "wireit",
-    "build:node:prebuild:arm": "wireit",
-    "build:node:prebuild:arm64": "wireit",
-    "build:node:prebuild:x64": "wireit",
-    "build:node:prebuild:ia32": "wireit",
```
kraenhansen, line 89: Simply passing `PREBUILD_ARCH` env on CI instead, as I don't expect cross-compiling to be a frequently used script on one of our developers machines.

packages/realm/package.json
```diff
@@ -46,12 +46,12 @@
...
       "default": "./dist/experimental/base-url.js"
     },
-    "./scripts/submit-analytics": "./scripts/submit-analytics.mjs",
+    "./scripts/submit-analytics": "./dist/scripts/submit-analytics.js",
     "./react-native.config.js": "./react-native.config.js",
     "./package.json": "./package.json"
   },
   "imports": {
-    "#realm.node": "./binding/build/realm.node"
+    "#realm.node": "./prebuilds/node/realm.node"
```
kraenhansen, line 54: I suggest co-locating all the prebuilds.

packages/realm/binding/node/node.cmake
```diff
@@ -32,7 +32,7 @@ target_compile_definitions(realm-js PRIVATE
     NAPI_VERSION=6
 )
 
-if(REALM_JS_BUILD_CORE_FROM_SOURCE AND TARGET ObjectStore)
+if(TARGET ObjectStore)
```
kraenhansen, line 35: This option got removed and is now obsolete in this file.

packages/realm/binding/jsi/jsi_init.cpp
```diff
@@ -1,81 +0,0 @@
-////////////////////////////////////////////////////////////////////////////
```
kraenhansen, line 1: The contents of this file is included in the generated jsi_init.cpp.

packages/realm/binding/jsi/CMakeLists.txt
```diff
@@ -1,19 +0,0 @@
-# Using Node.js to resolve the path to the "react-native" package.
```
kraenhansen, line 1: Our jsi target is removed, because the Android target and iOS Podspec points directly to `jsi_init.cpp`.

packages/realm/binding/ios/CMakeLists.txt
```diff
@@ -1,11 +0,0 @@
-add_library(realm-js-ios STATIC
```
kraenhansen, line 1: This is now driven entirely by the podspec, making this file obsolete.

.github/workflows/pr-realm-js.yml
```diff
@@ -189,68 +214,172 @@ jobs:
...
-      - name: Build Android
-        if: ${{ matrix.variant.os == 'android' }}
-        run: npm run build:android --workspace realm -- --arch=${{matrix.variant.arch}}
+          PREBUILD_ARCH: ${{ matrix.arch }}
 
-      # Due to a limitation in upload-artifact a redundant file is needed to force
-      # preserving paths (https://github.com/actions/upload-artifact/issues/174)
       - name: Upload prebuild artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
kneth, line 223: Closes https://github.com/realm/realm-js/issues/6530

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None
 
 ### Enhancements
-* None
+* Building for iOS and Android has been optimized for compatibility with future React Native versions, by deferring compilation of JSI dependent code to the dependent app's build.
```
kneth, line 7: - Should we mention that building from source is only possible if you clone the git repository?
- Should we mention that the package now requires a lot more disk space?

packages/realm/RealmJS.podspec
kneth: Should we merge or close https://github.com/realm/realm-js/pull/5053?

packages/realm/bindgen/src/templates/jsi.ts
kneth: But will we accidentally overlook something?

packages/realm/binding/android/CMakeLists.txt
```diff
@@ -1,128 +1,74 @@
...
-endif()
-
 if (ANDROID_ABI MATCHES "^armeabi")
-    target_compile_definitions(realm-js-android PUBLIC REALM_WRAP_MEMMOVE=1)
-    target_link_options(realm-js-android PUBLIC -Wl,--wrap=memmove -Wl,--wrap=memcpy)
+    target_compile_definitions(realm-js-android-binding PUBLIC REALM_WRAP_MEMMOVE=1)
+    target_link_options(realm-js-android-binding PUBLIC -Wl,--wrap=memmove -Wl,--wrap=memcpy)
 else()
-    target_compile_definitions(realm-js-android PUBLIC REALM_WRAP_MEMMOVE=0)
+    target_compile_definitions(realm-js-android-binding PUBLIC REALM_WRAP_MEMMOVE=0)
```
kneth, line 29: Is it time to remove this workaround for an old Android version?

packages/realm/binding/node/CMakeLists.txt
```diff
@@ -0,0 +1,141 @@
...
+        set(check_flag ${flag})
+    endif()
+
+    check_cxx_compiler_flag(${check_flag} HAVE${check_flag})
+    if(HAVE${check_flag})
+        add_compile_options($<$<COMPILE_LANGUAGE:CXX>:${flag}>)
+    endif()
+endfunction()
+
+set(CMAKE_CXX_STANDARD 20)
```
kneth, line 70: Is there a way to inherit it from Core's CMakeLists.txt?

packages/realm/binding/node/CMakeLists.txt
```diff
@@ -0,0 +1,141 @@
...
+include(NodeJSTargets)
+
+# Ninja buffers output so we need to tell tools to use colors even though stdout isn't a tty.
+if(CMAKE_GENERATOR MATCHES "^Ninja")
+    if (NOT MSVC)
+        add_cxx_flag_if_supported(-fdiagnostics-color)
+    endif()
+endif()
+
+
```
kneth, line 90: Can be removed

packages/realm/src/scripts/build/cli.ts
```diff
@@ -0,0 +1,185 @@
+////////////////////////////////////////////////////////////////////////////
```
kneth, line 1: I think we can close https://github.com/realm/realm-js/issues/3708 😄

packages/realm/src/scripts/build/xcode.ts
kneth: tvOS is likely the most interesting one but I have no clue how difficult it is to run tests on tvsimulator

packages/realm/src/scripts/submit-analytics.ts
kneth: TypeScript first team :smile:

packages/realm/src/scripts/submit-analytics.ts
```diff
@@ -122,17 +116,15 @@ function writePackageJson(packagePath, content) {
...
-    isDisabled |= process.env["NODE_ENV"] === "production" || process.env["NODE_ENV"] === "test";
+  if ("REALM_DISABLE_ANALYTICS" in process.env || "CI" in process.env) {
+    return true;
+  } else if (
+    // NODE_ENV is commonly used by JavaScript framework
+    process.env["NODE_ENV"] === "production" ||
+    process.env["NODE_ENV"] === "test"
+  ) {
+    return true;
   }
```
kneth, line 127: I think you need a `return false` in the end - otherwise `isAnalyticsDsiabled()` will return `undefined`

kneth (commented):
My first pass, and I like it :smile:

The updates to the files in `contrib` have been requested for a while to help onboarding - great to see it

I would like to take a second pass later this week

.github/workflows/pr-realm-js.yml
kraenhansen: Not entirely, but it does bring us closer. Before closing that issue, I'd like for the test jobs to be split out and only depend on the artifacts that they actually need:
> refactor our workflow into jobs where a test run only depends on the upload of the prebuilt artefact that it actually needs to complete.

packages/realm/binding/node/CMakeLists.txt
kraenhansen: What? The newline?

CHANGELOG.md
kraenhansen: > Should we mention that building from source is only possible if you clone the git repository?

This isn't different from the current state, so I don't think that's needed here.

> requires a lot more disk space?

Good question 🤔 I'm unsure if that's super relevant as that shouldn't affect the final app's size.

.github/workflows/pr-realm-js.yml
```diff
@@ -189,68 +214,172 @@ jobs:
...
+            packages/realm/prebuilds/realm-*-napi-*.tar.gz
 
-  ios-xcframework:
-    name: Generate Info.plist with all frameworks in place
-    needs: [build]
-    if: ${{ success() || failure() }}
-    runs-on: macos-latest
+  prebuild-apple-archives:
+    name: Prebuild ${{ matrix.platform }} (Apple React Native)
+    runs-on: macos-latest-large
```
kneth, line 233: Do we need a large runner?

.github/workflows/pr-realm-js.yml
```diff
@@ -189,68 +214,172 @@ jobs:
...
+          # Including README.md to pin paths to the root of the repository
+          path: |
+            README.md
+            packages/realm/bindgen/vendor/realm-core/build-xcode/include/
+            packages/realm/bindgen/vendor/realm-core/build-xcode/${{ matrix.platform }}.xcarchive/Products/usr/local/lib/librealm-combined.a
+          if-no-files-found: error
+
+  prebuild-apple:
+    name: Combine Xcframework (Apple React Native)
+    runs-on: macos-latest-large
```
kneth, line 281: Do we need a large runner?

.github/workflows/pr-realm-js.yml
```diff
@@ -189,68 +214,172 @@ jobs:
...
+
+      # On CI, file timestamps change with every run so instead rely on file content hashing
+      # https://reactnative.dev/docs/build-speed#using-this-approach-on-a-ci
+      - name: Configure ccache
+        run: ccache --set-config="compiler_check=content"
+
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu' # See 'Supported distributions' for available options
```
kneth, line 355: Should we add it to `env`?

CHANGELOG.md
kneth: Let's leave it as it is

contrib/building.md
```diff
@@ -197,35 +197,60 @@ Other editors should also be able to be configured to use the `compile_commands.
 
 ## Building Realm JS
 
+In most cases, its not required to build the SDK explicitly. You can either simply download the package from `npm` or run one of the test scripts in either of the `integration-tests/environments` (which will drive the dependent build-scripts automatically). If you want to invoke these scripts manually, see the individual sections below:
```
kneth, line 200:
```suggestion
In most cases, it's not required to build the SDK explicitly. You can either simply download the package from `npm` or run one of the test scripts in either of the `integration-tests/environments` (which will drive the dependent build-scripts automatically). If you want to invoke these scripts manually, see the individual sections below:
```

contrib/debug-rn-examples.md
```diff
@@ -103,13 +103,11 @@ Since we want to keep the simplicity for our users, changes made in order to deb
...
 Make sure to build binaries for Android and iOS in packages/realm by running:
 
-```
-npm run build:ios --workspace=realm
-```
-or
-
-```
-npm run build:android --workspace=realm
+```bash
```
kneth, line 106:
```suggestion
```sh
```

packages/realm/binding/android/build.gradle
```diff
@@ -0,0 +1,97 @@
...
+            packagingOptions {
+                doNotStrip "**/librealm.so"
+            }
+        }
+    }
+
+    externalNativeBuild {
+        cmake {
+            path file('CMakeLists.txt')
+            version '3.22.1'
```
kneth, line 79: We use 3.29.2 in our PR workflow

packages/realm/binding/node/CMakeLists.txt
kneth: yes

kneth, line 90:
```suggestion
```

kneth (approved):
I suppose we don't need `COMPATIBILITY.md` in the future.  Please consider to add a note at the top of the file about working with multiple RN versions. And remove it from the check-list in the PR template.

It is very good to see this project coming to an end. It will give us more confidence when releasing!

packages/realm/src/scripts/build/xcode.ts
kraenhansen: I'd say `xros` would also be an interesting platform to support.

packages/realm/src/scripts/build/cli.ts
kraenhansen: I'd like to get the Node build into this CLI too, before closing #3708

.github/workflows/pr-realm-js.yml
kraenhansen: I'm experimenting with the `large` ... it's currently `xlarge` https://github.com/realm/realm-js/blob/b121047ac34e632bac452d1627ab9b8734929018/.github/workflows/pr-realm-js.yml#L77-L79

We might be able to get smaller, but I'd like to experiment with that separately from this PR 👍

.github/workflows/pr-realm-js.yml
kraenhansen: Same as above.

packages/realm/binding/android/build.gradle
kraenhansen: I don't think the version of CMake matters too much here, as long as it meets the minimum requirements of our CMakeList.txt file. We're aligning this with the version with React Native to cause the least amount of friction to developers: https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/build.gradle.kts#L44-L47

.github/workflows/pr-realm-js.yml
kneth: Sounds good to experiment later

packages/realm/binding/android/build.gradle
kneth: Why not say 3.22.1 in our PR workflow?

kraenhansen: In general, I don't see a reason not to. I just wanted to keep the changes of this PR to a minimum and as you pointed out 3.29.2 is what's currently being used: https://github.com/realm/realm-js/blob/b121047ac34e632bac452d1627ab9b8734929018/.github/workflows/pr-realm-js.yml#L116

packages/realm/binding/android/CMakeLists.txt
```diff
@@ -1,128 +1,74 @@
...
-add_library(realm-js-android SHARED
-    # $<TARGET_OBJECTS:realm-js-shared>
-    hack.cpp
-    platform.cpp
-    jni_utils.cpp
-    io_realm_react_RealmReactModule.cpp
-    ${REACT_NATIVE_ROOT_DIR}/ReactAndroid/src/main/jni/react/turbomodule/ReactCommon/CallInvokerHolder.cpp
+add_library(realm-js-android-binding SHARED
+    ../jsi/jsi_init.cpp
+    src/main/cpp/hack.cpp
```
nhachicha, line 12: I think we can remove this hack, since it's only targeting `armeabi`/`ARMv5` 32-bit Android which was removed in NDK r17

packages/realm/binding/android/CMakeLists.txt
```diff
@@ -1,128 +1,74 @@
...
-        COMMAND ${CMAKE_COMMAND} -E tar xfz "fbjni-0.2.2.aar"
-        WORKING_DIRECTORY "${FBJNI_ROOT_DIR}"
-    )
-endif()
-
 if (ANDROID_ABI MATCHES "^armeabi")
-    target_compile_definitions(realm-js-android PUBLIC REALM_WRAP_MEMMOVE=1)
-    target_link_options(realm-js-android PUBLIC -Wl,--wrap=memmove -Wl,--wrap=memcpy)
+    target_compile_definitions(realm-js-android-binding PUBLIC REALM_WRAP_MEMMOVE=1)
+    target_link_options(realm-js-android-binding PUBLIC -Wl,--wrap=memmove -Wl,--wrap=memcpy)
```
nhachicha, line 27: The flag should be removed if we remove the `hack.cpp` (see above)

packages/realm/binding/android/build.gradle
```diff
@@ -0,0 +1,97 @@
...
+    buildToolsVersion rootProject.hasProperty('buildToolsVersion') ? rootProject.buildToolsVersion : '28.0.3'
+    buildFeatures {
+        prefab = true
+    }
+    defaultConfig {
+        minSdkVersion rootProject.hasProperty('minSdkVersion') ? rootProject.minSdkVersion : 16
+        targetSdkVersion rootProject.hasProperty('targetSdkVersion') ? rootProject.targetSdkVersion : 28
+        externalNativeBuild {
+            cmake {
+                arguments "-DANDROID_STL=c++_shared",
```
nhachicha, line 58: Try adding `-DCMAKE_CXX_VISIBILITY_PRESET=hidden` to avoid exporting Core symbols, this will reduce a little bit the binary size and help integrate with another SDK compiling Core statically

kraenhansen: I suspected that was handled by this: https://github.com/realm/realm-js/blob/f84f3321764831b096848002c7c2bfbffe30f3e6/packages/realm/binding/android/CMakeLists.txt#L35

kraenhansen: With [this change](https://github.com/realm/realm-js/pull/6650/commits/ec9be5e91be5c8aa917a426c8dc393ded8f724bd) I managed to bring down the .so's from ~ 100MB each to:

```
 12M    binding/android/src/main/jniLibs/arm64-v8a/librealm.so
6.9M    binding/android/src/main/jniLibs/armeabi-v7a/librealm.so
 12M    binding/android/src/main/jniLibs/x86/librealm.so
 13M    binding/android/src/main/jniLibs/x86_64/librealm.so
 ```

For some reason, building for release (I test this calling `./gradlew realm:assembleRelease` in the integration test app), I was running `realm:configureCMakeRelWithDebInfo` - now it runs `realm:configureCMakeRelease`.

Comment (nhachicha) [reply packages/realm/binding/android/build.gradle:58]: These are the numbers we have for Kotlin as well https://mvnrepository.com/artifact/io.realm.kotlin/cinterop-android/1.16.0 👍

**Review by elle-j — APPROVED** [review]
Amazing effort, research, and execution of this project, both from you and @takameyer  🎉  Glad to see this close to landing!

File: packages/realm/binding/apple/RealmReact.mm (hunk 5 above)
Comment (kraenhansen) [reply packages/realm/binding/apple/RealmReact.mm:150]: Oh it was.

File: packages/realm/binding/apple/RealmReact.mm [hunk 45]
```diff
@@ -144,10 +144,10 @@ - (void)setBridge:(RCTBridge *)bridge {
             // because the work is performed before the microtask queue is flushed - see sequence
             // diagram at https://bit.ly/3kexhHm. It might be possible to further optimize this,
             // e.g. only flush the queue a maximum of once per frame, but this seems reasonable.
-            if (!waitingForUiFlush) {
-              waitingForUiFlush = true;
+            if (!self->waitingForUiFlush) {
```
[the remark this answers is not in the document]
kraenhansen: Yep 🙂

packages/realm/src/scripts/build/android.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export const SUPPORTED_ARCHITECTURES = [
+  "x86",
+  "armeabi-v7a",
+  "arm64-v8a",
+  "x86_64",
+] as const satisfies readonly AndroidArchitecture[];
+
+export function validateArchitectures(
+  values: readonly (AndroidArchitecture | "all" | "infer")[],
```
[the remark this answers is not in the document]
elle-j: Sounds good, or extending the implementation with validation to fail early.

packages/realm/RealmJS.podspec
kraenhansen: Perhaps 🤔 I haven't double checked the current state of the Node.js binding.

packages/realm/binding/node/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+    set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
+    add_link_options(-fuse-ld=gold)
+    add_link_options(-Wl,-z,noexecstack)
+    add_link_options(-Wl,--warn-execstack)
+    elseif(NODE_ARCH STREQUAL "x64")
+    set(CMAKE_TOOLCHAIN_FILE "${BINDGEN_DIR}/vendor/realm-core/tools/cmake/x86_64-linux-gnu.toolchain.cmake")
+    add_link_options(-fuse-ld=gold)
+    add_link_options(-Wl,-z,noexecstack)
+    add_link_options(-Wl,--warn-execstack)
+    endif()
```
kraenhansen, line 43:
```suggestion
    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13")
    if(NODE_ARCH STREQUAL "arm64")
    set(CMAKE_OSX_ARCHITECTURES "arm64" CACHE STRING "")
    else()
    set(CMAKE_OSX_ARCHITECTURES "x86_64" CACHE STRING "")
    endif()
elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
    if(NODE_ARCH STREQUAL "arm")
        set(CMAKE_TOOLCHAIN_FILE "${BINDGEN_DIR}/vendor/realm-core/tools/cmake/armv7-linux-gnueabihf.toolchain.cmake")
        add_link_options(-fuse-ld=bfd)
        # due to how multiarch works in debian this is needed to link to the correct system libraries
        set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
    elseif(NODE_ARCH STREQUAL "arm64")
        set(CMAKE_TOOLCHAIN_FILE "${BINDGEN_DIR}/vendor/realm-core/tools/cmake/aarch64-linux-gnu.toolchain.cmake")
        set(CMAKE_IGNORE_PATH "/usr/lib/x86_64-linux-gnu")
        add_link_options(-fuse-ld=gold)
        add_link_options(-Wl,-z,noexecstack)
        add_link_options(-Wl,--warn-execstack)
    elseif(NODE_ARCH STREQUAL "x64")
        set(CMAKE_TOOLCHAIN_FILE "${BINDGEN_DIR}/vendor/realm-core/tools/cmake/x86_64-linux-gnu.toolchain.cmake")
        add_link_options(-fuse-ld=gold)
        add_link_options(-Wl,-z,noexecstack)
        add_link_options(-Wl,--warn-execstack)
    endif()
```

packages/realm/binding/node/CMakeLists.txt
kraenhansen: Might be, but technically the Core CMake project is a dependency of ours, so that would be a bit reversed 🤔

packages/realm/src/scripts/build/android.ts
```diff
@@ -0,0 +1,180 @@
...
+        "CMAKE_MAKE_PROGRAM=ninja",
+        "-D",
+        "CMAKE_C_COMPILER_LAUNCHER=ccache",
+        "-D",
+        "CMAKE_CXX_COMPILER_LAUNCHER=ccache",
+        "-D",
+        `ANDROID_NDK=${ndkPath}`,
+        "-D",
+        `ANDROID_ABI=${architecture}`,
+        // TODO: Do we need both the above and below?
```
kraenhansen, line 120:
```suggestion
```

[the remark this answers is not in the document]
kraenhansen: I actually deleted the line this was referring to after looking into it.

contrib/building.md
```diff
@@ -197,35 +197,60 @@ Other editors should also be able to be configured to use the `compile_commands.
...
+
 ### Building for iOS
 
 You can build and bundle for iOS by running the following command from the root directory:
 
 ```sh
-npm run build:ios --workspace realm
-npm run bundle --workspace realm
+# Pre-build Realm Core into an XCFramework
+npm run prebuild-apple --workspace realm
```
takameyer, line 216: Super nitpicky, but curious why we use a `-` for one separator and `:` for the other.

packages/realm/src/scripts/build/cli.ts
```diff
@@ -0,0 +1,182 @@
...
+
+program
+  .command("build-android")
+  .description("Build native code for Android platforms")
+  .option("--clean", "Delete any build directory first", false)
+  .addOption(androidArchOption)
+  .addOption(configurationOption)
+  .option("--ndk-version <version>", "The NDK version used when building", android.DEFAULT_NDK_VERSION)
+  .option("--skip-collecting-headers", "Skip collecting headers from the build directory and copy them to the SDK")
+  // .option("--skip-generating-version-file", "Skip generating a Version.java file")
```
takameyer, line 157: Delete this line?

packages/realm/src/scripts/build/xcode.ts
```diff
@@ -0,0 +1,141 @@
...
+  },
+  archive({ cwd, scheme, targets = [], configuration, sdkName, destinations = [], archivePath }: XcodeArchiveOptions) {
+    const { status } = spawnSync(
+      "xcodebuild",
+      [
+        "archive",
+        "-scheme",
+        scheme,
+        ...targets.flatMap((target) => ["-target", target]),
+        // "-project"
```
takameyer, line 86: Remove this line?

takameyer (approved):
What an effort! Thanks for finishing this extremely difficult task.  This will relieve a lot of stress as new versions of Expo and React Native are released.

packages/realm/binding/node/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
+    set(CMAKE_SYSTEM_VERSION 8.1)
+    if(CMAKE_GENERATOR_PLATFORM MATCHES "^[Ww][Ii][Nn]32$")
+        set(VCPKG_TARGET_TRIPLET "x86-windows-static" CACHE STRING "")
+    elseif(CMAKE_GENERATOR_PLATFORM MATCHES "^[Xx]64$")
+        set(VCPKG_TARGET_TRIPLET "x64-windows-static" CACHE STRING "")
+    else()
+        message(FATAL_ERROR "Unable to deduce vcpkg target triplet (from '${NODE_ARCH}')")
+    endif()
+elseif(APPLE)
```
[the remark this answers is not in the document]
kraenhansen: I believe I've fixed this 👍

contrib/building.md
kraenhansen: The honest reason is because I don't want `prebuild-node` to act as an NPM pre-script for the existing `build:node` and it would if it was `prebuild:node` 🙈 We could revisit this later, but this was a fast and (only slightly dirty workaround).

packages/realm/src/scripts/build/android.ts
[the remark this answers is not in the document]
kraenhansen: I went with `pickArchitectures` 🙈

integration-tests/environments/react-native-test-app/package.json
```diff
@@ -107,6 +123,18 @@
           "default": "1"
         }
       }
+    },
+    "realm:build-android:debug": {
+      "command": "cd android && ./gradlew realm:assembleDebug",
```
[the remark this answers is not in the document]
kraenhansen: We talked about this face-2-face 👍

packages/realm/src/scripts/build/android.ts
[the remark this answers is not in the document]
elle-j: [apple.ts](https://github.com/realm/realm-js/blob/0d3d38d3fdd1bd5040ad903f9868679493c5ec28/packages/realm/src/scripts/build/apple.ts#L64) could also benefit from that I think 🙂

PR #6665, opened 2024-05-13
Fix performance tests

elle-j:
## What, How & Why?

Fixes the broken performance tests and makes the type outputs from the tests consistent (no matter how you pass in the property schema).

### DIFF
--- a/integration-tests/tests/src/performance-tests/property-reads.ts
+++ b/integration-tests/tests/src/performance-tests/property-reads.ts
@@ -16,53 +16,52 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import Realm, { ObjectSchema, PropertySchema } from "realm";
+import { expect } from "chai";
+import Realm, { BSON, ObjectSchema, PropertySchema, PropertySchemaShorthand } from "realm";
 
 import { describePerformance } from "../utils/benchmark";
 
 type Value = ((realm: Realm) => unknown) | unknown;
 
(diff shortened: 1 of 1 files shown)


integration-tests/tests/src/performance-tests/property-reads.ts
```diff
@@ -16,53 +16,59 @@
...
-  if (typeof type === "object") {
-    const prefix = type.optional ? "optional " : "";
-    if (type.objectType) {
-      return prefix + `${type.type}<${type.objectType}>`;
-    } else {
-      return prefix + type.type;
-    }
-  } else {
-    return type;
+const COLLECTION_MARKERS: Readonly<Record<string, string>> = {
```
elle-j, line 26: Widening the type here for better ergonomics when bracket accessing it by an unknown string (and that's the only way it's used in this file).

integration-tests/tests/src/performance-tests/property-reads.ts
```diff
@@ -16,53 +16,59 @@
...
+};
+
+/**
+ * Get a representative consistent name of the type depending on the schema.
+ *
+ * @example
+ * "int?[]"                                            -> "int?[]"
+ * { type: "list", objectType: "int", optional: true } -> "int?[]"
+ */
+function getTypeDisplayName(schema: PropertySchemaShorthand | PropertySchema) {
```
elle-j, line 39: Renaming from `getTypeName` to disambiguate it from the various usages of `type` we have.

kneth (approved):

kraenhansen (approved):

PR #6673, opened 2024-05-22
Adding --latest-local to the baas test server CLI

kraenhansen:
## What, How & Why?

I got annoyed that our BaaS test server always had to pull the latest baas image and I found it cumbersome to figure out the latest baas image that I had locally. So I turned this into a runtime argument for our BaaS test server command:

```
npm start --workspace @realm/baas-test-server -- --latest-local
```

It this works well for people, we might turn this into the default behaviour and instead add a `--pull-latest` flag.

## ☑️ ToDos
* [x] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/integration-tests/baas-test-server/cli.ts
+++ b/integration-tests/baas-test-server/cli.ts
@@ -84,7 +84,11 @@ yargs(hideBin(process.argv))
   .command(
     ["docker [githash]"],
     "Runs the BaaS test image using Docker",
-    (yargs) => yargs.positional("githash", { type: "string" }).option("branch", { default: "master" }),
+    (yargs) =>
+      yargs
+        .positional("githash", { type: "string" })
+        .option("branch", { default: "master" })
+        .option("latest-local", { default: false, boolean: true }),
     wrapCommand(async (argv) => {
       const { AWS_PROFILE, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY } = process.env;
       assert(AWS_ACCESS_KEY_ID && AWS_SECRET_ACCESS_KEY, "Missing AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY env");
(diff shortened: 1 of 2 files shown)


kneth (approved):
Merge when tests pass :-)

integration-tests/baas-test-server/cli.ts
```diff
@@ -95,12 +99,19 @@ yargs(hideBin(process.argv))
...
+          image: argv.githash,
+          accessKeyId: AWS_ACCESS_KEY_ID,
+          secretAccessKey: AWS_SECRET_ACCESS_KEY,
+        });
+      } else if (argv["latest-local"]) {
+        const id = docker.getLatestLocalId();
+        docker.spawnBaaS({ image: id, accessKeyId: AWS_ACCESS_KEY_ID, secretAccessKey: AWS_SECRET_ACCESS_KEY });
       } else {
         const tag = await docker.fetchBaasTag(argv.branch);
         assert(AWS_PROFILE, "Missing AWS_PROFILE env");
```
elle-j, line 112: Would it make sense to put this `assert` before fetching?

integration-tests/baas-test-server/cli.ts
```diff
@@ -95,12 +99,19 @@ yargs(hideBin(process.argv))
...
         docker.pullBaas({ profile: AWS_PROFILE, tag });
```
elle-j, line 113: We may want to rename this `tag` property to `image` if doing it for the `spawnBaaS()` argument.

elle-j (approved):

integration-tests/baas-test-server/cli.ts
kraenhansen: It's always a tag when it comes from baasaas 👍

integration-tests/baas-test-server/cli.ts
kraenhansen: Perhaps 🤔 I think I like this to be next to the line that relies on it, to increase the likelihood that they're moved together in a future refactoring.

elle-j: 🙂

PR #6675, opened 2024-05-23
Adding types for passing objects when constructing objects

kraenhansen:
## What, How & Why?

This adds types to allow passing object links and embedded objects when creating objects using a class-based model constructor:

```typescript
import Realm from "realm";

class Contact extends Realm.Object<Contact, "name"> {
    name!: string;
    address!: Address;

    static schema: Realm.ObjectSchema = {
        name: "Contact",
        properties: {
            name: "string",
            address: "Address",
        },
    };
}

class Address extends Realm.Object<Address, "street" | "city"> {
    street!: string;
    city!: string;

    static schema: Realm.ObjectSchema = {
        name: "Address",
        embedded: true,
        properties: {
            street: "string",
            city: "string",
        },
    };
}

Realm.deleteFile({});
let realm = new Realm({ schema: [Contact, Address] });
realm.write(() => {
    const alice = new Contact(realm, {
        name: "Alice",
        address: {
            street: "Grøndals Parkvej 2A",
            // city: "Vanløse"
        },
    });

    const bob = new Contact(realm, {
        name: "Bob",
        address: {
            street: "Griffenfeldsgade 14B",
            city: "København N",
        }
    });
});

realm.objects<Contact>(Contact.schema.name).forEach((c) => {
    console.log(`${c.name} at ${c.address?.street}, ${c.address?.city}`);
});
```

Unfortunately, the `RequiredProperties` passed as type parameter (`"street" | "city"`) when declaring `Address` doesn't cascade as expected, which is why `alice` can be created without passing `city`. As an alternative `RealmObjectRemappedModelPart` type could be extended to make all properties required, but I haven't been able to express the constructor type from the instance type `T[K]` in a way that I could infer the `RequiredProperties` correctly.

## ☑️ ToDos
* [ ] Fix the required properties
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/src/Object.ts
+++ b/packages/realm/src/Object.ts
@@ -74,7 +74,7 @@ type CreationContext = {
 };
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type AnyRealmObject = RealmObject<any>;
+export type AnyRealmObject = RealmObject<any, any>;
 
 export const KEY_ARRAY = Symbol("Object#keys");
 export const KEY_SET = Symbol("Object#keySet");
--- a/packages/realm/src/Unmanaged.ts
+++ b/packages/realm/src/Unmanaged.ts
@@ -45,12 +45,22 @@ type RealmDictionaryRemappedModelPart<T> = {
     : never;
 };
 
+/**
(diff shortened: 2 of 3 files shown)


kraenhansen:

PR #6678, opened 2024-05-23
Add internal annotation to fix docs build

elle-j:
## What, How & Why?

Our docs build broke due to a function not marked as `@internal`.

(Also adding to the lint workflow to check the public types generated.)


kraenhansen (approved):

kraenhansen:
We should probably run this during the lint workflow:

```
npm run check-types --workspace realm
```

elle-j:
Good idea 👍 I'll add that

kneth (approved):

PR #6686, opened 2024-05-27
Simplify RealmJS.podspec to avoid passing the version

kraenhansen:
## What, How & Why?

I believe this was only used by https://github.com/realm/realm-js/blob/6593a851cd0a7f49fb85ad13bb7c4e1fd245f346/react-native/ios/RealmReact/RealmAnalytics.mm#L69C23-L69C43 and we've later refactored the analytics to send this information from the SDK: [here](https://github.com/realm/realm-js/blob/fc633f70b69596d84f7dce5a2d7f1fac9f8e25e1/packages/realm/src/platform/react-native/device-info.ts#L21) and [here](https://github.com/realm/realm-js/blob/fc633f70b69596d84f7dce5a2d7f1fac9f8e25e1/packages/realm/src/platform/react-native/device-info.ts#L65).


elle-j (approved):
Good catch 👍

PR #6694, opened 2024-05-31
RJS-1413: Add support for a logical counter as a presentation data type

### LINKED ISSUE #4089: Add support for counters
### Problem

Realm Core supports a special case of integers: counters. Counters are in particular useful for synced Realms.

See also https://github.com/realm/realm-core/issues/5056

### Solution

We'll be adding a new property type named "counter", which will use the "int" core type underneath but return instances of a `Counter` class instead of raw `number`.

```typescript
declare class Counter {
  public get value(): number;
  public increment(by = 1);
  public decrement(by = 1);
  public set(value: number);

  // Implements https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf
  // this would allow using an instance of `Counter` to be coerced into a `number`.
  public valueOf(): number {
    return this.value;
  }
}
```

Because we cannot overload the increment operators `++`, `+=`, `--` and `-=` explicitly in JS (these simply call "set" / assignment underneath), we suggest throwing if these are used, to prevent users from relying on these operators which wouldn't use the proper counter semantics.

We discussed implementing a heuristic (if the setter is invoked just after the getter, perhaps users are performing a `+=`), but we wouldn't be able to distinguish that from a `*= 2`, in which case it would be misleading to use the counter semantics.

_No response_

elle-j:
## What, How & Why?

A `Counter` class and `counter` presentation data type have been added, representing a logical counter.

Note that `counter` types are **not** supported as:
* `Mixed` values
* Primary keys
* Inside collections
* Query arguments for placeholders (use `Counter.value`)

## Usage

**Valid property schema:**
```typescript
// Shorthand
"counter"
"counter?"

// Object form
{
  type: "int",
  presentation: "counter",
  optional: <true/false/undefined>
}
```

**Creating an object with a counter:**
```typescript
realm.write(() => {
  realm.create(MyObject, { _id: "123", counter: 0 });
});
```

**Updating the count:**
```typescript
realm.write(() => {
  object.counter.increment();
  object.counter.value; // 1
  object.counter.decrement(2);
  object.counter.value; // -1
  object.counter.set(100);
  object.counter.value; // 100
});
```

**Creating a counter from a previously `null` value:**
```ts
realm.write(() => {
  realm.create(MyObject, { _id: "123", counter: 0 }, UpdateMode.Modified);
});
```

## Notes

* **No implicit coercion to `number`**:
  * We decided not to support implicit coercion into number via `valueOf` partly due to TypeScript [not supporting it](https://github.com/microsoft/TypeScript/issues/2361).
* **Potential future reference**
  * We decided to remove support for collections of counters.
    * If this becomes relevant in the future, [this is the commit](https://github.com/realm/realm-js/pull/6694/commits/b975c563a60f3b05a0f22a859c1855ead8d1afa7) that removed related implementation and tests.

## Test overview

<details>
<summary>Click to expand</summary>

```
  Counter
    create and access
      via 'realm.create()'
        ✓ can create and access (input: number)
        ✓ can create and access (input: Counter)
        ✓ can create and access (input: default value)
        ✓ can create nullable counter with int or null
        ✓ returns different reference for each access
    update
      Counter.value
        ✓ increments
        ✓ decrements
        ✓ sets
      Realm object counter property
        ✓ updates nullable counter from int -> null -> int via setter
        ✓ updates nullable counter from int -> null -> int via UpdateMode: modified
        ✓ updates nullable counter from int -> null -> int via UpdateMode: all
    filtering
      ✓ filters objects with counters
    Realm.schema
      ✓ includes `presentation: 'counter'` in the canonical property schema
    invalid operations
      ✓ throws when calling increment with non-integer
      ✓ throws when calling decrement with non-integer
      ✓ throws when calling set with non-integer
      ✓ throws when setting 'Counter.value' directly
      ✓ throws when updating outside write transaction
      ✓ throws when setting a non-nullable counter via setter
      ✓ throws when setting a non-nullable counter via UpdateMode: modified
      ✓ throws when setting a non-nullable counter via UpdateMode: all
      ✓ throws when setting a nullable counter from number -> number via setter
      ✓ throws when setting a nullable counter from number -> number via UpdateMode: modified
      ✓ throws when setting a nullable counter from number -> number via UpdateMode: all
      ✓ throws when setting an int property to a counter
      ✓ throws when getting the count on an invalidated obj
      ✓ throws when setting a mixed to a counter via object creation
      ✓ throws when setting a mixed to a counter via object setter
      ✓ throws when adding a counter to mixed collections via object creation
      ✓ throws when adding a counter to mixed collections via setters
      ✓ throws when using counter as query argument

  normalizePropertySchema
    Shorthand notation
      Valid combinations
        'counter'
          ✔ normalizes 'counter'
          ✔ normalizes 'counter?'
      Invalid shorthand notation
        ✔ throws when normalizing 'counter[]'
        ✔ throws when normalizing 'counter?[]'
        ✔ throws when normalizing 'counter{}'
        ✔ throws when normalizing 'counter?{}'
        ✔ throws when normalizing 'counter<>'
        ✔ throws when normalizing 'counter?<>'
    Object notation
      Valid object notation
        'counter'
          ✔ normalizes { type: 'int', presentation: 'counter' }
          ✔ normalizes { type: 'int', presentation: 'counter', optional: false }
          ✔ normalizes { type: 'int', presentation: 'counter', optional: true }
      Invalid object notation
        Mixing shorthand inside object notation
          ✔ throws when normalizing { type: 'counter' }
          ✔ throws when normalizing { type: 'list', objectType: 'counter?' }
        'counter' & collection combinations
          ✔ throws when normalizing { type: 'list', objectType: 'int', presentation: 'counter' }
          ✔ throws when normalizing { type: 'list', objectType: 'int', presentation: 'counter', optional: false }
          ✔ throws when normalizing { type: 'list', objectType: 'int', presentation: 'counter', optional: true }
          ✔ throws when normalizing { type: 'dictionary', objectType: 'int', presentation: 'counter' }
          ✔ throws when normalizing { type: 'dictionary', objectType: 'int', presentation: 'counter', optional: false }
          ✔ throws when normalizing { type: 'dictionary', objectType: 'int', presentation: 'counter', optional: true }
          ✔ throws when normalizing { type: 'set', objectType: 'int', presentation: 'counter' }
          ✔ throws when normalizing { type: 'set', objectType: 'int', presentation: 'counter', optional: false }
          ✔ throws when normalizing { type: 'set', objectType: 'int', presentation: 'counter', optional: true }
        Indexed and primary key combinations
          ✔ throws when normalizing { type: 'int', presentation: 'counter' } (primary key)
```
</details>

This closes #4089

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+## vNext (TBD)
+
+### Deprecations
+* None
+
+### Enhancements
+* None
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
+* None
+
+### Compatibility
+* React Native >= v0.71.4
+* Realm Studio v15.0.0.
+* File format: generates Realms with format v24 (reads and upgrades file format v10).
+
+### Internal
+<!-- * Either mention core version or upgrade -->
+<!-- * Using Realm Core vX.Y.Z -->
+<!-- * Upgraded Realm Core from vX.Y.Z to vA.B.C -->
+
 ## 12.10.0-rc.0 (2024-05-31)
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
 * Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
 * A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. ([#6694](https://github.com/realm/realm-js/pull/6694))
+  * See the [API docs](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/Realm.Types.Counter.html) for more information about the usage, or get a high-level introduction about counters in the [documentation](https://www.mongodb.com/docs/atlas/app-services/sync/details/conflict-resolution/#counters).
 ```typescript
 class MyObject extends Realm.Object {
   _id!: BSON.ObjectId;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,14 @@
 * None
 
 ### Enhancements
-* None
+* Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
+* Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None
+* After compacting, a file upgrade would be triggered. This could cause loss of data for synced Realms. ([realm/realm-core#7747](https://github.com/realm/realm-core/issues/7747), since 12.7.0-rc.0)
+* The function `immediatelyRunFileActions` was not added to the bindgen's opt-list. This could lead to the error `TypeError: app.internal.immediatelyRunFileActions is not a function`. ([#6708](https://github.com/realm/realm-js/issues/6708), since v12.8.0)
+* Encrypted files on Windows had a maximum size of 2 GB even on x64 due to internal usage of `off_t`, which is a 32-bit type on 64-bit Windows. ([realm/realm-core#7698](https://github.com/realm/realm-core/pull/7698), since the introduction of encryption support on Windows - likely in v1.11.0)
+* Tokenizing strings for full-text search could lead to undefined behavior. ([realm/realm-core#7698](https://github.com/realm/realm-core/pull/7698), since v11.3.0-rc.0)
 
 ### Compatibility
 * React Native >= v0.71.4
@@ -16,15 +19,11 @@
 * File format: generates Realms with format v24 (reads and upgrades file format v10).
 
 ### Internal
-<!-- * Either mention core version or upgrade -->
-<!-- * Using Realm Core vX.Y.Z -->
-<!-- * Upgraded Realm Core from vX.Y.Z to vA.B.C -->
+* Upgraded Realm Core from v14.7.0 to v14.10.0. ([#6701](https://github.com/realm/realm-js/issues/6701))
 
 ## 12.10.0-rc.0 (2024-05-31)
 
 ### Enhancements
-* Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
-* Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
 * A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. ([#6694](https://github.com/realm/realm-js/pull/6694))
   * See the [API docs](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/Realm.Types.Counter.html) for more information about the usage, or get a high-level introduction about counters in the [documentation](https://www.mongodb.com/docs/atlas/app-services/sync/details/conflict-resolution/#counters).
 ```typescript
@@ -58,19 +57,11 @@ realm.write(() => {
 
 ### Fixed
 * A non-streaming progress notifier would not immediately call its callback after registration. Instead you would have to wait for a download message to be received to get your first update - if you were already caught up when you registered the notifier you could end up waiting a long time for the server to deliver a download that would call/expire your notifier. ([#7627](https://github.com/realm/realm-core/issues/7627), since v12.8.0)
-* After compacting, a file upgrade would be triggered. This could cause loss of data for synced Realms. ([realm/realm-core#7747](https://github.com/realm/realm-core/issues/7747), since 12.7.0-rc.0)
-* The function `immediatelyRunFileActions` was not added to the bindgen's opt-list. This could lead to the error `TypeError: app.internal.immediatelyRunFileActions is not a function`. ([#6708](https://github.com/realm/realm-js/issues/6708), since v12.8.0)
-* Encrypted files on Windows had a maximum size of 2 GB even on x64 due to internal usage of `off_t`, which is a 32-bit type on 64-bit Windows. ([realm/realm-core#7698](https://github.com/realm/realm-core/pull/7698), since the introduction of encryption support on Windows - likely in v1.11.0)
-* Tokenizing strings for full-text search could lead to undefined behavior. ([realm/realm-core#7698](https://github.com/realm/realm-core/pull/7698), since v11.3.0-rc.0)
-
 
 ### Compatibility
 * React Native >= v0.71.4
 * Realm Studio v15.0.0.
 * File format: generates Realms with format v24 (reads and upgrades file format v10).
-
-### Internal
-* Upgraded Realm Core from v14.7.0 to v14.10.0. ([#6701](https://github.com/realm/realm-js/issues/6701))
 
 ## 12.9.0 (2024-05-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 * Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
 * Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
 * A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. ([#6694](https://github.com/realm/realm-js/pull/6694))
-```ts
+```typescript
 class MyObject extends Realm.Object {
   _id!: BSON.ObjectId;
   counter!: Realm.Types.Counter;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 ### Enhancements
 * Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
 * Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
-* A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. See the [API docs](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/Realm.Types.Counter.html) for more information. ([#6694](https://github.com/realm/realm-js/pull/6694))
+* A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. ([#6694](https://github.com/realm/realm-js/pull/6694))
 ```ts
 class MyObject extends Realm.Object {
   _id!: BSON.ObjectId;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,4 @@
-## vNext (TBD)
-
-### Deprecations
-* None
+## 12.10.0-rc.0 (2024-05-31)
 
 ### Enhancements
 * Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,35 @@
 ### Enhancements
 * Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
 * Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
+* A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. See the [API docs](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/Realm.Types.Counter.html) for more information. ([#6694](https://github.com/realm/realm-js/pull/6694))
+```ts
+class MyObject extends Realm.Object {
+  _id!: BSON.ObjectId;
+  counter!: Realm.Types.Counter;
+
+  static schema: ObjectSchema = {
+    name: "MyObject",
+    primaryKey: "_id",
+    properties: {
+      _id: { type: "objectId", default: () => new BSON.ObjectId() },
+      counter: "counter",
+      // or: counter: { type: "int", presentation: "counter" },
+    },
+  };
+}
+
+const realm = await Realm.open({ schema: [MyObject] });
+const object = realm.write(() => {
+  return realm.create(MyObject, { counter: 0 });
+});
+
+realm.write(() => {
+  object.counter.increment();
+  object.counter.value; // 1
+  object.counter.decrement(2);
+  object.counter.value; // -1
+});
+```
 
 ### Fixed
 * A non-streaming progress notifier would not immediately call its callback after registration. Instead you would have to wait for a download message to be received to get your first update - if you were already caught up when you registered the notifier you could end up waiting a long time for the server to deliver a download that would call/expire your notifier. ([#7627](https://github.com/realm/realm-core/issues/7627), since v12.8.0)
--- a/integration-tests/tests/src/tests/counter.ts
+++ b/integration-tests/tests/src/tests/counter.ts
@@ -282,4 +282,119 @@ describe("Counter", () => {
       });
     });
   });
+
+  describe("Update", () => {
+    describe("Counter.value", () => {
+      it("increments", function (this: RealmContext) {
+        const { counter } = this.realm.write(() => {
+          return this.realm.create<IWithCounter>(WithCounterSchema.name, {
+            counter: 0,
+          });
+        });
+        expectCounter(counter);
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.increment(0);
+        });
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.increment();
+        });
+        expect(counter.value).equals(1);
+
+        this.realm.write(() => {
+          counter.increment(19);
+        });
+        expect(counter.value).equals(20);
+
+        this.realm.write(() => {
+          counter.increment(-20);
+        });
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.increment();
+          counter.increment();
+          counter.increment();
+        });
+        expect(counter.value).equals(3);
+      });
+
+      it("decrements", function (this: RealmContext) {
+        const { counter } = this.realm.write(() => {
+          return this.realm.create<IWithCounter>(WithCounterSchema.name, {
+            counter: 0,
+          });
+        });
+        expectCounter(counter);
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.decrement(0);
+        });
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.decrement();
+        });
+        expect(counter.value).equals(-1);
+
+        this.realm.write(() => {
+          counter.decrement(19);
+        });
+        expect(counter.value).equals(-20);
+
+        this.realm.write(() => {
+          counter.decrement(-20);
+        });
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.decrement();
+          counter.decrement();
+          counter.decrement();
+        });
+        expect(counter.value).equals(-3);
+      });
+
+      it("sets", function (this: RealmContext) {
+        const { counter } = this.realm.write(() => {
+          return this.realm.create<IWithCounter>(WithCounterSchema.name, {
+            counter: 0,
+          });
+        });
+        expectCounter(counter);
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.set(0);
+        });
+        expect(counter.value).equals(0);
+
+        this.realm.write(() => {
+          counter.set(1);
+        });
+        expect(counter.value).equals(1);
+
+        this.realm.write(() => {
+          counter.set(20);
+        });
+        expect(counter.value).equals(20);
+
+        this.realm.write(() => {
+          counter.set(100_000);
+        });
+        expect(counter.value).equals(100_000);
+
+        this.realm.write(() => {
+          counter.set(1);
+          counter.set(2);
+          counter.set(3);
+        });
+        expect(counter.value).equals(3);
+      });
+    });
+  });
 });
--- a/integration-tests/tests/src/tests/counter.ts
+++ b/integration-tests/tests/src/tests/counter.ts
@@ -444,8 +444,7 @@ describe("Counter", () => {
         expect(counter.value).equals(-3);
       });
 
-      // TODO(lj): To be implemented.
-      it.skip("sets", function (this: RealmContext) {
+      it("sets", function (this: RealmContext) {
         const { counter } = this.realm.write(() => {
           return this.realm.create<IWithCounter>(WithCounterSchema.name, {
             counter: 0,
@@ -683,8 +682,7 @@ describe("Counter", () => {
       expect(counter.value).equals(10);
     });
 
-    // TODO(lj): To be implemented.
-    it.skip("throws when setting to non-integer", function (this: RealmContext) {
+    it("throws when setting to non-integer", function (this: RealmContext) {
       const { counter } = this.realm.write(() => {
         return this.realm.create<IWithCounter>(WithCounterSchema.name, {
           counter: 10,
@@ -697,14 +695,14 @@ describe("Counter", () => {
         this.realm.write(() => {
           counter.set(1.1);
         });
-      }).to.throw("Expected 'by' to be an integer, got a floating point number");
+      }).to.throw("Expected 'value' to be an integer, got a floating point number");
       expect(counter.value).equals(10);
 
       expect(() => {
         this.realm.write(() => {
           counter.set(NaN);
         });
-      }).to.throw("Expected 'by' to be an integer, got NaN");
+      }).to.throw("Expected 'value' to be an integer, got NaN");
       expect(counter.value).equals(10);
 
       expect(() => {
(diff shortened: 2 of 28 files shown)


packages/realm/src/Counter.ts
elle-j, line 1: Adding a bit more docs onto this `Counter` class as we don't have it yet for the MDB docs. Please see if you think something should be reformulated/added/removed, etc.

packages/realm/src/Counter.ts
```diff
@@ -0,0 +1,140 @@
...
+export class Counter {
+  /** @internal */
+  private readonly [REALM]: Realm;
+  /** @internal */
+  private readonly [OBJ]: binding.Obj;
+  /** @internal */
+  private readonly [COLUMN_KEY]: binding.ColKey;
+
+  /** @internal */
+  constructor(realm: Realm, obj: binding.Obj, columnKey: binding.ColKey) {
```
elle-j, line 78: P.S. Despite `@internal`, the constructors still show up in the API ref docs (for other classes as well).

packages/realm/src/Object.ts
```diff
@@ -43,25 +43,27 @@ import {
...
-   * performance but also have implications for how changes across devices are
-   * merged. For most use cases, the behavior will match the intuitive behavior
-   * of how changes should be merged, but if updating an entire object is
-   * considered an atomic operation, this mode should not be used.
+   * If an existing object exists (determined by a matching primary key), only
+   * properties where the value has actually changed will be updated. This improves
+   * notifications and server side performance but also have implications for how
+   * changes across devices are merged. For most use cases, the behavior will match
+   * the intuitive behavior of how changes should be merged, but if updating an
+   * entire object is considered an atomic operation, this mode should not be used.
```
elle-j, line 61: The only change here is adding `(determined by a matching primary key)`.

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+      it("can create and access (input: number)", function (this: RealmContext) {
+        for (let i = 0; i < initialValuesList.length; i++) {
+          const input = initialValuesList[i];
+          const { counter } = this.realm.write(() => {
+            return this.realm.create<IWithCounter>(WithCounterSchema.name, {
+              counter: input,
+            });
+          });
+
+          const expectedNumObjects = i + 1;
```
kneth, line 124:
```suggestion
          const expectedNumObjects = initialValuesList.length;
```

packages/realm/src/Unmanaged.ts
elle-j, line 1: cred @nirinchev

packages/realm/src/Unmanaged.ts
```diff
@@ -16,23 +16,29 @@
...
+import type { AnyRealmObject, Collection, Counter, Dictionary, List, Realm, RealmSet } from "./internal";
 
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any -- We define these once to avoid using "any" through the code */
 export type AnyCollection = Collection<any, any, any, any, any>;
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any -- We define these once to avoid using "any" through the code */
 export type AnyDictionary = Dictionary<any>;
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any -- We define these once to avoid using "any" through the code */
 export type AnyList = List<any>;
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any -- We define these once to avoid using "any" through the code */
+export type AnySet = RealmSet<any>;
```
elle-j, line 28: I'm also adding an unmanaged set (as an array), which I thought we had already done.

packages/realm/src/tests/schema-normalization.test.ts
elle-j, line 1: I've grouped certain tests into suites in this file, so not all changes here are actually relevant changes (more indentation changes). I'll mark the relevant changes with a PR comment like this.

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
+          optional: true,
+        });
+      });
 
-    itNormalizes("Person<>", {
-      type: "set",
-      objectType: "Person",
-      optional: false,
-    });
+      describe("'counter'", () => {
```
elle-j, line 128: Relevant

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
-        indexed: true,
-      },
-      {
-        type: "string",
-        indexed: true,
-        optional: false,
-      },
-      { isPrimaryKey: false },
-    );
+      itThrowsWhenNormalizing("counter[]", "Counters cannot be used in collections");
```
elle-j, line 272: Relevant: This and the next 5 single-line additions.

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
+      });
 
-    itThrowsWhenNormalizing(
-      {
-        // @ts-expect-error Passing in the wrong type
-        type: "int?[]",
-      },
-      "Cannot use shorthand '[]' and '?' in 'type' or 'objectType' when defining property objects",
-    );
+      describe("'counter'", () => {
```
elle-j, line 534: Relevant

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
+
+        itThrowsWhenNormalizing(
+          {
+            type: "list",
+            objectType: "int?",
+          },
+          "Cannot use shorthand '?' in 'type' or 'objectType' when defining property objects",
+        );
+
+        itThrowsWhenNormalizing(
```
elle-j, line 1018: Relevant

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
+          {
+            // @ts-expect-error Passing in the wrong type.
+            type: "counter",
+          },
+          "Cannot use shorthand 'counter' in 'type' or 'objectType' when defining property objects. To use presentation types such as 'counter', use the field 'presentation'",
+        );
+
+        itThrowsWhenNormalizing(
```
elle-j, line 1026: Relevant

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
-    itThrowsWhenNormalizing(
-      {
-        type: "string",
-        indexed: false,
-        optional: false,
-      },
-      "Primary keys must always be indexed.",
-      { isPrimaryKey: true },
-    );
+      describe("'counter' & collection combinations", () => {
```
elle-j, line 1035: Relevant

integration-tests/tests/src/tests/counter.ts
kneth: I know that the variable `i` is declared in the scope outside the loop (and each iteration has its own scope), but I always find it a bit hard to read.

packages/realm/src/tests/schema-normalization.test.ts
```diff
@@ -27,982 +27,1130 @@ const OBJECT_NAME = "MyObject";
...
+        itThrowsWhenNormalizing(
+          {
+            type: "string",
+            indexed: "full-text",
+          },
+          "Primary keys cannot be full-text indexed.",
+          { isPrimaryKey: true },
+        );
+
+        itThrowsWhenNormalizing(
```
elle-j, line 1144: Relevant

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+};
+
+interface IWithDefaultCounter {
+  counterWithDefault: Counter;
+}
+
+const WithDefaultCounterSchema: ObjectSchema = {
+  name: "WithDefaultCounter",
+  properties: {
+    counterWithDefault: { type: "int", presentation: "counter", default: 0 },
```
kraenhansen, line 60: Would this be more interesting if the `default` was different than `0`?

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+      WithDefaultCounterSchema,
+      WithRegularIntSchema,
+      WithMixedSchema,
+    ],
+  });
+
+  const initialValuesList = [-100, 0, 1.0, 1000] as const;
+
+  describe("Create and access", () => {
+    describe("Via 'realm.create()'", () => {
```
kraenhansen, line 114:
```suggestion
  describe("create and access", () => {
    describe("via 'realm.create()'", () => {
```

kneth (commented):
I understand the the introduction of the "presentation" concept has added to the complexity of implementing Counters. The concept is interesting, and I am looking forward on how we can use it in the future.

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+          for (const input of initialNumValues) {
+            const { counter } = this.realm.create<IWithCounter>(WithCounterSchema.name, {
+              counter: input,
+            });
+            expectCounter(counter);
+            expect(counter.value).equals(input);
+
+            initialCounterValues.push(counter);
+          }
+        });
```
kraenhansen, line 146: A more functional alternative - not feeling strong for this, just a slightly more concise way of expressing this that doesn't require explicit type declaration of `initialCounterValues` nor array manipulation via `push`.

```suggestion
        // First create Realm objects with counters.
        const initialCounterValues = this.realm.write(() => {
          return initialNumValues.map(input => {
            const { counter } = this.realm.create<IWithCounter>(WithCounterSchema.name, {
              counter: input,
            });
            expectCounter(counter);
            expect(counter.value).equals(input);
            return counter;
            });
          }
        });
```

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+  name: "WithMixed",
+  properties: {
+    mixed: "mixed",
+    list: "mixed[]",
+    dictionary: "mixed{}",
+    set: "mixed<>",
+  },
+};
+
+function expectCounter(value: unknown): asserts value is Counter {
```
kraenhansen, line 92: 💪

integration-tests/tests/src/tests/counter.ts
elle-j: Yes it would! 🙂

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+  expect(value).to.be.instanceOf(Counter);
+}
+
+function expectKeys(dictionary: Realm.Dictionary, keys: string[]) {
+  expect(Object.keys(dictionary)).members(keys);
+}
```
kraenhansen, line 98: I wonder if usage of `expectKeys` could be replaced with `expect(dictionary).to.have.all.keys(keys)` as per https://www.chaijs.com/api/bdd/#method_keys?

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+
+      it("can create and access (input: default value)", function (this: RealmContext) {
+        const { counterWithDefault } = this.realm.write(() => {
+          // Pass an empty object in order to use the default value from the schema.
+          return this.realm.create<IWithDefaultCounter>(WithDefaultCounterSchema.name, {});
+        });
+
+        expect(this.realm.objects(WithDefaultCounterSchema.name).length).equals(1);
+        expectCounter(counterWithDefault);
+        expect(counterWithDefault.value).equals(0);
```
kraenhansen, line 172: As per my earlier comment, I would expect the default to be `0` even if not provided explicitly through the schema.

integration-tests/tests/src/tests/counter.ts
kraenhansen: If the title is not describing an entity (like the `Counter` class name), yes. One idea about the BDD nomenclature is that it becomes a whole sentence when concatenated, making it easier to read in my opinion.

packages/realm/src/assert.ts
```diff
@@ -64,6 +64,10 @@ assert.number = (value: unknown, target?: string): asserts value is number => {
   assert(typeof value === "number", () => new TypeAssertionError("a number", value, target));
 };
 
+assert.integer = (value: unknown, target?: string): asserts value is number => {
+  assert(Number.isInteger(value), () => new TypeAssertionError("an integer", value, target));
+};
```
kraenhansen, line 69: Great addition!

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+            );
+          });
+          expectCounter(object.nullableCounter);
+          expect(object.nullableCounter.value).equals(-100_000);
+        });
+      }
+    });
+  });
+
+  describe("Filtering", () => {
```
kraenhansen, line 423: While I understand why this goes here, it's also interesting in terms of how we structure our tests going forward as it could also have been an addendum to https://github.com/realm/realm-js/blob/main/integration-tests/tests/src/tests/queries.ts

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+      expect(counter.value).equals(10);
+
+      expect(() => {
+        this.realm.write(() => {
+          // @ts-expect-error Testing incorrect type.
+          counter.set(BigInt(1));
+        });
+      }).to.throw("Expected 'value' to be an integer, got a bigint");
+      expect(counter.value).equals(10);
+    });
```
kraenhansen, line 616: I'm wondering if this could benefit from a loop, to ensure consistency if a case is added in the future:

```
for (const operation of ["increment", "decrement", "set"]) {
```

integration-tests/tests/src/tests/counter.ts
elle-j: `to.have.all.keys` does not seem to work for custom classes. But `members` will check that the provided keys all exist (so it will also fail if more than the provided keys exist).

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+        return { counter, objectWithMixed };
+      });
+      expectCounter(counter);
+      expect(counter.value).equals(10);
+
+      expect(() => {
+        this.realm.write(() => {
+          objectWithMixed.mixed = counter;
+        });
+      }).to.throw("Using a Counter as a Mixed value is not supported");
```
kraenhansen, line 850: Perhaps we should capture an issue to track lifting this restriction in the future?

integration-tests/tests/src/tests/counter.ts
```diff
@@ -0,0 +1,994 @@
...
+        });
+      });
+      expectCounter(counter);
+
+      const objects = this.realm.objects(WithCounterSchema.name);
+      expect(objects.length).equals(1);
+
+      expect(() => {
+        objects.filtered("counter = $0", counter);
+      }).to.throw("Using a Counter as a query argument is not supported. Use 'Counter.value'");
```
kraenhansen, line 991: It might be interesting to eventually lift this restriction too.

packages/realm/src/errors.ts
```diff
@@ -45,6 +45,14 @@ export class TypeAssertionError extends AssertionError {
...
     } else if (typeof value === "function") {
       return `a function or class named ${value.name}`;
+    } else if (typeof value === "number") {
+      let type = "a number";
+      if (Number.isNaN(value)) {
+        type = "NaN";
+      } else if (!Number.isInteger(value)) {
+        type = "a floating point number";
+      }
+      return type;
```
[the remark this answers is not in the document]
elle-j: Thanks for catching that, will update.

integration-tests/tests/src/tests/counter.ts
elle-j: Yeah I think for the counters tests it still made sense to keep most things as it's easier to remember what's allowed in terms of counters. But agree that it would be valuable for us to decide on a clear structure of our tests, where it's easy to quickly reason about where a new test should go are where a current test is located.

integration-tests/tests/src/tests/counter.ts
elle-j: I think it's more a server-side restriction at the moment, with not being able to merge it correctly. @nirinchev did you know the specific restriction here?

integration-tests/tests/src/tests/counter.ts
elle-j: Yup I agree. Went with our earlier agreement to push this to a future counter version.

integration-tests/tests/src/tests/counter.ts
kraenhansen: Should this go into `integration-tests/tests/src/utils/expects.ts` as well?

elle-j: Sounds good 👍

packages/realm/src/assert.ts
elle-j: Ty 👍

integration-tests/tests/src/tests/counter.ts
elle-j: Can take a look at what that refactor would look like 👍

elle-j:
RC release train will take off now 🚋

CHANGELOG.md
```diff
@@ -6,6 +6,53 @@
...
+
+### Internal
+<!-- * Either mention core version or upgrade -->
+<!-- * Using Realm Core vX.Y.Z -->
+<!-- * Upgraded Realm Core from vX.Y.Z to vA.B.C -->
+
+## 12.10.0-rc.0 (2024-05-31)
+
+### Enhancements
+* A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. ([#6694](https://github.com/realm/realm-js/pull/6694))
```
kneth, line 26:
```suggestion
* A `counter` presentation data type has been introduced. The `int` data type can now be used as a logical counter for performing numeric updates that need to be synchronized as sequentially consistent events rather than individual reassignments of the number. For a high level introduction, read about [counters in the documentation](https://www.mongodb.com/docs/atlas/app-services/sync/details/conflict-resolution/#counters). ([#6694](https://github.com/realm/realm-js/pull/6694))
```

kneth (approved):
Great work. As we have discussed the PR before the RC, I don't have any further comments (except a suggestion to link to the documentation).

Let's get it merged and released!

packages/realm/src/Counter.ts
```diff
@@ -0,0 +1,149 @@
...
+ * to be synchronized as sequentially consistent events rather than individual
+ * reassignments of the number.
+ *
+ * For instance, offline Client 1 and Client 2 which both see `Counter.value`
+ * as `0`, can both call `Counter.increment(1)`. Once online, the value will
+ * converge to `2`.
+ * @note
+ * __Not supported__:
+ *
+ * Counter types are __not__ supported as:
```
kraenhansen, line 36: My eyes stopped at this, since I haven't seen use use this markdown syntax elsewhere.

I wonder if it's possible / better to use an h2 here instead?

```suggestion
 * ## Not supported
 *
 * Counter types are __not__ supported as:
```

Or since this the header is sort of repeated with the pretext to the list of items, it would be equally clear (and less verbose) to just leave out the heading entirely.

```suggestion
 * Counter types are __not__ supported as:
```

CHANGELOG.md
elle-j: Thanks, good link! (I update the suggestion slightly to add another link as well) 👍

packages/realm/src/Counter.ts
kraenhansen: This is a bit weird 🤔 Perhaps there's something misconfigured in the way we generate docs then.

packages/realm/src/PropertyHelpers.ts
```diff
@@ -121,6 +125,45 @@ function embeddedSet({ typeHelpers: { toBinding }, columnKey }: PropertyOptions)
...
+            if (isUninitialized || resettingToNull) {
+              defaultSet(options)(obj, value);
+              return;
+            }
+          }
+          throw new Error(
+            "You can only reset a Counter instance when initializing a previously " +
+              "null Counter or resetting a nullable Counter to null. To update the " +
+              "value of the Counter, use its instance methods.",
+          );
```
kraenhansen, line 157: Not blocking, but I'm not a super-fan of duplicating `defaultSet(options)(obj, value)` here - is there a way to isolate the logic to determine if a set is allowed and have calling `defaultSet` depend on just that?

packages/realm/src/Counter.ts
elle-j: It's a bold syntax, but yeah a heading may be more suitable 👍 Tested with h3 which renders a bit better than h2 I think 🙃

packages/realm/src/PropertyHelpers.ts
elle-j: Do you mean calling `defaultSet()` only in one place here? One thing I want to avoid is unnecessarily calling into Core (`obj.getAny()`).

packages/realm/src/Counter.ts
elle-j: ✅

integration-tests/tests/src/tests/counter.ts
elle-j: Good suggestion Kraen, I've updated it.

packages/realm/src/Counter.ts
```diff
@@ -0,0 +1,148 @@
...
+ * ### Nullability
+ *
+ * The above property schema can be extended to allow a nullable counter.
+ * A `Counter` never stores `null` values itself, but the counter property
+ * on the {@link Realm.Object} (e.g. `myRealmObject.myCounter`) can be `null`.
+ *
+ * ### Not a live object
+ *
+ * The `Counter` instance itself is *not* a live object. Calling `myCounter.value`
+ * will always return the most recent locally persisted (and synced) state, and
```
kraenhansen, line 73: > the most recent locally persisted

That makes it live, doesn't it?

packages/realm/src/PropertyHelpers.ts
kraenhansen: > calling `defaultSet()` only in one place here?

Right.

> avoid is unnecessarily calling into Core

That makes sense, but as long as the call into core is on the right-hand-side of a logical operator, it knows not to evaluate it, if it's not needed. Notice how this doesn't throw:

```js
> function foo() { throw new Error("!"); }
undefined
> const isAllowed = 2 > 1 || foo();
undefined
> isAllowed
true
```

Personally I would probably either make a local `const isAllowed` variable or define a separate `function isAllowed(obj, isCreating)` if that expression becomes too complex.

packages/realm/src/Realm.ts
```diff
@@ -651,8 +662,11 @@ export class Realm {
       if (extras) {
         objectSchema.ctor = extras.constructor;
       }
+      // Some tests require that the properties exist on the object, even
+      // if the value is `undefined`. Thus, explicitly set to `undefined`.
```
kraenhansen, line 666: This comment seems more appropriate for the PR. I do think this is by design - the canonical schema is verbose and therefore includes keys even if the value of those are undefined. We could potentially lift that restriction 🤔

packages/realm/src/Unmanaged.ts
```diff
@@ -16,23 +16,29 @@
...
 
 type ExtractPropertyNamesOfType<T, PropType> = {
   [K in keyof T]: T[K] extends PropType ? K : never;
 }[keyof T];
 
+type ExtractPropertyNamesOfTypeExclNullability<T, PropType> = {
```
kraenhansen, line 34:
```suggestion
type ExtractPropertyNamesOfTypeExcludingNullability<T, PropType> = {
```

packages/realm/bindgen/vendor/realm-core
kraenhansen, line 1: Is this update still needed?

kraenhansen (approved):
This turned out not as simple as we thought on the surface of it. Great job @elle-j!

packages/realm/src/PropertyHelpers.ts
elle-j: Gotcha, yeah with the Core part I was referring more to keeping some of the local variables for readability that sometimes require evaluating it beforehand. The logic could still be expressed in one variable:

```ts
set(obj, value, isCreating) {
  // We only allow resetting a counter this way (e.g. realmObject.counter = 5)
  // when it is first created, or when a nullable/optional counter was
  // previously `null`, or when resetting a nullable counter to `null`.
  const isAllowed =
    isCreating || (optional && (value === null || value === undefined || obj.getAny(columnKey) === null));

  if (isAllowed) {
    defaultSet(options)(obj, value);
  } else {
    throw new Error(
      "You can only reset a Counter instance when initializing a previously " +
        "null Counter or resetting a nullable Counter to null. To update the " +
        "value of the Counter, use its instance methods.",
    );
  }
},
```

packages/realm/src/Counter.ts
```diff
@@ -0,0 +1,148 @@
...
+ * ### Creating a counter
+ *
+ * Use a `number` when creating your counter on a {@link Realm.Object}.
+ * ```ts
+ * realm.write(() => {
+ *   realm.create(MyObject, { _id: "123", counter: 0 });
+ * });
+ * ```
+ *
+ * To update a `null` counter, use {@link UpdateMode.Modified} or {@link UpdateMode.All}.
```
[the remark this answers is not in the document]
elle-j: I've [clarified](https://github.com/realm/realm-js/pull/6694/commits/9f1693f3451d7ae9bf551af64c9c04304a5b8e8c) it now and also moved it under "Nullability" 🙂

packages/realm/src/Counter.ts
elle-j: The distinction I made here between the Counter itself and its value may not be correct in terms of "live", thanks for pointing it out 🙂

packages/realm/src/Realm.ts
elle-j: Hmm, I don't at this point see an issue with having the fields set to `undefined`. It may be helpful to see all fields in the canonical version? But if we decide to change this at some point, we should consider the behavioral change for people using the `in` operator as `"key1" in mySchema` will go from `true` to `false`.

(I'll remove the comment tho) 👍

packages/realm/src/Unmanaged.ts
```diff
@@ -45,6 +51,20 @@ type RealmDictionaryRemappedModelPart<T> = {
...
+ */
+type RealmSetRemappedModelPart<T> = {
+  [K in ExtractPropertyNamesOfType<T, AnySet>]?: T[K] extends RealmSet<infer GT> ? Array<GT | Unmanaged<GT>> : never;
+};
+
+/**
+ * Exchanges properties defined as a {@link Counter} with a `number`.
+ */
+type RealmCounterRemappedModelPart<T> = {
+  [K in ExtractPropertyNamesOfTypeExclNullability<T, Counter>]?: Counter | number | Exclude<T[K], Counter>;
```
elle-j, line 65:
```suggestion
  [K in ExtractPropertyNamesOfTypeExcludingNullability<T, Counter>]?: Counter | number | Exclude<T[K], Counter>;
```

packages/realm/src/Unmanaged.ts
```diff
@@ -53,6 +73,7 @@ export type OmittedRealmTypes<T> = Omit<
   | ExtractPropertyNamesOfType<T, Function> // TODO: Figure out the use-case for this
   | ExtractPropertyNamesOfType<T, AnyCollection>
   | ExtractPropertyNamesOfType<T, AnyDictionary>
+  | ExtractPropertyNamesOfTypeExclNullability<T, Counter>
```
elle-j, line 76:
```suggestion
  | ExtractPropertyNamesOfTypeExcludingNullability<T, Counter>
```

packages/realm/bindgen/vendor/realm-core
elle-j: This branch is using Core 14.10.0 (now in main), and *additionally* the necessary spec change.

PR #6704, opened 2024-06-04
Sanity checking BaaS

kraenhansen:
## What, How & Why?

I believe we're seeing failures on https://github.com/realm/realm-js/actions/runs/9356672750/job/25755122839?pr=6650 which are unrelated to the changes made in the PR. This is an attempt to verify that assumption.


kraenhansen:
Closing since the change to BaaS got reverted 👍

PR #6712, opened 2024-06-10
Enables Ccache for our RNTA on iOS

kraenhansen:
## What, How & Why?

This uses https://github.com/facebook/react-native/pull/42051 to enabled the use of Ccache when running the React Native Test App tests on iOS.


gagik (approved):

integration-tests/environments/react-native-test-app/ios/Podfile
```diff
@@ -2,6 +2,8 @@ require_relative '../../../../node_modules/react-native-test-app/test_app'
 
 workspace 'RealmTests.xcworkspace'
 
+ENV["USE_CCACHE"] = Pod::Executable::which('ccache') ? "1" : "0"
```
kraenhansen, line 5: This is using [Pod::Executable::which](https://rubydoc.info/gems/cocoapods/Pod/Executable#which-class_method):
> Returns the absolute path to the binary with the given name on the current ‘PATH`, or `nil` if none is found.

elle-j, line 5: For consistency 👍 
```suggestion
ENV['USE_CCACHE'] = Pod::Executable::which('ccache') ? '1' : '0'
```

elle-j (approved):

kneth (approved):

integration-tests/environments/react-native-test-app/ios/Podfile
kraenhansen: I like consistency 💙 Thanks

integration-tests/environments/react-native-test-app/ios/Podfile
```diff
@@ -2,6 +2,8 @@ require_relative '../../../../node_modules/react-native-test-app/test_app'
 
 workspace 'RealmTests.xcworkspace'
 
+ENV["USE_CCACHE"] = ENV["USE_CCACHE"] || Pod::Executable::which('ccache') ? "1" : "0"
```
elle-j, line 5: Think this got reverted 😛 
```suggestion
ENV['USE_CCACHE'] = ENV['USE_CCACHE'] || Pod::Executable::which('ccache') ? '1' : '0'
```

kraenhansen:
For some reason the "Post ccache" step doesn't seem to report ccache being used 🤔
But, I'm merging this for now since I've verified this works locally and want to depend on it from #6650.

Issue #6283, opened 2023-11-27
`@realm/react`: Passing a `realm` as prop to `RealmProvider`


kraenhansen:
We have users that have complex code opening the Realm while they would like to use the default context exported.

We want to make their use-case easier, by providing an overload to the `RealmProvider` to take a single `realm` instance as prop instead of taking configuration parameters.

In this case, we could have the `closeOnUnmount` default to `false`.

kraenhansen:
@bimusiek I've created this to track better support for your use-case.

Please add any details you might have to help us solve this in a way that makes most sense for you and your team.

kraenhansen:
Another use-case for this, is the use of `react-native-navigation` as mentioned in https://github.com/realm/realm-js/issues/6233#issue-1977300237 where screens are mounted in separate component trees. This also discuss a similar need of having `app` passed to the `AppProvider`.

PR #6714, opened 2024-06-10
RJS-2648: Add ability for Realm instance to be used in RealmProvider

gagik:
## What, How & Why?

@realm/devdocs 

### What, How
1. The default `RealmProvider` (or any empty provider created from `createRealmContext()`) is the most generic provider which has mutually exclusive PropTypes that accepts either an existing realm **_or_** Realm configuration fields as props.
```jsx
import { RealmProvider } from `@realm/react`;
...
return <>
  {/* Valid */}
  <RealmProvider realm={new Realm(...)}> ... </RealmProvider>
  {/* Valid */}
  <RealmProvider schema={[catSchema]} path="test.realm" realmRef={ref}> ... </RealmProvider>
  {/* Invalid, would raise a type exception */}
  <RealmProvider realm={new Realm()} schema={[catSchema]}> ... </RealmProvider>
</>
```
2. Provider created from using an existing Realm instance, i.e. `createRealmContext(realm)` is a simple wrapper with no props, just children.  **NOTE: This provider is not necessary, you can use the hooks without it as long as you import from the correct context!**
```jsx
const { RealmProvider, useRealm } = createRealmContext(realm);
...
<>
   {/* This provider is valid but not needed */}
  <RealmProvider> ... </RealmProvider>
  <AnotherComponent>
  {/*  When using createRealmContext(realm) with an existing Realm,
       you can use the hooks outside of the scope of the provider! */}
  </AnotherComponent>
</>
```
3. Provider created from a configuration, i.e. `createRealmContext({schema: [...])` allows you to use all the Realm configuration settings as well as the existing special `@realm/react` props.
```jsx
const { RealmProvider } = createRealmContext({schema: [catSchema]});
...
<RealmProvider closeOnUnmount={false} realmRef={ref} path="thing.realm" schema={...}> ... </RealmProvider>
```

#### Nesting
If two default providers are nested, the deepest provider will take precedence, i.e.
```jsx
<RealmProvider realm={realmA}>
     <RealmProvider realm={realmB}>
            {/* Running useRealm() here will return realmB */}
      </RealmProvider>
</RealmProvider>
```
Same applies for configuration-based providers.

### Why?
Unlocks new usecases and abilities for the `RealmProvider`, where one might want to use their own managed Realm in the context.

This closes https://github.com/realm/realm-js/issues/6283

## ☑️ ToDos

* [X] 📝 Changelog entry
* [X] 🚦 Tests


CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
 ### Deprecations
 * None
 
+### Enhancements
+* Added the ability to use an existing Realm instance in `RealmProvider` and `createRealmContext`. ([#6714](https://github.com/realm/realm-js/pull/6714))
+```jsx
+// Using RealmProvider
```
gagik, line 10: I thought it'd be nice to include an example in the changelog and I noticed it has been done before but maybe it should go somewhere else.

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -161,6 +191,45 @@ export function createRealmProvider(
...
 
+/**
+ * Generates a flexible `RealmProvider` which is either based on configuration or based on a realm
+ * depending on its props.
+ * @param RealmContext - The context that will contain the Realm instance
+ * @returns a RealmProvider component that provides context to all context hooks
+ */
+export function createFlexibleRealmProvider(RealmContext: React.Context<Realm | null>): FlexibleRealmProviderFC {
+  return ({ realm, ...restProps }) => {
+    return useMemo(() => {
```
gagik, line 202: I think `useMemo` is appropriate here but may be worth discussing if there's better alternatives

packages/realm-react/src/RealmContext.ts
```diff
@@ -0,0 +1,109 @@
...
+   *
+   *   return (
+   *     <RealmProvider schema={[Task, User]} path={"data.realm"} sync={syncConfig}>
+   *       <App/>
+   *     </RealmProvider>
+   *   )
+   * }
+   * ```
+   * @param props - The {@link Realm.Configuration} for this Realm defaults to
+   * the config passed to `createRealmProvider`, but individual config keys can
```
gagik, line 44: The original text said passed to `createRealmProvider` but this is a private method isn't it? `createRealmContext` seems more appropriate.

packages/realm-react/src/__tests__/RealmProvider.test.tsx
```diff
@@ -43,317 +48,556 @@ const catSchema: Realm.ObjectSchema = {
...
+
+        await waitFor(() => queryByTestId("testContainer"));
+
+        expect(queryByTestId("fallbackContainer")).toBeNull();
+        expect(queryByTestId("testContainer")).not.toBeNull();
+      });
+    });
+  });
+
+  describe("with an existing Realm instance", () => {
```
gagik, line 280: tried to cover a good ground with testing but let me know if there's anything that may be worthwhile to test more.

gagik (commented):
Some things worth discussing.

CHANGELOG.md
kneth: Definitely a good idea. It gives our users a taste of the new feature, and they can dive deeper into the documentation

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,75 @@ type ProviderProps = PartialRealmConfiguration & {
...
+export type RealmProviderWithConfigurationProps = RestrictivePick<
+  RealmProviderProps,
+  "closeOnUnmount" | "realmRef" | "fallback" | "children" | keyof PartialRealmConfiguration
+>;
+
+/**
+ * Represents the provider returned from creating context with no arguments (including the default context).
+ * Supports either passing a `realm` as a property or the schema configuration.
+ */
+export type FlexibleRealmProviderFC = React.FC<
```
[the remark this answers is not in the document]
kneth: We use "flexible" for other concepts (flexible sync, and upcoming flexible schema). Maybe "Relaxed" or "Generalized"??

elle-j (commented):
Nice to see this feature coming alive 🎉  Leaving my comments from a first pass. (Haven't gone through all of it yet.)

packages/realm-react/src/RealmProvider.tsx
gagik: Hm, good point, I was under impression that tests were handling this case but now that I think about it none of them seem to be doing exactly this change

packages/realm-react/src/index.tsx
```diff
@@ -135,11 +54,51 @@ type RealmContext = {
...
+ *    properties: {
+ *      ...
+ *    },
+ *  };
+ * }
+ * const {useRealm, useQuery, useObject, RealmProvider} = createRealmContext();
+ * ...
+ * <RealmProvider schema={[Task]}></RealmProvider>
+ * ```
+ *  @example
```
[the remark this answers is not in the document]
gagik: Good catch, I feel like it'd be nice to have better linting for spaces and things in the codebase. autoformat would help a lot

packages/realm-react/src/index.tsx
```diff
@@ -135,11 +54,51 @@ type RealmContext = {
...
+ * ```
+ * const realm = await Realm.open({ path: "example.realm" });
```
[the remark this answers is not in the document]
gagik: I agree with the other place since we had the schema already defined and it's nice to show equivalence to the config-based provider but I feel like for purpose of this manually managed realm-centric bit it's not really necessary to mention a schema.

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,73 @@ type ProviderProps = PartialRealmConfiguration & {
...
+  "closeOnUnmount" | "realmRef" | "fallback" | "children" | keyof PartialRealmConfiguration
+>;
+
+/**
+ * Represents the provider returned from creating context with no arguments (including the default context).
+ * Supports either passing a `realm` as a property or the schema configuration.
+ */
+export type GeneralizedRealmProviderFC = React.FC<
+  RealmProviderWithRealmInstanceProps | RealmProviderWithConfigurationProps
+>;
```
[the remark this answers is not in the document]
gagik: This is definetely simpler, I like the new names but with some there's some complications:
1.  It's a very subtle distinction but `RealmProviderFromRealmProps != RealmProviderWithRealmProps` .

If the provider was created from `createRealmContext(realm)` then there's no `realm` to pass as props again (unless we want to support overriding it after creation but I think it'll lead to people passing it twice a lot, it'd make sense for them to use the DynamicRealmProvider then).
This is not the case for `RealmProviderFromConfigurationProps` since you can basically override all the options you pass to `createRealmContext`, so `RealmProviderFromConfigurationProps === RealmProviderWithConfigurationProps`

I also personally like the existence of a singular `RealmProviderProps` and doing the picking from there because it gives us a one point of entry for all the props a user sees when they get a `DynamicRealmProvider` and then we can create restrictive subsets for different scenarios.

```ts
export type DynamicRealmProvider = React.FC<RealmProviderFromConfigurationProps | RealmProviderFromRealmProps>;
```


2. This `DynamicRealmProvider` will allow you to set both `realm` **and** configuration props at the same time. We can throw an error in this case ourselves on runtime but I think it's nicer UX to enforce it through types (unless it can seem like maintenece cost is too high... I think it's worth it). I don't think TS it smart enough to understand the union and separate it when enforcing props, which is why I use `RestrictedPick`, it allows for mutually exclusive props to be used as a union.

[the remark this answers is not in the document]
elle-j: Yeah I think if we could prevent having to think too much about `From`/`With` distinction in the types that'd be good as well 🙂 So I guess there are 4 allowed combinations for props then:
* With configuration
* With realm
* Just empty (incl. children)
* With configuration *or* realm (dynamic)

E.g.:
```ts
type RealmProviderWithConfigurationProps = {
  closeOnUnmount?: boolean;
  realmRef?: React.MutableRefObject<Realm | null>;
  fallback?: React.ComponentType<unknown> | React.ReactElement | null | undefined;
  children: React.ReactNode;
} & PartialRealmConfiguration;

type RealmProviderWithRealmProps = {
  realm: Realm; // Note no `?`
  children: React.ReactNode;
};

type RealmProviderWithEmptyProps = {
  children: React.ReactNode;
};

type DynamicRealmProviderProps = { /* TODO (further below) */ };
```

This gives us 4 provider types (avoiding to mix `From` and `With` in these types used):

```ts
export type RealmProviderWithConfiguration = React.FC<RealmProviderWithConfigurationProps>;
export type RealmProviderWithRealm = React.FC<RealmProviderWithRealmProps>;
export type RealmProviderWithEmpty = React.FC<RealmProviderWithEmptyProps>;
export type DynamicRealmProvider = React.FC<DynamicRealmProviderProps>;

export type RealmProvider = RealmProviderWithConfiguration | RealmProviderWithEmpty | DynamicRealmProvider;
```

To address the `DynamicRealmProvider`, as you pointed we should have a type that:
* Allows all keys of `RealmProviderWithRealmProps`, while disallowing (`never | undefined`) the excess or non-shared keys of `RealmProviderWithConfigurationProps`.
* Or, allows all keys of `RealmProviderWithConfigurationProps`, while disallowing the excess or non-shared keys of `RealmProviderWithRealmProps`.

In order to not have to maintain adding/modifying individual keys in the types I believe this should work:

```ts
/**
 * A type with all properties of `T`, as well as any excess
 * properties on `Excess` set to `never | undefined`.
 */
type WithExcessAsNever<T, Excess> = T & { [K in keyof Omit<Excess, keyof T>]?: never };

type DynamicRealmProviderProps =
  | WithExcessAsNever<RealmProviderWithRealmProps, RealmProviderWithConfigurationProps>
  | WithExcessAsNever<RealmProviderWithConfigurationProps, RealmProviderWithRealmProps>;
```

Perhaps ☝️ could even be simplified further.

The `create` functions are similar to before (haven't renamed the `From` in these names as its *from* the argument passed):
```ts
export function createRealmProviderFromRealm(
  // Note: Was previously `Realm | null` but seems like the callers always pass a realm here.
  realm: Realm,
  ...
): RealmProviderWithEmpty { ... }
export function createRealmProviderFromConfiguration(...): RealmProviderWithConfiguration { ... }
export function createDynamicRealmProvider(): DynamicRealmProvider { ... }
export function createRealmProvider(...): RealmProvider { ... }
```

packages/realm-react/src/index.tsx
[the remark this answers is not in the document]
elle-j: Yeah I agree, I believe we gave it a go at some point but faced some issues due to what we need to allow for our headers in each file (starting with `///`).

packages/realm-react/src/RealmContext.ts
```diff
@@ -0,0 +1,109 @@
...
+   * const realm = useRealm();
+   * ```
+   * @returns a realm instance
+   */
+  useRealm: ReturnType<typeof createUseRealm>;
+
+  /**
+   * Returns a {@link Realm.Collection} of {@link Realm.Object}s from a given type.
+   * The hook will update on any changes to any object in the collection
+   * and return an empty array if the collection is empty.
```
kraenhansen, line 64: > an empty array if the collection is empty

Is that true? I was in the understanding that it just returns the empty collection?

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+   */
+  fallback?: React.ComponentType<unknown> | React.ReactElement | null | undefined;
   children: React.ReactNode;
-};
+} & PartialRealmConfiguration;
+
+/**
+  Represents the provider returned from using an existing realm at context creation i.e. `createRealmContext(new Realm))`.
+ */
+export type RealmProviderFromRealmFC = React.FC<Pick<RealmProviderProps, "children">>;
```
kraenhansen, line 54: To me `FC` here feels a bit like an implementation detail.
```suggestion
export type RealmProviderFromRealm = React.FC<Pick<RealmProviderProps, "children">>;
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+export type RealmProviderFromConfiguration = React.FC<
+  Pick<RealmProviderProps, "closeOnUnmount" | "realmRef" | "fallback" | "children" | keyof PartialRealmConfiguration>
+>;
+
+/**
+ * Explicitly sets the unpicked properties of a type to never instead of dropping them like in Pick.
+ * Useful for ensuring different prop types are mutually exclusive as React expects the union type
+ * of different prop types to include all the fields.
+ */
+type RestrictivePick<T, K extends keyof T> = Pick<T, K> & { [RestrictedKey in keyof Omit<T, K>]?: never };
```
kraenhansen, line 68: Perhaps this makes more sense to put in https://github.com/realm/realm-js/blob/main/packages/realm-react/src/helpers.ts or perhaps https://github.com/realm/realm-js/blob/main/packages/realm-react/src/types.ts 🤔

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+/**
+  Represents the provider returned from using an existing realm at context creation i.e. `createRealmContext(new Realm))`.
+ */
+export type RealmProviderFromRealmFC = React.FC<Pick<RealmProviderProps, "children">>;
+
+/**
+ * Represents the provider returned from using a Realm configuration at context creation i.e. `createRealmContext({schema: []}))`.
+ */
+export type RealmProviderFromConfiguration = React.FC<
+  Pick<RealmProviderProps, "closeOnUnmount" | "realmRef" | "fallback" | "children" | keyof PartialRealmConfiguration>
```
kraenhansen, line 60: Would this be equivalent?

```suggestion
  Omit<RealmProviderProps, "realm">
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+ * props should be disallowed.
+ */
+export type RealmProviderWithRealmProps = RestrictivePick<RealmProviderProps, "realm" | "children">;
+
+/*
+ * Represents properties of a generic RealmProvider where Realm configuration props are used and no realm is set
+ */
+export type RealmProviderWithConfigurationProps = RestrictivePick<
+  RealmProviderProps,
+  "closeOnUnmount" | "realmRef" | "fallback" | "children" | keyof PartialRealmConfiguration
```
kraenhansen, line 81: Would this be equivalent?

```suggestion
  Omit<RealmProviderProps, "realm">
```

packages/realm-react/src/RealmProvider.tsx
kraenhansen: Alternatively, perhaps these keys could be extracted to its own type.

packages/realm-react/src/index.tsx
```diff
@@ -19,109 +19,28 @@
...
+} from "./RealmProvider";
 import { createUseObject } from "./useObject";
 import { createUseQuery } from "./useQuery";
 import { createUseRealm } from "./useRealm";
+import { RealmContext } from "./RealmContext";
 
 export type { UseObjectHook } from "./useObject";
 export type { UseQueryHook, QueryHookOptions, QueryHookClassBasedOptions } from "./useQuery";
 
-type RealmContext = {
```
kraenhansen, line 30: I believe this was documented here for the docs to generate correctly, did you try to run the docs command to compare the different outputs? I did [upstream an issue](https://github.com/TypeStrong/typedoc/issues/2521) which seems to have been fixed and will be released in [an upcoming `typedoc@0.26.0` release](https://github.com/TypeStrong/typedoc/issues/2567).

packages/realm-react/src/index.tsx
```diff
@@ -135,11 +54,51 @@ type RealmContext = {
  * @param realmConfig - {@link Realm.Configuration} used to open the Realm
  * @returns An object containing a `RealmProvider` component, and `useRealm`, `useQuery` and `useObject` hooks
  */
-export const createRealmContext: (realmConfig?: Realm.Configuration) => RealmContext = (
-  realmConfig: Realm.Configuration = {},
-) => {
-  const RealmContext = createContext<Realm | null>(null);
-  const RealmProvider = createRealmProvider(realmConfig, RealmContext);
+export function createRealmContext(realmConfig: Realm.Configuration): RealmContext<RealmProviderFromConfiguration>;
```
kraenhansen, line 57: I would probably have expected to see this moved tot he new `src/RealmContext.ts` file 🤔

kraenhansen (approved):
I really like they way you're expressing the two mutually exclusive modes of the provider and also tying that back with the `createRealmContext` - Great stuff! 🎉 I have minor comments, but nothing blocking a merge.

kraenhansen:
I think it would be good to have a few tests using `@ts-expect-error` comments to assert the mutual exclusivity of the different props.

packages/realm-react/src/index.tsx
```diff
@@ -135,11 +54,51 @@ type RealmContext = {
...
+ * ...
+ * <RealmProvider realm={realm}></RealmProvider>
+ * ```
+ * @returns An object containing a `RealmProvider` component, and `useRealm`, `useQuery` and `useObject` hooks
+ */
+export function createRealmContext(): RealmContext<DynamicRealmProvider>;
+export function createRealmContext(
+  realmOrConfig?: Realm | Realm.Configuration,
+): RealmContext<RealmProviderFromConfiguration | RealmProviderFromRealmFC | DynamicRealmProvider> {
+  const RealmContext = createContext<Realm | null>(realmOrConfig instanceof Realm ? realmOrConfig : null);
```
kraenhansen, line 100: A side-effect of passing the `Realm` instance as the `defaultValue` argument of `createContext` is that users providing the Realm via `createRealmContext` actually doesn't have to wrap consumers (including the hooks derived from the context) in a `RealmProvider`. This is a great feature and I think it should be documented (and tested if it isn't already).

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Report the originating error that caused a client reset to occur. ([realm/realm-core#6154](https://github.com/realm/realm-core/issues/6154))
 * Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions. ([realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734))
 
+
```
elle-j, line 10:
```suggestion
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -16,7 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import React, { useContext, useEffect, useRef, useState } from "react";
+import React, { useContext, useEffect, useMemo, useRef, useState } from "react";
```
elle-j, line 19:
```suggestion
import React, { useContext, useEffect, useRef, useState } from "react";
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+
+/**
+ * Explicitly sets the unpicked properties of a type to never instead of dropping them like in Pick.
+ * Useful for ensuring different prop types are mutually exclusive as React expects the union type
+ * of different prop types to include all the fields.
+ */
+type RestrictivePick<T, K extends keyof T> = Pick<T, K> & { [RestrictedKey in keyof Omit<T, K>]?: never };
+
+/**
+ * Represents properties of a generic RealmProvider where the realm is set and therefore all other
```
elle-j, line 71:
```suggestion
 * Represents properties of a {@link DynamicRealmProvider} where the realm is set and therefore all other
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+ * props should be disallowed.
+ */
+export type RealmProviderWithRealmProps = RestrictivePick<RealmProviderProps, "realm" | "children">;
+
+/*
+ * Represents properties of a generic RealmProvider where Realm configuration props are used and no realm is set
```
elle-j, line 77:
```suggestion
/**
 * Represents properties of a {@link DynamicRealmProvider} where Realm configuration props are used and no realm is set.
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +41,71 @@ type ProviderProps = PartialRealmConfiguration & {
...
+>;
+
+/**
+ * Represents the provider returned from creating context with no arguments (including the default context).
+ * Supports either passing a `realm` as a property or the schema configuration.
+ */
+export type DynamicRealmProvider = React.FC<RealmProviderWithRealmProps | RealmProviderWithConfigurationProps>;
+
+export function createRealmProviderFromRealm(
+  realm: Realm | null,
```
elle-j, line 91: It looks like all the callers of this function passes a `Realm` instance?
```suggestion
  realm: Realm,
```

packages/realm-react/src/__tests__/RealmProvider.test.tsx
```diff
@@ -43,317 +44,556 @@ const catSchema: Realm.ObjectSchema = {
...
-    await waitFor(() => expect(result.current).not.toBe(null));
-    const realm = result.current;
-    expect(realm.schema[0].name).toBe("dog");
-    unmount();
-    expect(realm.isClosed).toBe(false);
-  });
+    it("returns the configured realm with useRealm", async () => {
+      const wrapper = ({ children }: { children: React.ReactNode }) => <RealmProvider>{children}</RealmProvider>;
+      const { result } = renderHook(() => useRealm(), { wrapper });
+      await waitFor(() => expect(result.current).not.toBe(null));
```
elle-j, line 64: Can we update the `not.toBe(null)`s to the more specific `toBeInstanceOf(Realm)` instead? 🙂

packages/realm-react/src/__tests__/RealmProvider.test.tsx
```diff
@@ -43,317 +44,556 @@ const catSchema: Realm.ObjectSchema = {
...
+      const realm = result.current;
+      expect(realm).not.toBe(null);
```
elle-j, line 66: This `expect()` may not be needed given the earlier `waitFor()`.

packages/realm-react/src/__tests__/RealmProvider.test.tsx
```diff
@@ -43,317 +44,556 @@ const catSchema: Realm.ObjectSchema = {
...
-          {toggleComponent && (
-            <View testID="secondRealmProvider">
+      const { result, unmount } = renderHook(() => useRealm(), { wrapper });
+      await waitFor(() => expect(result.current).not.toBe(null));
+      const realm = result.current;
+      unmount();
+      expect(realm.isClosed).toBe(false);
+    });
+
+    it("will override the the configuration provided in createRealmContext", async () => {
```
elle-j, line 90:
```suggestion
    it("will override the configuration provided in createRealmContext", async () => {
```

packages/realm-react/src/RealmProvider.tsx
elle-j, line 1: For the selection of props I think it'd be beneficial to find a way to not have to maintain those types too much when/if modifying our realm configuration (or realm-react-specific props). Perhaps you can find some combination of what you have right now and the ideas discussed in [this thread](https://github.com/realm/realm-js/pull/6714#discussion_r1639505429).

elle-j (approved):
Awesome work and nicely documented! 🙂

packages/realm-react/src/RealmContext.ts
gagik: this is a copy-paste from the current documentation, may need to check

packages/realm-react/src/index.tsx
gagik: Ah I see, didn't think about type-based doc generation, will check

packages/realm-react/src/index.tsx
gagik: that's a great point. should we even return a RealmProvider in this case then? seems misleading unless we plan to add other props to this in the future

packages/realm-react/src/index.tsx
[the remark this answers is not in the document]
gagik:
```ts
const realm = new Realm({ schema: [...] });
```
I can do this instead

packages/realm-react/src/RealmProvider.tsx
gagik: `helpers.ts` seems to fit better, will move there

kneth (approved):

packages/realm-react/src/RealmContext.ts
gagik: Hm yeah it does seem like this isn't really the case, it always return a collection and not an array. Not sure what the useful of that line is anyways so I think I'll just remove it

packages/realm-react/src/index.tsx
gagik: @kraenhansen where was the issue before? 

I don't see a RealmContext in docs at all since this wasn't being exported before:
Was it with the Provider?
https://www.mongodb.com/docs/realm-sdks/react/latest/functions/RealmProvider.html

From what I can see docs seem largely unchanged 🤞

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -161,6 +193,48 @@ export function createRealmProvider(
...
+ * Generates a `RealmProvider` which is either based on a configuration
+ * or based on a realm, depending on its props.
+ * @param RealmContext - The context that will contain the Realm instance
+ * @returns a RealmProvider component that provides context to all context hooks
+ */
+export function createDynamicRealmProvider(RealmContext: React.Context<Realm | null>): DynamicRealmProvider {
+  return ({ realm, ...restProps }) => {
+    if (realm) {
+      const { children, ...disallowedProps } = restProps;
+      if (!isEqual(disallowedProps, {})) {
```
gagik, line 206: @kraenhansen @elle-j @kneth 
I think this is the last thing worth taking a look at. Added this so there can be runtime validation and a human-friendly error in case they pass both types of props. If no comments about that, should be ready to merge.

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -161,6 +193,47 @@ export function createRealmProvider(
...
+/**
+ * Generates a `RealmProvider` which is either based on a configuration
+ * or based on a realm, depending on its props.
+ * @param RealmContext - The context that will contain the Realm instance
+ * @returns a RealmProvider component that provides context to all context hooks
+ */
+export function createDynamicRealmProvider(RealmContext: React.Context<Realm | null>): DynamicRealmProvider {
+  return ({ realm, children, ...configurationProps }) => {
+    if (realm) {
+      if (!isEqual(configurationProps, {})) {
```
gagik, line 205: @kraenhansen @elle-j @kneth
I think this is the last thing worth taking a look at. Added this so there can be runtime validation and a human-friendly error in case they pass both types of props. If no comments about that, should be ready to merge.

kraenhansen: This could also be simplified (limiting dependence on the lodash dependency), although I don't have a strong opinion here:

```suggestion
      if (Object.keys(configurationProps).length === 0) {
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -161,6 +193,47 @@ export function createRealmProvider(
...
+        throw new Error("Cannot use configuration props when using an existing Realm instance.");
+      }
+
+      const RealmProviderFromRealm = createRealmProviderFromRealm(realm, RealmContext);
+      return <RealmProviderFromRealm>{children}</RealmProviderFromRealm>;
+    } else {
+      const RealmProviderFromConfig = createRealmProviderFromConfig({}, RealmContext);
+      return <RealmProviderFromConfig {...configurationProps}>{children}</RealmProviderFromConfig>;
```
kraenhansen, line 213: NIT: Since these are never in scope together, it could be simpler to just name them `RealmProvider` here.

```suggestion
      const RealmProvider = createRealmProviderFromRealm(realm, RealmContext);
      return <RealmProvider>{children}</RealmProvider>;
    } else {
      const RealmProvider = createRealmProviderFromConfig({}, RealmContext);
      return <RealmProvider {...configurationProps}>{children}</RealmProvider>;
```

packages/realm-react/CHANGELOG.md
```diff
@@ -4,7 +4,39 @@
...
+function MyApp() {
+  return (
+	<>
+    <RealmProvider> 
+      ...
+    </RealmProvider>
+	<AnotherComponent>
+	{/* Note: The hooks returned from `createRealmContext` using an existing Realm can be used outside of the scope of the provider! */}
+	</AnotherComponent>
+	</>
```
kraenhansen, line 37:
```suggestion
  return (
    <>
      <RealmProvider> 
        ...
      </RealmProvider>
      <AnotherComponent>
        {/* Note: The hooks returned from `createRealmContext` using an existing Realm can be used outside of the scope of the provider! */}
      </AnotherComponent>
    </>
```

kraenhansen (approved):

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -41,43 +39,77 @@ type ProviderProps = PartialRealmConfiguration & {
...
+/** Props used for a Realm instance-based Realm provider */
+type RealmProviderRealmProps = {
+  /**
+   * The Realm instance to be used by the provider.
+   */
+  realm: Realm;
   children: React.ReactNode;
 };
 
+type RealmProviderProps = RealmProviderConfigurationProps & RealmProviderRealmProps;
```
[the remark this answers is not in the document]
elle-j: 🤟

PR #6720, opened 2024-06-12
Fix starting local Docker

elle-j:
## What, How & Why?

For the tests, when running `baas-test-server docker` (or `npm start`) the docker images output was malformed causing `JSON.parse()` to throw.


kraenhansen (approved):
@kneth saw this as well and fixed it by upgrading docker 🙂 But you can merge if you don't want to do that 😄

elle-j:
Probably good to do both 😉

kneth (approved):
Merge when you are ready

PR #6730, opened 2024-06-17
Revert "RJS-2717: Replaced lodash with lodash.isequal"

kneth:
Reverts realm/realm-js#6633

It seems to break https://github.com/realm/realm-js/pull/6729


elle-j (commented):
[Looks](https://github.com/realm/realm-js/actions/runs/9545669228/job/26306922212?pr=6729) like the only change that should be needed (instead of reverting) is to:
* Replace the 2 occurrences of:
```js
import { isEqual } from "lodash.isequal";
```
* With:
```js
import isEqual from "lodash.isequal";
```

kneth:
Closing as it can be solved: https://github.com/realm/realm-js/pull/6731 (similar to what @elle-j suggests above).

Issue #6653, opened 2024-05-03
Support "bridgeless" mode by exposing a TurboModule on iOS


kraenhansen:
Historically our library has always been bridgeless, in the sense that we never really used the bridge to communicate from JS and instead relied on internal APIs to get a pointer to the JS runtime and expose our C++ library code through JS by constructing JS objects and functions via JSC and later JSI calls.

We did migrate the way we load our native code on Android to use a TurboModule, but the mechanism we use to initialize our native binding on iOS is still a hack and rely on the bridge. Since React Native 0.74.0 enables bridgeless by default when the new architecture is enabled, this refactor is overdue.

Relevant documentation:
- [Cross-Platform Turbo Native Modules with C++](https://github.com/reactwg/react-native-new-architecture/blob/main/docs/turbo-modules-xplat.md)
- [Supporting Custom C++ Types](https://github.com/reactwg/react-native-new-architecture/blob/main/docs/cxx-custom-types.md)

PR #6718, opened 2024-06-12
React Native bridgeless

kraenhansen:
## What, How & Why?

This closes #6653 by introducing two implementations of the way we register our native module on both iOS and Android conditionally based on the enablement of the new architecture in the consuming app.

This needs to be rebased on `main` once https://github.com/realm/realm-js/pull/6650 gets merged.

## ☑️ ToDos
* [ ] Ensure `DEBUG` definition gets propagated and the expected code is called when "invalidating".
* [ ] Ensure the SSL helper gets passed correctly on Android.
* [ ] Add testing of new / old architecture on CI.
* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 📝 Update `COMPATIBILITY.md`
* [ ] 🚦 Tests
* [ ] 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)
* [ ] 📱 Check the React Native/other sample apps work if necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

PR #6737, opened 2024-06-18
RJS-2815: React Native bridgeless support (take 2)

kraenhansen:
## What, How & Why?

This closes https://github.com/realm/realm-js/issues/6653 by introducing a new implementations of the way we register our native module on both iOS and Android. I suggest relying on the legacy `NativeModules` compatibility layer for now, since the current state of pure C++ TurboModules are largely undocumented and experimental: We could revisit this decision in the future by introducing an alternative implementation, conditional on the new architecture being enabled in the app.

This approach aligns with [others vendors having exposing their native C++ / Rust based native libraries through JSI](https://ditto.live/blog/bridging-react-native-and-rust-via-jsi).

This still injects the binding into the JS global, from where the "JS wrapper" of our native module is able to read and delete it again. With these changes, we're now deferring the injection of this (and the loading of our .so on Android) until the very last moment it's needed, essentially getting the benefits of deferred loading that TurboModules bring:

https://github.com/realm/realm-js/blob/d774f3543df1161afef1f8db8c69b83a906993b3/packages/realm/binding/android/src/main/java/io/realm/react/RealmReactModule.java#L63-L66

Since we're implementing and injecting our binding onto the a global using JSI, we're not relying on the legacy bridge compatibility layer for every call from JS into C++ and I therefore expect this to be as fast as a TurboModule would be.

I did attempt to inject the "JS call invoker" for the "UI flush workaround" only if bridgeless is disabled, as https://github.com/facebook/react-native/pull/43396 fixed this for bridgeless and we don't want to pay the cost if it isn't needed, but it didn't work. We could still experiment with using the JSI [Runtime::drainMicrotasks()](https://github.com/facebook/hermes/blob/cf111e67f958e5734fc2546e428e9c6238b73364/API/jsi/jsi/jsi.h#L246) API if available.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


integration-tests/environments/react-native-test-app/harness/runner.js
```diff
@@ -139,7 +139,7 @@ async function run(spawnLogcat) {
     if (PLATFORM === "android") {
       // Start the log cat (skipping any initial pid from an old run)
       if (spawnLogcat) {
-        logcat.start("com.realmreactnativetests", true).catch(console.error);
+        logcat.start("com.microsoft.reacttestapp", true).catch(console.error);
```
kraenhansen, line 142: Technically no longer needed, but it does fix an issue of spawning logcat when running locally.

packages/realm/binding/android/src/main/cpp/io_realm_react_RealmReactModule.h
```diff
@@ -1,45 +0,0 @@
-/* DO NOT EDIT THIS FILE - it is machine generated */
```
kraenhansen, line 1: This file is not needed as no code is referring to these symbols externally.

packages/realm/binding/jsi/flush_ui_queue_workaround.cpp
```diff
@@ -0,0 +1,62 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "flush_ui_queue_workaround.h"
+
+#include <ReactCommon/CallInvoker.h>
+
+namespace realm::js::flush_ui_workaround {
```
kraenhansen, line 23: As part of https://github.com/realm/realm-js/pull/6718 I did a refactor of the "flush UI workaround" which I decided to keep in this PR.

packages/realm/binding/android/src/main/java/io/realm/react/RealmReactModule.java
```diff
@@ -16,110 +16,103 @@
...
     }
 
-    public RealmReactModule(ReactApplicationContext reactContext) {
-        super(reactContext);
+    @ReactMethod(isBlockingSynchronousMethod = true)
+    public boolean injectModuleIntoJSGlobal() {
+        // Calling loadLibrary multiple times will be ignored
+        // We do it here instead of statically to allow the app load faster if Realm isn't accessed.
+        // Effectively emulating the behaviour of TurboModules.
+        SoLoader.loadLibrary("realm");
```
kneth, line 66: Should we wrap it in a `try`/`catch` to provide a better error mesage?

packages/realm/binding/android/src/main/cpp/io_realm_react_RealmReactModule.cpp
```diff
@@ -144,12 +140,17 @@ JNIEXPORT void JNICALL Java_io_realm_react_RealmReactModule_setupFlushUiQueue(JN
...
+    realm::js::flush_ui_workaround::inject_js_call_invoker(nativePointer->getCallInvoker());
 }
+
+extern "C" JNIEXPORT void JNICALL Java_io_realm_react_RealmReactModule_invalidateCaches(JNIEnv* env, jobject thiz)
+{
+    // Disable the flush ui workaround
+    realm::js::flush_ui_workaround::reset_js_call_invoker();
+    __android_log_print(ANDROID_LOG_VERBOSE, "Realm", "Invalidating caches");
+#if DEBUG
+    realm_jsi_close_sync_sessions();
```
kneth, line 153: Please include the comments as in `RealmReactModule.m`

kneth (approved):
My comments are not blocking but merely suggestions and questions to clarify a few things (for me)

packages/realm/binding/android/src/main/cpp/io_realm_react_RealmReactModule.cpp
```diff
@@ -74,61 +69,62 @@ JNIEXPORT void JNI_OnUnload(JavaVM* vm, void*)
...
-JNIEXPORT void JNICALL Java_io_realm_react_RealmReactModule_setDefaultRealmFileDirectory(JNIEnv* env, jobject,
-                                                                                         jstring fileDir,
-                                                                                         jobject javaAssetManager)
+extern "C" JNIEXPORT void JNICALL Java_io_realm_react_RealmReactModule_injectModuleIntoJSGlobal(JNIEnv* env,
+                                                                                                jobject thiz,
+                                                                                                jlong runtime_pointer)
 {
-    __android_log_print(ANDROID_LOG_VERBOSE, "JSRealm", "setDefaultRealmFileDirectory");
+
+    __android_log_print(ANDROID_LOG_VERBOSE, "Realm", "install");
```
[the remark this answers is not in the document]
kraenhansen: Core isn't necessarily loaded into memory at the point of calling this.

packages/realm/binding/android/src/main/java/io/realm/react/RealmReactModule.java
kraenhansen: I imagine the SoLoader will provide sufficient information - we can add to this if we see common patterns emerge, but right now I don't know what I would write to make it more actionable for developers 🤔 It used to be the case that people forgot to `react-native link` but that's a thing of the past.

packages/realm/RealmJS.podspec
```diff
@@ -40,10 +40,20 @@ Pod::Spec.new do |s|
   # @see https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#platform-ios
   s.source                 = { :http => 'https://github.com/realm/realm-js/blob/main/CONTRIBUTING.md#how-to-debug-react-native-podspec' }
 
+
   s.source_files           = 'binding/jsi/*.cpp',
-                             'binding/apple/*.mm'
+                             'binding/apple/platform.mm',
```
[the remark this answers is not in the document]
kraenhansen: Nope, I think I just went the more explicit route due to the other files in that list.

kraenhansen:
I included a fix in #acb518d4f6d244cd31172a9463a6d3076293e6d5 to solve #6787.

ferrannp:
I have still not being able to migrate my project to new architecture but super happy to see this ongoing. Just writing to say thank you @kraenhansen !

kraenhansen:
@ferrannp Honestly, a comment like that means more than you might know 🙂 Thank you!

Issue #5500, opened 2023-02-26
Reenable tests


kneth:
A number of FLX tests are timing out, and we need to investigate why.

For now, the following tests have been disabled by https://github.com/realm/realm-js/pull/5491/commits/826a9a585f2aff65f9a0724766bc85302606abf3

* `client reset fails, the error handler is called`
* `handles discard local simulated client reset with`
* `handles simulated client reset with recovery with`

kneth:
The function to trigger a client reset is not able to delete any client files. Therefore the client reset is not triggered. The same function works fine with partition based sync.

kraenhansen:
Similarly to realm/realm-cpp#218 we can now depend on an endpoint of the Admin API to trigger client resets from tests: See https://github.com/10gen/baas/pull/12398 and [adminbaas.sh#L122-L126](https://github.com/10gen/baas/blob/7acb87591c81673de455b6c2fe4ba7fa612639cf/etc/scripts/adminbaas.sh#L122-L126) for context.

kraenhansen:
Just noticed the `get_file_ident` is already in the Realm Core spec: https://github.com/realm/realm-core/blob/970c4fd89f3c9b18ffd26f075534992e2fd8fa51/bindgen/spec.yml#L1397

PR #6733, opened 2024-06-17
Exposing an internal API to get file ident of a `SyncSession`

kraenhansen:
## What, How & Why?

This is prerequisite for #5500.

PR #6738, opened 2024-06-19
RJS-2204: Re-enabled client reset tests

kraenhansen:
## What, How & Why?

Fixes https://github.com/realm/realm-js/issues/5500 by calling the admin API to perform client reset.
This also brings a refactor of the app importer to take the client as an argument, since we want to instantiate and call this outside of the context of importing apps.

Note: This should be rebased once https://github.com/realm/realm-js/pull/6733 merge.

### DIFF
--- a/integration-tests/tests/src/tests/sync/client-reset.ts
+++ b/integration-tests/tests/src/tests/sync/client-reset.ts
@@ -32,6 +32,7 @@ import { DogSchema, PersonSchema } from "../../schemas/person-and-dog-with-objec
 import { expectClientResetError } from "../../utils/expect-sync-error";
 import { createPromiseHandle } from "../../utils/promise-handle";
 import { buildAppConfig } from "../../utils/build-app-config";
+import { baasAdminClient } from "../../utils/baas-admin-api";
 
 const FlexiblePersonSchema = { ...PersonSchema, properties: { ...PersonSchema.properties, nonQueryable: "string?" } };
 const FlexibleDogSchema = { ...DogSchema, properties: { ...DogSchema.properties, nonQueryable: "string?" } };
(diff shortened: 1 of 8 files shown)


integration-tests/tests/src/tests/sync/client-reset.ts
```diff
@@ -53,18 +54,16 @@ function getPartitionValue() {
...
-    count--;
+async function triggerClientReset(app: App, syncSession: Realm.App.Sync.Session): Promise<void> {
+  const { fileIdent } = syncSession as unknown as Record<string, unknown>;
+  if (typeof fileIdent !== "bigint") {
+    throw new Error("Expected the internal file ident");
   }
-  throw new Error(`Cannot trigger client reset in ${maxAttempts} attempts`);
+  await baasAdminClient.ensureLogIn();
+  const { _id } = await baasAdminClient.getAppByClientAppId(app.id);
+  syncSession.pause();
```
kneth, line 64: Do we need a connection listener to ensure it is safe to call `forceSyncReset()`?

kraenhansen: Safe in what way? That we have access to BaaS from the test suite or perhaps that the Realm is ready to get the reset? (for the latter we are doing a `await syncSession.uploadAllLocalChanges()` before calling `triggerClientReset`).

kneth: `pause()` will not immediately pause the connection, and I wonder if we need to wait for it before calling `forceSyncReset()`?

kraenhansen: Ah 🤔 I honestly don't know .. I think it's safe to assume `resume` will internally block for anything async that `pause` started on the sync client thread.

elle-j (approved):
LGTM 💪

kneth (approved):

PR #6742, opened 2024-06-20
Creating symlinks for `ccache` on CI instead of modifying `PATH`

kraenhansen:
## What, How & Why?

I hope this will help CI use ccache more (it seems the integration test job for RN iOS isn't hitting caches at all).

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -517,7 +517,7 @@ jobs:
     env:
       DEVELOPER_DIR: /Applications/Xcode_14.3.1.app
       IOS_DEVICE_NAME: iPhone 14
-    runs-on: macos-latest-xlarge
+    runs-on: macos-latest-large
     timeout-minutes: 60
     steps:
       - name: Checkout code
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -182,16 +182,7 @@ jobs:
...
-
-      - name: Prepend ccache executables to the PATH
-        if: ${{ runner.os != 'Windows' }}
-        run: |
-          echo "/usr/lib/ccache:/usr/local/opt/ccache/libexec" >> $GITHUB_PATH
-
-      # in CI file timestamps change with every run so instead rely on file content hashing
-      # https://reactnative.dev/docs/build-speed#using-this-approach-on-a-ci
-      - name: Configure ccache
-        run: ccache --set-config="compiler_check=content"
```
kraenhansen, line 194: As per https://github.com/hendrikmuhs/ccache-action/pull/147 this is already set by the action on macOS.

.github/workflows/pr-realm-js.yml
```diff
@@ -182,16 +182,7 @@ jobs:
         with:
           key: ${{ runner.os }}-${{ matrix.os }}-${{ matrix.arch }}
           max-size: '2.0G'
-
-      - name: Prepend ccache executables to the PATH
-        if: ${{ runner.os != 'Windows' }}
-        run: |
-          echo "/usr/lib/ccache:/usr/local/opt/ccache/libexec" >> $GITHUB_PATH
```
kraenhansen, line 189: The action will create symlinks in `/usr/local/bin`, which is why we no longer need to modify the path.

kraenhansen:
I included this in https://github.com/realm/realm-js/pull/6746

PR #6743, opened 2024-06-21
RJS-2636: Add progress notifications and tests

gagik:
## What, How & Why?

Adds progress notifications for flexible sync and tests for it.

```typescript
realm.syncSession?.addProgressNotification(
  ProgressDirection.Upload,
  ProgressMode.ReportIndefinitely,
  (estimate) => console.log(`progress: ${estimate}/1.0`)
);
```

This closes https://github.com/realm/realm-js/issues/6256

@realm/devdocs 

## ☑️ ToDos

* [X] 📝 Changelog entry
* [X] 🚦 Tests
* [ ] ~📦 Updated internal package version in consuming `package.json`s (if updating internal packages)~
* [ ] ~📱 Check the React Native/other sample apps work if necessary~


integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -49,6 +50,7 @@ const PersonForSyncSchema: Realm.ObjectSchema = {
   },
 };
 
+// TODO: Fix this.
```
gagik, line 53: This is unrelated to this PR, I can create a follow-up issue but waitForConnectionState seems to be a complete noop right now, may mean some tests either don't need it or are meaningless.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -437,6 +442,264 @@ describe("Flexible sync", function () {
...
+
+          // There should be at least one point where the progress is not yet finished.
+          expect(callback.args.find(([estimate]) => estimate < 1)).to.not.be.undefined;
+
+          expect(callback.withArgs(1.0).called).to.be.true;
+        });
+      });
+
+      describe(`with ProgressDirection.Download`, function () {
+        it("should call callback with 1.0 when there is nothing to download", async function (this: RealmContext) {
```
gagik, line 643: this is weird... why is there divergence with upload and download?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -437,6 +442,264 @@ describe("Flexible sync", function () {
...
+            },
+          });
+
+          realm2.syncSession?.addProgressNotification(
+            Realm.ProgressDirection.Download,
+            Realm.ProgressMode.ReportIndefinitely,
+            callback,
+          );
+
+          await realm2.objects(Person).subscribe();
```
gagik, line 572: can't think of a good way to have a proper download-upload situation, any ideas? another weirdness is that the estimates are
[ 1, 1, 1, 1, 1, 1, ... 1, ] no real progress
uploads on the other hand are more reasonable [1, 0, 0.3123, 0,4112.... ,1]

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -286,16 +288,14 @@ describe("SessionTest", () => {
...
-          progressCalled = true;
-        }),
-        throwAfterTimeout(5000),
-      ]);
-      expect(progressCalled).to.be.true;
+
+      const progressCallback = spy();
+      await Realm.open(config).progress(progressCallback);
+
+      expect(progressCallback.called).is.true;
```
gagik, line 298: Not going to spend much time refactoring the other tests since they related to the deprecated method, just thought this was a nice opportunity for sinon (also just timeout).

gagik:
I am still battling some CI issues but I figure it may be worth requesting some reviews for now.

integration-tests/tests/src/tests/sync/sync-session.ts
gagik: Also, there's a sinon-chai integration and I will switch to that soon, maybe in a different PR if there's not enough time for now

integration-tests/tests/package.json
```diff
@@ -84,6 +84,8 @@
     "@types/chai-as-promised": "^7.1.5",
     "@types/jsrsasign": "^10.5.4",
     "@types/mocha": "^10.0.0",
+    "@types/sinon": "^9.0.5",
+    "assert": "^2.1.0",
```
gagik, line 88: This polyfill seems necessary for electron... will double check

integration-tests/tests/src/tests/sync/sync-session.ts
gagik: https://github.com/realm/realm-js/issues/6759

gagik:
Okay, I am 99% certain that last failing tests is a result of GHA caching, and I tried to delete the cache manually many times but seems to keep using, @kraenhansen may have a better idea of what is going on here... but will have to leave it at that since this now 62 commits deep into making it fully green 😄

Will leave it at this point since there's no chance to have a proper reviews but feel free to continue from here to get it merged. I'd say it is pretty complete and although tests are far from ideal, there is also no clear behavior expectations yet and much of it is just up to core, so should be good to merge beyond any review comments you can have. Feel free to get in touch if needed in next few days.

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -167,18 +175,31 @@ export class ProgressRealmPromise implements Promise<Realm> {
...
     const transferable = binding.Int64.intToNum(transferableArg);
     for (const listener of this.listeners) {
-      listener(transferred, transferable);
+      if (listener.length === 1) {
+        const estimateListener = listener as DynamicProgressNotificationCallback;
+        estimateListener(progressEstimate);
+      } else {
+        const pbsListener = listener as PartitionBasedSyncProgressNotificationCallback;
+        pbsListener(transferred, transferable);
+      }
```
elle-j, line 202:
```suggestion
      if (isEstimateProgressNotification(listener)) {
        listener(progressEstimate);
      } else {
        listener(transferred, transferable);
      }
```

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
...
-export type ProgressNotificationCallback =
+/** @deprecated */
+export type PartitionBasedSyncProgressNotificationCallback =
   /**
+   * This notification callback only supports Partition Based Sync.
    * @param transferred - The current number of bytes already transferred
    * @param transferable - The total number of transferable bytes (i.e. the number of bytes already transferred plus the number of bytes pending transfer)
+   * @deprecated - This notification callback will be removed in next major release. Use {@link DynamicProgressNotificationCallback} instead.
+   * @since 1.12.0
    */
```
elle-j, line 64: This could be moved onto the `type` itself 👍

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
...
 }
 
-export type ProgressNotificationCallback =
+/** @deprecated */
+export type PartitionBasedSyncProgressNotificationCallback =
   /**
+   * This notification callback only supports Partition Based Sync.
    * @param transferred - The current number of bytes already transferred
    * @param transferable - The total number of transferable bytes (i.e. the number of bytes already transferred plus the number of bytes pending transfer)
+   * @deprecated - This notification callback will be removed in next major release. Use {@link DynamicProgressNotificationCallback} instead.
```
elle-j, line 62:
```suggestion
   * @deprecated - This notification callback will be removed in the next major release. Use {@link DynamicProgressNotificationCallback} instead.
```

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
...
+   * @since 1.12.0
    */
   (transferred: number, transferable: number) => void;
 
+export type DynamicProgressNotificationCallback =
+  /**
+   * This notification callback supports both Partition Based Sync and Flexible Sync.
+   * @param estimate - An estimate between 0.0 and 1.0 of how much have been transferred.
+   * @since 12.10.0
+   */
```
elle-j, line 72: This could be moved onto the `type` itself 👍

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
...
+   * @deprecated - This notification callback will be removed in next major release. Use {@link DynamicProgressNotificationCallback} instead.
+   * @since 1.12.0
    */
   (transferred: number, transferable: number) => void;
 
+export type DynamicProgressNotificationCallback =
+  /**
+   * This notification callback supports both Partition Based Sync and Flexible Sync.
+   * @param estimate - An estimate between 0.0 and 1.0 of how much have been transferred.
+   * @since 12.10.0
```
elle-j, line 71: `@since` should be updated (or removed 🤔 ) @kneth , when do we usually add `@since`?

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
...
   /**
+   * This notification callback only supports Partition Based Sync.
    * @param transferred - The current number of bytes already transferred
    * @param transferable - The total number of transferable bytes (i.e. the number of bytes already transferred plus the number of bytes pending transfer)
+   * @deprecated - This notification callback will be removed in next major release. Use {@link DynamicProgressNotificationCallback} instead.
+   * @since 1.12.0
    */
   (transferred: number, transferable: number) => void;
 
+export type DynamicProgressNotificationCallback =
```
elle-j, line 67: In this context `Dynamic` may be ambiguous. It may seem implicit since the progress itself is dynamic, whic PBS also is. But I understand you're referring to PBS-only vs Flx-or-PBS. At another point I saw you used `isEstimateProgress..`. What do you think of that name instead (`EstimateProgressNotificationCallback`)?

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
...
+  /**
+   * This notification callback supports both Partition Based Sync and Flexible Sync.
+   * @param estimate - An estimate between 0.0 and 1.0 of how much have been transferred.
+   * @since 12.10.0
+   */
+  (estimate: number) => void;
+
+export type ProgressNotificationCallback =
```
elle-j, line 75: Could we add some short API docs here describing the callback in general, and then let the specific callbacks include more details in their respective docs.

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -93,6 +109,21 @@ function toBindingDirection(direction: ProgressDirection) {
   }
 }
 
+function isEstimateProgressNotification(cb: ProgressNotificationCallback): cb is DynamicProgressNotificationCallback {
```
elle-j, line 112: 1. I believe a `Callback` postfix is missing from the name.
2. Could we use the fully spelled out name `callback` for increased consistency?

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -93,6 +109,21 @@ function toBindingDirection(direction: ProgressDirection) {
...
+function toBindingProgressNotificationCallback(cb: ProgressNotificationCallback) {
+  if (isEstimateProgressNotification(cb)) {
+    return (transferredBytes: binding.Int64, transferrableBytes: binding.Int64, progressEstimate: number) =>
+      cb(progressEstimate);
+  } else {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    return (transferredBytes: binding.Int64, transferrableBytes: binding.Int64, _: number) =>
+      cb(transferrableBytes, transferrableBytes);
+  }
+}
```
elle-j, line 125: * Nit: Argument name
* The 2nd callback was accidentally called without `transferredBytes`.
```suggestion
function toBindingProgressNotificationCallback(callback: ProgressNotificationCallback) {
  if (isEstimateProgressNotification(callback)) {
    return (_: binding.Int64, __: binding.Int64, progressEstimate: number) => callback(progressEstimate);
  } else {
    return (transferredBytes: binding.Int64, transferrableBytes: binding.Int64, _: number) =>
      callback(transferredBytes, transferrableBytes);
  }
}
```

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -251,7 +282,7 @@ const PROGRESS_LISTENERS = new Listeners<
 >({
   add(callback, weakInternal, internal, direction, mode) {
     const token = internal.registerProgressNotifier(
-      (transferred, transferable) => callback(Number(transferred), Number(transferable)),
+      toBindingProgressNotificationCallback(callback),
```
elle-j, line 285: 💪

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -441,12 +472,17 @@ export class SyncSession {
    * Can be either:
    *  - `reportIndefinitely` - the registration will stay active until the callback is unregistered
    *  - `forCurrentlyOutstandingWork` - the registration will be active until only the currently transferable bytes are synced
-   * @param callback - Called with the following arguments:
-   * 1. `transferred`: The current number of bytes already transferred
-   * 2. `transferable`: The total number of transferable bytes (the number of bytes already transferred plus the number of bytes pending transfer)
+   * @param callback - see {@link DynamicProgressNotificationCallback} and {@link PartitionBasedSyncProgressNotificationCallback}
```
elle-j, line 475: Perhaps something similar to this could suffice (mainly because the type itself exists by the parameter; so similarly to e.g. `mode`, we don't refer to the `ProgressMode`s available in the doc description here 👍 ).
```suggestion
   * @param callback - The function to call when the progress is updated.
```

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -441,12 +472,17 @@ export class SyncSession {
...
+   * @throws if signature of `callback` doesn't match requirements for sync mode
    * @since 1.12.0
    */
   addProgressNotification(direction: ProgressDirection, mode: ProgressMode, callback: ProgressNotificationCallback) {
+    if (this.config.flexible) {
+      assert(
+        isEstimateProgressNotification(callback),
+        `For flexible sync the callback can only take one argument - got ${callback.length} arguments`,
+      );
+    }
```
elle-j, line 485: I think this may be a breaking change, since it's still been possible to call the API, and we haven't documented that it should have thrown before. Hmm, could we warn here instead?

elle-j (commented):
Comments from a first pass here, looks really nice 🙂

packages/realm/src/app-services/SyncSession.ts
```diff
@@ -53,13 +53,29 @@ export enum ProgressMode {
   ForCurrentlyOutstandingWork = "forCurrentlyOutstandingWork",
 }
 
-export type ProgressNotificationCallback =
+/** @deprecated */
+export type PartitionBasedSyncProgressNotificationCallback =
```
elle-j, line 57: For the new public types that got introduced, they also need to be exported to users similar to [this](https://github.com/realm/realm-js/blob/a72bbc2c13ca2afc99b438ffc0aa207351c6b101/packages/realm/src/Realm.ts#L1326).

For deprecated ones such as the PBS-specific one, we list them below [this line](https://github.com/realm/realm-js/blob/a72bbc2c13ca2afc99b438ffc0aa207351c6b101/packages/realm/src/Realm.ts#L1369).

integration-tests/tests/package.json
kraenhansen: I'd like more context on this change.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
     });
   });
+
+  describe("Progress notification", function () {
+    this.timeout(5000);
+    before(async function (this: UserContext) {
+      // TODO: Investigate this.
+      // Use a new user for this test suite as not doing so causes some session issues in other tests
+      this.user = await this.app.logIn(Credentials.anonymous(false));
+    });
```
kraenhansen, line 2091: Are you aware of the `authenticateUserBefore` hook? If its important to avoid reusing, you could path https://github.com/realm/realm-js/blob/f8aefa20ebd03a9178b79ea8fa0ed72bf206696e/integration-tests/tests/src/hooks/authenticate-user-before.ts#L24 to take a boolean argument?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+      // @ts-expect-error Using an internal API
+      this.realm = new Realm({
+        schema: [Person, Dog],
+        sync: {
+          flexible: true,
+          user: this.user,
+          _sessionStopPolicy: SessionStopPolicy.Immediately,
+        },
+      });
+    });
```
kraenhansen, line 2102: Is there a reason you're not using `openRealmBeforeEach` here?

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+
+    it("only estimate callback is allowed", async function (this: RealmContext) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
+      const callback = (estimate: number) => {};
+      this.realm.syncSession?.addProgressNotification(
+        Realm.ProgressDirection.Download,
+        Realm.ProgressMode.ForCurrentlyOutstandingWork,
+        callback,
+      );
+    });
```
kraenhansen, line 2112: What is this testing specifically? Passing a single argument arrow function as callback?

integration-tests/tests/package.json
gagik: It seemed necessary for sinon but I had since then changed the version to something which has more compatibility so it may not be needed anymore

packages/realm/src/app-services/SyncSession.ts
kraenhansen: I agree - this feels more breaking than a deprecation (which might be okay? but should be explicit and released accordingly) if it actually throws instead of simply warning.

integration-tests/tests/src/tests/sync/flexible.ts
gagik: Hm, I was moving back and forth between using that or my own as I had other cleanup hooks to check different things but now it seems like this can be fully covered by`openRealmBeforeEach`, feel free to change it.

integration-tests/tests/src/tests/sync/flexible.ts
gagik: Yes, this has to do with deprecated API vs non-deprecated

packages/realm/src/app-services/SyncSession.ts
gagik: agree with the callback, these bits of code are a mix of mine and largely @kneth so he may want to weigh in as well

packages/realm/src/ProgressRealmPromise.ts
```diff
@@ -167,18 +175,31 @@ export class ProgressRealmPromise implements Promise<Realm> {
    */
   progress(callback: ProgressNotificationCallback): this {
     this.listeners.add(callback);
+    if (callback.length === 1) {
+      const estimateCallback = callback as DynamicProgressNotificationCallback;
+      estimateCallback(1.0);
+    } else {
+      const pbsCallback = callback as PartitionBasedSyncProgressNotificationCallback;
+      pbsCallback(0.0, 0.0);
+    }
```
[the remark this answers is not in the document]
gagik: I think the manual trigger was to imitate past behavior, it may only be relevant for `estimateCallback` rather than `pbsCallback` since I believe before .progresss would usually have these values at initialization, I didn't look too much into this as this was hitting a previous. Agree with adding a comment here to clear up why.

gagik:
Just tried to reply to some of the comments that may have been hard to guess without my input. Beyond that, going to be away so feel free to go ahead with what seems best 😄

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+          });
+
+          await realm.syncSession?.uploadAllLocalChanges();
+          subscription.unsubscribe();
+
+          // TODO: Find a way to fix
+          // There should be at least one point where the progress is not yet finished.
+          // expect(callback.args.find(([estimate]) => estimate < 1)).to.not.be.undefined;
+
+          expect(callback.withArgs(1.0)).called;
```
elle-j, line 2179: Unsure if `withArgs()` is the correct usage here since that seems to create a new spy (according to their API docs). If using that, you probably want to put the `called` onto the spy instead (similar to the above where `callback.notCalled` is used):
```js
expect(callback.withArgs(1.0).called).to.be.true
```

However, there are also these APIs:
```js
callback.calledWith(..)
callback.calledWithExactly(..)
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+          );
+
+          await this.realm.syncSession?.uploadAllLocalChanges();
+          await this.realm.syncSession?.downloadAllServerChanges();
+
+          // TODO: This callback should not be called at this stage but seems flakey
+          // and gets called with 1.0 at times, likely because of a race condition.
+          expect(callback.notCalled || callback.calledWith(1.0)).is.true;
+
+          realm.objects(Person).subscribe();
```
elle-j, line 2200:
```suggestion
          await realm.objects(Person).subscribe();
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+            callback,
+          );
+
+          await this.realm.syncSession?.uploadAllLocalChanges();
+          await this.realm.syncSession?.downloadAllServerChanges();
+
+          // TODO: This callback should not be called at this stage but seems flakey
+          // and gets called with 1.0 at times, likely because of a race condition.
+          expect(callback.notCalled || callback.calledWith(1.0)).is.true;
+          const subscription = await realm.objects(Person).subscribe();
```
elle-j, line 2161: Nit: May want to rename this to e.g. `persons` as it's a `Results` being returned, rather than a `Subscription`.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+          await realm.syncSession?.uploadAllLocalChanges();
+          const subscription2 = await realm2.objects(Person).subscribe();
+          await realm2.syncSession?.downloadAllServerChanges();
+
+          expect(callback.args.find(([estimate]) => estimate < 1)).to.not.be.undefined;
+          expect(callback.withArgs(1.0).callCount).is.greaterThanOrEqual(2);
+
+          subscription.unsubscribe();
+          subscription2.unsubscribe();
+          realm2.close();
```
elle-j, line 2295: If using the `openRealmBeforeEach()` hook, I think we could remove these explicit `close()` calls, even if it's on a different realm. The hook will also call `Realm.clearTestState()` after each test which closes and deletes all realm files in the default directory.

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+              name: "Heavy",
+              age: 36,
+              data: new ArrayBuffer(1_000_000),
+            });
+          });
+
+          await realm.syncSession?.uploadAllLocalChanges();
+          await realm2.syncSession?.downloadAllServerChanges();
+
+          expect(callback).has.callCount(oldCallCount);
```
elle-j, line 2352: I think earlier the `expect(callback.callCount).to.be.equal(oldCallCount)` was used, we could align there 👍

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -2075,4 +2081,400 @@ describe("Flexible sync", function () {
...
+            Realm.ProgressDirection.Download,
+            Realm.ProgressMode.ForCurrentlyOutstandingWork,
+            callback,
+          );
+
+          // TODO: Doesn't work
+          // There should be at least one point where the progress is not yet finished.
+          // expect(callback.args.find(([estimate]) => estimate < 1)).to.not.be.undefined;
+
+          expect(callback.withArgs(1.0)).calledOnce;
```
elle-j, line 2472: Note: We could align with the APIs used 👍

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -47,6 +50,7 @@ import { expectClientResetError } from "../../utils/expect-sync-error";
 import { createSyncConfig } from "../../utils/open-realm";
 import { createPromiseHandle } from "../../utils/promise-handle";
 import { buildAppConfig } from "../../utils/build-app-config";
+import { spy } from "sinon";
```
elle-j, line 53: Nit: Can move up to the 3rd party imports 🙂

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -22,8 +22,9 @@ import { authenticateUserBefore, importAppBefore, openRealmBefore } from "../../
 import { DogSchema } from "../../schemas/person-and-dog-with-object-ids";
 import { getRegisteredEmailPassCredentials } from "../../utils/credentials";
 import { generatePartition } from "../../utils/generators";
-import { sleep, throwAfterTimeout } from "../../utils/sleep";
+import { sleep } from "../../utils/sleep";
 import { buildAppConfig } from "../../utils/build-app-config";
+import { spy } from "sinon";
```
elle-j, line 27: Nit: Can move up to the 3rd party imports 🙂

integration-tests/tests/src/tests/sync/sync-session.ts
```diff
@@ -286,16 +287,14 @@ describe("SessionTest", () => {
...
-      await Promise.race([
-        Realm.open(config).progress(() => {
-          progressCalled = true;
-        }),
-        throwAfterTimeout(5000),
-      ]);
-      expect(progressCalled).to.be.true;
+
+      const progressCallback = spy();
+      await Realm.open(config).progress(progressCallback);
```
elle-j, line 295: 👍

CHANGELOG.md
```diff
@@ -1,10 +1,31 @@
...
+  ProgressMode.ReportIndefinitely,
+  (estimate) => console.log(`progress estimate: ${estimate}/1.0`),
+);
+/** @deprecated */
+realm.syncSession.addProgressNotification(
+  Realm.ProgressDirection.Upload,
+  ProgressMode.ReportIndefinitely,
+  (transferred, transferable) => console.log(`progress: ${(transferred / transferable)}/1.0`),
+);
+```
```
elle-j, line 18: Nice info 🙂 I'm wondering if the info about the new API is better served along with the Enhancement entry and refer to that one instead? We could still mention deprecated API in the current section.

CHANGELOG.md
```diff
@@ -1,10 +1,31 @@
...
 
 ### Enhancements
-* None
+* Added progress notifications support for flexible sync using a new callback argument. ([#6256](https://github.com/realm/realm-js/issues/6256))
+```ts
```
elle-j, line 22:
```suggestion
```typescript
```

CHANGELOG.md
```diff
@@ -1,10 +1,31 @@
...
+realm.syncSession.addProgressNotification(
```
elle-j, line 23:
```suggestion
realm.syncSession?.addProgressNotification(
```

elle-j (commented):
Great to see many tests here 🙂 Left some suggestions and thoughts

packages/realm/src/app-services/SyncSession.ts
elle-j: We could update it to e.g.:
```ts
if (this.config.flexible && !isEstimateProgressNotificationCallback(callback)) {
  /* eslint-disable-next-line no-console */
  console.warn("Deprecated progress notifier used. Please use 'EstimateProgressNotificationCallback' which accepts the estimate as the only argument.");
}
```

But I think we could even skip the warning, since the warning is part of the deprecation in the docs. (We don't usually warn for other deprecations.)

For now I've updated this to just assert that it's a function.

packages/realm/src/app-services/SyncSession.ts
elle-j: @kneth, I removed the `@since` annotation for now.

elle-j:
Note on Android tests: Still experiencing [signature mismatch](https://github.com/realm/realm-js/actions/runs/9796825413/job/27053644710?pr=6743).

([Reference](https://github.com/realm/realm-js/pull/6760))

kraenhansen:
@elle-j with https://github.com/realm/realm-js/pull/6769 merged, I'd expect the Android tests to run if this branch is rebased.

PR #6746, opened 2024-06-21
Refactor PR workflow into separate test jobs

kraenhansen:
## What, How & Why?

My work on https://github.com/realm/realm-js/pull/6742 turned into a refactor of the PR workflow separating the test job into three, each depending only on the jobs producing the artifacts they need. Taking us one step closer towards #6530.

- This makes it possible for the tests to start as soon as the artefacts they depend on become available (such that the iOS tests doesn't have to wait for the Node.js prebuilds and vice-versa).
- It also means, that if the prebuild of the Node native module fails the job can be re-run without the iOS tests having to re-run if they already passed.
- I figured out the reason and fixed the issue with our iOS test app not utilizing ccache and shaved off ⅓ (~ 10 minutes) of the runtime on that 🏃💨
- I also managed to remove a few workarounds and unneeded steps in the process.

### DIFF
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -517,7 +517,7 @@ jobs:
     env:
       DEVELOPER_DIR: /Applications/Xcode_14.3.1.app
       IOS_DEVICE_NAME: iPhone 14
-    runs-on: macos-latest-xlarge
+    runs-on: macos-latest-large
     timeout-minutes: 60
     steps:
       - name: Checkout code
@@ -545,13 +545,13 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-node-
 
-      - name: Restore React Native cache
-        uses: actions/cache@v4
-        with:
-          path: '**/Pods'
-          key: ${{ runner.os }}-react-native-test-app-${{ hashFiles('**/Podfile.lock', './src/**', './vendor/**') }}
-          restore-keys: |
-            ${{ runner.os }}-react-native-test-app-
+      # - name: Restore React Native cache
+      #   uses: actions/cache@v4
+      #   with:
+      #     path: '**/Pods'
+      #     key: ${{ runner.os }}-react-native-test-app-${{ hashFiles('**/Podfile.lock', './src/**', './vendor/**') }}
+      #     restore-keys: |
+      #       ${{ runner.os }}-react-native-test-app-
 
       - name: MSVC Setup
         if: ${{ runner.os == 'Windows' }}
--- a/.github/workflows/pr-realm-js.yml
+++ b/.github/workflows/pr-realm-js.yml
@@ -182,16 +182,7 @@ jobs:
         with:
           key: ${{ runner.os }}-${{ matrix.os }}-${{ matrix.arch }}
           max-size: '2.0G'
-
-      - name: Prepend ccache executables to the PATH
-        if: ${{ runner.os != 'Windows' }}
-        run: |
-          echo "/usr/lib/ccache:/usr/local/opt/ccache/libexec" >> $GITHUB_PATH
-
-      # in CI file timestamps change with every run so instead rely on file content hashing
-      # https://reactnative.dev/docs/build-speed#using-this-approach-on-a-ci
-      - name: Configure ccache
-        run: ccache --set-config="compiler_check=content"
+          create-symlink: true
 
       - name: Install dependencies
         # Ignoring scripts to prevent a prebuilt from getting fetched / built
@@ -265,10 +256,8 @@ jobs:
         with:
           key: ${{ runner.os }}-${{ matrix.platform }}
           max-size: '2.0G'
-      # On CI, file timestamps change with every run so instead rely on file content hashing
-      # https://reactnative.dev/docs/build-speed#using-this-approach-on-a-ci
-      - name: Configure ccache
-        run: ccache --set-config="compiler_check=content"
+          create-symlink: true
+          
       - name: Install dependencies
         run: npm ci
       - name: Build archive
@@ -351,11 +340,7 @@ jobs:
         with:
           key: ${{ runner.os }}-android-${{ matrix.architecture }}
           max-size: '2.0G'
-
-      # On CI, file timestamps change with every run so instead rely on file content hashing
-      # https://reactnative.dev/docs/build-speed#using-this-approach-on-a-ci
-      - name: Configure ccache
-        run: ccache --set-config="compiler_check=content"
+          create-symlink: true
 
       - name: Setup Java
         uses: actions/setup-java@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-realm-js.yml
```diff
@@ -131,23 +138,9 @@ jobs:
...
         uses: actions/setup-node@v4
         with:
-          node-version: 20.11.1
+          node-version: 20
           cache: npm
 
-      - name: Get NPM cache directory
-        id: npm-cache-dir
-        shell: bash
-        run: echo "dir=$(npm config get cache)" >> $GITHUB_OUTPUT
```
kraenhansen, line 140: Not needed as this is already handled by `actions/setup-node@v4`.

.github/workflows/pr-realm-js.yml
```diff
@@ -131,23 +138,9 @@ jobs:
       - name: Setup node version
         uses: actions/setup-node@v4
         with:
-          node-version: 20.11.1
```
kraenhansen, line 134: This was for a workaround in `prebuild` which is not released and this PR includes an upgrade to get the fix.

.github/workflows/pr-realm-js.yml
```diff
@@ -349,13 +332,9 @@ jobs:
...
         uses: hendrikmuhs/ccache-action@v1
         with:
-          key: ${{ runner.os }}-android-${{ matrix.architecture }}
+          key: ${{ github.job }}-${{ matrix.architecture }}
           max-size: '2.0G'
-
-      # On CI, file timestamps change with every run so instead rely on file content hashing
-      # https://reactnative.dev/docs/build-speed#using-this-approach-on-a-ci
-      - name: Configure ccache
-        run: ccache --set-config="compiler_check=content"
```
kraenhansen, line 358: This is already handled by `hendrikmuhs/ccache-action@v1`.

.github/workflows/pr-realm-js.yml
```diff
@@ -440,90 +430,110 @@ jobs:
...
+          node-version: 20
+          cache: npm
+
+      # - name: Restore React Native cache
+      #   uses: actions/cache@v4
+      #   with:
+      #     path: '**/Pods'
+      #     key: ${{ runner.os }}-react-native-test-app-${{ hashFiles('**/Podfile.lock', './src/**', './vendor/**') }}
+      #     restore-keys: |
+      #       ${{ runner.os }}-react-native-test-app-
```
kraenhansen, line 512: We should re-enable this to avoid the extraction of boost while running `pod install`, as it takes ~ 4½ minutes 🤯

.github/workflows/pr-realm-js.yml
```diff
@@ -440,90 +430,110 @@ jobs:
...
+        id: baas
+        uses: ./.github/actions/baas-test-server
+        with:
+          branch: ${{ env.BAAS_BRANCH }}
+        env:
+          BAASAAS_KEY: ${{ secrets.BAASAAS_KEY }}
+
+      - name: Create Mocha Remote Context
+        id: mocha-env
+        run: echo "context=syncLogLevel=warn,longTimeoutMs=${{ env.LONG_TIMEOUT }},baseUrl=${{ steps.baas.outputs.baas-url }}" >> $GITHUB_OUTPUT
```
kraenhansen, line 474: Little annoying that this is now duplicated in the workspace code 🤔

.github/workflows/pr-realm-js.yml
```diff
@@ -580,27 +570,98 @@ jobs:
...
-
-      - name: Run ${{matrix.variant.target}} (${{ matrix.variant.os}} / ${{ matrix.variant.environment }})
-        if: ${{ matrix.variant.os == 'ios' }}
+      # Because the Xcode project injects its own CCACHE_CONFIGPATH,
+      # the configuration set by the ccache action doesn't propagate.
+      - name: Expose Ccache config to Xcode
+        run: |
+          echo "CCACHE_DIR=`ccache --get-config cache_dir`" >> $GITHUB_ENV
+          echo "CCACHE_MAXSIZE=`ccache --get-config max_size`" >> $GITHUB_ENV
+          echo "CCACHE_COMPILERCHECK=`ccache --get-config compiler_check`" >> $GITHUB_ENV
```
kraenhansen, line 579: This was the fix for the main issue.
- The React Native ccache scripts use `CCACHE_CONFIGPATH`: https://github.com/facebook/react-native/blob/f7fe68861a52a0428698a9352d39a6e39afefe42/packages/react-native/scripts/xcode/ccache-clang.sh#L12
- This makes ccache disregard the system-wide configuration, specifically the `cache_dir` which is unexpected to the `hendrikmuhs/ccache-action@v1` action.

integration-tests/environments/react-native-test-app/package.json
```diff
@@ -109,7 +109,7 @@
       }
     },
     "pod-install:ci": {
-      "command": "pod-install",
+      "command": "cd ios && pod install --verbose",
```
kraenhansen, line 112: This helped me debug the progress and just gives a bit more output.

packages/realm/package.json
```diff
@@ -336,7 +336,7 @@
     "cross-env": "^7.0.3",
     "glob": "^10.3.12",
     "mocha": "^10.1.0",
-    "prebuild": "^12.1.0",
+    "prebuild": "^13.0.1",
```
kraenhansen, line 339: See the note above on upgrading "node".

.github/workflows/pr-realm-js.yml
```diff
@@ -440,90 +430,110 @@ jobs:
...
-        if: ${{ matrix.variant.environment == 'react-native-test-app' }}
-        run: ccache --set-config="compiler_check=content"
-
-      # Hermes doesn't work with Cocoapods 1.15.0
-      # https://forums.developer.apple.com/forums/thread/745518
-      - name: Install older Cocoapods version
-        if: ${{ matrix.variant.os == 'ios' }}
-        uses: maxim-lobanov/setup-cocoapods@v1
-        with:
-          version: 1.14.3
```
kraenhansen, line 511: This got fixed in 1.15.2 which is now installed on the runner and there's no need for the workaround anymore.

.github/workflows/pr-realm-js.yml
kraenhansen: I re-enabled this ...

gagik (approved):
Awesome! Quicker CI is always great 🚀

.github/workflows/pr-realm-js.yml
```diff
@@ -386,47 +366,57 @@ jobs:
...
-      SPAWN_LOGCAT: true
-      BAAS_BRANCH: master
-      # Pin the Xcode version
-      DEVELOPER_DIR: /Applications/Xcode_14.3.1.app
-      IOS_DEVICE_NAME: iPhone 14
-    runs-on: ${{ matrix.variant.runner }}
+  node-electron-tests:
+    name: Test ${{ matrix.environment }} on ${{ matrix.runner }} (${{matrix.script_name}})
+    needs:
+      - generate-jsi
```
kneth, line 381: Why is JSI needed for node? Shouldn't it be `node-prebuild`?

kraenhansen: Great catch - `generate-jsi` is not needed for the Node / Electron tests 👍

elle-j (approved):
This will be very helpful!

PR #6753, opened 2024-06-24
RJS-2846: Fix path configuration when using sync

gagik:
## What, How & Why?
A bug was introduced in 12.8.0 where a custom path was being ignored when using a sync configuration. This fixes it and introduces a new test to track it.

This closes https://github.com/realm/realm-js/issues/6754

## ☑️ ToDos

* [X] 📝 Changelog entry
* [X]~ 📝 `Compatibility` label is updated or copied from previous entry~
* [ ] ~📝 Update `COMPATIBILITY.md`~
* [X] 🚦 Tests
* [ ]~ 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)~
* [ ] ~📱 Check the React Native/other sample apps work if necessary~

### DIFF
--- a/integration-tests/tests/src/tests/sync/flexible.ts
+++ b/integration-tests/tests/src/tests/sync/flexible.ts
@@ -411,7 +411,7 @@ describe("Flexible sync", function () {
             closeRealm(realm2);
           });
 
-          it("sets the correct path from the configuration", async function () {
+          it("sets custom path from the configuration", async function () {
             const realm = await Realm.open({
               path: "custom_path.realm",
               ...getSuccessConfig(this.user, {}),
--- a/integration-tests/tests/src/tests/sync/flexible.ts
+++ b/integration-tests/tests/src/tests/sync/flexible.ts
@@ -411,6 +411,15 @@ describe("Flexible sync", function () {
             closeRealm(realm2);
           });
 
+          it("sets the correct path from the configuration", async function () {
+            const realm = await Realm.open({
+              path: "custom_path.realm.realm",
+              ...getSuccessConfig(this.user, {}),
+            });
+
+            expect(realm.path).to.contain("custom_path.realm");
+          });
+
           it("does not update the subscriptions on second open if rerunOnOpen is false", async function (this: RealmContext) {
             const realm = await testSuccess(this.user, { rerunOnOpen: false }, false);
             if (!realm) throw new Error("Valid realm was not returned from testSuccess");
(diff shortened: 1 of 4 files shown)


integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -411,6 +411,15 @@ describe("Flexible sync", function () {
             closeRealm(realm2);
           });
 
+          it("sets the correct path from the configuration", async function () {
+            const realm = await Realm.open({
+              path: "custom_path.realm.realm",
```
kraenhansen, line 416:
```suggestion
              path: "custom_path.realm",
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -411,6 +411,15 @@ describe("Flexible sync", function () {
             closeRealm(realm2);
           });
 
+          it("sets the correct path from the configuration", async function () {
```
elle-j, line 414: We could also move this to another file, e.g. [sync/realm.ts](https://github.com/realm/realm-js/blob/main/integration-tests/tests/src/tests/sync/realm.ts), as this one relates to flexible sync specific tests.

elle-j, line 414: Alternative
```suggestion
          it("sets custom path from the configuration", async function () {
```

integration-tests/tests/src/tests/sync/flexible.ts
```diff
@@ -411,6 +411,15 @@ describe("Flexible sync", function () {
...
+            const realm = await Realm.open({
+              path: "custom_path.realm",
+              ...getSuccessConfig(this.user, {}),
+            });
+
+            expect(realm.path).to.contain("custom_path.realm");
```
elle-j, line 420: For an exact string to match, see e.g. the [local realm test](https://github.com/realm/realm-js/blob/1b24533db033d1cae13f2da62c3cb081cddb44e6/integration-tests/tests/src/tests/realm-constructor.ts#L52) 👍

elle-j (approved):
Good catch!

integration-tests/tests/src/tests/sync/flexible.ts
elle-j: I believe it incorporates `user.id` if I'm not misremembering.

gagik: Got it now!

kneth (approved):

PR #6784, opened 2024-07-10
Test collection and object listeners for updates on links

elle-j:
## What, How & Why?

Adds more tests for collection and object listeners for updates on links to track the difference in behavior.


packages/realm/src/Collection.ts
```diff
@@ -158,6 +158,9 @@ export abstract class Collection<
    * @note `deletions` and `oldModifications` report the indices in the collection before the change happened,
    * while `insertions` and `newModifications` report the indices into the new version of the collection.
    * @throws A {@link TypeAssertionError} if `callback` is not a function.
+   * @note
```
kneth, line 161: Good explanation of why it is called initially!

kneth (approved):

Issue #6233, opened 2023-11-04
Better Multi React DOM support


aight8:
### Problem
I use react-native-navigation, which uses many React DOM's (every Screen is a separate DOM)
Every screen shares a common Wrapper, which initializes Realm, App and optionally the User Context.
The problem is that:
1) AppProvider and RealmProvider creates and configure a fresh instance of Realm.App and Realm object for the DOM. I just want to initialize and reuse the global realm and realm.app object externally. Also the observing doesn't work like this, I don't know exactly why, but I think that the global realm and realm.app object only corresponds to the last changed screen.
2) I have tried to simply set Realm and Realm.App through the Context.Provider but they are not exported. Also, the AppProvider contains some extra logic (also renders AuthOperationProvider).

```jsx
<AppProvider
	appRef={appRef}
	id={'XXXXX'}
>
	<ConditionalWrap
		condition={userRequired}
		wrap={(children) => (
			<UserProvider fallback={<Text>UserProvider fallback</Text>}>
				{children}
			</UserProvider>
		)}
	>
		<RealmProvider
			realmRef={realmRef}
			closeOnUnmount={false}
			deleteRealmIfMigrationNeeded={true}
			schema={[Model]}
		>
			{children}
		</RealmProvider>
	</ConditionalWrap>
</AppProvider>
```

### Solution
```ts
const realmApp = new Realm.App({})
const realm = new Realm({})

// Here I can setup the single instance:

realmApp.addEventListener(() => {
    // something
})
realmApp.currentUser?.addEventListener(() => {
    // change Screen (e.g. login screen)
})
```

```jsx
<AppProvider
	...
	appInstance={realmApp}
	...
>
	<ConditionalWrap
		condition={userRequired}
		wrap={(children) => (
			<UserProvider fallback={<Text>UserProvider fallback</Text>}>
				{children}
			</UserProvider>
		)}
	>
		<RealmProvider
			...
			realmInstance={realm} // explicitly also closeOnUnmount
			...
		>
			{children}
		</RealmProvider>
	</ConditionalWrap>
</AppProvider>
```

As result the Realm and Realm.App Object is managed externally and as side effect the code React DOM stays little bit cleaner.

### Alternatives
There is no real alternatives. I currently forked realm-js, exported the React Context.Provider, manually add AuthOperationProvider etc.

### How important is this improvement for you?
Dealbreaker

### Feature would mainly be used with
Atlas Device Sync

kneth:
[...]

PR #6788, opened 2024-07-11
RJS-2861: Deprecate AppConfiguration.app and add app instance to App Provider

gagik:
@realm/devdocs 
## What, How & Why?
Allow the client to use an existing `Realm.App` in `@realm/react`. Helps support usecases like #6233 better.

This closes https://github.com/realm/realm-js/issues/6785

## ☑️ ToDos

* [X] 📝 Changelog entry
* [X] 🚦 Tests
* [X] 🔔 Mention `@realm/devdocs` if documentation changes are needed

### DIFF
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -103,7 +103,7 @@ export function AppProvider({ children, app, ...appWithConfigurationProps }: Dyn
       throw new Error("Cannot use configuration props when using an existing App instance.");
     }
     return <AppProviderWithApp app={app}>{children}</AppProviderWithApp>;
-  } else if (app != null) {
+  } else if (typeof app === "object") {
     throw new Error(
       `The "app" prop is used to use an existing Realm.App instance with an AppProvider. Either remove it or pass a valid Realm.App.`,
     );
@@ -127,9 +127,9 @@ function AppProviderWithApp({ app, children }: React.PropsWithChildren<AppProvid
 function AppProviderWithConfiguration({
   appRef,
   children,
-  ...configurationProps
+  ...config
 }: React.PropsWithChildren<AppProviderWithConfigurationProps>) {
-  const configuration = useRef<Realm.AppConfiguration>(configurationProps);
+  const configRef = useRef<Realm.AppConfiguration>(config);
 
   const [app, setApp] = useState<Realm.App>(() => new Realm.App(configuration.current));
 
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -104,7 +104,7 @@ export function AppProvider({
 }: DynamicAppProviderProps): React.ReactNode {
   if (appInstance != null) {
     if (Object.keys(appWithConfigurationProps).length > 0) {
-      throw new Error("Cannot use configuration props when using an existing App.");
+      throw new Error("Cannot use configuration props when using an existing App instance.");
     }
 
     return <AppProviderWithApp appInstance={appInstance}>{children}</AppProviderWithApp>;
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -104,20 +104,20 @@ export function AppProvider(props: DynamicAppProviderWithConfigurationProps): Re
  * @param app - The {@link Realm.App} for the provider.
  */
 export function AppProvider(props: DynamicAppProviderWithAppProps): React.ReactNode;
-export function AppProvider({ children, app, ...appWithConfigurationProps }: DynamicAppProviderProps): React.ReactNode {
+export function AppProvider({ children, app, ...config }: DynamicAppProviderProps): React.ReactNode {
   if (app instanceof App) {
-    if (Object.keys(appWithConfigurationProps).length > 0) {
+    if (Object.keys(config).length > 0) {
       throw new Error("Cannot use configuration props when using an existing App instance.");
     }
     return <AppProviderWithApp app={app}>{children}</AppProviderWithApp>;
-  } else if (typeof app === "object") {
+  } else if (typeof app !== "undefined") {
     throw new Error(
-      `The "app" prop is used to use an existing Realm.App instance with an AppProvider. Either remove it or pass a valid Realm.App.`,
+      `The "app" prop is used to pass an existing Realm.App instance into an AppProvider. Either remove it or pass a valid Realm.App.`,
     );
   }
 
   return (
-    <AppProviderWithConfiguration {...(appWithConfigurationProps as AppProviderWithConfigurationProps)}>
+    <AppProviderWithConfiguration {...(config as AppProviderWithConfigurationProps)}>
       {children}
     </AppProviderWithConfiguration>
   );
--- a/packages/realm-react/src/AppProvider.tsx
+++ b/packages/realm-react/src/AppProvider.tsx
@@ -78,9 +78,8 @@ type DynamicAppProviderWithConfigurationProps = RestrictivePick<
 >;
 
 /**
- * Props for the AppProvider component. You can either pass an existing app through the `app` prop
- * or props that replicate the configuration that is used to create a Realm.App instance:
- * https://www.mongodb.com/docs/realm-sdks/js/latest/Realm.App.html#~AppConfiguration
+ * Props for the AppProvider component. You can either pass an existing {@link Realm.App} through the `app` prop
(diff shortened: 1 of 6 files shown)


packages/realm-react/src/RealmProvider.tsx
```diff
@@ -88,7 +88,7 @@ export type DynamicsRealmProviderWithConfigurationProps = RestrictivePick<
  * Supports either {@link RealmProviderRealmProps} or {@link RealmProviderConfigurationProps}.
  */
 export type DynamicRealmProvider = React.FC<
```
gagik, line 90: I decided to apply the naming scheme of RealmProvider for consistency

packages/realm/src/app-services/App.ts
```diff
@@ -286,6 +288,15 @@ export class App<
       assert.function(fetch, "fetch");
     }
 
+    // TODO: Delete this warning once the app field has been removed from the SDK.
+    if (app !== undefined) {
+      internal.defaultLogger({
```
gagik, line 293: I figured this is better than a normal `console.warn` but at the same time, this may be redundant to do, the app is already being marked as deprecated by TypeScript so potentially they could get this warning indirectly. Unless we have JS users...

packages/realm-react/src/AppProvider.tsx
```diff
@@ -62,20 +58,84 @@ type AppProviderProps = Realm.AppConfiguration & {
...
-export const AppProvider: React.FC<AppProviderProps> = ({ children, appRef, ...appProps }) => {
-  const configuration = useRef<Realm.AppConfiguration>(appProps);
+export function AppProvider(props: DynamicAppProviderWithConfigurationProps): React.ReactNode;
+/**
+ * React component providing a Realm App instance on the context for the
+ * sync hooks to use. An `AppProvider` is required for an app to use the hooks.
+ * @param appInstance - The {@link Realm.App} for the provider.
+ */
+export function AppProvider(props: DynamicAppProviderWithAppProps): React.ReactNode;
+export function AppProvider({ children, app, ...appWithConfigurationProps }: DynamicAppProviderProps): React.ReactNode {
```
kraenhansen, line 100: The `appWithConfigurationProps` name seems a bit verbose to me, could we simplify and just call it `config`?

packages/realm-react/src/AppProvider.tsx
```diff
@@ -62,20 +58,84 @@ type AppProviderProps = Realm.AppConfiguration & {
...
+  if (app instanceof App) {
+    if (Object.keys(appWithConfigurationProps).length > 0) {
+      throw new Error("Cannot use configuration props when using an existing App instance.");
+    }
+    return <AppProviderWithApp app={app}>{children}</AppProviderWithApp>;
+  } else if (app != null) {
```
kraenhansen, line 106: I'd like for us to avoid loose (not) equality.

```suggestion
  } else if (typeof app === "object") {
```

packages/realm-react/src/AppProvider.tsx
```diff
@@ -62,20 +58,84 @@ type AppProviderProps = Realm.AppConfiguration & {
...
+    <AppContext.Provider value={app}>
+      <AuthOperationProvider>{children}</AuthOperationProvider>
+    </AppContext.Provider>
+  );
+}
+
+function AppProviderWithConfiguration({
+  appRef,
+  children,
+  ...configurationProps
```
kraenhansen, line 130: By similar logic as the comment above ☝️ 
```suggestion
  ...config
```

packages/realm-react/src/AppProvider.tsx
```diff
@@ -62,20 +58,84 @@ type AppProviderProps = Realm.AppConfiguration & {
...
+}: React.PropsWithChildren<AppProviderWithConfigurationProps>) {
+  const configuration = useRef<Realm.AppConfiguration>(configurationProps);
```
kraenhansen, line 132: I believe it's fairly common practice to suffix these with `Ref`.

```suggestion
  const configRef = useRef<Realm.AppConfiguration>(configurationProps);
```

packages/realm-react/src/__tests__/AppProvider.test.tsx
```diff
@@ -42,72 +43,146 @@ describe("AppProvider", () => {
...
+
+      expect(newSchemaNameContainer).toHaveTextContent(differentAppInstanceId);
     });
+  });
 
-    const appIdText = await waitFor(() => getByTestId("appIdText"));
+  it("throws an error if the app is not a Realm.App", () => {
+    expect(() =>
+      render(
+        // @ts-expect-error The app prop type should be Realm.App
```
kraenhansen, line 170: Great use of `@ts-expect-error` 👍

packages/realm/src/app-services/App.ts
```diff
@@ -157,6 +158,7 @@ export type AppConfiguration = {
 };
 
 /**
+ * @deprecated LocalAppConfiguration is no longer used by Atlas Device Sync. It will be removed in future SDK releases and should not be used.
```
kraenhansen, line 161: No need to mention the symbol in its documentation.

```suggestion
 * @deprecated No longer used by Atlas Device Sync. It will be removed in future SDK releases and should not be used.
```

packages/realm/src/app-services/App.ts
```diff
@@ -119,6 +119,7 @@ export type AppConfiguration = {
   baseUrl?: string;
 
   /**
+   * @deprecated LocalAppConfiguration is no longer used by Atlas Device Sync. It will be removed from in future SDK releases.
```
kraenhansen, line 122: No need to mention the symbol in its documentation.

```suggestion
   * @deprecated No longer used by Atlas Device Sync. It will be removed from in future SDK releases.
```

packages/realm/src/app-services/App.ts
kraenhansen: > Unless we have JS users...

I'm fairly certain we do. At least using Node.js.

packages/realm-react/src/AppProvider.tsx
```diff
@@ -48,12 +49,7 @@ const AuthOperationProvider: React.FC<AuthOperationProps> = ({ children }) => {
   );
 };
 
-/**
- * Props for the AppProvider component. These replicate the options which
- * can be used to create a Realm.App instance:
- * https://www.mongodb.com/docs/realm-sdks/js/latest/Realm.App.html#~AppConfiguration
- */
```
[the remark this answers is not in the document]
gagik: It was moved down to the dynamic provider but I will add comments to these inidividual ones too

packages/realm-react/src/AppProvider.tsx
gagik: in this context, isn't more loose what we want? This kind of assumes that maybe they use JavaScript so ended up passing anything, so could be even not an object, I'd generally not have this but since we have this breaking change and until now they could have gotten away with anything in the app field, seemed necessary. Otherwise object is actually disallowed by TypeScript. I could do `!app`.

packages/realm/src/app-services/App.ts
gagik: Yeah I did find that utility but did seem a bit abandoned. I guess might be best to just do `console.warn` it's just that we'd potentially be console logging regardless of the development environment which seems a bit hostile but I guess not a big deal for this very niche scenario.

packages/realm-react/src/AppProvider.tsx
```diff
@@ -62,23 +59,93 @@ type AppProviderProps = Realm.AppConfiguration & {
...
+ * @param app - The {@link Realm.App} for the provider.
+ */
+export function AppProvider(props: DynamicAppProviderWithAppProps): React.ReactNode;
+export function AppProvider({ children, app, ...appWithConfigurationProps }: DynamicAppProviderProps): React.ReactNode {
+  if (app instanceof App) {
+    if (Object.keys(appWithConfigurationProps).length > 0) {
+      throw new Error("Cannot use configuration props when using an existing App instance.");
+    }
+    return <AppProviderWithApp app={app}>{children}</AppProviderWithApp>;
+  } else if (typeof app === "object") {
```
kraenhansen, line 113: Yeah - either this or expect that it's `undefined`.

```suggestion
  } else if (typeof app !== "undefined") {
```

kraenhansen (approved):

packages/realm-react/src/AppProvider.tsx
gagik: sounds good, we should probably make strict inequality a lint rule

kneth (approved):

PR #6789, opened 2024-07-15
fix(toJSON): get the generic type T for toJSON response

### LINKED ISSUE #5165: Generic type for toJSON
### Problem

In our application, we're using .toJSON a lot because we work with immutability and so we copy the data by using js spreading operation and when the spreading takes place the data is not available anymore, toJSON solves that problem. I wasn't seeing any problem with using toJSON for that matter, but after upgrading Realm to a newer version, I noticed that the type is returning something different `Record<string, unknown>`. Is there a way I could type it without having to force the type? The cast doesn't work here because they are a lot different.

image

from Realm types folder
image

current realm version:    11.0.0
current react native version:   0.70.4
current typescript version:  4.8.3

### Solution

It would be better if the type was a generic instead of Record<string,unknown>

Would be a major improvement

### LINKED ISSUE #6728: Inaccurate Return Type for `toJSON` Method in `RealmObject`
I've noticed a potential issue with the `toJSON` method's return type in the `RealmObject` class. According to the current TypeScript definition, the `toJSON` method is expected to return a `DefaultObject`. However, considering the generic type `T` used throughout the `RealmObject<T>` class, it seems more appropriate for `toJSON` to return a type `T` instead, especially since the method processes and returns a plain object representation of the instance.

### Current Implementation:
```typescript
export class RealmObject<T = DefaultObject, RequiredProperties extends keyof OmittedRealmTypes<T> = never> {

  // ...

  toJSON(_?: string, cache?: unknown): DefaultObject;
  toJSON(_?: string, cache = new JSONCacheMap()): DefaultObject {
    // ...
    const result: DefaultObject = {};
    // Implementation details...
    return result;
  }

  // ...
}
  
```

### Source:
[Object.ts#L383-L410](https://github.com/realm/realm-js/blob/4fca13e3248e35cd17c818ece93b47710bcc8bb7/packages/realm/src/Object.ts#L383-L410)

### Documentation Reference:
[RealmObject.toJSON](https://www.mongodb.com/docs/realm-sdks/js/latest/classes/Object.html#toJSON)

### Suggested Change:
It might be more accurate for the `toJSON` method to return a type `T` to reflect the actual data structure of the `RealmObject` instance. This change would align the method's return type with the generic type used in the class definition.

Could you please review this and consider updating the TypeScript definition if appropriate? Thank you for looking into this matter.

vinibgoulart:
## What, How & Why?
fixing the toJSON generic type return

closes #6728
closes #5165

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/src/Object.ts
+++ b/packages/realm/src/Object.ts
@@ -382,9 +382,9 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
    * and [flatted](https://www.npmjs.com/package/flatted) to stringify Realm entities that have circular structures.
    * @returns A plain object.
    */
-  toJSON(_?: string, cache?: unknown): DefaultObject;
+  toJSON(_?: string, cache?: unknown): T | Partial<T>;
   /** @internal */
-  toJSON(_?: string, cache = new JSONCacheMap()): DefaultObject {
+  toJSON(_?: string, cache = new JSONCacheMap<Partial<T>>()): T | Partial<T> {
     // Construct a reference-id of table-name & primaryKey if it exists, or fall back to objectId.
 
     // Check if current objectId has already processed, to keep object references the same.
@@ -392,7 +392,7 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
     if (existing) {
       return existing;
     }
-    const result: DefaultObject = {};
+    const result: Partial<T> = {};
     cache.add(this, result);
     // Move all enumerable keys to result, triggering any specific toJSON implementation in the process.
     for (const key in this) {
@@ -402,7 +402,7 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
       }
       if (value instanceof RealmObject || value instanceof OrderedCollection || value instanceof Dictionary) {
         // recursively trigger `toJSON` for Realm instances with the same cache.
-        result[key] = value.toJSON(key, cache);
+        result[key] = value.toJSON(key, cache)
       } else {
         // Other cases, including null and undefined.
         result[key] = value;
--- a/packages/realm/src/JSONCacheMap.ts
+++ b/packages/realm/src/JSONCacheMap.ts
@@ -19,8 +19,8 @@
 import { DefaultObject, INTERNAL, RealmObject } from "./internal";
 
 /** @internal */
-export class JSONCacheMap extends Map<number, Map<string, DefaultObject>> {
-  add<T>(object: RealmObject<T>, value: DefaultObject) {
+export class JSONCacheMap<T = DefaultObject> extends Map<number, Map<string, T>> {
+  add<U>(object: RealmObject<U>, value: T) {
     const tableKey = object[INTERNAL].table.key;
     let cachedMap = this.get(tableKey);
     if (!cachedMap) {
@@ -29,7 +29,7 @@ export class JSONCacheMap extends Map<number, Map<string, DefaultObject>> {
     }
     cachedMap.set(object._objectKey(), value);
   }
-  find<T>(object: RealmObject<T>) {
+  find<U>(object: RealmObject<U>) {
     return this.get(object[INTERNAL].table.key)?.get(object._objectKey());
   }
 }


vinibgoulart:
@cla-bot check

packages/realm/src/Object.ts
```diff
@@ -382,17 +382,17 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
    * and [flatted](https://www.npmjs.com/package/flatted) to stringify Realm entities that have circular structures.
    * @returns A plain object.
    */
-  toJSON(_?: string, cache?: unknown): DefaultObject;
+  toJSON(_?: string, cache?: unknown): T | Partial<T>;
```
kraenhansen, line 385: Why the `Partial<T>`?

Also, a property of type `Realm.List` is expected to return an array of values, in which case, using the `T` directly would be incorrect.

vinibgoulart: You're right, we need to mapping the array case. I'll check how can I improve it.

About the partial, when we are creating and mapping the values to create the object we have a partial of T, and we save in the cache, if some we have some break in the flow the cache will still be there (I assume) and the fn will return the partial of T, this is the mutabillity problem. Should I consider that we will never break and always return a complete object?

kraenhansen: > we will never break and always return a complete object?

I not 100% sure I follow. But what I think you're referring to is the fact that the object graph might contain cycles.
As I understand the `toJSON` semantics, it needs to transform objects into JSON serializable values, which might contain cycles and that's exactly what the implementation tries to do. I.e. if two objects refer to each other the cache would be used to detect the cycle and the resulting object would be two plain objects cyclicly referring to each other.

CacaoRick: I think should return `OmittedRealmTypes<T>`

In document https://www.mongodb.com/docs/atlas/device-sdks/sdk/react-native/model-data/define-a-realm-object-model/

```javascript
class Book extends Realm.Object<Book> {
  name!: string;
  price?: number;

    name: 'Book',
    properties: {
      name: {type: 'string', indexed: true},
      price: 'int?',
    },
  };
}
```

`T` is `Book` (extends `Realm.Object`) so `T` has `Realm.Object`'s methods, properties

but `toJSON()` remove those methods, properties like `OmittedRealmTypes<T>`

I patched it locally to `toJSON(_?: string, cache?: unknown): OmittedRealmTypes<T>;` and didn't encounter any issues while using it.

packages/realm/src/Object.ts
```diff
@@ -382,9 +382,9 @@ export class RealmObject<T = DefaultObject, RequiredProperties extends keyof Omi
    * and [flatted](https://www.npmjs.com/package/flatted) to stringify Realm entities that have circular structures.
    * @returns A plain object.
    */
-  toJSON(_?: string, cache?: unknown): OmittedRealmTypes<T>;
+  toJSON(_?: string, cache?: unknown): OmittedRealmTypes<T | Unmanaged<T>>;
   /** @internal */
-  toJSON(_?: string, cache = new JSONCacheMap<T>()): OmittedRealmTypes<T> {
+  toJSON(_?: string, cache = new JSONCacheMap<T>()): OmittedRealmTypes<T | Unmanaged<T>> {
```
CacaoRick, line 387: Thank you for your commit.

`Unmanaged<T>` will remap all properties to optional.

If a property is not optional in its `Realm.Object` schema, instances created by that Object will certainly have that property and it will not be optional.

Will this cause the object returned by `toJSON()` to need to be remapped with type definitions in order to be used correctly?

Do you believe OmittedRealmTypes<T> is sufficient?

CacaoRick (approved):

tehvgg:
Is there potential for allowing a generic to override the return type? My objects are based on models we've exported from my client's swagger and converted to TS interfaces, so I already have a type I'd like to use as a `toJSON()` return

```typescript
toJSON<R = OmittedRealmTypes<T>>= (_?: string, cache = new JSONCacheMap<T>()): R;

const json = obj.toJSON<MyModel>();
```

currently I can't just override `toJSON` and change the return type because it's technically not compatible with the `DefaultObject` on the original, and I suspect I'd have cases where it also wouldn't match the new return type proposed here, even if they would be compatible enough for my purposes

```typescript
toJSON(...args: Parameters<Realm.Object<MyRealmObject>['toJSON']>): MyModel {
  return super.toJSON(...args) as ReturnType<MyRealmObject['toJSON']>;
}
```

Issue #6787, opened 2024-07-11
Gradle lint error with 12.10.0 and yarn errors with 12.11.1


tmac24utm:
### How frequently does the bug occur?

Always

### Description

Let me preface by saying I assume this must be an issue with my setup, or there would be more reports. However, I can't seem to get by the lint error and it appears to be the patch for this that causes the other errors.

<details>
<summary>package.json</summary>

```
{
  "name": "project-name",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "node ./scripts/android.mjs",
    "ios": "node ./scripts/ios.mjs",
    "start": "react-native start",
    "test": "yarn lint && yarn build && jest --ci --reporters=default --reporters=jest-junit",
    "test:coverage": "jest --coverage",
    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
    "build": "tsc --build tsconfig.json",
    "lintfix": "eslint . --ext .js,.ts,.tsx --fix",
    "install-ios": "bundle install && yarn pod-install"
  },
  "dependencies": {
    "@formatjs/intl-getcanonicallocales": "2.3.0",
    "@formatjs/intl-locale": "4.0.0",
    "@formatjs/intl-numberformat": "8.10.3",
    "@formatjs/intl-pluralrules": "5.2.14",
    "@formatjs/intl-relativetimeformat": "11.2.14",
    "@native-html/table-plugin": "5.3.1",
    "@notifee/react-native": "7.8.2",
    "@react-native-async-storage/async-storage": "1.23.1",
    "@react-native-community/blur": "4.4.0",
    "@react-native-community/datetimepicker": "8.1.1",
    "@react-native-community/geolocation": "3.2.1",
    "@react-native-community/netinfo": "11.3.2",
    "@react-native-firebase/app": "20.1.0",
    "@react-native-firebase/firestore": "20.1.0",
    "@react-native-firebase/messaging": "20.1.0",
    "@react-native-picker/picker": "2.7.7",
    "@react-navigation/drawer": "6.7.0",
    "@react-navigation/material-top-tabs": "6.6.13",
    "@react-navigation/native": "6.1.17",
    "@react-navigation/stack": "6.4.0",
    "@rneui/base": "4.0.0-rc.8",
    "@rneui/themed": "4.0.0-rc.8",
    "@semantic-release/commit-analyzer": "13.0.0",
    "@semantic-release/release-notes-generator": "14.0.1",
[...]

PR #6790, opened 2024-07-16
Adding lint of the Android code to be built on developers machine

kraenhansen:
## What, How & Why?

This would have caught #6787.


integration-tests/environments/react-native-test-app/package.json
```diff
@@ -127,6 +128,12 @@
       "dependencies": [
         "../../../packages/realm/:bindgen:jsi"
       ]
+    },
+    "realm:lint-android": {
+      "command": "cd android && ./gradlew realm:lint",
+      "dependencies": [
+        "../../../packages/realm/:bindgen:jsi"
```
kraenhansen, line 135: Need to double-check that this dependency is actually needed as it slows down the execution quite significantly 🤔

kraenhansen:
Drafting this as it needs an update to build the Android prebuilds firs. To me this adds too much runtime and complexity to the workflow for the benefit it gives.

kraenhansen:

Issue #6772, opened 2024-06-28
Refactor throwing uncaught exceptions from callbacks dispatched onto the event loop from C++


kraenhansen:
We should find a more general approach (ideally driven by the binding generator templates) to throw uncaught exceptions when JS callbacks dispatched on the event loop from C++, throws. We're currently wrapping some of our listener APIs in a pattern that catch errors from callback and use `setImmediate` to rethrow from a new task scheduled from JS, but we have many callbacks that doesn't have this manual "workaround".

I've pushed https://github.com/realm/realm-js/tree/kh/fix-throwing-callbacks as a starting point, disabling the workaround and adding a few failing tests that are now causing aborts.

My current thinking is that we should implement platform specific "EventLoopDispatchers" wrapping Core's `util::EventLoopDispatcher` used here: https://github.com/realm/realm-js/blob/f8aefa20ebd03a9178b79ea8fa0ed72bf206696e/packages/realm/bindgen/src/templates/node.ts#L588
Our platform dispatcher should wrap the lambda with a try-catch which reports the error as an uncaught exception using platform specific runtime APIs.

PR #4725, opened 2022-07-13
Flush RN UI queue on Android

tomduncalf:
## What, How & Why?

This makes the tests run (almost) reliably on Android. I get 5 failures but for seemingly unrelated reasons. Previously many of the roundtrip tests would fail at random. 

Note: it seems that the JNI header generator default has changed `jclass` to `jobject`. I just went with it but I'm not sure if this might cause issues!

Fixes #4389 for Android (already fixed on iOS)

## ☑️ ToDos

* [x] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 🚦 Tests
* [x] 🔀 Executed flexible sync tests locally if modifying flexible sync
* [ ] 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)
* [ ] 📱 Check the React Native/other sample apps work if necessary
* [ ] 📝 Public documentation PR created or is not necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [ ] typescript definitions file is updated
* [ ] jsdoc files updated
* [ ] Chrome debug API is updated if API is available on React Native

PR #4389, opened 2022-03-02
TTd/fix rn queue issue invoke async

tomduncalf:
## What, How & Why?

This closes # ???

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 🚦 Tests
* [ ] 📱 Check the React Native/other sample apps work if necessary
* [ ] 📝 Public documentation PR created or is not necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary

*If this PR adds or changes public API's:*
* [ ] typescript definitions file is updated
* [ ] jsdoc files updated
* [ ] Chrome debug API is updated if API is available on React Native

PR #6791, opened 2024-07-16
RJS-2852: Adding a `CallInvoker`-based scheduler for Core on React Native

### LINKED ISSUE #2655: Change listeners on Results should always update React Native UI

kraenhansen:
## What, How & Why?

This implements a `realm::util::Scheduler` for the React Native platform through the React Native [CallInvoker](https://github.com/facebook/react-native/blob/5bb57eb2b3884ef0c60551ee80880d95814eb71c/packages/react-native/ReactCommon/callinvoker/ReactCommon/CallInvoker.h#L27) API.

It's my current understanding that all active implementations of this API that a React Native app will construct, utilize the [RuntimeSchedulerCallInvoker](https://github.com/facebook/react-native/blob/5bb57eb2b3884ef0c60551ee80880d95814eb71c/packages/react-native/ReactCommon/react/renderer/runtimescheduler/RuntimeSchedulerCallInvoker.cpp) which in turn drop into either [RuntimeScheduler_Legacy](https://github.com/facebook/react-native/blob/5bb57eb2b3884ef0c60551ee80880d95814eb71c/packages/react-native/ReactCommon/react/renderer/runtimescheduler/RuntimeScheduler_Legacy.cpp) or [RuntimeScheduler_Modern](https://github.com/facebook/react-native/blob/5bb57eb2b3884ef0c60551ee80880d95814eb71c/packages/react-native/ReactCommon/react/renderer/runtimescheduler/RuntimeScheduler_Modern.cpp), both of which have invariants that we want when Core schedules calling of listeners:
1. Running tasks using a [RuntimeExecutor](https://github.com/facebook/react-native/blob/5bb57eb2b3884ef0c60551ee80880d95814eb71c/packages/react-native/ReactCommon/runtimeexecutor/ReactCommon/RuntimeExecutor.h#L17-L26) (ensuring the function will be called when it is safe to do so)
2. Draining microtasks after execution (effectively solving https://github.com/facebook/react-native/issues/33006)
3. Reporting uncaught `JSError` exceptions via `handleJSError` (effectively solving https://github.com/realm/realm-js/issues/6772 on React Native)

This fixes #2655 and #3571 and #6772 on React Native (iOS and Android) and replace the workaround introduced with https://github.com/realm/realm-js/pull/4389, https://github.com/realm/realm-js/pull/4725 and https://github.com/realm/realm-js/pull/5258/commits/03c30e7c35cd0726b8ae975fb4129b2bb18d993f.


packages/realm/binding/jsi/react_scheduler.cpp
```diff
@@ -0,0 +1,97 @@
...
+    void invoke(util::UniqueFunction<void()>&& func) override
+    {
+        m_js_call_invoker->invokeAsync(
+            // Using low priority to avoid blocking rendering
+            facebook::react::SchedulerPriority::LowPriority,
+            // Wrapping the func in a shared_ptr to ensure it outlives the invocation
+            [func = std::make_shared<util::UniqueFunction<void()>>(std::move(func))] {
+                (*func)();
+            });
+    }
```
tgoyne, line 68: `[ptr = func.release())] { util::UniqueFunction<void()>(ptr)(); }`

LowPriority seems suspect to me. We do want it to be below user input, but processing refreshes promptly is pretty important.

packages/realm/binding/jsi/react_scheduler.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+namespace realm::js::react_scheduler {
+
+using Scheduler = realm::util::Scheduler;
+
+// std::shared_ptr<facebook::react::CallInvoker> js_call_invoker_{};
+std::shared_ptr<Scheduler> scheduler_{};
+
+
+class ReactScheduler : public realm::util::Scheduler {
```
tgoyne, line 36: This is very minor, but this type and the global variable should be in an anonymous namespace so that the symbols are not exported.

packages/realm/binding/jsi/react_scheduler.cpp
kraenhansen: These are the available values https://github.com/facebook/react-native/blob/8d0cbbf0e6435e0a29d1d5d218b93c9eef1a2dcd/packages/react-native/ReactCommon/callinvoker/ReactCommon/SchedulerPriority.h#L14-L20

```
enum class SchedulerPriority : int {
  ImmediatePriority = 1,
  UserBlockingPriority = 2,
  NormalPriority = 3,
  LowPriority = 4,
  IdlePriority = 5,
};
```

Unfortunately not greatly documented 🤔

tgoyne: Glancing at the implementation, it looks like it always runs the highest priority waiting task unless a lower priority task has hit a timeout (250ms for UserBlocking, 5s for Normal, 10s for Low). If there's some specific Normal priority thing that we know we need to not interrupt then using Low makes sense, but otherwise I would default to Normal?

kraenhansen:
```
[ptr = func.release())] { util::UniqueFunction<void()>(ptr)(); }
```

I try to refrain from c-style castings 🤔 It seems to work, but what ensure the original function owned by `func` ever gets destructed?

packages/realm/binding/apple/RealmReactModule.mm
```diff
@@ -79,7 +79,7 @@ - (void)dealloc {
   // See https://github.com/facebook/react-native/pull/43396#issuecomment-2178586017 for context
```
kneth, line 79: Should we remove the comments?

packages/realm/binding/apple/RealmReactModule.mm
```diff
@@ -79,7 +79,7 @@ - (void)dealloc {
...
   // If it was, we could use the enablement of "microtasks" to avoid the overhead of calling the invokeAsync on every call from C++ into JS.
```
kneth, line 80: Should we remove the comments?

kneth (approved):
It is a better solution that the old workaround.

Please include an entry in `CHANGELOG.md` - maybe under "Internal" if you don't consider it a bug fix.

.watchmanconfig
```diff
@@ -3,6 +3,8 @@
     "react-native/node_modules",
     "packages/realm-web",
     "packages/realm-web-integration-tests",
-    ".wireit"
+    ".wireit",
+    "packages/realm/.wireit",
+    "integration-tests/environments/react-native-test-app/.wireit"
```
kraenhansen, line 8: I got annoyed by these being indexed by watchman.

fealebenpae (approved):

packages/realm/binding/jsi/react_scheduler.cpp
tgoyne: There's no casts involved there. That's constructing a new UniqueFunction using the impl pointer (which takes ownership of the pointer) and then immediately invoking it. The function is then destroyed at the end of the statement as it's an unnamed temporary.

kraenhansen: I don't know how I misread that as a cast 🙂 What about the case where the function is dropped and never called, that would leak the function, right? (since no new `UniqueFunction` takes ownership.

tgoyne: This relies on CallInvoker guaranteeing that it'll invoke the function it's given exactly once; more than once and it's a use-after-free and zero times is a memory leak. This is generally a very safe thing to rely on as basically everything gets way more complicated otherwise.

tgoyne (approved):

PR #6798, opened 2024-07-19
Refactored Node test environments to share more

kraenhansen:
## What, How & Why?

This is a refactor of the node test environment + the node-based dev environment (running tests without mocha-remote).


integration-tests/environments/node/index.cjs
```diff
@@ -18,25 +18,14 @@
...
-v8.setFlagsFromString("--expose_gc");
 
 const client = new Client({
   title: `Node.js v${process.versions.node} on ${os.platform()} (via CommonJS)`,
   tests(context) {
-    // Exposing the Realm constructor as a global
-    global.fs = fs;
-    global.path = path;
-    global.environment = { ...context, node: true };
-    global.gc = vm.runInNewContext("gc");
```
kraenhansen, line 35: The main idea behind this refactor is to move this into a shared "node/src/setup-globals.ts" used across these two "runners" and the test ran with bare mocha (without mocha-remote - used when debugging native code).

integration-tests/tests/src/node/index.ts
```diff
@@ -16,7 +16,14 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-// NOTE: This file is only supposed to be imported from a Node.js environment
+if (typeof process.versions.node !== "string") {
+  throw new Error("This file is only supposed to be imported from a Node.js environment!");
```
kraenhansen, line 20: Asserting that this gets imported from node instead of simply commenting the intent.

integration-tests/tests/src/setup-globals.ts
```diff
@@ -54,6 +54,10 @@ chai.use(chaiRealmObjects);
 const DEFAULT_LONG_TIMEOUT = 30 * 1000; // 30s
 
 describe("Test Harness", function (this: Mocha.Suite) {
+  // Inject the root suite title from the global
+  if (typeof title === "string" && this.parent?.root) {
+    this.parent.title = title;
+  }
```
kraenhansen, line 60: This was needed to get the title from "inject-dev-environment.ts" to propagate as expected.

gagik (approved):
LGTM

PR #6800, opened 2024-07-19
RJS-2854: Adding a `ThreadSafeFunction`-based scheduler for Core on Node.js

### LINKED ISSUE #6772: Refactor throwing uncaught exceptions from callbacks dispatched onto the event loop from C++

kraenhansen:
## What, How & Why?

This closes #6772 by implementing a `utils::Scheduler` based on the [`Napi::ThreadSafeFunction`](https://github.com/nodejs/node-addon-api/blob/main/doc/threadsafe_function.md) primitive.

Note: This depends on #6798 which I expect to merge first and then rebase this branch.

Note: In order for the uncaught exceptions thrown by the Node.js addon, the node process must be started with a `--force-node-api-uncaught-exceptions-policy` flag enabled. This was added under a flag to "not breaking existing running Node-API add-ons.".

https://github.com/nodejs/node/blob/cd39578e53d6502de03f15620d15835b86954ac8/src/node_api.cc#L85-L115

Alternatively the following warning will be logged:

> (node:17922) [DEP0168] DeprecationWarning: Uncaught N-API callback exception detected, please run node with option --force-node-api-uncaught-exceptions-policy=true to handle those exceptions properly.

## ☑️ ToDos
* [ ] Rebase when #6798 merge.
* [ ] Rebase when https://github.com/realm/realm-core/pull/7903 merge and gets released.
* [ ] Investigate passing the `enforceUncaughtExceptionPolicy` template parameter to avoid the warning when throwing uncaught exceptions.
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
I thought my implementation was done, but it turns out tests were still failing 🤔 This is likely never going to get merged.

Issue #5459, opened 2023-02-20
RealmProvider `fallback` property should include progress notifications


takameyer:
### Problem

Currently, there is no clear way to include display progress notifications when opening a synced Realm.

We already have complaints about this (see issue #5144).

### Solution

The best solution would be for the `fallback` component rendered by the `RealmProvider` to include properties for the progress notification and re-render when this changes.  This would allow more context into the `fallback` in order to determine that progress is being made in opening the `realm`.

```tsx
const RealmFallback = ({transferred, transferable}) => {...}
```

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

Atlas Device Sync

BubbleTrouble14:
Hi there. Any news on this topic ?

PR #6801, opened 2024-07-19
RJS-2187: Add progress information to RealmProvider fallback.

gagik:
@realm/devdocs 
## What, How & Why?
Solves https://github.com/realm/realm-js/issues/5459

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [X] 🚦 Tests
* [X] 🔔 Mention `@realm/devdocs` if documentation changes are needed


packages/realm-react/src/RealmProvider.tsx
```diff
@@ -158,14 +162,26 @@ export function createRealmProviderFromConfig(
...
       const initRealm = async () => {
-        const openRealm = await Realm.open(configuration.current);
+        const openRealmPromise = Realm.open(configuration.current);
+        if (configuration.current.sync?.flexible) {
+          try {
+            openRealmPromise.progress((estimate: number) => {
+              setProgress(estimate);
+            });
+          } catch (error) {
+            console.warn("Progress information with @realm/react work with realm version >=12.12.0.");
```
gagik, line 181: Not a fan of this. Should we consider bumping up the compatibility instead?

kraenhansen: That would be fine by me, especially if we signal it through the `peerDependencies`.

COMPATIBILITY.md
```diff
@@ -2,6 +2,7 @@
 
 | Realm JavaScript    | @realm/react      | React Native       | Expo     | Hermes | npm    | node   |
 |---------------------|-------------------|--------------------|----------|--------|--------|--------|
+| 12.12.0             | >= 0.10.0         | >= 0.71.4          | >= 48    | ✅     | >= 10  | >= 20  |
```
gagik, line 5: Is this how we do it?

kraenhansen: As far as I understand `realm@12.12.0` is compatible with `@realm/react` which are smaller than `0.10.0`, which is not what this row is communicating. I.e. this could be turned into "optional peer dependencies" of "realm", not the other way around.

packages/realm-react/CHANGELOG.md
```diff
@@ -4,13 +4,29 @@
 * None
 
 ### Enhancements
-* None
+* Added the ability to get `progress` information in `fallback` component of `RealmProvider` when opening a synced Realm. ([#6785](https://github.com/realm/realm-js/issues/6785))
+```tsx
+import { RealmProvider, RealmProviderFallback } from "@realm/react";
+
+const Fallback: RealmProviderFallback = ({ progress }) => {
+	return <Text>Loading: {progress}/1.0</Text>;
```
kraenhansen, line 12:
```suggestion
	return <Text>Loading: {(100 * progress).toFixed()} %</Text>;
```

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -27,6 +27,10 @@ type PartialRealmConfiguration = Omit<Partial<Realm.Configuration>, "sync"> & {
   sync?: Partial<Realm.SyncConfiguration>;
 };
 
+export type RealmFallbackComponent = React.ComponentType<{
+  progress: number | null;
```
kraenhansen, line 31: `null` is normally used for the absence of an object, perhaps `undefined` would be better here? Or just some sane number like `0` if the progress is unknown?

packages/realm-react/src/RealmProvider.tsx
```diff
@@ -158,14 +162,18 @@ export function createRealmProviderFromConfig(
       }
     }, [realm]);
 
+    const [progress, setProgress] = useState<number | null>(null);
```
kraenhansen, line 165: As per my previous comment.

```suggestion
    const [progress, setProgress] = useState<number>(0);
```

packages/realm-react/src/__tests__/mocks.ts
```diff
@@ -0,0 +1,94 @@
...
+  const progressRealmPromise = new MockedProgressRealmPromise(
+    (resolve) => {
+      setTimeout(() => resolve(new Realm()), delay);
+    },
+    {
+      progress: (callback) => {
+        if (progressValues instanceof Array) {
+          const sendProgress = () => {
+            // Uses act as this causes a component state update.
+            act(() => callback(progressValues[progressIndex]));
```
kraenhansen, line 77: I wonder if this should be moved into the calling callback instead 🤔

packages/realm-react/src/__tests__/mocks.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+export function mockRealmOpen(
+  options: {
+    /** Progress values which the `Realm.open(...).progress(...)` will receive in an equal interval. */
+    progressValues?: number[];
+    /** Duration of the Realm.open in milliseconds */
+    delay?: number;
+  } = {},
+): MockedProgressRealmPromise {
+  const { progressValues, delay = 100 } = options;
```
kraenhansen, line 65: Could this be a bit faster and still reliable?
```suggestion
  const { progressValues, delay = 10 } = options;
```

packages/realm-react/src/__tests__/mocks.ts
```diff
@@ -0,0 +1,94 @@
...
+      progress: (callback) => {
+        if (progressValues instanceof Array) {
+          const sendProgress = () => {
+            // Uses act as this causes a component state update.
+            act(() => callback(progressValues[progressIndex]));
+            progressIndex++;
+
+            if (progressIndex <= progressValues.length) {
+              // Send the next progress update in equidistant time
+              setTimeout(sendProgress, delay / progressValues.length);
```
kraenhansen, line 82: This would make the last progress arrive when the realm resolves - potentially racing that. Would it make more sense to divide by `progressValues.length + 1`?
```suggestion
              setTimeout(sendProgress, delay / (progressValues.length + 1));
```

packages/realm-react/src/__tests__/mocks.ts
```diff
@@ -0,0 +1,94 @@
...
+  ) {
+    super(callback);
+    this.progressHandler = options?.progress;
+  }
+
+  get [Symbol.toStringTag]() {
+    return "MockedProgressRealmPromise";
+  }
+
+  cancel = () => this;
```
kraenhansen, line 43: Could this easily clear outstanding timeouts? Alternatively throw that it isn't implemented.

packages/realm-react/src/__tests__/mocks.ts
```diff
@@ -0,0 +1,94 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import { act } from "@testing-library/react-native";
+import { EstimateProgressNotificationCallback, ProgressRealmPromise, Realm } from "realm";
+
+/**
+ * Mocks {@link Realm.ProgressRealmPromise} with a custom
+ * promise completion and progress handler.
+ */
+export class MockedProgressRealmPromise extends Promise<Realm> implements ProgressRealmPromise {
```
kraenhansen, line 26: 🎉

packages/realm-react/src/__tests__/mocks.ts
```diff
@@ -0,0 +1,94 @@
...
+              setTimeout(sendProgress, delay / progressValues.length);
+            }
+          };
+          sendProgress();
+        }
+      },
+    },
+  );
+
+  const delayedRealmOpen = jest.spyOn(Realm, "open");
```
kraenhansen, line 91: (How) is this mock restored? Perhaps this side-effect could be moved into the test where you could also call [jest.restoreAllMocks()](https://jestjs.io/docs/jest-object#jestrestoreallmocks) to avoid this state bleeding into other tests.

packages/realm-react/src/RealmProvider.tsx
gagik: thing is depending on the type of progress the next value could either be 1 or 0, kind of hard to come up with a good default to show right away

packages/realm-react/src/__tests__/mocks.ts
gagik: I agree, seemed reliable but would be best to avoid potential flakiness

packages/realm-react/src/__tests__/mocks.ts
gagik: I have been running the cleaning but I agree it's not the most inituitive. Maybe even extending this jest with like `jest.mockRealmOpen()` could be the most consistent way to get these mock helpers in (and then intuitively `jest.restoreAllMocks()` cleans them)

packages/realm-react/src/__tests__/mocks.ts
gagik: Ah, the calling callback in this case is `RealmProvider` so I do not have much control over that sadly; I could create the option to let the user define a function wrapper but seems too much for a convenience method

packages/realm-react/src/__tests__/mocks.ts
gagik: Seems like extending jest isn't the norm as it is with extending `expect` so I guess we either can define a hook with cleanup built in (although then if we have multiple mocks we'd be running the cleanup too much) or use the function as is and just make sure to run the `restoreAllMocks`

packages/realm-react/CHANGELOG.md
```diff
@@ -4,13 +4,29 @@
...
+  );
+}
+```
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
 * None
 
 ### Compatibility
+* Realm JavaScript >= v12.12.0
```
kneth, line 29: Should we add a note at the top (under `## vNext (TBD)`) about the bump of the Realm JS version?

kneth (approved):

packages/realm-react/CHANGELOG.md
gagik: Will merge this and change changelog in the useProgress hook PR

PR #6812, opened 2024-07-24
RJS-2870: Remove an unused header from the iOS and unused import from Android binding

### LINKED ISSUE #6808: Realm does not build on iOS with `use_frameworks! :linkage => :static` as of version 12.12.0

Always

### Description

Our React Native app fails to build when `use_frameworks! :linkage => :static` is set in the Podfile. This issue started occurring after upgrading to Realm JS v12.12.0. Given that this configuration was working in previous versions, we expected it to continue working after a minor version upgrade.

I believe the issue was introduced by #6791.

This issue is critical for our project as we rely on react-native-firebase, which requires the use_frameworks! :linkage => :static configuration.

### Stacktrace & log output

```shell
❌  (/Users/gvanderclay/workspace/my-app/node_modules/realm/binding/apple/RealmReactModule.mm:24:9)

  22 | #import <React/RCTInvalidating.h>
  23 | #import <ReactCommon/CallInvoker.h>
> 24 | #import <React-featureflags/react/featureflags/ReactNativeFeatureFlags.h>
     |         ^ 'React-featureflags/react/featureflags/ReactNativeFeatureFlags.h' file not found
  25 | #import <jsi/jsi.h>
  26 |
  27 | #import <arpa/inet.h>
```

Always

### Reproduction Steps

I have created a minimal reproducible example to demonstrate the issue. The repository is available [here](https://github.com/gvanderclay/realm-useframeworks-build-error), and it includes detailed instructions on how to reproduce the build failure.

Note: The example uses Expo with expo-build-properties to set use_frameworks!, but the issue persists regardless of this context.

### Version

12.12.0

Local Database only

### Are you using encryption?

Yes

iOS only

### Build environment

_No response_

### Cocoapods version

1.15.2

kraenhansen:
## What, How & Why?

This closes #6808 by removing an import which is no longer used.
The fix was verified by the original creator of the issue here: https://github.com/realm/realm-js/issues/6808#issuecomment-2248096323

## ☑️ ToDos
* [x] 📝 Changelog entry
* [x] 🚦 Tests (by the original reporter)
* [x] 🔔 Mention `@realm/devdocs` if documentation changes are needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@
 * None
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None
+* Fixed compiler error when building binding for React Native iOS: "'React-featureflags/react/featureflags/ReactNativeFeatureFlags.h' file not found" ([#6808](https://github.com/realm/realm-js/issues/6808), since v12.12.0)
 
 ### Compatibility
 * React Native >= v0.71.4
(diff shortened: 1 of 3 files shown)


gagik (approved):

sunrise1002:
So in this file 
`node_modules/realm/binding/android/src/main/java/io/realm/react/RealmReactModule.java`
we also have to remove this line for android 🤔
`import com.facebook.react.internal.featureflags.ReactNativeFeatureFlags;`

gagik (approved):

PR #6814, opened 2024-07-25
Avoid circular imports

### LINKED ISSUE #6565: Bundling with Metro warns with "Require cycles are allowed, but can result in uninitialized values."
We used to resolve these require cycles as part of producing a rollup bundle, but these will now surface as warnings when bundling with Metro.

We could probably revisit the decision to use a `internal.ts` to control the import order and instead focus on breaking the cycles:

https://github.com/realm/realm-js/blob/1e1e48a0f906d471b4a046168ed8cd628ef2d70f/packages/realm/src/internal.ts#L21

kraenhansen:
## What, How & Why?

This closes #6565 by refactoring imports away from the "internal" pattern (which we introduced when we used Rollup to bundle into a single file) and introducing an `indirect` object which gets classes injected once they're initialized. This also untangles the recursively referencing collections accessors introduced with collections in mixed, into separate files.

This also adds a dependency in the "madge" tool enabling a new script to check for circular imports:

```
npm run check-circular-imports --workspace realm
```

Although this still fails with three cases of circular references 🤔 I need to decide how to deal with those before adding it as a command which is ran as part of our lint workflow on CI.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
@gagik I recognize these are a lot of changes, but most were pretty mechanical. I'd love to get this in soon-ish to avoid the codebase from diverging while I'm PTO 🙏

packages/realm/src/Counter.ts
```diff
@@ -84,7 +88,7 @@ export class Counter {
 
   /** @internal */
   constructor(realm: Realm, obj: binding.Obj, columnKey: binding.ColKey) {
-    if (!(realm instanceof Realm) || !(obj instanceof binding.Obj)) {
+    if (!(obj instanceof binding.Obj)) {
```
gagik, line 91: is there a reason we're removing this check?

packages/realm/src/assert.ts
```diff
@@ -138,17 +141,6 @@ assert.never = (value: never, target?: string): never => {
 
 // SDK specific
 
-assert.primaryKey = (value: unknown, target?: string): asserts value is PrimaryKey => {
```
gagik, line 141: Is this no longer needed?

packages/realm/src/collection-accessors/Dictionary.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+/** @internal */
+export function isJsOrRealmDictionary(value: unknown): value is Dictionary | Record<string, unknown> {
+  return isPOJO(value) || value instanceof indirect.Dictionary;
+}
+
+/** @internal */
+export function isPOJO(value: unknown): value is Record<string, unknown> {
```
gagik, line 135: Not sure I understand what POJO is. Plain old JavaScript object? I guess this is Java world stuff?

gagik: I'd prefer `isJsPlainObject` or at least a comment here

gagik (approved):
Nice changes!

packages/realm/src/collection-accessors/Dictionary.ts
kraenhansen: I agree, but I'd rather do that in a different PR as this is merely copying code around.

packages/realm/src/assert.ts
kraenhansen: I simplified the only calling code to no longer need it.

packages/realm/src/Counter.ts
kraenhansen: I didn't want the file to depend on `Realm` as that would introduce a circular dependency. I think this check is sufficiently strong as it's mainly guarding against misuse of the constructor, from outside of the SDK.

kneth (approved):
The `import`s are much easier to read. And I like that you have isolated the collection accessors in separate files.

PR #6819, opened 2024-07-26
RJS-2649: Add more additional overload to useQuery to allow exhaustive-deps to work

bimusiek:
## What, How & Why?
Related to https://github.com/realm/realm-js/issues/6284


kraenhansen:
Very interesting approach - I'd be happy to merge this, although this hook is turning into a bit of an overload nightmare 🙈

bimusiek:
@kraenhansen Updated. Not sure why my `npm` is changing lock to `dev: true` everywhere, but I have manually reverted it back to what is on the main.

Crossing my fingers to have this change on next release, we will finally be able to use official package rather than our fork 👍

packages/realm-react/src/useQuery.tsx
```diff
@@ -132,19 +142,47 @@ export function createUseQuery(useRealm: () => Realm): UseQueryHook {
...
+    } else if (
+      typeof typeOrOptionsOrQuery === "function" &&
+      Array.isArray(queryOrDeps) &&
+      typeof depsOrPartialOptions === "object" &&
+      depsOrPartialOptions !== null
+    ) {
       /* eslint-disable-next-line react-hooks/rules-of-hooks -- We're calling `useQuery` once in any of the brances */
-      return useQuery(typeOrOptions, Array.isArray(queryOrDeps) ? queryOrDeps : deps);
+      return useQuery(
+        { ...(depsOrPartialOptions as QuaryHookPartialOptions<T>), query: typeOrOptionsOrQuery as QueryCallback<T> },
```
kraenhansen, line 183: Would it be possible to write this using a type assert function instead? I.e. something which test the value and return `: assert arg is QueryHookPartialOptions<T>` instead of using the `as QueryHookPartialOptions<T>`?

packages/realm-react/src/useQuery.tsx
```diff
@@ -31,6 +31,8 @@ export type QueryHookOptions<T> = {
   keyPaths?: string | string[];
 };
 
+export type QuaryHookPartialOptions<T> = Omit<QueryHookOptions<T>, "query">;
```
kraenhansen, line 34: A small typo:
```suggestion
export type QueryHookPartialOptions<T> = Omit<QueryHookOptions<T>, "query">;
```

I wonder if it would be easier to declare a base type (without the `query`) and declare `QueryHookOptions<T>` as a union with that base type and `{ query: ... }`?

bimusiek:
@kraenhansen Typeguards were a good suggestion. I went further and refactored the code quite a bit to make it more readable.

kraenhansen:
Just did a quick read through ... and I like it! 💙 Thanks a lot!

bimusiek:
@kraenhansen Is there a timeline for when is this going to be merged? We would love to switch from fork to official realm react package.

kraenhansen:
This was just published as `@realm/react@v0.11.0` 👍 Thanks for the nudge 🙂

PR #6820, opened 2024-07-26
Refactored binding wrapper

kraenhansen:
## What, How & Why?

As a prerequisite for the rebase of the WASM / browser binding, I had to rework the generated code that wraps the native module and expose it in a TypeScript friendlier way than free functions. Instead of having platform specific wrappers that implement platform specific code to load the native module I suggest exporting an `injectNativeModule` function which takes the native module (from somewhere) and wraps it into the `binding` namespace / object.

This has multiple benefits:
- We'll be able to maintain the platform specific code which load the native module as part of our `src/platform` files, which enables us to leverage TypeScript to achieve type-safety.
- We'll be able to call this `injectNativeModule` once the native module has loaded (possibly asynchronously as in the case of the WASM module).
- I've implemented this as a TypeScript source file, instead of the JS + d.ts separation that the current wrappers use. This had the potential to become more cumbersome to write, but it turned out fairly simple as the native module just has `any` type at the point of injection. This has the added benefit that we can get rid of a lot of the complexity in the `./binding` directory as well as the babel transpile step require to get a CommonJS variant of the wrappers.

## ☑️ ToDos
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests


kraenhansen:
I rebased this on `main` as #6814 was merged.

packages/realm/bindgen/src/templates/wrapper.ts
```diff
@@ -0,0 +1,405 @@
+import { TemplateContext } from "@realm/bindgen/context";
```
kraenhansen, line 1: The removal of the platform specific wrappers and the addition of this file are the biggest part of the changes proposed here. Originally we had the thought that keeping these separate could allow for platform specific optimizations but in reality we haven't relied on that and the separation gives rise to extra unneeded complexity.

kneth (approved):

packages/realm/bindgen/src/templates/wrapper.ts
```diff
@@ -0,0 +1,405 @@
...
+    );
+  }
+
+  // Check the support for template instances used
+  for (const template of Object.keys(rawSpec.templates)) {
+    assert(
+      Object.keys(TEMPLATE_MAPPING).includes(template),
+      `Spec declares an unsupported template instance: "${template}"`,
+    );
+  }
```
elle-j, line 209: Nit: Alternatively, this could be refactored into some `assertSupportedTypes()` (or the like) function (with the `rawSpec` param), and let `generate()` only focus on generating output.

packages/realm/package.json
elle-j, line 1: Nice simplification here

packages/realm/src/binding/NativeBigInt.ts
```diff
@@ -0,0 +1,40 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+import type { binding } from "./wrapper.generated";
+
+export const NativeBigInt: typeof binding.Int64 = Object.freeze({
+  add(a: binding.Int64, b: binding.Int64) {
+    return a + b;
+  },
+  equals(a: binding.Int64, b: binding.Int64) {
```
elle-j, line 25: To conform to prior behavior:
```suggestion
  equals(a: binding.Int64, b: binding.Int64 | number | string) {
```

packages/realm/src/binding/PolyfilledBigInt.ts
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import { Long } from "bson";
+
+import type { binding } from "./wrapper.generated";
+import { assert } from "../assert";
+
+export const PolyfilledBigInt: typeof binding.Int64 = Object.freeze({
```
elle-j, line 24: Interesting solution for the RN issue 🙂

packages/realm/src/binding/PolyfilledBigInt.ts
```diff
@@ -0,0 +1,48 @@
...
+  add(a: binding.Int64, b: binding.Int64) {
+    assert.instanceOf(a, Long);
+    assert.instanceOf(b, Long);
+    return a.add(b) as unknown as binding.Int64;
+  },
+  equals(a: binding.Int64, b: binding.Int64) {
+    assert.instanceOf(a, Long);
+    assert.instanceOf(b, Long);
+    return a.equals(b);
+  },
```
elle-j, line 34: If updating the previous `equals()` to allow numbers and strings, this would likely need to be updated as well, perhaps something like:
(BSON's Long `equals()` also takes numbers and strings)
```suggestion
  equals(a: binding.Int64, b: binding.Int64 | number | string) {
    assert.instanceOf(a, Long);
    assert(
      a instanceof Long || typeof a === "number" || typeof a === "string",
      "Expected a 'BSON.Long', or number, or string.",
    );
    return a.equals(b);
  },
```

packages/realm/src/binding/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+    func((...args: Args) => {
+      // Any errors in this function should flow into the Promise chain, rather than out to the caller,
+      // since callers of async callbacks aren't expecting exceptions.
+      try {
+        if (args.length < 1 || args.length > 2) throw Error("invalid args length " + args.length);
+        // The last argument is always an error
+        const error = args[args.length - 1];
+        if (error) {
+          reject(error);
+        } else if (args.length == 2) {
```
elle-j, line 37:
```suggestion
        } else if (args.length === 2) {
```

packages/realm/src/platform/react-native/binding.ts
```diff
@@ -16,7 +16,37 @@
...
+      private native: unknown;
+      constructor(obj: unknown) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- See "createWeakRef" protocol in the jsi bindgen template
+        this.native = (nativeModule as any).createWeakRef(obj);
+      }
+      deref() {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- See "createWeakRef" protocol in the jsi bindgen template
+        return (nativeModule as any).lockWeakRef(this.native);
+      }
+    },
```
elle-j, line 46: Is this also a polyfill? If so, perhaps we could declare and define it `PolyfillWeakRef` similar to the bigint polyfill.

elle-j (approved):
Nice refactor 🙂 LGTM after addressing the comments 👌

packages/realm/bindgen/src/templates/wrapper.ts
```diff
@@ -0,0 +1,405 @@
...
+import { BoundSpec, Class, Enum, Method, NamedType, Property, Struct, Type } from "@realm/bindgen/bound-model";
+import assert from "node:assert";
+import { eslintFormatter } from "../formatters";
+import { doJsPasses } from "../js-passes";
+
+function generateEnumDeclaration(e: Enum) {
+  return `export const enum ${e.jsName} { ${e.enumerators.map(({ jsName, value }) => `${jsName} = ${value}`)} };`;
+}
+
+const PRIMITIVES_MAPPING: Record<string, string | undefined> = {
```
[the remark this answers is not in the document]
kraenhansen: Because not all possible string properties result in a `string` upon lookup.

If omitted `PRIMITIVES_MAPPING["invalid"]` would have the type `string` when in fact it'll be `undefined`.

packages/realm/bindgen/src/templates/wrapper.ts
kraenhansen: This is copy-paste from the former `typescript.ts` template: https://github.com/realm/realm-js/blob/ff3d889df4bb1da1ae40074cbcbea8bf8410223f/packages/realm/bindgen/src/templates/typescript.ts#L135-L150

packages/realm/bindgen/src/templates/wrapper.ts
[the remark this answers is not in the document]
kraenhansen: This is missing because the eslint config is different for these files: https://github.com/realm/realm-js/blob/nh/wasm/emscripten_target/packages/realm/bindgen/.eslintrc.json

We could probably delete that file and commit the `--fix` but I'll defer that to another PR.

packages/realm/src/binding/NativeBigInt.ts
kraenhansen: I do see it mentioned here: https://github.com/realm/realm-js/blob/ff3d889df4bb1da1ae40074cbcbea8bf8410223f/packages/realm/bindgen/src/templates/typescript.ts#L201C32-L201C50 but I wonder if anything is relying on it or if it would be okay to simplify 🤔 I mean, what is the expected semantics here? For one thing, must the two things being compared match in type or will they get coerced?

elle-j: I think the ambiguity here is what type of equality our `equals()` method is supposed to be. The previous behavior mimics loose equality, so if it's a number, their numeric values are compared; whereas if it's a string, the string gets converted to a `BigInt`.

PR #6833, opened 2024-08-12
Fix realm-react progress tests

gagik:
## What, How & Why?
Minor fixes from previous PR feedback

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/src/indirect.ts
+++ b/packages/realm/src/indirect.ts
@@ -47,9 +47,14 @@ type Indirects = {
  */
 export const indirect = {} as Indirects;
 
+const IGNORED_PROPS = new Set([
+  // See https://github.com/realm/realm-js/issues/6522
+  "$$typeof",
+]);
+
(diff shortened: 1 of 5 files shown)


packages/realm/src/indirect.ts
```diff
@@ -47,9 +47,14 @@ type Indirects = {
  */
 export const indirect = {} as Indirects;
 
+const IGNORED_PROPS = new Set([
+  // See https://github.com/realm/realm-js/issues/6522
+  "$$typeof",
```
gagik, line 52: This was causing me issues before and existed pre-refactor; was the new refactor meant to resolve this? @kraenhansen

kraenhansen: I didn't think of that issue 👍 good catch.

kraenhansen (approved):

PR #6849, opened 2024-08-18
Update expo instructions

kadikraman:
### 1. use `npx expo install` instead of `npm install`
`npx expo install` will use your configured package manager, `npm install` will use `npm` so won't work if you're using a different package manager

### 2. update expo go instructions
The instructions said "consider using development build" which implies there are other ways to get it to work within Expo Go, which isn't the case. This adds a small tweak to make it clearer what to do next.


kadikraman:
@cla-bot check

kraenhansen (approved):
Thanks a lot @kadikraman!

PR #6922, opened 2024-11-06
Add option to exclude Realm file from iCloud backup

### LINKED ISSUE #4139: Add option to exclude realm from iCloud backup
### Problem
Some apps might want to exclude realm from the iCloud backup. Currently there is no configuration option to achieve this behavior. I am aware that there are manual ways to solve this, e.g. manipulating the resource keys of files created by realm. However, I can understand developers who would rather not make such adjustments, as they may not be able to fully understand the effects. E.g., Apple writes in the documentation about the corresponding [NSURLIsExcludedFromBackupKey](https://developer.apple.com/documentation/foundation/nsurlisexcludedfrombackupkey) parameter:

> Set this property each time you save a file because some common file operations cause this property to reset to false.

It also feels to me that this option should be configurable via Realm itself, since Realm is a database with its own synchronisation capabilities. React Native Async storage for example offers a [configuration option](https://react-native-async-storage.github.io/async-storage/docs/advanced/backup/) for this purpose. Also, I've seen that for some files there is already an exclusion rule defined (see [PR](https://github.com/realm/realm-cocoa/pull/5270/commits/ce8be5ed6aec5b0c73cd62ff55f9a62d0c97373d)). Maybe you can tie in here? 

### Solution
Provide a configuration parameter to exclude realm file from iCloud backup.

danibonilha:
## What, How & Why?
This PR adds a flag on Realm Config to exclude realm files from iCloud backup and also removes a reference from a guide that has been deleted.

This closes #4139

## Draft discussion
@kraenhansen I'm starting this as a draft to make sure I'm on the right path, and to align some things:

1.  In order to add this flag, some changes are required on @realm/core, to basically add `exclude_from_icloud_backup` on `spec.yml` and `shared_realm.hpp`, how is that coordinated ?

2.  You mentioned testing might be tricky for this one, it's a bit concerning if it's too difficult because I might not have the bandwidth to do it, but I can give it a try if there are some really similar examples.

3. This implementation is enough to ENABLE excluding from backup, but it cannot be DISABLED once a realm file has been set with  `excludedFromBackup`,  since we are just checking for `true` on the flag and then setting the `NSURLIsExcludedFromBackupKey` key, there's no check to remove this key when the flag is false again. Which means that, if a realm file has been created with `excludeFromIcloudBackup=false` and then set to `excludeFromIcloudBackup=true` the file will be correctly excluded, but if the same file if set again to `excludeFromIcloudBackup=false` it will remain excluded. I see two ways to handle this:   

 - 1. Stick with this implementation and explain the behavior via docs;
 - 2. Always check flag status and update files accordingly; 

- Particularly I don't think there's much usage of setting this flag dynamically, so i'd probably go with the first option since realm does provide an easy way to copy a file with different configs if needed, although the second one is more flexible, it might take longer to be done as I believe I'd need to always be checking if the files already have the excludedFromBackup key set. 

4. About the .md file on how to test this fix,  where would that live?

## Testing the fix
Generate the build files and test it on a sample react-native app on an IOS simulator, adding `excludeFromIcloudBackup: true` to the Realm configuration. 

To validate if the file has the exclude backup attribute, you need to check the files attributes with `xattr` on your terminal, in order to do so you need the path of realm files from the simulator Documents' folder, you can easily get that first by running:
```shell
open `xcrun simctl get_app_container booted com.your.app.bundleId data`/Documents
```
This will open a Finder window with the files, you can just drag and drop it to the terminal after adding `xattr`
 ```shell
 xattr [realm file simulator path]
 ```
 if this command returns:
 ```shell
 com.apple.metadata:com_apple_backup_excludeItem
 ```
It means the file has been successfully marked as excluded from backup 🎉 
if it returns nothing, it means the file has no attributes therefore is not excluded from backup. 

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/bindgen/js_opt_in_spec.yml
+++ b/packages/realm/bindgen/js_opt_in_spec.yml
@@ -51,6 +51,7 @@ records:
       - initialization_function
       - should_compact_on_launch_function
       - automatically_handle_backlinks_in_migrations
+      - exclude_from_icloud_backup
 
   UserIdentity:
     fields:
@@ -238,6 +239,7 @@ classes:
       - remove_file
       - remove_directory
       - get_cpu_arch
+      - after_realm_open
 
   WeakSyncSession:
     methods:
--- a/packages/realm/bindgen/js_opt_in_spec.yml
+++ b/packages/realm/bindgen/js_opt_in_spec.yml
@@ -51,7 +51,6 @@ records:
       - initialization_function
       - should_compact_on_launch_function
       - automatically_handle_backlinks_in_migrations
-      - exclude_from_icloud_backup
 
   UserIdentity:
     fields:
--- a/packages/realm/bindgen/js_spec.yml
+++ b/packages/realm/bindgen/js_spec.yml
@@ -15,6 +15,7 @@ classes:
       remove_file: '(path: const std::string&)'
       remove_directory: '(path: const std::string&)'
       get_cpu_arch: () -> std::string
+      after_realm_open: '(sharedRealm: std::shared_ptr<Realm>)'
       # print: (const char* fmt, ...) # can't expose varargs directly. Could expose a fixed overload.
 
   WeakSyncSession:
--- a/packages/realm/bindgen/js_spec.yml
+++ b/packages/realm/bindgen/js_spec.yml
@@ -15,7 +15,7 @@ classes:
       remove_file: '(path: const std::string&)'
       remove_directory: '(path: const std::string&)'
       get_cpu_arch: () -> std::string
-      after_realm_open: '(realm: SharedRealm)'
+      after_realm_open: '(realm: SharedRealm, exclude_from_icloud_backup: bool)'
       # print: (const char* fmt, ...) # can't expose varargs directly. Could expose a fixed overload.
 
   WeakSyncSession:
--- a/packages/realm/bindgen/js_spec.yml
+++ b/packages/realm/bindgen/js_spec.yml
@@ -15,7 +15,7 @@ classes:
       remove_file: '(path: const std::string&)'
       remove_directory: '(path: const std::string&)'
       get_cpu_arch: () -> std::string
-      after_realm_open: '(sharedRealm: std::shared_ptr<Realm>)'
+      after_realm_open: '(realm: SharedRealm)'
       # print: (const char* fmt, ...) # can't expose varargs directly. Could expose a fixed overload.
 
   WeakSyncSession:
--- a/packages/realm/binding/apple/platform.mm
+++ b/packages/realm/binding/apple/platform.mm
@@ -18,6 +18,7 @@
 
 #include "../platform.hpp"
 
+#include <memory>
 #include <realm/util/to_string.hpp>
 
 #include <stdarg.h>
@@ -33,6 +34,10 @@ static NSString *error_description(NSError *error) {
     return underlyingError.localizedDescription;
   }
   return error.localizedDescription;
+}
+
+static void RLMAddSkipBackupAttributeToItemAtPath(std::string_view path) {
+    [[NSURL fileURLWithPath:@(path.data())] setResourceValue:@YES forKey:NSURLIsExcludedFromBackupKey error:nil];
 }
 
 static std::string s_default_realm_directory;
@@ -156,6 +161,18 @@ void JsPlatformHelpers::remove_file(const std::string &path)
                                                       filePath.UTF8String, error_description(error).UTF8String));
         }
     }
+}
+
+void JsPlatformHelpers::after_realm_open(const std::shared_ptr<Realm> sharedRealm) {
+     if(sharedRealm->config().exclude_from_icloud_backup) {
+       RLMAddSkipBackupAttributeToItemAtPath(sharedRealm->config().path);
+       RLMAddSkipBackupAttributeToItemAtPath(sharedRealm->config().path +
+                                             ".lock");
+       RLMAddSkipBackupAttributeToItemAtPath(sharedRealm->config().path +
+                                             ".note");
+       RLMAddSkipBackupAttributeToItemAtPath(sharedRealm->config().path +
+                                                ".management");
+     }
 }
 
 void JsPlatformHelpers::remove_directory(const std::string &path)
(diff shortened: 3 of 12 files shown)


packages/realm/bindgen/js_opt_in_spec.yml
```diff
@@ -51,6 +51,7 @@ records:
       - initialization_function
       - should_compact_on_launch_function
       - automatically_handle_backlinks_in_migrations
+      - exclude_from_icloud_backup
```
kraenhansen, line 54: These values refer to fields of the `RealmConfig` defined by Realm Core: https://github.com/realm/realm-core/blob/43a1aac3dc41edf80a5be093a8a65e839b0ad225/src/realm/object-store/shared_realm.hpp#L94-L194

If it's important for this behavior to be configurable, an alternative that wouldn't require changes to Core would be pass this as a bool argument to `after_realm_open` 🤔 Or rename `after_realm_open` to `exclude_from_icloud_backup` and call it conditionally in `Realm.ts`.

packages/realm/bindgen/js_spec.yml
```diff
@@ -15,6 +15,7 @@ classes:
       remove_file: '(path: const std::string&)'
       remove_directory: '(path: const std::string&)'
       get_cpu_arch: () -> std::string
+      after_realm_open: '(sharedRealm: std::shared_ptr<Realm>)'
```
kraenhansen, line 18: I suggest naming this `realm` as all references to `Realm` classes in the codebase are to objects wrapped in `shared_ptr`. We also refer to these are `SharedRealm` (made possible by [this line in Realm Core's spec.yml](https://github.com/realm/realm-core/blob/85eeca41654cc9070ad81a21a4b1d153ad467c33/bindgen/spec.yml#L928-L929))

```suggestion
      after_realm_open: '(realm: SharedRealm)'
```

packages/realm/binding/apple/platform.mm
```diff
@@ -158,6 +163,18 @@
     }
 }
 
+void JsPlatformHelpers::after_realm_open(const std::shared_ptr<Realm> sharedRealm) {
+     if(sharedRealm->config().exclude_from_icloud_backup) {
+       RLMAddSkipBackupAttributeToItemAtPath(sharedRealm->config().path);
```
kraenhansen, line 168: Nit: It could make sense to introduce a local `const auto path = sharedRealm->config().path` to keep things DRY.

packages/realm/binding/platform.hpp
```diff
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <string>
+#include "realm/object-store/shared_realm.hpp"
```
kraenhansen, line 22: As this is referring to an include outside of the JS SDK code-base:

```suggestion
#include <realm/object-store/shared_realm.hpp>
```

kraenhansen:
> In order to add this flag, some changes are required on @realm/core, to basically add `exclude_from_icloud_backup` on `spec.yml` and `shared_realm.hpp`, how is that coordinated?

This can be done through a PR to https://github.com/realm/realm-core and and updating to the submodule of this repo to point to the PR commit in realm-core. But, since this won't be generally supported by the other SDKs I would rather that we keep changes to this codebase. A possible alternative is suggested in this comment: https://github.com/realm/realm-js/pull/6922#discussion_r1832256555. If you _really_ want this to be supported generally across SDKs, I'd say the logic calling into `setResourceValue` should live in Realm Core too.

> Always check flag status and update files accordingly

I like this idea 👍 I agree this would make it more flexible.

> it might take longer to be done as I believe I'd need to always be checking if the files already have the excludedFromBackup key set.

Would it throw if you set it when it's already set?

packages/realm/bindgen/js_opt_in_spec.yml
danibonilha: Oh at first I thought the only way to add parameters to the config was adding on realm-core, but I see now we can scope it just to realm-js, so I'll go with the parameter option.

packages/realm/binding/apple/platform.mm
```diff
@@ -18,6 +18,7 @@
 
 #include "../platform.hpp"
 
+#include <memory>
```
kraenhansen, line 21: Is this strictly needed?

kraenhansen (approved):

danibonilha:
> But, since this won't be generally supported by the other SDKs I would rather that we keep changes to this codebase.

Agreed, I've added a new implementation with a parameter 

>Would it throw if you set it when it's already set? If not, I wouldn't be too worried of setting it while it's already set, as we'd expect it to be called once per app open.

It does not throw, so I went ahead and update it to dynamically change based on the attribute

>Perhaps a guide-testing-exclude-icloud-backup.md

I've created this file, but also added a script to make it easier to test, I have created for me but I just figured it would be helpful to have it in this project

Oh I also updated the `ninja` requirement, as I was having trouble running for android and it was because this package was missing, so I added a step with the OpenJDK.

kraenhansen:
Awesome! I assume you've already verified this fixes the issue in your own (or some test) app?

packages/realm/binding/apple/platform.mm
danibonilha: I don't think so, it's probably just a leftover

danibonilha:
> Awesome! I assume you've already verified this fixes the issue in your own (or some test) app?

Yes I have, it's working perfectly on my app: 

`excludeFromIcloudBackup: true`

image


`excludeFromIcloudBackup: false `

image

I'm having issues though running the android app, but it might be my environment or something, is there a check on the CI that builds the android sample app ? Since there's no implementation there, just having it the app building should be enough.

kraenhansen:
I reworded the note in the changelog and I ran the tests locally on Android, iOS and Node.js 👍 I'll go ahead with the merge and release of this 👍

kraenhansen:
I'm seeing some compilation errors on Node.js for Windows that I need to resolve before releasing this 😕 Stay tuned.

danibonilha:
> I'm seeing some compilation errors on Node.js for Windows that I need to resolve before releasing this 😕 Stay tuned.

Could it be related maybe to using unnamed parameters on platform.cpp ?

kraenhansen:
Sorry about the wait on this one. Would you mind taking a look and providing a review on https://github.com/realm/realm-js/pull/6927?

PR #6943, opened 2024-12-06
Nested unmanaged types

### LINKED ISSUE #6037: Type issues with Unmanaged<T, RequiredProperties>

Always

### Description

When using Realm classes, we noticed a few type issues all related to `Unmanaged<T, RequiredKeys>`. 

1. When specifying the `RequiredProperties` in `Realm.Object<T, RequiredProperties>`, the properties of type Realm.List or Realm.Dictionary are not allowed. 
2. If `RequiredProperties` is not provided when defining the class, the constructor accepts string as values. 
3. Unmanaged nested Realm.Object is not allowed in `Realm.create`. It's also reported in #5594

The details are available in the mini repo I attached. 

If anyone is interested, I made a patch to resolve the issue 1 & 3 for now. https://gist.github.com/Yupeng-li/655eaf75f46086c4345a78e27dfa3c5e

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

Please check this [codesandbox mini repo](https://codesandbox.io/s/thirsty-cherry-2d3s8d?file=/src/App.tsx) for details. (Please ignore the browser view on the right hand side. ) 

### Version

11.7 & 11.9

Local Database only

### Are you using encryption?

Yes

All

### Build environment

### Cocoapods version

_No response_

liamjones:
## What, How & Why?
Ensures `Unmanaged` works when a `Realm.Object` contains other `Realm.Objects` (or Lists, etc of them).

This closes #6037

## ☑️ ToDos
* [ ] 👨‍💻 Expand checks to existing things that already work; List, Dictionary, Set & Counter
* [ ] ❓ Why did the Counter PR add `ExtractPropertyNamesOfTypeExcludingNullability` instead of using `ExtractPropertyNamesOfType`?
* [ ] 👨‍💻 Expand checks to include things that don't currently work; nested Object, List<Object>, etc so tests fail
* [ ] 👨‍💻 Update Unmanaged so that it supports the new things and tests pass again
* [ ] ❓ Should `type:tests` be run as part of a higher level `build` or `test`?
* [ ] ❗ String still accepted as `values`, matches `Unmanaged<DefaultObject, never>`?
* [ ] ❓ If a Realm.Object has required types shouldn't they be automatically required in the Unmanaged version during `realm.create()` too? Is this possible or will there be a circular type reference?
* [ ] 📝 Changelog entry
* [ ] 🚦 Tests

### DIFF
--- a/packages/realm/package.json
+++ b/packages/realm/package.json
@@ -73,7 +73,7 @@
   ],
   "scripts": {
     "test": "wireit",
-    "test:types": "tsc --project type-tests/tsconfig.json",
+    "test:types": "wireit",
     "lint": "eslint --ext .js,.mjs,.ts .",
     "prebuild": "tsx ./src/scripts/build/cli.ts",
     "prebuild-apple": "wireit",
@@ -101,6 +101,12 @@
       "env": {
         "TSX_TSCONFIG_PATH": "tsconfig.tests.json"
       }
+    },
+    "test:types": {
+      "command": "tsc --project type-tests/tsconfig.json",
+      "dependencies": [
+        "build:ts"
+      ]
     },
     "prebuild-apple": {
       "command": "tsx ./src/scripts/build/cli.ts build-apple",
--- a/packages/realm/type-tests/nested-unmanaged-tests.ts
+++ b/packages/realm/type-tests/nested-unmanaged-tests.ts
@@ -1,0 +1,107 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2024 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+/* eslint-disable @typescript-eslint/no-unused-vars -- We're just testing types */
+
+// This file checks that the Unmanaged type re-mappings work correctly for Realm types, especially when there are nested classes
+
+import Realm from "realm";
+
+class RealmClassWithRequiredParams extends Realm.Object<
+  RealmClassWithRequiredParams,
+  "aMandatoryString" | "anOptionalString"
+> {
+  public aMandatoryString!: Realm.Types.String;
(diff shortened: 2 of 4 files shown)


liamjones:
Hi @kraenhansen,

This PR has been slow going due to other work cropping up but I thought I might as well get a draft PR up so I can start discussing bits of it with the team.

There's a couple of TODOs you might be able to help with immediately:

I've got `type:tests` working again in a sensible manner (I think?): https://github.com/realm/realm-js/pull/6943/commits/b0be3468dfa9e78b52ec5a61ac73109eaaebc460. The question; should `type:tests` be run as part of a higher-level `build` or `test` command rather than only being run directly? Maybe it should just be an extra [pr-lint check command](https://github.com/realm/realm-js/blob/05deac9b58143f654698d06f8766a840ae9af2c8/.github/workflows/pr-linting.yml#L22)? 

The second query, and this might be one for the original PR author instead, do you know why the `Counter` PR added `ExtractPropertyNamesOfTypeExcludingNullability` instead of using `ExtractPropertyNamesOfType`? It wasn't immediately obvious to me.

liamjones:
Unfortunately, I won't have time to finish this properly right now.

I'm stuck with an issue whereby `RequiredProperties` in `Unmanaged` is complaining that "Type string is not assignable to type Exclude<ExtractPropertyNamesOfTypeExcludingNullability<T, Counter>, keyof AnyRealmObject | ExtractPropertyNamesOfType<...> | ExtractPropertyNamesOfTypeExcludingNullability<...>> | ... 5 more ... | Exclude<...>"": https://github.com/realm/realm-js/compare/main...living-with:realm-js:nested-unmanaged-types

Since we `skipLibCheck` in our ts-config, this doesn't overly affect our usage, but it's not in a good publishable state for the project itself.

liamjones:
@kraenhansen As I'm currently stuck with getting the PR into a mergeable state I'm patching the types in a fork for our purposes (https://github.com/realm/realm-js/compare/main...living-with:realm-js:nested-unmanaged-types) and then transferring the changes into our project using [patch-package](https://github.com/ds300/patch-package).

My fork's commits are on top of the community/v20.1.0 tag and I'm using `npm test --workspace @realm/node-tests` to check things are good + produce the `d.ts` files.

One niggle I'm hitting, the build still seems to produce device sync info in the types but the `realm` types inside my project's node modules from v20.1.0 don't seem to contain that anymore so I'm getting more difference between the generated types than I expect (fork on the left, realm 20.1.0 inside a project's node_modules on the right):

image

How do I generate the type files without the device sync entries?

kraenhansen:
> How do I generate the type files without the device sync entries?

These types were removed from v20 as part of [the deprecation of device sync](https://github.com/realm/realm-js/blob/main/DEPRECATION.md), so any v20 release will have those removed.

liamjones:
Oh, I'm an idiot, I thought I branched from `community` but I branched from `main`. 🫣 Thank you.

PR #6907, opened 2024-10-22
fix: cannot read property size of undefined (#6626)

MichelGutner:
## What, How & Why?

I fixed the TypeError: Cannot read property 'size' of undefined, which occurred due to attempting to access the size property of an undefined object. I added a check to ensure the object exists before accessing its properties to prevent this error.

This closes #6626

## ☑️ ToDos

* [ ] 📝 Changelog entry
* [ ] 📝 `Compatibility` label is updated or copied from previous entry
* [ ] 📝 Update `COMPATIBILITY.md`
* [ ] 🚦 Tests
* [ ] 📦 Updated internal package version in consuming `package.json`s (if updating internal packages)
* [ ] 📱 Check the React Native/other sample apps work if necessary
* [ ] 💥 `Breaking` label has been applied or is not necessary
* [ ] 🔔 Mention `@realm/devdocs` if documentation changes are needed

PR #6955, opened 2025-01-08
fix(realm/Results): provide fallback condition

### LINKED ISSUE #6954: [React Native] TypeError: Cannot read property 'size' of undefined

Always

### Description

The app crashes on fresh install because of a problem in the [Results.js](node_modules/realm/dist/Results.js) file.

### Stacktrace & log output

```shell
Warning: TypeError: Cannot read property 'size' of undefined
```

Always

### Reproduction Steps

Our project was built using:
- react-native: 0.71.2
- realm: 11.9.0
- @realm/react: 0.4.3

And had two schemas with no relation between them.

After upgrading to the latest versions of all the packages, the app crashed on runtime giving the following output:
`Warning: TypeError: Cannot read property 'size' of undefined`.

The issue has been solved by going to the Results.js file in the dist folder and updating the getter `length` to this:
`
get length() {
        return this.internal ? this.internal.size() : 0;
    }
`

### Version

12.14.0

Local Database only

### Are you using encryption?

Yes

iOS and Android, all versions

### Build environment

- "react-native": "^0.76.2"
-  "@react-native/metro-config": "0.76.5"
- "@realm/react": "^0.11.0"
- "realm": "^12.14.0"
- Node version: 18.17.1
- Android env:
  - ndkVersion: "26.1.10909125"
  - kotlinVersion: "1.9.24"
  - buildToolsVersion = "35.0.0"
  - minSdkVersion = 24
  - compileSdkVersion = 35
  - targetSdkVersion = 34
- iOS env:
  - Minimum version: 13.0
  - Swift version: 5.0 
  - Cocoapods version: 1.15.2

### Cocoapods version

1.15.2

guillemsarda:
## What, How & Why?
Small change in the getter `length` from Results' class.
This solves an issue that causes an app crash when upgrading from react-native: 0.71.2, realm: 11.9.0 and @realm/react: 0.4.3 to the corresponding latest versions.
The issue is solved by providing a fallback condition in the getter (i.e., when `this.internal` is not defined).

This closes #6954

### DIFF
--- a/packages/realm/src/Results.ts
+++ b/packages/realm/src/Results.ts
@@ -93,7 +93,7 @@ export class Results<T = unknown> extends OrderedCollection<
   }
 
   get length() {
-    return this.internal ? this.internal.size() : 0;
+    return this.internal?.size() ?? 0;
   }
 
   set length(value: number) {
--- a/packages/realm/src/Results.ts
+++ b/packages/realm/src/Results.ts
@@ -92,8 +92,8 @@ export class Results<T = unknown> extends OrderedCollection<
     throw new Error("Modifying a Results collection is not supported.");
   }
 
-  get length(): number {
-    return this.internal.size();
+  get length() {
+    return this.internal ? this.internal.size() : 0;
   }
 
   set length(value: number) {


packages/realm/src/Results.ts
```diff
@@ -92,8 +92,8 @@ export class Results<T = unknown> extends OrderedCollection<
     throw new Error("Modifying a Results collection is not supported.");
   }
 
-  get length(): number {
-    return this.internal.size();
+  get length() {
+    return this.internal ? this.internal.size() : 0;
```
albertms10, line 96: Maybe briefer: 😄 

```suggestion
    return this.internal?.size() ?? 0;
```

guillemsarda: That would work, too!

kraenhansen:
Thanks for your interest in solving this. We've had other users experiencing this issue and [another PR](https://github.com/realm/realm-js/pull/6907) trying to fix it.

It is however an invariant of the Results to have an `internal` value, so instead of handling the possibility of this internal going missing, the proper solution would be to figure out why this internal doesn't have a value in the first place and fix that.

This pattern of `internal` objects are something we're using lots of places and I don't want to merge a fix for a problem with an undetermined root cause.

PR #6977, opened 2025-02-19
Adds failing test case for issue #6889 + fix

### LINKED ISSUE #6889: Bug on nested array linking objects update, (bug not happening on Realm 12.2.1)

Always

### Description

So we have to update to RN 0.74.5 (from 0.72) and we were on 12.2.1 realm version. Unfortunately, keeping same realm version, makes the app block on white screen, and through logcat I can see it's just saying " Building an exports object", but never proceeds. 
So, I tried with 12.12.1 and 12.13.1 realm versions, which do not have this problem and everything goes smoothly, except from one use case.

I have a parent object `Participant` which has a subTable array of `ParticipantMeta` objects. `ParticipantMeta` are linking objects to the parent `Participant`.

When i create a Participant object with some metas, it's correctly added on the db.
When i query it, and re-add it using `realm.create` with `Realm.UpdateMode.ALL`, without any change at all, the meta list array becomes empty.

I have created a reproducible "minimal" repo that you can see on the logs the problem.

Update to RN 0.74 is crucial as we need to target api 34, and be able to release on playstore. Currently, this is our only blocker.

### Stacktrace & log output

```shell
no crash. Only from the logs we can see that before updating the object we have full array and after the update, the array of the subTable is emptied.
```

Always

### Reproduction Steps
https://github.com/angelos3lex/Realm-js-12.13.1-bug-repro
Check this repo ^
(Mainly `App.tsx` `onPress` of the button is where the whole logic begins.)

_PS: Didn't test the reproducible example on iOS, but on my real case app it's also happening on iOS._

### Version

12.12.1, 12.13.1

Local Database only

### Are you using encryption?

No

android, iOS, realm 12.12+, RN 0.74.5

### Build environment

no debugger

### Cocoapods version

didn't test the reproducible example on iOS, but on my real case app it's also happening on iOS.

angelos3lex:
#6889 issue, failing test ticket

If you remove lines 664-666, (the part with updating the managed object):
```
this.realm.write(() => {
    this.realm.create(PersonWithId, john, UpdateMode.All);
});
```

It works well. So the problem is when you update a managed object, it somehow loses the collection.

Update: Pushed a [commit](https://github.com/realm/realm-js/pull/6977/commits/625c997a1b3f1623533689fa373af2b0eae7ea4a) that fixes the issue.

### DIFF
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -634,6 +634,45 @@ describe("Realm.Object", () => {
         expect(persons.length).equals(1);
       });
 
+      describe("with collection", () => {
+        it("can be created, fetched, updated and refetched without affecting collection", function (this: Mocha.Context &
+          RealmContext) {
+          const _id = new Realm.BSON.ObjectId();
+
+          const friend = {
+            age: 18,
+            name: "Friend's Name",
+            _id: new Realm.BSON.ObjectId(),
+          };
+
+          this.realm.write(() => {
+            this.realm.create(
+              PersonWithId,
+              {
+                _id,
+                name: "John Doe",
+                age: 42,
+                friends: [friend],
+              },
+              UpdateMode.All,
+            );
+          });
+
+          const john = this.realm.objectForPrimaryKey(PersonWithId, _id);
+          if (!john) throw new Error("Object not found");
+
+          this.realm.write(() => {
+            this.realm.create(PersonWithId, john, UpdateMode.All);
+          });
+
+          const johnAfterUpdate = this.realm.objectForPrimaryKey(PersonWithId, _id);
+          if (!johnAfterUpdate) throw new Error("Object not found");
+
+          expect(john.friends.length).equals(1);
+          expect(johnAfterUpdate.friends.length).equals(1);
+        });
+      });
+
       describe("applying 'UpdateMode' recursively", () => {
         let aliceId: Realm.BSON.ObjectId;
         let bobId: Realm.BSON.ObjectId;
--- a/integration-tests/tests/src/tests/objects.ts
+++ b/integration-tests/tests/src/tests/objects.ts
@@ -637,36 +637,27 @@ describe("Realm.Object", () => {
       describe("with collection", () => {
         it("can be created, fetched, updated and refetched without affecting collection", function (this: Mocha.Context &
           RealmContext) {
-          const _id = new Realm.BSON.ObjectId();
-
-          const friend = {
-            age: 18,
-            name: "Friend's Name",
-            _id: new Realm.BSON.ObjectId(),
-          };
-
-          this.realm.write(() => {
+          const john = this.realm.write(() =>
             this.realm.create(
               PersonWithId,
               {
-                _id,
+                _id: new Realm.BSON.ObjectId(),
                 name: "John Doe",
                 age: 42,
-                friends: [friend],
+                friends: [
+                  {
+                    age: 18,
+                    name: "Friend's Name",
+                    _id: new Realm.BSON.ObjectId(),
+                  },
+                ],
               },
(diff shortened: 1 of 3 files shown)


angelos3lex:
@cla-bot check

integration-tests/tests/src/tests/objects.ts
```diff
@@ -634,6 +634,45 @@ describe("Realm.Object", () => {
...
+              },
+              UpdateMode.All,
+            );
+          });
+
+          const john = this.realm.objectForPrimaryKey(PersonWithId, _id);
+          if (!john) throw new Error("Object not found");
+
+          this.realm.write(() => {
+            this.realm.create(PersonWithId, john, UpdateMode.All);
```
kraenhansen, line 665: While I agree this should be a possible invocation of the API, can you help me understand the use-case for it? When would you ever call the `create` and pass an existing object directly? What would be the purpose of that?

I'm asking because I suspect there might be other (less synthetic) ways of hitting the same (or similar) bug, which won't be tested here.

angelos3lex: @kraenhansen In our case, we get the child object through db, inside a write transaction we change a field of it, and we then want to trigger an update on the parent. So we query the parent, and perform a change in a field of it, then call realm.create, which ends up loosing subLists that we didn't even change in any transaction.

We need to update the whole object because that's required by our design, changing a child must trigger a whole update on a parent so that the parent updates some `localUpdatedOn` etc things we need.

In the example above, I made no change for simplicity, but we could for example change:
```
this.realm.write(() => {
    john.friends[0].age = 20;
    this.realm.create(PersonWithId, john, UpdateMode.All);
});
```
which would result in loosing the friends sublist.

I understand that we are not required to perform realm.create since we already change (inside a write transaction) the fields we desire, but due to our architecture, we reuse some functions that work either for managed or unmanaged objects, an then the problem arises when we do it for managed objects.

kraenhansen: Do I understand this correctly? `john` might or might not be managed and to keep things simply, you pass it to `realm.create` in any case?

> We need to update the whole object because that's required by our design, changing a child must trigger a whole update on a parent

You might want to benefit from passing a second "keypaths" argument to the `addListener` method to have changes to `friends` fire events on the "parent":

```javascript
const realm = new Realm({ schema: [{ name: "Person", properties: { name: "string", friends: "Person[]" } }] });
const alice = realm.write(() => realm.create("Person", { name: "Alice", friends: [{ name: "Bob" }] }));

alice.addListener(() => {
  console.log("Alice changed!", alice);
}, "friends.*");

setTimeout(() => {
  console.log("Update!");
  realm.write(() => {
    alice.friends[0].name = "Charlie";
  });
}, 1000);
```

yields

```
Alice changed! Person {
  name: 'Alice',
  friends: Realm.List [ Person { name: 'Bob', friends: [Realm.List] } ]
}
Update!
Alice changed! Person {
  name: 'Alice',
  friends: Realm.List [ Person { name: 'Charlie', friends: [Realm.List] } ]
}
```

packages/realm/src/property-accessors/Array.ts
```diff
@@ -87,11 +87,12 @@ export function createArrayPropertyAccessor({
         assert.inTransaction(realm);
         assert.iterable(values);
 
+        const valuesCopiedRef = Array.from(values);
```
kraenhansen: Using `Array.from` will eagerly consume the `values` iterator, which might be okay, but depending on the size of the collection, it could cause unnecessary memory pressure. I suggest using a snapshot instead.

kraenhansen:
Tests (finally) passed over in https://github.com/realm/realm-js/pull/6985, making this PR is good to merge 🎉  Thanks a lot for your patience on this one.

